PR #38, opened 2015-01-03
Update to Rust nightly 2015-01-02

ayosec:
* Use `#[derive]` instead of `#[deriving]`
* Load `std::c_str::ToCStr` to be able to use `to_c_str`
* `iter::repeat(0).take(..).collect()` instead of `Vec::from_elem(0, ..)`

Tested with `rustc 0.13.0-nightly (c89417130 2015-01-02 21:56:13 +0000)`

There are some warnings about `deriving`, but it seems that they are out of this repository:

```
<std macros>:5:11: 5:71 warning: `deriving` is deprecated; use `derive`
<std macros>:5         #[deriving(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
```

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -8,7 +8,7 @@ pub use self::consts::Errno::*;
 
 pub type SysResult<T> = Result<T, SysError>;
 
-#[deriving(Clone, PartialEq, Copy)]
+#[derive(Clone, PartialEq, Copy)]
 pub struct SysError {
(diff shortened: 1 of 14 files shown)


carllerche:

ayosec:
> I think this has already been handled by #36.

My bad. I didn't check if there was some other PR with the same patch.

PR #40, opened 2015-01-03
replaced deriving with derive and added use std::c_str::ToCStr where nec...

MarkusJais:
replaced deriving with derive and added use std::c_str::ToCStr where necessary. now compiles again against latest Rust version.


carllerche:

MarkusJais:
Only saw #36 after I submitted mine :-)
nix-rust is cool, keep up the great work!

PR #41, opened 2015-01-04
updates for latest rust changes. (derive, ToCStr)

rrichardson:


carllerche:

PR #43, opened 2015-01-06
Fix breakage caused by the removal of unboxed closures

fhartwig:
Unboxed closures have been removed from the language, so this change is needed to compile with the current rust nightly.

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -62,7 +62,7 @@ mod cpuset_attribs {
     }
 }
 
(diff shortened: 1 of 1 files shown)


fhartwig:
I somehow managed to forget apostrophe and for some reason it compiled on my machine. This is now fixed.

carllerche:
Thanks for the PR! I rebased and merged it 7321706cb976cd39204e90bb6fa246195560d653

PR #44, opened 2015-01-07
ToCStr removal fallout

vhbit:
- ToCStr removal
- split of `Show` and `String` traits 
- `globs` feature is enabled by default


vhbit:
Note: works with master but changes haven't yet hit nightly

rrichardson:
Apparently the removal of #[feature(globs)]  caused it to break in Travis, did globs become non feature gated?

vhbit:
Yes, although it didn't yet hit nightly

PR #47, opened 2015-01-10
changed uint and int to usize and isize respectively

rrichardson:
OK. I used a much smarted regex for sed this time.


carllerche:
Committed here: 380d4020f10583797fe7bab6fe95923934660605

PR #48, opened 2015-01-16
Add `termios` bindings.

utkarshkukreti:
This includes almost all constants/functions from `/usr/include/sys/termios.h` on OSX 10.9.5. Looking for feedback on the naming/uses of `bitflags!` and `enum`s.

For the record, I used [this script][0] I wrote to get the values of these constants.

[0]: https://gist.github.com/5703c8ae132569420fce

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -14,6 +14,8 @@ pub mod socket;
 
 pub mod stat;
 
+pub mod termios;
+
 #[cfg(target_os = "linux")]
 pub mod utsname;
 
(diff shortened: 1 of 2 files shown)


carllerche:
Looking good. Let me know when this can be merged (I assume the WIP implies that it is still a work in progress).

utkarshkukreti:
I modified `tcgetattr` a bit to make it possible to use it without unsafe code (like [sys::stat::stat](https://github.com/carllerche/nix-rust/blob/b31eee4aa55623ebf8d47b6ccde5302a86ba68a4/src/sys/stat.rs#L57)) in https://github.com/carllerche/nix-rust/commit/a757e4ae9b07ff198a4067a8e871c57eeceb3163. What do you think? If that's okay, this is ready to be merged.

carllerche:
Thanks. I merged this da900c6932254fa5e3b2a44b3ac9160fd275d261

PR #55, opened 2015-01-27
Fix warnings + remove `std::io` dependency.

utkarshkukreti:


carllerche:
Thanks! Always appreciated :heart:

PR #56, opened 2015-01-30
Replace all uses of `Path` with `CString`.

utkarshkukreti:
All the `Path`s were being converted to `CString`s on every call to any
function that accepted them, which is an unnecessary allocation (and the
possibility of a panic which should probably happen outside the
functions in this crate) which can be easily avoided now.

You can fix existing code by converting `Path`s to `CString` using
something like:

    CString::from_slice(path.container_as_bytes())

and then passing that to the function(s).

Or just not create `Path`s and pass a `&[u8]` to `CString::from_slice`
directly.

    foo(&Path::new("..."))

to

    foo(&CString::from_slice(b"..."))

[breaking-change]


carllerche:
I'm somewhat torn. On one hand, I agree w/ all the points that you mention. On the other, the goal of nix is to provide "rusty" interfaces to the posix APIs.

I'm wondering if another option may be to implement a Path like type that is tailored for interfacing w/ the posix C apis and also doesn't have the problems you mentioned. I'm not actually convinced that this is the right strategy, but perhaps you have thoughts?

utkarshkukreti:
Would that new Path-like type basically be a newtype for `CString`, or something more than that?

carllerche:
It could provide the conveniences of Path as well... unsure.

The argument that you have is that you want to avoid the allocation, which means that you need to be manipulating CStrings in the first place, no?

utkarshkukreti:
Yes, I don't think we need to store anything more than a `CString` in that new `Path` type.

If we want a more "rusty" API, we could even add some methods to it e.g.

    let path = Path::new("foo.txt");
    path.open(...);
    path.stat();
    path.chmod(...);
    path.chown(...);

but that may be out of nix's scope.. not sure.

carllerche:
I have thought about this some more. Here is my suggestion.

Create a `NixPath` trait that is implemented for `Path`, `CString`, `&[u8]`, etc... The only thing it does is implement a function like:

```rust
fn with_nix_path<T, F: FnOnce(&[u8]) -> T>(&self, f: F) -> Result<T, PathError>
```

Which would return the ptr to the c string, null terminated. The implementation of with_nix_path would be pretty straight forward for `&[u8]` (it would just pass self. For `Path`, it could be implemented without an allocation. Stack allocate PATH_MAX (would have to figure out how much this is on various platforms), and copy the bytes in, appending null. It could also be implemented without a panic, by scanning for a null within the body of the path and returning a `Err`.

Thoughts?

utkarshkukreti:
That's much better than my proposal!

How would we handle returning `PathError` from the functions which return a `SysResult`?

carllerche:
Is there an invalid argument error perhaps? If not, we could possibly add an extra variant to handle errors that are bubbled up from nix vs. syscalls.

utkarshkukreti:
There's `EINVAL` for `Invalid argument` and `ENOENT` for `No such file or directory`, but I don't think we should return anything other than the actual value of `errno`..

carllerche:
What do you think we should do?

utkarshkukreti:
How about

```
pub enum Error {
  Sys(Errno), // or SysCall
  InvalidString // or InvalidPath or anything
}
```

and return `Result<T, Error>` from all functions?

There are only 2 functions I could find that can panic, and 1 that allocates: `execve` allocates 2 `Vec`s which could panic, and `eventfd` asserts the presence of some global function. We can (probably) fix the `execve` allocation as the sizes of the args and env must be less than the stack size limit, and I'm sure something can be done about `eventfd` too.

If we do all this, nix will have no allocations and no panics.

carllerche:
I don't think `Error` is the right name since `Error` from std will become a thing, but this general strategy looks good. I would probably use `InvalidPath` and maybe call it `NixError`?

I think we are in agreement for the strategy. I am looking forward to a PR :smile:

utkarshkukreti:
`NixError` and `InvalidPath` sound good. I'm working on a PR :)

utkarshkukreti:
> `fn with_nix_path<T, F: FnOnce(&[u8]) -> T>(&self, f: F) -> Result<T, PathError>`

Did you mean `F: FnOnce(*const c_char) -> T` here?

carllerche:
Either way doesn't matter since one transmutes to the other. So, which ever one you want :smile:

utkarshkukreti:
@carllerche does this look like it's going in the right direction: https://gist.github.com/464d19a5bb305ea4098c? I still need to shuffle around some functions and test on Linux.

carllerche:
Looks good to me so far :+1:

utkarshkukreti:
Submitted a new PR #61.

PR #57, opened 2015-02-01
Added NONE constant to WaitPidFlag and added  unit test for fork

MarkusJais:
I've added a small unit test for the fork function. I've also added "NONE" to waitpid flags so that it is possible to use it without an option (e.g. like passing 0 in the C world).

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -11,6 +11,7 @@ mod ffi {
 
 bitflags!(
     flags WaitPidFlag: c_int {
+        const NONE = 0x00000000,
         const WNOHANG = 0x00000001,
     }
 );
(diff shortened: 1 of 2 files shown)


utkarshkukreti:
Since `NONE` is not defined in any C headers, and passing `0` to seems to be a special case, maybe we should make `waitpid` it accept an `Option<WaitPidFlag>`?

MarkusJais:
Good point. Option<WaitPidFlag> sounds like a good solution.

carllerche:
+1 for Option<WaitPidFlag>

MarkusJais:
Updated the Pull Request. waitpid now takes an Option<WaitPidFlag>

carllerche:
Thanks, merged 9a17a4d200b33f66a6d61dc06a9df7ddef01d735

PR #59, opened 2015-02-06
path -> old_path

fhartwig:
The build is currently broken on nightly due to the introduction of the new Path API (which doesn't yet support conversion to `&[u8]` as far as I can see).
This fixes the build by continuing to use the old Path API.
(This should become unnecessary if #56 is merged).

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1,4 +1,4 @@
-use std::path::Path;
+use std::old_path::Path;
 use libc::{c_int, mode_t};
 use errno::{SysResult, SysError};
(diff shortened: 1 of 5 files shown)


carllerche:
@utkarshkukreti thoughts since you are working on #56?

fhartwig:
Sorry, I missed at least one module due to OS-conditional compilation. Will push a fix in a few minutes.

fhartwig:
Ok, seems to be fine now. Next time I'll check my changes on linux, promise :-)

carllerche:
Thanks for the PR

PR #60, opened 2015-02-06
added test for stat function

MarkusJais:
I've added a small test for the stat function.

For some tests I can only test that the value is a positive integer as some values would be different on each developers machine (e.g. user id, inode number, etc).

I've also added an import for the old posix::Path as this is currently needed for the File type. It will be removed as soon as the new io module is more stable and can create files.

If the stat call returns Err I panic! as this indicates a serious problem on that machine.

What do you think?

I will add more unit tests in the future.


MarkusJais:
Seems to break as std:.path:Path has now "as_vec" method. I will look into this.

MarkusJais:
One way to fix the error in utils.rs would be:

```rust
impl ToCStr for Path {
    fn to_c_str(&self) -> CString {
        CString::from_slice(self.to_str().unwrap_or("").as_bytes())
    }
}
```

Would this be an option or are there better and more efficient ways to do that?

carllerche:
I think this is a recent breakage in rust. However, you should look at #61 which is working to solve the Path problem in a better way. Besides that, looks good. Once it passes CI I will merge :smile:

MarkusJais:
OK, will get the latest version and look at #61.

tests/stat.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        use std::old_io::{File, Open, ReadWrite};
+        use std::old_io::fs::{unlink};
+        use std::old_path::posix::Path as OldPath; // temporary until new File type is available that takes std::path::Path
+        use std::path::Path as NewPath;
+
+        let filename = "target/foo.txt";
+        let old_io_path = OldPath::new(filename);  // for creating File
+
+        let test_file = File::create(&old_io_path);
```
utkarshkukreti: Maybe you should use the functions present in nix itself to create and unlink files?

MarkusJais: Will do.

MarkusJais:
Now it builds.

I've added the unlink function to  unistd. This allowed me to use only nix functions to create, close and remove the test file for the stat function.

I've added asserts to the unlink, close and open functions to make sure everything is fine with creating and removing the file.

I will add more unit tests for the functions in unistd in the coming days.

Hope this is ok like this. More will follow :-)

tests/stat.rs
```diff
@@ -0,0 +1,43 @@
...
+        use nix::fcntl::open;
+        use nix::unistd::{close, unlink};
+        use nix::fcntl::O_CREAT;
+        use nix::sys::stat::S_IWUSR;
+
+        let filename = b"target/foo.txt";
+        let fd_res = open(filename, O_CREAT, S_IWUSR);
+        assert!(fd_res.is_ok()); // creating a file in "target" should always work
+        let close_res = close(fd_res.ok().unwrap()); // close right here. We use the file only for the stat test
+        assert!(close_res.is_ok());
```
utkarshkukreti: You can also use `Result::unwrap()` here. It'll do the unwrap + assert in one step and even print the `Err` value if it's an `Err`.

```rust
let fd_res = open(...).unwrap();
let close_res = close(fd_res).unwrap();
```

MarkusJais:
Followed Utkarsh's suggestion and used unwrap. This allowed me to get rid of the not really necessary variables without compiler warnings about unused code variables or return values.

carllerche:
Thanks! Squashed & merged: f898b825566ab0738dbfa3a1a45010d37060aed0

PR #61, opened 2015-02-07
Add `NixPath`, `NixError`, and `NixResult`.

utkarshkukreti:
This needs a very careful review as most functions don't have any tests, so I don't know if I've broken anything. We're just a couple of steps away from removing all `std` dependencies. :smile:


src/mount.rs
```diff
@@ -68,42 +66,47 @@ mod ffi {
...
-            fstype.as_ref().map(|s| s.as_ptr()).unwrap_or(ptr::null()),
-            flags.bits,
-            data.map(|s| s.as_ptr() as *const c_void).unwrap_or(ptr::null()))
-    };
-
-    from_ffi(res)
+        data: Option<P4>) -> NixResult<()> {
+    use libc;
+
+    let res = try!(try!(try!(try!(
```
utkarshkukreti: Any ideas to improve this?

src/nix.rs
```diff
@@ -0,0 +1,49 @@
+use libc::c_char;
```
utkarshkukreti: Any ideas for a better file name?

MarkusJais: libc::c_char sounds good to me if you want to indicate that it is a C char.

MarkusJais:
Just curious: Why remove all std dependencies?

src/mount.rs
carllerche: ewww yeah... there probably is a way to clean it up, but I wouldn't be too worried about it ATM. This is ok for now

src/nix.rs
carllerche: You could also put this stuff in `lib.rs` but I don't mind `nix.rs` either.

carllerche:
Looks good to me. Could you smash it into one commit?

The lack of tests is definitely sucky at this point. I originally started this as a dumping ground for all the bindings I had spread out in various libs... but it has grown to the point where it really should have tests.

I will want to coordinate merging this PR w/ updating mio at least (I can do the work, just let me know when it is ready to go).

Finally, I wouldn't worry too much about removing the `std` dep. If it's easy to do, seems ok, but the only real time that not depending on `std` is useful is when writing a kernel (or something similar). Since this binds OS apis, this isn't a requirement.

carllerche:
Also, I'm going to apply, PRs, so could you rebase? :smile:

src/nix.rs
utkarshkukreti: As we're already exporting the contents of this module on the top level, I've made this module private..

utkarshkukreti:
I shuffled some functions around in `errno` module. Let me know if everything looks good, I'll squash all the commits so you can merge it.

carllerche:
Seems good, squash it. I will try to start updating mio to coordinate the change.

utkarshkukreti:
Done!

carllerche:
Looking good. I will try updating mio tomorrow and then merge everything. Thanks!

utkarshkukreti:
(I added some more `derive`s to `NixError` and force pushed.)

carllerche:
Merged. Thanks (effb423fdb26843612edca6a748e479522089c1d)

utkarshkukreti:
@MarkusJais Regarding the std dependency: I want to use nix as a safe alternative to calling the unsafe `libc` functions directly in a `no_std` project. Although I only need to use about a dozen or so functions, which I can copy paste into my project, if it's not too hard to remove the dependency in nix, I think that it would be a good idea.

PR #64, opened 2015-02-11
Add tests for `nix::sys::termios::tcgetattr()`.

utkarshkukreti:
Apparently cargo doesn't automatically run tests inside nested folders.


carllerche:
Thanks! merged 954a1e6002c5c8dce4e88fb77e5fe398cb0871be

PR #66, opened 2015-02-13
iOS: term support is really redundant

vhbit:


carllerche:
I believe it, but I am no expert here. @utkarshkukreti did the original work. What do you think?

utkarshkukreti:
Are these headers/functions not present in iOS? (I also know next to nothing about iOS..)

vhbit:
They might be presented as iOS shares a lot with OS X. But the main thing is that there is no official way to write and run apps which use term - apps are sandboxed and are GUI-only. The only workaround is to jailbreak device. 

I bet a person who jailbreaks device and runs term apps written in Rust on it, will have enough expertise to open another PR.

utkarshkukreti:
Sounds good. :+1:

carllerche:
Pushed, thanks 19f3690627c45ab91ca849d1cb0c9781716c4367

PR #67, opened 2015-02-13
improved the stat test and included a test for fstat.

MarkusJais:


carllerche:
Looks good. Would you mind squashing into on commit? I'm not 100% sure what github thinks is going on here.

MarkusJais:
I was confused myself after the pull request :-)

I will try to squash this. Might take a few days as I have limited computer access at home right now.

PR #71, opened 2015-02-17
Fix type mismatch: errno now i32

little-arhat:
Fix type mismatch: errno now i32


carllerche:
Thanks. Merged 0e361c0341f7ab60e21d2fda818e352de53ac510

PR #73, opened 2015-02-18
Add inotify bindings

mbazarov:


carllerche:
Hi there. Thanks for the PR. One quick thought. It looks like you are taking a path as an argument. So maybe instead of adding `AsCString`, using `NixPath` would be better? `NixPath` is able to get a C String without allocating as well or panicking if the input is not valid (contains null bytes).

Thoughts?

kamalmarhubi:
Hi @mbazarov! I'm going through the open PRs on nix. Are you still interested in getting inotify support in? If so here are a couple of ideas base on how the library seems to have developed since you opened this (almost a year ago!):

* put the constants and extern fn declarations in a PR to [libc](https://github.com/rust-lang-nursery/libc)
* make this PR just contain the wrappers you have starting at about [line 80](https://github.com/carllerche/nix-rust/pull/73/files#diff-992886e985802f70fe6584c248c1fa2dR80)
* wrap up the constants with bitflags (see the stat module for [an example](https://github.com/carllerche/nix-rust/blob/0529fb8e324157bb02908eac5cfd9d0fa99c130f/src/sys/stat.rs#L20-L51))

kamalmarhubi:
I'm closing this because it's been inactive for a while. Feel free to reopen it any time if you're still interested in getting it in!

PR #74, opened 2015-02-19
Fix for latest rust by updating a conversion.

reem:


utkarshkukreti:
#71

reem:
@utkarshkukreti #71 looks like it has 30 random commits, whats going on there?

carllerche:
Dup of #71. I merged the other.

PR #79, opened 2015-02-20
Remove hash feature declaration

teotwaki:
The `hash` feature is no longer recognised by the compiler. It would seem this declaration is no longer needed.


carllerche:
Thanks :heart: merged 99ad14f6ed4f1ddb87523d646af18947b68f8297

PR #80, opened 2015-02-20
Use new ffi:CStr instead of c_str_to_bytes_with_nul

teotwaki:
I'm not 100% sure how to test this one. Feel free to discard it if you don't like the dereferencing aspect without it being thoroughly checked; this being an unsafe area and all.


carllerche:
Thanks. Merged 4abe2fc86d0fcaedc5dff1a841c6276705a976a5

PR #81, opened 2015-02-21
Re-export in6_addr

vhbit:


carllerche:
Sorry for the delay. It looks like this got pulled in: https://github.com/carllerche/nix-rust/blob/master/src/sys/socket/mod.rs#L36

PR #87, opened 2015-03-07
Miscellaneous changes; make it easier to manually integrate existing low level *nix code with nix-rust

koute:


carllerche:
So sorry for the delay. In what case do you need sockaddr_storage_to_addr exposed?

koute:
Currently I need it for converting the `sockaddr_storage` I get from calling `getaddrinfo`.

Since nix-rust is pretty low level as-is I don't think it would hurt to expose `sockaddr_storage_to_addr`; it's marked unsafe anyway.

carllerche:
Thanks. I rebased & pushed to master. 772aa42381ac467a18d6264cd0d0041dc65a6a6c

PR #92, opened 2015-04-03
Fd ist now RawFd (rustup)

fhartwig:


fhartwig:
Tests pass for me, travis failure seems to be unrelated to the changes.

fhartwig:
I've now pushed the changes required for the next nightly to this branch as well.

fhartwig:
Meh, missed the linux-specific bits again. Sorry, fixing it now.

Gekkio:
For some reason unknown to me, Clone isn't implemented for statically sized arrays larger than 32 elements, so you can't derive Clone for UtsName. A manual implementation works (fixed-size arrays still have a Copy implementation...?):

```rust
impl Clone for UtsName {
    fn clone(&self) -> UtsName {
        UtsName {
            sysname: self.sysname,
            nodename: self.nodename,
            release: self.release,
            version: self.version,
            machine: self.machine,
            domainname: self.domainname
        }
    }
}
```

With your changes and the above UtsName fix I can compile nix-rust on Ubuntu 64-bit successfully :)

CraZySacX:
You can change the clone implementation to

```rust
impl Clone for UtsName { fn clone(&self) -> UtsName { *self } }
```

Per the comment in this issue https://github.com/rust-lang/rust/pull/23860

fhartwig:
@CraZySacX @Gekkio Sorry for the breakage, I'm abusing travis as my compiler, since I'm not near a Linux machine. Thanks for your help, should be ok now.

PR #94, opened 2015-04-04
Compatibility with rust latest

zmack:
Hey, I was trying to get some stuff done with mio on rust head, and just ended up in this sed-yak-shave, hope you can get some use out of it.

The only tricky bit is the clone implementation in UtsName, which i'm still not sure is 100% correct, but at least it gets tests passing.

Anyway, on this particular branch tests are green on today's rust. Hope it helps :smile_cat:

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -6,8 +6,8 @@ use sys::stat::Mode;
 pub use self::consts::*;
 pub use self::ffi::flock;
 
-// Re-export Fd defined in std
-pub type Fd = ::std::os::unix::io::Fd;
+// Re-export RawFd defined in std
+pub type RawFd = ::std::os::unix::io::RawFd;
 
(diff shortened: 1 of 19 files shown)


carllerche:
Yeah, I've been trying to get nix running on Rust beta. I already applied some of these fixes, but haven't yet pushed. I will attempt to get nix running on the latest nightly at least.

fhartwig:
Re: the UtsName issue: apparently the recommended workaround is the one that I've used in  #92

zmack:
@fhartwig hah, nice, i was wondering about that one.

Good discussion on that as well down [here](https://github.com/rust-lang/rust/pull/23860), thanks for pointing me to it.

carllerche:
Thanks for the PR. Master should be fixed for Rust 1.0 Beta. I ended up merging #92

PR #101, opened 2015-04-14
Make `sigaction` unsafe

### LINKED ISSUE #90: `sigaction` should probably be `unsafe`
If you access thread-local data in a signal handler or panic, you probably invoke UB.

http://users.rust-lang.org/t/unix-signals-in-rust/733/3

tbu-:
This is done because interrupted execution has soundness bugs regarding
thread-local storage. Fixes #90.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -350,12 +350,11 @@ impl SigAction {
     }
 }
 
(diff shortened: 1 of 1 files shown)


carllerche:
thanks!

PR #104, opened 2015-04-19
Fix build on latest nightly

### LINKED ISSUE #103: Build fails on latest nightly
https://github.com/rust-lang/rust/pull/24461 seems to have broken the build on nightly. The `impl`s generated by the`sockopt_impl` macro violate this. I'm trying to find a good way to fix this, but my macro-fu is pretty weak.

fhartwig:
I'm not at all sure this is how this should be fixed, but on the off chance that this is useful to you, I thought I'd make a pr out if my local fixes anyway. Feel free to ignore it, though.

Fixes #103

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -277,32 +277,28 @@ pub enum SockLevel {
 
 /// Represents a socket option that can be accessed or set. Used as an argument
 /// to `getsockopt` and `setsockopt`.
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks for the PR. I ended up going a different route for this though. I opted not to expose more generics to the end user and to simplify the `SockOpt` trait.

Issue #97, opened 2015-04-09
Improve ARM support


carllerche:
This really requires CI infrastructure. I filed an issue with Travis: https://github.com/travis-ci/travis-ci/issues/3376. Also, it would require ARM builds of Rust, etc...

kamalmarhubi:
I filed #258 for CI on ARM. We should be able to use QEMU as is done in libc.

asomers:
We currently have ten different ARM builds running in CI, using QEMU.

Issue #95, opened 2015-04-08
cpuset_attribs not defined for ARM Linux


posborne:
No cpuset_attributes are defined for arch "arm" and os "linux".  This results in compilation errors when cross-compiling for triples like "arm-unknown-linux-gnueabihf" used by boards like the raspberry pi and beaglebone black.

I  copied the cpuset_attribs from android and these worked for my use case.

```
   Compiling nix v0.3.3
.../nix-0.3.3/src/sched.rs:94:25: 94:52 error: failed to resolve. Use of undeclared type or module `cpuset_attribs`
.../nix-0.3.3/src/sched.rs:94     cpu_mask: [CpuMask; cpuset_attribs::CPU_SETSIZE/cpuset_attribs::CPU_MASK_BITS]
```

carllerche:
There isn't that great ARM support right now, mostly because it isn't hooked into CI. If you have a PR adding `cpuset_attribs` for ARM linux, I will pull it in. In the mean time, I opened up an issue for improving ARM support: #97.

PR #105, opened 2015-04-21
sched: add support for arm linux using same cpuset_attribs as android

posborne:
The limit of 32 cores may not actually be a limit with arm-linux, but
I am not aware of anything in excess of 32 processors out there currently
and this is what I have been running for awhile now on a beaglebone
black (`--target=arm-unknown-linux-gnueabihf`).

This change addresses #95 and relates to #97.

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -63,7 +63,7 @@ mod cpuset_attribs {
     }
 }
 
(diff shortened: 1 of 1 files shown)


carllerche:
Merged thanks: bc84ac1358d49a5391724f28dc55e178efdf18bc

PR #106, opened 2015-04-23
Fix build on nightly

fhartwig:


fhartwig:
`rand` hasn't been updated yet, so the test won't build.

fhartwig:
I've amended the commit to include the new `rand` version, so it should be fine now.

carllerche:
Thanks!

PR #111, opened 2015-05-03
feat: add execvpe

houqp:


houqp:
oops, i forgot something, will send another PR with fix.

PR #112, opened 2015-05-03
feat: add execvpe

houqp:


houqp:
hmm... weird, it passed for beat, but not nightly, i am using the nightly build too.

houqp:
OK, it passed on nightly now without any change...

carllerche:
Looks mostly good. I don't believe that this function is supported on Darwin (and possibly other OSes). Could you guard the API with cfg flags?

houqp:
Good catch, I moved it into linux module now :)

carllerche:
Thanks! merged.

PR #113, opened 2015-05-05
Add lstat

dead10ck:
Adds a wrapper for the `lstat` function.

### DIFF
--- a/src/sys/stat.rs
+++ b/src/sys/stat.rs
@@ -1,7 +1,7 @@
 pub use libc::dev_t;
 pub use libc::stat as FileStat;
 
-use {Error, Result, NixPath, AsExtStr,from_ffi};
+use {Error, Result, NixPath, AsExtStr, from_ffi};
 use errno::Errno;
 use fcntl::Fd;
 use libc::mode_t;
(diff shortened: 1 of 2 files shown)


carllerche:
Hey there. Thanks for the PR. I tried running the tests on my laptop (OS X) and I get the following test failure:

```
---- test_stat::test_stat_fstat_lstat stdout ----
	thread 'test_stat::test_stat_fstat_lstat' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os(17) }', /Users/rustbuild/src/rust-buildbot/slave/beta-dist-rustc-mac/build/src/libcore/result.rs:729
```

Any ideas?

dead10ck:
Hmm, error 17 is a "file exists" error, which probably means either "foo" or "bar" existed in the target folder before the test was run. Those files might still be there if you had at some point in the past run those tests and they failed, meaning they didn't get deleted in the end. Can you try making sure they are deleted and then run the test again?

dead10ck:
However, I just ran this on my OS X laptop as well, and it fails because `st_blocks` was not 0. I think this is pretty strange, because it passes on Linux, meaning that `stat` reports that the symlink has 0 blocks allocated. I'm not sure how a symlink can have `st_size` > 0, but have no blocks allocated.

carllerche:
Cool. Let me know when I should test again.

dead10ck:
It should be good to go.

carllerche:
Thanks. Merged: ecbbf1a8176130ca845b23c3e9dfe37d45da3c13

PR #114, opened 2015-05-09
ioctl: implement generic support for the ioctl syscall and supporting…

### LINKED ISSUE #108: Make ioctl usable for more than just TIOCGWINSZ
The version of ioctl that is currently present in nix is not particularly useful (only has support for TIOCGWINSZ) and the design would require that the library itself be modified in order to add additional ioctls.  I would like to see nix support a good lower-level ioctl implementation that can be used to raise the level of abstraction for any rust code that needs to send ioctls to regular and special files.

My immediate need for something like this is in implementing a library that provides access to Linux spidev devices (https://www.kernel.org/doc/Documentation/spi/spidev / http://lxr.free-electrons.com/source/include/uapi/linux/spi/spidev.h), but there are many, many devices out there that require ioctl for normal operation.  I think there is probably some interface that can be provided that provides a little more help to implementers than what is currently available (and more than the libc ioctl).

A good starting place seems to be looking at the low-level macros available in the ioctl header: http://lxr.free-electrons.com/source/include/uapi/asm-generic/ioctl.h

I am going to play with things and see if I can come up with a more concrete proposal for a refined design for ioctls.

posborne:
… functions

This commit provides a new implementation for ioctl that is much
more generic, allowing for clients to use send any manner of ioctl
requests at special files.  The implementation provides two main features
that help to raise the level of abstraction over that provided by libc.

1. The module now provides functions that provide the same functionality
   as the linux kernel _IO* macros.  These are used frequently in the
   linux kernel for building ops for ioctls.  The use of these helper
   functions are not required.
2. Functions are provided for the 3 main types of ioctl usage patterns
   (read, write, and execute).  For many subystems, the read() call
   which returns a Result<T> and the write calls taking a &T provide
   a nice interface.

All of the methods wrapping ioctl are unsafe and will probably need to
remain that way unless knowledge of the semantics of every possible
ioctl call are added to the nix library.  The best that exists for
ioctls are some conventions around the op, but even these conventions
are really only used for newer devices added to the kernel.

This change resolves #108

### DIFF
--- a/src/sys/ioctl.rs
+++ b/src/sys/ioctl.rs
@@ -1,45 +1,210 @@
-use libc;
+///! Provide helpers for making ioctl system calls
+///!
+///! # Overview of IOCTLs
+///!
+///! The `ioctl` system call is a widely support system
+///! call on *nix systems providing access to functions
+///! and data that do not fit nicely into the standard
+///! read and write operations on a file itself.  It is
+///! common to see ioctls used for the following purposes:
+///!
+///! * Provide read/write access to out-of-band data related
+///!   to a device such as configuration (for instance, setting
+///!   serial port options)
+///! * Provide a mechanism for performing full-duplex data
+///!   transfers (for instance, xfer on SPI devices).
+///! * Provide access to control functions on a device (for example,
+///!   on Linux you can send commands like pause, resume, and eject
+///!   to the CDROM device.
+///! * Do whatever else the device driver creator thought made most sense.
+///!
+///! Ioctls are synchronous system calls and are similar to read and
+///! write calls in that regard.
+///!
+///! The prototype for the ioctl system call in libc is as follows:
+///!
+///! ```c
+///! int ioctl(int fd, unsigned long request, ...);
+///! ```
+///!
+///! Typically, an ioctl takes 3 parameters as arguments:
+///! 1. An open file descriptor, `fd`.
+///! 2. An device-dependennt request code or operation.  This request
+///!    code is referred to as `op` in this module.
+///! 3. Either a pointer to a location in memory or an integer.  This
+///!    number of pointer may either be used by the kernel or written
+///!    to by the kernel depending on how the operation is documented
+///!    to work.
+///!
+///! The `op` request code is essentially an arbitrary integer having
+///! a device-driver specific meaning.  Over time, it proved difficult
+///! for various driver implementors to use this field sanely, so a
+///! convention with macros was introduced to the Linux Kernel that
+///! is used by most newer drivers.  See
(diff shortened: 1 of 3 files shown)


posborne:
Here's the code I have been developing along with this functionality.  Viewing this along with the code here should help with understanding why some of the implementation is the way it is.  Let me know if there are questions....

https://github.com/posborne/rust-spidev/blob/nix-upstream/src/spidevioctl.rs

carllerche:
Hey, sorry for the delay. This looks good for the most part (of course, I have personally never used `ioctl`, so I will take your word on the details).

One problem though, it doesn't seem to compile on OS X. It looks like it is related to your cast to `as i32` [here](https://github.com/carllerche/nix-rust/pull/114/files#diff-664f8052e0c4be85caa69d4addb49e82R190). I'm not 100% sure why this is happening though since the fn def seems to take a `c_int` (though it probably should be set to cast to `c_int` vs. `i32`)

The error:

```
src/sys/ioctl.rs:190:38: 190:47 error: mismatched types:
 expected `u64`,
    found `i32`
(expected u64,
    found i32) [E0308]
src/sys/ioctl.rs:190     convert_ioctl_res(libc_ioctl(fd, op as i32, data_ptr))
                                                          ^~~~~~~~~
src/sys/ioctl.rs:202:38: 202:47 error: mismatched types:
 expected `u64`,
    found `i32`
(expected u64,
    found i32) [E0308]
src/sys/ioctl.rs:202     convert_ioctl_res(libc_ioctl(fd, op as i32, data as *const T))
                                                          ^~~~~~~~~
src/sys/ioctl.rs:209:47: 209:56 error: mismatched types:
 expected `u64`,
    found `i32`
(expected u64,
    found i32) [E0308]
src/sys/ioctl.rs:209     convert_ioctl_res(unsafe { libc_ioctl(fd, op as i32) })

posborne:
@carllerche, thanks for taking a look.  I'll test on my mac and make the change to use the libc types (c_int here) where appropriate.

carllerche:
I should probably see if I can get travis to run tests on OS X.

posborne:
This turned out to be more anything that I thought.  The type signatures for ioctl are actually different on linux than on bsd platforms.

* BSD: https://github.com/rust-lang/rust/blob/master/src/liblibc/lib.rs#L6078
* Linux: https://github.com/rust-lang/rust/blob/master/src/liblibc/lib.rs#L6114

I tested on mac and fixed up.  Travis tested Linux and it seems to be passing now.

carllerche:
Thanks! Merged: e29ea45de7cce4365843a0ee0dfca0aff14e84f9

PR #115, opened 2015-05-09
Feature/freebsd

andrewjstone:
Library now Builds on freebsd with `cargo build`.

`cargo test` still does not work. It crashes the compiler.


carllerche:
Thanks, merged: 1a2d6d51f4f585e975eb3fb420904427d1f706f4

PR #118, opened 2015-05-16
cleaned up mq unit tests

MarkusJais:
Just some cleanups for the mq unit tests.

Next step is adding the missing mq functions and more tests.


MarkusJais:
I've also added getpid and getppid functions. They wrap the unsafe functions from the Rust library.
They don't return Result but only pid_t as those functions always return a valid value (man page).

carllerche:
Thanks!

PR #121, opened 2015-05-20
Fix AsExtStr to return CStrings

### LINKED ISSUE #117: `impl AsExtStr for OsStr` does not convert to a proper C string
I started noticing some strange behavior when trying to use the `lstat` function I recently submitted; `lstat` seemed to be returning `ENOENT` when the file did indeed exist, and opened correctly with `File::open()`. I began investigating, and I think I've found the cause: the implementation of `AsExtStr` for `OsStr`:

```rust
/// Converts a value to an external (FFI) string representation
trait AsExtStr {
    fn as_ext_str(&self) -> *const libc::c_char;
}

impl AsExtStr for OsStr {
    fn as_ext_str(&self) -> *const libc::c_char {
        self.as_bytes().as_ptr() as *const libc::c_char
    }
}
```

It appears the problem is that this just borrows the byte slice from the `OsStr` and then simply casts to a `libc::c_char`. If I am not mistaken, believe that this does not properly convert to a null-terminated C string. I was able to construct a rather interesting test case which showcases this (you can paste this at the bottom of `src/lib.rs` to test for yourself):

```rust
#[cfg(test)] use libc::funcs::c95::stdio;
#[cfg(test)] use std::ffi::CStr;
#[test]
#[allow(unused_variables)]
fn test_as_ext_str() {
    let foo_str = "foo";
    let bar_str = "bar";
    let s = Path::new(foo_str).as_os_str();
    let ext_str = s.as_ext_str();
    unsafe { stdio::puts(ext_str); }
    let cstr = unsafe { CStr::from_ptr(ext_str) };
    let back_to_str = std::str::from_utf8(cstr.to_bytes()).unwrap();
    assert_eq!(foo_str, back_to_str);
}
```

This is the output:

```sh
dead10ck at arch-laptop in ~/src/rust/nix-rust (master●)
$ cargo test test_as_ext_str
     Running target/debug/nix-21d2363c10db6dc6

running 1 test
foobarsrc/lib.rs
test test_as_ext_str ... FAILED

failures:

---- test_as_ext_str stdout ----
	thread 'test_as_ext_str' panicked at 'assertion failed: `(left == right)` (left: `"foo"`, right: `"foobarsrc/lib.rs"`)', src/lib.rs:167



failures:
    test_as_ext_str

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured

thread '<main>' panicked at 'Some tests failed', /home/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libtest/lib.rs:259

```

You can see just below the `running 1 test` line the output of the call to `libc`'s `puts` function. The input string was "foo", but because the pointer that `as_ext_str` returned was not properly null-terminated, it also printed out some other strings that were adjacent in memory.

This will likely be non-trivial to fix, since `AsExtStr` is used in many places throughout the crate; a proper `CStr` or `CString` will probably need to be constructed from the `OsStr` argument, and that should be returned instead of a raw pointer. What do you think?

### LINKED ISSUE #120: unistd::pivot_root() is possibly broken
While implementing a fix for #117, I noticed that `unistd::pivot_root()` is possibly broken.

On 9 Feb, effb423 introduced the `NixPath` trait, which at first took a closure which has a `*const libc::c_char` as an argument, and `unistd::pivot_root()` was changed to use `with_nix_path()` and its closure's arguments as arguments in the call to `sys::syscall::syscall()`.

Then on 26 Feb, e8a58c8 changed `NixPath` to take an `&OsStr` instead of a `*const libc::c_char`, but `unistd::pivot_root()` was not updated. So it appears that `unistd::pivot_root()` is now passing `&OsStr`s to `sys::syscall::syscall()`. I imagine this wasn't caught because `unistd::pivot_root()` doesn't appear to be covered by any unit tests.

dead10ck:
As described in #117, the `AsExtStr` trait is defined to return a raw `*const libc::c_char`. Its `impl` for `OsStr` simply borrowed the byte slice from its `OsStr` argument and cast it to a `*const libc::c_char`, which does not construct a proper null-terminated C string.

This changes the `AsExtStr` trait to return a `Result<CString>`, and changes the `impl` for `OsStr` to allocate a new owned `CString` from its `OsStr` argument and return that instead.

Fixes #117
Fixes #120

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -73,7 +73,8 @@ mod ffi {
 
 pub fn open<P: ?Sized + NixPath>(path: &P, oflag: OFlag, mode: Mode) -> Result<Fd> {
     let fd = try!(path.with_nix_path(|osstr| {
-        unsafe { ffi::open(osstr.as_ext_str(), oflag.bits(), mode.bits() as mode_t) }
+        let cstr = try!(osstr.as_ext_str());
+        Ok(unsafe { ffi::open(cstr.as_ptr(), oflag.bits(), mode.bits() as mode_t) })
     }));
 
     if fd < 0 {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,14 +48,17 @@ pub mod unistd;
  *
  */
 
+use std::convert;
 use std::{ptr, result};
 use std::path::{Path, PathBuf};
+use std::ffi;
 
 pub type Result<T> = result::Result<T, Error>;
 
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Error {
     Sys(errno::Errno),
+    NulError(ffi::NulError),
     InvalidPath,
 }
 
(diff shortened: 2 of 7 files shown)


carllerche:
Hey there. I think a better way to do this would be to get rid of `AsExtStr` entirely and fix `NixPath` to invoke the function with a `CStr` instead. `NixPath` already ensures that the string is null terminated.

https://github.com/carllerche/nix-rust/blob/master/src/lib.rs#L89-L109

If you don't want to do it, I can fix it :smile:

carllerche:
I do appreciate the PR :heart:

dead10ck:
Oh, ok, no problem; you know your lib better than I do, and are much more experienced with Rust than I am! I probably won't have time to do it soon, though.

carllerche:
Hey there. I pushed this out: 21fb1394a111d19079da90a4d31171b0d6adc27c I think this fixes the issues.

PR #124, opened 2015-05-25
Add sys::socket send and recv

phildawes:
Adds the libc recv and send functions to sys::socket. I wanted this for doing low-level tcp with nix.

Let me know if you need me to improve the patch in some way, or if there's a better way to achieve the same thing. Thanks!


carllerche:
Thanks! merged: e71077d824f8d5e7003c00c0c1e939ba5955be8f

PR #125, opened 2015-05-26
Documentation: epoll usage example added

hoxnox:
I wrote example of epoll usage. If you wish, you could add it into the project so it would be easier for beginners to deal with it.

### DIFF
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -1,3 +1,154 @@
+//! Linux epoll binding for Rust.
+//!
+//! Example:
+//!
+//!     extern crate nix;
+//!     use nix::sys::epoll;
+//!     use nix::sys::epoll::EpollOp;
+//!     use nix::sys::socket;
+//!     use std::ascii::AsciiExt;
+//!     use std::net;
+//!
+//!     fn main() {
+//!         let addrs = vec!(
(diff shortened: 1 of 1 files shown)


carllerche:
Hey, thanks for the PR. Documentation is definitely appreciated. Were you having trouble getting started? If so, have you used epoll before? 

The example seems a bit long to include as inline documentation, but it would probably be OK to include in `examples/epoll.rs`. Also, it will get the tests passing :smile: (doc examples are evaluated as part of tests and this example doesn't terminate).

hoxnox:
epoll, poll, select - logic is the same. Not much new things. I wrote example only to understand how you implemented it in rust.
One thing is not clear - buffer for read. If I would use unistd::read as following, it not work.

                let mut data: Vec<u8> = Vec::with_capacity(2048);
                nix::unistd::read(fd, &mut data);

carllerche:
The reason it didn't work is that, even though `Vec::with_capacity(2048)` allocates 2048 bytes under the hood, it only exposes indexes that were written to (to avoid accessing uninitialized memory). The easiest way to get a buffer is `let buf = [0; 2048];` (I think).

hoxnox:
It's clear for C developer, but confuses C++ developer =).

carllerche:
I'm going to close this for now, but if you have an example to add, feel free to open another PR.

PR #126, opened 2015-05-29
Basic OpenBSD support.

ltratt:
Some of the tests are currently unrunnable, but the basic library is at least
buildable with this patch.


carllerche:
Thanks! merged

PR #127, opened 2015-06-04
added wait system call and unit test

MarkusJais:
wait in C is equivalent to

 waitpid(-1, &status, 0);

so I just called the waitpid function and did not wrap the C function.


carllerche:
Merged, thanks! 1d9bbacdfbdf21e8af25a19a91f746bed1087b5e

PR #128, opened 2015-06-05
Add pwrite and pread

fhartwig:
This adds pread and pwrite. I'm unsure if these should take a `RawFd` or
a `&File`. The latter has the benefit that the user can't accidentally pass
e.g. a fd representing a pipe or a socket, but it complicates things if the
caller only has a RawFd and no associated `File` object. What do you think?

### DIFF
--- a/src/sys/uio.rs
+++ b/src/sys/uio.rs
@@ -9,7 +9,7 @@ use std::os::unix::io::RawFd;
 
 mod ffi {
     use super::IoVec;
-    use libc::{ssize_t, c_int};
+    use libc::{ssize_t, c_int, size_t, off_t, c_void};
     use std::os::unix::io::RawFd;
 
     extern {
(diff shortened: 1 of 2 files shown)


fhartwig:
I've also replaced the deprecated `fs::soft_link` in the tests to fix the compiler warning. I hope it's okay that I added that to this pr.

carllerche:
Good question. You can go from `RawFd` -> `File` using `FromRawFd ` (though it is still flagged as unstable). I am wondering exactly what nix is supposed to be. Whether it should mirror the system APIs w/ some rusty sugar or try to make positions like you are suggesting (aka, take a `File` vs. `RawFd).

For now, I would say lets stick with `RawFd` since that is what the rest of nix does. The PR looks good, I am going to merge it. Thanks!

PR #134, opened 2015-06-15
Fix types for pread/pwrite.

### LINKED ISSUE #131: ARM fix
I recently compiled nix for my beagleboard black - it went smoothly, but I got two compile errors, both in uio.rs.

The parameter `offset` to ffi::pread and ffi::pwrite are i32 on my arm. Maybe they should be isize? I am not sure how to express that in rust in a portable manner.

SirVer:
This fixes compile for 32 bit systems like the ARM I am working on.

Fixes #131.

[Review on Reviewable](https://reviewable.io/reviews/carllerche/nix-rust/134)

### DIFF
--- a/src/sys/uio.rs
+++ b/src/sys/uio.rs
@@ -52,10 +52,10 @@ pub fn readv(fd: RawFd, iov: &mut [IoVec<&mut [u8]>]) -> Result<usize> {
     return Ok(res as usize)
 }
 
-pub fn pwrite(fd: RawFd, buf: &[u8], offset: i64) -> Result<usize> {
+pub fn pwrite(fd: RawFd, buf: &[u8], offset: off_t) -> Result<usize> {
     let res = unsafe {
         ffi::pwrite(fd, buf.as_ptr() as *const c_void, buf.len() as size_t,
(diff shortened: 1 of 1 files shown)


src/sys/uio.rs
```diff
@@ -55,7 +55,7 @@ pub fn readv(fd: RawFd, iov: &mut [IoVec<&mut [u8]>]) -> Result<usize> {
 pub fn pwrite(fd: RawFd, buf: &[u8], offset: i64) -> Result<usize> {
     let res = unsafe {
         ffi::pwrite(fd, buf.as_ptr() as *const c_void, buf.len() as size_t,
-                    offset)
+                    offset as off_t)
```
posborne: Would it not be better to change the type signature for `pread`/`pwrite` to take offset as an `off_t` rather than casting here?

carllerche: :+1: I agree. @SirVer Do you want to update it?

SirVer:
Updated.

carllerche:
Thanks! Merged: d9808975608e7818be2d564357469dab0a31e1d2

PR #137, opened 2015-06-24
Bump libc dependency up to version 0.1.8

jnicholls:
libc::TCP_KEEPIDLE is not defined for linux in libc v0.1.4, so this bump will alleviate that issue.


carllerche:
Thanks!

jnicholls:
You bet!

PR #141, opened 2015-07-02
Remove sa_tramp from macos sigaction

### LINKED ISSUE #140: Illegal hardware instruction on Mac OS when using signals
Hi.  I have the following program
```rust
extern crate nix;
extern crate libc;

use std::thread;
use nix::sys::signal;

extern "C" fn handle_signal(val: libc::c_int) {
    println!("Handling");
}

fn main() {
    let sig_action = signal::SigAction::new(
        handle_signal,
        signal::SockFlag::empty(),
        signal::SigSet::empty()
        );
    unsafe { let _ = signal::sigaction(signal::SIGINT, &sig_action).unwrap(); }
    loop {
        thread::sleep_ms(100);
    }
}
```

When I run this program and hit ^C I get the output "Illegal instruction: 4" and the program crashes

My system information is
libc: v0.1.8
nix: v0.3.7
Mac OS: 10.10.3
rustc --version:  1.2.0-nightly (9ad0063a1 2015-06-20)

I believe this has to do with the implementation of [sigaction](https://github.com/carllerche/nix-rust/blob/master/src/sys/signal.rs#L267).  When I reimplement this struct myself like so the sigaction function call works
```rust
#[repr(C)]
pub struct sigaction {
    pub sa_handler: extern fn(libc::c_int),
    pub sa_sigaction: extern fn(libc::c_int, *mut siginfo_t, *mut libc::c_void),
    pub sa_mask: sigset_t,
    pub sa_flags: libc::c_int,
    pub sa_restorer: extern fn(),
}

type sigaction_t = self::sigaction;
```
Thanks in advance

nikklassen:
Fixes #140  This current implementation follows the struct definition for __sigaction in [include/sys/signal.h](http://www.opensource.apple.com/source/xnu/xnu-2782.1.97/bsd/sys/signal.h).  This changes it to the implementation to ```sigaction``` expected by the sigaction function in include/signal.h.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -266,7 +266,6 @@ pub mod signal {
     #[allow(missing_copy_implementations)]
     pub struct sigaction {
         pub sa_handler: extern fn(libc::c_int),
-        sa_tramp: *mut libc::c_void,
         pub sa_mask: sigset_t,
         pub sa_flags: SockFlag,
     }


carllerche:
Thanks!

PR #143, opened 2015-07-02
Add support for ptrace

joekain:
This is working for me on Linux in a project I'm working on.

One thing I've noticed is that the status from waitpid isn't made available which makes it difficult to check why the tracee is stopped.  I have a change to waidpid that I will submit as a separate PR once I get it working on FreeBSD.

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -3,44 +3,50 @@ use libc::c_int;
 pub use self::consts::*;
 pub use self::consts::Errno::*;
 
-/// Returns the platform-specific value of errno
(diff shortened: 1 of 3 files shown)


carllerche:
I skimmed quickly and it looks good as far as I can tell :smile: If this is ready to go, I can squash & merge if you would like.

joekain:
That sounds great, thanks!

carllerche:
Merged, thanks: a0fc60700f9e8b93e570e51460b10201f721b635

Issue #144, opened 2015-07-04
Intermittent failures running cargo test


joekain:
I was trying to track down some intermittent failures with some new tests I was writing and write a script to run `cargo test` over and over.  With master I see 3 distinct failures signatures:

1. This is due to an off-by-one error in sys::test_uio::test_readv.  I have a change to fix this and will submit a PR.
        failures:

        ---- sys::test_uio::test_readv stdout ----
        thread 'sys::test_uio::test_readv' panicked at 'Rng.gen_range called with low >= high', /home/joseph/.cargo/registry/src/github.com-0a35038f75765ae4/rand-0.3.8/src/lib.rs:441

        failures:
        sys::test_uio::test_readv

2. I think this is due to tests running tests in parallel and wait() waking up for a different child than expected.  Is there a way to run tests in isolation?
        failures:

        ---- test_unistd::test_wait stdout ----
        	thread 'test_unistd::test_wait' panicked at 'assertion failed: `(left == right)` (left: `Ok(Exited(21710))`, right: `Ok(Exited(21714))`)', test/test_unistd.rs:41

        ---- test_unistd::test_execve stdout ----
        	thread 'test_unistd::test_execve' panicked at 'called `Result::unwrap()` on an `Err` value: Sys(ECHILD)', ../src/libcore/result.rs:731

        failures:
            test_unistd::test_execve
            test_unistd::test_wait
    
3. This one takes longer to reproduce.  I'm not sure what the problem is here.
        failures:

        ---- sys::test_termios::test_tcgetattr stdout ----
        	thread 'sys::test_termios::test_tcgetattr' panicked at 'assertion failed: `(left == right)` (left: `Some(Sys(ENOTTY))`, right: `Some(Sys(EBADF))`)', test/sys/test_termios.rs:15

        failures:
            sys::test_termios::test_tcgetattr

Please let me know if I should file 3 separate issues for these.

carllerche:
Merged this. Thanks!

joekain:
Oops, I guess I should have filed 3 separate issues.  Problems 2 and 3 still persist.  I'll file new issues for them.

PR #145, opened 2015-07-04
Fix off-by-one error calling Rng::gen_range

joekain:
Fix the boundary condition so that we never call

    Rng::gen_range(64, 64).

Helps #144


carllerche:
Thanks!

PR #147, opened 2015-07-05
ioctl work

cmr:


cmr:
cc @ltratt @utkarshkukreti @posborne @kennytm @fhartwig, people who have touched the previous ioctl implementation.

cmr:
Wow, some of the tests are really volatile..

joekain:
@cmr, yes some tests do seem to be failing intermittently.  I filed #144 about this yesterday.  I think I can fix one of the problems.  The one that hit your commit might be harder to fix.

cmr:
Yeah, not sure about the tcgetattr test. For the waitpid one, setting RUST_TEST_THREADS=1 should disable all paralellism I think?

joekain:
@cmr, yes RUST_TEST_THREADS=1 should work but it seems heavy handed.  Only the one wait() test need to be serialized.

posborne:
@cmr, ioctls's are "fun" aren't they?  I took the last shot at ioctl.  Initially, I think this implementation should cover the use cases I have had for library like https://github.com/posborne/rust-spidev while providing better cross-platform support.

The macros for `ioctl!` don't feel quite intuitive, but I don't have a counter-proposal at this time.  The use of `bad` makes sense but could also be confusing as this type is pretty common, at least in Linux.

cmr:
Fortunately, the "bad" is only exposed internally.

I have no particular feelings about the syntax that the `ioctl!` macro accepts, but it shouldn't be used frequently. Ideally, `nix` would expose everything.

ltratt:
Given this doesn't seem to touch the OpenBSD parts I added, I don't have any useful comments :)

carllerche:
cc @posborne. Hoping to get some feedback (given my lack of ioctl knowledge)

cmr:
@ltratt Note that this patch *removes* support for all other platforms but Linux, as I haven't personally looked into their command encoding etc.

posborne:
> Ideally nix would expose everything

I really feel that this is an unrealistic expectation based on the real work I have done with ioctls in the past, particularly on embedded Linux systems where you end up talking to a larger variety of devices than one might see in a desktop or server environment.  I think we need to leave  the door open for users and library providers to be able to define their own ioctls.  The question for me is whether it is even worth trying to "automatically" provide any kind of interface for the specific ioctls available on any given system.

Providing an automatic interface that provides a good and safe interface to most ioctls is, in my opinion, a losing battle (which is why new drivers are not accepted using ioctl in Linux).  I would rather not provide methods for accessing system interfaces than to provide ones that are broken.

I'm working to see if I can port my SPI library to using the new interface as a trial.  Do you have example code using the interface that you can show as well?

src/sys/ioctl/platform/linux.rs
```diff
@@ -0,0 +1,415 @@
...
+            $crate::sys::ioctl::ioctl(fd, $nr as $crate::sys::ioctl::libc::c_ulong, data)
+        }
+    );
+    (none $name:ident with $ioty:expr, $nr:expr) => (
+        pub unsafe fn $name(fd: $crate::sys::ioctl::libc::c_int) -> $crate::sys::ioctl::libc::c_int {
+            $crate::sys::ioctl::ioctl(fd, io!($ioty, $nr) as $crate::sys::ioctl::libc::c_ulong)
+        }
+    );
+    (read $name:ident with $ioty:expr, $nr:expr; $ty:ty) => (
+        pub unsafe fn $name(fd: $crate::sys::ioctl::libc::c_int, val: *mut $ty) -> $crate::sys::ioctl::libc::c_int {
```
posborne: The previous implementation returned a `nix::Result` with `Ok` for `res >= 0`.  0 is typical for a return value but a length is sometimes returned that is greater than 0[1].  The error returned also handled getting error information from the errno.

Unless there is a good reason to do else-wise, I think we would want to do the same here.

[1] http://man7.org/linux/man-pages/man2/ioctl.2.html

src/sys/ioctl/platform/linux.rs
```diff
@@ -0,0 +1,415 @@
...
+    pub const NONE: u8 = 1;
+    #[doc(hidden)]
+    pub const READ: u8 = 2;
+    #[doc(hidden)]
+    pub const WRITE: u8 = 4;
+    #[doc(hidden)]
+    pub const SIZEBITS: u8 = 13;
+    #[doc(hidden)]
+    pub const DIRBITS: u8 = 3;
+}
```
posborne: The mips and powerpc constants appear to be identical.  Combine them?

cmr:
>I really feel that this is an unrealistic expectation based on the real work I have done with ioctls in the past, particularly on embedded Linux systems where you end up talking to a larger variety of devices than one might see in a desktop or server environment.

If it's in a header, the scripts in `etc` can extract it. If it's not exposed by a header distributed with the Linux kernel, then you can still bind it manually. I'm not saying that the implementation details exposed by `ioc` etc should be hidden, just that it really ought to be possible to bind everything in mainline. Third-party modules are an important exception that makes the rule.

> Providing an automatic interface that provides a good and safe interface to most ioctls is, in my opinion, a losing battle (which is why new drivers are not accepted using ioctl in Linux). I would rather not provide methods for accessing system interfaces than to provide ones that are broken.

I agree -- does this implementation try to do that? In particular, nothing is safe.

src/sys/ioctl/platform/linux.rs
cmr: That's an excellent idea.

posborne: The old interface also was provided `ioctl::read` with the following signature:

```rust
pub unsafe fn read<T>(fd: RawFd, op: ioctl_op_t) -> Result<T> {
...
}
```

Whereas the current interface provides the somewhat less safe and friendly interface (abbreviated for clarity):

```rust
(read $name:ident with $ioty:expr, $nr:expr; $ty:ty) => (
    pub unsafe fn $name(fd: c_int, val: *mut $ty) -> c_int {
    ...
    }
);
```

The main difference being that the user must always pass in a `*mut ty` in the new implementation where they would just get back an owned value in the new implementation.  I see no difference in safety between the two implementations (both are unsafe) but there is more the user could screw up in the new interface.  The old implementation also provided `read_into<T>(fd: RawFd, op: ioctl_op_t, data: &mut T) -> Result<c_int>` and `read_into_ptr<T>(fd: RawFd, op: ioctl_op_t, data_ptr: *mut T) -> Result<c_int>` as they are useful in some cases.  The first mentioned, however, is the most ergonomic for the cases in which it can be used.

cmr:
@posborne I wrote the original library this impl comes from for use in https://github.com/cmr/evdev

posborne:
> I agree -- does this implementation try to do that? In particular, nothing is safe.

I didn't spot anything obviously wrong while looking over the ioctls that were present (versus the ones that are left commented) that was obviously problematic.  At the same time, most of the really interesting ioctls require read/writing to a structure which cannot be handled currently (would probably require something like bindgen for the structures, but that can get very messy).

For evdev, it looks like you added the structure definitions for those structures that are part of the kernel API to nix itself.  I think these (and the ioctl constants) should probably live with your library outside of nix (Awesome lib BTW; I may find opportunity to use that myself sometime soon).

carllerche:
@cmr also, what was the reason for removing the debug impls for the various epoll types?

carllerche:
I personally have no opinion about this, so I can merge once there is consensus that it is good to go! :smile:

cmr:
@carllerche bitflags now provides the `Debug` impls.

cmr:
@posborne I'd prefer it if `nix` were the one canonical source for all raw Linux kernel APIs. It has to live somewhere, why not all here?

carllerche:
@posborne ping, what are your thoughts re: this?

posborne:
Sorry for the slow response.

> @posborne I'd prefer it if nix were the one canonical source for all raw Linux kernel APIs.

I think nix might be able to support all major ways of accessing kernel APIs (e.g. ioctl, nl80211, sysfs, functionfs) and some APIs that are commonly desired, but supporting *all* APIs seems overly ambitious and likely to make maintenance of the library difficult.  There is real complexity in many of the APIs and capturing this for all APIs would involve a lot of code.

> It has to live somewhere, why not all here?

I think it is easier to maintain several crates, each with a specific purpose, than to maintain a monolithic crate that tries to do too much.  With rust, we have a great packaging and distribution system in cargo and crates.io.  For SPI devices, users can use the spidev crate I have created.  For evdev access, they can use the evdev crate.

cmr:
@posborne While I agree that that is certainly the case for higher-level interfaces, the `ioctl`s and structs used in them will be the same everywhere.

cmr:
I've removed the contentious code.

cmr:
@posborne So, removing the various wrapper functions leaves this whole implementation in a precarious position. The macros make less sense when you're not generating all of the code with macros. I think I'd rather just maintain my `ioctl` crate than try and force this particular approach on `nix`.

posborne:
@cmr, I do think the macros make it easier for describing ioctl APIs, even if we do not include generated or non-generated ioctl APIs in nix from the get-go.  I think it is important to have a moderately stable API from the start; I may take a stab at changing the API exposed as discussed in comments eariler.  If so, I'll let you know where those changes are so you can pull them into this PR.

posborne:
Ok, I made some additional changes to return a result type rather than just returning the integer from the underlying call.  In order to build, I had to rebase on master and make one other change.  You should be able to cherry-pick my two commits (if deemed acceptable) onto your branch: https://github.com/posborne/nix-rust/commits/ioctl-rework

Here's the relevant diff: https://github.com/posborne/nix-rust/compare/eb837f8766add26f9b86e634206aaa05b022d87d...posborne:ioctl-rework

carllerche:
@cmr @posborne Sorry for the delay. What is the status of this PR? It seems that you are reaching a consensus?

cmr:
Goodness, sorry for the delay myself... I think what @posborne suggested (and implemented) is excellent, and if he's happy with it, I'm happy with it!

posborne:
I think I'm OK moving forward with what we ended up with.  Since it breaks API compatibility with previous versions, this should be a version bump (with semantic versioning, I believe a bump to the next 0.x is appropriate).

With my changes, the one part I think I may have wrong is with the following macro... https://github.com/posborne/nix-rust/compare/eb837f8766add26f9b86e634206aaa05b022d87d...posborne:ioctl-rework#diff-0f39576519037fe2d243b366759ea808R94

I think that needs to be in a mod or something to prevent polluting the namespace of the code that uses the macro?  I feel like I've seen code that does that since I wrote this.

cmr:
@posborne Are you worried about the `let res`? If so, the Rust macro system makes that not a problem: it doesn't add an identifier that the client's code can observe, and it won't clobber any `res` that the client has.

posborne:
@cmr, That was my concern.  If it's not a problem, excellent!  Good one Rust!

carllerche:
So, this PR is ready to go then? Let me know and I will merge. Also, it would probably be a good idea to bump the version as part of the PR.

posborne:
I opened up a separate PR that has the original changes from @cmr and my changes rebased on master: https://github.com/carllerche/nix-rust/pull/172

carllerche:
Zomg! Finally merged! :tada:

posborne:
Excellent!  I've been using this new stuff for a new lib[1] I am working on, so I'm glad to see it merged up.

[1] https://github.com/posborne/rust-i2cdev

carllerche:
@posborne @cmr I can cut a new crates io release when you think it is good. Now or later if you want to let it bake.

PR #149, opened 2015-07-06
Use c_long in ptrace fns for arch-independence

halorgium:


carllerche:
This has been merged.

PR #157, opened 2015-07-10
Add status to WaitStatus

joekain:
* Extend the enums in WaitStatus to include all process states (signaled,
  stopped, exited, continued).
* Decode status from waitpid
* Return appropate WaitStatus
* Update tests to use the new WaitStatus
* Add new tests for specific status values

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -2,6 +2,8 @@ use libc::{pid_t, c_int};
 use errno::Errno;
 use {Error, Result};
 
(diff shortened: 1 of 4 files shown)


joekain:
The BSD support is best effort, I can't manage to get things running at all on FreeBSD so I haven't been able to test it.

carllerche:
Thanks! Merged: c20401b51c2b54c1687b6c30edf20feb3e5eb411

PR #158, opened 2015-07-10
Fix unit tests as blank files can still have blocks

posix4e:
$ stat baz
  File: ‘baz’
  Size: 0         	Blocks: 16         IO Block: 4096   regular empty file
Device: 26h/38d	Inode: 6835152     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/ posix4e)   Gid: ( 1000/ posix4e)
Access: 2015-07-10 11:11:21.846851777 -0700
Modify: 2015-07-10 11:11:21.846851777 -0700
Change: 2015-07-10 11:11:21.846851777 -0700
 Birth: -


posix4e:
Also down with just adding a comment for confused people like me

PR #161, opened 2015-07-12
Remove unused import.

ltratt:
This is needed in order to compile nix-rust on 1.2 beta.


carllerche:
This seems to break the build on OS X: https://travis-ci.org/carllerche/nix-rust/builds/70635904 Though, CI does not include beta yet I think

ltratt:
Ah, yes, the problem was obvious once I saw the Travis error. I've fixed the problem and force pushed an update, so the pull request now builds on Travis.

carllerche:
Ooops, this got merged a while back but I forgot to close this :)

a384b1d7be2d634c4a564e70c77af3a466782f75

PR #162, opened 2015-07-13
added mq_getattr for the mqueue modul

MarkusJais:
This needs a careful review. It took me a while to write this without the test crashed with Signal 4.

Please have a look at these 2 lines in particular (line 102/103)

let mut attr = Box::new(MqAttr { mq_flags: 0, mq_maxmsg: 0, mq_msgsize: 0, mq_curmsgs: 0 });
let res = unsafe { ffi::mq_getattr(mqd, &mut *attr) };

Is this correct? I read on the docs that Box::new should be used. In the C code a pointer to the struct is passed. I first had is stack allocated but that lead to all the crashes. See the issue ticked I opened and later closed for more information. 

https://github.com/carllerche/nix-rust/issues/142

If this works, Rust users won't have to deal with this low level C stuff :-)

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -33,23 +33,25 @@ mod consts {
 }
 
 mod ffi {
-    use libc::{c_char, size_t, ssize_t, c_uint, c_int, mode_t};
+    use libc::{c_char, size_t, ssize_t, c_uint, c_int};
     use super::MQd;
     use super::MqAttr;
 
-    extern {
-        pub fn mq_open(name: *const c_char, oflag: c_int, mode: mode_t, attr: *const MqAttr) -> MQd;
+    extern "C" {
+        pub fn mq_open(name: *const c_char, oflag: c_int, ...) -> MQd;
 
-        pub fn mq_close (mqdes: MQd) -> c_int;
+        pub fn mq_close (mqd: MQd) -> c_int;
 
-        pub fn mq_receive (mqdes: MQd, msg_ptr: *const c_char, msg_len: size_t, msq_prio: *const c_uint) -> ssize_t;
+        pub fn mq_receive (mqd: MQd, msg_ptr: *const c_char, msg_len: size_t, msq_prio: *const c_uint) -> ssize_t;
 
-        pub fn mq_send (mqdes: MQd, msg_ptr: *const c_char, msg_len: size_t, msq_prio: c_uint) -> c_int;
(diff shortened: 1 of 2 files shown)


posborne:
Use of `Box` and the heap should not be necessary here and is likely only covering up a deeper issue.  I looked at this a bit and was able to reproduce the problem you saw.  I think the issue is that your definition of `MqAttr` does not match libc's in some cases.

I looked at a few implementations and one thing that shows up is that many of them include padding[1] that is not accounted for in your structure.  On the stack, if this padding was written out it would clobber something else on the stack.  On the heap... maybe it doesn't clobber anything obvious.

This was certainly what was going on with my system...

```
$ cat /usr/include/x86_64-linux-gnu/bits/mqueue.h
...
struct mq_attr
{
  __syscall_slong_t mq_flags;	/* Message queue flags.  */
  __syscall_slong_t mq_maxmsg;	/* Maximum number of messages.  */
  __syscall_slong_t mq_msgsize;	/* Maximum message size.  */
  __syscall_slong_t mq_curmsgs;	/* Number of messages currently queued.  */
  __syscall_slong_t __pad[4];
};
```

I'm not sure what the best solution is in general, but I think the problem exists for both heap and stack allocations.

MarkusJais:
Thanks. That makes a lot of sense. I agree that in that case Box won't help.
What OS did you use? Linux? Which version?

If the __pad[4] is not available on all Linux/libc versions it might get complicated. One could make different versions of the struct for different Linux version but if this padding is an implementation detail that might change with libc versions it will be difficult.

I will do some digging. Thanks for pointing this out.

MarkusJais:
/usr/src/linux-headers-3.16.0-28/include/uapi/linux/mqueue.h 
has a similar struct:

struct mq_attr {
    __kernel_long_t mq_flags;   /* message queue flags          */
    __kernel_long_t mq_maxmsg;  /* maximum number of messages       */
    __kernel_long_t mq_msgsize; /* maximum message size         */
    __kernel_long_t mq_curmsgs; /* number of messages currently queued  */
    __kernel_long_t __reserved[4];  /* ignored for input, zeroed for output */
};

I assume that __syscall_slong_t and __kernel_long_t are the same size and equivalent of a c_long in Rust but I wonder if there are any guarantees.

MarkusJais:
Seems like someone had a similar problem in Python:

http://permalink.gmane.org/gmane.comp.python.ctypes/4029

If that __pad[4] is always there on Linux it would work when I change the struct. I will git it a try.

MarkusJais:
I'll close this one as the code is broken. I may already have a fix. Next PR soon.

PR #163, opened 2015-07-18
added mq_getattr for Linux (2nd try)

MarkusJais:
This is the 2nd attempt for mq_getattr which hopefully solves the broken stack problem.

Now again without Box and stack allocated.

I have added a c_long[4] array to the MqAttr as this is there on Linux systems. Without it the memory is corrupt and some unit tests result in an error with Signal 4 (which is very bad!).

See 
/usr/include/linux/mqueue.h

Again, please carefully review this. mq_open is now also changed to take varargs (...) which is how it works in Linux.
(/usr/include/mqueue.h)

If this is ok, added the remaining mqueue functions should be a lot easier and I will be able to complete the mqueue support for nix-rust.

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -33,29 +33,39 @@ mod consts {
 }
 
 mod ffi {
-    use libc::{c_char, size_t, ssize_t, c_uint, c_int, mode_t};
+    use libc::{c_char, size_t, ssize_t, c_uint, c_int};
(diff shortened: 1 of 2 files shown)


carllerche:
Looks good as far as I can tell. Thanks :+1:

MarkusJais:
Thanks for merging. Working in the missing functions.

PR #164, opened 2015-07-18
Fix deprecated item

steveklabnik:
warning: use of deprecated item: replaced with std::os::unix::fs::symlink and std::os::windows::fs::{symlink_file, symlink_dir}, #[warn(deprecated)] on by default

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ exclude     = [
 
 eventfd = []
 execvpe = []
+1_0 = []
 
 [dependencies]
 libc     = "0.1.8"
--- a/test/test_stat.rs
+++ b/test/test_stat.rs
@@ -1,4 +1,4 @@
-use std::fs;
+use std::os::unix::fs;
 use std::str;
 
 use libc::consts::os::posix88;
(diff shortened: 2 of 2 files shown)


joekain:
I tried a similar fix (in my fork) and ran into the same failure with Rust 1.0.  Is there some way to conditionally compile based on Rust version or standard lib version?

steveklabnik:
Ahhh yeah. I'll modify this...

steveklabnik:
this is one way to do this, though travis will still fail without the flag.

@carllerche what do you think?

carllerche:
I'm not really a fan of having to pass a feature flag to support older versions of stable Rust. IMO Rust should provide a built in version cfg or something.

steveklabnik:
Yeah, it's true.

joekain:
I'm also not a fan of having to pass the feature flag but I guess at this point Rust 1.0.0 and Rust 1.1.0 couldn't retroactively add a version cfg.

carllerche:
@joekain True, which is indicative that is an important thing to add!

carllerche:
Closing for now. Not going to take a feature flag for this.

PR #168, opened 2015-07-30
Fix eventfd -- switch from fcntl::fd to RawFd.

jamwt:
Build was broken when eventfd feature is enabled.


jamwt:
Okay, this got merged elsewhere.

PR #171, opened 2015-08-10
Add missing exec functions

nikklassen:
I also wonder if it would be worthwhile to change the env array of string to take in something compatible with the return value of std::env::vars like Vec<(CString, CString)> (or Vec<(T, T)> where T: Into<CString>).  I also personally find it awkward to create CStrings for this interface, so I think it would make sense to take a T: Into<Vec<u8>> like CString's new function does.  But I'd like to get some input.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -5,7 +5,7 @@ use errno::Errno;
 use fcntl::{fcntl, OFlag, O_NONBLOCK, O_CLOEXEC, FD_CLOEXEC};
 use fcntl::FcntlArg::{F_SETFD, F_SETFL};
 use libc::{c_char, c_void, c_int, size_t, pid_t, off_t};
-use std::{mem, ptr};
+use std::mem;
 use std::ffi::CString;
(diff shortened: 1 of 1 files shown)


nikklassen:
It would also be nice to add an interface for setting the environ variable, but I don't know what the best signature would be for that function.

carllerche:
It's tricky. I think the exact guidelines for what level APIs are included in nix hasn't been strictly defined yet.

My current implicit goal is to provide safe APIs that are zero cost (or, as close to zero cost as possible). The current goal is to not necessarily provide "pretty" APIs. However, I have been debating whether this would be a worthwhile effort. Aka, try to reorganize the lib to have both simple & low level APIs.

carllerche:
Merged, thanks! d1d345514a381e56385fde9eca987893b8be4c17

nikklassen:
I agree that you should be consistent with whatever you chose, low level vs simple.  I'm happy to help with simplifying the APIs if that's what you chose to go with.

PR #172, opened 2015-08-13
Additional Ioctl Changes on Top of Original Branch

posborne:
This branch includes my additional changes to close out https://github.com/carllerche/nix-rust/pull/147 originally opened by @cmr.

The changes were rebased on master and a few commits were added to add remaining portions discussed during the course of the review.


posborne:
In order to further test these changes I started to port the code I have in my rust-spidev library to use the new interface.  I found a couple bugs and made some improvements off of that effort.  I still need to test these changes on hardware, so more changes may be on there way based on that testing.

https://github.com/posborne/rust-spidev/compare/nix-ioctl-changes

carllerche:
Do you want me to merge this now or wait a bit?

posborne:
I'll try to get some cycles on my Pi and Beaglebone with the updated spidev code tonight to verify that all appears to be well and then I think it should be good to merge.

posborne:
I was able to test out my changes on hardware this evening and after fixing one more bug in my own port, everything seems to be working.

The one area where things are not as they should be (but for which I do not have an easy solution) is with the spidev transfer call.  This ioctl is encoded as a write (`iow!()`), however, the kernel can modify the buffer depending on the contents of the buffer itself.  With the current macros, it is assumed that an ioctl that is declared with `write buf` should take a `*const $ty` when it is actually mutating the data pointed to by the pointer in this case.

Since the code is unsafe and rust has no idea what the kernel is doing, everything seemed to work OK, but I am somewhat concerned that the mutation of the buffer could lead to possible memory safety issues in the future (maybe with future optimizations in the compiler/llvm).

This was avoided in the previous API as creating the op and performing logic read/writes/readwrites with the op were logically separate.

With all that said, I think I am OK with merging and we can try to figure out some way to solve this problem in the future.

PR #173, opened 2015-08-16
Add flock(2) support

agrover:
Adds basic support for calling flock().

I could see going further and adding the equivalent of std::sync::Mutex's MutexGuard for automatic unlocking at the end of scope. Not sure if that would fit with the rather low-level nature of Nix, though.


carllerche:
Indeed, determining the scope of nix is tricky...

Anyway, thanks for the PR!

PR #175, opened 2015-08-22
Fix signature of linux `clone` function

tailhook:
In fact function returns pid of new process (or tid of new thread)

The signature of the function change, but it's only return value changed from returning nothing to `pid_t` so I believe that I believe that the amount of code it may break is low.

On the other hand I don't even understand how to use `clone` that does *not* return id of newly created process (may be useful for threads?)

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -1,5 +1,5 @@
 use std::mem;
-use libc::{c_int, c_uint, c_void, c_ulong};
+use libc::{c_int, c_uint, c_void, c_ulong, pid_t};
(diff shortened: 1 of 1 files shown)


tailhook:
@carllerche, any issues with this? The patch is quite trivial.

carllerche:
Sorry, been trying to keep up w/ mio. Looking at this now.

carllerche:
seeeems good.

tailhook:
No problem. Thanks!

PR #176, opened 2015-08-23
Add memfd functionality

diwic:
Hi!

I've done some experiments with [memfds](https://dvdhrm.wordpress.com/2014/06/10/memfd_create2/) lately, and it seems to work just fine here. I think it would be a good addition to nix and I've tried to follow existing nix conventions.


carllerche:
Thanks! Merged: 4f69bc358d7ab644ae766fde8bb5bdfd631f789b

PR #182, opened 2015-08-27
Fix the struct packing of EpollEvent

### LINKED ISSUE #181: epoll does not work on Raspberry Pi
Hi,

I'm facing problems with epoll on Raspberry Pi. I've figured out it's due to different size of the EpollEvent structure. The structure defined in rust comes out packed but the native structure is not packed (i.e. it's 16 bytes wide). In my opinion, it's common pattern for the whole arm architecture, so the code in `sys/epoll.rs` should be changed to something like this:

```rust
#[cfg(target_arch = "arm")]
#[derive(Clone, Copy)]
#[repr(C)]
pub struct EpollEvent {
    ...
}

#[cfg(target_arch = "arm")]
#[test]
fn test_epoll_event_size() {
    ...
}

#[cfg(not(target_arch = "arm"))]
#[derive(Clone, Copy)]
#[repr(C, packed)]
pub struct EpollEvent {
    ...
}
```

I can send you a pull request if you want.

alex-gulyas:
According to the [Linux source code](https://github.com/torvalds/linux/blob/95803812cfb3ece8ee1bb3747611ead48300fca2/include/uapi/linux/eventpoll.h#L47-L62), EpollEvent should only use packing on x86_64.

Fixes #181.

### DIFF
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -50,19 +50,26 @@ pub struct EpollEvent {
     pub data: u64
 }
 
-#[cfg(not(target_arch = "x86_64"))]
-#[test]
-fn test_epoll_event_size() {
-    use std::mem::size_of;
-    assert_eq!(size_of::<EpollEvent>(), 16);
-}
-
 #[cfg(target_arch = "x86_64")]
 #[derive(Clone, Copy)]
 #[repr(C, packed)]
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks! merged: c6cad77d10836f12ecda7bee5fb7a57ee6ea4f89

PR #183, opened 2015-08-31
added support and tests for mq_setattr

MarkusJais:
now finally the PR for mq_setattr.

On Linux only O_NONBLOCK can be set.

If the PR is ok I'll add convenience methods for setting and removing that flag to make it clearer and easier.


MarkusJais:
looks like it fails on a 32 machine as I am casting something to i64 (long in C).
I'll have a look how to fix this.

MarkusJais:
Now it builds on 32bit systems as well.

Next steps:
- adding convenience methods for setting O_NONBLOCK that are easier to use than mq_setattr
- adding documentation

carllerche:
Merged, thanks! fad667e4e2c72889b55bcc7a21d87fa72218e101

PR #185, opened 2015-09-01
Ioctl cleanup

posborne:
This set of changes addresses the inability to expose a public interface by making generating functions for `ioctl!` public.  It also includes documentation improvements and some code cleanup.

I have some other ideas for ways to enhance the ioctl! macros to support other common use cases better but am holding off on them for the moment.

### DIFF
--- a/src/sys/ioctl/mod.rs
+++ b/src/sys/ioctl/mod.rs
@@ -30,13 +30,61 @@
 //! What does this module support?
 //! ===============================
 //!
-//! This library provides the `ioctl!` macro, for binding `ioctl`s. It also tries
-//! to bind every `ioctl` supported by the system with said macro, but
-//! some `ioctl`s requires some amount of manual work (usually by
-//! providing `struct` declaration) that this library does not support yet.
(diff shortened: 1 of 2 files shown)


carllerche:
This PR doesn't seem to be passing CI atm.

posborne:
The doctest was missing a `cfg` guard as it will only pass under Linux.  That has been added now (and I rebased on master), so it should pass now.

The biggest change here is making `ioctl!` generate public functions now as was brought up in https://github.com/carllerche/nix-rust/issues/184.  @jmemson does bring up a good point that in most cases you will not really want to directly export unsafe code.  I think the public probably makes sense as then you have the option to export or not via the use of a module namespace as is  shown in the example that is added.

posborne:
@carllerche, this one should be ready for review again.

carllerche:
looks good to me! I will merge.

carllerche:
Merged: 0fd0337e0f55b0f5b976c22bbe97fe49a2875bf0

PR #186, opened 2015-09-03
Add signalfd support

alex-gulyas:
This PR adds support for `signalfd`, a Linux exclusive method to receive signals via a file descriptor.

Also includes:
- added support for blocking signals in `sys::signal`: `signalfd` requires blocking the signals from their normal signal handlers
- added some helper functions to `sys::signal::SigSet`: querying set membership, signal blocking helpers, etc...
- unit tests for signal blocking and `signalfd` reading

Tested with Linux + Rust 1.2 on: x86, x86_64, arm(raspberry pi)

If and when this gets merged, I will add support for signal handling via signalfd + epoll to `mio`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ exclude     = [
 ]
 
 [features]
-
 eventfd = []
 execvpe = []
 
@@ -30,6 +29,11 @@ path    = "nix-test"
 version = "*"
 
 [[test]]
-
 name = "test"
 path = "test/test.rs"
+
+[[test]]
+name = "test-signalfd"
+path = "test/test_signalfd.rs"
+harness = false
+test = true
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -17,6 +17,9 @@ pub mod ioctl;
 
 pub mod signal;
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub mod signalfd;
+
 pub mod socket;
 
 pub mod stat;
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -158,7 +158,7 @@ pub mod signal {
     );
 
     bitflags!{
-        flags HowFlag: libc:c_int {
+        flags HowFlag: libc::c_int {
             const SIG_BLOCK   = 1,
             const SIG_UNBLOCK = 2,
             const SIG_SETMASK = 3,
@@ -237,7 +237,7 @@ pub mod signal {
     );
 
     bitflags!{
-        flags HowFlag: libc:c_int {
+        flags HowFlag: libc::c_int {
             const SIG_BLOCK   = 1,
             const SIG_UNBLOCK = 2,
             const SIG_SETMASK = 3,
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -288,20 +288,22 @@ pub mod signal {
 }
 
 mod ffi {
-    use libc;
+    use libc::{c_int, pid_t};
     use super::signal::{sigaction, sigset_t};
 
     #[allow(improper_ctypes)]
     extern {
-        pub fn sigaction(signum: libc::c_int,
+        pub fn sigaction(signum: c_int,
                          act: *const sigaction,
-                         oldact: *mut sigaction) -> libc::c_int;
+                         oldact: *mut sigaction) -> c_int;
 
-        pub fn sigaddset(set: *mut sigset_t, signum: libc::c_int) -> libc::c_int;
-        pub fn sigdelset(set: *mut sigset_t, signum: libc::c_int) -> libc::c_int;
-        pub fn sigemptyset(set: *mut sigset_t) -> libc::c_int;
+        pub fn sigaddset(set: *mut sigset_t, signum: c_int) -> c_int;
+        pub fn sigdelset(set: *mut sigset_t, signum: c_int) -> c_int;
+        pub fn sigemptyset(set: *mut sigset_t) -> c_int;
+        pub fn sigfillset(set: *mut sigset_t) -> c_int;
+        pub fn sigismember(set: *const sigset_t, signum: c_int) -> c_int;
 
-        pub fn kill(pid: libc::pid_t, signum: libc::c_int) -> libc::c_int;
+        pub fn kill(pid: pid_t, signum: c_int) -> c_int;
     }
 }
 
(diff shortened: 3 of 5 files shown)


geofft:
You might want to look at https://github.com/carllerche/mio/issues/16 and https://ldpreload.com/blog/signalfd-is-useless . (If you beat me to actually implementing signals for mio I won't complain at all. I'd been planning to get to it in the next week or two, feel free to ping me on IRC if you want to coordinate.)

So I think that signalfd is fine as a notification mechanism that _at least one_ signal has been delivered on Linux (using a separate sigwaitinfo thread as a cross-platform fallback, but you might as well avoid the thread on Linux), but the actual data in the signalfd is not reliably useful. If you get, say, two SIGCHLDs before you're scheduled again, you only read one `struct signalfd_siginfo` out of the fd. So you can't actually reliably implement anything with that data: you have to `waitpid(-1, &status, WNOHANG)` in a loop until you've gotten all the children, and at that point, you got all the data you would have gotten out of the `struct signalfd_siginfo` anyway. A similar argument holds for just about all the other signals like `SIGIO` (you might have gotten multiple notifications), with the exception of the POSIX realtime signals, which do not coalesce.

If you have a use case for getting info out of `struct signalfd_siginfo` (e.g. you know you only have one child, or one file sending `SIGIO`, or whatever), sure, nix is a low-level binding, but my guess is that it's not actually useful to do anything other than report the existence of a signal, and report the data that's useful for POSIX realtime signals. Any patch adding support to mio should actively avoid assuming that the data in `struct signalfd_siginfo` (or `struct siginfo`, for that matter, the same arguments apply) is complete.

alex-gulyas:
Hi!

First of all, thanks for pointing out these issues, I wasn't aware of the true complexity of unix signal handling.

My plan for the mio signalfd support is much simpler than the ones mentioned in carllerche/mio#16. I didn't plan to add a new method to the `Handler` trait, instead I'm going to implement `Evented` for the `SignalFd` struct. This way mio is not responsible for any signal related tasks, you just create and register the fd with the event loop, and then read from it once it becomes readable, like you do with sockets. Blocking the signals is also the users responsibility.

My use case is a long running daemon which can listen for a few signals to reload configuration, shutdown gracefully, etc... This daemon uses tcp sockets to communicate with it's clients, and since I already had an epoll event loop for that, I decided to implement this simple way of signal handling.

Now I understand that this is not the proper way to implement signal handling, and I don't mind if this will not get merged, but it works for my use case, and I regrettably don't have the time to fix it. Once you or somebody else implements it correctly, I will probably port my app to use the correct version, but until then I have to work with what I already have.

geofft:
Ah! Yeah, if you're not doing realtime signals with data (which _do_ work properly, but it's a very niche use case) and you're just doing hand-delivered signals for configuration, then you don't care about coalescing, but you also don't care about data. In that case, I'd vaguely suggest not exposing new things in `siginfo` or any of the data fields in `signalfd_siginfo` besides `ssi_siginfo`, so that there aren't any implicit promises about their validity/usefulness, and leaving a comment about it. (I'm also curious who the existing user of the members of `siginfo` is, but I have half a suspicion that it's libstd and this was copied from libstd....)

I am pretty sure it's true that you can just make signals an event on all platforms instead of adding a method to `Handler`, and that does seem like a cleaner design.

alex-gulyas:
As per your suggestion, I expanded the documentation to warn about signal discarding.

However, I did not remove the fields from `signalfd_siginfo`, because the reliability issues are caused by the the Linux signal implementation, and not by this PR. As I see it, the nix crate offers low level wrappers around system functions, but does not offer higher level guarantees about them, therefore there is no reason to remove potential functionality, even if it is not usable in every (or most) use case(s).

**On the mio signal handling:**
If I've understood correctly, there is no way to make signal handling 100% reliable on Linux, unless you eliminate the possibility to receive signals faster than you can process them\*, or Linux starts to support queuing multiple signals of the same type.

I think if a program needs a reliable comm. channel on a platform that doesn't support queuing multiple signals of the same type, it shouldn't use signals. And once this unreliability is accepted, I think it should be safe to add signalfd support to mio without adding higher level guarantees. (Obviously not without proper documentation.)

\* If you expect signals from 10 child processes, do you create 10 threads listening for them? Even if you assume that signals will be distributed evenly between the threads, so one thread could watch 2 children (1 signal processed + 1 pending), you will still need O(n/2) threads, and this only works if no child is expected to send out more than one signal in a short period of time.

geofft:
> As I see it, the nix crate offers low level wrappers around system functions, but does not offer higher level guarantees about them, therefore there is no reason to remove potential functionality, even if it is not usable in every (or most) use case(s).

Sure, that's reasonable. I care more that mio (or an external library for mio) exposes a hard-to-misuse interface.

(Looking harder I think it's possible to use the realtime signals for lots of use cases, too, in which case coalescing doesn't happen.)

> If you expect signals from 10 child processes, do you create 10 threads listening for them?

What you do is you treat signals as an edge-triggered interface. If you are signaled with `SIGCHLD` (whether that's through an actual signal handler, through `sigwaitinfo` etc., or through `signalfd`), what you know is that _at least one_ child has exited since the last time you were scheduled. While you can get a `siginfo` or `signalfd_siginfo` out of the initial delivery, it only informs you about one of the children. There may be more, and you won't get renotified for the other children. So you have to loop through them all anyway, and at that point, knowing `si_pid` / `ssi_pid` wasn't particularly helpful. So on receiving `SIGCHLD`, you call `waitpid(-1, &status, WNOHANG)` in a loop until it stops returning children. (Or preferably you loop over each child you're intentionally tracking, so that if some library code is secretly spawning a child without telling you, you don't break it.) But you certainly can reap all the children from a single thread / mainloop.

Lots of the other interesting siginfo fields _can't_ even be received through `signalfd`, as far as I can tell. In particular, all the synchronous signals (the ones that want to terminate your program, like `SIGSEGV`, `SIGSYS`, `SIGFPE`, etc.) can't usefully be delayed until you get around to reading from a `signalfd`, since there's no way for the program to continue. If you want to fix up a segfault by doing some userspace paging trickery, you have to do it from a signal handler as an interrupt. (And they're directed to the faulting thread, so you can't even have another thread handle it.) So, as far as I can tell, `ssi_addr`, `ssi_trapno`, etc. can't possibly be set.

> I think if a program needs a reliable comm. channel on a platform that doesn't support queuing multiple signals of the same type, it shouldn't use signals. And once this unreliability is accepted, I think it should be safe to add signalfd support to mio without adding higher level guarantees. (Obviously not without proper documentation.)

:+1: (though you can arguably use realtime signals if you really want)

geofft:
Looked at the code (since @carllerche asked): the bindings look correct and I'm happy to see this merged. The only thing is that Android API levels below 21 (Lollipop) have a handful of these functions as inlines in the header files, so we'll need to do something like rust-lang/rust@a8dbb92b471cae1d3f8225857f5553311dd8aeb3 for it to work there.

FWIW, I'm pretty convinced now that exposing all the fields in `signalfd_siginfo` is the right thing to do.

carllerche:
I rebased & merged. Thanks.

36432e1a13f4340a437e6250f4c665742439a5aa

PR #189, opened 2015-09-15
quotactl() API

polachok:
I'm fairly new to rust, so I'm looking for some early feedback on this API. Am I moving into right direction with this?

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -47,3 +47,10 @@ pub mod time;
               target_arch = "arm")),
           )]
 pub mod ptrace;
+
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86",
+              target_arch = "x86_64",
+              target_arch = "arm")),
+          )]
+pub mod quota;
--- a/src/sys/quota.rs
+++ b/src/sys/quota.rs
@@ -1,0 +1,126 @@
+use {Result, NixPath, from_ffi};
+use errno::Errno;
+use libc::{c_int, c_char};
+
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86",
+              target_arch = "x86_64",
+              target_arch = "arm")),
+          )]
+pub mod quota {
+	use libc::c_int;
+
+	pub struct QuotaCmd(pub QuotaSubCmd, pub QuotaType);
+	pub type QuotaSubCmd = c_int;
+
+	impl QuotaCmd {
+		pub fn as_int(&self) -> c_int {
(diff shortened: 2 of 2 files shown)


src/sys/quotactl.rs
```diff
@@ -0,0 +1,99 @@
+use {Result, NixPath, from_ffi};
+use errno::Errno;
+use libc::{c_int, c_char};
+
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86",
+              target_arch = "x86_64",
+              target_arch = "arm")),
+          )]
+pub mod quotactl {
```
carllerche: I am not sure why this module needs to be public.

carllerche:
Looks mostly fine to me. I added a minor nitpick. API style wise, nix doesn't really have a consistent guideline yet (something I am hoping to work on), what you came up with is in line with other APIs in nix though... so seems good :)

src/sys/mod.rs
```diff
@@ -47,3 +47,10 @@ pub mod time;
               target_arch = "arm")),
           )]
 pub mod ptrace;
+
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86",
+              target_arch = "x86_64",
+              target_arch = "arm")),
+          )]
+pub mod quotactl;
```
utkarshkukreti: This should probably be named `quota` as the corresponding C header file is `sys/quota.h`.

polachok: fixed

src/sys/quota.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+pub fn quotactl_get<P: ?Sized + NixPath>(which: quota::QuotaType, special: &P, id: c_int, dqblk: &mut quota::Dqblk) -> Result<()> {
+	use std::mem;
+	unsafe {
+		quotactl(quota::QuotaCmd::new(quota::Q_GETQUOTA, which), Some(special), id, mem::transmute(dqblk))
+	}
+}
+
+pub fn quotactl_set<P: ?Sized + NixPath>(which: quota::QuotaType, special: &P, id: c_int, dqblk: &mut quota::Dqblk) -> Result<()> {
```
polachok: quotactl() is not supposed to mutate dqblk here, would it be safe to pass a &quota::Dqblk and then transmute it?

polachok:
I think it's ready, what do you think?

carllerche:
Merged! Thanks (8da9b314e973e756c8384a6110fb85c899d64561)

PR #191, opened 2015-09-19
added convenience functions for setting O_NONBLOCK on message queues

MarkusJais:
Just 2 convenience functions to make setting and removing O_NONBLOCK on message queues easier. The mapping of the original C function is not very user friendly. These 2 new methods make it easier (and safer) to use.


carllerche:
Thanks! Merged: 5ef1f7acd3b325fc642aa4af6ff25a9dfd0ac618

PR #192, opened 2015-09-20
sys/event: add kevent_ts to better expose kevent timeout

luca-barbieri:
This allows to specify no timeout and allows to specify any timespec
timeout for kevent.


carllerche:
Merged! Thanks: c17d5b32a77e49bc2c53b21f25a7ebf9be607cba

PR #198, opened 2015-10-08
Fix handling of sockaddr_un lengths

### LINKED ISSUE #169: Doesn't support abstract UNIX sockets
To use a socket in the abstract namespace, one `bind()`s a socket to a path beginning with a NUL byte. Unfortunately, there's an [explicit test](https://github.com/carllerche/nix-rust/blob/master/src/lib.rs#L99) that fails on a NUL encountered in a path. Additionally, functions like `strcpy()` [are called](https://hoverbear.github.io/raft/src/nix/sys/socket/addr.rs.html#351) in parts of the library, and won't work if there's a NUL that's not the last byte in the string.

### LINKED ISSUE #177: getpeername() panics on AF_UNIX sockets because sockaddr_un varies in size
```rust
extern crate nix;
use nix::sys::socket::*;
use nix::unistd::*;

fn main() {
    let (fd1, fd2) = socketpair(AddressFamily::Unix, SockType::Stream, 0, SockFlag::empty()).unwrap();
    getpeername(fd1).unwrap();
}
```
does this:
```
geofft@titan:~/src/rust/scratch/t$ target/debug/t
thread '<main>' panicked at 'assertion failed: len as usize == mem::size_of::<sockaddr_un>()', /home/geofft/src/rust/nix-rust/src/sys/socket/mod.rs:721
```

The panic is in `sockaddr_storage_to_addr`:
```rust
        consts::AF_UNIX => {
            assert!(len as usize == mem::size_of::<sockaddr_un>());
            Ok(SockAddr::Unix(UnixAddr(*(addr as *const _ as *const sockaddr_un))))
        }
```

From some print statements, it looks like `len` is 2, whereas the size of the structure is 110. strace traces this as

```
getpeername(3, {sa_family=AF_LOCAL, NULL}, [2]) = 0
```

It looks like it's valid for a `sockaddr_un` to report a length shorter than the maximum size, and we're supposed to treat this as a flexible array member or something. The easiest answer may be for UnixAddr to explicitly track its own size, although I'm wondering if really SockAddr should track size for all variants.

Probably a prereq for / should be fixed alongside #169.

(Also, `getpeername` has no tests and probably should have at least one.)

geofft:
The returned length of AF_UNIX sockaddrs is significant, and generally
does not match the length of the entire structure. For filesystem
sockets, this is ignorable because the path is also NUL-terminated, but
for unbound sockets (e.g., a socketpair) or abstract-namespace
sockets (a Linux extension where the address is an arbitrary
bytestring), we need to keep track of the length.

Fixes #177. Also add a UnixAddr::new_abstract function and some better
handling of abstract-namespace socket addresses to fix #169.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -2,8 +2,8 @@ use {Result, Error, NixPath};
 use super::{consts, sa_family_t};
 use errno::Errno;
 use libc;
-use std::{fmt, hash, mem, net};
-use std::ffi::{CStr, OsStr};
+use std::{fmt, hash, mem, net, ptr};
+use std::ffi::OsStr;
 use std::path::Path;
 use std::os::unix::ffi::OsStrExt;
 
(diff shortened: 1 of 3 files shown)


carllerche:
Thanks! Merged ad87c3bdc99357f66af9260bce1eb6ae2cbfda98

PR #199, opened 2015-10-10
Add fsync, fdatasync; fix some comments

jsgf:
- add fsync, fdatasync
- fix some comments
- add implementations of `From` and `Error` for `nix::Error`


carllerche:
thanks!

PR #203, opened 2015-11-09
Bump .travis.yml to Rust 1.4.0

polachok:


carllerche:
I'm keeping the explicit Rust version at the lowest version of Rust that nix supports. That way, dropping support for an older version of Rust is explicit.

PR #210, opened 2015-12-03
update libc to 0.2.2

marjakm:
Nix is broken on nightly arm because of:
https://github.com/rust-lang/rust/pull/29775

This fixes nix for arm nightly, but breaks on arm stable. Should work on x86 stable and nightly.
Maybe this could be put into a separate nightly branch that could be pointed to in Cargo.toml?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,8 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc     = "0.1.12"
-bitflags = "0.3.2"
+libc     = "0.2.2"
+bitflags = "0.3.3"
 
 [dev-dependencies]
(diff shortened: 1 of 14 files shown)


carllerche:
Yuck, it looks like this breaks CI. What broke on ARM?

marjakm:
Here are the build errors:
https://gist.github.com/marjakm/792e126363ff07b9bd1a

libc c_char on ARM Linux was i8 but is now u8:
https://github.com/rust-lang/rust/issues/29774

I'll look into the errors on mac and 1.1.0, I only tested with linux and 1.4.0 before.

carllerche:
Closing this one in favor of #219. Thanks!

PR #202, opened 2015-11-08
[WIP] Netlink bits

polachok:
Makes it (almost) possible to create a netlink socket.
See #200 
Not really because socket() doesn't accept protocol value, and netlink requires it.

PR #211, opened 2015-12-03
socket(): add protocol argument (breaking change)

polachok:
This is unfortunate, but some socket types (ie netlink on linux) require protocol argument to be set.
See #202


carllerche:
Merged. Thanks. Sorry for the delay.

polachok:
Can we have #202 in now?

PR #217, opened 2015-12-18
add NSIG constant

fiveop:
glibc defines this constant as "the total number of signals defined.
Since the signal numbers are allocated consecutively, NSIG is also one
greater than the largest defined signal number."


carllerche:
Thanks! Merged eb9718625708a951a615a18b6583561937875fd0

PR #231, opened 2016-01-06
Bring back mount

kamalmarhubi:
Fixes https://github.com/carllerche/nix-rust/issues/85

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,8 @@ name = "test-signalfd"
 path = "test/test_signalfd.rs"
 harness = false
 test = true
+
+[[test]]
+name = "test-mount"
+path = "test/test_mount.rs"
+harness = false
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -210,6 +210,23 @@ impl NixPath for PathBuf {
     }
 }
 
+/// Treats `None` as an empty string.
+impl<'a, NP: ?Sized + NixPath>  NixPath for Option<&'a NP> {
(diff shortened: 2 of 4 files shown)


kamalmarhubi:
(Rebased)

carllerche:
Hmmm... it doesn't seem to pass CI. Unsure why.

kamalmarhubi:
Nightly I think is some nightly weirdness. 1.1.0 is because the test is using `Result::expect()`, which was added in 1.4.0. I'll update it to use `unwrap()` or `unwrap_or_else()` given the desire to test on 1.3.0.

kamalmarhubi:
@carllerche I've bashed Travis into submission*. Please check the test to see if what's going on makes sense. Thanks!

* well, Travis forced me to disable one of the tests in CI, but we'll make do.

carllerche:
Instead of re-implementing the test runner, is it possible to use `Once` (call once) construct to do initialization?

kamalmarhubi:
Sadly not. The `unshare` call must happen before any threads are spawned. The built-in runner does not have any way to get code to run before that.

kamalmarhubi:
Other alternatives involve doing weird things with fork / exec and exit statuses to communicate success or failure to the runner.

kamalmarhubi:
@carllerche ping for comments on the test runner

kamalmarhubi:
I'll merge this soon, leaving the "test runner" as is. I filed https://github.com/carllerche/nix-rust/issues/248 to track its improvement.

PR #232, opened 2016-01-09
Fix name of MNT_DETACH flag

kamalmarhubi:


src/mount.rs
```diff
@@ -43,7 +43,7 @@ bitflags!(
 bitflags!(
     flags MntFlags: c_int {
         const MNT_FORCE   = 1 << 0,
-        const MNT_DETATCH = 1 << 1,
+        const MNT_DETACH = 1 << 1,
```
matklad: It is probably a good idea to preserve alignment here.

kamalmarhubi: Ah yes. Made the change, thanks!

carllerche:
Thanks! Merged: 0fca8b360776403c292d1fc7742028eec711df70

PR #233, opened 2016-01-09
typo: MNT_DETATCH -> MNT_DETACH

matklad:
I believe the T should not be there: http://linux.die.net/man/2/umount2


matklad:
oups, this is a duplicate of https://github.com/carllerche/nix-rust/pull/232

PR #234, opened 2016-01-09
Add POSIX function raise()

fiveop:


fiveop:
Will reopen freshly rebased to master, so that travis runs again with all the previous fixes applied.

PR #235, opened 2016-01-09
Publish SockFlags from nix::sys::signal.

fiveop:
This brings them in alignment with other things republished from
the OS specific nix::sys::signal::signal.


carllerche:
Thanks. Merged: e295907ee684c075250b7204e98724cd7484a374

PR #237, opened 2016-01-13
ci: Track stable and beta in place of specific versions

kamalmarhubi:


carllerche:
Hey, I'm ok w/ adding stable & beta, but I would like to keep the `1.1.0`, which represents the oldest supported Rust. Maybe add a comment above `1.1.0` saying "Oldest version of Rust supported by nix"

I'm also willing to bump that to `1.3.0` :)

kamalmarhubi:
I'm happy to support any version, just so long as we make it explicit. When you say "oldest supported Rust" do you mean by Rust core team? Or by nix?

carllerche:
By nix. Running CI against that version (1.3.0) will hopefully ensure that nix doesn't break compat with rust 1.3.0.

Basically, I only want to bump to newer Rust versions explicitly (vs by accident) and if there is a good reason to do so.

kamalmarhubi:
I've changed this to test on all released versions 1.1.0+. The test should pass in about 45 minutes, at which point I'll merge it. :-)

PR #240, opened 2016-01-15
add aarch64 support

brayniac:
I would like to get aarch64 support merged in. It compiles successfully when targeting `aarch64-unknown-linux-gnu` Also tested through a project using mio which I have actually run on a machine using this architecture.


carllerche:
Thanks!

Sorry for the delay :)

Issue #49, opened 2015-01-17
sigaction struct on OS X


vbuslov:
I've been looking on signal-handling code in nix-rust to implement signal support in mio and it seems that sigaction struct definition for OS X is a bit off.

Relevant definition from 10.10:
```
/* union for signal handlers */
union __sigaction_u {
        void    (*__sa_handler)(int);
        void    (*__sa_sigaction)(int, struct __siginfo *,
                       void *);
};

/* Signal vector template for Kernel user boundary */
struct  __sigaction {
        union __sigaction_u __sigaction_u;  /* signal handler */
        void    (*sa_tramp)(void *, int, int, siginfo_t *, void *);
        sigset_t sa_mask;               /* signal mask to apply */
        int     sa_flags;               /* see signal options below */
};

/*                                                                                                                                                                              
 * Signal vector "template" used in sigaction call.                                                                                                                             
 */
struct  sigaction {
        union __sigaction_u __sigaction_u;  /* signal handler */
	sigset_t sa_mask;               /* signal mask to apply */
        int     sa_flags;               /* see signal options below */
};
```

struct definition from nix:
```
    #[cfg(any(target_os = "macos", target_os = "ios"))]
    #[repr(C)]
    #[allow(missing_copy_implementations)]
    pub struct sigaction {
        pub sa_handler: extern fn(libc::c_int),
        sa_tramp: *mut libc::c_void,
        pub sa_mask: sigset_t,
        pub sa_flags: SockFlag,
    }
```

So it seems that nix defines sigaction according to private kernel-user boundary OS X struct instead of public struct which actually used as param type to sigaction function.
Also I need the sa_sigaction(instead of sa_handler which is currently used in nix struct) handler to obtain siginfo_t structure and I'm struggling to define c-union in Rust. Any suggestions?

vhbit:
@vbuslov there is no way to define unions in Rust, in case of data with the same size (like pointer to a function) workaround is to use `mem::transmute`.

fiveop:
I have the same problem and was thinking of using one or two enums to encapsulate the differences between `sa_handler`, `sa_sigaction` as well as the difference between an actual function pointer and `SIG_IGN` and `SIG_DFL`. I'll work on a proposal.

carllerche:
Great! Looking forward to it.

fiveop:
[...]

PR #241, opened 2016-01-20
Account for the different signal handler types.

fiveop:
This commit would be an API breaking change: `nix::sys::signals::sigaction` get's a different signature.

It currently leaves the responsibility of passing the `SA_SIGINFO` flag if necessary to the user.

This might solve part of #49.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -464,6 +464,15 @@ impl AsRef<sigset_t> for SigSet {
     }
 }
 
+pub use self::signal::siginfo;
+
+pub enum SigHandler {
+    SigDfl,
+    SigIgn,
+    Handler(extern fn(SigNum)),
+    SigAction(extern fn(SigNum, *mut siginfo, *mut libc::c_void))
+}
+
 type sigaction_t = self::signal::sigaction;
 
 pub struct SigAction {
@@ -471,9 +480,14 @@ pub struct SigAction {
 }
 
 impl SigAction {
-    pub fn new(handler: extern fn(libc::c_int), flags: SockFlag, mask: SigSet) -> SigAction {
+    pub fn new(handler: SigHandler, flags: SockFlag, mask: SigSet) -> SigAction {
         let mut s = unsafe { mem::uninitialized::<sigaction_t>() };
-        s.sa_handler = handler;
+        s.sa_handler = match handler {
+            SigHandler::SigDfl => unsafe { mem::transmute(libc::SIG_DFL) },
+            SigHandler::SigIgn => unsafe { mem::transmute(libc::SIG_IGN) },
+            SigHandler::Handler(f) => f,
+            SigHandler::SigAction(f) => unsafe { mem::transmute(f) },
+        };
         s.sa_flags = flags;
         s.sa_mask = mask.sigset;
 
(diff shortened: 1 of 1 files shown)


fiveop:
Thinking more about this, the constructor function `new` is now unsafe. If you pass in a `extern fn(SigNum, *mut siginfo, *mut libc::c_void)` you also have to pass flags with `SA_SIGINFO` and for `extern fn(SigNum)` the flag must not be set.

I see two solutions: Either add/remove those flags everytime (just in case) change the return value to a Result and signal an error, if the flags are not appropriatly set. Both go against the idea of zero cost abstractions. However, we cannot provide a safe interface without them.

The third option is to mark the method unsafe and describe in a comment why. This whay the caller know what he has to do, to ensure safety.

kamalmarhubi:
I'm not following the reasoning behind not wanting to force the correct setting of `SA_SIGINFO` based on the enum variant. This seems like a perfect place for nix to help the user do the right thing.

fiveop:
My problem was (and still is) that it is no longer a zero cost abstraction. But all our error handling code already adds additional costs, so I opted to push a variant that sets or unsets the flag as necessary.

src/sys/signal.rs
```diff
@@ -471,10 +480,20 @@ pub struct SigAction {
...
-        s.sa_flags = flags;
+        s.sa_handler = match handler {
+            SigHandler::SigDfl => unsafe { mem::transmute(libc::SIG_DFL) },
+            SigHandler::SigIgn => unsafe { mem::transmute(libc::SIG_IGN) },
+            SigHandler::Handler(f) => f,
+            SigHandler::SigAction(f) => unsafe { mem::transmute(f) },
+        };
+        s.sa_flags = match handler {
+            SigHandler::SigAction(_) => flags | SA_SIGINFO,
+            _ => flags - SA_SIGINFO,
```
kamalmarhubi: I think `flags & !SA_SIGINFO` is nicer when removing a single flag. As it is, I had to go and double check that we actually had a bitflags type here.

kamalmarhubi:
Regarding "zero cost", I think your commit comment captures what that means. It's ok to put in a few extra instructions for the infrequent thing, but the actual calling of the signal handlers will have no cost above what you'd expect in C. This differs from, say, Go which handles signals by passing them to channels, or Python, where the signal handlers are Python functions and so there's some per-signal overhead there.

src/sys/signal.rs
fiveop: I prefer the minus sign, which represents set difference. And that is exactly what we want to express and achieve here. In addition it is more concise.
I do not think that preventing confusion is a valid argument since the type is stated in the parameter list and (now) by convention we use bitflags for such types.

fiveop:
Unless there are further concerns I would like to merge this, in particular, because I want to use it (and apparently others too).

src/sys/signal.rs
kamalmarhubi: That is a very reasonable position!

kamalmarhubi:
Sounds good to me!

fiveop:
Merged as 645c48f2890cd5ef993f6a73970a693e26ca5f91...645e37d22e5453d9aaf8634095a4150789e74b0e.

PR #245, opened 2016-01-25
Added if_nametoindex (and necessary module based on Cs net/if.h)

mbr:
Hopefully this crate is the right home for this snippet. I've tried to follow conventions, but I'm not sure I've gotten it right (especially the error handling). Please let me know if work needs to be done on this.

### DIFF
--- a/src/net/if_.rs
+++ b/src/net/if_.rs
@@ -5,10 +5,10 @@
 
 use libc::{c_uint, if_nametoindex};
 use std::ffi::{CString, NulError};
-use ::Error;
+use ::{Result, Error};
 
 /// Resolve an interface into a interface number.
-pub fn name_to_index(name: &str) -> Result<c_uint, Error> {
+pub fn name_to_index(name: &str) -> Result<c_uint> {
     let name = match CString::new(name) {
         Err(e) => match e { NulError(..) => {
             // A NulError indicates that a '\0' was found inside the string,
--- a/src/net/if_.rs
+++ b/src/net/if_.rs
@@ -5,40 +5,29 @@
 
 use libc::{c_uint, if_nametoindex};
 use std::ffi::{CString, NulError};
-use std::io;
-
-/// Error that can occur during interface name resolution.
-#[derive(Debug)]
-pub enum NameToIndexError {
-    /// Failed to allocate a C-style string to for the syscall
-    NulError,
-    IOError(io::Error),
-}
-
-impl From<NulError> for NameToIndexError {
-    fn from(_: NulError) -> NameToIndexError {
-        NameToIndexError::NulError
-    }
-}
-
-impl From<io::Error> for NameToIndexError {
-    fn from(e: io::Error) -> NameToIndexError {
-        NameToIndexError::IOError(e)
-    }
-}
+use ::Error;
 
 /// Resolve an interface into a interface number.
-pub fn name_to_index(name: &str) -> Result<c_uint, NameToIndexError> {
-    let name = try!(CString::new(name));
+pub fn name_to_index(name: &str) -> Result<c_uint, Error> {
+    let name = match CString::new(name) {
+        Err(e) => match e { NulError(..) => {
+            // A NulError indicates that a '\0' was found inside the string,
+            // making it impossible to create valid C-String. To avoid having
+            // to create a new error type for this rather rare case,
+            // nix::Error's invalid_argument() constructor is used.
+            //
+            // We match the NulError individually here to ensure to avoid
+            // accidentally returning invalid_argument() for errors other than
+            // NulError (which currently don't exist).
+            return Err(Error::invalid_argument());
+        }},
+        Ok(s) => s
+    };
 
     let if_index;
     unsafe {
         if_index = if_nametoindex(name.as_ptr());
     }
 
-    if if_index == 0 {
-        return Err(NameToIndexError::from(io::Error::last_os_error()));
-    }
-
-    Ok(if_index)
+    if if_index == 0 { Err(Error::last()) } else { Ok(if_index) }
 }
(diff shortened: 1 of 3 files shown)


kamalmarhubi:
Hi! Thanks for the PR. I think this is a good place for something like that to live. I'd also like to see the reverse, but I think that can wait until it's added to libc.

src/net/if_.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl From<io::Error> for NameToIndexError {
+    fn from(e: io::Error) -> NameToIndexError {
+        NameToIndexError::IOError(e)
+    }
+}
+
+extern {
+    fn if_nametoindex(ifname: *const c_char) -> c_uint;
```
kamalmarhubi: use this from libc instead of making our own exter: http://rust-lang-nursery.github.io/libc/x86_64-unknown-linux-gnu/libc/fn.if_nametoindex.html

src/net/if_.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::ffi::{CString, NulError};
+use std::io;
+
+/// Error that can occur during interface name resolution.
+#[derive(Debug)]
+pub enum NameToIndexError {
+    /// Failed to allocate a C-style string to for the syscall
+    NulError,
+    IOError(io::Error),
+}
```
kamalmarhubi: can you use [`nix:Error`](http://rustdoc.s3-website-us-east-1.amazonaws.com/nix/master/linux/nix/enum.Error.html) instead of your own error?

src/net/if_.rs
```diff
@@ -0,0 +1,48 @@
...
+        NameToIndexError::IOError(e)
+    }
+}
+
+extern {
+    fn if_nametoindex(ifname: *const c_char) -> c_uint;
+}
+
+/// Resolve an interface into a interface number.
+pub fn name_to_index(name: &str) -> Result<c_uint, NameToIndexError> {
```
kamalmarhubi: A few things:

* use the functions name as in C
* return a `nix::Result<c_uint>`, which is equivalent to swapping in `nix::Error` for the error type
* use a [`NixPath`](http://rustdoc.s3-website-us-east-1.amazonaws.com/nix/master/linux/nix/trait.NixPath.html) bound for the string argument (the name is bad, and it'll likely change soon, see #221 if you want more info!)

This would look like:

``` rust
pub fn if_nametoindex<S: ?Sized + NixPath>(target: &P) -> Result<c_uint> {
    let res = try!(name.with_nix_path(|name| { /* rest of body here */ }));

    from_ffi(res)
}
```

The [`nix::from_ffi()`](http://rustdoc.s3-website-us-east-1.amazonaws.com/nix/master/linux/nix/fn.from_ffi.html) function handles errno translation for you. :-)

src/net/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod if_;
```
kamalmarhubi: could you add a tiny comment explaining the name is to avoid the keyword issue? (I'm assuming that's why, anyway)

kamalmarhubi:
FYI I filed https://github.com/rust-lang-nursery/libc/pull/153 to add `if_indextoname` to libc.

src/net/if_.rs
mbr: Damn, I could've sworn I already replaced it. Fixed in 8972270

src/net/if_.rs
mbr: I looked into it before implementing my own, but there's the NulError to take care of. I believe it may be worthwhile to add that to nix::Error though? It's documented in https://doc.rust-lang.org/std/ffi/struct.NulError.html.

Alternatives are just returning not found if a NulError occured or, as a last resort, unpacking (I'd like to avoid that).

mbr: A fourth option is now in 71c6836: Just returned Error::invalid_argument() if a string containig a null-byte is passed.

src/net/if_.rs
mbr: > A few things:
> 
> use the function's name as in C

Fixed in 2cb7878

> return a nix::Result<c_uint>, which is equivalent to swapping in nix::Error for the error type

Fixed in 5ef9689

> use a NixPath bound for the string argument (the name is bad, and it'll likely change soon, see #221 if you want more info!)

That currently seems a little messy and missing an implementation for ``&str``? Interface names are not Paths and I don't want to require passing in CStrings, because at that point it stops being a Rust-Interface, I think.

> The nix::from_ffi() function handles errno translation for you. :-)

If i am reading this right, won't ``from_ffi()`` just return ``()`` ? At least the compiler seems to agree with me. Additionally, it only works for ``c_int``, but ``if_nametoindex`` returns unsigned values.

mbr:
That concludes my changes. The code looks much cleaner and shorter now. I would appreciate it though if you could take care of the NixPath-conversion.

src/net/if_.rs
kamalmarhubi: > That currently seems a little messy and missing an implementation for &str? Interface names are not Paths and I don't want to require passing in CStrings, because at that point it stops being a Rust-Interface, I think.

This is a good consideration. I've copied your note over to to #221 where we're discussing string handling in nix.

kamalmarhubi:
This is great! I've rebased and fixed the conflicts that arose, added the NixPath stuff, and made it useable on non-Linux as well. Currently waiting for CI to go green before merging. I also sent out #250 to address the `str` issue you mentioned with `NixPath`.

Thanks again!

kamalmarhubi:
Since it's not showing up here, the CI for the patch set I'm planning to merge: https://travis-ci.org/kamalmarhubi/nix-rust

kamalmarhubi:
CI went green, merged as: https://github.com/carllerche/nix-rust/compare/e8f014d...beec2f4 :-)

mbr:
Perfekt, thanks! =) Now I just need to fix the typo in the function documentation ("a" => "an")... =)

PR #246, opened 2016-01-26
Normalize tabs to spaces

arcnmx:

PR #247, opened 2016-01-26
Errno::result()

arcnmx:
A replacement of `from_ffi`, and updates all methods to use it. Partial preparation for #230 where `Error` will be replaced by `Errno`.

Changes are on top of #246, so merge that first!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,8 +18,9 @@ extern crate libc;
 #[cfg(test)]
 extern crate nix_test as nixtest;
 
-// Re-export some libc constants
+// Re-exports
 pub use libc::{c_int, c_void};
+pub use errno::{Errno, Result};
 
 pub mod errno;
 pub mod features;
@@ -42,12 +43,12 @@ pub mod unistd;
 
 /*
  *
- * ===== Result / Error =====
+ * ===== Error =====
  *
  */
 
 use libc::c_char;
-use std::{ptr, result};
+use std::ptr;
 use std::ffi::CStr;
 use std::path::{Path, PathBuf};
 use std::os::unix::ffi::OsStrExt;
@@ -55,8 +56,6 @@ use std::io;
 use std::fmt;
 use std::error;
 use libc::PATH_MAX;
-
-pub type Result<T> = result::Result<T, Error>;
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum Error {
@@ -184,13 +183,4 @@ impl NixPath for PathBuf {
     fn with_nix_path<T, F>(&self, f: F) -> Result<T> where F: FnOnce(&CStr) -> T {
         self.as_os_str().as_bytes().with_nix_path(f)
     }
-}
-
-#[inline]
-pub fn from_ffi(res: libc::c_int) -> Result<()> {
-    if res != 0 {
-        return Err(Error::Sys(errno::Errno::last()));
-    }
-
-    Ok(())
 }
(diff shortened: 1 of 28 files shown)


kamalmarhubi:
I'm going to take a look at this tomorrow. I also think we'll want input from an extra reviewer or two since it's a major change to the API.

arcnmx:
Well, to be fair the only API change is renaming `from_ffr` to `Errno::result` :P

But yeah, it touches a lot of the codebase! Not sure who else to pull in.

kamalmarhubi:
> Well, to be fair the only API change is renaming `from_ffi` to `Errno::result` :P

I thought this was changing the return type of every function in the library from `std::result::Result<_, nix::NixError>` to `std::result::Result<_, nix::errno::Errno>`? Am I confused?

arcnmx:
No, this doesn't change anything externally (except `from_ffi`, which is a public function). Signatures are otherwise all the same, and `nix::Error` is still used because this doesn't include the `NixString` changes (and therefore `InvalidPath` is still an applicable error).

src/lib.rs
```diff
@@ -56,8 +57,6 @@ use std::fmt;
 use std::error;
 use libc::PATH_MAX;
 
-pub type Result<T> = result::Result<T, Error>;
```
kamalmarhubi: My confusion stemmed from moving this into the `errno` module. What's the rationale?

kamalmarhubi:
Ah, I was confused. See inline comment https://github.com/carllerche/nix-rust/pull/247#discussion_r50795350

arcnmx:
Essentially, this is just a PR to change every function to use `from_ffi`. It also just happens to rename it to `Errno::result()`, for two reasons:

1. It's more self-documenting, etc. It returns a `nix::Result` from a POSIX function return value containing an `Errno`.
2. It prepares for the change that actually does swap `nix::Result` to `Result<T, Errno>`. None of the code this touches will need to be changed as part of #230 (which will actually change the function signatures).

src/lib.rs
arcnmx: Oh, it should probably be moved back. Mostly it's there because that's where #230 had it.

kamalmarhubi:
Thanks for the clarification!

src/lib.rs
kamalmarhubi: Mind moving it back? :-)

arcnmx: Yeah, I'll do it when I get a chance sometime tomorrow night. Don't let that keep you from reviewing the rest of it though!

kamalmarhubi: Based on what you said, and when I'm more awake, I'm pretty sure this will be fine. I'd been meaning to do a `from_ffi` sweep soon, so...

arcnmx:
Moved `Result` back.

kamalmarhubi:
CI failed on apple. Please apply:

```diff
diff --git a/src/sys/event.rs b/src/sys/event.rs
index eb99f7d..1563a74 100644
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -1,7 +1,8 @@
 /* TOOD: Implement for other kqueue based systems
  */
 
-use errno::{Errno, Result};
+use Result;
+use errno::Errno;
 #[cfg(not(target_os = "netbsd"))]
 use libc::{timespec, time_t, c_int, c_long, uintptr_t};
 #[cfg(target_os = "netbsd")]
```

:-)

kamalmarhubi:
Ok I'm going over this once more and then let's get the damned thing merged!

posborne:
Did a quick review.  I'm happy with this change.

kamalmarhubi:
Rebased and merged as https://github.com/carllerche/nix-rust/compare/01e8416...f167e8f

PR #249, opened 2016-01-27
IPv6 addresses are always in big endian order

jedisct1:
No matter what the architecture is, IPv6 addresses are always in big endian order in the C structure.

Parsing IPv6 addresses used to be broken in Nix on little-endian architectures. For example, `[::1]` resulted in address `[::100]` being effectively stored.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -280,21 +280,13 @@ pub struct Ipv6Addr(pub libc::in6_addr);
 
 macro_rules! to_u8_array {
     ($($num:ident),*) => {
-        if cfg!(target_endian = "big") {
-            [ $(($num>>8) as u8, ($num&0xff) as u8,)* ]
-        } else {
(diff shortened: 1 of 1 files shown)


fiveop:
I suggest adding a comment that refers to [RFC 1700](https://tools.ietf.org/html/rfc1700).

kamalmarhubi:
@fiveop what's the relevant section?

fiveop:
If I understand it correctly, it specifies that all IP related RFCs are to picture data in big endian order, which results big endian becoming the 'network byte order'. See for example the linux man page on ipv6. I'll add a corresponding comment and commit this.

kamalmarhubi:
:+1:

fiveop:
Merged as 7502db05f562daac1e353fd51d9072b81571f8fb...645c48f2890cd5ef993f6a73970a693e26ca5f91.

PR #250, opened 2016-01-27
Implement NixPath for str and OsStr

kamalmarhubi:
This is a stop gap improvement until the NixPath reform is figured out.

refs #221

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,7 +50,7 @@ pub mod unistd;
 
 use libc::c_char;
 use std::{ptr, result};
-use std::ffi::CStr;
+use std::ffi::{CStr, OsStr};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
@arcnmx any thoughts?

I realized on the way to this that we're currently doing a terrible job of interior NULs. There's nothing new on that front in this PR, as you can have interior NULs in a `Path` as easily as you can in a `str`.

arcnmx:
Seems fine to me!

Interior nuls should probably trigger invalidpath errors, just like they do in `libstd`. This is ignored for all types, so yeah, this PR doesn't change the status quo. It's something to do as part of the `NixString` reform for sure.

kamalmarhubi:
Great, rebased, waiting for CI then will merge.

Issue #58, opened 2015-02-02
Several `const`s should be `bitflags`


utkarshkukreti:
Some random examples (there are many more):

* [MmapFlag](https://github.com/carllerche/nix-rust/blob/9a17a4d200b33f66a6d61dc06a9df7ddef01d735/src/sys/mman.rs#L13-L32)
* [MmapProt](https://github.com/carllerche/nix-rust/blob/9a17a4d200b33f66a6d61dc06a9df7ddef01d735/src/sys/mman.rs#L34-L41)
* [SockMessageFlags](https://github.com/carllerche/nix-rust/blob/9a17a4d200b33f66a6d61dc06a9df7ddef01d735/src/sys/socket.rs#L221-L225)

carllerche:
Looks like you are correct :)

fiveop:
After #243 only flags in `src/sys/socket/consts.rs` should remain to be converted.

kamalmarhubi:
@fiveop are you working on the remaining flags?

fiveop:
I think we can close this now.

kamalmarhubi:
@fiveop did you convert the ones in the original issue, or go over all flags? If the former, I think we could keep this open pending an audit to make sure we're using bitflags all over.

fiveop:
I am pretty sure that I found all of them, but please do check again.

kamalmarhubi:
@fiveop skimmed all the source, looks good!

PR #256, opened 2016-01-29
Convert MSG_ flags to bitflags! type.

fiveop:
This should be all the flags for #58.


kamalmarhubi:
Merged as https://github.com/carllerche/nix-rust/compare/e04842e...9835270

PR #262, opened 2016-02-09
signal: Allow raw_pointer_derive on SigAction struct

kamalmarhubi:
This will fix the build on pre-1.6.0 Rust.


fiveop:
I am sorry, that I just checked against my local rust. However, this brings up the question (again), whether it would be sufficient to just support stable and beta (and nightly to a certain degree).

kamalmarhubi:
@fiveop no worries! It's a process issue more than anything else that is slipped in. :-)

PR #266, opened 2016-02-12
linux: Add sendfile(2)

kamalmarhubi:


kamalmarhubi:
r? @arcnmx 
since why not

kamalmarhubi:
@homu r? @arcnmx

kamalmarhubi:
hrm

kamalmarhubi:
Oh, r? is part of rust-highfive I think.

kamalmarhubi:
r+

kamalmarhubi:
@homu r+

kamalmarhubi:
@homu r-

kamalmarhubi:
woo it works

kamalmarhubi:
@arcnmx would you mind taking a look? :-)

arcnmx:
Where's test_sendfile? o:

kamalmarhubi:
oh dear rebase fail

kamalmarhubi:
@arcnmx fixed

arcnmx:
Pretty straightforward, looks good! My only comment is that the test would deadlock on a system without buffered pipes, but that's not the case on Linux so..!

@homu r+

arcnmx:
Fancy, it used an older test result o:

kamalmarhubi:
Cool!

@arcnmx out of curiosity, what common systems don't have buffered pipes?

arcnmx:
No idea, probably none, but it's not specified so it's somewhat like relying on implementation details.

> Applications should not rely on a particular capacity: an application should be designed so that a reading process consumes data as soon as it is available, so that a writing process does not remain blocked.

PR #270, opened 2016-02-19
Improvements and new syscalls

abbradar:
This most likely should have been split for the review, but I don't see how except for maybe splitting each commit into its own PR -- each change is independent of other ones.

Issue #190, opened 2015-09-15
Figure out the future of nix


carllerche:
**tl;dr** Should nix be split up into multiple crates named `nix-*` and what should the API style be?

## The situation

Nix originated as a posix binding dumping ground for various crates that I had been working on. I tended to need the same bindings over and over again, so I just moved them into Nix. I thought this would be useful for other people, so I accepted PRs for various system APIs as they have been submitted.

Everything has been fine so far, but a few issues have arisen.

As the API coverage has increased, whenever a breaking change is made, the semver must be incremented. So, if the `ioctl` module changes an API as it figures out the best way to expose functionality, the version must be bumped and downstream crates must change their dependencies. For example, [Mio](//github.com/carllerche/mio) depends on nix, but not on any `ioctl` features, so bumping the dependency is pretty annoying.

Secondly, the API style has been somewhat ad hoc, especially with more complicated system APIs like `fcntl`.

## Crate structure

There is value in having a single place (nix) to find system APIs that have a coherent API style and a single documentation site. I think that each specific API silo, like sockets, ioctl, mqueue, signals, etc... could be in their own crate named `nix_*`. Common features like errors, `NixPath`, etc... would live in `nix_core`. The `nix` crate would be a facade depending on all the sub crates with liberal dependencies (probably `"*"`) which would allow the consumer to optionally be able to pin each sub crate to whatever.

Libraries like Mio could then depend only on the features that are needed.

Also, it would allow delegating responsibility better. I, for one, know nothing about `ioctl`. I would like to stop being the blocker.

## API style

This is still something that I haven't figured out. I do know that the goal of nix is **comprehensive, safe, zero cost bindings to system APIs  on *nix OSes**.

So, what does that mean for nix? As of now, there are currently a variety of styles. For example, most APIs are 1-1 bindings to the equivalent OS api. This leads to APIs like [`fcntl`](https://github.com/carllerche/nix-rust/blob/master/src/fcntl.rs#L97-L122) which uses a convoluted `FcntlArg` strategy. [`ioctl`](https://github.com/carllerche/nix-rust/blob/master/src/sys/ioctl/mod.rs#L66-L77) on the other hand uses a (fairly impressive) macro...

The question is, would it be better to expand beyond the 1-1 rule. For example, in the `fnctl` case, the API would most likely be much cleaner and more "rusty" if each "fcntl arg" was moved to an individual function. There could, for example, be a public `fcntl` mod with a `dupfd` function. Using it would be:

```rust
[...]

PR #273, opened 2016-02-19
wait: support ptrace events for Linux

abbradar:
Part of #270 -- see https://github.com/abbradar/nix-rust/commit/b2d3b4afa07bd97e4b842ffc0be57cf0ed3e636b for previous discussion.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -42,7 +42,7 @@ const WSTOPPED: WaitPidFlag = WUNTRACED;
 pub enum WaitStatus {
     Exited(pid_t, i8),
     Signaled(pid_t, signal::SigNum, bool),
-    Stopped(pid_t, signal::SigNum),
+    Stopped(pid_t, signal::SigNum, c_int),
     Continued(pid_t),
     StillAlive
 }
@@ -51,6 +51,7 @@ pub enum WaitStatus {
           target_os = "android"))]
 mod status {
     use sys::signal;
+    use libc::c_int;
 
     pub fn exited(status: i32) -> bool {
         (status & 0x7F) == 0
@@ -80,6 +81,10 @@ mod status {
         ((status & 0xFF00) >> 8) as signal::SigNum
     }
 
+    pub fn stop_additional(status: i32) -> c_int {
+        (status >> 16) as c_int
+    }
+
     pub fn continued(status: i32) -> bool {
         status == 0xFFFF
     }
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Copying part of my comment on the previous discussion:

> I'm also a bit put off by adding a meaningless member on non-Linux systems. Trying to think of alternatives...

I'm going to hold off on this for a bit while I think about stuff related to #190. :-)

kamalmarhubi:
Coming back to this, I'd be happier if the extra variant and methods were `cfg`'ed to only be present on Linux, or whichever systems support it.

luser:
I was thinking about this again, and how to handle this in a way that isn't terrible across platforms. What would you think about adding another entry to the `WaitStatus` enum, like `WaitStatus::PtraceEvent`? Then the `if status::stopped` case could check for this and return that value. It'd still be useless on non-Linux systems, but it seems like it'd be easier to work with a platform-conditional enum variant than a conditional field *inside* an enum variant. `#[cfg(...)]` seems to work OK on enum variants and match arms:
https://play.rust-lang.org/?gist=15519135f09722bf6d8653912592c079&version=stable&backtrace=0

abbradar:
Sigh, I _still_ haven't returned to this (haven't actually got a chance to touch Rust for a long time already). I understand that this can be frustrating, and also -- if anyone wants to continue this work please do so!

I hope I can return to this someday...

kamalmarhubi:
@abbradar this is not frustrating at all! And thanks for your contributions :-)

kamalmarhubi:
@luser 
> What would you think about adding another entry to the `WaitStatus` enum, like `WaitStatus::PtraceEvent`? 

Could you sketch out what you mean a tiny bit?

luser:
Oh, I can do you one better! I actually wrote the patch because I wanted to use it for something, I just forgot to do anything with it. Here it is:
https://github.com/luser/nix/commit/92a3cf77265e7402c555ab77d0ae3e3920914d36

Here's the code I wrote that uses it:
https://github.com/luser/tracetree/blob/master/src/main.rs#L65

I didn't actually try to make that code portable yet, but making it build should be straightforward. (Making it actually work right is probably harder).

Susurrus:
Is this still relevant? If so, @abbradar would you be willing to update this patch?

luser:
This was superseded by https://github.com/nix-rust/nix/pull/438

You can see an example of using it here: https://github.com/luser/tracetree/blob/2dcde59a9b2bd8dd6343a147a6d09f5d4428068e/src/main.rs#L141

Susurrus:
Okay, then I'm going to close this. Thanks, @luser.

PR #280, opened 2016-02-26
Add sigwait

0x1997:


src/sys/signal.rs
```diff
@@ -370,6 +372,15 @@ impl SigSet {
         try!(pthread_sigmask(how, Some(self), Some(&mut oldmask)));
         Ok(oldmask)
     }
+
+    /// Suspends execution of the calling thread until one of the signals in the
+    /// signal mask becomes pending, and returns the accepted signal.
+    pub fn wait(&self) -> Result<SigNum> {
+        let mut signum: SigNum = unsafe { mem::uninitialized() };
```
fiveop: Why don't you just initialize it to `0`? SigNum is just an alias for an integer type.

posborne: What is present is valid and marginally more performant than initializing to zero.  It is analogous to what would be done in C.  We just need a spot on the stack that the kernel will write to.

kamalmarhubi: I'm not sure if I like the use of `mem::uninitialized()`, but it's using the pattern already prevalent in the file. Given that, I think it's fine, and could be revisited file-wide. Thoughts @fiveop?

fiveop: It makes sense. I propose adding a note to CONVENTIONS stating that we want to use `std::mem::uninitialized` in such cases.

fiveop:
@homu r+

PR #283, opened 2016-02-28
Bumped bitflags version to 0.4 and removed warning

KalitaAlexey:
Bumped bitflags version to 0.4 and removed warning about derive with raw pointer


fiveop:
I suggest splitting this PR into two, because bumping the bitflags versions should be mergeable immediately.

KalitaAlexey:
@fiveop I removed last commit.

fiveop:
@homu r+

Issue #226, opened 2015-12-29
Export types from function signatures, or libc in general


viraptor:
I've found that there's a number of types like pid_t which are either function results or arguments, but they're not explicitly exported. This means either I have to depend on libc myself and track the version nix depends on, or use a hacky solution (pid_t is available via nix::sys::ioctl::libc::pid_t).

It would be great if the types used in public signatures were either available through nix in a more straightforward way, or if the libc was available via nix::libc.

kamalmarhubi:
Many of them are in `sys/types.h` ([POSIX spec](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html)), so exporting at least those ones under `nix::sys::types` seems reasonable.

carllerche:
This should be fixed... However, the exact strategy is still TBD (#190).

If an FFI type is used in a function signature, it should be exported by libc. However, nix may want to create its own types and wrap the libc FFI types in order to provide stricter type safety as well as custom functions on the types.

viraptor:
The problem with types exported by libc is that cargo is a bit silly with pulling multiple libraries. (https://github.com/rust-lang/cargo/issues/2064)

This results in "it's T, but not the right version of T" errors unless the local cargo.toml is kept in sync with nix's cargo.toml all the time. If nix forwarded just the plain pid_t (and other affected types), it would make the maintenance easier, because libc wouldn't have to be specified anymore.

kamalmarhubi:
> nix forwarded just the plain pid_t

What do you mean by this? I'm not all caught up on how the multiple versions of a Crate stuff works.

viraptor:
It's hard to explain in practice beyond "bad things happen if you don't ensure there's only one libc"... I'll try to create a minimal example.

fiveop:
Suppose we have two versions of libc (A and B). Further assume that we have to libraries: foo and bar.
foo uses libc version A and bar, while bar uses libc version B.

Now if bar exports a function like this:

```rust
extern crate libc;
use libc;
fn baz(libc::c_int) {}
```
and foo uses it as follows

```rust
extern crate libc;
extern crate bar;
fn call_baz() {
   let a : libc::c_int = 3;
   baz(a);
}
```
then you'd get an error. Because `a` has type `libc::c_int of version A`, while `baz` expects an argument of type `libc::c_int of version B`.

[...]

PR #284, opened 2016-02-28
Make crate libc import public.

fiveop:
On currently nightly I got an error regarding the public export of libc in sys::ioctl. It now requires us to import the extern crate publicly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,7 +13,7 @@
 #[macro_use]
 extern crate bitflags;
 
-extern crate libc;
+pub extern crate libc;
 
 #[cfg(test)]
 extern crate nix_test as nixtest;


kamalmarhubi:
I don't fully understand the implications of this. Seems related to #226?

posborne:
II'm OK with this change and I don't see another way to solve the problem while keeping the `ioctl` stuff the way it is now (using macros).

fiveop:
The other option is not exporting libc from `ioctl` and apparently that breaks things.

posborne:
Yet another option would be to change the `nix::sys::ioctl` to use functions that are part of the nix API rather than generating code that directly uses libc.  I am somewhat partial to this solution and maybe eliminating the macros completely.

Then again, I'm not sure how many users are depending on the macros.  I think they were an interesting experiment but I'm not sure they really make sense in nix and certainly seem to clash with the other exposed APIs to some degree.

fiveop:
For now, I'd like to get the build working again.

For the discussion on whether to export libc or not we already have #226.

posborne:
Fair enough, @homu r+

PR #288, opened 2016-03-04
libc: re-export libc properly

posborne:
With rust 1.7, the following warning was being emitted by the compiler:

    warning: `pub extern crate` does not work as expected and should
    not be used. Likely to become an error. Prefer `extern crate` and
    `pub use`.

Based on https://github.com/rust-lang/rust/issues/26775 it appears that
this change is the current best approach for properly exporting the
libc dependency so it may be used outside of nix.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,12 +16,17 @@ extern crate bitflags;
 #[macro_use]
 extern crate cfg_if;
 
-pub extern crate libc;
-
 #[cfg(test)]
 extern crate nix_test as nixtest;
 
-// Re-exports
+// In rust 1.8+ this should be `pub extern crate libc` but prior
+// to https://github.com/rust-lang/rust/issues/26775 being resolved
+// it is necessary to get a little creative.
+pub mod libc {
+    extern crate libc;
+    pub use self::libc::*;
+}
+
 pub use libc::{c_int, c_void};
 pub use errno::Errno;
 


posborne:
Relates to https://github.com/nix-rust/nix/pull/284.

r? @fiveop

fiveop:
That is confusing. We just added the `pub` to `extern crate` because of an error in nightly, with a message stating that we cannot `pub use` a crate, that is private. The CI for this PR fails for exactly this reason in beta and nightly, now.

We need to figure out how to properly reexport a library in bea/nightly.

posborne:
Agreed.

src/lib.rs
```diff
@@ -13,12 +13,15 @@
...
-
 #[cfg(test)]
 extern crate nix_test as nixtest;
 
+mod exports {
+    extern crate libc;
+}
+
 // Re-exports
+pub use exports::libc;
```
kamalmarhubi: can you not do

```
extern crate libc;
pub use libc;
```

? I think I've done that before...

posborne:
Ok, I believe what is present now should result in the correct behavior prior to the recent changes coming out of https://github.com/rust-lang/rust/issues/26775 and should also have the same behavior for beta/nightly.

kamalmarhubi:
Looks good here, assuming it works. I'll let @fiveop take another look though.

fiveop:
@homu r+

PR #291, opened 2016-03-05
conventions: Codify use of newtypes

kamalmarhubi:


CONVENTIONS.md
```diff
@@ -19,7 +19,16 @@ We use the functions exported from [libc][libc] instead of writing our own
...
 
 We use the `struct` definitions from [libc][libc] internally instead of writing
-our own.
+our own. If we want to add methods to a libc type, we use the newtype pattern.
+For example,
+
+```rust
+pub struct SigSet(libc::sigset_t);
+
+impl SigSet {
```
kamalmarhubi: I'm not sure what the right thing to suggest for naming standards is though. Should this be Sigset?

posborne: `SigSet` feels right to me in this case.

fiveop: I agree. So, let's add another sentence that mentions that we use the rust convention of camel case names for structs. Then we can merge this PR.

kamalmarhubi: how's the change?

kamalmarhubi:
One thing I've wondered about is

```rust
pub struct SigSet(libc::sigset_t);
```

vs

```rust
pub struct SigSet {
    inner: libc::sigset_t,
}
```

There's also a possibility of introducing a something like the Rust project's `*Inner` traits: https://github.com/rust-lang/rust/blob/master/src/libstd/sys/common/mod.rs#L54-L76

CONVENTIONS.md
posborne: Change looks good to me.

kamalmarhubi:
Deferring choice of how to represent the newtype. @homu r=@posborne

Issue #251, opened 2016-01-27
Investigate flaky tests: test_wait_signal, test_wait, test_execve


kamalmarhubi:
I've seen these fail occasionally
- `sys::test_wait::test_wait_signal`
- `test_unistd::test_wait`
- `test_unistd::test_execve`

Maybe some others

kamalmarhubi:
Including output from @viraptor's report in #260:

```
test test_unistd::test_execve ... FAILED
test test_unistd::test_wait ... FAILED

failures:

---- test_unistd::test_execve stdout ----
        thread 'test_unistd::test_execve' panicked at 'called `Result::unwrap()` on an `Err` value: Sys(ECHILD)', ../src/libcore/result.rs:741

---- test_unistd::test_wait stdout ----
        thread 'test_unistd::test_wait' panicked at 'assertion failed: `(left == right)` (left: `Ok(Exited(2837, 0))`, right: `Ok(Exited(2843, 0))`)', test/test_unistd.rs:41


failures:
    test_unistd::test_execve
    test_unistd::test_wait
```

kamalmarhubi:
new flaky test:

```
test sys::signal::tests::test_sigwait ... Process didn't exit successfully: `/Users/travis/build/kamalmarhubi/nix-rust/target/x86_64-apple-darwin/debug/nix-3f7e4935a20dbf98` (signal: 30)
```

kamalmarhubi:
@posborne noted in #292:
> I was mostly looking at the test_unistd failures which make calls out
to fork() and then make subsequent calls to wait(). In that case there
is one parent and the wait() called could (and frequently does) get some
random child pid back because it just happened to terminate. That is
why when one of the test fails so does the other one.

posborne:
test_sigwait still seems to fail sometimes even with the single threaded execution.  Must be a separate problem.

kamalmarhubi:
Thoughts on these issues:

> I was mostly looking at the test_unistd failures which make calls out
to fork() and then make subsequent calls to wait(). In that case there
is one parent and the wait() called could (and frequently does) get some
random child pid back because it just happened to terminate.

I'm going to make them wait on the child they spawn instead of any old child. I'm still not sure what is going on in `test_sigwait`. I couldn't see any failures on my linux machine when running single-threaded.

dhylands:
Those errors (mentioned on Jan 31) look like the ones I saw as well.

kamalmarhubi:
[...]

PR #292, opened 2016-03-05
testing: increase stability by removing thread parallelism

posborne:
Currently, several of the tests are failing intermittently.  After
some research it appears that these failures only occur when thread
parallelism is enabled (as is the case by default).  To test, I just
ran the failing tests over and over.  I would consistently see errors
when running the following:

    $ while true; do target/debug/test-7ec4d9681e812f6a; done

When I forced single threaded execution, I no longer saw failures:

    $ while true; do RUST_TEST_THREADS=1 target/debug/test-7ec4d9681e812f6a; done

I was mostly looking at the test_unistd failures which make calls out
to fork() and then make subsequent calls to wait().  In that case there
is one parent and the wait() called could (and frequently does) get some
random child pid back because it just happened to terminate.  That is
why when one of the test fails so does the other one.

I couldn't think of an obvious fix other than preventing thread
parallelism in the short term.  The tests still run very quickly.

https://github.com/nix-rust/nix/issues/251

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ os:
   - osx
 
 env:
-  - ARCH=x86_64
-  - ARCH=i686
+  - RUST_TEST_THREADS=1 ARCH=x86_64
+  - RUST_TEST_THREADS=1 ARCH=i686
 
 script:
   - curl -sSL https://raw.githubusercontent.com/carllerche/travis-rust-matrix/master/test | RUST_TEST_THREADS=1 bash


kamalmarhubi:
This is at least a good stopgap for spurious CI failures.

@homu r+

kamalmarhubi:
> I was mostly looking at the test_unistd failures which make calls out
to fork() and then make subsequent calls to wait(). In that case there
is one parent and the wait() called could (and frequently does) get some
random child pid back because it just happened to terminate. That is
why when one of the test fails so does the other one.

This is great to know. I hadn't looked at *why* there were failures.

posborne:
Agreed that this still doesn't feel ideal.  Unit tests are a pretty difficult thing to achieve for libs like this (and calling the actual fork() is obviously not a "good" test in the traditional sense).  For testing C code, I have made use of linker flags to wrap actual system calls with mocks but I haven't been able to think of a good, portable way of handling this in rust.

A generic mocking utilty for testing could be pretty useful across the ecosystem if we can think of a good way of handling.

http://stackoverflow.com/questions/3662856/how-to-reimplement-or-wrap-a-syscall-function-in-linux

PR #297, opened 2016-03-07
Docker cross test infrastructure

posborne:
These commits contain the first portion of new infrastructure that will allow us to build for and test on a much larger set of platforms.  Nix and the tests are *not* compiling on many platforms right now, but I believe at this point that most of those problems are issues with nix (or libc) rather than the test infrastructure.

I want to get this out there so we can start filing issues and other people can start doing testing and development to resolve the issues popping up on various platforms.  None of this is integrated in with travis at this point -- that will be phase 3.  Phase 2 will be getting qemu going to run the tests for non-host platforms (this might be tricky when running in a container as `/dev/kvm` might need to be passed through).

This is based on the testing infrastructure in libc but modified quite a bit -- The libc infrastructure doesn't seem to be a great experience for testing locally.  If this approach works out, we might want to push this approach toward libc.

The Dockerfile/Image that goes along with this is currently here:
* https://github.com/posborne/docker-rust-cross
* https://hub.docker.com/r/posborne/rust-cross/

I would have no problem moving this into the org if that makes sense down the line.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,35 @@
----
+language: rust
 sudo: required
+dist: trusty
 services:
   - docker
-language: rust
+
 rust:
   - 1.1.0  # Oldest supported version
-  - 1.2.0
-  - 1.3.0
-  - 1.4.0
-  - 1.5.0
-  - 1.6.0
-  - 1.7.0
   - stable
   - beta
   - nightly
 
 script:
-  - sh ci/run-travis.sh
+  - bash ci/run-travis.sh
+
+env:
+  - ARCH=x86_64
 
 os:
   - linux
   - osx
 
-env:
-  - ARCH=x86_64
-  - ARCH=i686
-
 # Failures on nightly shouldn't fail the overall build.
 matrix:
   fast_finish: true
   include:
+    - os: linux
+      env: ARCH=i686
+      rust: stable
+    - os: osx
+      env: ARCH=i686
+      rust: stable
     - os: linux
       env: TARGET=aarch64-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:arm
       rust: 1.7.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 ---
-sudo: false
+sudo: required
+services:
+  - docker
 language: rust
 rust:
   - 1.1.0  # Oldest supported version
@@ -7,31 +9,49 @@ rust:
   - 1.3.0
   - 1.4.0
   - 1.5.0
+  - 1.6.0
+  - 1.7.0
   - stable
   - beta
   - nightly
 
-# Failures on nightly shouldn't fail the overall build.
-matrix:
-  allow_failures:
-    - rust: nightly
+script:
+  - sh ci/run-travis.sh
 
 os:
   - linux
   - osx
 
 env:
-  - RUST_TEST_THREADS=1 ARCH=x86_64
-  - RUST_TEST_THREADS=1 ARCH=i686
+  - ARCH=x86_64
+  - ARCH=i686
 
-script:
-  - curl -sSL https://raw.githubusercontent.com/carllerche/travis-rust-matrix/master/test | RUST_TEST_THREADS=1 bash
-  - cargo doc --no-deps
-
-addons:
-  apt:
-    packages:
-      - gcc-multilib
+# Failures on nightly shouldn't fail the overall build.
+matrix:
+  fast_finish: true
+  include:
+    - os: linux
+      env: TARGET=aarch64-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:arm
+      rust: 1.7.0
+    - os: linux
+      env: TARGET=arm-unknown-linux-gnueabihf DOCKER_IMAGE=posborne/rust-cross:arm
+      rust: 1.7.0
+    - os: linux
+      env: TARGET=mips-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:mips
+      rust: 1.7.0
+    - os: linux
+      env: TARGET=mipsel-unknwon-linux-gnu DOCKER_IMAGE=posborne/rust-cross:mips
+      rust: 1.7.0
+    - os: linux
+      env: TARGET=arm-linux-androideabi DOCKER_IMAGE=posborne/rust-cross:android
+      rust: 1.7.0
+  allow_failures:
+    - rust: nightly
+    - env: TARGET=aarch64-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:arm
+    - env: TARGET=arm-unknown-linux-gnueabihf DOCKER_IMAGE=posborne/rust-cross:arm
+    - env: TARGET=mips-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:mips
+    - env: TARGET=mipsel-unknwon-linux-gnu DOCKER_IMAGE=posborne/rust-cross:mips
+    - env: TARGET=arm-linux-androideabi DOCKER_IMAGE=posborne/rust-cross:android
 
 # Deploy documentation to S3 for specific branches. At some
 # point, it would be nice to also support building docs for
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc     = "0.2.8"
+libc     = { git = "https://github.com/rust-lang/libc.git" }
 bitflags = "0.4"
 cfg-if = "0.1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc     = { git = "https://github.com/rust-lang/libc.git" }
+libc     = "0.2.8"
 bitflags = "0.4"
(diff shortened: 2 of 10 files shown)


kamalmarhubi:
I am really excited to look at this!

kamalmarhubi:
> (this might be tricky when running in a container as /dev/kvm might need to be passed through)

Could we prefer software emulation? That's what libc does for freebsd / netbsd / android on ARM. Travis can't even give us /dev/kvm if they wanted to: it's not available on GCE where they run. Or on EC2 for that matter...

ci/run.sh
```diff
@@ -0,0 +1,124 @@
...
+
+echo "======================================================="
+echo "TESTING VERSION: ${VERSION}, TARGET: ${TARGET}"
+echo "======================================================="
+
+#
+# Tell cargo what linker to use and whatever else is required
+#
+configure_cargo() {
+  rm -rf .cargo
```
kamalmarhubi: this seems potentially distructive

ci/run.sh
```diff
@@ -0,0 +1,124 @@
...
+
+# This is a hack as we cannot currently
+# ask cargo what test files it generated:
+# https://github.com/rust-lang/cargo/issues/1924
+find_binaries() {
+  target_base_dir="${BUILD_DIR}/${TARGET}/debug"
+
+  # find [[test]] sections and print the first line and
+  # hack it to what we want from there.  Also "nix" for
+  # tests that are implicitly prsent
```
kamalmarhubi: nit: present

ci/run.sh
```diff
@@ -0,0 +1,124 @@
...
+  for test_base in $( awk '/\[\[test\]\]/{getline; print}' "${MANIFEST_PATH}" | \
```
kamalmarhubi: you might be able to use `cargo read-manifest` and JSON processing here instead of fake-parsing the manifest

kamalmarhubi:
Couple of comments so far. I am at a locale with slow internet, and so I can't play with it because pulling the image will take a while.

ci/run.sh
posborne: Yes.   Currently, this is only executed in the container (and not in a directory that is part of a shared volume) -- May need to re-evaluate depending on how things evolve.

posborne:
>> (this might be tricky when running in a container as /dev/kvm might need to be passed through)
> 
> Could we prefer software emulation? That's what libc does for freebsd / netbsd / android on ARM. Travis can't even give us /dev/kvm if they wanted to: it's not available on GCE where they run. Or on EC2 for that matter...

I think we can.  I am just getting around to actually testing the qemu bits and will give it a good try.  I'l have some time free next week (changing day jobs and have a week off), so I should have some time to round out testing and get things ready to go for travis.

ci/run.sh
kamalmarhubi: I guess I mean that if I accidentally run this from the top level, and I have some cargo config, I will lose it. since I keep those out of git, I would then be sad.

kamalmarhubi: eg, I'd rather save the existing one, and have a trap to put the original back on exit.

kamalmarhubi:
>  I am just getting around to actually testing the qemu bits and will give it a good try.

:+1: 

@posborne do you want to get this merged sooner than later?

posborne:
@kamalmarhubi I am going to take one more pass at it, but then I would like to get something into master so people can start testing/fixing on various issues.  Phase 2 will be figuring out how best to integrate it with CI -- We will probably need to dump travis-rust-matrix, but that has been showing some strain lately anyhow.

I'm rebuilding the docker image right now as several dependencies were missing previously (the image is quite large now as the qemu-* packages are not small).

posborne:
Ok, things seem to be working, so I think this is ready for a review.  I have the extra platforms tested set to not fail things right now until we get things sorted.  As we fix things, we can remove those versions from the `allow_failures` set.

r? @kamalmarhubi @fiveop

Cargo.toml
```diff
@@ -20,7 +20,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc     = "0.2.8"
+libc     = { git = "https://github.com/rust-lang/libc.git" }
 bitflags = "0.4"
```
fiveop: Why this change in this PR?

.travis.yml
```diff
@@ -1,37 +1,57 @@
...
   - linux
   - osx
 
-env:
-  - RUST_TEST_THREADS=1 ARCH=x86_64
-  - RUST_TEST_THREADS=1 ARCH=i686
-
-script:
-  - curl -sSL https://raw.githubusercontent.com/carllerche/travis-rust-matrix/master/test | RUST_TEST_THREADS=1 bash
-  - cargo doc --no-deps
```
fiveop: Did this deploy the docs somewhere? Or just build them?

Cargo.toml
posborne: Ooops, I will remove this.  I had this locally for testing and it slipped in.

.travis.yml
posborne: Deployment is done later on and should be unchanged.

.travis.yml
```diff
@@ -1,37 +1,57 @@
----
-sudo: false
 language: rust
+sudo: required
```
kamalmarhubi: I really wish they'd let this be set per matrix entry... we're going to be slower to schedule because of this.

/me is planning to spend time on https://github.com/kamalmarhubi/containy-thing this week because of this silliness

ci/cargo-config
```diff
@@ -0,0 +1,31 @@
...
+linker = "gcc"
+
+[target.x86_64-unknown-linux-musl]
+linker = "gcc"
+
+[target.i686-unknown-linux-gnu]
+linker = "gcc"
+
+[target.i686-unknown-linux-musl]
+linker = "gcc"
```
kamalmarhubi: is this the default? if so we should maybe leave it out in case Rust ever changes to lld?

ci/README.md
```diff
@@ -0,0 +1,42 @@
...
+determined by looking at the contents of the script.  The docker image
+used is [posborne/rust-cross][posborne/rust-cross].
+
+Running Test for Specific Architectures/Versions
+------------------------------------------------
+
+Suppose we have a failing test with Rust 1.6/1.7 on the raspberry pi.  In
+that case, we can run the following:
+
+    $ RUST_VERSIONS="1.6.0 1.7.0" RUST_TARGETS="arm-unknown-linux-gnueabihf" ci/run-all.sh
```
kamalmarhubi: does this actually work? I don't see either variable referred to later

ci/run-travis.sh
```diff
@@ -0,0 +1,44 @@
...
+# - TARGET: e.g. arm-unknown-linux-gnueabihf
+
+set -ex
+
+BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+if [ "$TRAVIS_OS_NAME" = "linux" ]; then
+  OS=unknown-linux-gnu
+else
+  OS=apple-darwin
```
kamalmarhubi: can you maek this an `elif`, or the whole thing a `case` and abort if it's neither linux nor darwin?

ci/run.sh
```diff
@@ -0,0 +1,124 @@
...
+BUILD_DIR="."
+
+VERSION="$1"
+TARGET="$2"
+
+export RUST_TEST_THREADS=1
+
+echo "======================================================="
+echo "TESTING VERSION: ${VERSION}, TARGET: ${TARGET}"
+echo "======================================================="
```
kamalmarhubi: minor: can you move this below function definitions?

ci/run.sh
kamalmarhubi: if you get jq installed, then this works:

```
$ for bin in $(cargo read-manifest | jq '.targets[].name'); do echo $bin; done
"rustfmt"
"cargo-fmt"
"rustfmt"
"system"
```

kamalmarhubi: Ahhh I didn't actually read the linked bug. That is silly an annoying.

.travis.yml
posborne: Agreed.  It is annoying -- I removed some combos from the matrix to make overall build time somewhat reasonable, but that is a tradeoff and we could, for instance, miss a failure on some intermediate rust version.

ci/README.md
posborne: No, I'll need to update this doc.  There is an equivalent, but it would be something like this:

```
DOCKER_IMAGE=posborne/rust-cross:arm RUST_VERSION="1.7.0" RUST_TARGET="arm-unknown-linux-gnueabihf" ci/run-docker.sh
```

Right now the containers only container 1.7.0.  Adding other versions is not hard but it also starts to blow up the image size.  I figured stable was the most valuable version to test against if you are going to test against anything.

ci/run-travis.sh
posborne: Sure.  This bit was taken from libc, but I agree that it is probably better to be a bit more strict: https://github.com/rust-lang/libc/blob/master/ci/run-travis.sh#L11

ci/run.sh
kamalmarhubi: It may still be possible to collect them up by doing `$bin{-,}*` or something like that

.travis.yml
kamalmarhubi: It was more a comment than anything else. It's been on my mind.

kamalmarhubi:
awesome!!!

@homu r+

PR #301, opened 2016-03-07
Add support for MIPS targets

### LINKED ISSUE #300: Add MIPS support
I've tried cross-compiling this create for a MIPS target (mipsel to be precise, a Creator Ci20 board) and realized that MIPS support is missing. It should be relatively simple to add it though, I'll try preparing a PR with it.

gabrielesvelto:
Closes #300

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -87,7 +87,7 @@ mod cpuset_attribs {
     }
 }
 
-#[cfg(all(target_arch = "arm", any(target_os = "linux", target_os = "android")))]
+#[cfg(all(any(target_arch = "arm", target_arch = "mips"), target_os = "android"))]
 mod cpuset_attribs {
     use super::CpuMask;
     // bionic only supports up to 32 independent CPUs, instead of 1024.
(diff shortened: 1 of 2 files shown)


gabrielesvelto:
I've slightly reorganized the code dealing with CPU sets so that ARM and MIPS have a CPU_SETSIZE limited to 32 only under Android (i.e. when using bionic) but use the full Linux CPU_SETSIZE otherwise. To work correctly this also needs a fix to the libc crate which I've pushed here: https://github.com/rust-lang/libc/pull/217

posborne:
Awesome, MIPS/MIPSEL are targets that will be tested (at least for compilation) when #297 (or whatever follows and integrates with travis) lands.

I'll hold off on approving until the changes in libc are published.  If we can get the CI changes in places and merged in soonish @humu can check that everything works.

gabrielesvelto:
The change to libc has landed and it's available in version 0.2.8 of the crate: https://github.com/rust-lang/libc/commit/2278a549559c38872b4338cb002ecc2a80d860dc

gabrielesvelto:
The dependency in libc has landed, is there anything else needed before this can be merged?

kamalmarhubi:
@homu r+

@posborne I figure we can fix any issues that come up in CI when we turn it on.

PR #302, opened 2016-03-08
tests: Disable test_sigwait on apple platforms

kamalmarhubi:
After #292 was merged, this flakiness remained. I observe it only on
Darwin, hence the targetted disabling until there's been more
investigation.


kamalmarhubi:
r? @posborne 
cc @dhylands

posborne:
@homu r+

Issue #264, opened 2016-02-12
Use definitions from libc instead of writing our own


kamalmarhubi:

kamalmarhubi:
I'm going to start going through module by module over the next week or so.

kamalmarhubi:
Thinking about this a bit more, we may want our own analog of the `sys` module pattern. Right now it's hard to track down where items come in because the `cfg` attrs are all over the place. @arcnmx you have a recently opened RFC on a related topic. Thoughts?

kamalmarhubi:
Ah, discussion of conditional compilation module pattern would belong in #98.

kamalmarhubi:
**Mentored bug**

This is something the maintainers will be working on on and off, but it will take some time. It's a good place to start, and I'm happy to help anyone who wants to contribute patches towards it.

The intention is to remove all `extern` declarations from nix, and rely on them from the [libc crate](https://github.com/rust-lang/libc) instead.

This can be done a module at a time. In some cases, functions declared in nix are not yet in libc, so this may require some patches to libc. I've made a few PRs there, and have a good idea of where things belong, so I can help with those PRs as well.

**Example PR** doing this for the `unistd` module: https://github.com/nix-rust/nix/pull/294

kamalmarhubi:
The basic steps for a PR would be:

1. remove the ffi submodule
2. replace references to `ffi::foo` by `libc::foo`
  - if `libc::foo` is missing, create a PR on the libc crate to add it
3. make sure to run the tests (`cargo test`)
4. open the PR!

Susurrus:
Supplanted by #316 and #748.

PR #304, opened 2016-03-10
uio: Use bindings from libc instead of our own

kamalmarhubi:
Refs https://github.com/nix-rust/nix/issues/264


posborne:
Looks good and CI passes.  @homu r+

PR #309, opened 2016-03-10
Update signalfd

### LINKED ISSUE #307: Build fails with signalfd feature enabled
Unresolved imports in `src/sys/signalfd.rs`:
- `sys::signal::signal::siginfo`
- `sys::signal::sigset_t`

alex-gulyas:
Update `signalfd` to use types and functions from `libc`.

It is a breaking change, because:
- renamed pub const `CREATE_NEW_FD` to `SIGNALFD_NEW`
- renamed pub const `SIGINFO_SIZE` to `SIGNALFD_SIGINFO_SIZE`
- removed pub const `SIGINFO_PADDING`

Fixes #307

### DIFF
--- a/src/sys/signalfd.rs
+++ b/src/sys/signalfd.rs
@@ -15,32 +15,25 @@
 //!
 //! Please note that signal discarding is not specific to `signalfd`, but also happens with regular
 //! signal handlers.
-use libc::{c_int, pid_t, uid_t};
+use libc;
 use unistd;
-use {Errno, Result};
-use sys::signal::signal::siginfo as signal_siginfo;
+use {Error, Errno, Result};
 pub use sys::signal::{self, SigSet};
+pub use libc::signalfd_siginfo as siginfo;
 
 use std::os::unix::io::{RawFd, AsRawFd};
 use std::mem;
 
-mod ffi {
-    use libc::c_int;
-    use sys::signal::sigset_t;
-
-    extern {
-        pub fn signalfd(fd: c_int, mask: *const sigset_t, flags: c_int) -> c_int;
-    }
-}
 
 bitflags!{
-    flags SfdFlags: c_int {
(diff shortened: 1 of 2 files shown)


src/sys/signalfd.rs
```diff
@@ -15,32 +15,25 @@
...
-    extern {
-        pub fn signalfd(fd: c_int, mask: *const sigset_t, flags: c_int) -> c_int;
-    }
-}
 
 bitflags!{
-    flags SfdFlags: c_int {
-        const SFD_NONBLOCK  = 0o00004000, // O_NONBLOCK
-        const SFD_CLOEXEC   = 0o02000000, // O_CLOEXEC
+    flags SfdFlags: libc::c_int {
```
posborne: In most code we bring `c_int` and other common types into the module namespace rather than accessing through `libc` -- not sure if this is worth promoting as a convention or if it is an important style element (obviously, it makes no functional difference).  CC @kamalmarhubi

posborne:
LGTM, but I'll let another reviewer in.  r? @kamalmarhubi

src/sys/signalfd.rs
fiveop: We do not have a convention on this yet. I would prefer module level imports, because they are shorter in general. This is a purely aesthetic argument, and thus not very strong. But I have no other arguments for or against either option.

This does, however, not have any relevance for this PR.

fiveop:
@homu r+

src/sys/signalfd.rs
kamalmarhubi: I've just been following the convention in the file. I don't have strong feelings in either direction. For common types, I agree module level is shorter, and for nix we're pretty clear on where they would be coming from.

kamalmarhubi:
Thanks @alex-gulyas for noticing and fixing this!

PR #310, opened 2016-03-11
Use sigwait function from libc instead of own ffi function.

fiveop:
Some left over ffi function that can be replaced now that we have libc 2.8.


utkarshkukreti:
@homu r+

Issue #214, opened 2015-12-14
Export all symbols at crate root as in libc


kamalmarhubi:
As part of implementing [RFC 1291](https://github.com/rust-lang/rfcs/blob/master/text/1291-promote-libc.md), the structure of the libc crate was changed. It now exports all symbols at the top level. I think it's worth considering the same structure for nix.

Susurrus:
I don't think this makes sense, primarily from a documentation point of view. It's nice being able to provide docs for entire modules for some of the wrapper types that `nix` implements. It also makes it easier for people from from C to find the man pages. I think it made sense for `libc` to do this given that they don't care about documentation as they're such a low-level dependency, but for `nix` we want to have a really good documentation story that includes being able to reuse all the existing docs on libc stuff.

I'm closing this as a WONTFIX because of the above.

PR #311, opened 2016-03-11
Add context module.

fiveop:
The module wraps context handling related functions and structs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,6 +49,13 @@ pub mod net;
 pub mod sched;
 
 pub mod sys;
+
+// This can be implemented for other platforms as soon as libc
+// provides bindings for them.
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86", target_arch = "x86_64")))]
+pub mod ucontext;
+
 pub mod unistd;
 
 /*
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -1,4 +1,3 @@
-
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub mod epoll;
 
--- a/src/ucontext.rs
+++ b/src/ucontext.rs
@@ -1,0 +1,25 @@
+use libc;
+use {Errno, Result};
+use std::mem;
+
+#[derive(Clone, Copy)]
+pub struct UContext {
+    context: libc::ucontext_t,
+}
+
+impl UContext {
+    pub fn get() -> Result<UContext> {
+        let mut context: libc::ucontext_t = unsafe { mem::uninitialized() };
+        let res = unsafe {
+            libc::getcontext(&mut context as *mut libc::ucontext_t)
+        };
+        Errno::result(res).map(|_| UContext { context: context })
+    }
+
+    pub fn set(&self) -> Result<()> {
+        let res = unsafe {
+            libc::setcontext(&self.context as *const libc::ucontext_t)
+        };
+        Errno::result(res).map(drop)
+    }
+}


src/sys/context.rs
```diff
@@ -0,0 +1,25 @@
+use libc;
+use {Errno, Result};
+use std::mem;
+
+#[derive(Clone, Copy)]
+pub struct UContext {
+    context: libc::ucontext_t,
```
kamalmarhubi: for direct wrappers like this, I kind of prefer not naming the inner field, and just accessing it via `.0`. thoughts?

src/sys/context.rs
```diff
@@ -0,0 +1,25 @@
+use libc;
```
kamalmarhubi: convention thusfar is using the header name, in which case this should be at `src/ucontext.rs`. There is #214 to discuss how to handle it though...

src/sys/context.rs
```diff
@@ -0,0 +1,25 @@
...
+    context: libc::ucontext_t,
+}
+
+impl UContext {
+    pub fn get() -> Result<UContext> {
+        let mut context: libc::ucontext_t = unsafe { mem::uninitialized() };
+        let res = unsafe {
+            libc::getcontext(&mut context as *mut libc::ucontext_t)
+        };
+        Errno::result(res).map(|_| UContext { context: context })
```
kamalmarhubi: if we use tuple struct newtypes, this can be `.map(UContext)`

src/sys/mod.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86", target_arch = "x86_64")))]
```
kamalmarhubi: can you add a comment that the gating is because libc doesn't include it for other platform / architectures? the functions themselves are part of POSIX: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/ucontext.h.html

kamalmarhubi: Hm I saw [this comment](https://github.com/rust-lang/libc/pull/196#issuecomment-189918276) on the libc PR where added the struct. Is it not possible to add more platforms?

src/sys/context.rs
fiveop: I would not mind. Feel free to add a corresponding section to the CONVENTIONS, as you have already talked about in the other issue. We should also open a issue that tracks the refactoring process.

Unless we have public members in our structs anywhere, it will not even be a breaking change.

I will leave this unchanged until we have that convention merged into master.

src/sys/context.rs
fiveop: I did not notice that, but should have thought about it. Thank you. This might be a good addition to our conventions file.

src/sys/context.rs
fiveop: No, it cannot. Because `res != context`.

src/sys/mod.rs
fiveop: They could be added. However, testing a struct if you do not have a platform involves a full CI run, so it is a bit tedious. I will add a comment for now.

src/lib.rs
```diff
@@ -30,6 +30,12 @@ pub mod libc {
 pub use libc::{c_int, c_void};
 pub use errno::Errno;
 
+// This can be implemented for other platforms as soon as libc
+// provides bindings for them.
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86", target_arch = "x86_64")))]
+pub mod context;
```
kamalmarhubi: this should be `ucontext`?

src/sys/context.rs
kamalmarhubi: sounds good!

src/sys/context.rs
kamalmarhubi: good point :-)

kamalmarhubi:
once you fix the `pub mod`, r=me

fiveop:
@homu r=@kamalmarhubi

PR #314, opened 2016-03-13
Rename flags to conform to conventions.

### LINKED ISSUE #290: SaFlag and SigFlag types should be called SaFlags and SigFlags
However, that would be a breaking change again. Are we doing 0.5.1 or 0.6.0 next? How frequently can we release?

fiveop:
Resolves #290.


kamalmarhubi:
@homu r+

PR #315, opened 2016-03-15
Add libc_bitflags convenience macro

kamalmarhubi:
We define many bitflags types with values from the libc crate. Currently
these look like this:

    bitflags!{
        flags ProtFlags: libc::c_int {
            const PROT_NONE      = libc::PROT_NONE,
            const PROT_READ      = libc::PROT_READ,
            const PROT_WRITE     = libc::PROT_WRITE,
            const PROT_EXEC      = libc::PROT_EXEC,
            const PROT_GROWSDOWN = libc::PROT_GROWSDOWN,
            const PROT_GROWSUP   = libc::PROT_GROWSUP,
        }
    }

There's some repetition which is tedious. With the new macro, the above
can instead be written

    libc_bitflags!{
        flags ProtFlags: libc::c_int {
            PROT_NONE,
            PROT_READ,
            PROT_WRITE,
            PROT_EXEC,
            PROT_GROWSDOWN,
            PROT_GROWSUP,
        }
    }

Thanks to Daniel Keep for the Little Book of Rust Macros, and for
helping with this macro.

Refs https://github.com/nix-rust/nix/issues/264

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,0 +1,188 @@
+/// The `libc_bitflags!` macro helps with a common use case of defining bitflags with values from
+/// the libc crate. It is used the same way as the `bitflags!` macro, except that only the name of
+/// the flag value has to be given.
+///
+/// The `libc` crate must be in scope with the name `libc`.
+///
+/// # Example
+/// ```
+/// libc_bitflags!{
+///     flags ProtFlags: libc::c_int {
+///         PROT_NONE,
+///         PROT_READ,
+///         PROT_WRITE,
+///         PROT_EXEC,
+///         #[cfg(any(target_os = "linux", target_os = "android"))]
+///         PROT_GROWSDOWN,
+///         #[cfg(any(target_os = "linux", target_os = "android"))]
+///         PROT_GROWSUP,
+///     }
+/// }
+/// ```
+macro_rules! libc_bitflags {
+    // (non-pub) Exit rule.
+    (@call_bitflags
+        {
+            name: $BitFlags:ident,
+            type: $T:ty,
+            attrs: [$($attrs:tt)*],
+            flags: [$($flags:tt)*],
+        }
+    ) => {
+        bitflags! {
+            $($attrs)*
+            flags $BitFlags: $T {
+                $($flags)*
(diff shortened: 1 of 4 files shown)


src/macros.rs
```diff
@@ -0,0 +1,194 @@
+/// The `libc_bitflags!` macro helps with a common use case of defining bitflags with values from
+/// the libc crate. It is used the same way as the `bitflags!` macro, except that only the name of
+/// the flag value has to be given.
+///
+/// The `libc` crate must be in scope with the name `libc`.
+///
+/// # Example
+/// ```
+/// # #[macro_use] extern crate nix;
```
fiveop: I would not make this publicly available.

fiveop:
I like this change.

Can you change all occurences instead of just one, right away. It should not be so many.

src/macros.rs
kamalmarhubi: ah this was me misunderstanding how macro visibility worked. good catch!

kamalmarhubi:
Going through the rest now.

posborne:
Got my approval as well, although I have not gone through the macros with a fine-tooth comb (but agreed with @fiveop on mot making the macro public).  I would be fine with merging this even if we don't have 100% of libc flags converted.

Once you are happy with things, r=me is fine.

kamalmarhubi:
I changed visibility of the macro, and added the use of the macro to the conventions doc.

> I would be fine with merging this even if we don't have 100% of libc flags converted.

Having spent most of the morning today trying to convert ~60 uses with medium success, I'd much rather merge this and do the rest incrementally. The conversion should happen but isn't high priority.

Leaving this to one of you to r+ taking into account the conventions change.

posborne:
Looks good to me.  @homu r+

kamalmarhubi:
@homu retry 7038da0

kamalmarhubi:
@homu retry

kamalmarhubi:
One day I'll get the hang of @homu r+ 7038da0

kamalmarhubi:
Oh I guess that should have been r=@posborne, ah well

PR #318, opened 2016-03-16
testing: remove refs to test_nix in run-all.sh

posborne:
Somewhere in the last set of changes, a few references to
`test_nix` were not removed.  Since that function doesn't exist,
the tests would never be run on those targets.

This script is only used when testing locally right now.


kamalmarhubi:
@homu r+

PR #322, opened 2016-03-17
travis: only require sudo on docker/i686 builds

posborne:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,11 @@ services:
 
 rust:
   - 1.1.0  # Oldest supported version
+  - 1.2.0
+  - 1.3.0
+  - 1.4.0
+  - 1.5.0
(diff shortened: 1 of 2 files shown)


posborne:
Based on http://stackoverflow.com/questions/34668468/use-sudo-true-with-single-element-of-build-matrix-in-travis-ci

posborne:
With the added coverage, this certainly doesn't make things faster (especially at peak hours), but it is faster than it would be for the same amount of coverage using sudo for everything.

r? @kamalmarhubi

posborne:
Ping @kamalmarhubi

kamalmarhubi:
Ooh! I didn't know you could specify `sudo` per-matrix entry. Nice!

@homu r+ a6967d6

PR #324, opened 2016-03-19
Allow to specify interface index for ipv6_mreq.

### LINKED ISSUE #323: ipv6_mreq does not allow interface to be set
The ability to set an interface index value should be exposed in https://github.com/nix-rust/nix/blob/5e3c954596f12512ba8728c437345a529ec48b0a/src/sys/socket/multicast.rs#L34

Related to https://github.com/carllerche/mio/issues/377

fiveop:
Fixes #323

### DIFF
--- a/src/sys/socket/multicast.rs
+++ b/src/sys/socket/multicast.rs
@@ -20,7 +20,7 @@ impl ip_mreq {
     pub fn new(group: Ipv4Addr, interface: Option<Ipv4Addr>) -> ip_mreq {
         ip_mreq {
             imr_multiaddr: group.0,
-            imr_interface: interface.unwrap_or(Ipv4Addr::any()).0
+            imr_interface: interface.unwrap_or(Ipv4Addr::any()).0,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/sys/socket/multicast.rs
```diff
@@ -37,4 +37,10 @@ impl ipv6_mreq {
             ipv6mr_interface: 0,
         }
     }
+    pub fn new(group: Ipv6Addr, interface: c_uint) -> ipv6_mreq {
```
posborne: This (function overloading based on number of arguments) isn't supported in rust AFAIK -- you can do it based on type by using traits but I don't think that works in this case.  We'll either have to break backwards compatibility or introduce a new function name.  I think the  latter probably makes sense.

kamalmarhubi: Yeah, overloading by argument count doesn't work, as evident from CI failure:

```
src/sys/socket/multicast.rs:40:5: 45:6 error: duplicate associated function [E0201]
```

Given that these aren't actually system functions, I think we'd be best off if *both* had more representative names. I also think there's work to be done here with moving to newtypes around the libc types..

My preference is to defer fixing #323 until we can get libc types + newtypes sorted out here.

fiveop: I figured that the sorting out would take some time, and wanted to provide a quick solution. Given that we will probably change the interface in a breaking way anyway in the future, I did not see a problem with that.

fiveop:
I will close this. We can fix it, when we know how we want to fix it.

PR #326, opened 2016-03-20
Fix arm and x86 ci

posborne:
See the individual commit message for more detail.  ARM is now passing (but we don't run mount tests any more when in the docker container).  For x86_64/i686, there were some changes in the posborne/rust-cross stuff.

### DIFF
--- a/ci/run-all.sh
+++ b/ci/run-all.sh
@@ -11,7 +11,8 @@ RUN_DOCKER="${BASE_DIR}/ci/run-docker.sh"
 
 export RUST_VERSION=1.7.0
 
-export DOCKER_IMAGE=posborne/rust-cross:base
+export DOCKER_IMAGE=posborne/rust-cross:x86
+RUST_TARGET=i686-unknown-linux-gnu ${RUN_DOCKER}
 RUST_TARGET=x86_64-unknown-linux-gnu ${RUN_DOCKER}
 RUST_TARGET=x86_64-unknown-linux-musl ${RUN_DOCKER}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,8 +70,6 @@ matrix:
       sudo: true
   allow_failures:
     - rust: nightly
-    - env: TARGET=aarch64-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:arm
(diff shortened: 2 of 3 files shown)


ci/run-all.sh
```diff
@@ -11,7 +11,8 @@ RUN_DOCKER="${BASE_DIR}/ci/run-docker.sh"
 
 export RUST_VERSION=1.7.0
 
-export DOCKER_IMAGE=posborne/rust-cross:base
+export DOCKER_IMAGE=posborne/rust-cross:x86
+RUST_TARGET=i686-unknown-linux-gnu ${RUN_DOCKER}
```
posborne: This didn't work previously but does not because the `x86` image includes `gcc-multilib` now so we can run i686 binaries.

test/test_mount.rs
```diff
@@ -207,15 +207,22 @@ macro_rules! run_tests {
...
-                     test_mount_bind};
-    setup_namespaces();
-
-    run_tests!(test_mount_tmpfs_without_flags_allows_rwx,
-               test_mount_rdonly_disallows_write,
-               test_mount_noexec_disallows_exec,
-               test_mount_bind);
+    use std::env;
+
+    if env::var("DOCKER_ENVIRONMENT").is_ok() {
```
kamalmarhubi: nit: explicitly check if it's 1

kamalmarhubi: actually probably better is to change the panic in `setup_namespaces` to just exit after printing a warning. not having unprivileged namespaces available is not a test failure. thoughts?

kamalmarhubi:
r=me after whichever fix you think makes more sense

test/test_mount.rs
fiveop: On my local machine the test always fails because of that.

kamalmarhubi: Ok I just sent #327 to allow it to fail if the reason is `unshare` not supported.

@posborne you should be able to rebase on top of that once it's accepted, and drop changes in this file.

kamalmarhubi: I'm not really satisfied with this test runner anyway...

kamalmarhubi:
Update after #327 merge: r=me after dropping changes to test_mount.rs

posborne:
Changes to mount tests have been dropped in lieu of #327.  @homu r=kamalmarhubi (pending tests pass)

PR #327, opened 2016-03-21
mount: Do not treat unshare(2) failure as test failure

kamalmarhubi:
The mount test runner uses unprivileged user namespaces. Previously,
failure from `unshare(2)` to create the user namespace would fail the
test. This changes that to simply print an error and exit successfully.

Refs https://github.com/nix-rust/nix/pull/326


fiveop:
@homu r+ (assuming tests run through)

PR #330, opened 2016-03-28
epoll: Add EPOLLEXCLUSIVE flag.

kubo39:
EPOLLEXCLUSIVE flag is available in Linux4.5.

### DIFF
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -27,6 +27,7 @@ bitflags!(
         const EPOLLERR = 0x008,
         const EPOLLHUP = 0x010,
         const EPOLLRDHUP = 0x2000,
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Thanks for the PR! Any chance I can also get you to open a PR to add it to the libc crate? The place to add `EPOLL*` flags is likely here: https://github.com/rust-lang/libc/blob/cd925e0ce1958b3e8561932cbe1fad5b328a8b95/src/unix/notbsd/mod.rs#L424-L434

In general, we're [moving to using definitions from the libc crate](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#libc-constants-functions-and-structs), so adding it there will make it easier to clean up this module when we shift to using libc definitions. :-)

kamalmarhubi:
@homu r+

PR #332, opened 2016-03-29
unistd: Redesign the enum returned by fork()

kamalmarhubi:
This changes the name of the enum returned by `fork()` to `ForkResult`,
and changes the `Parent` variant to be struct-like.

The result can be matched like

    use nix::unistd::ForkResult::*;
    match fork().unwrap() {
        Parent { child } => { ... }
        Child => { ... }
    }

using the shorthand matching syntax for struct-like enum variants.

This is a breaking change.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -12,33 +12,33 @@ use std::os::unix::io::RawFd;
 pub use self::linux::*;
 
 #[derive(Clone, Copy)]
-pub enum Fork {
-    Parent(pid_t),
+pub enum ForkResult {
+    Parent {
+        child: pid_t
+    },
     Child
 }
 
-impl Fork {
+impl ForkResult {
     pub fn is_child(&self) -> bool {
         match *self {
-            Fork::Child => true,
+            ForkResult::Child => true,
             _ => false
         }
     }
 
     pub fn is_parent(&self) -> bool {
-        match *self {
-            Fork::Parent(_) => true,
-            _ => false
-        }
+        !self.is_child()
     }
 }
 
-pub fn fork() -> Result<Fork> {
+pub fn fork() -> Result<ForkResult> {
+    use self::ForkResult::*;
     let res = unsafe { libc::fork() };
 
     Errno::result(res).map(|res| match res {
-        0 => Fork::Child,
-        res => Fork::Parent(res)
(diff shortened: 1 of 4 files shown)


kamalmarhubi:
**Review note**: I'd like more than one person to comment on this, and to figure out if this should be made a convention.

I can't immediately think of anywhere eles in nix that we do something similar to `fork`. Anyone have any handy?

src/unistd.rs
```diff
@@ -27,7 +29,7 @@ impl Fork {
 
     pub fn is_parent(&self) -> bool {
         match *self {
-            Fork::Parent(_) => true,
+            Fork::Parent {..}  => true,
             _ => false
         }
     }
```
fiveop: pub fn is_parent(&self) -> bool {
        !is_child(self)
    }

while you're at it? :)

fiveop:
I could not guess the intent of the struct before looking how it was used (as `fork()` result). Why don't we call it ForkResult? It does not represent a fork, it represents the result of a fork.

kamalmarhubi:
I'll do that once I get back to the computer.

kamalmarhubi:
@fiveop done!

posborne:
Looks good to me.  I can't think of other similar calls off hand, but it wouldn't surprise me if one exists.  I don't think we support it at present, but `forkpty` would certainly look the same (as it calls `fork()`).

kamalmarhubi:
Not yet, but soon: https://github.com/rust-lang/libc/pull/246

fiveop:
@homu r+

What lead you to mark our fork wrapper as inline? I do not disagree. I think it makes sence for our wrapper functions that only add error wrapping (I don't want to call it handling). Maybe this is another thing for our conventions?

kamalmarhubi:
> What lead you to mark our fork wrapper as inline? I do not disagree. I think it makes sence for our wrapper functions that only add error wrapping (I don't want to call it handling).

Vaguely, I think I'd been looking at something in std and saw some things that were slightly more complex than wrappers marked as `#[inline]`. The `#[inline]` attribute is only a suggestion, anyway.

> Maybe this is another thing for our conventions?

This overall is an interesting point. Opened #338 to discuss.

PR #333, opened 2016-03-29
tests: Unwrap result of fork in tests

kamalmarhubi:
All tests were panicking in the `Err(..)` case. This makes the code a
little cleaner by moving the panic into a call to `expect()`, tidying up
the match arms.


kamalmarhubi:
Uh I'll reopen this; missed some tests.

PR #336, opened 2016-03-30
unistd: Add pause(2)

kamalmarhubi:


kamalmarhubi:
This one is weird because if it ever returns, the return value will be `Err(EINTR)`.

kamalmarhubi:
> RETURN VALUE

>    Since pause() suspends thread execution indefinitely unless interrupted by a signal, there is no successful completion return value. A value of -1 is returned and errno is set to indicate the error. 

> ERRORS

>    The pause() function will fail if:

>    [EINTR]
>        A signal is caught by the calling process and control is returned from the signal-catching function.

―http://pubs.opengroup.org/onlinepubs/007908799/xsh/pause.html

fiveop:
@homu r+

PR #337, opened 2016-03-31
Fixed warning with raw_pointer_derive

Razican:
There is no need to allow raw_pointer_derive, and it gives a warning in Rust stable. This fixes that warning.


fiveop:
The CI result shows you why we cannot just remove it. We currently support older versions of rust, where the error was still a thing.

kamalmarhubi:
I had a plan for this, but I forgot what it was!

PR #340, opened 2016-04-01
Fix raw_pointer_derive warning

kamalmarhubi:
This commit adds a small build script to detect if we need to
`#[allow(raw_pointer_derive)]` and makes the attribute conditional.

Refs #337

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ exclude     = [
   "test/**/*"
 ]
 
+build = "build.rs"
+
 [features]
 eventfd = []
 execvpe = []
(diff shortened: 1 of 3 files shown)


posborne:
Huh, I tried a couple versions locally and the tests pass.

@homu retry

posborne:
Ah, now I see.  It is only when we are building the docs that things choke.  It might work to invert the logic so that we allow unless we say that we should not have that behavior (might still get warnings when building docs, but who cares).

kamalmarhubi:
Ah thanks for looking. I tried at every released version locally and was confused.

kamalmarhubi:
Flipped the logic, and filed https://github.com/rust-lang/cargo/issues/2539

posborne:
@homu r+

PR #344, opened 2016-04-07
Allow to specify signal when calling clone.

### LINKED ISSUE #343: Expose termination signal option in clone
I'm talking about this commit:
https://github.com/nix-rust/nix/commit/53cae89c6892b1f125ef69a769ed8147fcba4d92

The issue is that according to a [man page](http://linux.die.net/man/2/clone) you need to mix a signal name to a flags in clone function, like this:
```
pid = clone(childFunc, stackTop, CLONE_NEWUTS | SIGCHLD, argv[1]);
```

But with bitflags you can't do it. E.g. `CloneFlags::from_bits(SIGCHLD)` returns `None`.

Any good idea how to fix it? Revert the commit?

fiveop:
This is my suggestion on how to fix #343.

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -1,5 +1,6 @@
 use std::mem;
 use std::os::unix::io::RawFd;
+use std::option::Option;
 use libc::{self, c_int, c_void, c_ulong, pid_t};
 use {Errno, Result};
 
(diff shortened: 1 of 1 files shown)


src/sched.rs
```diff
@@ -197,15 +198,16 @@ pub fn sched_setaffinity(pid: isize, cpuset: &CpuSet) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-pub fn clone(mut cb: CloneCb, stack: &mut [u8], flags: CloneFlags) -> Result<pid_t> {
+pub fn clone(mut cb: CloneCb, stack: &mut [u8], flags: CloneFlags, signal: Option<c_int>) -> Result<pid_t> {
```
posborne: Signal should be constrained to an 8-bit unsigned value (`u8`/`c_char`) at compile time given the allowable mask for signal: http://lxr.free-electrons.com/source/include/uapi/linux/sched.h#L7.  Other values could clobber the other flags.

posborne:
I don't have a strong opinion on this one.  Allowing for the signal to be included does seem important and I don't see any obvious problems with this approach.  @kamalmarhubi

src/sched.rs
fiveop: I would prefer to let signal remain of type `c_int`, the type for signals. In order to address your (valid) concern, we could add a check into our wrapper, that signals an appropriate error.

kamalmarhubi:
Interesting. I'd have to read the entire man page to figure out if there are any other similar issues so we don't end up changing the signature repeatedly.

kamalmarhubi:
I just opened a PR with an alternative approach at #348.

kamalmarhubi:
Given discussion on #348 and #343, I think this is the most acceptable approach. If / when we add an enum for signals, we can adjust the type of the argument.

@homu r+ b06432b

Issue #341, opened 2016-04-04
Remove or revamp sys::syscall module


kamalmarhubi:
This came up in [a comment from @pnkfelix on #339](https://github.com/nix-rust/nix/issues/339#issuecomment-205407123).

This module appears to be Linux specific. It declares

``` rust
extern {
    pub fn syscall(num: Syscall, ...) -> c_int;
}
```

for all architectures, with `Syscall` being architecture dependent alias that is `c_long` on all listed architectures.

On Linux, [`syscall(2)`][syscall] should return a `long`, so this declaration is incorrect. We should switch to using definitions from `libc` and tidy up or remove this module.

Related: #264

kamalmarhubi:
This PR is related: https://github.com/rust-lang/libc/pull/374

Susurrus:
Closed by #747

PR #345, opened 2016-04-07
Use libc's syscall declaration.

fiveop:
Another proposal for fixing an issue ( #341 ). I decided against removing the module, since the constants defined in there are used and they are not defined in libc.

### DIFF
--- a/src/sys/syscall.rs
+++ b/src/sys/syscall.rs
@@ -1,14 +1,32 @@
 //! Indirect system call
-//!
-use libc::c_int;
+
+pub use self::os::*;
+
+#[cfg(any(target_os = "linux",
+          target_os = "android",
+          target_os = "emscripten"))]
+mod os {
+    use libc::c_long;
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
@fiveop 
> since the constants defined in there are used and they are not defined in libc.

Do you mean the constants are used inside nix, or by users of nix? I think if within nix, we should get rid of any such uses. For users of nix, I think the eventual destiny of this module is to be removed, so it's a matter of when they have to work around that. Or do you think the module should remain, even if all constants are added to libc?

kamalmarhubi:
A heroic PR on libc is in progress to bring in the syscall numbers: https://github.com/rust-lang/libc/pull/374

kamalmarhubi:
At least on Linux

fiveop:
I was about to close this PR, when you started your triaging spree. I think #341 is enough of a reminder, and once I or someone else feels like it, can start afresh. So let's close it.

PR #349, opened 2016-04-13
make execvpe available for all platforms

Mic92:
- execvpe is as GNU extension of libc.                                                       
- redone in Rust to make it available on all systems.                          
- this also makes the feature flag obsolete.

NOTE:

- this my first contribution to a rust project, so watch out for beginner's mistake :)
- is there a way to deprecate feature flags, so the code will still compile with a warning?

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -373,13 +373,37 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
     unsafe { libc::sleep(seconds) }
 }
 
+pub fn execvpe(filename: &CString, args: &[CString], env: &[CString]) -> Result<()> {
+    use std::env;
+    use std::ffi::OsString;
+    use std::ffi::OsStr;
+    use std::os::unix::ffi::OsStrExt;
+
+    if filename.as_bytes().iter().find(|c| **c == b'/').is_some() {
+        return execve(filename, args, env);
+    }
+
+    let paths = match env::var_os("PATH") {
+        Some(val) => val,
+        None => OsString::from("/usr/local/bin:/bin:/usr/bin"),
+    };
+
+    let name = OsStr::from_bytes(&filename.as_bytes());
+    let mut res = Err(Error::Sys(Errno::ENOENT));
+    for path in env::split_paths(&paths) {
+        let p = path.with_file_name(name);
+        let p2 = &CString::new(p.as_os_str().as_bytes()).unwrap();
+        res = execve(p2, args, env);
+    }
+
+    return res;
+}
+
 #[cfg(any(target_os = "linux", target_os = "android"))]
 mod linux {
     use sys::syscall::{syscall, SYSPIVOTROOT};
     use {Errno, Result, NixPath};
 
-    #[cfg(feature = "execvpe")]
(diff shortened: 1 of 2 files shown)


Mic92:
done

Mic92:
anything else left?

kamalmarhubi:
Hi @Mic92 sorry for the delay. Taking a quick look!

kamalmarhubi:
> - this my first contribution to a rust project, so watch out for beginner's mistake :)

Yay! And thanks for picking our project!

> - is there a way to deprecate feature flags, so the code will still compile with a warning?

I'm not sure. I haven't messed with Cargo features that much!

src/unistd.rs
```diff
@@ -373,13 +373,37 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
...
+    use std::ffi::OsStr;
+    use std::os::unix::ffi::OsStrExt;
+
+    if filename.as_bytes().iter().find(|c| **c == b'/').is_some() {
+        return execve(filename, args, env);
+    }
+
+    let paths = match env::var_os("PATH") {
+        Some(val) => val,
+        None => OsString::from("/usr/local/bin:/bin:/usr/bin"),
```
kamalmarhubi: Is this default specified somewhere?

kamalmarhubi:
Ok having taken a look: I've always been unsure of this function in nix, and have considered proposing its removal. In particular, it doesn't match the string passing style of other parts of the library. On the other hand, it is somewhat convenient.

I'd like feedback from a couple of other maintainers on how they feel about it. @fiveop @posborne?

kamalmarhubi:
This function also does more work than most nix functions. It's not wrapping an existing library function but actually reimplementing it.

src/unistd.rs
Mic92: I took the same value as musl libc.
If you want to stick to the POSIX standard, it is only "/bin/:usr/bin" according to confstr(3).

Mic92:
The reason I have not using an existing one is, because it is not available in all libc implementations. My first version was using the libc one, but it is actually more code
and requires users to set a feature flag, which will not work on any platform.

arcnmx:
> This function also does more work than most nix functions. It's not wrapping an existing library function but actually reimplementing it.

Yeah, this makes me wary about including it as more than just a libc wrapper... Also that one important difference between this and the glibc version is that glibc will use the stack / `alloca` to avoid allocations when possible.

> I've always been unsure of this function in nix, and have considered proposing its removal. In particular, it doesn't match the string passing style of other parts of the library.

Note that this applies to the whole `exec` family of functions, which do need a bit of a revamp. `execvpe` isn't necessarily unique here, though it has bitrot in particular because the CI doesn't test that feature flag. I'd been holding off on fixing them while waiting for #230 to be resolved, but just threw up #358 now with some of my old work to gather feedback.

Mic92:
I was already wondering, why exec* wants CString. Stack allocated arrays is something what I do not want to implement. Well, than I use my own implementation of execvpe.

Artoria2e5:
It might make sense to move this code out as a separate crate (in the new signature of course)

Issue #316, opened 2016-03-16
Add some constants to libc


kamalmarhubi:
**NB** This issue describes making changes to the [libc](https://github.com/rust-lang/libc) crate, not to nix. The issue is here because these are constants we'd like to use from libc as part of work on #264, but are not yet available in the libc crate.

I've been going through bitflags as part of #315, and taking note of some places where we need to get constants upstreamed.
- [x] `sys::memfd::MFD_*` (rust-lang/libc#836)
- [x] `sys::socket::consts::MSG_*`
- [x]  `sys:socket::consts::SOL_*` (https://github.com/rust-lang/libc/pull/228)

posborne:
I have a PR open for `sys:socket::consts::SOL_*`: https://github.com/rust-lang/libc/pull/228 although those are not flags.

kamalmarhubi:
Definitely add any non-flags up above too. I'm just looking at flags quite closely because of #315.

PR #352, opened 2016-04-14
Added sys::socket::consts::MSG_*

maennchen:
Partially solve #316.


fiveop:
Thank you for the PR. However, #316 is about getting the constants defined in rust-lang/libc and then use those constants for our bitflags. I would prefer adding the constants to libc first, because they have testing infrastructure in place that checks if the constants are defined with the same values in C headers.

kamalmarhubi:
Ah the description in #316 is a bit unclear. I'll update it.

kamalmarhubi:
Triaging PRs, and closing this one for inactivity.

Issue #346, opened 2016-04-08
Warning when compiling


szagi3891:
```
grzegorz@grzegorz:~/Pulpit/rust$ git clone https://github.com/nix-rust/nix
Cloning into 'nix'...
remote: Counting objects: 2904, done.
remote: Total 2904 (delta 0), reused 0 (delta 0), pack-reused 2904
Receiving objects: 100% (2904/2904), 751.82 KiB | 214.00 KiB/s, done.
Resolving deltas: 100% (2043/2043), done.
Checking connectivity... done.
grzegorz@grzegorz:~/Pulpit/rust/nix$ cargo build
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading bitflags v0.4.0
 Downloading semver v0.1.20
 Downloading rustc_version v0.1.7
   Compiling libc v0.2.9
   Compiling semver v0.1.20
   Compiling rustc_version v0.1.7
   Compiling bitflags v0.4.0
   Compiling cfg-if v0.1.0
   Compiling nix v0.5.1-pre (file:///home/grzegorz/Pulpit/rust/nix)
src/sched.rs:208:20: 208:34 warning: `extern "C" fn(*mut Box<std::ops::FnMut() -> isize>) -> i32 {sched::clone::callback}` is now zero-sized and has to be cast to a pointer before transmuting to `*const extern "C" fn(*const Box<std::ops::FnMut() -> isize>) -> i32`, #[warn(transmute_from_fn_item_types)] on by default
src/sched.rs:208         ffi::clone(mem::transmute(callback), ptr as *mut c_void, flags.bits(), &mut cb)
                                    ^~~~~~~~~~~~~~
src/sched.rs:208:20: 208:34 warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
src/sched.rs:208:20: 208:34 note: for more information, see issue #19925 <https://github.com/rust-lang/rust/issues/19925>
grzegorz@grzegorz:~/Pulpit/rust/nix$
```

can you fix it?

posborne:
@szagi3891 What version of rustc?

fiveop:
This is the relevant Rust issue: https://github.com/rust-lang/rust/issues/19925 .

szagi3891:
It is the latest version. Today installed.  Nightly  (1.9)

```
gszeliga@dev4:~/workspace/nix$ cargo version
cargo 0.10.0-nightly (f20ca62 2016-04-06)
```

[...]

PR #353, opened 2016-04-14
Cast function item to function pointer in order to appease compiler.

fiveop:
This is necessary because of compiler changes. For further information
look at rust-lang/rust#19925.

This solves #346.

I know that there are other problems with the affected function. We are thinking about how to best pass the arguments to the function and we don't want to define the ffi's ourselves. However, I would like to get the warning out of our tree as soon as possible. We do not need to wait for our polish.


kamalmarhubi:
> However, I would like to get the warning out of our tree as soon as possible. We do not need to wait for our polish.

:+1:

@homu r+ 47dc5fe

PR #354, opened 2016-04-15
fcntl: Use bindings from libc instead of our own

brianp:
**Un-finished** and looking for help. As #264 was tagged "mentor, good first bug"

Ref #264

I couldn't find `F_GET_SEALS`, or `F_ADD_SEALS` in libc.

Is this a case where I should be making a PR over there? I'm not sure where exactly to address it in libc.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1,109 +1,26 @@
 use {Errno, Result, NixPath};
-use libc::{c_int, c_uint};
+use libc::{self, c_int, c_uint};
 use sys::stat::Mode;
 use std::os::unix::io::RawFd;
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 use sys::uio::IoVec;  // For vmsplice
-#[cfg(any(target_os = "linux", target_os = "android"))]
-use libc;
 
 pub use self::consts::*;
-pub use self::ffi::flock;
 
+// TODO: The remainder of the ffi module should be removed afer work on
+// https://github.com/rust-lang/libc/issues/235 is resolved.
 #[allow(dead_code)]
 mod ffi {
-    pub use libc::{open, fcntl};
-    pub use self::os::*;
-    pub use libc::flock as libc_flock;
-    pub use libc::{LOCK_SH, LOCK_EX, LOCK_NB, LOCK_UN};
+    use libc::c_int;
 
-    #[cfg(any(target_os = "linux", target_os = "android"))]
-    mod os {
-        use libc::{c_int, c_short, off_t, pid_t};
-
-        #[repr(C)]
-        #[derive(Clone, Copy, Default, Debug)]
-        pub struct flock {
-            pub l_type: c_short,
-            pub l_whence: c_short,
-            pub l_start: off_t,
-            pub l_len: off_t,
-            pub l_pid: pid_t,
-
-            // not actually here, but brings in line with freebsd
-            pub l_sysid: c_int,
-        }
-
-        pub const F_DUPFD:         c_int = 0;
-        pub const F_DUPFD_CLOEXEC: c_int = 1030;
-        pub const F_GETFD:         c_int = 1;
-        pub const F_SETFD:         c_int = 2;
-        pub const F_GETFL:         c_int = 3;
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Hi @brianp, thanks for the PR! Taking a look now!

kamalmarhubi:
Oh re `F_GET_SEAL` and `F_SET_SEALS` in libc, this is blocked by needing to figure out https://github.com/rust-lang/libc/issues/235

src/fcntl.rs
```diff
@@ -9,101 +9,26 @@ use sys::uio::IoVec;  // For vmsplice
...
-        pub const F_DUPFD:         c_int = 0;
-        pub const F_DUPFD_CLOEXEC: c_int = 1030;
-        pub const F_GETFD:         c_int = 1;
-        pub const F_SETFD:         c_int = 2;
-        pub const F_GETFL:         c_int = 3;
-        pub const F_SETFL:         c_int = 4;
-        pub const F_SETLK:         c_int = 6;
-        pub const F_SETLKW:        c_int = 7;
-        pub const F_GETLK:         c_int = 5;
+        use libc::{c_int};
```
kamalmarhubi: no need for braces if just importing one name

src/fcntl.rs
```diff
@@ -9,101 +9,26 @@ use sys::uio::IoVec;  // For vmsplice
...
-        #[cfg(target_os = "netbsd")]
-        pub const F_CLOSEM:        c_int = 10;
-        #[cfg(target_os = "netbsd")]
-        pub const F_MAXFD:         c_int = 11;
-        #[cfg(target_os = "netbsd")]
-        pub const F_GETNOSIGPIPE:  c_int = 13;
-        #[cfg(target_os = "netbsd")]
-        pub const F_SETNOSIGPIPE:  c_int = 14;
-    }
+    mod os { }
```
kamalmarhubi: I can't tell, but it looks like this module can be deleted altogether

src/fcntl.rs
```diff
@@ -9,101 +9,26 @@ use sys::uio::IoVec;  // For vmsplice
 use libc;
```
kamalmarhubi: delete this import and the `#[cfg]` attribute, and change line 2 to `use libc::{self, c_int, c_uint}`. This is one reason for the Apple build failing: `libc` the module was only imported on Linux.

kamalmarhubi:
I think if you make the change with the `libc` import, the build should pass. I'm still not sure exactly what to do with `F_{GET,SET}_SEALS`.

src/fcntl.rs
brianp: Oh yeah this can come out. I left it because the flags designated the other OS block per system for targets `target_os = "linux", target_os = "android"` and then included everything in `os::*` so non linux and android systems couldn't find it. 

I'll clean that up.

brianp:
Thanks for the reply. If I can help out with rust-lang/libc#235 in any way let me know!

I'll tidy up this pr but will still leave `F_{GET,SET}_SEALS` until the above is addressed.

src/fcntl.rs
```diff
@@ -1,109 +1,24 @@
...
-        #[cfg(target_os = "netbsd")]
-        pub const F_CLOSEM:        c_int = 10;
-        #[cfg(target_os = "netbsd")]
-        pub const F_MAXFD:         c_int = 11;
-        #[cfg(target_os = "netbsd")]
-        pub const F_GETNOSIGPIPE:  c_int = 13;
-        #[cfg(target_os = "netbsd")]
-        pub const F_SETNOSIGPIPE:  c_int = 14;
-    }
+    pub const F_ADD_SEALS: c_int = 1033;
```
kamalmarhubi: add a TODO explaining why these are left here, and reference https://github.com/rust-lang/libc/issues/235

kamalmarhubi:
one small comment, otherwise this looks good. could you squash your commits down so we don't have the broken one in our history?

brianp:
Happy to! I'll take care of it now.

kamalmarhubi:
Thanks @brianp!

@homu r+ 7568340

PR #356, opened 2016-04-16
mount: Use bindings from libc instead of our own

brianp:
Ref #264

Changing internal usage of `ffi::mount` and `ffi:umount` to make use of libc.


posborne:
Thanks!  Looks good @homu r+

PR #357, opened 2016-04-18
Use Void in exec return type

arcnmx:
Indicates that these methods cannot return successfully. This does introduce a new dependency; an empty enum could be used instead but then you would lose the convenience `unreachable` methods exposed by the `void` crate, and standardizing on one type, etc...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,9 +23,10 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc     = "0.2.8"
+libc = "0.2.8"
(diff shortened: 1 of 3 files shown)


fiveop:
I approve, but a new dependency probably warrants a second opionion.

posborne:
Given its well-constrained purpose and small size, I am on board.  I fixed support for rust back to 1.0 for the 0.1.2 released (regressed when `no_std` support was added).  @humu r+

CC @reem

kamalmarhubi:
@homu r=@posborne

:-)

PR #359, opened 2016-04-19
Add some missing SFlags

arcnmx:
Also reordered just because I wasn't sure where to insert the new ones.

### DIFF
--- a/src/sys/stat.rs
+++ b/src/sys/stat.rs
@@ -18,11 +18,14 @@ mod ffi {
 
 bitflags!(
     flags SFlag: mode_t {
-        const S_IFREG  = 0o100000,
-        const S_IFCHR  = 0o020000,
-        const S_IFBLK  = 0o060000,
         const S_IFIFO  = 0o010000,
-        const S_IFSOCK = 0o140000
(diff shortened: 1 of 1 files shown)


utkarshkukreti:
The `libc` crate seems to have the values of these flags. Maybe you should just use those?

src/sys/stat.rs
```diff
@@ -18,30 +18,33 @@ mod ffi {
...
+        const S_IRGRP = libc::S_IRGRP,
+        const S_IWGRP = libc::S_IWGRP,
+        const S_IXGRP = libc::S_IXGRP,
+
+        const S_IRWXO = libc::S_IRWXO,
+        const S_IROTH = libc::S_IROTH,
+        const S_IWOTH = libc::S_IWOTH,
+        const S_IXOTH = libc::S_IXOTH,
 
         const S_ISUID = 0o4000,
```
utkarshkukreti: @arcnmx You missed these 3. r=me after you replace these :)

arcnmx:
Oop I didn't look closely, they were the wrong type! Let's see if this works o:

@homu r=utkarshkukreti

PR #361, opened 2016-04-25
Add missing signals.

fiveop:
In preparation of turning them into an enumeration.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -11,36 +11,47 @@ pub use libc::{
     SIGINT,
     SIGQUIT,
     SIGILL,
+    SIGTRAP,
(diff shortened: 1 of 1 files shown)


fiveop:
Added rust-lang/libc#265 to get this to work.

fiveop:
Now, all we need is a new libc version for this.

fiveop:
This should compile now. Will @homu run another test, if we accept it?

kamalmarhubi:
Could you bump the libc requirement in `Cargo.toml` to the release that includes these names. Once you do that, r=me.

Re @homu, we'll find out!

fiveop:
@homu r=@kamalmarhubi

PR #362, opened 2016-04-25
Signal enum

fiveop:
This is work in progress. I post this pull request as a request for discussion. I mark the lines I have doubts or ideas about.

Ignore the first commit. It is the same as PR #361, which I wanted to base this change off.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -2,11 +2,93 @@
 // See http://rust-lang.org/COPYRIGHT.
 
 use libc;
-use {Errno, Result};
+use {Errno, Error, Result};
 use std::mem;
 use std::ptr;
 
-pub use libc::{
+// Currently there is only one definition of c_int in libc, as well as only one
+// type for signal constants.
+// We would prefer to use the libc::c_int alias in the repr attribute. Unfortunately
+// this is not (yet) possible.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(i32)]
+pub enum Signal {
+    SIGHUP = libc::SIGHUP,
+    SIGINT = libc::SIGINT,
+    SIGQUIT = libc::SIGQUIT,
+    SIGILL = libc::SIGILL,
+    SIGTRAP = libc::SIGTRAP,
+    SIGABRT = libc::SIGABRT,
+    SIGBUS = libc::SIGBUS,
+    SIGFPE = libc::SIGFPE,
+    SIGKILL = libc::SIGKILL,
+    SIGUSR1 = libc::SIGUSR1,
+    SIGSEGV = libc::SIGSEGV,
+    SIGUSR2 = libc::SIGUSR2,
+    SIGPIPE = libc::SIGPIPE,
+    SIGALRM = libc::SIGALRM,
+    SIGTERM = libc::SIGTERM,
+    #[cfg(not(target_os = "macos"))]
+    SIGSTKFLT = libc::SIGSTKFLT,
+    SIGCHLD = libc::SIGCHLD,
+    SIGCONT = libc::SIGCONT,
+    SIGSTOP = libc::SIGSTOP,
+    SIGTSTP = libc::SIGTSTP,
+    SIGTTIN = libc::SIGTTIN,
+    SIGTTOU = libc::SIGTTOU,
+    SIGURG = libc::SIGURG,
+    SIGXCPU = libc::SIGXCPU,
+    SIGXFSZ = libc::SIGXFSZ,
+    SIGVTALRM = libc::SIGVTALRM,
+    SIGPROF = libc::SIGPROF,
+    SIGWINCH = libc::SIGWINCH,
+    SIGIO = libc::SIGIO,
+    #[cfg(not(target_os = "macos"))]
+    SIGPWR = libc::SIGPWR,
+    SIGSYS = libc::SIGSYS,
+    #[cfg(target_os = "macos")]
+    SIGEMT = libc::SIGEMT,
+    #[cfg(target_os = "macos")]
+    SIGINFO = libc::SIGINFO,
+}
+
+pub use self::Signal::*;
+
+#[cfg(not(target_os = "macos"))]
+const SIGNALS: [Signal; 31] = [
+    SIGHUP,
+    SIGINT,
+    SIGQUIT,
+    SIGILL,
+    SIGTRAP,
+    SIGABRT,
+    SIGBUS,
+    SIGFPE,
+    SIGKILL,
+    SIGUSR1,
+    SIGSEGV,
+    SIGUSR2,
+    SIGPIPE,
+    SIGALRM,
+    SIGTERM,
+    SIGSTKFLT,
+    SIGCHLD,
+    SIGCONT,
+    SIGSTOP,
+    SIGTSTP,
+    SIGTTIN,
+    SIGTTOU,
+    SIGURG,
+    SIGXCPU,
+    SIGXFSZ,
+    SIGVTALRM,
+    SIGPROF,
+    SIGWINCH,
+    SIGIO,
+    SIGPWR,
+    SIGSYS];
+#[cfg(target_os = "macos")]
+const SIGNALS: [Signal; 31] = [
     SIGHUP,
     SIGINT,
     SIGQUIT,
@@ -36,24 +118,49 @@ pub use libc::{
     SIGWINCH,
     SIGIO,
     SIGSYS,
-};
-
-#[cfg(target_os = "macos")]
-pub use libc::{
     SIGEMT,
-    SIGINFO,
-};
-
-#[cfg(not(target_os = "macos"))]
-pub use libc::{
-    SIGPWR,
-    SIGSTKFLT,
-    SIGIOT, // Alias for SIGABRT
-    SIGPOLL, // Alias for SIGIO
-    SIGUNUSED, // Alias for 31
-};
+    SIGINFO];
 
 pub const NSIG: libc::c_int = 32;
+
+pub struct SignalIterator {
+    next: usize,
+}
+
+impl Iterator for SignalIterator {
+    type Item = Signal;
+
+    fn next(&mut self) -> Option<Signal> {
+        if self.next < SIGNALS.len() {
+            let next_signal = SIGNALS[self.next];
+            self.next += 1;
+            Some(next_signal)
+        } else {
+            None
+        }
+    }
+}
+
+impl Signal {
+    pub fn iterator() -> SignalIterator {
+        SignalIterator{next: 0}
+    }
+
+    // We do not implement the From trait, because it is supposed to be infallible.
+    // With Rust RFC 1542 comes the appropriate trait TryFrom. Once it is
+    // implemented, we'll replace this function.
+    #[inline]
+    pub fn from_c_int(signum: libc::c_int) -> Result<Signal> {
+        match 0 < signum && signum < NSIG {
+            true => Ok(unsafe { mem::transmute(signum) }),
+            false => Err(Error::invalid_argument()),
+        }
+    }
+}
+
+pub const SIGIOT : Signal = SIGABRT;
+pub const SIGPOLL : Signal = SIGIO;
+pub const SIGUNUSED : Signal = SIGSYS;
 
 bitflags!{
     flags SaFlags: libc::c_int {
@@ -80,7 +187,6 @@ pub struct SigSet {
     sigset: libc::sigset_t
 }
 
-pub type SigNum = libc::c_int;
 
 impl SigSet {
     pub fn all() -> SigSet {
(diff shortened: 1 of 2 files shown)


src/sys/signal.rs
```diff
@@ -6,44 +6,125 @@ use {Errno, Result};
...
-    SIGWINCH,
-    SIGUSR1,
-    SIGUSR2,
-};
-
-// This doesn't always exist, but when it does, it's 7
-pub const SIGEMT: libc::c_int = 7;
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Signal {
+    SIGHUP = libc::SIGHUP as isize,
```
fiveop: I would have prefered `#[repr(libc::c_int)]` but that does not seem to be possible. Maybe that's worth an RFC for the language.

src/sys/signal.rs
```diff
@@ -6,44 +6,125 @@ use {Errno, Result};
...
+    SIGIO = libc::SIGIO as isize,
+    #[cfg(not(target_os = "macos"))]
+    SIGPWR = libc::SIGPWR as isize,
+    SIGSYS = libc::SIGSYS as isize,
+    #[cfg(target_os = "macos")]
+    SIGEMT = libc::SIGEMT as isize,
+    #[cfg(target_os = "macos")]
+    SIGINFO = libc::SIGEMT as isize,
+}
+
```
fiveop: The following type and function are only needed internally. They are used in the implementation of the iterator and in `wait.rs`. Because of the latter I hade to make them public (for now).

I propose putting the whole code into a private module (src/internal/sys/signal.rs) and then reexport what is supposed to be public here.

src/sys/signal.rs
```diff
@@ -6,44 +6,125 @@ use {Errno, Result};
...
+pub const SIGUNUSED : Signal = SIGSYS;
 
 pub const NSIG: libc::c_int = 32;
 
+struct SignalIterator {
+    signum : libc::c_int,
+}
+
+impl Iterator for SignalIterator {
+    type Item = Signal;
```
fiveop: I considered a nother method of iterating over enumerations. It uses an array of all elements and uses the slice module to produce and iterator. I think the following is a little clearer than an array which contains all the elements of the enumeration again. In particular, when the elements of the enumeration differ between platforms.

src/sys/signal.rs
```diff
@@ -6,44 +6,125 @@ use {Errno, Result};
...
+            self.signum += 1;
+            Some(signal_from(self.signum))
+        } else {
+            None
+        }
+    }
+}
+
+impl Signal {
+    fn iterator() -> SignalIterator {
```
fiveop: This should probably be public.

src/sys/signal.rs
```diff
@@ -86,40 +166,33 @@ impl SigSet {
         SigSet { sigset: sigset }
     }
 
-    pub fn add(&mut self, signum: SigNum) -> Result<()> {
-        let res = unsafe { libc::sigaddset(&mut self.sigset as *mut libc::sigset_t, signum) };
-
-        Errno::result(res).map(drop)
+    pub fn add(&mut self, signal: Signal) {
```
utkarshkukreti: I think we should rename this to `insert` to match the *Map and *Set APIs.

fiveop: We could do that. Though, I would want to do that in a separate PR

fiveop:
@kamalmarhubi I commented on yours. Now, it is your turn ;)

src/sys/signal.rs
kamalmarhubi: I came across the same issue in #364 though I decided to go with `#[repr(i32)]` since the type of the constants in libc crate (`c_int` is `i32` unconditionally), and cast to `c_int` at use. Thinking about it a bit more, I'm not super happy about it, because if libc ever makes `c_int` conditional, we'll have trouble.

kamalmarhubi: And yeah, I think it would be good if the language supported type aliases in `#[repr]` attributes.

src/sys/signal.rs
```diff
@@ -6,44 +6,125 @@ use {Errno, Result};
...
+    SIGPWR = libc::SIGPWR as isize,
+    SIGSYS = libc::SIGSYS as isize,
+    #[cfg(target_os = "macos")]
+    SIGEMT = libc::SIGEMT as isize,
+    #[cfg(target_os = "macos")]
+    SIGINFO = libc::SIGEMT as isize,
+}
+
+// TODO: NOT PUBLIC :/
+pub type SigNum = libc::c_int;
```
kamalmarhubi: I wonder if we could instead define `From<T> for Signal` for various integral types? Our code and user code could then write `Signal::from(9)` for example.

src/sys/signal.rs
kamalmarhubi: For hiding, your proposal is the right thing to do. Though there's no need to put everything in that module I think: circular dependencies among modules in a crate are ok. You could have an internal module depend on this module, and define `signal_from`. Then this module and `wait` could both depend on that module.

src/sys/signal.rs
kamalmarhubi: I'm ok with pushing that change to a later PR. This one can just be the type changing to an enum.

src/sys/signal.rs
```diff
@@ -86,40 +166,33 @@ impl SigSet {
...
     }
 
-    pub fn contains(&self, signum: SigNum) -> Result<bool> {
-        let res = unsafe { libc::sigismember(&self.sigset as *const libc::sigset_t, signum) };
-
-        match try!(Errno::result(res)) {
-            1 => Ok(true),
-            0 => Ok(false),
-            _ => unreachable!("unexpected value from sigismember"),
+    pub fn extend(&mut self, other: &SigSet) {
```
kamalmarhubi: later PR could then also make this take `FromIterator<Item=Signal>`, I like this :-)

src/sys/signal.rs
kamalmarhubi: I think this approach is better.

src/sys/signal.rs
kamalmarhubi: I wonder if it should be `all_signals` or something instead?

kamalmarhubi:
Not for this PR: one thing from looking more closely at this module than I have before: do the thread functions really belong as `impl SigSet`? I think they would more belong on the thread object via some extensions trait.

src/sys/signal.rs
fiveop: I worte a draft for an RFC regarding this: https://github.com/fiveop/rfcs/blob/master/text/0000-type-aliases-in-enumeration-repr-attritubes.md

src/sys/signal.rs
fiveop: I'll implement that.

src/sys/signal.rs
fiveop: I have two problems with this: Its slower. It cannot be used with enumerations whose variant representation dont form a single interval in the natural numbers. Therefore, I actually prefer the other approach, even though it is more verbose and probably makes the executable slightly bigger.

src/sys/signal.rs
fiveop: I thought about the name as well, though I think iterator is pretty descriptive.

src/sys/signal.rs
fiveop: Signal::from(9) fits to the signal_from method above.In that case, I agree. However, I do not want to provide users access to it (as mentioned above).

src/sys/signal.rs
utkarshkukreti: `all_signals` would be a little redundant. How about `all`?

fiveop: `all` (or `all_signals`) implies to me some sort of collection. Why not just use that as the name of the function. `iterator` returns an `Iterator`. `all` could also be an array or some other sort of collection.

src/sys/signal.rs
fiveop: After my last change it is no longer slower. However, we still cannot use its pattern for non consecutive enums.

fiveop:
I worked on this PR again, after adapting code that uses the `sys::signal` module.

Since I have to write functions with C-ABI as signal handlers, which have to take `SigNum` instead of `Signal` because of type safety, I noticed that it does make sense to export the conversion function from `SigNum` to `Signal`, so the `From` trait fits the bill perfectly.

I am now pretty happy with the state of the PR. Any more remarks?

src/sys/signal.rs
```diff
@@ -6,41 +6,92 @@ use {Errno, Result};
...
+    SIGPWR = libc::SIGPWR,
+    SIGSYS = libc::SIGSYS,
+    #[cfg(target_os = "macos")]
+    SIGEMT = libc::SIGEMT,
+    #[cfg(target_os = "macos")]
+    SIGINFO = libc::SIGEMT,
+}
+
+
+pub type SigNum = libc::c_int;
```
kamalmarhubi: one thing I'm wondering is what to do we get out of an extra indirection here? posix specifies they be `int`:

> The <signal.h> header shall define the following macros that are used to refer to the signals that occur in the system. Signals defined here begin with the letters SIG followed by an uppercase letter. The macros shall expand to positive integer constant expressions with type int and distinct values. The value 0 is reserved for use as the null signal (see kill()). Additional implementation-defined signals may occur in the system.

—http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html

src/sys/signal.rs
```diff
@@ -6,41 +6,92 @@ use {Errno, Result};
...
+    }
+}
+
+pub use self::Signal::*;
+
+pub const SIGIOT : Signal = SIGABRT;
+pub const SIGPOLL : Signal = SIGIO;
+pub const SIGUNUSED : Signal = SIGSYS;
 
 pub const NSIG: libc::c_int = 32;
```
kamalmarhubi: is this correct on all platforms with the conditionally included variants?

src/sys/signal.rs
kamalmarhubi: perhaps it's better to always have a static array of all the enum variants then?

src/sys/signal.rs
fiveop: It signals that we do not only want any int but one out of the range for signals.

src/sys/signal.rs
fiveop: Yes. That is what the other PR with the missing signals is for. I counted and hunted the missing ones down :)

src/sys/signal.rs
kamalmarhubi: if the concern is returning a slice iterator, we could newtype it to make it opaque. that would simplify implementation and work for non-contiguous enums.

fiveop: It probably is. I'll change it.

fiveop: Why would returning a slice iterator be a problem?

src/sys/signal.rs
kamalmarhubi: that seems annoying! I guess this is another place where a static array would help: NSIG would be `.len()` on it. There must be some generally accepted way of doing this!

fiveop: `.len()` + 1, because there are actually 31 signals (+ the special 0 signal your quote above mentions :)).

src/sys/signal.rs
kamalmarhubi: but a type alias doesn't do anything to help with that?

kamalmarhubi: oh as documentation somehow, I see. hmmmm...

fiveop: I admit, that one can argue, that it hides the fact that it's unsafe.

fiveop:
I changed how the iterator is created.

src/sys/signal.rs
kamalmarhubi: it's part of the public interface, so a change to the implementation would be breaking.

(my kingdom for an [`impl Trait`](https://github.com/rust-lang/rfcs/pull/1522)!)

fiveop: So we should export our own iterator.

src/sys/signal.rs
kamalmarhubi: I'm now wondering if we should be using `From` for this. From docs:

> **Note: this trait must not fail.** If the conversion can fail, use a dedicated method which return an `Option<T>` or a `Result<T, E>`.
―http://doc.rust-lang.org/std/convert/trait.From.html

utkarshkukreti: The [TryFrom/TryInto RFC](https://github.com/rust-lang/rfcs/pull/1542) was accepted 2 days ago.

fiveop: I will provide our own implementation and add a comment as well as issue regarding the new traits, once we merge this.

kamalmarhubi: @utkarshkukreti amazing! I knew it was being worked on, hadn't realized it was already written and merged.

fiveop:
I worked on this again. Two things changed:

* The `From` implementation was replaced by the `signal_from` function because `From` may not fail. A comment regarding the comming trait `TryFrom` was added.
* `Signal::iterator` now returns our own `Iterator` struct, so that changes in implementation won't break the interface.

kamalmarhubi:
> * The `From` implementation was replaced by the `signal_from` function because `From` may not fail. A comment regarding the comming trait `TryFrom` was added.

Given this change, the motivation for having `SigNum` in our public API seems a bit weaker. Do you think we can remove it now?

kamalmarhubi:
Also @fiveop thanks for being so patient through all the comments!

fiveop:
I removed `SigNum`. With `signal_from` failing properly it makes sense.

Any more concerns?

src/lib.rs
```diff
@@ -8,7 +8,6 @@
 // latest bitflags triggers a rustc bug with cross-crate macro expansions causing dead_code
 // warnings even though the macro expands into something with allow(dead_code)
 #![allow(dead_code)]
-#![deny(warnings)]
```
kamalmarhubi: is there a reason this was removed in this PR?

src/sys/signal.rs
kamalmarhubi: now that the iterator uses an internal array, should `NSIG` be public? should it ever have been public?

kamalmarhubi: also if the `SIGNALS` array is moved up to module level, can this be `SIGNALS.len() + 1`, or is Rust's constexpr support not strong enough for that?

src/sys/signal.rs
```diff
@@ -2,56 +2,162 @@
...
+    SIGEMT = libc::SIGEMT,
+    #[cfg(target_os = "macos")]
+    SIGINFO = libc::SIGEMT,
+}
+
+// We do not implement the From trait, because it is supposed to be infallible.
+// With Rust RFC 1542 comes the appropriate trait TryFrom. Once it is
+// implemented, we'll replace this function.
+#[inline]
+pub fn signal_from(signum: libc::c_int) -> Result<Signal> {
```
kamalmarhubi: Could this perhaps be `Signal::from_c_int`?

kamalmarhubi:
Ok I think I will promise not to make any more requests after these three I just made!

src/lib.rs
fiveop: Probably, because it wouldn't compile otherwise on whatever was the current nightly then. I'll recheck it, when I look at it again.

src/sys/signal.rs
fiveop: Certainly

fiveop:
I only see two messages? What's the third concern?

src/lib.rs
kamalmarhubi: No worries, was just curious.

src/sys/signal.rs
```diff
@@ -2,56 +2,162 @@
...
+    pub fn iterator() -> SignalIterator {
+        SignalIterator{next: 0}
+    }
+}
+
+pub use self::Signal::*;
+
+pub const SIGIOT : Signal = SIGABRT;
+pub const SIGPOLL : Signal = SIGIO;
+pub const SIGUNUSED : Signal = SIGSYS;
```
kamalmarhubi: are these aliases cross-platform?

kamalmarhubi:
@fiveop 
> I only see two messages? What's the third concern?

Honestly no idea! Maybe I can't count past one correctly. Anyway, feel free to r=me once `signal_from` changes. Thanks so much for doing this!

src/sys/signal.rs
fiveop: I still use NSIG in my code (in order to create an function array of NSIG size for handlers). I'll give the `len() + 1` definition a try.

fiveop: And the answer is "No".

fiveop:
finally done?

kamalmarhubi:
@homu r+

:tada:

kamalmarhubi:
@homu r-

kamalmarhubi:
Some failures on Apple?

src/sys/signal.rs
```diff
@@ -2,11 +2,60 @@
...
+    SIGPROF = libc::SIGPROF,
+    SIGWINCH = libc::SIGWINCH,
+    SIGIO = libc::SIGIO,
+    #[cfg(not(target_os = "macos"))]
+    SIGPWR = libc::SIGPWR,
+    SIGSYS = libc::SIGSYS,
+    #[cfg(target_os = "macos")]
+    SIGEMT = libc::SIGEMT,
+    #[cfg(target_os = "macos")]
+    SIGINFO = libc::SIGEMT,
```
utkarshkukreti: @kamalmarhubi typo here?

utkarshkukreti: cc @fiveop

fiveop:
this should be good to merge now as well (finally! :))

posborne:
Although @kamalmarhubi did most of the reviewing before, I looked things over again and it looks good to me.  @homu r+

PR #364, opened 2016-04-30
Add getrlimit(2) and setrlimit(2)

kamalmarhubi:


src/sys/resource.rs
```diff
@@ -0,0 +1,48 @@
+use std::mem;
+
+use libc::{self, c_int};
+
+use {Errno, Result};
+
+#[repr(i32)]
+pub enum Resource {
```
kamalmarhubi: should this be `Resource` or `Rlimit` / `RLimit`?

kamalmarhubi: @fiveop thoughts? since you added the convention on enums...

src/sys/resource.rs
```diff
@@ -0,0 +1,49 @@
+use std::mem;
+
+use libc::{self, c_int};
+pub use libc::{rlimit, RLIM_INFINITY};
```
kamalmarhubi: I *think* these should be reexported, but I'm not sure if we are consistently doing that.

src/sys/resource.rs
fiveop: I would prefer `Resource`, because it very clearly expresses what it stands for.

src/sys/resource.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use {Errno, Result};
+
+#[repr(i32)]
```
fiveop: Using the type `i32` here is dangerous. Who says that the constants are represented by 32bit integers on all (current and future) platforms?

I got around that in my pull request by using the default (`isize`) and casting the constants to `isize` in the enumeration, for example: `RLIMIT_CORE = libc::RLIMIT_CORE as isize`.

I would prefer to write `#[repr(libc::c_int)]` or whatever type is used for these constants at the moment in libc. But even that can differ from platform to platform.

Is there a way to conditionalize the repr line on platforms? We certainly can conditionalise the whole enumeration definition on the platform, but that is a lot of repetition, though that certainly can be dealt with using macros.

src/sys/resource.rs
fiveop: What about `RLIM_SAVED_CUR` and `RLIM_SAVED_MAX`?

Why don't we wrap rlimit and provide proper accessor methods? I admit that it would be purely cosmetic in this case, as far as I can see.

src/sys/resource.rs
kamalmarhubi: > Is there a way to conditionalize the repr line on platforms?

This should work:

```rust
#[cfg_attr(target_os = "linux", repr(i32))]
#[cfg_attr(target_os = "some_other", repr(i16))]
pub enum Resource {
    ...
}
```

kamalmarhubi: Also I kind of answered this in https://github.com/nix-rust/nix/pull/362#discussion_r61669895

kamalmarhubi:
Apple failure is because `RLIMIT_RSS` isn't defined in libc for the platform. I have https://github.com/rust-lang/libc/pull/272 to fix it.

src/sys/resource.rs
posborne: `Resource` as a name still feels a little off give the values in the enum.  You are not really getting the rlimit of a resource, you are getting the value of a specific resource limit in the system (one "resource" may have multiple related limits).

I'm not completely sold on this, but I could see naming the enum `ResourceLimit` and allowing get/set on the limit.  E.g.

```rust
let limit =  ResourceLimit::RLIMIT_STACK.get().unwrap();
ResourceLimit::RLIMIT_STACK.set(limit).unwrap();
```

kamalmarhubi: yeah I thought of this kind of approach but I don't know yet how I feel about going so far from the C API.

In related news: I just created https://github.com/nix-rust/rfcs and will create RFC 0000 soon (probably tomorrow) to bootstrap the process. Then we'll have a good place to discuss stuff like this.

src/sys/resource.rs
kamalmarhubi: forgotten! fixed.

src/sys/resource.rs
fiveop: If we add the `get` and `set` methods to the type, the name `ResourceLimit` does make more sense. However, if we don't do that, the variants of the enum actually represent the Resource, whose limit is set/get by the methods `setrlimit`, `getrlimit`.

The more I think about it, the more I like @posborne's proposal.

src/sys/resource.rs
fiveop: What about the rlimit wrapper?

kamalmarhubi: oh I was deciding not to do that since we aren't doing anything with the type. I wanted it to be easy to specify inline, eg

```
setrlimit(RLIMIT_STACK, rlimit { rlim_cur: 1024*1024, rlim_max: RLIM_INFINITY }).unwrap()
```

fiveop: What about `ResourceLimit::new(1024 * 1024, RLIM_INFINITY)` ?

fiveop:
With the new libc version, this should be mergeable. Could you address the open discussion topics, so that we can get this merged.

kamalmarhubi:
I've been busy but should get to this in the next few days. Thanks for the ping!

fiveop:
Any chance for an update?

Susurrus:
@kamalmarhubi You still planning on merging this?

ccope:
I experimented with this today, and ran into strange behavior when trying to change RLIMIT_STACK inside `main()`. I think the rust runtime interferes by initializing the heap and creating guard pages before `main()` gets invoked. I made an issue here: https://github.com/rust-lang/rust/issues/43879

Susurrus:
@ccope It appears that `setrlimit` can't be implemented with the current rustc implementation based on the conversation in rust-lang/rust#43879, correct?

ccope:
I was only specifically testing testing setrlimit with RLIMIT_STACK, and while it doesn't work in rust < 1.20.0 inside the main thread (due to the main thread guard page), I am unable to find any issues* with rust 1.20.0+.

*: No matter what version of rust you use, you need to make sure that variables larger than the initial stack size don't get inlined into the main function before you can adjust the stack size.

Susurrus:
@ccope Can we expose these failures at run time or will things just Not Work and the user will be confused? Additionally we should make sure to clearly document all the invariants involved with using this functionality.

@kamalmarhubi are you still willing to work on this PR and get it merged? If not, @ccope, you willing to step in here?

ccope:
@Susurrus rust will exit with a stack overflow error in rust 1.20+, and the user will have to figure out why RLIMIT_STACK didin't take effect. I'm not sure there's a way to implement automatic checking for this, but documentation would help.

Susurrus:
@ccope As long as this is clearly documented, I think it'll be fine, though I wonder if this falls under Rust's definition of `unsafe` given that it's dealing with memory.

Additionally, I'm not certain I like the API we expose here. I think we're mimicking the C API a little more closely than we should. For example I'd see us replacing the use of `RLIM_INFINITY` with an `Option` type. I also think we should hide the `rlimit` struct entirely from the user and instead do a `(Option<HardLimit>, Option<SoftLimit>)` instead.

Susurrus:
@kamalmarhubi I'm going to go ahead and close this PR because you haven't touched this in over a year.

@ccope If you're interested in pushing this forward, please clone into your repo and file your own PR that we can then iterate on.

Issue #238, opened 2016-01-13
Decide on a rust version support policy


kamalmarhubi:
I just sent a [PR to track stable and beta channels of Rust](https://github.com/carllerche/nix-rust/pull/237). This would drop "official" support for 1.1, on the assumption that users are going to track stable.

Thoughts?

carllerche:
I commented on #237 saying that I think 1.3.0 is good for the oldest official Rust support for now.

kamalmarhubi:
On the topic, what are the pros/cons of supporting older versions? Eg could we support as far back as 1.0.0? (I only started writing Rust a month or two ago, so I don't know what's changed.)

carllerche:
Mostly a question of who is using nix and if they "freeze" their version of Rust. I know that some production Rust users who depend on "unstable" Rust features have frozen version of Rust used internally and only occasionally update it.

That's really the only reason.

kamalmarhubi:
Ok. I've made it 1.1.0+ for now. I'll look into the errors with 1.0.0 to see how reasonable it is to attempt to support all released versions.

fiveop:
I would prefer not to allow failures on nightly. I see three cases why it might fail:

1. We introduced a bug in nix. => We fix it.
2. The language changed in such a way, that our build breaks. => We fix it (if necessary with feature guards).
3. Rust nightly has a bug. => We wait a day or two.

kamalmarhubi:
I've seen broken nightly make my PRs go red at least a couple of times. On the flip side, the stability guarantee means we shouldn't see any language changes that cause breakage. And if they do happen, there will be an entire beta cycle to fix them. We an also keep tabs on nightly ourselves, since it'll be run on every PR.

kamalmarhubi:
Or summarising:

> 1. We introduced a bug in nix. => We fix it.

We should see it in other releases and fix it.

> 2. The language changed in such a way, that our build breaks. => We fix it (if necessary with feature guards).

Shouldn't happen.

> Rust nightly has a bug. => We wait a day or two.

Annoying and disruptive. (Subjective, of course.)

kamalmarhubi:
@arcnmx any thoughts on this? You have some changes in mind that seem to assume bleeding edge rust...

arcnmx:
> I would prefer not to allow failures on nightly.

[...]

PR #365, opened 2016-05-01
Add mkstemp(3)

antifuchs:
I've missed this call in a project that uses nix, so I added it to the `unistd` module; the placement is only a half-truth (as the header is `stdlib.h` on most systems, but some related functions are documented to live in `unistd.h` on BSD-derivatives), so I'm not sure my PR is completely right. But then, there is no `stdlib` module yet! Any guidance here is much appreciated! (-:

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -5,8 +5,10 @@ use fcntl::{fcntl, OFlag, O_NONBLOCK, O_CLOEXEC, FD_CLOEXEC};
 use fcntl::FcntlArg::{F_SETFD, F_SETFL};
 use libc::{self, c_char, c_void, c_int, c_uint, size_t, pid_t, off_t, uid_t, gid_t};
 use std::mem;
-use std::ffi::CString;
+use std::ffi::{CString, OsStr};
+use std::os::unix::ffi::OsStrExt;
 use std::os::unix::io::RawFd;
+use std::path::{PathBuf, Path};
 use void::Void;
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
@@ -375,16 +377,23 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
 }
 
 #[inline]
-pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<RawFd> {
-    let res = try!(template.with_nix_path(|path| {
-        let mut path_copy = path.to_bytes_with_nul().to_owned();
-        let c_template: *mut c_char = path_copy.as_mut_ptr() as *mut c_char;
+pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
+    let res = template.with_nix_path(|path| {
+        let owned_path = path.to_owned();
+        let path_ptr = owned_path.into_raw();
         unsafe {
-            libc::mkstemp(c_template)
+            (libc::mkstemp(path_ptr), CString::from_raw(path_ptr))
         }
-    }));
-    Errno::result(res)
-
+    });
+    match res {
+        Ok((fd, pathname)) => {
+            try!(Errno::result(fd));
+            Ok((fd, Path::new(OsStr::from_bytes(pathname.as_bytes())).to_owned()))
+        }
+        Err(e) => {
+            Err(e)
+        }
+    }
 }
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -374,6 +374,19 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
     unsafe { libc::sleep(seconds) }
 }
 
+#[inline]
+pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<RawFd> {
+    let res = try!(template.with_nix_path(|path| {
+        let mut path_copy = path.to_bytes_with_nul().to_owned();
+        let c_template: *mut c_char = path_copy.as_mut_ptr() as *mut c_char;
+        unsafe {
+            libc::mkstemp(c_template)
+        }
+    }));
+    Errno::result(res)
+
+}
+
 #[cfg(any(target_os = "linux", target_os = "android"))]
 mod linux {
     use sys::syscall::{syscall, SYSPIVOTROOT};
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -47,10 +47,11 @@ fn test_wait() {
 
 #[test]
 fn test_mkstemp() {
-    let result = mkstemp("/tmp/tempfile.XXXXXXXX");
+    let result = mkstemp("/tmp/nix_tempfile.XXXXXXXX");
     match result {
-        Ok(fd) => {
+        Ok((fd, path)) => {
             close(fd).unwrap();
+            unlink(path.as_path()).unwrap();
         }
         Err(e) => panic!("mkstemp failed: {}", e)
     }
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -50,7 +50,7 @@ fn test_mkstemp() {
     let result = mkstemp("/tmp/tempfile.XXXXXXXX");
     match result {
         Ok(fd) => {
-            close(fd).expect("Couldn't close the file descriptor");
+            close(fd).unwrap();
         }
         Err(e) => panic!("mkstemp failed: {}", e)
     }
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -45,6 +45,16 @@ fn test_wait() {
     }
 }
 
+#[test]
+fn test_mkstemp() {
+    let result = mkstemp("/tmp/tempfile.XXXXXXXX");
+    match result {
+        Ok(fd) => {
+            close(fd).expect("Couldn't close the file descriptor");
+        }
+        Err(e) => panic!("mkstemp failed: {}", e)
+    }
+}
 
 #[test]
 fn test_getpid() {


src/unistd.rs
```diff
@@ -374,6 +374,19 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
...
+pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<RawFd> {
+    let res = try!(template.with_nix_path(|path| {
+        let mut path_copy = path.to_bytes_with_nul().to_owned();
+        let c_template: *mut c_char = path_copy.as_mut_ptr() as *mut c_char;
+        unsafe {
+            libc::mkstemp(c_template)
+        }
+    }));
+    Errno::result(res)
+
```
posborne: Probably move the newline to before the final expression rather than after it.

posborne:
Looks good to me.  I agree that the placement within unistd is a bit odd.  I'll let others take a look before merging to decide if we want something like a `stdlib` module or something else -- most `stdlib.h` functions have not been implemented as they have suitable alternatives in `std`.

Right now, I would lean toward creating that module versus putting it in `unistd`.

kamalmarhubi:
> Right now, I would lean toward creating that module versus putting it in `unistd`.

Same.

One thing that's slightly concerning: we are unable to communicate the actual file name back to the caller: because of how `NixPath` works, `libc::mkstemp` modifies a stack allocated buffer with a copy of the passed in template.

Pinging #221 #230 since this use case might matter for them.

kamalmarhubi:
@antifuchs 
> I've missed this call in a project that uses nix, so I added it to the `unistd` module

Aside: for temporary files, I usually use either the [tempdir](https://crates.io/crates/tempdir) or [tempfile](https://crates.io/crates/tempfile) crates depending on the exact use case. tempfile's focus is on creating an unnamed file that is not linked into the filesystem. Most often I use tempdir and create files in there.

kamalmarhubi:
Test failures are for Rust < 1.4, which didn't have `Result::expect`. I really like that method. :/

Ping #238.

antifuchs:
Thanks a lot for the comments! Those are some excellent points (and agreed to a new module)! My use case is specifically for a fresh file descriptor (not a `File`) opened to an unlinked file, so this exact library call was perfect for me (and e.g. `tempfile` wouldn't be).

While I'm not sure what one would use the filename of an unlinked file for, losing the value isn't great either - we could either require a `&mut str` as the template, or return the actual path as a fresh string, both approximately equally displeasing for different reasons (-:

I'm going to remove the depentency on `Result::expect`, and will move that pesky newline, too.

kamalmarhubi:
I thought that `mkstemp` will result in a linked file that you have to unlink yourself.

/me checks

kamalmarhubi:
Yeah, mkstemp results in a linked file:

```
> cat tmpfiletest.c
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>
#include <unistd.h>

int main(void) {
        char filename[128];
        int fd;
        struct stat stat_buf;

        strncpy(filename, "/tmp/my-awesome-tmpfile.XXXXXX", sizeof(filename));

        fd = mkstemp(filename);
        if (fd == -1) {
                perror("mkstemp");
                exit(EXIT_FAILURE);
        }
        printf("file name is %s\n", filename);

        if (stat(filename, &stat_buf)) {
                perror("stat");
                exit(EXIT_FAILURE);
        }
        printf("file's inode: %ld\n", stat_buf.st_ino);
        pause();

        return 0;
}
$ gcc -Wall -Wextra -Werror tmpfiletest.c
$ ./a.out &
file name is /tmp/my-awesome-tmpfile.bwEU7q
file's inode: 3830951
$ ls -l /tmp/my-awesome-tmpfile.bwEU7q
-rw------- 1 kamal kamal 0 May  1 19:00 /tmp/my-awesome-tmpfile.bwEU7q
$ man ls
$ ls -il /tmp/my-awesome-tmpfile.bwEU7q
3830951 -rw------- 1 kamal kamal 0 May  1 19:00 /tmp/my-awesome-tmpfile.bwEU7q
```

kamalmarhubi:
Non-portably, the best approach on Linux is to use `open(2)` with `O_TMPFILE`.

kamalmarhubi:
> we could either require a `&mut str` as the template

ugh strings and nix are so annoying. this is not ideal because `str` is always valid unicode, and we'd have to document the nul-termination and no inner nul requirement on the `str`.

/me goes to finish setting up the RFCs repo so we can get past #221.

antifuchs:
Ahaha, eep, I totally missed the non-unlinkedness. You're completely right (and gosh, I should have taken a look at my `/tmp`). I'll adjust the test accordingly.

kamalmarhubi:
The best I can think of for the template right now is to take it as `&mut [CStr](http://doc.rust-lang.org/std/ffi/struct.CStr.html)`. I think once we figure out `NixString` (#221, #230) we *may* be able to open it up to `&mut T where T: NixString`, but not sure.

src/unistd.rs
```diff
@@ -375,16 +377,23 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
 }
 
 #[inline]
-pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<RawFd> {
-    let res = try!(template.with_nix_path(|path| {
-        let mut path_copy = path.to_bytes_with_nul().to_owned();
-        let c_template: *mut c_char = path_copy.as_mut_ptr() as *mut c_char;
+pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
```
kamalmarhubi: I don't like this too much because it forces an allocation and copy. See https://github.com/nix-rust/nix/pull/365#issuecomment-216374309 for another option.

antifuchs: Yeah, I don't much like that. Mutating the input parameter feels like a hack and a bit of a safety hazard (though the `mut` mark makes it a bit less terrifying), to be honest. Plus, `CStr` input parameters are always kind of messy - I just had reason to use `memfd_create`, and its signature is kind of a pain.

kamalmarhubi: It won't be a safety hazard unless you unsafely obtained the mutable reference :-)

kamalmarhubi: `memfd_create` is like `open` so it should be fine?

arcnmx:
> The best I can think of for the template right now is to take it as &mut CStr

As unfortunate as the type is, it does fit best... There's no way to create one though, even `CString` doesn't impl `DerefMut` in the way `String` does (and `&mut str` isn't even that well of a supported type). nix would have to provide some extension traits, at which point we may just be better off going with `&mut [u8]`, and panic/error on invalid input (null bytes in wrong spot).

> I think once we figure out NixString (#221, #230) we may be able to open it up to &mut T where T: NixString, but not sure.

that will be very tricky, especially if we continue to support fallible one-way conversions...

kamalmarhubi:
@arcnmx 
> There's no way to create one though, even `CString` doesn't impl `DerefMut` in the way `String` does (and `&mut str` isn't even that well of a supported type).

Huh, that's unfortunate. I'll take a look at adding at least the `DerefMut` impl, which seems like an oversight.

It would be nice to have an RFC like https://github.com/rust-lang/rfcs/pull/1309 for `CString` to fill in some gaps. Not all the gaps, since some of it doesn't make sense with the specified API of `CStr` not storing a length.

> that will be very tricky, especially if we continue to support fallible one-way conversions...

Maybe a different trait for the mutable argument case?

arcnmx:
> Huh, that's unfortunate. I'll take a look at adding at least the DerefMut impl, which seems like an oversight.

At the same time maybe try adding some `&mut self` methods to `CStr` like `unsafe to_bytes_mut` and `unsafe to_bytes_with_nul_mut`. `from_bytes_with_nul_mut` and `from_bytes_with_null_mut_unchecked` would be nice too..! Otherwise there's not much you can really do with a `&mut CStr`...

(oh god these long names though, why couldn't we have just named them `bytes` and `inner_bytes` instead...)

> It would be nice to have an RFC like rust-lang/rfcs#1309 for CString to fill in some gaps. Not all the gaps, since some of it doesn't make sense with the specified API of CStr not storing a length.

I'm not sure this is necessary, since the conversion from `&CStr` to `&OsStr` is already cost-free (minus the length check). Maybe we could add `from_c_str()` to `std::os::unix::ffi::OsStrExt` for convenience, since the current conversion is `OsStrExt::from_bytes(cstr.to_bytes())`

philippkeller:
Since I also wanted to implement mkstemp in unistd I found this open PR.

I fixed it. This is the working diff: [nix-mkstemp.diff](https://gist.github.com/philippkeller/e92f1ecb86641e93846e2ebd7e9d53d5)

A few remarks

- @antifuchs if you don't care about this open PR then I can start a new one, I didn't want to steal away your code and claim it's mine. Your code only needed like 2-3 line of fixing.
- made it working with Rust 1.2 which needed `to_bytes_with_nul()`. I think the implementation is memory safe but would like to have a pair of eyes checking that
- added documentation
- replaced `Path` by `PathBuf` so it's more in line with `getcwd`
- I didn't move it into another module, I would like to do that but in a separate PR (probably moving all stdio methods out)
- it's true that unistd doesn't *need* `mkstmp` since there is the crate tempdir and tempfile, but I'd love to see this here for completeness

fiveop:
You can create a fresh PR that builds on antifuchs's work:

* add antifuchs' fork as a remote to your local git repository (edit `.git/config`)
  ```
  [remote "antifuchs"]
  url = git@github.com:antifuchs/nix.git
  fetch = +refs/heads/*:refs/remotes/antifuchs/*
   ```
* fetch from him: `git fetch antifuchs`
* checkout his branch: `git checkout youllhavetolookupthebranchnameyourself`
* add your commits
* push to your fork: `git push -u yourremotename whateverthebranchiscalled`
* and open a PR

(I'm sorry, if I bore you with things you already know).

philippkeller:
@fiveop thanks for the explanation. No, I didn't know this already. What's the advantage of doing it the way you suggested against just doing a 100% fresh PR with his `mkstemp` and `test_mkstemp`?

fiveop:
His commits will show up in the git history (instead of just his code).

philippkeller:
@fiveop thanks, I opened https://github.com/nix-rust/nix/issues/428 now, I guess this PR can be closed then

antifuchs:
Ahh, sorry for letting this sit so long & thanks for working on this, @philippkeller! I'll close this PR now in favor of yours!

kamalmarhubi:
@fiveop @philippkeller an improvement on the get-the-PR workflow: `git fetch origin refs/pulls/365/head && git checkout -b my-new-branch-name FETCH_HEAD` :-)

fiveop:
thanks :)

PR #367, opened 2016-05-02
Add SO_ORIGINAL_DST

justinlatimer:
In Linux, the SO_ORIGINAL_DST socket option can be used to get the original destination, which can be needed if the connection is translated by a NAT, i.e. iptables. In C, this information can be obtained by ``getsockopt(fd, SOL_IP, SO_ORIGINAL_DST, &remote, &remote_len)`` and returns a ``sockaddr_in`` struct. I've added a binding for this option.

Thanks!

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -2,6 +2,8 @@ use super::{ffi, consts, GetSockOpt, SetSockOpt};
 use {Errno, Result};
 use sys::time::TimeVal;
 use libc::{c_int, uint8_t, c_void, socklen_t};
+#[cfg(target_os = "linux")]
+use libc::sockaddr_in;
 use std::mem;
 use std::os::unix::io::RawFd;
 
@@ -168,6 +170,8 @@ sockopt_impl!(GetOnly, SockType, consts::SOL_SOCKET, consts::SO_TYPE, super::Soc
           target_os = "linux",
           target_os = "nacl"))]
 sockopt_impl!(GetOnly, AcceptConn, consts::SOL_SOCKET, consts::SO_ACCEPTCONN, bool);
+#[cfg(target_os = "linux")]
+sockopt_impl!(GetOnly, OriginalDst, consts::SOL_IP, consts::SO_ORIGINAL_DST, sockaddr_in);
 
 /*
  *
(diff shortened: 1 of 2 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -1,7 +1,7 @@
 use super::{ffi, consts, GetSockOpt, SetSockOpt};
 use {Errno, Result};
 use sys::time::TimeVal;
-use libc::{c_int, uint8_t, c_void, socklen_t};
+use libc::{c_int, uint8_t, c_void, socklen_t, sockaddr_in};
```
fiveop: This needs to be a conditional use only for linux (see Travis-CI result).

fiveop:
When you fix the use error on apple, we can merge this.

justinlatimer:
@fiveop Updated to fix the apple error, thanks!

src/sys/socket/sockopt.rs
```diff
@@ -2,6 +2,8 @@ use super::{ffi, consts, GetSockOpt, SetSockOpt};
 use {Errno, Result};
 use sys::time::TimeVal;
 use libc::{c_int, uint8_t, c_void, socklen_t};
+#[cfg(target_os = "linux")]
+use libc::{sockaddr_in};
```
kamalmarhubi: tiny nit: no need for braces if there's just a single name imported

justinlatimer:
@kamalmarhubi braces removed, thanks!

kamalmarhubi:
@justinlatimer yay thanks! could you squash down to a single commit, and we'll get this merged :-)

justinlatimer:
@kamalmarhubi all squashed, thanks!

kamalmarhubi:
yay!

@homu r+ 3954580

sorz:
Should this return `sockaddr` instead of `sockaddr_in` for compatible with IPv6?

asomers:
@sorz I can't find any documentation on `SO_ORIGINAL_DST` anywhere.  I see another symbol defined named `IP6T_SO_ORIGINAL_DST`, but I don't know exactly what either does.  You'll need to dive into the Linux source, I'm afraid, if you want to find out.

PR #274, opened 2016-02-19
New syscalls

abbradar:
Part of #270

PR #369, opened 2016-05-04
Use Wrapping for intended underflow of unsigned integer value.

fiveop:
This fixes an warning made error on nightly builds.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -114,8 +114,12 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
 #[inline]
 pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<uid_t>, group: Option<gid_t>) -> Result<()> {
     let res = try!(path.with_nix_path(|cstr| {
-        // We use `0 - 1` to get `-1 : {u,g}id_t` which is specified as the no-op value for chown(3).
-        unsafe { libc::chown(cstr.as_ptr(), owner.unwrap_or(0 - 1), group.unwrap_or(0 - 1)) }
+        // According to the POSIX specification, -1 is used to indicate that
+        // owner and group, respectively, are not to be changed. Since uid_t and
+        // gid_t are unsigned types, we use wrapping_sub to get '-1'.
+        unsafe { libc::chown(cstr.as_ptr(),
+                             owner.unwrap_or((0 as uid_t).wrapping_sub(1)),
+                             group.unwrap_or((0 as gid_t).wrapping_sub(1))) }
     }));
 
     Errno::result(res).map(drop)


src/unistd.rs
```diff
@@ -111,11 +111,15 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+
```
utkarshkukreti: Any reason for adding these 2 lines?

fiveop: No. I'll fix it.

src/unistd.rs
```diff
@@ -114,8 +114,10 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
 #[inline]
 pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<uid_t>, group: Option<gid_t>) -> Result<()> {
     let res = try!(path.with_nix_path(|cstr| {
-        // We use `0 - 1` to get `-1 : {u,g}id_t` which is specified as the no-op value for chown(3).
```
utkarshkukreti: I think this comment is still relevant. What about you @fiveop?

fiveop: I thought about it; but then, we are not documenting all the other functions properly. So, for consistency I removed it. :)

If you do mind, I'll put it in again.

utkarshkukreti: I think this is kind of a weird value for a no-op and that line makes it a little clearer what we're trying to do with `(Wrapping(0 as uid_t) - Wrapping(1 as uid_t)).0`. Any opinions @kamalmarhubi?

fiveop: I'll add a comment.

kamalmarhubi: I remember asking for the comment back in https://github.com/nix-rust/nix/pull/274#discussion_r53517520 so... :-)

(I think it warrants the comment because it's a weird thing to do.)

fiveop:
I readded the (reworded) comment.

src/unistd.rs
```diff
@@ -114,8 +114,13 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
...
-        unsafe { libc::chown(cstr.as_ptr(), owner.unwrap_or(0 - 1), group.unwrap_or(0 - 1)) }
+        // According to the POSIX specification, -1 is used to indicate that
+        // owner and group, respectively, are not to be changed. Since uid_t and
+        // gid_t are unsigned types, we use std::num::Wrapping to get '-1'.
+        use std::num::Wrapping;
+        unsafe { libc::chown(cstr.as_ptr(),
+                             owner.unwrap_or((Wrapping(0 as uid_t) - Wrapping(1 as uid_t)).0),
+                             group.unwrap_or((Wrapping(0 as gid_t) - Wrapping(1 as gid_t)).0)) }
```
kamalmarhubi: one last nit, can we use `0.wrapping_sub(1)` instead of using `Wrapping`? I think it'll be easier to read and doesn't require an extra import.

fiveop: Certainly! If I had known the function, I would have used it right away.

kamalmarhubi:
@homu r+ 2139c90

PR #370, opened 2016-05-05
Provide accessors to sigmask of UContext.

fiveop:


src/ucontext.rs
```diff
@@ -22,4 +23,13 @@ impl UContext {
         };
         Errno::result(res).map(drop)
     }
+
+    #[allow(mutable_transmutes)]
+    pub fn mut_sigmask(&mut self) -> &mut SigSet {
```
fiveop: `mut_sigmask` or `sigmask_mut`?

fiveop: For `DerefMut` it's `deref_mut`, so I went with `sigmask_mut`.

fiveop: Another example that supports my choice of name:

https://github.com/rust-lang/rust/commit/9370d3a05107bdcac6fb62d3bd5380e1d7c79b5b

src/ucontext.rs
```diff
@@ -22,4 +23,13 @@ impl UContext {
         };
         Errno::result(res).map(drop)
     }
+
+    #[allow(mutable_transmutes)]
```
kamalmarhubi: This attribute can be dropped if you do

```
unsafe { mem::transmute(&mut self.context.uc_sigmask) }
```

fiveop: Thanks you. I have missed that in some other code of mine as well. Any other comments?

fiveop:
Can we merge this now?

posborne:
Looks good.  @homu r+

PR #371, opened 2016-05-10
Exclude UContext::get(),set() on musl, not available from libc.

justinlatimer:
It looks like ``libc`` doesn't provide ``libc::getcontext`` or ``libc::setcontext`` when targeting musl, so I've excluded the ``ucontext`` module from being built when targeting musl which allows the module to be built successfully.

Thanks!

### DIFF
--- a/src/ucontext.rs
+++ b/src/ucontext.rs
@@ -1,5 +1,7 @@
 use libc;
+#[cfg(not(target_env = "musl"))]
 use {Errno, Result};
+#[cfg(not(target_env = "musl"))]
 use std::mem;
 
 #[derive(Clone, Copy)]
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Could you just disable those methods on the `UContext` type instead? The struct itself is defined for musl, so it'd be nice to still have it. (See https://github.com/rust-lang/libc/pull/255#issuecomment-207072979 for a tiny bit of context.)

justinlatimer:
@kamalmarhubi updated to only exclude functions that don't compile (and their imports) + squashed.

Thanks!

kamalmarhubi:
Thanks @justinlatimer!

@homu r+ e3e92d7

kamalmarhubi:
@fiveop I decided that since this was more conservative that the original where the struct was also removed, it was fine. Hope that makes sense!

PR #372, opened 2016-05-11
Fix a typo in travis config.

justinlatimer:
Fix a typo in the target name for the config of ``mipsel-unknown-linux-gnu``. Build will probably still fail though, but will hopefully get further :wink:

Thanks!


posborne:
Yep, will probably still fail, but looks good.  @homu r+

PR #374, opened 2016-05-26
Fix deployment of docs from travis.

### LINKED ISSUE #320: Docs are not being updated in S3
Compare, for example:

rustdoc.s3-website-us-east-1.amazonaws.com/nix/master/linux/src/nix/sys/uio.rs.html
https://github.com/nix-rust/nix/blob/master/src/sys/uio.rs

The S3 version is missing the fact that #304 got rid of `mod ffi` in that module.

justinlatimer:
The problem in the repo line in the travis config, I guess it hasn't been updated since ownership was transferred. See: https://travis-ci.org/nix-rust/nix/jobs/131751025#L614

Should resolve #320 

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,6 @@ deploy:
   acl: public_read
   on:
     condition: "\"$TRAVIS_RUST_VERSION/$ARCH\" == \"1.1.0/x86_64\""
-    repo: carllerche/nix-rust
+    repo: nix-rust/nix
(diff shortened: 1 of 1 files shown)


fiveop:
@homu r+

(I can see changes of c1334f5068b9c404be64325bbf90c0b7ff4c00d0 on the documentation page, so it probably worked)

PR #377, opened 2016-06-10
Added lseek and seek64

aoprisan:
I have added lseek and seek64 to unistd, the last one targeting Linux/Android only. I wasn't sure where to place the Whence enum, or if it's a nice of doing, I am quite fresh to Rust.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -214,6 +214,40 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
     Errno::result(res).map(|r| r as usize)
 }
 
+pub enum Whence {
+    SeekSet,
+    SeekCur,
+    SeekEnd,
+    SeekData,
+    SeekHole
+}
+
+impl Whence {
+    fn to_libc_type(&self) -> c_int {
+        match self {
+            &Whence::SeekSet => libc::SEEK_SET,
+            &Whence::SeekCur => libc::SEEK_CUR,
+            &Whence::SeekEnd => libc::SEEK_END,
+            &Whence::SeekData => 3,
+            &Whence::SeekHole => 4
+        }
+    }
+
+}
+
+pub fn lseek(fd: RawFd, offset: libc::off_t, whence: Whence) -> Result<libc::off_t> {
+    let res = unsafe { libc::lseek(fd, offset, whence.to_libc_type()) };
+
+    Errno::result(res).map(|r| r as libc::off_t)
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn lseek64(fd: RawFd, offset: libc::off64_t, whence: Whence) -> Result<libc::off64_t> {
+    let res = unsafe { libc::lseek64(fd, offset, whence.to_libc_type()) };
+
+    Errno::result(res).map(|r| r as libc::off64_t)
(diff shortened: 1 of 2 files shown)


test/test_unistd.rs
```diff
@@ -112,6 +119,38 @@ macro_rules! execve_test_factory(
...
+    let mut tmp = tempfile().unwrap();
+    tmp.write(CONTENTS).unwrap();
+
+    lseek64(tmp.as_raw_fd(), 5, Whence::SeekSet).unwrap();
+
+    let mut buf = String::new();
+     tmp.read_to_string(&mut buf).unwrap();
+     assert_eq!(b"f123456", buf.as_bytes());
+
+     close(tmp.as_raw_fd()).unwrap();
```
posborne: nit: spacing in function seems inconsistent (some of it is 5 spaces)

src/unistd.rs
```diff
@@ -214,6 +214,36 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
...
+    SeekCur,
+    SeekEnd
+}
+
+impl Whence {
+    fn to_libc_type(&self) -> c_int {
+        match self {
+            &Whence::SeekSet => 0,
+            &Whence::SeekCur => 1,
+            &Whence::SeekEnd => 2
```
posborne: Might be worth pursuing adding these constant definitions to libc where we will have testing that they match the libc implementations.

src/unistd.rs
```diff
@@ -214,6 +214,36 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
     Errno::result(res).map(|r| r as usize)
 }
 
+pub enum Whence {
+    SeekSet,
+    SeekCur,
+    SeekEnd
```
posborne: Linux >= 3.1 also supports `SEEK_DATA` and `SEEK_HOLE` which might be worth supporting as well.

posborne:
Please also squash commits down to a single commit as well.

test/test_unistd.rs
aoprisan: Sorry abou that, I rushed a bit, I will fix it.

src/unistd.rs
aoprisan: Should I use those constants directly ? What made me not do it in the first place was that the type is too permissive, any c_int could be sent; and that I wan't sure how much of libc's types should be in nix's API.

src/unistd.rs
aoprisan: If I use the constants from libc, which do you think it's the best place to add these two ?

src/unistd.rs
posborne: No, I think the enum is a good thing.  I would just change the returned value to be the constant from libc.

aoprisan:
All done, let me know if anything else has to be fixed.

posborne:
Sorry for the delay @aoprisan.  A couple minor things remain, but we can probably address these in the future:

* I think one of the tests still has a minor indentation issue
* I would like to eventually see the additional constants added to the libc crate and refer to the constants here
* It may make sense to conditionally expose the extra linux constant for Linux only (will be necessitated by point 2)

That being said, that can all be fixed later.  So, @homu r+

PR #379, opened 2016-06-10
Add missing wait flag WUNTRACED for non-Linux systems

nikklassen:
My understanding is that this flag is required by POSIX, so all systems should allow for it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc = "0.2.11"
+libc = "0.2.13"
 bitflags = "0.4"
 cfg-if = "0.1.0"
 void = "1.0.2"
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -16,6 +16,7 @@ mod ffi {
 bitflags!(
     flags WaitPidFlag: c_int {
         const WNOHANG     = 0x00000001,
+        const WUNTRACED   = 0x00000002,
     }
 );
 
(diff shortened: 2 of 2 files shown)


posborne:
Hi @nikklassen, thanks for the change!

We have been on a push to move all constant definitions like this to the [libc crate](https://github.com/rust-lang/libc).  It appears that `WNOHANG` is present there but `WUNTRACED` is not (currently).

I think our preferred change here would be:
1. Add `WUNTRACED` to libc
2. Issue PR against nix modifying the bitfield and also updating to use the constants defined within libc.

Adding to libc gets us additional verification on the constant (via the testing done in libc) and also makes the constant available to Rust users that may not be using nix.

nikklassen:
Sounds good. I'll look into that.

posborne:
Hi @nikklassen Any luck?  Need any help?  Want somebody else to take over the task?

nikklassen:
I got my pull request to libc accepted, I'm just waiting for them to bump the version so that I can update the code here. Unfortunately I don't know how long that'll take

posborne:
> I got my pull request to libc accepted, I'm just waiting for them to bump the version so that I can update the code here. Unfortunately I don't know how long that'll take

Sounds good.  @alexcrichton any chance of a version bump soon?  I have another change into libc that I am waiting on as well.  The alternative (CC @kamalmarhubi) would be to have nix master track libc in git and then to synchronize on releases (this has some risk but not too big as there doesn't usually seem to be a big delay in libc releases).

alexcrichton:
Sure! I basically just do a release of libc whenever someone requests it, so feel free to request it at any time :)

fiveop:
Now, replacing the two numbers by their respective libc constants is all that is left to do.

fiveop:
@homu r+

PR #383, opened 2016-07-03
eventfd: Follow nix conventions

kamalmarhubi:
This commit revamps to eventfd to follow nix conventions:
  - drop in-crate FFI definitions
  - rename EventFdFlag to EfdFlags

Additionally, it changes the initval argument to be a libc::c_uint,
matching the actual type.


posborne:
Looks good to me.  @homu r+

PR #384, opened 2016-07-09
Add epoll_create1.

kubo39:


src/sys/epoll.rs
```diff
@@ -80,6 +88,13 @@ pub fn epoll_create() -> Result<RawFd> {
 }
 
 #[inline]
+pub fn epoll_create1(flags: EpollFdFlag) -> Result<RawFd> {
+    let res = unsafe { ffi::epoll_create1(flags.bits() | EPOLL_CLOEXEC.bits) };
```
alex-gulyas: - The way you are adding the default flag is actually enforcing it, therefore making it impossible for someone to call `epoll_create1` without the `EPOLL_CLOEXEC` flag.
- I don't think that any of the flags should be turned on by default, especially not without documentation. (Principle of Least Surprise)

kubo39: In Rust's standard library close-on-exec flag is set by default.
I'm not sure why so, however, I think it's security reason.
Thugh, someone might be surpised when I force to using `fcntl(fd, F_SETFD, !O_CLOEXEC)` if he want.
Okay, I'll fix it.

src/sys/epoll.rs
```diff
@@ -8,12 +8,20 @@ mod ffi {
 
     extern {
         pub fn epoll_create(size: c_int) -> c_int;
+        pub fn epoll_create1(flags: c_int) -> c_int;
         pub fn epoll_ctl(epfd: c_int, op: c_int, fd: c_int, event: *const EpollEvent) -> c_int;
         pub fn epoll_wait(epfd: c_int, events: *mut EpollEvent, max_events: c_int, timeout: c_int) -> c_int;
```
posborne: Hi @kubo39 , although this is not directly related to your PR we have been moving to defining ffi constants entirely in libc and using those rather than defining our own (see [CONVENTIONS.md](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#libc-constants-functions-and-structs).  it looks like the existing constants are present but `epoll_create1` is not.  Can you please do the following:

1. Open up a PR to libc adding `epoll_create1`
2. Update `epoll.rs` to use the ffi definitions from libc.

Also, thanks for the change!

fiveop:
There is a new libc version with your changes. Now, you can amend the pull request to use the new libc version and your additions to it. After that, we can merge this pull request.

posborne:
@kubo39 Can you just rebase your commits on master rather than having the merge commit back into your branch?  Other than that, looks good to me.

fiveop:
This will probably create a merge conflict, because of the libc version. Please use the existing form of the dependency targeting the git repository instead of a fixed version.

kubo39:
I'm sorry for being late.

posborne:
@kubo39 I think we want to remove the extern functions from nix and just depend on them directly from libc.

src/sys/epoll.rs
```diff
@@ -8,12 +8,20 @@ mod ffi {
...
     }
 }
 
 bitflags!(
+    flags EpollFdFlag: c_int {
+        const EPOLL_NONBLOCK = 0x800,
+        const EPOLL_CLOEXEC = 0x80000
```
posborne: These constants should also come out of libc if possible.

kamalmarhubi:
Work on this seems to be continuing in #410, so I'm closing this one. Thanks for starting on this @kubo39!

PR #385, opened 2016-07-10
Publish docs to gh-pages, add coveralls, update README

posborne:
See the commit messages for some additional details.  This started out with just an update to publish the docs to gh-pages but snowballed into a few other changes along the way.


posborne:
r? @kamalmarhubi @fiveop When I started out, I was thinking that the docs weren't being updated at all.  It looks like they might be after all.

So, reasons to move to Github Pages:

* Don't need to worry abut maintaining the seperate S3 bucket (I believe this done by @carllerche now), although there probably is not a significant amount of work/cost here.
* Access goes along with contributors.
* Most other projects I have seen publish docs to gh-pages

Downsides:

* Currently, we are just publishing the docs for Linux on x86_64 with the default options on master.  We should be able to make OSX work fine, but we'll need to get a bit creative.

README.md
```diff
@@ -1,18 +1,32 @@
...
-API documentation generated by rustdoc:
+As an example of what Nix provides, examine the differences between what is
+exposed by libc and nix for the
+[gethostname](http://man7.org/linux/man-pages/man2/gethostname.2.html) system
+call:
 
-- Linux: [master](http://rustdoc.s3-website-us-east-1.amazonaws.com/nix/master/linux/nix/)
-- OS X: [master](http://rustdoc.s3-website-us-east-1.amazonaws.com/nix/master/osx/nix/)
+```rust,ignore
+// libc api (unsafe, requires handing return code/errno)
```
fiveop: handling?

README.md
```diff
@@ -25,10 +39,14 @@ nix = "0.6.0"
 
 Then, add this to your crate root:
 
-```rust
+```rust,ignore
```
fiveop: Out of curiosity: what does the `ignore` 'language' achieve? A cursory search didn't bring up anything useful.

posborne: I didn't add it yet, but was planning on adding support for [skeptic](https://github.com/brson/rust-skeptic)

fiveop:
Looks good, except for that typo! :) I think the benefits outweigh the drawbacks of switching to github pages.

posborne:
> Looks good, except for that typo! :) I think the benefits outweigh the drawbacks of switching to github pages.

Cool, I'll fix that up.  Was traveling yesterday.

fiveop:
@homu r+

kamalmarhubi:
Big :+1: for this!

Issue #363, opened 2016-04-28
Release procedure and policy


kamalmarhubi:
As far as I know, only @carllerche can make releases. We need to figure out a procedure that works for the active maintainers.

We should also come up with a release schedule. There have been over 100 commits since the last release, which was less than two months ago, so we're probably averaging ~10 commits per week. 

@fiveop proposed every other week or so in https://github.com/nix-rust/nix/issues/290#issuecomment-195381212. I'm in favour of a fairly quick cycle to make sure changes are routinely getting onto crates.io.

Depending on how careful we are or want to be, we can mostly make 0.0.1 bumps except on API changes. Or we can just always make 0.1.0 bumps. I don't have strong feelings.

posborne:
I'm in favor of a faster release cycle, particularly when users report issues and work to get a PR merged.  In those cases, the reporter has probably already been inconvenienced by bumping into the issue and it is nice to provide a release in a timely fashion.  Doing releases more frequently reduce the overall impact of cutting a release (smaller deltas).

I think we should try to follow semantic versioning and let that dictate how we do version bumps.

Adding the nix core team to the crate but leaving @carllerche as the named owner (although I would nominate @kamalmarhubi as a potential co-named owner) seems to make sense to me so long as there is a documented release process in place that we can all follow (PR with changelog prior to release perhaps approved by another core team member).  There is pretty good documentation on team ownership here: http://doc.crates.io/crates-io.html#cargo-owner

carllerche:
Happy to do whatever, let me know

posborne:
@nix-rust/nix-maintainers Any other thoughts?  I'm currently in favor of granting co-named ownership to @kamalmarhubi and team ownership to @nix-rust/nix-maintainers.  Any objections?

kamalmarhubi:
I have no thoughts other than let's be able to make releases! @carllerche can you run at least the following command:

```
cargo owner --add github:nix-rust:maintainers
```
?

posborne:
Agreed.  We need to cut a release.  @carllerche I aggree with @kamalmarhubi that getting the maintainers access should be enough to get the ball rolling.

kamalmarhubi:
@carllerche gave it a shot, but was not having luck with GitHub permissions. I opened up the nix-rust org but I don't think he had a chance to try it again. This is a crates.io usability pain point for sure.

kamalmarhubi:
[...]

PR #387, opened 2016-07-12
Use git master as libc version

bugaevc:
As mentioned by @fiveop in #363,

> We should use the Github version of libc for development and a fixed version for releases.

So this changes the required libc version to GitHub's master.

The dependency is supposed to be manually changed to a fixed version when doing a release, out of the GitHub repository (i.e. nix's master always tracks libc's master).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


posborne:
Looks good to me.  Cargo will not allow things to be pushed to crates.io with the dependency not pointing to a version also on crates.  @homu r+

PR #391, opened 2016-07-22
Introduce CHANGELOG.md

fiveop:
- Added initial CHANGELOG.md with Changes from Version 0.6.0 and the
  upcoming changes for version 0.7.0.
- Added RELEASE_PROCEDURE.md detailing what changes should be made
  to CHANGELOG.md during a release.
- Added sections about CHANGELOG.md in CONVENTIONS.md and
  CONTRIBUTIONS.md.

PR #392, opened 2016-08-03
Replace ffi module by libc functions in mqueue.rs

fiveop:
This is almost finished. I still need to check if I introduced any breaking changes by changing signatures. I would want to record this in the change log, however, for that we still need to merge #391.

- [x] update change log
- [x] run rustfmt on `src/mqueue.rs`

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -4,7 +4,7 @@
 
 use {Errno, Result};
 
-use libc::{self, c_char,c_long,mode_t,mqd_t,size_t};
+use libc::{self, c_char, c_long, mode_t, mqd_t, size_t};
 use std::ffi::CString;
 use sys::stat::Mode;
 use std::mem;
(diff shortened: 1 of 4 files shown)


src/mqueue.rs
```diff
@@ -4,142 +4,126 @@
...
+
+bitflags!{
+    flags MQ_OFlag: libc::c_int {
+        const O_RDONLY    = 0o00000000,
+        const O_WRONLY    = 0o00000001,
+        const O_RDWR      = 0o00000002,
+        const O_CREAT     = 0o00000100,
+        const O_EXCL      = 0o00000200,
+        const O_NONBLOCK  = 0o00004000,
+        const O_CLOEXEC   = 0o02000000,
```
posborne: Might be better to use the libc constants as the values in the enum?

src/mqueue.rs
```diff
@@ -4,142 +4,126 @@
...
+    }
 }
 
+impl MqAttr {
+    pub fn new(mq_flags: c_long, mq_maxmsg: c_long, mq_msgsize: c_long, mq_curmsgs: c_long) -> MqAttr {
+        let mut attr = unsafe { mem::uninitialized::<libc::mq_attr>() };
+        attr.mq_flags = mq_flags;
+        attr.mq_maxmsg = mq_maxmsg;
+        attr.mq_msgsize = mq_msgsize;
+        attr.mq_curmsgs = mq_curmsgs;
```
posborne: Is there any reason to not use the struct initializer syntax in this case rather than creating a mutable `mq_attr` and setting every field?

posborne:
Changes generally look good.  Probably run rustfmt on the file while your making changes (a few lines are getting a bit long and its nice to have consistent formatting).

src/mqueue.rs
fiveop: Yes, it would be better. I will change that.

It would be even better to use `OFlag` from `fcntl.rs`. But `Oflag` has more flags than are allowed for the mqueue functions. I suppose, the other flags would just be ignored if passed into the functions. And just to be safe we could warn about their usage in mqueue functions. I have not made my mind up yet, and this PR is just for getting rid of the `ffi` module. There is much more that could be done to clean up this module.

src/mqueue.rs
fiveop: There is a private field for padding.

fiveop:
I'll have a look at rustfmt. I never used it before. (Maybe something for the conventions or contribution guidelines?)

fiveop:
I think this can be merged now.

fiveop:
Given that no further comments were made, I'll merge this tomorrow.

fiveop:
@homu r=@fiveop

PR #394, opened 2016-08-04
Use specific version 0.3.4 of tempdir

fiveop:
Version 0.3.5 is broken for Rust 1.1.


posborne:
Specifying 0.3.4 in this way does not prevent 0.3.5 from being selected (as I'm sure CI will show).

fiveop:
Rust 1.2 compiles. I do not mind getting rid of 1.1 as well.

fiveop:
I retract this PR in support of #395.

PR #395, opened 2016-08-05
No longer support Rust 1.1

fiveop:
Since tempdir is broken on Rust 1.1 I propose to push the minimum supported Rust version to 1.2.


posborne:
@homu r+

PR #396, opened 2016-08-06
Fix nix on FreeBSD

asomers:


asomers:
Grr, the tests failed at runtime on OSX.  The problem isn't obvious from the console output and I don't have an OSX machine to develop on.  Could somebody who does take a look at the changes?  Something funny is going on, because by inspection sendmsg/recvmsg never should've worked on 64-bit OSX in the first place.

src/sys/socket/ffi.rs
```diff
@@ -37,7 +37,7 @@ pub struct cmsghdr {
     pub cmsg_len: type_of_cmsg_len,
     pub cmsg_level: c_int,
     pub cmsg_type: c_int,
-    pub cmsg_data: [type_of_cmsg_len; 0]
+    pub cmsg_data: [usize; 0]
```
fiveop: Instead of using a standard rust type here and thus affecting (though probably not breaking) other platforms, you should fix the type definition in line 14 using the appropriate type alias from libc.

src/sys/socket/mod.rs
```diff
@@ -201,12 +200,8 @@ pub enum ControlMessage<'a> {
...
 
 fn cmsg_align(len: usize) -> usize {
-    let round_to = mem::size_of::<type_of_cmsg_len>();
-    if len % round_to == 0 {
-        len
-    } else {
-        len + round_to - (len % round_to)
-    }
+    let align_bytes = mem::size_of::<usize>() - 1;
+    (len + align_bytes) & !align_bytes
```
fiveop: The fix is fine, but regarding `usize` (see above)

fiveop:
I left my comments in the code. Thank you for the changes and fixes. Regarding the failing test, I have the same problem as you.

posborne:
> Regarding the failing test, I have the same problem as you.

I have a machine running OSX and can take a look.

posborne:
Also, CC @geofft who initially added support.

src/sys/socket/ffi.rs
```diff
@@ -37,7 +37,7 @@ pub struct cmsghdr {
     pub cmsg_len: type_of_cmsg_len,
     pub cmsg_level: c_int,
     pub cmsg_type: c_int,
-    pub cmsg_data: [usize; 0]
+    pub cmsg_data: [size_t; 0]
```
fiveop: Is this correct for all platforms? Or similarly, was using `type_of_cmsg_len` _totally_ wrong?

asomers: `type_of_cmsg_len` for `cmsg_data` was totally wrong.  If you look at `CMSG_DATA` on Linux, it basically says that the data field begins at a location suitable for an aligned `struct cmsghdr`.  On FreeBSD, `CMSG_DATA` basically says that the data field should begin at a location suitably aligned for a `register_t`, which is either a `__int32_t` or an `__int64_t` depending on the platform.

posborne:
I started looking at fixing this for OSX and it looks like our best bet for fixing is probably to rely a bit more heavily on the (presumably correct) ffi definitions for things like `msghdr` that are present in libc (if those are wrong, fixes should definitely be pushed upstream).  Some platform type mangling will still be required to get the casts right.

Basically, I think most everything in `sys::socket::ffi` should disappear.

asomers:
I'll be getting access to a MAC later this evening, and I'll try out your suggestion.

posborne:
Changes look good to me.  We can probably get rid of some code (leverage libc further) in the future, but I don't want to gait this PR on that action.

Would you mind rebasing and squashing a few of the fixup commits?  Things have gotten a little bit messy at this point.

Once that is done, r=me.

asomers:
Sure.  But what does "r=me" mean?

posborne:
Sorry, that is just code to other core team members that if they get around to looking at this first and that action has been taking, they may tell the @homu bot that it has been reviewed by me (and the bot will then merge the commit after doing some final checks).

fiveop:
I cleaned up the PR in #397 and merged it. Thank you.

fiveop:
One more thing @asomers: could you update CHANGELOG.md regarding new features, breaking changes, and fixes?

PR #397, opened 2016-08-10
Fix nix on FreeBSD (rebased on master)

fiveop:


fiveop:
@posborne could you have a look at the build failures. Given that the failing builds, fail due to some problems with docker images you created, you probably know what the problem is.

posborne:
> @posborne could you have a look at the build failures. Given that the failing builds, fail due to some problems with docker images you created, you probably know what the problem is.

Strange.  I'll try to figure out what is going on.  Thanks for the rebase.

posborne:
Update: Looks like there was a change to remove the `/.dockerinit` file in newer version of docker which is why this is failing.  I'll prepare a PR that fixes that.  https://github.com/docker/docker/issues/18355

fiveop:
@homu r=@posborne

posborne:
Thanks @asomers @fiveop

posborne:
As an FYI, I'm taking a fresh look at some of the docker stuff to see if I can come up with some kind of circus act in order to get some BSD builds into the mix.

PR #398, opened 2016-08-11
ci: fix docker container check

posborne:
Newer version of docker no longer have the /.dockerinit file present
when the container is executing, so that is no longer an option.
When executing in a container, we do know that we will be executing
as pid 1 -- this is probably not ever the case on the host system.

r? @fiveop

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -6,7 +6,7 @@
 set -e
 
 # This should only be run in a docker container, so verify that
-if [ ! -f /.dockerinit ]; then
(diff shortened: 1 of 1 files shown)


fiveop:
Given that the correct build (pr) is green, it would probably be alright to merge this. Let's see whether homu let's me. Why did the push build trigger?

@homu r+

posborne:
> Why did the push build trigger?

I accidentally pushed my branch to the wrong remote (this repo) at first.

PR #228, opened 2015-12-31
Add everything from poll.h.

utkarshkukreti:
(Only OSX and Linux support.)

PR #399, opened 2016-08-11
Use libc in poll.rs

fiveop:
I'll add a change log commit, once this is reviewed.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,74 +1,48 @@
-use libc::c_int;
+use libc;
 use {Errno, Result};
 
-pub use self::ffi::PollFd;
-pub use self::ffi::consts::*;
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct PollFd {
+    pollfd: libc::pollfd,
+}
 
-mod ffi {
-    use libc::c_int;
-    pub use self::consts::*;
-
-    #[derive(Clone, Copy, Debug)]
-    #[repr(C)]
-    pub struct PollFd {
-        pub fd: c_int,
-        pub events: EventFlags,
-        pub revents: EventFlags
-    }
-
-    #[cfg(target_os = "linux")]
-    pub mod consts {
-        use libc::{c_short, c_ulong};
-
-        bitflags! {
-            flags EventFlags: c_short {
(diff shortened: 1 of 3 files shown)


src/poll.rs
```diff
@@ -1,74 +1,48 @@
...
-                const POLLRDBAND = 0x0080,
-                const POLLWRBAND = 0x0100,
-                const POLLERR    = 0x0008,
-                const POLLHUP    = 0x0010,
-                const POLLNVAL   = 0x0020,
-            }
-        }
-
-        pub type nfds_t = c_uint;
+    pub fn revents(&self) -> Option<EventFlags> {
```
utkarshkukreti: We'll also need at least read/write access to `events` and possibly write access to `revents`. This looks a little less ergonomic than direct access to the fields we have right now :( Maybe we can keep the struct definition as is? Are there any reasons to switch other than libc having a better test suite to verify the structure?

fiveop: I was wondering about that, but I could only come up with read access to `revents` as necessary after creation. So, thank you for the feedback.
You text sounds like you have used poll somewhere. Can you point me code that uses it?

utkarshkukreti: I added these bindings to nix in #228 as I planned to use it in a project but I never got around to it since then. A [quick code search](https://github.com/search?utf8=%E2%9C%93&q=%22nix%3A%3Apoll%22+language%3Arust&type=Code&ref=searchresults) on Github shows atleast 2 people are using it: https://github.com/tailhook/stator and https://github.com/hjr3/carp-rs; they might have some useful feedback :)

fiveop: Both example only need read access to `revents` after creation.

fiveop:
I would also like to merge this soon, if noone has any further comments. I'll be happy to add read/write accessors to `PollFd`s member fields, when they are needed.

posborne:
Hi @fiveop, changes look good to me.  Based on my own usage of poll in the past, it would be very useful to be able to modify poll interests.  For instance, with non-blocking sockets, it is useful to register for `POLLOUT` to get notifications of when a socket connects and then unregister that interest later.

I think we can add that later without issues, so @homu r+

PR #401, opened 2016-08-15
Get rid of a few warnings during compilation of tests

fiveop:


posborne:
Tests were failing due to an issue with grabbing 1.7.0 that was transient.  Reran and things passed.  Changes look good @humu r+

Issue #360, opened 2016-04-19
clone() is broken and unsafe with CLONE_VM flag


arcnmx:
`clone()` has a few issues when used with `CLONE_VM`, unless it is used together with `CLONE_VFORK`.

1. it takes a `Box<FnMut>` by value, which will be moved, dropped, and destroyed when the function returns back to the caller.
    - essentially means it will unconditionally segfault when passed `CLONE_VM`
2. Even if it did work, the mutable borrow of the stack will end once the `clone()` call returns, even if the cloned process is still running, meaning you can modify the child process' stack as it runs.

I was actually going to suggest that its prototype be changed to an unboxed `FnOnce`, since the caller's stack will never fall out from under you unless `CLONE_VM` is used... but I think it also needs to disallow the `CLONE_VM` flag entirely (unless `CLONE_VM | CLONE_VFORK`). Perhaps have a second unsafe version for that use.

bugaevc:
`clone()` also [doesn't impose any limitations](https://nix-rust.github.io/nix/nix/sched/type.CloneCb.html) on the callback -- it can be `!Send`, it can reference variables that go out of scope as soon as the `clone()` call is made... basically, `clone()` enables every unsafety Rust was built to prevent. This is a _very_ serious soundness hole; for a lack of a better solution I propose to make `clone()` `unsafe` ASAP.

Susurrus:
The point of `nix` is to provide safe wrappers around a lot of these primitives. The final suggestion from @arcnmx seems like it's a safe version, even if it has reduced functionality. I think we should offer that use case instead and punt on a complete but safe version for now, unless someone objects to that.

Susurrus:
Coming back to this as I'd like to fix up some bugs, I think having both an unsafe and safe-but-restricted clone makes a lot of sense here. @arcnmx it seems like you have an understanding of what the API should be, would you be able to whip up a PR for this?

@bugaevc Is there anything we should do to the existing `clone()` function besides make it unsafe? I'd also propose a rename to `clone_unsafe` or something along those lines so that the safe version can be `clone()` and it will encourage people to use the safe one over the unsafe one.

bugaevc:
Yes, we could make it accept the same kind of closures `thread::spawn` accepts: `Send + 'static`. That itself would be safe, and if anyone wants to use a non-`Send` or non-`'static` closure they can do the `unsafe` dance on top of that (e.g. by casting any reference to and from a `*const u8`).

[...]

PR #402, opened 2016-08-15
Replace parts of ffi module by libc functions ind sched.rs

fiveop:
Not exporting `CpuMask` anymore, should be the only API change of this PR. Since the type is not used in any other exported signature or type, I think it is fine to no longer export it.

r? @kamalmarhubi @posborne

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -1,204 +1,110 @@
 use std::mem;
 use std::os::unix::io::RawFd;
 use std::option::Option;
-use libc::{self, c_int, c_void, c_ulong, pid_t};
-use {Errno, Result};
+use libc::{self, c_int, c_void, pid_t};
+use {Errno, Error, Result};
 
 // For some functions taking with a parameter of type CloneFlags,
 // only a subset of these flags have an effect.
-bitflags!{
-    flags CloneFlags: c_int {
-        const CLONE_VM             = libc::CLONE_VM,
-        const CLONE_FS             = libc::CLONE_FS,
-        const CLONE_FILES          = libc::CLONE_FILES,
-        const CLONE_SIGHAND        = libc::CLONE_SIGHAND,
-        const CLONE_PTRACE         = libc::CLONE_PTRACE,
-        const CLONE_VFORK          = libc::CLONE_VFORK,
-        const CLONE_PARENT         = libc::CLONE_PARENT,
-        const CLONE_THREAD         = libc::CLONE_THREAD,
-        const CLONE_NEWNS          = libc::CLONE_NEWNS,
-        const CLONE_SYSVSEM        = libc::CLONE_SYSVSEM,
-        const CLONE_SETTLS         = libc::CLONE_SETTLS,
-        const CLONE_PARENT_SETTID  = libc::CLONE_PARENT_SETTID,
-        const CLONE_CHILD_CLEARTID = libc::CLONE_CHILD_CLEARTID,
-        const CLONE_DETACHED       = libc::CLONE_DETACHED,
-        const CLONE_UNTRACED       = libc::CLONE_UNTRACED,
-        const CLONE_CHILD_SETTID   = libc::CLONE_CHILD_SETTID,
-        // TODO: Once, we use a version containing
-        // https://github.com/rust-lang-nursery/libc/pull/147
-        // get rid of the casts.
-        const CLONE_NEWUTS         = libc::CLONE_NEWUTS as c_int,
-        const CLONE_NEWIPC         = libc::CLONE_NEWIPC as c_int,
-        const CLONE_NEWUSER        = libc::CLONE_NEWUSER as c_int,
-        const CLONE_NEWPID         = libc::CLONE_NEWPID as c_int,
-        const CLONE_NEWNET         = libc::CLONE_NEWNET as c_int,
-        const CLONE_IO             = libc::CLONE_IO as c_int,
-    }
-}
-
-// Support a maximum CPU set of 1024 nodes
-#[cfg(all(target_arch = "x86_64", target_os = "linux"))]
-mod cpuset_attribs {
-    use super::CpuMask;
-    pub const CPU_SETSIZE:           usize = 1024;
-    pub const CPU_MASK_BITS:         usize = 64;
-
-    #[inline]
-    pub fn set_cpu_mask_flag(cur: CpuMask, bit: usize) -> CpuMask {
-        cur | (1u64 << bit)
-    }
-
-    #[inline]
-    pub fn clear_cpu_mask_flag(cur: CpuMask, bit: usize) -> CpuMask {
-        cur & !(1u64 << bit)
-    }
-}
-
-#[cfg(all(target_arch = "x86", target_os = "linux"))]
-mod cpuset_attribs {
-    use super::CpuMask;
-    pub const CPU_SETSIZE:           usize = 1024;
-    pub const CPU_MASK_BITS:         usize = 32;
-
-    #[inline]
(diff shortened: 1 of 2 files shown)


src/sched.rs
```diff
@@ -1,204 +1,127 @@
...
-        pub fn setns(fd: c_int, nstype: c_int) -> c_int;
-
-        // Set the current CPU set that a task is allowed to run on
-        pub fn sched_setaffinity(__pid: pid_t, __cpusetsize: size_t, __cpuset: *const CpuSet) -> c_int;
+        pub fn clone(cb: *const CloneCb,
+                     child_stack: *mut c_void,
+                     flags: c_int,
+                     arg: *mut super::CloneCb,
+                     ...)
+                     -> c_int;
```
posborne: Can this be replaced with the [libc clone](https://doc.rust-lang.org/beta/libc/fn.clone.html)?

src/sched.rs
```diff
@@ -1,204 +1,127 @@
...
-        cur | (1u32 << bit)
-    }
-
-    #[inline]
-    pub fn clear_cpu_mask_flag(cur: CpuMask, bit: usize) -> CpuMask {
-        cur & !(1u32 << bit)
-    }
+#[inline]
+fn clear_cpu_mask_flag(cur: CpuMask, bit: usize) -> CpuMask {
+    cur & !(cpuset_attribs::FLAG_BIT << bit)
```
posborne: I'm not sure these helpers are really necessary given the pretty singular purpose of the code at the call site and the fact that these are only called once.

posborne:
Overall looks good.  Interested to hear feedback on the comments  but neither of those seem gating.  Your commit message has a typo ("ind") that you might as well amend to fix.

src/sched.rs
fiveop: You are right, I have not thought of that.

src/sched.rs
fiveop: Probably, but understanding what is happning and how `clone` is supposed to be used, was to much for me yesterday evening, so given that it does not share any types or code with the other functions in that file, I left it for another commit.

posborne:
Test failures look legitimate (typo).  Once that is fixed, looks good r=me.

src/sched.rs
```diff
@@ -1,204 +1,116 @@
...
 // A single CPU mask word
-pub type CpuMask = c_ulong;
+type CpuMask = c_ulong;
 
 // Structure representing the CPU set to apply
 #[repr(C)]
 #[derive(Clone, Copy)]
 pub struct CpuSet {
-    cpu_mask: [CpuMask; cpuset_attribs::CPU_SETSIZE/cpuset_attribs::CPU_MASK_BITS]
+    cpu_mask: [CpuMask; CPU_SETSIZE as usize / cpuset_attribs::CPU_MASK_BITS],
```
kamalmarhubi: would it make more sense to wrap `libc::cpu_set_t` here?

src/sched.rs
```diff
@@ -1,204 +1,116 @@
...
-        }
+        CpuSet { cpu_mask: unsafe { mem::zeroed() } }
     }
 
     pub fn set(&mut self, field: usize) {
         let word = field / cpuset_attribs::CPU_MASK_BITS;
         let bit = field % cpuset_attribs::CPU_MASK_BITS;
 
-        self.cpu_mask[word] = cpuset_attribs::set_cpu_mask_flag(self.cpu_mask[word], bit);
+        self.cpu_mask[word] = self.cpu_mask[word] | (cpuset_attribs::FLAG_BIT << bit);
```
kamalmarhubi: `libc::CPU_SET`?

src/sched.rs
```diff
@@ -1,204 +1,116 @@
...
-    cpu_mask: [CpuMask; cpuset_attribs::CPU_SETSIZE/cpuset_attribs::CPU_MASK_BITS]
+    cpu_mask: [CpuMask; CPU_SETSIZE as usize / cpuset_attribs::CPU_MASK_BITS],
 }
 
 impl CpuSet {
     pub fn new() -> CpuSet {
-        CpuSet {
-            cpu_mask: unsafe { mem::zeroed() }
-        }
+        CpuSet { cpu_mask: unsafe { mem::zeroed() } }
```
kamalmarhubi: `libc::CPU_ZERO`?

src/sched.rs
```diff
@@ -1,204 +1,116 @@
...
-        self.cpu_mask[word] = cpuset_attribs::set_cpu_mask_flag(self.cpu_mask[word], bit);
+        self.cpu_mask[word] = self.cpu_mask[word] | (cpuset_attribs::FLAG_BIT << bit);
     }
 
     pub fn unset(&mut self, field: usize) {
         let word = field / cpuset_attribs::CPU_MASK_BITS;
         let bit = field % cpuset_attribs::CPU_MASK_BITS;
 
-        self.cpu_mask[word] = cpuset_attribs::clear_cpu_mask_flag(self.cpu_mask[word], bit);
+        self.cpu_mask[word] = self.cpu_mask_[word] & !(cpuset_attribs::FLAG_BIT << bit);
```
kamalmarhubi: we'd have to add it to libc, but `CPU_CLR`

src/sched.rs
kamalmarhubi: Yeah there's some fanciness going on here to allow passing a Rust closure as the function to run in the new process. There's also #360 which is concerning.

kamalmarhubi:
Few more comments, mostly to do with using even more libc :-)

src/sched.rs
fiveop: Since the field in `cpu_set_t` is private, it would complicate the set and get methods below needlessly.

src/sched.rs
fiveop: Is there an Android version?

(What kind of functions are these? Are they implementations of C macros?)

src/sched.rs
fiveop: Yes.

Though, I will first prepare a patch to add this and `CPU_SET` for the android platform.

src/sched.rs
fiveop: I'll add it and use it.

fiveop:
Thank you for your comments. #392 is also awaiting a merge (or more comments).

src/sched.rs
fiveop: When we use the`CPU_` functions (macros), the previous comment becomes obsolete.

fiveop:
I changed the PR based on @kamalmarhubi's comments. However, I did not use `CPU_ZERO`, because just `mem::zeroed` is nicer than `mem::uninitialized` + `CPU_ZERO`.

Maybe someone can come up with a macro, that captures the three very similar functions. I could not get around implementing it in at least two rules in order to differentiate between `&self` and `&self mut`. And two cases for three functions is not worth it.

What also should be looked at are the return values. Since we want to provide safe interfaces, I felt that such a check was necessary.

fiveop:
Given that @posborne was fine with the changes and I addressed (almost) all of @kamalmarhubi's concerns and in the intrest of getting forward with this project, I'll take @posborne's invitation:

@homu r=@posborne

How about a release in the next few weeks? @kamalmarhubi could write up how it is done in the RELEASE_PROCEDURE.md, and then I could do it in the future in regular intervals.

posborne:
> How about a release in the next few weeks? @kamalmarhubi could write up how it is done in the RELEASE_PROCEDURE.md, and then I could do it in the future in regular intervals.

Absolutely, it may make sense to push for it earlier.  I saw some requests for it on nix and I think @carllerche said he was planning to cut a mio release soon.

PR #403, opened 2016-08-21
Fix kqueue-related definitions on *BSD

asomers:
The kqueue-related definitions contained errors for all non-OSX operating
systems.  I split them up into separate sections for macos, freebsd, openbsd,
netbsd and dragonfly, and I removed NOTE_SIGNAL from everything besides OSX
because it's only defined within the kernel.

### DIFF
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -64,7 +64,7 @@ mod ffi {
     }
 }
 
-#[cfg(not(any(target_os = "dragonfly", target_os = "netbsd")))]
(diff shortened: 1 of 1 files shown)


asomers:
Hold off on merging.  I just discovered that the website I was using as a reference is out-of-date.  I'll amend this PR shortly.

posborne:
Thanks @asomers, I have been looking at a few different appraoches for adding *BSD to our test matrix (using QEMU) but haven't gotten around to actually implementing anything as of yet.

PR #404, opened 2016-08-21
Fix kqueue-related definitions on *BSD

asomers:
The kqueue-related definitions contained errors for all non-OSX operating
systems.  I split them up into separate sections for macos, freebsd, openbsd,
netbsd and dragonfly, and I removed NOTE_SIGNAL from everything besides OSX
because it's only defined within the kernel.

### DIFF
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -64,7 +64,7 @@ mod ffi {
     }
 }
 
-#[cfg(not(any(target_os = "dragonfly", target_os = "netbsd")))]
+#[cfg(target_os = "macos")]
 #[repr(i16)]
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum EventFilter {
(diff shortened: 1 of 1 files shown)


src/sys/event.rs
```diff
@@ -96,6 +96,27 @@ pub enum EventFilter {
...
     EVFILT_USER = -9,
+    EVFILT_FS = -10,
+    EVFILT_SYSCOUNT = 10,
+}
+
+#[cfg(target_os = "freebsd")]
+#[repr(i16)] // u_short
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum EventFilter {
+    EVFILT_READ = -1,
```
posborne: Can we use the constants from libc instead (adding as needed).  They have better tests for such things there and it is the way we have been moving things.

asomers:
Yeah, we can do that.  Would you like me to move all of the structure definitions to libc as well?

posborne:
> Yeah, we can do that. Would you like me to move all of the structure definitions to libc as well?

Awesome.  Yeah, I think the structures are probably something that would be good to contribute to libc as well.

asomers:
About to open a new PR that moves most definitions into libc.

PR #405, opened 2016-08-25
Implement `Clone` for `FdSet` on Mac/iOS

murarth:


murarth:
`Clone` is implemented for `FdSet` on all other platforms, excluding Mac/iOS. Seems to be an oversight.

fiveop:
Thank you.

@homu r+

PR #407, opened 2016-08-26
Add FcntlArg::F_FULLFSYNC

untitaker:
https://developer.apple.com/library/ios/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -46,6 +46,8 @@ pub enum FcntlArg<'a> {
     F_ADD_SEALS(SealFlag),
     #[cfg(target_os = "linux")]
     F_GET_SEALS,
+    #[cfg(any(target_os = "macos", target_os = "ios"))]
+    F_FULLFSYNC,
 
     // TODO: Rest of flags
 }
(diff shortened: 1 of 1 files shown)


src/fcntl.rs
```diff
@@ -69,6 +71,8 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
             F_ADD_SEALS(flag) => libc::fcntl(fd, ffi::F_ADD_SEALS, flag.bits()),
             #[cfg(target_os = "linux")]
             F_GET_SEALS => libc::fcntl(fd, ffi::F_GET_SEALS),
+            #[cfg(target_os = "darwin")]
+            F_FULLFSYNC => libc::fcntl(fd, ffi::F_FULLFSYNC),
```
fiveop: Darwin is not a valid `target_os`. You have to use either `macos`, `ios` or both, depending on where that flag exists. This is also, why this line does not cause our apple build to fail. Otherwise it would complain about an unkown identifier `ffi::F_FULLFSYNC` (see the `ffi` module definition further up in the file).

untitaker: Sorry, I was confused because of https://github.com/untitaker/nix/blob/69c5d686ae5c64ea758ceaab482e5a3789397ee2/nix-test/build.rs#L10

untitaker:
I don't have Mac OS X, so I'll test on Travis. I don't know if iOS supports that flag, but [this thread](https://lists.apple.com/archives/darwin-userlevel/2014/Apr/msg00000.html) seems to imply so.

untitaker:
Okay, the tests seem to pass now!

fiveop:
Thank you.

@homu r+

PR #415, opened 2016-09-05
Move kqueue-related definitions from nix to libc

asomers:
See https://github.com/rust-lang/libc/pull/379 for the companion commit to libc.  If and when that commit is merged, I'll update nix's Cargo.toml file and add it to this PR.


fiveop:
Our master branch requires libc's github master branch. So you don't need to change our Cargo.toml.

src/sys/event.rs
```diff
@@ -3,278 +3,237 @@
...
-        #[cfg(target_os = "netbsd")]
-        pub fn kevent(
-            kq: c_int,
-            changelist: *const kevent,
-            nchanges: size_t,
-            eventlist: *mut kevent,
-            nevents: size_t,
-            timeout: *const timespec) -> c_int;
-    }
+// Redefine kevent in terms of programmer-friendly enums and bitfields.
```
kamalmarhubi: I think it would be preferable here to just have an inner `libc::kevent` struct. Then we don't have to duplicate the definition, and can get rid of the test that they match. This would require some setters though, for the enum / flags parts. I guess I should look at some kqueue code to see how it's used first. Never done any.

kamalmarhubi: @fiveop any thoughts here?

test/sys/test_event.rs
```diff
@@ -0,0 +1,30 @@
+use nix::sys::event;
+use libc;
+use std::mem::{size_of, size_of_val};
+use std::ptr;
+
+#[test]
+/// Check that nix::event::KEvent has the same layout as libc::kevent
+/// This would be easier with offsetof operators.
```
kamalmarhubi: if we do go this route, then the test should set distinct non-zero values in the fields and also check for equality.

src/sys/event.rs
asomers: I tried that first.  But I found it very inconvenient to require setters in my application code.  It's much more convenient just to `let ke = KEvent{whatever}`

test/sys/test_event.rs
asomers: Good idea.

src/sys/event.rs
fiveop: The rest of nix just wraps libc's structs and provides constructor fuctions:

```rust
pub struct NixStruct {
   struct: libc::struct
}
impl NixStruct {
   pub fn new(all the fields as parameters) {
     // do the dirty convertion here
   }
}
```

I like it and would like to keep it consistent.

test/sys/test_event.rs
fiveop: I really don't like this.

src/sys/event.rs
kamalmarhubi: One issue here is that there are 6 args. I think it works ok for things with 2 maybe 3. With 6 it becomes hard to know what they are from reading.

asomers: @fiveop what about testing fields?  It's easy to do
```
let k = KEvent{...}
...
match k.filter {
    nix.event.EVFILT_AIO => something
    _ => something else
}
```
but that becomes harder if the filter field is no longer an enum.

fiveop: With an accessor you still have an enum. See for example `EpollEvent::events()`.

asomers: @fiveop that's only because EpollEvent and EpollEventKind are both defined within Nix.  But kevent is defined within libc, and libc doesn't do enums, only constants.  That's why I redefined KEvent within nix.

fiveop: I'm sorry. I forgot that my change to EpollEvent is not merged yet. You can see, how it can look like at https://github.com/nix-rust/nix/pull/410/files.

asomers: Ok @fiveop I think I see what you want.  I'll give it a shot.

asomers:
@fiveop is this more like what you had in mind?

src/sys/event.rs
```diff
@@ -3,280 +3,255 @@
...
 
-#[cfg(not(any(target_os = "dragonfly", target_os = "netbsd")))]
+#[cfg(any(target_os = "macos",
+          target_os = "freebsd",
+          target_os = "dragonfly"))]
+type type_of_event_flag = u16;
+#[cfg(any(target_os = "macos",
+          target_os = "freebsd",
+          target_os = "dragonfly"))]
 bitflags!(
```
fiveop: I think you can now use the `libc_bitflags` macro, that allows to reduce repetition. You can find examples elsewhere in nix.

asomers:
All three build failures seem unrelated to this PR.

kamalmarhubi:
@asomers thanks!

@homu r+

asomers:
The failure is due to the gcc dependency not working with Rust 1.2.  It is unrelated to this PR.

fiveop:
I would have prefered to have waited until we get the build fully working again. A rebase instead of a merge from master would have also been nicer.

kamalmarhubi:
Sorry about that.

PR #416, opened 2016-09-06
add unistd::getcwd and unistd::mkdir

philippkeller:
As a (late) followup of [this withdrawn PR](https://github.com/rust-lang/libc/pull/326) I have added getcwd (wrapper around `libc::getcwd`) and mkdir (wrapper around `libc::mkdir`) and added testing.

A few notes:

 - I'm new to rust so I would appreciate some pair of eyes testing the code, plus I'm open for revision of code or general remarks about my coding style
 - I have run the tests both on OSX as on Linux (Ubuntu)
 - I've run `clippy` to see if my code is well formatted, however clippy issues many warnings about the project. I think I didn't add any more warnings
 - the methods in unistd are not documented so I also left out the documentation of `getcwd` and `mkdir`, although I think it'd probably be good to add some documentation, especially some example code how to use the methods
 - the base idea of `getcwd` is [taken from std](https://github.com/rust-lang/rust/blob/1.9.0/src/libstd/sys/unix/os.rs#L95-L119), should I mention that somewhere?

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -5,9 +5,10 @@ use fcntl::{fcntl, OFlag, O_NONBLOCK, O_CLOEXEC, FD_CLOEXEC};
 use fcntl::FcntlArg::{F_SETFD, F_SETFL};
 use libc::{self, c_char, c_void, c_int, c_uint, size_t, pid_t, off_t, uid_t, gid_t};
 use std::mem;
-use std::ffi::CString;
+use std::ffi::{CString,CStr};
 use std::os::unix::io::RawFd;
 use void::Void;
+use std::path::PathBuf;
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub use self::linux::*;
@@ -109,6 +110,45 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     }));
 
     Errno::result(res).map(drop)
+}
+
+// #[inline]
+// pub fn mkdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
+//     Errno::result(0)
+// }
+
+#[inline]
+pub fn getcwd() -> Result<PathBuf> {
+    let mut buf = Vec::with_capacity(512);
+    loop {
+        unsafe {
+            let ptr = buf.as_mut_ptr() as *mut libc::c_char;
+
+            // The buffer must be large enough to store the absolute pathname plus
+            // a terminating null byte, or else null is returned.
+            // To safely handle this we start with a reasonable size (512 bytes)
+            // and double the buffer size upon every error
+            if !libc::getcwd(ptr, buf.capacity()).is_null() {
+                let len = CStr::from_ptr(ptr).to_bytes().len();
+                buf.set_len(len);
+                buf.shrink_to_fit();
+                let s = try!(CString::new(buf).map_err(|_| Error::Sys(Errno::EILSEQ)));
+                let s = try!(s.into_string().map_err(|_| Error::Sys(Errno::EILSEQ)));
+                return Ok(PathBuf::from(&s));
+            } else {
+                let error = Errno::last();
+                if error == Errno::ERANGE {
+                    return Err(Error::Sys(error));
+                }                
+            }
+
+            // Trigger the internal buffer resizing logic of `Vec` by requiring
+            // more space than the current capacity.
+            let cap = buf.capacity();
+            buf.set_len(cap);
+            buf.reserve(1);
+        }
+    }    
 }
 
 #[inline]
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -3,12 +3,13 @@
 use {Errno, Error, Result, NixPath};
 use fcntl::{fcntl, OFlag, O_NONBLOCK, O_CLOEXEC, FD_CLOEXEC};
 use fcntl::FcntlArg::{F_SETFD, F_SETFL};
-use libc::{self, c_char, c_void, c_int, c_uint, size_t, pid_t, off_t, uid_t, gid_t};
+use libc::{self, c_char, c_void, c_int, c_uint, size_t, pid_t, off_t, uid_t, gid_t, mode_t};
 use std::mem;
 use std::ffi::{CString,CStr};
+use std::path::PathBuf;
 use std::os::unix::io::RawFd;
 use void::Void;
-use std::path::PathBuf;
+use sys::stat::Mode;
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub use self::linux::*;
@@ -112,10 +113,14 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-// #[inline]
-// pub fn mkdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
-//     Errno::result(0)
-// }
+#[inline]
+pub fn mkdir<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
+    let res = try!(path.with_nix_path(|cstr| {
+        unsafe { libc::mkdir(cstr.as_ptr(), mode.bits() as mode_t) }
+    }));
+
+    Errno::result(res).map(drop)
+}
 
 #[inline]
 pub fn getcwd() -> Result<PathBuf> {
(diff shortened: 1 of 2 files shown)


philippkeller:
oh, failing tests for Rust 1.2, I'll check the errors in a few hours

posborne:
> oh, failing tests for Rust 1.2, I'll check the errors in a few hours

Looks like the errors should be pretty easy to resolve.  At some point, I think we're OK with dropping 1.2 but in this case, it should be pretty easy to satisfy the compiler.

philippkeller:
@posborne indeed, I'm almost there. I ended up with an exact copy of the implementation in std (which I somehow was not able to make running but found out that I needed to add `use std::os::unix::ffi::OsStringExt` to bring the unix specific trait `from_vec` into scope)

src/unistd.rs
```diff
@@ -112,6 +114,50 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
...
+                buf.shrink_to_fit();
+                let s = try!(CString::new(buf).map_err(|_| Error::Sys(Errno::EILSEQ)));
+                let s = try!(s.into_string().map_err(|_| Error::Sys(Errno::EILSEQ)));
+                return Ok(PathBuf::from(&s));
+            } else {
+                let error = Errno::last();
+                // ERANGE means buffer was too small to store directory name
+                if error != Errno::ERANGE {
+                    return Err(Error::Sys(error));
+                }
```
posborne: Trailing whitespace.

philippkeller: Thanks for the hint. I've corrected those now. Any idea why `cargo clippy` did not warn about those?

test/test_unistd.rs
```diff
@@ -120,6 +123,31 @@ macro_rules! execve_test_factory(
...
 
 #[test]
+fn test_getcwd() {
+  // workaround for the fact that on os x TmpDir::new returns /var/folders/... but upon
+  // chdir into that directory getcwd returns /private/var/folders/...
+  let base = if cfg!(target_os = "macos") {
+    "/private/tmp/"
+  } else {
+    "/tmp/"
+  };
```
posborne: This seems a little fragile.  I think it would be OK to just verify that the nix `getcwd` results match those of libstd without hardcoding absolute base paths which may change based on the OS and system (My tmpfs may not be at `/tmp`).

posborne:
> the methods in unistd are not documented so I also left out the documentation of getcwd and mkdir, although I think it'd probably be good to add some documentation, especially some example code how to use the methods

I would like to see the documentation of nix improve.  We would welcome documentation (including examples) for new code.

test/test_unistd.rs
philippkeller: good idea, I'll change this and add documentation for `getcwd` and `mkdir`

philippkeller:
@posborne I fixed the test and added documentation for `getcwd` and `mkdir`. Can you have a look again?

src/unistd.rs
```diff
@@ -3,11 +3,14 @@
 use {Errno, Error, Result, NixPath};
 use fcntl::{fcntl, OFlag, O_NONBLOCK, O_CLOEXEC, FD_CLOEXEC};
 use fcntl::FcntlArg::{F_SETFD, F_SETFL};
-use libc::{self, c_char, c_void, c_int, c_uint, size_t, pid_t, off_t, uid_t, gid_t};
+use libc::{self, c_char, c_void, c_int, c_uint, size_t, pid_t, off_t, uid_t, gid_t, mode_t};
 use std::mem;
-use std::ffi::CString;
+use std::ffi::{CString,CStr,OsString};
```
posborne: nit: spaces after comma.

src/unistd.rs
```diff
@@ -111,6 +114,92 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
...
+/// use nix::sys::stat;
+/// use tempdir::TempDir;
+///
+/// fn main() {
+///     let mut tmp_dir = TempDir::new("test_mkdir").unwrap().into_path();
+///     tmp_dir.push("new_dir");
+///
+///     // create new directory and give read, write and execute rights to the owner
+///     match unistd::mkdir(&tmp_dir, stat::S_IRWXU) {
+///        Ok(_) => println!("created {:?}", tmp_dir.display()),
```
posborne: Explicitly calling `display` here should not be required.

src/unistd.rs
```diff
@@ -111,6 +114,92 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Creates new directory `path` with access rights `mode`.
+///
+/// # Errors
+///
+/// `Err` is returned in case of an error. There are several situations where mkdir might fail.
+/// For a full list consult `man mkdir(2)`
```
posborne: I think I would prefer a link to the man pages from man7.org, in this case http://man7.org/linux/man-pages/man2/mkdir.2.html

src/unistd.rs
```diff
@@ -111,6 +114,92 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Creates new directory `path` with access rights `mode`.
+///
+/// # Errors
+///
+/// `Err` is returned in case of an error. There are several situations where mkdir might fail.
```
posborne: First sentence does not need to be stated, drop it.

src/unistd.rs
```diff
@@ -111,6 +114,92 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
...
+///
+/// Err is returned if the current user doesn't have the permission to read or search a component of the current path.
+///
+/// # Example
+///
+/// ```rust
+/// use nix::unistd;
+///
+/// let dir = unistd::getcwd().expect("not allowed to get current directory");
+/// println!("The current directory is {:?}", dir.display());
```
posborne: Ditto on `display`.  If you want display in the formst string you should use `{}` not `{:?}` which uses the `Debug` trait.

src/unistd.rs
```diff
@@ -111,6 +114,92 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
...
+}
+
+/// Returns the current directory as a PathBuf
+///
+/// Err is returned if the current user doesn't have the permission to read or search a component of the current path.
+///
+/// # Example
+///
+/// ```rust
+/// use nix::unistd;
```
posborne: The previous example included `extern crate` declarations and this one doesn't.  I think they are not required in general for `$crate` (nix here).

src/unistd.rs
philippkeller: resolved

src/unistd.rs
philippkeller: resolved, thanks for the hint

src/unistd.rs
philippkeller: resolved, also added anchor to errors

src/unistd.rs
philippkeller: you're right. Resolved

src/unistd.rs
philippkeller: resolved

src/unistd.rs
philippkeller: No, this is required. The [doc says](https://doc.rust-lang.org/book/documentation.html#documentation-as-tests): 

> If the example does not contain extern crate, then extern crate <mycrate>; is inserted 

Since I need to do a `extern crate tempdir` I also need to do an extern crate nix here.

I think you want to hint that all examples should have the same style, that's why I now changed the doc of `getcwd` to have the same syntax as `mkdir` (including extern crate nix and an explicit main function)

test/test_unistd.rs
```diff
@@ -120,6 +123,24 @@ macro_rules! execve_test_factory(
...
+  // make path 500 chars longer so that buffer doubling in getcwd kicks in.
+  // Note: One path cannot be longer than 255 bytes (NAME_MAX)
+  // whole path cannot be longer than PATH_MAX (usually 4096 on linux, 1024 on macos)
+  for _ in 0..5 {
+    let newdir = iter::repeat("a").take(100).collect::<String>();
+    tmp_dir.push(newdir);
+    assert!(mkdir(tmp_dir.as_path(), stat::S_IRWXU).is_ok());
+  }
+  assert!(chdir(tmp_dir.as_path()).is_ok());
+  assert_eq!(getcwd().unwrap(), current_dir().unwrap());
```
posborne: I just noticed that this code is indented with 2-spaces for some reason.  Should be 4.

fiveop: Running `rustfmt` as a last step of preparing a PR should be added to our CONTRIBUTING.md.

posborne:
@homu r+

philippkeller:
Yay, thanks a lot @posborne for the helpful comments and the merge. I'm feeling proud - that was my first bigger PR into a rust project

fiveop:
Thank you!

Could you add entries to the CHANGELOG.md detailing your changes? Preferably before we cut the release.

PR #418, opened 2016-09-08
Expand on release procedure

kamalmarhubi:


posborne:
Looks good.  @homu r+

fiveop:
You need to update the libc dependency in cargo.toml before (to a concrete version) and after release (back to the git repository for development builds).

PR #425, opened 2016-09-09
Small fixes to RELEASE_PROCEDURE.md

fiveop:
I decided against adding a line for a local test of the commit, because we have the full build in the CI and you might miss something locally anyway.

### DIFF
--- a/RELEASE_PROCEDURE.md
+++ b/RELEASE_PROCEDURE.md
@@ -19,9 +19,9 @@ The release is prepared as follows:
   - In `README.md`, update the version in the Usage section to the new
     version.
 - Make a pull request.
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Haha, a difference of style opinion! I don't mind how this goes. Adding `origin` is probably a good idea though.

fiveop:
I only found references 'eg' in lists of 'internet slang' or 'internet abbreviations'. If I remember correctly, I once read in an english style guide/english grammar, that there is no reason not to write 'for example' instead of 'e.g.', the same holding for 'i.e.' and 'that is' :)

kamalmarhubi:
UK style is typically not to put periods. Eg, Guardian style guide: http://www.theguardian.com/guardian-observer-style-guide-a

Anyway, @homu r+

Issue #414, opened 2016-09-05
pipe2 doesn't set flags atomically?


oconnor663:
It looks like the `pipe2` function is implemented as a call to regular `pipe` followed by separate calls to `fcntl` (for compatibility with systems that don't support `pipe2`?). I think that's fine for O_NONBLOCK, but not for O_CLOEXEC. If a multithreaded program forks in between the creation of a pipe and when O_CLOEXEC is set, the child will inherit pipes that it isn't supposed to.

Is it possible to call the real `pipe2` libc function where it's supported, and to fall back to `pipe` only when necessary? I think this is what libstd does.

posborne:
> Is it possible to call the real pipe2 libc function where it's supported, and to fall back to pipe only when necessary? I think this is what libstd does.

I have no objection to this.  Mind opening up a PR?

oconnor663:
I'd love to put it together, but I could use your advice about the right way to do it. Rust's [libstd uses](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/pipe.rs#L33) some kind of [weak binding](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/weak.rs#L78) backed by function called `dlsym`. I'm don't know much about what that function's doing under the covers, but if you think that's the right way to go I'll read more about it. Are there any examples in nix already of doing this kind of capability detection?

posborne:
Personally, at least for the first pass, I am OK only supporting linux 2.6.27+ in nix as long as we make a note that further work would be required to support older kernels.  Alternatively, borrowing that macro/logic from libstd might also be an acceptable solution.

`dlsym` in this context is just seeing if the libc in used contains specific symbols at runtime.

oconnor663:
Interesting, it looks like musl already implements a similar fallback: http://git.musl-libc.org/cgit/musl/tree/src/unistd/pipe2.c

fiveop:
If I understand it, the situation cannot be improved, from where your PR got us. Can we close this now?

oconnor663:
Yep. Closing.

Does `libc` have any plans to support pipe2 on BSD platforms? What's the broader future of detecting this sort of thing dynamically?

PR #427, opened 2016-09-14
call pipe2 directly on Linux

oconnor663:
A first shot at fixing https://github.com/nix-rust/nix/issues/414. This approach keeps the old implementation for platforms other than `notbsd`, because `libc` only exposes `pipe2` in the `notbsd` module.

I've tested this by hand on my Linux machine in a couple ways:

- Create a toy program that opens a pipe and passes it to `cat`, which hags if `O_CLOEXEC` doesn't get set properly. Confirm that it doesn't hang, but that it does if I pass `0` as the flags to `libc::pipe2`.
- Delete the new implementation entirely, and delete the `cfg` guards on the old implementation, and confirm that above is still true.

I haven't actually tested a cross compilation build though. Is there a standard way to do that?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - `pipe2` now calls `libc::pipe2` where available. Previously it was emulated
   using `pipe`, which meant that setting `O_CLOEXEC` was not atomic.
+  ([#427](https://github.com/nix-rust/nix/pull/427))
 
 ## [0.7.0] 2016-09-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- `pipe2` now calls `libc::pipe2` where available. Previously it was emulated
+  using `pipe`, which meant that setting `O_CLOEXEC` was not atomic.
+
 ## [0.7.0] 2016-09-09
 
 ### Added
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -365,32 +365,28 @@ pub fn pipe() -> Result<(RawFd, RawFd)> {
           target_os = "android",
           target_os = "emscripten"))]
 pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
-    unsafe {
-        let mut fds: [c_int; 2] = mem::uninitialized();
+    let mut fds: [c_int; 2] = unsafe { mem::uninitialized() };
 
-        let res = libc::pipe2(fds.as_mut_ptr(), flags.bits());
+    let res = unsafe { libc::pipe2(fds.as_mut_ptr(), flags.bits()) };
 
-        try!(Errno::result(res));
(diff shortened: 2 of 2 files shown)


oconnor663:
It looks like support for this used to exist, but was removed in https://github.com/nix-rust/nix/commit/9e935330dd252d4ec640197a6040df2aaeb583c9. Am I repeating the mistakes of the past?

src/unistd.rs
```diff
@@ -374,7 +393,13 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
     }
 }
 
+#[cfg(not(any(target_os = "linux",
+              target_os = "android",
+              target_os = "emscripten")))]
 fn pipe2_setflags(fd1: RawFd, fd2: RawFd, flags: OFlag) -> Result<()> {
+    use fcntl::O_NONBLOCK;
```
fiveop: Why do you pull these `use` statements down from the file header?

src/unistd.rs
```diff
@@ -360,6 +360,25 @@ pub fn pipe() -> Result<(RawFd, RawFd)> {
     }
 }
 
+// libc only defines `pipe2` in `libc::notbsd`.
+#[cfg(any(target_os = "linux",
+          target_os = "android",
+          target_os = "emscripten"))]
+pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
+    unsafe {
```
fiveop: Could you use `unsafe` as finegrained as possible, even if you need more instances.

src/unistd.rs
oconnor663: Since this is the only function that uses these imports, and it's no longer building on Linux, we get "unused import" warnings if we leave these at the top. Is there another good way to avoid those?

src/unistd.rs
oconnor663: Done in b28005f. I notice GitHub's fancy new review system doesn't automatically collapse comments in the same way though.

src/unistd.rs
fiveop: Makes sense.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- `pipe2` now calls `libc::pipe2` where available. Previously it was emulated
+  using `pipe`, which meant that setting `O_CLOEXEC` was not atomic.
+
```
fiveop: `([#427](https://github.com/nix-rust/nix/pull/427))`

oconnor663:
Anything else I should do on my end to make this good to go?

fiveop:
@homu r+

It's perfect now. Thank you.

kamalmarhubi:
@oconnor663 I just came across the fact that the released version doesn't do this. I'll see if we can release this week to get it out there. And tthanks for making this change!

oconnor663:
@kamalmarhubi BSD support has also landed in `libc` master, though it hasn't been released there yet. We can expand this as soon as they release.

kamalmarhubi:
@oconnor663 oh cool! They are usually happy to release when asked. I'm trying to figure out if there are more things we want before asking. Otherwise we ought to be able to expand it quite soon indeed! :-)

PR #428, opened 2016-09-16
Add Mkstemp (fixed for rust 1.2)

philippkeller:
I fixed @antifuchs addition of `mkstmp` in https://github.com/nix-rust/nix/pull/365 by making it compile in Rust 1.2 and adding documentation.

A few remarks:

- made it working with Rust 1.2 which needed `to_bytes_with_nul()`. I think the implementation is memory safe but would like to have a pair of eyes checking that
- replaced Path by PathBuf so it's more in line with getcwd
- I didn't move it into another module. If this still the consensus then I would like to do that but in a separate PR (probably moving all stdio methods out)
- it's true that unistd doesn't need mkstmp since there is the crate tempdir and tempfile, but I'd love to see this here for completeness


philippkeller:
ping to @kamalmarhubi @arcnmx: could you check if that implementation is safe memory wise? I guess the question is if `OsStr::from_bytes(CStr::from_ptr(p).to_bytes())` copies the bytes away from the pointer `p` into an owned string.

arcnmx:
Yeah no that string is dropped by the time you use `PathBuf::from`, probably best to just move path_copy out of the closure instead.

src/unistd.rs
```diff
@@ -511,6 +511,40 @@ pub fn sleep(seconds: libc::c_uint) -> c_uint {
...
+/// ```
+#[inline]
+pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
+    let mut path = try!(template.with_nix_path(|path| {path.to_bytes_with_nul().to_owned()}));
+    let p = path.as_mut_ptr() as *mut _;
+    unsafe {
+        let fd = libc::mkstemp(p);
+        let pathname = OsStr::from_bytes(CStr::from_ptr(p).to_bytes());
+        try!(Errno::result(fd));
+        Ok((fd, PathBuf::from(pathname).to_owned()))
```
arcnmx: `PathBuf::from` already creates an owned PathBuf, so `to_owned()` is just making an additional copy. Also, it's unnecessary since `path` is already owned in the first place; you can just strip out the trailing nul byte and use `OsString::from_vec`

philippkeller: thanks, removed the double copying and used `OsString::from_vec` to avoid the constructing of the string from pointer. Nice side effects: less includes and less code in the `unsafe` block. Is there a nicer way to not have to use `.pop()`?

philippkeller: thanks for the tip, I added the `debug_assert`

philippkeller:
@fiveop I guess this is now ready to be merged, the code is now safe, shorter and easier to understand thanks to the input of @arcnmx

fiveop:
Looks good to me. Thank you and @antifuchs.

@homu r+

PR #430, opened 2016-09-17
Removes SIGSTKFLT when cross-compiling to MIPS.

tcr:
This is the only failing error when cross compiling to Tessel, which is a MIPS-based architecture. 

Apparently libc does not have SIGSTKFLT (Term Stack fault on coprocessor) on this system, and indeed it does not have a coprocessor.

This commit should preserve the semantics for all other architectures. Potentially this can be inverted in the future if more platforms are added to only *whitelist* ARM, x86, x86_64, but I thought this conditional would suffice.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -28,7 +28,7 @@ pub enum Signal {
     SIGPIPE = libc::SIGPIPE,
     SIGALRM = libc::SIGALRM,
     SIGTERM = libc::SIGTERM,
-    #[cfg(any(target_os = "linux", target_os = "android", target_os = "emscripten"))]
+    #[cfg(all(any(target_os = "linux", target_os = "android", target_os = "emscripten"), not(target_arch = "mips")))]
     SIGSTKFLT = libc::SIGSTKFLT,
     SIGCHLD = libc::SIGCHLD,
     SIGCONT = libc::SIGCONT,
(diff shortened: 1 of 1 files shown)


posborne:
@fiveop Looks like this condition is not possible on MIPS as an architecture, so this change seems to make sense.  https://www.linux-mips.org/archives/linux-mips/2006-05/msg00345.html

posborne:
I wish there was a better way to manage this (and there might be through macro magic), but I'm good with this change today in order to fix mips @homu r+

I still need to revisit getting the CI tests for MIPS to pass.  Until then, I would not say we officially support MIPS.  I would like to, of course.

tcr:
Thanks for finding the reference for this @posborne. If I can help with fixing the MIPS CI, lmk what I should look at.

posborne:
> If I can help with fixing the MIPS CI, lmk what I should look at.

Sure, that task is basically getting  CI passing.  The infrastructure built on top of docker is already present and you can run CI for mips by doing the following (also read https://github.com/nix-rust/nix/tree/master/ci):

```
 $ DOCKER_IMAGE=posborne/rust-cross:mips \
      RUST_VERSION=1.7.0 \
      RUST_TARGET=mips-unknown-linux-gnu ci/run-docker.sh
```

PR #431, opened 2016-09-27
Bump bitflags to 0.7

nox:


nox:
Somehow had forgotten the actual bump... This is fixed now and Travis should be happy.

src/macros.rs
```diff
@@ -150,7 +150,7 @@ macro_rules! libc_bitflags {
     // (non-pub) Entry rule.
     (
         $(#[$attr:meta])*
-        flags $BitFlags:ident: $T:ty {
+        pub flags $BitFlags:ident: $T:ty {
```
fiveop: I think this is a (or maybe even _the_) problem. The macro `libc_bitflags` already differentiated between pub and non-pub bitflags (as the comments indicate).

nox: The macro `libc_bitflags` was broken because `bitflags 0.4` did not support non-public flags.

nox: Oh wait, misunderstood what you mean. Will fix this.

nox:
1.2.0 and 1.7.0 are not supported by bitflags anymore, can we stop supporting them here? No idea what these beta linking failures are though.

fiveop:
I would not mind dropping support, see #432. I also don't know what the problem on the i686 builds for beta  (and nighlty) is.

posborne:
@nox We have dropped support for versions prior to 1.7.0 but are still depending on 1.7.0 for the time being.  I would consider dropping 1.7.0 but for now CI depends on it for testing non-host architectures.

fiveop:
I also opend a bitflags-issue asking whether the drop in support was intentional, though, I have not got an answer, yet.

nox:
1.7 is like an eon old, could we just merge this?

posborne:
> 1.7 is like an eon old, could we just merge this?

No, we cannot simply break the current oldest supported version.  To merge, we will need to do the additional work to increase the minimum version (including CI updates).

nox:
Yeah of course, that's obvious. Can it be done?

posborne:
> Yeah of course, that's obvious. Can it be done?

Yeah, I think we are open to it.  The larger chunk of work that needs to be bitten off is further improving the CI infrastructure to allow moving between versions more easily, probably based on https://github.com/japaric/trust.  Some discussion here: https://github.com/rust-embedded/rfcs/issues/16

dimbleby:
Recent `bitflags` doesn't generate dead code warnings.  So when this eventually gets merged, can remove [this code](https://github.com/nix-rust/nix/blob/ff4484d343dbd8f224d767b4499cc502c4c7f016/src/lib.rs#L8-L10).  At which point you'll find that it has been masking things:

```
warning: constant item is never used: `WSTOPPED`, #[warn(dead_code)] on by default
  --> src/sys/wait.rs:40:1
   |
40 | const WSTOPPED: WaitPidFlag = WUNTRACED;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

nox:
Can we land this?

posborne:
@nox I had forgotten about this and ended up merging #515 which was basically the same change submitted by @SimonSapin.  Closing this but thanks for the contribution (and sorry for missing this when going back through the backlog).

PR #438, opened 2016-10-16
wait: Support ptrace events for Linux

chaosagent:
Adds new WaitStatus value `PtraceEvent`. Implementation of #273 that only affects Linux/Android.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -191,7 +191,7 @@ fn decode(pid : pid_t, status: i32) -> WaitStatus {
     } else if status::signaled(status) {
         WaitStatus::Signaled(pid, status::term_signal(status), status::dumped_core(status))
     } else if status::stopped(status) {
-        if cfg!(any(target_os = "linux", target_os = "android")) {
+        #[cfg(any(target_os = "linux", target_os = "android"))] {
             let status_additional = status::stop_additional(status);
             if status_additional != 0 {
                 return WaitStatus::PtraceEvent(pid, status::stop_signal(status), status::stop_additional(status))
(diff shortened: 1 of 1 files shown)


chaosagent:
This works on newer versions of Rust; not really sure how to cleanly do a cfg on a block in previous versions.

src/sys/wait.rs
```diff
@@ -184,6 +191,12 @@ fn decode(pid : pid_t, status: i32) -> WaitStatus {
     } else if status::signaled(status) {
         WaitStatus::Signaled(pid, status::term_signal(status), status::dumped_core(status))
     } else if status::stopped(status) {
+        #[cfg(any(target_os = "linux", target_os = "android"))] {
```
kamalmarhubi: I didn't know `cfg` on blocks was possible, even if nightly only.

To do this on stable, I think your best approach is going to be adding an inner function like this:

```rust
cfg_if! {
    if #[cfg(any(target_os = "linux", target_os = "android"))] {
        fn decode_stopped(status) -> WaitStatus {
            // Linux code
        }
    } else {
        fn decode_stopped(status) -> WaitStatus {
            // Other platform code
        }

        decode_stopped(status)
    }
}
```

posborne: https://github.com/alexcrichton/cfg-if

kamalmarhubi: Oh yeah, should have pointed out where that comes from. We already have it in nix, so it's not a risk to use it.

kamalmarhubi:
@luser @abbradar could you take a look at this? It looks reasonable to me, but I haven't ptraced things much.

kamalmarhubi:
Thanks @chaosagent! I'm going to try and get someone who's used these APIs to see if they make sense. Should be able to merge soon though. :-)

chaosagent:
I've found [https://github.com/codius/codius-sandbox-core/blob/master/src/waitpid.rs](this) implementation of this, which works pretty much the same, if that's of any help.

abbradar:
On a quick glance this seems OK to me. I can't remember if it's possible that non-signal event still has ptrace information. Anyway, that's better than my old solution!

kamalmarhubi:
@chaosagent I'm really sorry for the delay on this PR. I'm going to assume that nothing has changed since November in terms of correctness.

@homu r+

Issue #435, opened 2016-10-07
Build failure in i686 beta


fiveop:
In recent builds for i686 and rust beta, there have been build failures do to linking problems. See for example:

https://travis-ci.org/nix-rust/nix/builds/165124003
https://travis-ci.org/nix-rust/nix/builds/164470698
https://travis-ci.org/nix-rust/nix/builds/164422919

fiveop:
I tried to recreate the issue on my machine by installing beta-i686-linux-unknown-gnu and running the build. However, the build runs through. This makes investigating the failure difficult.

@posborne, as the (usually) knowledgeable when it comes to builds, do you have any ideas?

posborne:
Hmm, interesting.  Looks like it is failing to link basically everything provided by `std`.  I am suspicious of whether something might have changed with TravisCI but will need to do some more digging to say anything for certain.

fiveop:
For reference: The build has the same problem, if I use the same scripts and environment as libc: https://travis-ci.org/fiveop/nix-rust/jobs/167885171

fiveop:
The build problem occurs since beta went from 1.12.0-beta.6 to 1.13.0-beta.1. The prerelease version was bumped less than a day ago (https://github.com/rust-lang/rust/commit/3ebfbdb3e50bf2d5d9bfe2bfaad6ece86067cd46), so I am suggest waiting for 1.13.0-beta.2, since someone mentioned that there were linker problems in beta and nightly.

TimNN:
This is probably rust-lang/rust#36793, I think.

(This has been fixed by rust-lang/rust#36866 on nightly, however the fix does not seem to have been include in 1.13.0-beta.2)

fiveop:
I suggest ignoring the beta build failure until 1.13.0-beta.3 (or newer) is released, so that work can continue.

fiveop:
The build features do not occur anymore, so we can revert the changes.

PR #440, opened 2016-10-20
Temporarily allow beta failures till 1.13.0-beta.3 or later is released

fiveop:
@kamalmarhubi @posborne, is that alright with you? I would let the Issue #435 open, till this is reverted, once we have a working beta again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,8 @@ matrix:
       sudo: true
   allow_failures:
     - rust: nightly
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Looks like we're failing on 1.2 as of the gcc crate's latest release. Filed https://github.com/alexcrichton/gcc-rs/issues/111 and opening a PR to bring back 1.2 support.

kamalmarhubi:
r=me with comment added :-)

fiveop:
I'll bypass homu, because the build will probably still fail.

PR #443, opened 2016-10-24
use tempfile in test_pwrite

asomers:
Use tempfile in test_pwrite so as not to pollute the source directory.


posborne:
@homu r+

Issue #432, opened 2016-10-03
Drop support for anything but stable, beta, (nightly)


fiveop:
libc's CI only tests against stable, beta and nighlty (and a single build seems to run against 1.0.0 for some reason). Since we are a wrapper for libc, I would like to drop support for anything but stable, beta (and working towards getting nighlty green).

If others are not comfortable with this, I would suggest to altenatively only support the n latest releases (in addition to beta and nightly), where n has to be clearly defined.

fiveop:
Closed in favour of #444.

PR #444, opened 2016-10-24
ci: drop official support for rustc < 1.7.0

posborne:
Supporting older versions of rust is causing CI to fail and is
a somewhat constant support burden with questionable value.  1.7.0
is the oldest I have seen that people have had a requirement on
supporting in recent history.

It may make sense to work toward a more official policy on
what version of rust we will support in the future.  Users of
older version of rust are, of course, welcome to use older versions
of nix and everything else in the ecosystem.

r? @kamalmarhubi @fiveop

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,7 @@ addons:
       - binutils-dev
 
 rust:
-  - 1.2.0  # Oldest supported version
-  - 1.7.0
-  - 1.8.0
+  - 1.7.0  # Oldest supported version
   - stable
   - beta
   - nightly
@@ -48,12 +46,6 @@ matrix:
     - os: osx
       env: ARCH=i686
       rust: stable
-    - os: osx
-      env: ARCH=x86_64
-      rust: 1.2.0
-    - os: osx
-      env: ARCH=i686
-      rust: 1.2.0
     # Docker builds for other targets
     - os: linux
       env: TARGET=aarch64-unknown-linux-gnu DOCKER_IMAGE=posborne/rust-cross:arm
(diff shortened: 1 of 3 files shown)


kamalmarhubi:
I'm ok with this so long as we mention it in the changelog and make it a "major" bump to 0.8.

kamalmarhubi:
(Seeing how this review business works)

fiveop:
As my issue #432 indicates, I'm all for it. Given changes are made as suggested by @kamalmarhubi r=me (or him I suppose :)).

posborne:
> As my issue #432 indicates, I'm all for it. Given changes are made as suggested by @kamalmarhubi r=me (or him I suppose :)).

Ah, missed that.  I'm still a big fan of having a minimum supported version that is tested by CI, but I agree that we don't need to support ancient versions.  Even 1.7 is older than I think we need to support, but that is where CI is at present.  More discussion on that over at rust-embedded/rfcs: https://github.com/rust-embedded/rfcs/issues/16

> Please add a note in the changelong, and bump the version to 0.8.0-pre.

Will do.

posborne:
@homu r=kamalmarhubi

posborne:
As a note, I have now also enabled nightly builds via https://nightli.es/ which I have been using on other projects for some time.  This will give us a better indication of when things may be broken on a new stable/beta/nightly release without having a contributor bump into it first.

kamalmarhubi:
@posborne does nightlies have a dashboard?

posborne:
> @posborne does nightlies have a dashboard?

They have a page called the dashboard when logged in.  It basically just shows rust projects under your account and whether nightli.es has setup webhooks or not.  Code is here: https://github.com/coderanger/nightlies.  It is quite barebones but will get us notifications via travis when things start failing.

kamalmarhubi:
Thought of an additional benefit bumping tro 1.7 is that we get `AsRef<CStr>` impls (https://github.com/rust-lang/rust/pull/30616). This will make the solution to #230 nicer.

PR #445, opened 2016-10-24
Make signal argument to kill optional

### LINKED ISSUE #441: Not possible to call kill(pid, 0)
According to the man page:

_If sig is 0, then no signal is sent, but error checking is still performed; this can be used to check for the existence of  a process ID or process group ID._

The current signature is `pub fn kill(pid: pid_t, signal: Signal)`, and I think it should use `Option<Signal>`.

fiveop:
Fixes #441

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -387,8 +387,12 @@ pub fn pthread_sigmask(how: SigFlags,
     Errno::result(res).map(drop)
 }
 
-pub fn kill(pid: libc::pid_t, signal: Signal) -> Result<()> {
-    let res = unsafe { libc::kill(pid, signal as libc::c_int) };
+pub fn kill<T: Into<Option<Signal>>>(pid: libc::pid_t, signal: T) -> Result<()> {
+    let res = unsafe { libc::kill(pid,
+                                  match signal.into() {
+                                      Some(s) => s as libc::c_int,
+                                      None => 0,
+                                  }) };
 
     Errno::result(res).map(drop)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#457](https://github.com/nix-rust/nix/pull/457))
 
 ### Changed
+- `kill`'s signature, defined in `::nix::sys::signal`, changed, so that the
+  signal parameter has type `T: Into<Option<Signal>>`. `None` as an argument
+  for that parameter will result in a 0 passed to libc's `kill`, while a
+  `Some`-argument will result in the previous behavior for the contained
+  `Signal`.
+  ([#445](https://github.com/nix-rust/nix/pull/410))
 - The minimum supported version of rustc is now 1.7.0.
   ([#444](https://github.com/nix-rust/nix/pull/444))
 - Implement `Send` for `KEvent`
--- a/test/sys/mod.rs
+++ b/test/sys/mod.rs
@@ -1,3 +1,4 @@
+mod test_signal;
 mod test_socket;
 mod test_sockopt;
 mod test_termios;
--- a/test/sys/test_signal.rs
+++ b/test/sys/test_signal.rs
@@ -1,0 +1,7 @@
+use nix::unistd::*;
+use nix::sys::signal::*;
+
+#[test]
+fn test_kill_none() {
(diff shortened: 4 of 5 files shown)


posborne:
I feel like a better approach might be to add a new enumerated value rather than making the signal an option.  A signal of 0 is still a signal but one that is just handled by the kernel without actually making it to the process.

I'm not sure this signal has an established name (it doesn't appear to in Linux: http://lxr.free-electrons.com/source/arch/arm/include/asm/signal.h?v=2.6.32#L31).  I suggest that we just call it `SIGZERO`.

As an added benefit, this won't break things for users using the existing API.

fiveop:
I disagree. 0 is no signal, it represents something different. Therefore, it does not belong in the enumeration. I would not have a problem with giving it it's own name, if I not already objected to it belonging to the operation.

Since we break the API left and right at the moment, in order to get things 'nice', I do not have a problem with breaking this. It is easy to adapt code that is affected by the change.

posborne:
Although 0 is not implemented as a signal to processes in most operating systems, I am still failing to see a strong reason to separate it.  Strace, for instance, goes so far as to name the `0` `SIG_0` in its output.

    kill(9999, SIG_0)                       = -1 ESRCH (No such process)

posborne:
Thinking about this more, I can be OK with the change (better than not having support for this useful operation).  Moving forward, we should probably have a push for doing more documentation on our functions, particularly when there are cases like this where things might not be immediately clear from the signature.  At present, I would need to read the source (and probably go deep on the man pages) in order to figure out what passing `None` to this function does.

With some docs added, r=me

fiveop:
I'll add a documentation comment.

src/sys/signal.rs
```diff
@@ -388,8 +388,12 @@ pub fn pthread_sigmask(how: SigFlags,
     Errno::result(res).map(drop)
 }
 
-pub fn kill(pid: libc::pid_t, signal: Signal) -> Result<()> {
-    let res = unsafe { libc::kill(pid, signal as libc::c_int) };
+pub fn kill(pid: libc::pid_t, signal: Option<Signal>) -> Result<()> {
```
kamalmarhubi: maybe we can take advantage of the [recently added](https://github.com/rust-lang/rust/pull/34828) `impl From<T> for Option<T>`. On recent enough Rust (1.12+), `kill(pid, SIGBLAH)` will still work, so it's a bit more ergonomic.

posborne: Interesting, I wasn't aware of that change.  Nice!

fiveop: What about us supporting Rust (1.7+), though? If someones crate wants to continue to support 1.7+, it has to change its source (which I don't have a problem with in the first place).

So asked differently: Are you saying, that we will passively take advantage of the new impl on new enough Rust versions anyway?

kamalmarhubi: Yeah, that was my thought. Though perhaps we should wait until we bump minimum version to 1.12, probably in 3-6 months? Not sure how to think about it.

kamalmarhubi: Trying to think about it differently: the change from `Signal` to `Option<Signal>` is breaking. The change from `Option<Signal>` to `T: Into<Option<Signal>>` is (I think?) not breaking thanks to `impl<T> From<T> for T`. (This depends on how you read "breaking" in signatures.) Our use of the `Into<Option<T>>` impl would be more akin to "progressive enhancement". A downstream crate requiring its use would have a 1.12+ minimum version, but we would not.

kamalmarhubi: In any case, if making use of the new impl, we should include tests that demonstrate it works in both cases.

fiveop: Sounds reasonable. And if I understand you and the matter correctly, then we could already use `T: Into<Option<Signal>>` right now.

kamalmarhubi: That's my belief!

fiveop:
@kamalmarhubi better?

kamalmarhubi:
Could you add a test to verify that calling it with `None` works? I want to be sure there's no weird inference stuff that requires us to specify `Option::<Signal>::None` or something like that. This could even be a `no_run` doc test, I think.

With that, r=me.

kamalmarhubi:
Also I checked and under 1.12 it allows `kill(pid, SIGKILL)`, ie no `Some`. So that's cool :-)

fiveop:
@homu r=@kamalmarhubi

PR #448, opened 2016-10-25
Add setresuid and setresgid

dgreid:
These were both recently added to libc, add wrappers.

Signed-off-by: Dylan Reid <dgreid@chromium.org>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added function `epoll_create1` and bitflags `EpollCreateFlags` in
   `::nix::sys::epoll` in order to support `::libc::epoll_create1`.
   ([#410](https://github.com/nix-rust/nix/pull/410))
+- Added `setresuid` and `setresgid` for Linux in `::nix::unistd`
+  ([#448](https://github.com/nix-rust/nix/pull/448))
 
 ### Changed
 - The minimum supported version of rustc is now 1.7.0.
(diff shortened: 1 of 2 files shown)


posborne:
Looks good.  A couple of minor things:

1. Go ahead and add an entry to [the CHANGELOG.md](https://github.com/nix-rust/nix/blob/master/CHANGELOG.md) with a note on the additions
2. Although most functions are not documented at present, I think we are moving in that direction.  Go ahead and add documentation for the functions which includes a summary description and link to the relevant online man pages (on man7.org)

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added function `epoll_create1` and bitflags `EpollCreateFlags` in
   `::nix::sys::epoll` in order to support `::libc::epoll_create1`.
   ([#410](https://github.com/nix-rust/nix/pull/410))
+- Added setresuid and setresgid for Linux
```
fiveop: Could you please markup the identifiers as code and reference the module they live in (`::nix::unistd`).

src/unistd.rs
```diff
@@ -587,6 +588,38 @@ mod linux {
         Errno::result(res).map(drop)
     }
 
+    /// Sets the real, effective, and saved uid
```
posborne: This is a real nit, but when rendered I think this will juxtapose the first line and "Further Reading", so it probably makes sense to add a period after the first line or do something like `([see setresuid(2)][...])`

posborne:
Thanks for making the updates!

posborne:
Looks good.  @homu r+

fiveop:
Thanks

PR #449, opened 2016-10-28
Documentation improvements

posborne:
This series of commits starts to fill in documentation that was not previously present for various calls, mostly in unistd for this pass.

r? @kamalmarhubi

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,8 +77,16 @@ use std::fmt;
 use std::error;
 use libc::PATH_MAX;
 
+/// Nix Result Type
 pub type Result<T> = result::Result<T, Error>;
 
+/// Nix Error Type
+///
+/// The nix error type provides a common way of dealing with
+/// various system system/libc calls that might fail.  Each
+/// error has a corresponding errno (usually the one from the
+/// underlying OS) to which it can be mapped in addition to
+/// implementing other common traits.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum Error {
     Sys(errno::Errno),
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -77,27 +77,40 @@ use std::fmt;
...
+/// Nix Result Type
 pub type Result<T> = result::Result<T, Error>;
 
+/// Nix Error Type
+///
+/// The nix error type provides a common way of dealing with
+/// various system system/libc calls that might fail.  Each
+/// error has a corresponding errno (usually the one from the
+/// underlying OS) to which it can be mapped in addition to
+/// implementing other common traits.
```
kamalmarhubi: should we doc `InvalidPath` or just punt on it in hopes we actually kill it soon?

src/unistd.rs
```diff
@@ -32,57 +37,144 @@ impl ForkResult {
...
+///
+/// After calling the fork system call (successfully) two processes will
+/// be created that are identical with the exception of their pid and the
+/// return value of this function.  As an example:
+///
+/// ```no_run
+/// use nix::unistd::{fork, ForkResult};
+///
+/// match fork() {
+///    Ok(ForkResult::Parent { child, .. }) => {
```
kamalmarhubi: `..`?

posborne: Basically, if other members were to ever be added in the future, things will keep working.

kamalmarhubi:
General thought: should we be linking just to Linux man pages at man7.org?

posborne:
> General thought: should we be linking just to Linux man pages at man7.org?

Although the man7.org pages are part of the Linux manpages project, Kerrisk has done a pretty good job of putting in notes on portability.  I think they are probably the single best source to link to if we are going to link to one source.  Linking to others might be something that could be done eventually, but at least initially I am relatively happy just linking to man7 unless there are major differences that users need to be aware of.

posborne:
I'm going to infer an approval from @kamalmarhubi and go ahead and get this merged as I have even more in the pipeline.  @homu r=kamalmarhubi

PR #452, opened 2016-10-29
Additional documentation additions

posborne:
This is a second round of doc additions.  There are also some funcitonal changes to APIs for getting/setting hostname that need to be reviewed.

r? @nix-rust/nix-maintainers

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1,5 +1,5 @@
-//! Standard symbolic constants and types
-//!
+//! Safe wrappers around functions found in libc "unistd.h" header
+
 use {Errno, Error, Result, NixPath};
 use fcntl::{fcntl, OFlag, O_CLOEXEC, FD_CLOEXEC};
 use fcntl::FcntlArg::F_SETFD;
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -411,6 +411,40 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
     Err(Error::Sys(Errno::last()))
 }
 
+/// Daemonize this process by detaching from the controlling terminal (see
+/// [daemon(3)](http://man7.org/linux/man-pages/man3/daemon.3.html)).
+///
+/// When a process is launched it is typically associated with a parent and it,
+/// in turn, by its controlling terminal/process.  In order for a process to run
+/// in the "background" it must daemonize itself by detaching itself.  Under
+/// posix, this is done by doing the following:
+///
+/// 1. Parent process (this one) forks
+/// 2. Parent process exits
+/// 3. Child process continues to run.
+///
+/// There are a couple options here whose names and meaning can be a bit
+/// confusing, so we'll describe the behavior for each state.
+///
+/// For `nochdir`:
+///
+/// * `nochdir = true`: The current working directory after daemonizing will
+///    be the current working directory.
+/// *  `nochdir = false`: The current working directory after daemonizing will
+///    be the root direcory, `/`.
+///
+/// For `noclose`:
+///
+/// * `noclose = true`: The process' current stdin, stdout, and stderr file
+///   descriptors will remain identical after daemonizing.
+/// * `noclose = false`: The process' stdin, stdout, and stderr will point to
+///   `/dev/null` after daemonizing.
+///
+/// The underlying implementation (in libc) calls both
+/// [fork(2)](http://man7.org/linux/man-pages/man2/fork.2.html) and
+/// [setsid(2)](http://man7.org/linux/man-pages/man2/setsid.2.html) and, as
+/// such, error that could be returned by either of those functions could also
+/// show up as errors here.
 pub fn daemon(nochdir: bool, noclose: bool) -> Result<()> {
     let res = unsafe { libc::daemon(nochdir as c_int, noclose as c_int) };
     Errno::result(res).map(drop)
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -344,6 +344,12 @@ fn to_exec_array(args: &[CString]) -> Vec<*const c_char> {
     args_p
 }
 
+/// Replace the current process image with a new one (see
+/// [exec(3)](http://man7.org/linux/man-pages/man3/exec.3.html)).
+///
+/// See the `::nix::unistd::execve` system call for additional details.  `execv`
+/// performs the same action but does not allow for customization of the
+/// environment for the new process.
 #[inline]
 pub fn execv(path: &CString, argv: &[CString]) -> Result<Void> {
     let args_p = to_exec_array(argv);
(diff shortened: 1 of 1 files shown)


src/unistd.rs
```diff
@@ -355,6 +361,24 @@ pub fn execv(path: &CString, argv: &[CString]) -> Result<Void> {
...
+/// command that is run. On success, this function will not return. Instead,
+/// the new program will run until it exits.
+///
+/// If an error occurs, this function will return with an indication of the
+/// cause of failure.  See
+/// [execve(2)#errors](http://man7.org/linux/man-pages/man2/execve.2.html#ERRORS)
+/// for a list of potential problems that maight cause execv to fail.
+///
+/// Both `::nix::unistd::execv` and `::nix::unistd::execve` take as arguments a
+/// slice of `::std::ffi::CString`s for `args` and `env`.  Each element in
```
fiveop: `execv` does not have the `env` parameter

src/unistd.rs
```diff
@@ -367,6 +391,15 @@ pub fn execve(path: &CString, args: &[CString], env: &[CString]) -> Result<Void>
...
 }
 
+/// Replace the current process image with a new one and replicate shell `PATH`
+/// searching behavior (see
+/// [exec(3)](http://man7.org/linux/man-pages/man3/exec.3.html)).
+///
+/// See `::nix::unistd::execve` for additoinal details.  `execvp` behaves the
+/// sme as execv except that it will examine the `PATH` environment variables
+/// for file names not specified with a leading slash.  For example, `execv`
+/// would not work if I specified "bash" for the path argument, but `execvp`
```
fiveop: I would prefer if the documentation where written consistently in the third person.

src/unistd.rs
```diff
@@ -378,12 +411,53 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
...
+/// * `noclose = true`: The process' current stdin, stdout, and stderr file
+///   descriptors will remain identical after daemonizing.
+/// * `noclose = false`: The process' stdin, stdout, and stderr will point to
+///   `/dev/null` after daemonizing.
+///
+/// The underlying implementation (in libc) calls both
+/// [fork(2)](http://man7.org/linux/man-pages/man2/fork.2.html) and
+/// [setsid(2)](http://man7.org/linux/man-pages/man2/setsid.2.html) and, as
+/// such, error that could be returned by either of those functions could also
+/// show up as errors here.
```
fiveop: What good is an abstraction, if we need to look at the implementation for documentation?

src/unistd.rs
```diff
@@ -378,12 +411,53 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
...
+/// When a process is launched it is typically associated with a parent and it,
+/// in turn, by its controlling terminal/process.  In order for a process to run
+/// in the "background" it must daemonize itself by detaching itself.  Under
+/// posix, this is done by doing the following:
+///
+/// 1. Parent process (this one) forks
+/// 2. Parent process exits
+/// 3. Child process continues to run.
+///
+/// There are a couple options here whose names and meaning can be a bit
```
fiveop: I would prefer having less prose here. Just explain what the parameters are for, not why it is necessary to document them.

src/unistd.rs
```diff
@@ -402,12 +476,39 @@ pub fn sethostname(name: &[u8]) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-pub fn gethostname(name: &mut [u8]) -> Result<()> {
-    let ptr = name.as_mut_ptr() as *mut c_char;
-    let len = name.len() as size_t;
+/// Get the host name and store it int the provided buffer, returning the actual
+/// length stored in the buffer (see
+/// [gethostname(2)](http://man7.org/linux/man-pages/man2/gethostname.2.html)).
```
fiveop: But we don't return the length (directly).

src/unistd.rs
```diff
@@ -402,12 +476,39 @@ pub fn sethostname(name: &[u8]) -> Result<()> {
...
+/// ```no_run
+/// use nix::unistd;
+///
+/// let mut buf = [0u8; 64];
+/// let hostname_cstr = unistd::gethostname(&mut buf).expect("Failed getting hostname");
+/// let hostname = hostname_cstr.to_str().expect("Hostname wasn't valid UTF-8");
+/// println!("Hostname: {}", hostname);
+/// ```
+pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
+    let ptr = buffer.as_mut_ptr() as *mut c_char;
```
fiveop: This is an aside with respect to this PR:

Shouldn't we change this method to just return the hostname if possible, by using our own buffer?

src/unistd.rs
posborne: Ah, yes.  This needs to be updated.  I went through a few iterations and ended up changing how this worked.

src/unistd.rs
posborne: We could and I considered this, but I wanted to provide a zero-cost abstraction first (the user may already have a buffer).  I would consider adding a variant as you mentioned -- not sure what it would be named exactly.

src/unistd.rs
posborne: I am referring users to the libc man pages, not to an implementation.  I think the former is acceptable.  This is basically the same approach taking by the man pages for this call.  We can go further with time, but I think this is a reasonable start.

src/unistd.rs
posborne: Sure, that can be accommodated.  In general, I am of the opinion that documentation changes moving forward should be fact-checked and merged first and improved in terms of content, quality, and consistency later, particularly given the somewhat poor state of our documentation at present.

posborne:
> Thanks for getting the documentation process started. In the long run, I would like to have some formal rules on how documentation should look like, because I think this will result in more consistent and more concise documentation.

Sure, I'm not opposed to aiming for consistency, but I think our primary goal needs to be improving the state of the documentation by at least having the basics for all public functionality.  Making it easy for doc changes to get merged and encouraging people to contribute are probably the things that will help us most.  We can update for consistency later.

src/unistd.rs
fiveop: Fair enough.

fiveop:
I agree with you that something, so long as it's not wrong, is better than nothing. I was just voicing what I would like to aim for.

r=me, after you've done whatever changes you deem necessary given my comments.

posborne:
> I agree with you that something, so long as it's not wrong, is better than nothing. I was just voicing what I would like to aim for.

Sure, I'll try to clean stuff up tonight.  Thanks for the review and the understanding!

src/unistd.rs
kamalmarhubi: Regarding this function, we cannot guarantee it is a valid `&CStr` that comes back:


> gethostname() returns the null-terminated hostname in the character array name, which has a length of len bytes.  If the null-terminated hostname is too large to fit, then the name is truncated, and __no error is returned__ (but see NOTES below).  POSIX.1 says that if such truncation occurs, then __it is unspecified whether the returned buffer includes a terminating null byte__.
```
―http://man7.org/linux/man-pages/man2/gethostname.2.html#DESCRIPTION

Comment (kamalmarhubi) [inline src/unistd.rs]: Also I'd prefer API changes in a separate PR from doc changes.

Comment (posborne) [inline src/unistd.rs]: > Regarding this function, we cannot guarantee it is a valid &CStr that comes back:

This is why I ensure that the last byte in the buffer is NUL.

File: CHANGELOG.md [hunk 7]
```diff
@@ -34,6 +34,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   accessible with the new method `events()` of `EpollEvent`. Instances of
   `EpollEvent` can be constructed using the new method `new()` of EpollEvent.
   ([#410](https://github.com/nix-rust/nix/pull/410))
+- `sethostname` now takes a `&str` instead of a `&[u8]` as this provides an API
+  that makes more sense in normal, correct usage of the API.
+- `gethostname` previously did not expose the actual length of the hostname
+  written from the underlying system call at all.  This has been updated to
+  return a `&CStr` within the provided buffer that is always properly
+  NUL-terminated (this is not guaranteed by the call with all platforms/libc
+  implementations).
```
kamalmarhubi: I would like both of these changes moved to a non-doc PR. That PR can include the docs for those functions, but I think it's best to have API changes not happen in a PR that only advertises doc changes.

src/unistd.rs
kamalmarhubi: Ah I missed that.

posborne:
I stripped this down to just the documentation commits and addressed comments, so getting this merged up.  @homu r=fiveop

kamalmarhubi:
@homu retry 173f2ac

kamalmarhubi:
@homu r=fiveop

PR #454, opened 2016-10-31
Bump minimum supported version of Rust

alexcrichton:
Many crates are coalescing around 1.9.0 with the stabilization of `std::panic`,
and it helps remove the build script w/ rustc version parsing for now as well.


posborne:
Thanks Alex, I'm good with this change.  Right now, CI is still testing (and passing) on 1.7.0 for some of the other non-x86 targets (another thing on my todo list).  For the 0.8 release, I am fine with going toward 1.9.0 as the minimum supported version.

Can you update the CHANGELOG.md file to indicate that 1.9.0 is now the minimum supported version (there is a note for the pending release about 1.7.0 present currently).

Thanks!

alexcrichton:
Oh! Actually reading more it looks like the version bump can be avoided. If the minimum version is 1.7.0 then the build script can still be removed as it's just checking for >= 1.6.0. Does that sound ok?

posborne:
Ah, sure enough.  Sounds good to me!  @homu r+

kamalmarhubi:
👍 I actually just did a `git pull` to go and remove the build script, and was confused about why I couldn't open it. Thanks @alexcrichton!

PR #455, opened 2016-11-02
Gethostname sethostname updates

posborne:
These changes have been previously discussed some with #452 but we decided it made sense to separate things out a bit more.

r? @kamalmarhubi @fiveop

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -506,12 +506,35 @@ pub fn sethostname(name: &str) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-pub fn gethostname(name: &mut [u8]) -> Result<()> {
-    let ptr = name.as_mut_ptr() as *mut c_char;
-    let len = name.len() as size_t;
+/// Get the host name and store it in the provided buffer, returning a pointer
+/// the CStr in that buffer on success (see
+/// [gethostname(2)](http://man7.org/linux/man-pages/man2/gethostname.2.html)).
+///
+/// This function call attempts to get the host name for the running system and
+/// store it in a provided buffer.  The buffer will be populated with bytes up
+/// to the length of the provided slice including a NUL terminating byte.  If
+/// the hostname is longer than the length provided, no error will be provided.
+/// The posix specification does not specify whether implementations will
+/// null-terminate in this case, but the nix implementation will ensure that the
+/// buffer is null terminated in this case.
+///
+/// ```no_run
+/// use nix::unistd;
+///
+/// let mut buf = [0u8; 64];
+/// let hostname_cstr = unistd::gethostname(&mut buf).expect("Failed getting hostname");
+/// let hostname = hostname_cstr.to_str().expect("Hostname wasn't valid UTF-8");
+/// println!("Hostname: {}", hostname);
+/// ```
+pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
+    let ptr = buffer.as_mut_ptr() as *mut c_char;
+    let len = buffer.len() as size_t;
 
     let res = unsafe { libc::gethostname(ptr, len) };
-    Errno::result(res).map(drop)
+    Errno::result(res).map(|_| {
+        buffer[len - 1] = 0; // ensure always null-terminated
+        unsafe { CStr::from_ptr(buffer.as_ptr() as *const c_char) }
+    })
 }
 
 pub fn close(fd: RawFd) -> Result<()> {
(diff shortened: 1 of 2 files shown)


src/unistd.rs
```diff
@@ -447,7 +447,14 @@ pub fn daemon(nochdir: bool, noclose: bool) -> Result<()> {
...
 
-pub fn sethostname(name: &[u8]) -> Result<()> {
+/// Set the system host name (see
+/// [gethostname(2)](http://man7.org/linux/man-pages/man2/gethostname.2.html)).
+///
+/// Given a name, attempt to update the system host name to the given string.
+/// On some systems, the host name is limited to as few as 64 bytes.  An error
+/// will be return if the name is not valid or the current process does not have
+/// permissions to update the host name.
+pub fn sethostname(name: &str) -> Result<()> {
```
kamalmarhubi: I'm not convinced `&str` is the right thing here. From your commit message:

> Although the underlying C API does take a pointer to a set of characters, *it is a requirement of almost every operating system that these bytes __not contain a premature NUL character__ or __other
special characters__*.  In other words, you want a `&str`.  Changing this to make the API make a bit more sense.

`&str` can contain a NUL byte anywhere, and is UTF-8, which may contain "special characters". Not really sure what to do here though!

PS commit message subject says `gethostname` when it should be `sethostname`.

src/unistd.rs
```diff
@@ -466,12 +473,39 @@ pub fn sethostname(name: &[u8]) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-pub fn gethostname(name: &mut [u8]) -> Result<()> {
-    let ptr = name.as_mut_ptr() as *mut c_char;
-    let len = name.len() as size_t;
+/// Get the host name and store it int the provided buffer, returning the actual
```
kamalmarhubi: store it *in* the

src/unistd.rs
```diff
@@ -466,12 +473,39 @@ pub fn sethostname(name: &[u8]) -> Result<()> {
...
+/// let hostname = hostname_cstr.to_str().expect("Hostname wasn't valid UTF-8");
+/// println!("Hostname: {}", hostname);
+/// ```
+pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
+    let ptr = buffer.as_mut_ptr() as *mut c_char;
+    let len = buffer.len() as size_t;
 
     let res = unsafe { libc::gethostname(ptr, len) };
-    Errno::result(res).map(drop)
+    Errno::result(res).map(move |_| {
```
kamalmarhubi: is `move` necessary?

src/unistd.rs
```diff
@@ -466,12 +473,39 @@ pub fn sethostname(name: &[u8]) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-pub fn gethostname(name: &mut [u8]) -> Result<()> {
-    let ptr = name.as_mut_ptr() as *mut c_char;
-    let len = name.len() as size_t;
+/// Get the host name and store it int the provided buffer, returning the actual
+/// length stored in the buffer (see
```
kamalmarhubi: the return value is a `&CStr` not any length?

src/unistd.rs
posborne: > &str can contain a NUL byte anywhere, and is UTF-8, which may contain "special characters". Not really sure what to do here though!

That's kind of where I landed as well.  Thinking about this more, I'm wondering what you think about the following:

```rust
pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
```

This would allow a greater variety of types (including strings) to work with the function and communicates a little more clearly that we are basically going to let the OS do validation.

src/unistd.rs
posborne: Yeah, I need to update this doc.  When I was first writing this I though the system call returned the length but it does not.

src/unistd.rs
posborne: I don't think so; I'll remove it.  Not sure what I was thinking.

src/unistd.rs
fiveop: Sounds reasonable.

lucab:
Any update on this? I'm interested in having a more idiomatic interface for hostname handling exposed through nix.

posborne:
> Any update on this? I'm interested in having a more idiomatic interface for hostname handling exposed through nix.

Sorry for the delay.  I kinda forgot I had opened this awhile back.  I'll update it and we'll see about getting things merged.

posborne:
Ok, @kamalmarhubi @fiveop I believe all comments on this one are addressed now (and I've remembered that this PR exists).  Ready to go?

posborne:
@homu r=kamalmarhubi :)

PR #456, opened 2016-11-02
Add openpty

cactorium:
Lemme know if anything needs to be fixed

### DIFF
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -1,12 +1,25 @@
 //! Create master and slave virtual pseudo-terminals (PTYs)
 
+use libc;
+
+pub use libc::pid_t as SessionId;
+pub use libc::winsize as Winsize;
+
 use std::ffi::CStr;
 use std::mem;
 use std::os::unix::prelude::*;
 
-use libc;
+use sys::termios::Termios;
+use {Errno, Result, Error, fcntl};
 
-use {Error, fcntl, Result};
+/// Representation of a master/slave pty pair
+///
+/// This is returned by `openpty`
+pub struct OpenptyResult {
+    pub master: RawFd,
+    pub slave: RawFd,
+}
+
 
 /// Representation of the Master device in a master/slave pty pair
 ///
@@ -161,4 +174,43 @@ pub fn unlockpt(fd: &PtyMaster) -> Result<()> {
     }
 
     Ok(())
+}
+
+
+/// Create a new pseudoterminal, returning the slave and master file descriptors
+/// in `OpenptyResult`
+/// (see [openpty](http://man7.org/linux/man-pages/man3/openpty.3.html)). 
+///
+/// If `winsize` is not `None`, the window size of the slave will be set to
+/// the values in `winsize`. If `termios` is not `None`, the pseudoterminal's
+/// terminal settings of the slave will be set to the values in `termios`.
+#[inline]
+pub fn openpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>>>(winsize: T, termios: U) -> Result<OpenptyResult> {
+    use std::ptr;
+
+    let mut slave: libc::c_int = -1;
+    let mut master: libc::c_int = -1;
+    let c_termios = match termios.into() {
+        Some(termios) => termios as *const Termios,
+        None => ptr::null() as *const Termios,
+    };
+    let c_winsize = match winsize.into() {
+        Some(ws) => ws as *const Winsize,
+        None => ptr::null() as *const Winsize,
+    };
+    let ret = unsafe {
+        libc::openpty(
+            &mut master as *mut libc::c_int,
+            &mut slave as *mut libc::c_int,
+            ptr::null_mut(),
+            c_termios as *mut libc::termios,
+            c_winsize as *mut Winsize)
+    };
+
+    Errno::result(ret)?;
+
+    Ok(OpenptyResult {
+        master: master,
+        slave: slave,
+    })
 }
--- a/test/test.rs
+++ b/test/test.rs
@@ -11,19 +11,17 @@ extern crate nix_test as nixtest;
 
 mod sys;
 mod test_fcntl;
+#[cfg(any(target_os = "linux"))]
+mod test_mq;
 mod test_net;
 mod test_nix_path;
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+mod test_poll;
+mod test_pty;
 #[cfg(any(target_os = "linux", target_os = "android"))]
 mod test_sendfile;
 mod test_stat;
 mod test_unistd;
-
-#[cfg(any(target_os = "linux"))]
-mod test_mq;
-
-#[cfg(any(target_os = "linux", target_os = "macos"))]
-mod test_poll;
-mod test_pty;
 
 use nixtest::assert_size_of;
 
--- a/test/test_pty.rs
+++ b/test/test_pty.rs
@@ -1,8 +1,11 @@
 use std::path::Path;
 use std::os::unix::prelude::*;
+
 use nix::fcntl::{O_RDWR, open};
 use nix::pty::*;
 use nix::sys::stat;
+use nix::sys::termios::*;
+use nix::unistd::{read, write, close};
 
 /// Test equivalence of `ptsname` and `ptsname_r`
 #[test]
@@ -89,4 +92,83 @@ fn test_open_ptty_pair() {
     // Open the slave device
     let slave_fd = open(Path::new(&slave_name), O_RDWR, stat::Mode::empty()).unwrap();
     assert!(slave_fd > 0);
+}
+
+#[test]
+fn test_openpty() {
+    let pty = openpty(None, None).unwrap();
+    assert!(pty.master > 0);
+    assert!(pty.slave > 0);
+
+    // Writing to one should be readable on the other one
+    let string = "foofoofoo\n";
+    let mut buf = [0u8; 16];
+    write(pty.master, string.as_bytes()).unwrap();
+    let len = read(pty.slave, &mut buf).unwrap();
+
+    assert_eq!(len, string.len());
+    assert_eq!(&buf[0..len], string.as_bytes());
+
+    // Read the echo as well
+    let echoed_string = "foofoofoo\r\n";
+    let len = read(pty.master, &mut buf).unwrap();
+    assert_eq!(len, echoed_string.len());
+    assert_eq!(&buf[0..len], echoed_string.as_bytes());
+
+    let string2 = "barbarbarbar\n";
+    let echoed_string2 = "barbarbarbar\r\n";
+    write(pty.slave, string2.as_bytes()).unwrap();
+    let len = read(pty.master, &mut buf).unwrap();
+
+    assert_eq!(len, echoed_string2.len());
+    assert_eq!(&buf[0..len], echoed_string2.as_bytes());
+
+    close(pty.master).unwrap();
+    close(pty.slave).unwrap();
+}
+
+#[test]
+fn test_openpty_with_termios() {
+    // Open one pty to get attributes for the second one
+    let mut termios = {
+        let pty = openpty(None, None).unwrap();
+        assert!(pty.master > 0);
+        assert!(pty.slave > 0);
+        let termios = tcgetattr(pty.master).unwrap();
+        close(pty.master).unwrap();
+        close(pty.slave).unwrap();
+        termios
+    };
+    termios.c_oflag &= !ONLCR;
+
+    let pty = openpty(None, &termios).unwrap();
+    // Must be valid file descriptors
+    assert!(pty.master > 0);
+    assert!(pty.slave > 0);
+
+    // Writing to one should be readable on the other one
+    let string = "foofoofoo\n";
+    let mut buf = [0u8; 16];
+    write(pty.master, string.as_bytes()).unwrap();
+    let len = read(pty.slave, &mut buf).unwrap();
+
+    assert_eq!(len, string.len());
+    assert_eq!(&buf[0..len], string.as_bytes());
+
+    // read the echo as well
+    let echoed_string = "foofoofoo\n";
+    let len = read(pty.master, &mut buf).unwrap();
+    assert_eq!(len, echoed_string.len());
+    assert_eq!(&buf[0..len], echoed_string.as_bytes());
+
+    let string2 = "barbarbarbar\n";
+    let echoed_string2 = "barbarbarbar\n";
+    write(pty.slave, string2.as_bytes()).unwrap();
+    let len = read(pty.master, &mut buf).unwrap();
+
+    assert_eq!(len, echoed_string2.len());
+    assert_eq!(&buf[0..len], echoed_string2.as_bytes());
+
+    close(pty.master).unwrap();
+    close(pty.slave).unwrap();
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#551](https://github.com/nix-rust/nix/pull/551))
 - Added `nix::pty::{grantpt, posix_openpt, ptsname/ptsname_r, unlockpt}`
   ([#556](https://github.com/nix-rust/nix/pull/556)
+- Added `nix::ptr::openpty`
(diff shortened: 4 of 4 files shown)


cactorium:
So it's only failing on aarch64, because apparently some iotctl used in `openpty()` isn't implemented on it. Should I just add a `#[cfg(not(target_arch="aarch64"))]` to the test?

posborne:
> So it's only failing on aarch64, because apparently some iotctl used in openpty() isn't implemented on it. Should I just add a #[cfg(not(target_arch="aarch64"))] to the test?

I think our preference would be to work with libc upstream to fix aarch64.  Or are you saying that openpty isn't supported at all on aarch64?

cactorium:
I'm not sure; the test run is failing with `test test_pty::test_openpty ... Unsupported ioctl: cmd=0x80045430` (which is `TIOCGPTN` for future reference), and returning `Sys(ENOSYS)`, which means "unimplemented" according to the man pages. So it's either in libc using the wrong syscalls on this platform (I'm guessing pretty unlikely), the aarch64 kernel not implementing it (also unlikely), or the QEMU emulation of the system not implementing it (probably this one?). Do you have any ideas on how to figure out which versions of these things are being run?

lucab:
I think you may have hit https://lists.gnu.org/archive/html/qemu-devel/2016-07/msg04286.html.

kamalmarhubi:
> Do you have any ideas on how to figure out which versions of these things are being run?
@posborne do you know offhand the version of qemu that's in your docker image?

posborne:
> @posborne do you know offhand the version of qemu that's in your docker image?

Unfortunately, no.  The base image I built stuff off of at the time was 15.10.

cactorium:
I think it's 0.10.50? I have no idea how Docker works, but I think I managed to pull the Docker image and start a container, and from inside a shell in that container:
```
docker pull posborne/rust-cross:android
# there was a lot of messing around between these two commands, but I think you only need these two
docker run --rm -it --name foo -p 5000:80 posborne/rust-cross:android /bin/sh
```

`/android/sdk/tools/emulator-arm -qemu -version` gave me 

```
QEMU PC emulator version 0.10.50Android, Copyright (c) 2003-2008 Fabrice Bellard

```

posborne:
Yeah, we can certainly bump.  I would really like to move the CI to something like https://github.com/japaric/trust

berkowski:
@posborne I could take a stab at getting `trust` CI up and running for `nix`.  I'd like to learn more about it for my own projects anyway.

posborne:
> @posborne I could take a stab at getting trust CI up and running for nix. I'd like to learn more about it for my own projects anyway.

@berkowski That would be awesome.

Susurrus:
@cactorium Can you rebase this so we can give it a run through now that we're testing code on more platforms?

cactorium:
Can do!

Susurrus:
@cactorium Sorry I didn't notice this the first time, but it looks like this should really be one commit, as you have a few smaller fixing commits. Can you squash this down into one? That would help me review it.

I'm also curious how this will end up interacting with #527. @cactorium Can you take a look at that issue, it kind of exploded into talking about how best to deal with the `termios` in `nix`? I don't want to merge this or that other one until we really know what we're getting ourselves into.

@berkowski If you have time can you take a look over this as well? This is probably of interest to you.

cactorium:
Err I think I messed up the commit history on this PR a lot, I guess pulling would've been a better choice than an actual rebase. Would you guys have any advice on cleaning up the commit history? I'll look at the other PR in a bit

Susurrus:
1. Do a `git reset upstream/master --hard` or use `origin` if that's what you called the `nix-rust/nix` repo.  
 2. Then cherry-pick your 3 main commits like `cherry-pick 397840a 0e3fcdb 75f44c9`.

Then you can squash them with `git rebase -i` and force-push again.

src/pty.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// Create a new pseudoterminal, returning the slave and master file descriptors
+/// in `OpenptyResult`
+/// (see [openpty](http://man7.org/linux/man-pages/man3/openpty.3.html)). 
+///
+/// If `winsize` is not `None`, the window size of the slave will be set to
+/// the values in `winsize`. If `termios` is not `None`, the pseudoterminal's
+/// terminal settings of the slave will be set to the values in `termios`.
+#[inline]
+pub fn openpty(winsize: Option<Winsize>, termios: Option<Termios>) -> Result<OpenptyResult> {
```
Susurrus: First, can't both of be borrows?

Second, we should be able to use a slightly different function declaration to give us better ergonomics when using a real struct for either of the arguments, something like:
```
pub fn openpty<T: Into<Option<&Winsize>>, U: Into<Option<&Termios>>>(winsize: T, termios: U) -> ... {
```
So now you can use `.into()` to convert the options to the struct representations and the caller doesn't need to wrap their struct reference in a `Some`. I don't know if the syntax above is exactly correct, but I think it should be. The [gtk-rs](https://github.com/gtk-rs/gtk) project did this quite successfully for all their string types and it's nice ([here](https://github.com/gtk-rs/gtk/blob/a40a4742ba09e11380e4cb1c51628e4aa4851d19/src/auto/tool_button.rs#L28) is an example).

test/test_pty.rs
```diff
@@ -0,0 +1,36 @@
+use nix::pty::openpty;
+use nix::unistd::{read, write, close};
+
+#[test]
+fn test_openpty() {
+    // TODO: figure out the right termios settings to pass in here
```
cactorium: Fixed

src/pty.rs
```diff
@@ -0,0 +1,50 @@
...
+    };
+    let c_winsize = match &winsize {
+        &Some(ref ws) => ws as *const Winsize,
+        &None => 0 as *const Winsize,
+    };
+    let ret = unsafe {
+        libc::openpty(
+            &mut master as *mut libc::c_int,
+            &mut slave as *mut libc::c_int,
+            0 as *mut libc::c_char,
```
cactorium: Fixed

src/pty.rs
```diff
@@ -0,0 +1,50 @@
...
+pub fn openpty(winsize: Option<Winsize>, termios: Option<Termios>) -> Result<OpenptyResult> {
+    let mut slave: libc::c_int = -1;
+    let mut master: libc::c_int = -1;
+    let c_termios = match &termios {
+        &Some(ref termios) => termios as *const Termios,
+        &None => 0 as *const Termios,
+    };
+    let c_winsize = match &winsize {
+        &Some(ref ws) => ws as *const Winsize,
+        &None => 0 as *const Winsize,
```
cactorium: Fixed

src/pty.rs
```diff
@@ -0,0 +1,50 @@
...
+/// If `winsize` is not `None`, the window size of the slave will be set to
+/// the values in `winsize`. If `termios` is not `None`, the pseudoterminal's
+/// terminal settings of the slave will be set to the values in `termios`.
+#[inline]
+pub fn openpty(winsize: Option<Winsize>, termios: Option<Termios>) -> Result<OpenptyResult> {
+    let mut slave: libc::c_int = -1;
+    let mut master: libc::c_int = -1;
+    let c_termios = match &termios {
+        &Some(ref termios) => termios as *const Termios,
+        &None => 0 as *const Termios,
```
cactorium: Fixed

src/pty.rs
cactorium: Fixed

test/test_pty.rs
cactorium: Any advice on what flags to play with? The only ones I can think of messing with don't do much other than turn CR on/off

Susurrus: It doesn't really matter what flags you set so long as the function gets tested with `None` and `Some(_)` types really. If the CR thing works, go ahead and add a test that tests that.

cactorium: After looking into it, there's not really a nice way to initialize a Termios object in a platform-agnostic manner until it implements `Default`, so I'd argue for pushing that fix later in a separate PR

Susurrus: Could you file a ticket against `libc` about this then? I'd actually prefer if we could resolve that before merging this if it's relatively easy.

cactorium: Fixing it in `libc` won't fix it because as of right now `Termios` is directly defined in `nix/sys/termios.rs`. If this lands before your PR, it'd make sense to fix it afterwards

Susurrus: Well that makes it pretty easy then, can you `#derive(Default)` on the multiple Termios struct declarations and then use that in your tests. I'll then make sure those tests don't break when resolving #527. This is being tracked in rust-lang/libc#573.

cactorium: Is it okay that `Default` won't necessarily set the fields to valid values? I have no idea if zero is a valid/correct default value for all the fields

Susurrus: `Default` should always set the fields to valid values, it kind of defeats the point if it doesn't. For the termios struct I'm pretty certain for the `termios` struct that all zeros is valid.

But for your tests, just create a termios struct out of uninitialized memory and use the various methods available to set the struct to your desired values. Does that work?

Susurrus:
@cactorium Can you please also add an entry to the CHANGELOG?

test/test_pty.rs
cactorium: So I ended up using `tcgetattr` to populate the `Termios`. Apparently this breaks on OS X because they don't support `tcgetattr`?

cactorium: Yeah, `tcgetattr(stdin)` is failing on OS X, added two print!() statements around that call and only the first one shows up in the log. Two obvious options are to either disable the test on OS X or switch back to just using `(None, None)`

Susurrus: You can only use `tcgetattr` on ttys, which I don't think stdin is, which explains why it fails on mac. What I would recommend is calling `tcgetattr` on one of the pttys you create and then use that to open one with that termios. So you'd do both of these in the same test. I'd make that a separate test than the one you already have that is testing reading & writing from a master/slave pair.

cactorium: Huh, I'm pretty sure it is on Linux-based platforms at least. From my understanding, the terminal emulator spawns a pty, gives the slave (or master? one of them) to the shell program, and draws stuff to the screen based on what it receives on its end. Anyways, that fix did it, both tests passing on all important test cases

test/test_pty.rs
```diff
@@ -0,0 +1,81 @@
+use nix::pty::openpty;
+use nix::sys::termios::*;
+use nix::unistd::{read, write, close};
+
+#[test]
+fn test_openpty() {
+    let pty = openpty(None, None).unwrap();
+    assert!(pty.master > 0); // must be valid file descriptors
```
Susurrus: Please capitalize the "m" in "must" and move this comment to the above line if you want to keep it, though I don't see this comment as very useful. Maybe just remove these trailing comments from the code.

Susurrus:
Can you add another test using a valid winsize argument? I can't easily figure out what that struct is, but it'd be nice to have verification that that works as well.

cactorium:
Well, that sets the size of the terminal window, and there's AFAIK really no nice way to test for it that doesn't involve `ioctl`

Susurrus:
Alright, then let's skip it.

cactorium:
Soo all's good then?

Susurrus:
Yes, but I'd like to wait til #590 is resolved before merging any new features.

Susurrus:
@cactorium Can you rebase this now that #556 just merged?

cactorium:
Done

src/pty.rs
```diff
@@ -1,12 +1,23 @@
...
 use std::ffi::CStr;
 use std::mem;
 use std::os::unix::prelude::*;
+use std::os::unix::io::RawFd;
 
-use libc;
+use sys::termios::Termios;
+use {Errno, Result, Error, fcntl};
+
+pub struct OpenptyResult {
```
Susurrus: This needs a doccomment since it's a public type.

src/pty.rs
```diff
@@ -162,3 +173,42 @@ pub fn unlockpt(fd: &PtyMaster) -> Result<()> {
...
+    let ret = unsafe {
+        libc::openpty(
+            &mut master as *mut libc::c_int,
+            &mut slave as *mut libc::c_int,
+            ptr::null_mut(),
+            c_termios as *mut libc::termios,
+            c_winsize as *mut Winsize)
+    };
+
+    let _ = try!(Errno::result(ret));
```
Susurrus: Can't this just be `Errno::result(ret)?;` instead? Can you explain what this line is doing for me?

test/test.rs
```diff
@@ -13,6 +13,7 @@ mod sys;
 mod test_fcntl;
 mod test_net;
 mod test_nix_path;
+mod test_pty;
```
Susurrus: Yeah, we should really alphabetize this list.

test/test_pty.rs
```diff
@@ -90,3 +93,81 @@ fn test_open_ptty_pair() {
...
     assert!(slave_fd > 0);
 }
+
+#[test]
+fn test_openpty() {
+    let pty = openpty(None, None).unwrap();
+    assert!(pty.master > 0);
+    assert!(pty.slave > 0);
+
+    // writing to one should be readable on the other one
```
Susurrus: Can you capitalize the first work in your line comments? Please do that to all of them as well.

test/test_pty.rs
```diff
@@ -90,3 +93,81 @@ fn test_open_ptty_pair() {
...
+
+    let string2 = "barbarbarbar\n";
+    let echoed_string2 = "barbarbarbar\n";
+    write(pty.slave, string2.as_bytes()).unwrap();
+    let len = read(pty.master, &mut buf).unwrap();
+
+    assert_eq!(len, echoed_string2.len());
+    assert_eq!(&buf[0..len], echoed_string2.as_bytes());
+
+    close(pty.master).unwrap();
```
Susurrus: Don't you normally want to close the slave first? Is there anywhere that suggest that closing the master will close the slave as well at the OS-level?

cactorium:
Sorry for taking so long to push changes, it's been a busy few weeks

src/pty.rs
cactorium: Done

src/pty.rs
cactorium: Fixed! I forgot about the new `?` operator, which is why I used a `try!` instead

test/test.rs
cactorium: Done

test/test_pty.rs
cactorium: Fixed

test/test_pty.rs
cactorium: I'm not sure; the best documentation I could find seemed to imply either way was okay (although nothing implies that closing either end will close the other one):

```
If fildes refers to the master side of a pseudo-terminal, and this is the last close, a SIGHUP signal shall be sent to the controlling process, if any, for which the slave side of the pseudo-terminal is the controlling terminal. It is unspecified whether closing the master side of the pseudo-terminal flushes all queued input and output.

If fildes refers to the slave side of a STREAMS-based pseudo-terminal, a zero-length message may be sent to the master.

```

cactorium:
Didn't notice the build fail, it's fixed now

Susurrus:
Looks good. Now just blocking on #590.

Susurrus:
@cactorium Can you rebase this on HEAD now that our new testing framework has merged? This is RTM after it passes those tests.

cactorium:
Done

Susurrus:
@asomers Can you check into why your FreeBSD buildbot isn't running?

cactorium:
Isn't that just disabled? It's got that DISABLE_TESTS=1 flag on all the recent builds I looked at: https://travis-ci.org/nix-rust/nix/builds

Susurrus:
It's not run through Travis CI. We have a separate build server that runs tests on FreeBSD. @asomers runs it.

asomers:
Buildbot saw the most recent change, yet did not decide to do a build.  I'll dig deeper after I get home tonight.

asomers:
I couldn't find anything obviously wrong with the BuildBot server.  I've written to the mailing list for help.  but in the meantime, I suggest we don't hold up PRs waiting for it to complete.  Let's go ahead and merge anything compiles for i686-unknown-freebsd, which is done through Travis.

PR #457, opened 2016-11-03
Add CLONE_NEWCGROUP

dgreid:
This is a new option to clone added earlier this year.  It was already added to rust/libc.


posborne:
Thanks! @homu r+

fiveop:
Could we get an entry for this in CHANGELOG.md?

Issue #459, opened 2016-11-06
signal::SigFlags should be enum instead of bitflags.


chaosagent:
http://man7.org/linux/man-pages/man3/pthread_sigmask.3.html

`SigFlags` represents the `how` parameter of pthread_sigmask, which takes 3 mutually exclusive options, which should be represented by an enum. The `SigFlags` name is pretty confusing too, as the underlying value is an option, not a flag.

chaosagent:
Fixing this would require some breaking changes, but I don't know what the policy on that is.

kamalmarhubi:
> Fixing this would require some breaking changes, but I don't know what the policy on that is.

A big part of the motivation for a changelog is to at least make it possible to see what has changed. Pre-1.0 we're fairly happy to make changes for API improvements. This is fixing the semantics of an API and should certainly be allowed, IMO.

kamalmarhubi:
For more on this parameter: http://man7.org/linux/man-pages/man2/sigprocmask.2.html#DESCRIPTION

Certainly agree this should be an enum.

fiveop:
#460 is merged, so this issue can be closed.

PR #460, opened 2016-11-06
Change SigFlags into an enum.

chaosagent:
Addresses #459.

This is a breaking change.

Should SigFlags be renamed to something more sensible?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   accessible with the new method `events()` of `EpollEvent`. Instances of
   `EpollEvent` can be constructed using the new method `new()` of EpollEvent.
   ([#410](https://github.com/nix-rust/nix/pull/410))
+- `SigFlags` in `::nix::sys::signal` has be renamed to `SigmaskHow` and its type
+  has changed from `bitflags` to `enum` in order to conform to our conventions.
+  ([#410](https://github.com/nix-rust/nix/pull/460))
 
 ### Fixed
 - Fixed using kqueue with `EVFILT_USER` on FreeBSD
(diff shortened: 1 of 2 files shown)


fiveop:
The PR looks good.

I agree that the name should be different. I cannot think of a good naming convention that would work for all enumerations (e.g. `SigEnum` just does not tell us anything). Therefore, I would propose to go with `SigHow`.

In addition, could you please add an entry to CHANGELOG.md describing the breaking change.

chaosagent:
I was thinking of something like `SigmaskHow` since this enum only applies to to the `pthread_sigmask` and the `sigmask` (unimplemented in nix) calls but not the rest of the signal functions. In contrast, `SigSet` applies to both `sigaction` and `sigmask`, so it deserves to be labelled generically as `Sig`. Another possibility is to just use `Request` as the standard way to specify `how`-like values in the library like with PtraceRequest, which is actually enumerated in C `libc` [1], but using `How` might be advantageous in that it reflects the argument names in `libc`. Thoughts?

I'm probably making way too much of a fuss about naming though ;)

I'll hold off on the CHANGELOG changes until we figure this out I guess.
[1] http://man7.org/linux/man-pages/man2/ptrace.2.html

chaosagent:
I went ahead and pushed a rename to `SigmaskHow` and an entry in `CHANGELOG.md`.

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   accessible with the new method `events()` of `EpollEvent`. Instances of
   `EpollEvent` can be constructed using the new method `new()` of EpollEvent.
   ([#410](https://github.com/nix-rust/nix/pull/410))
+- `SigFlags` in `::nix::sys::signal` has be renamed to `SigmaskHow` and its type
+  has changed from `bitflags` to `enum` in order to conform to our conventions.
+  ([#410](https://github.com/nix-rust/nix/pull/460))
```
fiveop: Link name is #410 instead of #460.

fiveop:
@homu r+

Issue #439, opened 2016-10-16
ptrace: Ptrace{Requests, Events, Options} are not in an enum


chaosagent:
https://github.com/nix-rust/nix/blob/master/src/sys/ptrace.rs#L14-L63

From CONVENTIONS.md:

```
We represent sets of constants that are intended as mutually exclusive arguments to parameters of functions by enumerations.
```

kamalmarhubi:
Thanks for the report! Saw your #438 as well. Going to comment there.

Susurrus:
I believe this is now resolved with #749.

PR #461, opened 2016-11-06
[WIP] Move Ptrace{Request, Event, Options} into enums and bitflags.

chaosagent:
Addresses #439.

This is a breaking change.

Currently, `PTRACE_EVENT`s aren't defined in libc, so they are hardcoded here.

- [ ] OSX support
- [ ] fix i686/arm

### DIFF
--- a/src/sys/ptrace.rs
+++ b/src/sys/ptrace.rs
@@ -1,66 +1,87 @@
 use {Errno, Error, Result};
-use libc::{pid_t, c_void, c_long};
+use libc::{pid_t, c_int, c_void, c_long};
 
-#[cfg(all(target_os = "linux",
-          any(target_arch = "x86",
-              target_arch = "x86_64",
-              target_arch = "arm")),
-          )]
+#[cfg(any(target_os = "linux", target_os = "android"))]
 pub mod ptrace {
-    use libc::c_int;
+    use libc;
 
-    pub type PtraceRequest = c_int;
+    #[repr(u32)]
+    pub enum PtraceRequest {
+        PTRACE_TRACEME     = libc::PTRACE_TRACEME,
+        PTRACE_PEEKTEXT    = libc::PTRACE_PEEKTEXT,
+        PTRACE_PEEKDATA    = libc::PTRACE_PEEKDATA,
+        PTRACE_PEEKUSER    = libc::PTRACE_PEEKUSER,
+        PTRACE_POKETEXT    = libc::PTRACE_POKETEXT,
+        PTRACE_POKEDATA    = libc::PTRACE_POKEDATA,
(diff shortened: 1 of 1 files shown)


chaosagent:
The `ptrace` ffi is included in `libc` now; should I make a new pull request to change that or do it here?

chaosagent:
We could export all the constants and use something like a `IntoPtraceRequest` trait to keep backwards compatibility, but that would be a bit confusing...

src/sys/ptrace.rs
```diff
@@ -1,66 +1,87 @@
...
+        PTRACE_EVENT_SECCOMP    = 7,
+        PTRACE_EVENT_STOP       = 128,
+    }
 
-    pub type PtraceEvent = c_int;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    impl PtraceEvent {
+        /// Creates a PtraceEvent from the extra bits of a wait status (status >> 16)
+        #[inline]
+        pub fn from_c_int(event: libc::c_uint) -> Option<PtraceEvent> {
```
fiveop: for `::nix::sys::signal::Signal` we used `Result<Signal>` as a return type for the corresponding function. I think this is better, because I don't think anyone would intentionally put a wrong `c_uint` into `from_c_int`, so the `None` return type is never expected, thus should probably be an error type.

src/sys/ptrace.rs
```diff
@@ -1,66 +1,87 @@
...
-    pub const PTRACE_SETREGSET:   PtraceRequest = 0x4205;
-    pub const PTRACE_SEIZE:       PtraceRequest = 0x4206;
-    pub const PTRACE_INTERRUPT:   PtraceRequest = 0x4207;
-    pub const PTRACE_LISTEN:      PtraceRequest = 0x4208;
-    pub const PTRACE_PEEKSIGINFO: PtraceRequest = 0x4209;
+    // These aren't currently in libc.
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    #[derive(Eq, PartialEq, Clone, Copy, Debug)]
+    #[repr(u32)]
+    pub enum PtraceEvent {
```
fiveop: I think we should use `#[repr(i32)]` and thus consistently the type, which is used by the ffi function.

src/sys/ptrace.rs
chaosagent: Yeah sounds good I'll change that soon

Susurrus:
Any willingness to push this over the finish line @chaosagent? This came up again because of #614. I'd love to get the `ptrace` module a little more cleaned up.

Susurrus:
@chaosagent Just wanted to ping you again on this as I'd really like to clean up our ptrace API. Would you be able to clean this up anytime in the near future?

Susurrus:
This was closed by #749.

PR #463, opened 2016-11-11
Change KEvent to treat udata as an intptr_t instead of a uintptr_t.

asomers:
This matches NetBSD's C definitions.  Other operating systems define
it as void*, despite not really being a pointer, but none actually
define it as uintptr_t.  Better to be right on NetBSD and wrong
everywhere else than wrong everywhere.  Plus, it's what mio expects.

Please include this PR in nix 0.8.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,11 +24,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - The minimum supported version of rustc is now 1.7.0.
   ([#444](https://github.com/nix-rust/nix/pull/444))
-- Implement `Send` for `KEvent`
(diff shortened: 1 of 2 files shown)


asomers:
The only test failure was due to a problem setting up the VM and was not caused by this PR.

fiveop:
I think it's fine, except for the missing CHANGELOG.md entry.

asomers:
Does it really need a CHANGELOG entry if it makes it into the same release as PR #415 ?  I consider this PR to just be a tweak to that one.

fiveop:
It does not need its own entry then, but we should link to both PRs.

asomers:
Ok, I'll update it.

fiveop:
@homu r+

PR #465, opened 2016-11-14
Avoid TempDir::into_path(), because it doesn't cleanup on Drop

asomers:


posborne:
Good catch, looks good to me.  @homu r+

PR #469, opened 2016-11-16
Fix intermittency in test_select

asomers:
Sometimes, on a heavily laden system, select would timeout and the test would
fail.  Fix it by removing the timeout.

### DIFF
--- a/test/sys/test_select.rs
+++ b/test/sys/test_select.rs
@@ -41,7 +41,7 @@ fn test_select() {
     fd_set.insert(r1);
     fd_set.insert(r2);
 
-    let mut timeout = TimeVal::seconds(1);
+    let mut timeout = TimeVal::seconds(10);
     assert_eq!(1, select(r2 + 1,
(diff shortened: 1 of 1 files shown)


fiveop:
Devils advocat: What if there is a bug and we wait the full 10 minutes till travis kills the job? Could we go with 10 seconds to alleviate the problem?

asomers:
There's still no guarantee that it will finish in 10s.  We could use a loop, but we might loop until the 10m are complete.  How bad would it be to tie up Travis for 10m?

fiveop:
It would be polite to not tie on up for 10m.

posborne:
Although it could fail at 10s, I think there are probably scores of things that are more likely to fail the build prior to that happening and I do think this is a "better" test in some sense (in addition to being more polite).  I might be running the tests, for instance, on some other infrastructure that does not time out.

posborne:
@homu r+

PR #470, opened 2016-11-16
Get rid of version suffix between versions

fiveop:


posborne:
Looks good.  @homu r+

PR #471, opened 2016-11-16
Push minimum supported version of Rust to 1.9.0

fiveop:
This PR is not fully prepared yet. I first want to see what breaks in the CI, if we do this. Once everything works, I will add a CHANGELOG.md entry.

This is done to allow #431 to go forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
       - binutils-dev
 
 rust:
-  - 1.7.0  # Oldest supported version
+  - 1.9.0  # Oldest supported version
   - stable
   - beta
   - nightly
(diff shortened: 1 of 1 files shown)


fiveop:
The release of [gcc-rs](https://github.com/alexcrichton/gcc-rs) 0.3.39 results in 1.7 builds breaking again. Since, I am a bit lost here, could you, @posborne, point me towards where I should be looking for the build failures that happen here. To me, it looks like 'infrastructure' problems (wrong files for old versions being requested). Given that you are more involved in the *odd* Platforms, you might have some insights, whose typing out does not require to much time.

posborne:
At present, Rust 1.7.0 is baked into the docker images we use for doing builds for other targets.  Some notes on how our current system works as well as a few others here: https://github.com/rust-embedded/rfcs/issues/16.  I would  like to move to being based on https://github.com/japaric/trust by @japaric.

I'm not 100% sure on the best way to reuse those contents are present (copying, git submodule, something else).  I've also been pretty busy lately with the day job, so my FOSS time has been decreased.

fiveop:
More and more dependencies don't run on older versions, so we have to update the minimal version. Since you do not have the time at the moment and I do not want to spend the time to get into working with these platforms, I propose that we drop support for those platforms for now. 

I propose to support only beta, stable, stable-1 and stable-2 (and work toward nightly support) on Linux/ MacOSX (x86/x86_64).

posborne:
> I propose that we drop support for those platforms for now.

We definitely should support these platforms, but in order to fix the build, I am fine with moving these targets to the ignored failures in travis.

fiveop:
I'll close this for now, since I won't be working on it.

PR #473, opened 2016-11-19
Allow receiving multiple FD over ControlMessage in unix sockets

### LINKED ISSUE #464: Passing multiple FD over an Unix Socket in a single Control Message doesn't work
For some reason, passing multiple FD over a Unix Domain Socket doesn't seem to work. Below is some test code that is supposed to send two fd over a Unix Socket along with 8 bytes of data. For some reason, only one FD is sent though, the other is lost in the void.

```rust
extern crate nix;

use std::os::unix::net::UnixDatagram;
use std::os::unix::io::{RawFd, AsRawFd};
use nix::sys::socket::{CmsgSpace, ControlMessage, MsgFlags, sendmsg, recvmsg};
use nix::sys::uio::IoVec;


fn main() {
    let (send, receive) = UnixDatagram::pair().unwrap();
    let thread = std::thread::spawn(move || {
        let mut buf = [0u8; 8];
        let iovec = [IoVec::from_mut_slice(&mut buf)];
        let mut space = CmsgSpace::<[RawFd; 2]>::new();
        let (paneid, fds) = match recvmsg(receive.as_raw_fd(), &iovec, Some(&mut space), MsgFlags::empty()) {
            Ok(msg) => {
                let mut iter = msg.cmsgs();
                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
                    (iovec[0].as_slice(), fds.to_vec())
                } else {
                    panic!();
                }
            },
            Err(_) => {
                panic!();
            }
        };
        println!("paneid = {:?}, fds = {:?}", paneid, fds);
    });

    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
    let iov = [IoVec::from_slice(&slice)];
    let arr = [0, 1]; // Pass stdin and stdout
    let cmsg = [ControlMessage::ScmRights(&arr)];
    sendmsg(send.as_raw_fd(), &iov, &cmsg, MsgFlags::empty(), None).unwrap();
    thread.join().unwrap();
}
```

The output is `paneid = [1, 2, 3, 4, 5, 6, 7, 8], fds = [6]`

roblabla:
Fixes #464

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -236,7 +236,6 @@ pub fn test_syscontrol() {
 }
 
 #[test]
-#[ignore]
 fn test_scm_rights_multiple_fd() {
     use std::os::unix::net::UnixDatagram;
     use std::os::unix::io::{RawFd, AsRawFd};
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -234,3 +234,91 @@ pub fn test_syscontrol() {
     // requires root privileges
     // connect(fd, &sockaddr).expect("connect failed");
 }
+
+#[test]
+#[ignore]
+fn test_scm_rights_multiple_fd() {
+    use std::os::unix::net::UnixDatagram;
+    use std::os::unix::io::{RawFd, AsRawFd};
+    use std::thread;
+    use nix::sys::socket::{CmsgSpace, ControlMessage, MsgFlags, sendmsg, recvmsg};
+    use nix::sys::uio::IoVec;
+
+    let (send, receive) = UnixDatagram::pair().unwrap();
+    let thread = thread::spawn(move || {
+        let mut buf = [0u8; 8];
+        let iovec = [IoVec::from_mut_slice(&mut buf)];
+        let mut space = CmsgSpace::<[RawFd; 2]>::new();
+        let (data, fds) = match recvmsg(receive.as_raw_fd(), &iovec, Some(&mut space), MsgFlags::empty()) {
+            Ok(msg) => {
+                let mut iter = msg.cmsgs();
+                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
+                    (iovec[0].as_slice(), fds.to_vec())
+                } else {
+                    panic!();
+                }
+            },
+            Err(_) => {
+                panic!();
+            }
+        };
+        assert_eq!(data, [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8]);
+        assert_eq!(fds.len(), 2);
+    });
+
+    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
+    let iov = [IoVec::from_slice(&slice)];
+    let arr = [0, 1]; // Pass stdin and stdout
+    let cmsg = [ControlMessage::ScmRights(&arr)];
+    sendmsg(send.as_raw_fd(), &iov, &cmsg, MsgFlags::empty(), None).unwrap();
+    thread.join().unwrap();
+}
+
+// Still buggy. See https://github.com/nix-rust/nix/issues/756
+#[test]
+#[ignore]
+fn test_scm_rights_multiple_cmsg() {
+    use std::os::unix::net::UnixDatagram;
+    use std::os::unix::io::{RawFd, AsRawFd};
+    use std::thread;
+    use nix::sys::socket::{CmsgSpace, ControlMessage, MsgFlags, sendmsg, recvmsg};
+    use nix::sys::uio::IoVec;
+
+    let (send, receive) = UnixDatagram::pair().unwrap();
+    let thread = thread::spawn(move || {
+        let mut buf = [0u8; 8];
+        let iovec = [IoVec::from_mut_slice(&mut buf)];
+        let mut space = CmsgSpace::<([RawFd; 1], CmsgSpace<[RawFd; 1]>)>::new();
+        let (data, fds) = match recvmsg(receive.as_raw_fd(), &iovec, Some(&mut space), MsgFlags::empty()) {
+            Ok(msg) => {
+                let mut iter = msg.cmsgs();
+                let fdone;
+                let fdtwo;
+                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
+                    fdone = fds[0];
+                } else {
+                    panic!();
+                }
+                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
+                    fdtwo = fds[0];
+                } else {
+                    panic!();
+                }
+                (iovec[0].as_slice(), [fdone, fdtwo])
+            },
+            Err(_) => {
+                panic!();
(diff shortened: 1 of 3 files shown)


fiveop:
Thanks, looks good to me. Could you add an entry to the CHANGELOG.md and write a test based on your failing example in #464?

roblabla:
There's still a problem with the second test (the one that sends multiple cmsgs). I believe this is due to a mistake in the way space/len of CmsgSpace is implemented (although I'm uncertain).

roblabla:
Nevermind my previous comment. I think the problem comes from https://github.com/nix-rust/nix/blob/80a34a57313519272428f3f19f54b62261ee8001/src/sys/socket/mod.rs#L241. This creates a new pointer, writes the FD in it, but the "old" buf pointer isn't incremented by the size of the FD. 

This means that when writing the second control message, it will overwrite the data of the first control message. I am unsure of how to fix this though, the code is pretty hard to follow and I'm unsure of how it interacts with lifetimes and all.

test/sys/test_socket.rs
```diff
@@ -180,3 +180,88 @@ pub fn test_unixdomain() {
 
     assert_eq!(&buf[..], b"hello");
 }
+
+#[test]
+fn test_scm_rights_multiple_fd() {
+    use std::os::unix::net::UnixDatagram;
```
fiveop: This import breaks ARM and 1.7.0 builds.

fiveop:
I now have an idea what the problem is. `copy_bytes` used by `encode_into` called in `sendmsg` updates `dst` so that it points after the bytes just copied into. This works for the `cmsghdr`. For the second `copy_bytes`, however, another slice `buf2` is used (which probably gets updated), thus `buf` (in `encode_into`) or `cmsg_buffer` (in `sendmsg`) is not updated. When two `ControlMessage`s are encoded, this is a problem.

kamalmarhubi:
@fiveop did your #489 fix this issue? If so, should we close this PR?

fiveop:
No. It was a step toward fixing it.

Susurrus:
What's necessary to wrap this PR up and get everything merged?

roblabla:
Well, I rebased the commits on the latest master, which makes this mergeable again. Worth noting that I ran the "Add some tests" commit locally (independently of the fix) to make sure the bugs were still present. Indeed they are.

After my fix, test_scm_rights_multiple_cmsg still fails though, and test_scm_rights_multiple_fd segfaults on armv7 and aarch64

Susurrus:
@roblabla Thanks for updating this!

Could you add some more details to the "Add some tests" commit and also some comments to the tests summarizing basically the discussion here for some context.

The segfaults are likely QEMU bugs as those are run on that. You can just ignore those tests for those targets.

But why is the "fixed" test still failing? Is the test not correct or is your "fix" not actually the correct fix? You didn't offer any commentary on that here, would you like a second pair of eyes on this or do you know how to fix it?

roblabla:
My main use when I made the PR was to fix my use-case, which is multiple FD. But while testing, I discovered there really were two bugs (depending on how you achieve multiple FD, with one cmsg containing multiple FD, or multiple cmsg containing each one FD).

I fixed one bug (multiple FD, one CMSG), but never quite managed to figure out the source of the other (multiple CMSG).

So the test is right (at least as far as I can see), but there is still a bug in nix somewhere. 

I'll add some details to the commit as soon as I'm on a work machine.

lucab:
Bump. Is there any way to make progress and eventually land this?

Susurrus:
@lucab I think we're basically waiting to resolve the second bug, which has not been resolved if I'm following this correctly.. If you're willing to dive into this, your help would be greatly appreciated and we could get this merged. @roblabla Can you confirm this is the current state of things and maybe provide any more context that's missing if you aren't able to continue to push this forward?

roblabla:
Well, this PR basically fixes what I needed fixed (I needed the ability to send multiple FD over one message, which is fixed), and I don't have the time to figure out what the other bug is.

IIRC, there's a problem in the way we encode multiple Control Messages which causes the kernel to only consider the first Control Message sent. The specifics of where the bug is, I have no idea. I mean, what needs to be done is take the failing test and figure out what's wrong with it.

In the meantime, maybe I could set the second test as ignored so we can land the first fix, and make an issue about the second bug so we can track further progress there ?

Susurrus:
@roblabla Yes, can you cut out the second test into it's own issue (someone can then take it and make a PR to fix it). And then we can merge your single fix and test case for it.

Susurrus:
This could also use a rebase and the CHANGELOG entry needs to be moved up into the current Unreleased section.

Susurrus:
@roblabla Just wanted to ping you on this as I'd love to get it merged!

roblabla:
Thanks for reminding me ! It had slipped from my mind ^^'.

Should I still remove the UnixDatagram stuff from the test ? Those types were added in 1.10, AFAIK we support minimum of 1.13 ?

Susurrus:
I think the `UnixDatagram` stuff is the bug we haven't figure out yet, yes? I assume at this point it isn't a wrong value in a constant. So can you create a separate issue for that and copy/paste the test code there and give as much context as you can towards the error?

roblabla:
I'm not sure we understood eachother ? I meant that I'm using `std::net::unix::UnixDatagram` in the tests, which [fiveop told me would break ARM 1.7 builds way back when](https://github.com/nix-rust/nix/pull/473#discussion_r89200708), so I was wondering if the comment was still valid - that is, if I should be using the nix functions to create a UnixDatagram (which I find less convenient).

#756 contains the details of what I know about the remaining bug.

Susurrus:
@roblabla We aim to support Rust 1.13, so if the tests pass on all supported platforms (which test the minimum supported Rust version), then we should be fine!

I saw #756, which does a good job detailing the other bug you found, so thanks for that.

And Travis finished but found a few failures, so those'll need to be addressed (I haven't looked at them).

roblabla:
So, it looks like it segfaults basically every single qemu instance. Besides, I also have failures on IOS and MacOS, but the output is empty 😕.

Should I just `#cfg[]` the tests out ?

No output :
- Mac : https://travis-ci.org/nix-rust/nix/jobs/270781266
- IOS : https://travis-ci.org/nix-rust/nix/jobs/270781262

List of qemu-related failures :
- aarch64-unknown-linux-gnu
- arm-unknown-linux-gnueabi
- armv7-unknown-linux-gnueabihf
- mips-unknown-linux-gnu
- mips64-unknown-linux-gnuabi64
- mips64el-unknown-linux-gnuabi64
- mipsel-unknown-linux-gnu
- powerpc64-unknown-linux-gnu
- powerpc64le-unknown-linux-gnu

Susurrus:
I guess so. Add a clear comment and use the `travis` feature with `#cfg` to ignore them.

Susurrus:
We would ideally file a bug with upstreaam and include that in the comment about why we exclude it.

Susurrus:
It's been about 3 months since there's been any movement on this, so I'm going to go ahead and close this.

PR #474, opened 2016-11-19
Fix multiple issues with Unix domain sockets

asomers:


src/sys/socket/addr.rs
```diff
@@ -502,7 +510,7 @@ impl SockAddr {
         match *self {
             SockAddr::Inet(InetAddr::V4(ref addr)) => (mem::transmute(addr), mem::size_of::<libc::sockaddr_in>() as libc::socklen_t),
             SockAddr::Inet(InetAddr::V6(ref addr)) => (mem::transmute(addr), mem::size_of::<libc::sockaddr_in6>() as libc::socklen_t),
-            SockAddr::Unix(UnixAddr(ref addr, len)) => (mem::transmute(addr), (len + mem::size_of::<libc::sa_family_t>()) as libc::socklen_t),
```
asomers: The old calculation only worked on Linux.  On FreeBSD (and probably OSX, but I don't have an OSX machine to test on), bind()ing a SockAddr created this way would chop off the last character.

src/sys/socket/mod.rs
```diff
@@ -627,7 +633,9 @@ pub unsafe fn sockaddr_storage_to_addr(
             Ok(SockAddr::Inet(InetAddr::V6((*(addr as *const _ as *const sockaddr_in6)))))
         }
         consts::AF_UNIX => {
-            Ok(SockAddr::Unix(UnixAddr(*(addr as *const _ as *const sockaddr_un), len)))
```
asomers: The old code forgot to consider the offset of sun_path.  The result was that getsockname would appear to return a path with one or two extra null bytes at the end.

fiveop:
Everything looks good. Commit message, CHANGELOG entry, tests, documentation. Thank you!

@homu r+

PR #478, opened 2016-11-25
Add support for system control sockets for XNU

conradev:
I added support for macOS and iOS system sockets, which can be used to control the kernel as described [here](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html).

To do this, I had to add in support for `ioctl` on those platforms, so I added in `ioctl` support for all BSD-based platforms. The API seems to be the same between [xnu](https://opensource.apple.com/source/xnu/xnu-3248.60.10/bsd/sys/ioccom.h.auto.html), [FreeBSD](https://github.com/freebsd/freebsd/blob/master/sys/sys/ioccom.h), [NetBSD](https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/sys/ioccom.h), [OpenBSD](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/ioccom.h?rev=1.5&content-type=text/x-cvsweb-markup) and [Dragonfly BSD](http://gitweb.dragonflybsd.org/dragonfly.git/blob/HEAD:/sys/sys/ioccom.h).

I added a test that runs on macOS and iOS for the functionality. Let me know if I need to make any changes!


conradev:
Do you think you'll be able to review this, @asomers?

src/sys/ioctl/mod.rs
```diff
@@ -145,3 +125,75 @@ extern "C" {
...
+            $crate::Errno::result($w)
+        }
+    );
+}
+
+/// Declare a wrapper function around an ioctl.
+#[macro_export]
+macro_rules! ioctl {
+    (bad $name:ident with $nr:expr) => (
+        pub unsafe fn $name(fd: $crate::sys::ioctl::libc::c_int,
```
asomers: This seems to be overloading the "bad" keyword.  AFAICT, in Linux the difference between _IOR and _IOR_BAD is in the interpretation of the size argument.  However, the bad macro here deals with callers who hard-code the ioctl number.  I know @conradev didn't introduce the bad keyword, but it would be nice to fix

src/sys/ioctl/platform/bsd.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub const INOUT: u32 = (IN|OUT);
+    pub const IOCPARM_MASK: u32 = 0x1fff;
+}
+
+pub use self::consts::*;
+
+#[macro_export]
+macro_rules! ioc {
+    ($inout:expr, $group:expr, $num:expr, $len:expr) => (
+        $inout | (($len as u32 & $crate::sys::ioctl::IOCPARM_MASK as u32) << 16) | (($group as u32) << 8) | ($num as u32)
```
asomers: The second `as u32` is superfluous, since `IOCPARAM_MASK` is already defined as `u32`.

src/sys/mod.rs
```diff
@@ -12,7 +12,8 @@ pub mod eventfd;
 #[cfg(target_os = "linux")]
 pub mod memfd;
 
-#[cfg(not(any(target_os = "ios", target_os = "freebsd", target_os = "dragonfly")))]
+#[cfg(not(any(target_os = "freebsd", target_os = "dragonfly")))]
```
asomers: It's fine to get rid of the "freebsd" here.  Probably fine to get rid of dragonfly, too, but I can't test on that platform.

conradev:
Made those changes! Let me know if you have further feedback :)

fiveop:
Could you rebase to master and push again (with -f), so that the CI is green? Could you also add an entry to our CHANGELOG.md to document the changes.

After that we can merge it.

conradev:
Rebased against master. I documented the additions, and the one existing API change!

fiveop:
@homu r+

thanks

PR #483, opened 2016-12-03
Add POSIX AIO support

asomers:
POSIX AIO is a standard for asynchronous file I/O.  Read, write, and
fsync operations can all take place in the background, with completion
notification delivered by a signal, by a new thread, by kqueue, or not
at all.

The SigEvent class, used for AIO notifications among other things, is
also added.

### DIFF
--- a/src/sys/aio.rs
+++ b/src/sys/aio.rs
@@ -1,0 +1,247 @@
+use {Error, Errno, Result};
+use std::os::unix::io::RawFd;
+use libc::{c_void, off_t, size_t};
+use libc;
+use std::mem;
+use std::ptr::{null, null_mut};
+use sys::signal::*;
+use sys::time::TimeSpec;
+
+/// Mode for `aio_fsync`.  Controls whether only data or both data and metadata
+/// are synced.
+#[repr(i32)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AioFsyncMode {
+    /// do it like `fsync`
+    O_SYNC = libc::O_SYNC,
+    /// on supported operating systems only, do it like `fdatasync`
+    #[cfg(any(target_os = "openbsd", target_os = "bitrig",
+              target_os = "netbsd", target_os = "macos", target_os = "ios",
+              target_os = "linux"))]
+    O_DSYNC = libc::O_DSYNC
+}
+
+/// When used with `lio_listio`, determines whether a given `aiocb` should be
+/// used for a read operation, a write operation, or ignored.  Has no effect for
+/// any other aio functions.
+#[repr(i32)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LioOpcode {
+    LIO_NOP = libc::LIO_NOP,
+    LIO_WRITE = libc::LIO_WRITE,
+    LIO_READ = libc::LIO_READ
+}
+
+/// Mode for `lio_listio`.
+#[repr(i32)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LioMode {
+    /// Requests that `lio_listio` block until all requested operations have
+    /// been completed
+    LIO_WAIT = libc::LIO_WAIT,
+    /// Requests that `lio_listio` return immediately
+    LIO_NOWAIT = libc::LIO_NOWAIT,
+}
+
+/// Return values for `aio_cancel`
+#[repr(i32)]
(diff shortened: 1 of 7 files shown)


asomers:
All failures are actually in the dependent gcc crate, which no longer compiles with rust 1.7.0.

asomers:
Could I please get some review of this PR?

src/sys/aio.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// `opcode` set to `LIO_READ`).  This method is useful when writing a const
+    /// buffer with `aio_write`, since from_mut_slice can't work with const
+    /// buffers.
+    // Note: another solution to the problem of writing const buffers would be
+    // to genericize AioCb for both &mut [u8] and &[u8] buffers.  aio_read could
+    // take the former and aio_write could take the latter.  However, then
+    // lio_listio wouldn't work, because that function needs a slice of AioCb,
+    // and they must all be the same type.  We're basically stuck with using an
+    // unsafe function, since aio (as designed in C) is an unsafe API.
+    pub unsafe fn from_slice(fd: RawFd, offs: off_t, buf: &[u8],
```
posborne: With the version of these functions that take references to buffers, I wonder if it would be reasonable to do something to ensure that the lifetime of the reference is at least as long as the lifetime of the `AioCb`.  Without this, there buffer could go out of scope in the calling context and we would be working with an invalid pointer.

I think this would typically be done with something like `PhantomData` (when not storing the reference itself).

src/sys/aio.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+/// An asynchronous version of `fsync`.
+pub fn aio_fsync(mode: AioFsyncMode, aiocb: &mut AioCb) -> Result<()> {
+    let p: *mut libc::aiocb = &mut aiocb.aiocb;
+    Errno::result(unsafe { libc::aio_fsync(mode as ::c_int, p) }).map(drop)
+}
+
+/// Asynchously reads from a file descriptor into a buffer
```
posborne: nit: Typo with "Asynchronously"

posborne:
Hi @asomers.  Deep apologies for the delay in getting you a review on this.  Having AIO support in nix will be a wonderful addition and this looks like it will work great.

I think my main concern right now is with the validity of buffers we are using for doing asynchronous reads/writes and ensuring that the lifetime of the buffer is at least as long as the asynchronous operation -- Without this, I think the kernel could easily end up stomping on memory (which would result in all manner of bad things).  Is this concern warranted?

I have mixed feelings on the AIO subsystem in general (my experience with Linux only) and am not sure where it is headed.  That being said, there are cases where it is beneficial for it to be used today and will be happy to accept this change (once we discuss a bit more).

asomers:
Point taken, @posborne .  I'll try to fix the buffer lifetime issue.  Regarding the AIO subsystem: I can understand why you're skeptical, because it's not well implemented on Linux.  On Linux, POSIX AIO performs poorly and the nonstandard libaio performs well.  However, in the BSD world, POSIX AIO performs well and there is no libaio.

Unfortunately, libaio does not have a compatible API at all.  The differences are even worse for higher level libraries.  mio and tokio-core basically need completely different implementations to use posix AIO vs libaio.

But POSIX AIO _does_ exist and work on Linux, and it works well on the BSDs, so I think it's worth including in nix.

posborne:
Thanks for the summary @asomers and points well taken.  I agree that exposing the Posix AIO interface (even if not performant with the glibc userspace implementation on linux) is the right thing to do in nix.  Mio or other libraries should be the ones to decide what abstraction to use on which host operating systems.

posborne:
Excellent, feeling better about this now that we track lifetimes.

I think it would be good to consider the following additional improvements in the future.  Let me know your thoughts:

1. Rather than having freestanding functions like `aio_read`, move these to be on the callback as you need to pass that in as a parameter anyway and it creates a more fluent API.
2. Consider automatic cancellation or other action on drop of an `AioCb`.  Right now, we are able to "leak" these very easily.

@homu r+

asomers:
Good idea regarding class vs instance methods, @posborne.  I'll take a swing at it.  I think there's also some runtime safety checks we could add, if we're willing to expand the `AioCb` structure.  For example, at runtime we could prohibit calling aio_read on an `AioCb` created with `from_slice`.  That kind of check can't be performed at compile time without breaking `lio_listio`.

posborne:
@asomers I think the runtime check sounds reasonable.

Issue #480, opened 2016-12-01
How to pass a null ptr to some functions?


hicqu:
look at this function:

`fn epoll_ctl(epfd: RawFd, op: EpollOp, fd: RawFd, event: &mut EpollEvent)`

if EpollOp == EpollDel, the event could be a NULL in C. But how to pass a NULL ptr to epoll_ctl in rust?

fiveop:
This is a missing feature of our interface. We had a similar problem with `kill` (see #445). Feel free to create a PR to fix this. We will incorporate it, even if it is a breaking change. We should probably check in the new implementation of `epoll_ctl`, whether `event` may be `None` (or `NULL` in C) depending on the other arguments.

hicqu:
OK, I think we can define these two functions in mio. Here is a simple test:

```
/// maybe we could accept these two functions into mio?
fn null_as_mut<T>() -> &'static mut T {
    unsafe { &mut *::std::ptr::null_mut::<T>() }
}

fn null_as_ref<T>() -> &'static T {
    unsafe { & *::std::ptr::null::<T>() }
}

/// simple test below

#[derive(Debug)]
struct T { f: i32, }

fn ref_t(t: &T) {
    // if pass nullptr, segment fault as expected.
    println!("{:?}", *t);   
}

fn mut_t(t: &mut T) {
    // if pass nullptr, segment fault as expected.
    println!("{:?}", *t);   
}

fn main() {
    let mut t: T = T { f: 100 };
    ref_t(&t);  // it's OK.

    mut_t(null_as_mut::<T>());  // segment fault.
    ref_t(null_as_ref::<T>());  // segment fault.
}
```

What do you think?

PR #484, opened 2016-12-03
Fix for #480 and add one test case for that.

hicqu:
r? @fiveop, thank you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Added struct `TimeSpec`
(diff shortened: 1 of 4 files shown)


hicqu:
why all checks have failed? It a big hit to me because it's the first time I make a pull request on github.
Please help me, thanks. :)

fiveop:
We don't need the pointer utilities. If you look at #445, you see that we solved the problem in the public interface by changing the parameter type of `signal` to `T: Into<Optional<Signal>>`, so that callers can specify `None`, if they don't want to provide a signal (or in this case no event).

I suggest (again), that you look at the changes made in #445 and base your PR on that.

Thank you for the PR.

hicqu:
Ok, I got that. I will use Option<&mut SockAddr> instead of pointer magic. New pull request will be later.

PR #486, opened 2016-12-08
fix #480 and add simple test cases for that.

### LINKED ISSUE #480: How to pass a null ptr to some functions?

hicqu:
r? @fiveop


src/sys/epoll.rs
```diff
@@ -72,9 +87,10 @@ pub fn epoll_create1(flags: EpollCreateFlags) -> Result<RawFd> {
 }
 
 #[inline]
-pub fn epoll_ctl(epfd: RawFd, op: EpollOp, fd: RawFd, event: &mut EpollEvent) -> Result<()> {
-    let res = unsafe { libc::epoll_ctl(epfd, op as c_int, fd, &mut event.event) };
-
+pub fn epoll_ctl<'a, T>(epfd: RawFd, op: EpollOp, fd: RawFd, event: T) -> Result<()>
+    where T: Into<&'a mut EpollEvent>
+{
+    let res = unsafe { libc::epoll_ctl(epfd, op as c_int, fd, &mut event.into().event) };
```
fiveop: In order for this to be safe, we should check, if `op` is `EPOLL_CTL_DEL` when `event` is NONE. Otherwise we might pass a null pointer into `libc::epoll_ctl`, when it expects something else.

src/sys/epoll.rs
```diff
@@ -44,10 +45,14 @@ pub struct EpollEvent {
 }
 
 impl EpollEvent {
-    pub fn new(events: EpollFlags, data: u64) -> EpollEvent {
+    pub fn new(events: EpollFlags, data: u64) -> Self {
         EpollEvent { event: libc::epoll_event { events: events.bits(), u64: data } }
     }
 
+    pub fn empty() -> Self {
```
fiveop: Can you give me an example of a use case for this function?

hicqu:
EpollEvent::empty() is a try for simplify this case:
```
let mut events: [EpollEvent; 128] = [EpollEvent::new(EpollFlags::empty(), 0 as u64); 128];
epoll_wait(epoll_fd, &mut events, -1);
```
Do you think that's a good idea? Maybe I should move that to another issue for more deep discussion.

As for another question, we should return an error when call `epoll_ctl(epoll_fd, EpollOp::EpollCtlAdd, None)` or just `panic!()`? I think return an error is a good idea. So which error should be return do you think?

hicqu:
Sorry for forgeting to edit CHANGLOG.  If all design are reviewed and OK, I will complete that.

fiveop:
I would return `EINVAL` in case op and event don't fit together. But we need to document that properly on that function.

hicqu:
Got it. I will fix all and make pull request again. Thank you!

hicqu:
I think everything is OK. Please check it again. Thanks!

hicqu:
Anything wrong?

fiveop:
No. I have not looked into this project for a while. Sorry.

@homu r=fiveop

PR #488, opened 2016-12-10
Allow failures of 1.7 builds

fiveop:
Libraries we depend on need newer versions of Rust. The current
stable release is 1.13.

This also disables the CI for the platforms/architectures MIPS, ARM
and Android. This cannot be helped as long as the test infrastructure
only runs on 1.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,9 @@ matrix:
       sudo: true
   allow_failures:
     - rust: nightly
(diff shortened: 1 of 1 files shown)


fiveop:
I take [this](https://github.com/nix-rust/nix/pull/471#issuecomment-264685756) comment as approval.

@homu r=@posborne

PR #489, opened 2016-12-10
Fix ControlMessage::encode_into when encoding multiple messages

fiveop:
copy_bytes updates dst so that it points after the bytes that were just
copied into it. encode_into did not advance the buffer in the same way
when encoding the data.

See #473


fiveop:
@homu r=fiveop

PR #490, opened 2016-12-11
Ensure child stack passed to clone is 16 byte aligned.

kevindejong:
The current implementation assumes that the array passed by the caller is word aligned (which I don't think Rust guarantees for [u8]) and a multiple of the word size.


fiveop:
I am not sure whether there are any platforms that need less then 16 byte alignment, however all should require a factor of 16, so this should at least not break anything and certainly improve things. (However, I think bionic on Android already does this in clone itself, for example.)

@homu +

fiveop:
@homu r+

PR #492, opened 2016-12-29
fix build for eventfd feature

domsj:


fiveop:
Thanks

@homu r=fiveop

PR #496, opened 2017-01-10
Always use libc's socketaddr_storage

fiveop:
Since rust-lang/rust#23425 is closed, we no longer need the special
case.


fiveop:
@homu r=@fiveop

PR #497, opened 2017-01-17
unistd: add fchdir(2)

lucab:
This introduces a wrapper for fchdir(2), allowing a process to change
directory based on an open file descriptor.

The underlying function is available in libc crate since 0.2.20.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#457](https://github.com/nix-rust/nix/pull/457))
 - Added `getpgrp` in `::nix::unistd`
   ([#491](https://github.com/nix-rust/nix/pull/491))
+- Added `fchdir` in `::nix::unistd`
+  ([#497](https://github.com/nix-rust/nix/pull/497))
 
 ### Changed
 - `epoll_ctl` now could accept None as argument `event`
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -249,6 +249,19 @@ pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Change the current working directory of the process to the one
+/// given as an open file descriptor (see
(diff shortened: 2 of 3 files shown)


lucab:
Note: this has been recently added in libc crate (https://github.com/rust-lang/libc/pull/496), and it's only available since 0.2.20 (released today).

posborne:
Excellent, thanks for the contribution @lucab.  Can you add an entry to the changelog for this addition?

posborne:
Also, it would be good to understand what might be happening with this failure.  We currently have some known issues with aarch64 which is why it is not gating the green check, but the failure does seem suspicious:

* aarch64: https://travis-ci.org/nix-rust/nix/jobs/192841376#L591
* arm-unknown-linux-gnueabihf: https://travis-ci.org/nix-rust/nix/jobs/192841377#L588
* mips: https://travis-ci.org/nix-rust/nix/jobs/192841378#L1335

It would appear that the system call is not working as desired in the test environment for these targets.

lucab:
Oops, I forgot the changelog, added now.

Regarding the cross-test failure, I think I'm hitting a weird combination of an old kernel on travis + some chroot/mount-ns trick by docker + (likely) some bug on aufs/overlayfs. `fchdir()` works, but the original root-fd seems to be leading somewhere else in the FS, possibly the lowerdir of a stacked fs. I've reworked the test to avoid grabbing a root-fd and use a self-contained tmpdir instead, now waiting for travis result of this (but it looks like it is not picking it up).

posborne:
@lucab Looks like macos has an issue that needs to be resolved in the tests now.

lucab:
@posborne ack. It turned out the test was not canonicalizing path, so it was fooled by symlinks. `test_fchdir` is now passing everywhere, despite some other (I guess unrelated) failures on cross-tests. PTAL.

posborne:
Well, that was fun :1st_place_medal:.

@homu r+

lucab:
@posborne great, thanks! Please ping here back once this gets into a released crate, so I can drop some private wrappers I have in a project which needs this feature.

PR #501, opened 2017-01-27
O_TMPFILE is includes O_DIRECTORY

Idolf:
At least does in in the linux kernel:

http://lxr.free-electrons.com/source/include/uapi/asm-generic/fcntl.h#L92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#491](https://github.com/nix-rust/nix/pull/491))
 - Added `fchdir` in `::nix::unistd`
   ([#497](https://github.com/nix-rust/nix/pull/497))
+- Corrected the value of `O_TMPFILE`.
+  ([#501](https://github.com/nix-rust/nix/pull/501))
 
 ### Changed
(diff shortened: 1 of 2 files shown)


posborne:
@Idolf  Since this code is Linux/Android specifc, I have no problem with the change.  Android is still on a 3.10 kernel on many systems, but I'm fine with the value returning an EINVAL on older kernels (this doesn't change that behavior).  Can you add a note on this to the CHANGELOG.md file?

Idolf:
Changelog update.

As far as I can tell, the value of `O_TMPFILE` has always been defined as `0o20000000 | O_DIRECTORY`, however I am just now noticing that the value of `O_DIRECTORY` is not architecture independent. :anguished:

I don't think that this fix will be sufficient.

posborne:
> I don't think that this fix will be sufficient.

`libc` defines `O_DIRECTORY` in an architecture-specific way, so we should move to using/reusing that value as we do elsewhere in nix.  We still want the enumerations we provide as part of the nix safety story, but we should use the actual value from libc.

The `O_TMPFILE` constant also really belongs in libc and not nix (obviously that hasn't been changes in this code as of yet at all and would require some contributions to libc).  If you are up for it, I would love a volunteer to help make this happen.

PR #508, opened 2017-02-13
Implement major/minor macros, correct mkdev

aidanhs:
It appears that the previous `mkdev` was based on the kernel headers (https://github.com/torvalds/linux/blob/v4.7/include/linux/kdev_t.h#L6) which (I guess) is the internal kernel dev_t. Scrolling down the file you can see some bitshifting operations to do conversions.

The new implementation(s) are based on [musl](http://git.musl-libc.org/cgit/musl/tree/include/sys/sysmacros.h?id=dbbb3734d8c0176feabd6c46e2e85bbc3b8a60af) and [glibc](https://github.molgen.mpg.de/git-mirror/glibc/blob/20003c49884422da7ffbc459cdeee768a6fee07b/sysdeps/unix/sysv/linux/sys/sysmacros.h#L38), which are in agreement about how dev_t should be handled.

(as it happens I suspect we could omit the shift by 32 since I don't see that in the kernel headers, but doesn't hurt to take the conservative route and mimic the libcs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#491](https://github.com/nix-rust/nix/pull/491))
 - Added `fchdir` in `::nix::unistd`
   ([#497](https://github.com/nix-rust/nix/pull/497))
+- Added `major` and `minor` in `::nix::sys::stat` for decomposing `dev_t`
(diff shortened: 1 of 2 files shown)


posborne:
There appear to be some issues with your current implementation; in general the changes seem reasonable.

aidanhs:
Sorry, thought it was spurious for some reason. Fixed it now.

posborne:
@homu r+

Issue #494, opened 2017-01-07
From<std::time::Duration> for nix::sys::time::TimeVal?


Roguelazer:
It might be more ergonomic to be able to pass `std::time::Duration` objects in place of timevals. Reasonable to send in a PR for?

fiveop:
Not as a replacement, but implementing `TimeValLike` for `std::time::Duration` is probably a good idea.

Please do make a PR.

PR #511, opened 2017-02-15
implement TimeValLike for std::time::Duration

Roguelazer:
This was suggested by @fiveop in #494

I'm not sure how you want me to handle the mismatch between `i64` everywhere in this api and `u64` in Duration; should I panic if someone passes a negative duration?


kamalmarhubi:
Shall hold off on this while #516 gets decided on? Thanks for pushing these changes though!

Susurrus:
I'm going to close this in favor of whatever comes out of #516, which I think we have on a good track. I don't think this implementation will be required.

PR #512, opened 2017-02-15
impl from<Duration> for TimeVal

### LINKED ISSUE #494: From<std::time::Duration> for nix::sys::time::TimeVal?

Roguelazer:
This should make working with various functions that take `TimeVal`s more ergonomic

Resolves #494.

### DIFF
--- a/src/sys/time.rs
+++ b/src/sys/time.rs
@@ -395,7 +395,8 @@ impl TimeVal {
 impl From<time::Duration> for TimeVal {
     fn from(d: time::Duration) -> Self {
         let micros = d.subsec_nanos() / 1_000;
-        TimeVal(timeval { tv_sec: d.as_secs() as i64, tv_usec: micros as i64})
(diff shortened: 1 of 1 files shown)


posborne:
@Roguelazer See the CI failures.  `i64` is not correct on all platforms for these casts.

Roguelazer:
I'm pretty sure I fixed it, although CI for OS X is hanging trying to set up the build VM.

kamalmarhubi:
Maybe add `Into<Duration>` as well?

kamalmarhubi:
As with #511, shall we hold off on this while #516 gets decided on?

Susurrus:
@kamalmarhubi I agree with your last comment, so let's go ahead and close this in favor of sorting all that out in one single issue.

PR #513, opened 2017-02-16
Fix unresolved name error when building with musl.

yxd-hde:
See: https://git.musl-libc.org/cgit/musl/tree/src/time/timer_create.c#n126


posborne:
Thanks!  @homu r+

posborne:
@homu r+

posborne:
@homu retry

PR #515, opened 2017-02-16
Update bitflags to 0.7

SimonSapin:
Brought to you by GNU sed.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -52,7 +52,7 @@ For example,
 
 ```rust
 libc_bitflags!{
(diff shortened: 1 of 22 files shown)


SimonSapin:
I made this to reduce the number of duplicated crates in Servo (we like to do that to keep the dependency graph form growing faster than it already does), but it turns out nix will probably be removed from the graph when ws updates mio from 0.5 to 0.6 so we likely won’t need this. But since it’s already done, here it is in case you want it.

posborne:
Thanks, change seems good.

posborne:
@homu r+

Issue #514, opened 2017-02-16
Provide baud constants for cfsetispeed and family


berkowski:
`cfgetospeed`, `cfgetispeed`, `cfsetospeed`, and `cfsetispeed` all work with the type `speed_t`, which is aliased to `c_uint`.

However the baud speeds are not their numerical value but system-defined constants.  For example, 9600 on linux is defined as `B9600 = 15`. 

Knowing what these constants are defined as on the target system is required to use these functions, but `nix` provides none and renders these functions unusable without knowledge from outside the crate.

`nix` should either provide aptly named `speed_t` constants that can be used with the baud related functions or an enum type that maps to valid values of `size_t`.

posborne:
Hey @berkowski.  We would definitely prefer to have an enum that matches to the right constants.  Would you be willing to put together a PR for this?

berkowski:
Sure, no problem.

PR #518, opened 2017-02-18
Added BaudRate enum for termios

berkowski:
Issue #514

Does not provide `BaudRate::EXTA` or `BaudRate::EXTB` constants. These seem to alias to `B19200` and `B38400` respectively and so break the 1:1 mapping needed by `From`.  I don't know their historic use.


src/sys/termios.rs
```diff
@@ -293,6 +359,82 @@ mod ffi {
...
+                    10006 => BaudRate::B576000,
+                    10007 => BaudRate::B921600,
+                    10010 => BaudRate::B1000000,
+                    10011 => BaudRate::B1152000,
+                    10012 => BaudRate::B1500000,
+                    10013 => BaudRate::B2000000,
+                    10014 => BaudRate::B2500000,
+                    10015 => BaudRate::B3000000,
+                    10016 => BaudRate::B3500000,
+                    10017 => BaudRate::B4000000,
```
posborne: I'm pretty sure this and many of the other constants you define here are completely incorrect as, looking at the linux kernel source (), the constants there are in octal.  These numbers are decimal and, as such, every constant you reference here greater than 7 is wrong.

Since there is better testing infrastructure for all libc variants in the `libc` crate, I believe we should introduce the constants there and add the type-safe enumeration providing access to this constants in nix.  That has been our basic approach.  Since `libc` uses `ctest`, it will verify the correctness of the constants.

berkowski: Yup, my mistake on the octals.  Good catch, thanks.

I will make the correction and push again while looking into `libc`

kamalmarhubi:
Could we push the constants up to libc? Can happen after this PR.

berkowski:
Working on it (see rust-lang/libc#530).  Just need to fixup a few more issues before it'll be merged.  I'll rework the `nix` side of things after.

kamalmarhubi:
Oh awesome. I was hoping to help but just found confusing facts. :/

berkowski:
Updated to use the newly-landed baud constants from libc (rust-lang/libc#530).

Any thoughts on adding tests to `test/sys/test_terimos.rs`?  The non-x86 ci tests are all pinned to rustc 1.7.0 and are failing due to other reasons.  The actual baud constants aren't getting touched.

posborne:
> Any thoughts on adding tests to test/sys/test_terimos.rs?

No, as with many things testing some of this stuff is a little tricky as there are not great devices in the kernel for doing all sorts of things.  It might be possible to set/get the baud rate on a PTY, but I'm OK with merging this change without that.

src/sys/termios.rs
```diff
@@ -125,6 +125,83 @@ mod ffi {
...
+                    B28800 => BaudRate::B28800,
+                    B57600 => BaudRate::B57600,
+                    B76800 => BaudRate::B76800,
+                    B115200 => BaudRate::B115200,
+                    B230400 => BaudRate::B230400,
+                    #[cfg(any(target_os = "netbsd", target_os = "freebsd"))]
+                    B460800 => BaudRate::B460800,
+                    #[cfg(any(target_os = "netbsd", target_os = "freebsd"))]
+                    B921600 => BaudRate::B921600,
+                    b @ _ => unreachable!("Invalid baud constant: {}", b),
```
posborne: I don't think this arm is necessary.  The compiler should be able to reason about this branch not being reachable and generate a compile error if something is off (which is better than this runtime error).

berkowski: I think it needs to be there to cover all branch cases.  `speed_t` is just an alias to something like `c_uint` so you need a catch-all case at the end to cover all possible values of `c_uint`.

For example, if you remove the similar case for the `linux` group of constants at line 452 you'll get the compile error:

```
   Compiling nix v0.8.0 (file:///home/zac/repos/rust/nix)
error[E0004]: non-exhaustive patterns: `_` not covered
   --> /home/zac/repos/rust/nix/src/sys/termios.rs:420:23
    |
419 |                 match s {
    |                       ^ pattern `_` not covered

error: aborting due to previous error

error: Could not compile `nix`.
```

posborne: Ah, yeah.  You are right.

posborne:
@homu r+

posborne:
@homu retry

Issue #516, opened 2017-02-16
Use Duration for all timeout arguments


Susurrus:
It looks like there may be other work on this area from perusing the existing issues and PRs, so maybe this is covered by that work already, but I wanted to use a Duration (or equivalent) for `poll::poll()`. Looks like there are a number of functions that could use this type treatment:

  * `poll::poll()`
  * `event::kevent()` (though this looks like a backcompat issue)
  * `epoll::epoll_wait()`

I'm not certain if there are rationales for why this hasn't been done already since it looks like some functions have been converted to use `TimeSpec`/`TimeVal` already.

kamalmarhubi:
I think part of the rationale was wanting to support old versions of Rust. `Duration` was only added in 1.3.0, and for a long time we supported back to 1.1.0. I think we now support ~1.7.0 so this could be revisited. @nix-rust/nix-maintainers thoughts?

arcnmx:
If the type is stable in all supported Rust versions, definite +1 to using proper time/duration types as arguments.

posborne:
Yeah, no objection from me.

Susurrus:
This brings up an interesting point, both for users and developers, that the minimum supported rust version isn't explicitly stated anywhere, though your CI only tests Rust 1.7. Can this be added to the readme?

posborne:
> Can this be added to the readme?

That seems reasonable. I do believe we have tried to make a note of any changes to supported version in the changelog.

kamalmarhubi:
1.7 is the current minimum, I believe.

Re taking `Duration`, I'd suggest taking `Into<Duration>` or perhaps `Into<TimeVal>` / `Into<TimeSpec>` as appropriate for the function.

Susurrus:
Why not just use Duration everywhere? #511 shows that if helper methods are desired they can just be implemented for Duration instead of introducing a new type. Or am I missing something and it's inappropriate or unergonomic in some cases?

Susurrus:
@posborne @kamalmarhubi Any comments on this?

posborne:
I agree that taking `Duration` would be more ergonomic and would be in favor of taking `Into<Duration>` (which will work just fine with a `Duration` due to reflexivity).  The only minor downside would be that we might need to do conversions more frequently if a syscall is made frequently (but even then the conversion overhead is approximately nothing compared to the syscall in most cases).

Susurrus:
[...]

PR #520, opened 2017-02-23
Add ppoll()

Susurrus:
This will currently fail CI, as the necessary changes haven't hit libc yet. That is tracked in rust-lang/libc#537. This does build on my computer using the changes tracked on that PR, so I'd appreciate any visual review of this code as it should be "done".

I also wanted to get this submitted so hopefully it'd be in the queue for the 0.8 release.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,3 +1,8 @@
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use sys::time::TimeSpec;
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use sys::signal::SigSet;
+
 use libc;
 use {Errno, Result};
 
@@ -45,5 +50,18 @@ pub fn poll(fds: &mut [PollFd], timeout: libc::c_int) -> Result<libc::c_int> {
                    timeout)
     };
 
+    Errno::result(res)
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn ppoll(fds: &mut [PollFd], timeout: TimeSpec, sigmask: SigSet) -> Result<libc::c_int> {
+
+
+    let res = unsafe {
+        libc::ppoll(fds.as_mut_ptr() as *mut libc::pollfd,
+                    fds.len() as libc::nfds_t,
+                    timeout.as_ref(),
+                    sigmask.as_ref())
+    };
     Errno::result(res)
 }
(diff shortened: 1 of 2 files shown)


src/poll.rs
```diff
@@ -47,3 +50,19 @@ pub fn poll(fds: &mut [PollFd], timeout: libc::c_int) -> Result<libc::c_int> {
 
     Errno::result(res)
 }
+
+pub fn ppoll(fds: &mut [PollFd], timeout: Duration) -> Result<libc::c_int> {
```
kamalmarhubi: can you take `timeout: TimeSpec` until we decide how to resolve #516?

kamalmarhubi: we should accept a sigmask here, too. @fiveop @chaosagent could you suggest how to represent it in nix?

kamalmarhubi: we should accept a sigmask here, too. @fiveop @chaosagent could you suggest how to represent it in nix?

Susurrus:
I wasn't sure if you wanted me to tack on additional commits to fix this or not, so I force-pushed a new version addressing both the sigmask issue (as I thought it could currently be solved) and using `TimeSpec` instead of `Duration`.

Susurrus:
I've pushed a rebased version to re-trigger Travis now that `ppoll()` was merged into `libc`.

kamalmarhubi:
@homu r+

Susurrus:
I'm not certain what's going on with travis and tempfile, which seems to be causing errors, but I'll make sure to only expose this function to non-Mac OSes:
```
```

Susurrus:
So `ppoll` is actually available on more OSes that this would imply, like OpenBSD and FreeBSD both support it through libc. But I don't think nix is worrying about those OSes right now, yes?

Anyways, this should resolve failures on Mac. I still don't know what's up with the other failing platforms, but they're `allow_failure`, so I guess I'll wait on you guys to tell me what, if anything, to do about those.

kamalmarhubi:
Looks like you didn't add your `cfg`? When you do could you also modify the changelog line to specify the OSes it's added for. I'll r+ again after and see if we can't get this past CI!

Susurrus:
@kamalmarhubi Whoops, don't know what happened there! Anyways, I've added the `cfg` and targets to the CHANGELOG.

kamalmarhubi:
thanks!

@homu r+

src/poll.rs
```diff
@@ -1,3 +1,6 @@
+use sys::time::TimeSpec;
```
fiveop: You have to conditional include this and `SigSet`. This is why the build fails on macos.

Susurrus:
Alright, I think I got it this time!

Note that all the Rust 1.7.0 errors are still occuring but we can't tell if they're a result of this change because they're all failing on the `tempfile` crate.

Susurrus:
All targets pass on Travis. This look good to merge?

fiveop:
@kamalmarhubi must still accept the changes. I think his concerns are addressed, but I do not want to press the link 'Dismiss review'.

fiveop:
@homu r+

PR #522, opened 2017-02-24
Specify Rust minimum version requirements

Susurrus:
I think this is important for 2 reasons:

1) Users know if they can use this library
2) Developers know what Rust features they can use in merge requests

What this doesn't do is establish any kind of policy for raising the Rust requirements, which is something  worth addressing in the near-future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ pub unsafe extern fn gethostname(name: *mut c_char, len: size_t) -> c_int;
 pub fn gethostname(name: &mut [u8]) -> Result<()>;
 ```
 
+## Requirements
+Rust >= 1.7.0
+
 ## Usage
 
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Thanks! There's an RFC issue to figure out the policy: https://github.com/nix-rust/rfcs/issues/4

@homu r+

fiveop:
If bitflags 0.7 does not run on rust 1.7 (see all our 1.7 builds breaking in the CI). >= 1.7 is certainly wrong.

Susurrus:
@fiveop Indeed, but that's what I was told! Anyways, I've determined that the min version is actually 1.9 and submitted a PR (#525) to resolve that for the docs and CI.

PR #524, opened 2017-02-25
epoll: Use libc_bitflags for EpollFlags

kamalmarhubi:
This is possible now that EPOLLEXCLUSIVE was added to libc in
https://github.com/rust-lang/libc/pull/526


fiveop:
@homu r+

PR #525, opened 2017-02-25
Update minimum rust version to 1.9.0

Susurrus:
This is largely because of features used in `tempfile` from 1.8 and 1.9. This was tested locally, so we'll see what CI says.

This does bring up the issue of the CI testing in the `ci` folder. That specifies the rust version as 1.7.0, but then when I looked at the docker images they seemed to only support 1.10, so maybe someone can comment on whether anything should be done about those things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ pub fn gethostname(name: &mut [u8]) -> Result<()>;
 ```
 
 ## Requirements
-Rust >= 1.7.0
+Rust >= 1.9.0
 
 ## Usage
 


Susurrus:
Looks like Rust 1.9 fails on ARM targets, which isn't surprising given the comments in the code. Maybe it's time to switch to using `trust` templates to do this testing? Though I don't know how this was supposed to work considering the `mips` targets weren't even supported until [Rust 1.14](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1140-2016-12-22)

posborne:
> That specifies the rust version as 1.7.0, but then when I looked at the docker images they seemed to only support 1.10

Yeah, I built newer images somewhere along the line with 1.10 but never got around to fully updating nix to use the new stuff.  I think the current is still using 1.7 images as it is pointed to `posborne/rust-cross:<platform>`.  The latest stuff from https://github.com/rust-embedded/docker-rust-cross results in images in the form `posborne/rust-cross-<platform>:<version>` being published.  That's not what we point to at the moment.

There are images for several of the platforms pushed for 1.10 now, but I haven't updated nix but it would be great to move to something a bit newer.  Long term, I think I prefer the approach that https://github.com/japaric/trust takes.

Susurrus:
So is there anything I should change in this PR, maybe move all the ARM stuff to use 1.10 and keep the other targets on 1.9?

This still leaves the question about `mips` targets pre-1.14. I don't even understand how they could compile since those targets didn't exist previously.

Susurrus:
Note that `tempfile` has clarified that their minimum supported version is 1.9.0 and they now do CI on that version (see Stebalien/tempfile@0b7e0d573a2351d4f0a6e4ed833bc6ffdbc0dd4f)

fiveop:
I suggest, that you only change the README.md in this PR. We certainly do not support lower versions. Fixin the CI is different problem, thus I would leave the other two files alone.

Susurrus:
Done.

fiveop:
@homu r+

After we are done setting up the new CI configuration, we should probably revisit this.

PR #526, opened 2017-02-25
Remove EPOLLWAKEUP and SIGEV_THREAD_ID when cross-compiling to MIPS.

Fulkerson:
With this nix builds for mipsel-unknown-linux-gnu.


posborne:
Excellent, I think we have some other changes coming down the pipeline that will further fix things up for MIPS and other cross-compilation targets.  This looks good to me, however.

Pending CI @homu r+

PR #527, opened 2017-02-25
Add cfmakeraw/cfsetspeed

Susurrus:
We'll see how this tests on other platforms, but it's supported on BSDs and Linux supposedly.

### DIFF
--- a/src/sys/termios.rs
+++ b/src/sys/termios.rs
@@ -1,647 +1,859 @@
+//! An interface for controlling asynchronous communication ports
+//!
+//! This interface provides a safe wrapper around the termios subsystem defined by POSIX. The
+//! underlying types are all implemented in libc for most platforms and either wrapped in safer
+//! types here or exported directly.
+//!
+//! If you are unfamiliar with the `termios` API, you should first read the
+//! [API documentation](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/termios.h.html) and
+//! then come back to understand how `nix` safely wraps it.
+//!
+//! It should be noted that this API incurs some runtime overhead above the base `libc` definitions.
+//! As this interface is not used with high-bandwidth information, this should be fine in most
+//! cases. The primary cost when using this API is that the `Termios` datatype here duplicates the
+//! standard fields of the underlying `termios` struct and uses safe type wrappers for those fields.
+//! This means that when crossing the FFI interface to the underlying C library, data is first
+//! copied into the underlying `termios` struct, then the operation is done, and the data is copied
+//! back (with additional sanity checking) into the safe wrapper types. The `termios` struct is
+//! relatively small across all platforms (on the order of 32-64 bytes).
+//!
+//! The following examples highlight some of the API use cases such that users coming from using C
+//! or reading the standard documentation will understand how to use the safe API exposed here.
+//!
+//! Example disabling processing of the end-of-file control character:
+//!
+//! ```
+//! # use self::nix::sys::termios::SpecialCharacterIndices::VEOF;
+//! # use self::nix::sys::termios::{_POSIX_VDISABLE, Termios};
+//! # let mut termios = unsafe { Termios::default_uninit() };
+//! termios.control_chars[VEOF as usize] = _POSIX_VDISABLE;
+//! ```
+//!
+//! The flags within `Termios` are defined as bitfields using the `bitflags` crate. This provides
+//! an interface for working with bitfields that is similar to working with the raw unsigned
+//! integer types but offers type safety because of the internal checking that values will always
+//! be a valid combination of the defined flags.
+//!
+//! An example showing some of the basic operations for interacting with the control flags:
+//!
+//! ```
+//! # use self::nix::sys::termios::{CS5, CSIZE, Termios};
+//! # let mut termios = unsafe { Termios::default_uninit() };
+//! termios.control_flags & CSIZE == CS5;
+//! termios.control_flags |= CS5;
+//! ```
+
 use {Errno, Result};
-use libc::c_int;
+use libc::{self, c_int, tcflag_t};
+use std::cell::{Ref, RefCell};
+use std::convert::From;
 use std::mem;
 use std::os::unix::io::RawFd;
 
-pub use self::ffi::consts::*;
-pub use self::ffi::consts::SetArg::*;
-pub use self::ffi::consts::FlushArg::*;
-pub use self::ffi::consts::FlowArg::*;
+/// Stores settings for the termios API
+///
+/// This is a wrapper around the `libc::termios` struct that provides a safe interface for the
+/// standard fields. The only safe way to obtain an instance of this struct is to extract it from
+/// an open port using `tcgetattr()`.
+#[derive(Clone)]
+pub struct Termios {
+    inner: RefCell<libc::termios>,
+    /// Input mode flags (see `termios.c_iflag` documentation)
+    pub input_flags: InputFlags,
+    /// Output mode flags (see `termios.c_oflag` documentation)
+    pub output_flags: OutputFlags,
+    /// Control mode flags (see `termios.c_cflag` documentation)
+    pub control_flags: ControlFlags,
+    /// Local mode flags (see `termios.c_lflag` documentation)
+    pub local_flags: LocalFlags,
+    /// Control characters (see `termios.c_cc` documentation)
+    pub control_chars: [libc::cc_t; libc::NCCS],
+}
 
-mod ffi {
-    pub use self::consts::*;
+impl Termios {
+    /// Exposes an immutable reference to the underlying `libc::termios` data structure.
+    ///
+    /// This can be used for interfacing with other FFI functions like:
+    ///
+    /// ```rust
+    /// # extern crate libc;
+    /// # extern crate nix;
+    /// # fn main() {
+    /// # use nix::sys::termios::Termios;
+    /// # let mut termios = unsafe { Termios::default_uninit() };
+    /// let inner_termios = termios.get_libc_termios();
+    /// unsafe { libc::cfgetispeed(&*inner_termios) };
+    /// # }
+    /// ```
+    ///
+    /// There is no public API exposed for functions that modify the underlying `libc::termios`
+    /// data because it requires additional work to maintain type safety.
+    // FIXME: Switch this over to use pub(crate)
+    #[doc(hidden)]
+    pub fn get_libc_termios(&self) -> Ref<libc::termios> {
+        {
+            let mut termios = self.inner.borrow_mut();
+            termios.c_iflag = self.input_flags.bits();
+            termios.c_oflag = self.output_flags.bits();
+            termios.c_cflag = self.control_flags.bits();
+            termios.c_lflag = self.local_flags.bits();
+            termios.c_cc = self.control_chars;
+        }
+        self.inner.borrow()
+    }
 
-    #[cfg(any(target_os = "macos", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd", target_os = "netbsd", target_os = "linux"))]
-    mod non_android {
-        use super::consts::*;
-        use libc::c_int;
+    /// Exposes the inner `libc::termios` datastore within `Termios`.
+    ///
+    /// This is unsafe because if this is used to modify the inner libc::termios struct, it will not
+    /// automatically update the safe wrapper type around it. Therefore we disable docs to
+    /// effectively limit its use to nix internals. In this case it should also be paired with a
+    /// call to `update_wrapper()` so that the wrapper-type and internal representation stay
+    /// consistent.
+    unsafe fn get_libc_termios_mut(&mut self) -> *mut libc::termios {
+        {
+            let mut termios = self.inner.borrow_mut();
+            termios.c_iflag = self.input_flags.bits();
+            termios.c_oflag = self.output_flags.bits();
+            termios.c_cflag = self.control_flags.bits();
+            termios.c_lflag = self.local_flags.bits();
+            termios.c_cc = self.control_chars;
+        }
+        self.inner.as_ptr()
+    }
 
-        // `Termios` contains bitflags which are not considered
-        // `foreign-function-safe` by the compiler.
-        #[allow(improper_ctypes)]
-        extern {
-            pub fn cfgetispeed(termios: *const Termios) -> speed_t;
-            pub fn cfgetospeed(termios: *const Termios) -> speed_t;
-            pub fn cfsetispeed(termios: *mut Termios, speed: speed_t) -> c_int;
-            pub fn cfsetospeed(termios: *mut Termios, speed: speed_t) -> c_int;
-            pub fn tcgetattr(fd: c_int, termios: *mut Termios) -> c_int;
-            pub fn tcsetattr(fd: c_int,
-                             optional_actions: c_int,
-                             termios: *const Termios) -> c_int;
-            pub fn tcdrain(fd: c_int) -> c_int;
-            pub fn tcflow(fd: c_int, action: c_int) -> c_int;
-            pub fn tcflush(fd: c_int, action: c_int) -> c_int;
-            pub fn tcsendbreak(fd: c_int, duration: c_int) -> c_int;
+    /// Allows for easily creating new Termios structs that will be overwritten with real data.
+    ///
+    /// This should only be used when the inner libc::termios struct will be overwritten before it's
+    /// read.
+    // FIXME: Switch this over to use pub(crate)
+    #[doc(hidden)]
+    pub unsafe fn default_uninit() -> Self {
+        Termios {
+            inner: RefCell::new(mem::uninitialized()),
+            input_flags: InputFlags::empty(),
+            output_flags: OutputFlags::empty(),
+            control_flags: ControlFlags::empty(),
+            local_flags: LocalFlags::empty(),
+            control_chars: [0 as libc::cc_t; NCCS],
         }
     }
 
-    #[cfg(any(target_os = "macos", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd", target_os = "netbsd", target_os = "linux"))]
-    pub use self::non_android::*;
-
-    // On Android before 5.0, Bionic directly inline these to ioctl() calls.
-    #[cfg(all(target_os = "android", not(target_arch = "mips")))]
-    mod android {
-        use libc;
-        use libc::c_int;
-        use super::consts::*;
-
-        const TCGETS: c_int = 0x5401;
-        const TCSBRK: c_int = 0x5409;
-        const TCXONC: c_int = 0x540a;
-        const TCFLSH: c_int = 0x540b;
-        const TCSBRKP: c_int = 0x5425;
-
-        pub unsafe fn cfgetispeed(termios: *const Termios) -> speed_t {
-            ((*termios).c_cflag & CBAUD).bits() as speed_t
-        }
-        pub unsafe fn cfgetospeed(termios: *const Termios) -> speed_t {
-            ((*termios).c_cflag & CBAUD).bits() as speed_t
-        }
-        pub unsafe fn cfsetispeed(termios: *mut Termios, speed: speed_t) -> c_int {
-            (*termios).c_cflag.remove(CBAUD);
-            (*termios).c_cflag.insert(ControlFlags::from_bits_truncate(speed) & CBAUD);
-            0
-        }
-        pub unsafe fn cfsetospeed(termios: *mut Termios, speed: speed_t) -> c_int {
-            (*termios).c_cflag.remove(CBAUD);
-            (*termios).c_cflag.insert(ControlFlags::from_bits_truncate(speed) & CBAUD);
-            0
-        }
-        pub unsafe fn tcgetattr(fd: c_int, termios: *mut Termios) -> c_int {
-            libc::ioctl(fd, TCGETS, termios)
-        }
-        pub unsafe fn tcsetattr(fd: c_int,
-                                optional_actions: c_int,
-                                termios: *const Termios) -> c_int {
-            libc::ioctl(fd, optional_actions, termios)
-        }
-        pub unsafe fn tcdrain(fd: c_int) -> c_int {
-            libc::ioctl(fd, TCSBRK, 1)
-        }
-        pub unsafe fn tcflow(fd: c_int, action: c_int) -> c_int {
-            libc::ioctl(fd, TCXONC, action)
-        }
-        pub unsafe fn tcflush(fd: c_int, action: c_int) -> c_int {
-            libc::ioctl(fd, TCFLSH, action)
-        }
-        pub unsafe fn tcsendbreak(fd: c_int, duration: c_int) -> c_int {
-            libc::ioctl(fd, TCSBRKP, duration)
-        }
+    /// Updates the wrapper values from the internal `libc::termios` data structure.
+    #[doc(hidden)]
+    pub fn update_wrapper(&mut self) {
+        let termios = *self.inner.borrow_mut();
+        self.input_flags = InputFlags::from_bits_truncate(termios.c_iflag);
+        self.output_flags = OutputFlags::from_bits_truncate(termios.c_oflag);
+        self.control_flags = ControlFlags::from_bits_truncate(termios.c_cflag);
+        self.local_flags = LocalFlags::from_bits_truncate(termios.c_lflag);
+        self.control_chars = termios.c_cc;
     }
+}
 
-    #[cfg(target_os = "android")]
-    pub use self::android::*;
-
-
-    #[cfg(any(target_os = "macos", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd", target_os = "netbsd"))]
-    pub mod consts {
-
-        use libc;
-
-        use libc::c_int;
-
-        pub type tcflag_t = libc::tcflag_t;
-        pub type cc_t = libc::cc_t;
-        pub type speed_t = libc::speed_t;
-
-        #[repr(C)]
-        #[derive(Clone, Copy)]
-        pub struct Termios {
-            pub c_iflag: InputFlags,
-            pub c_oflag: OutputFlags,
-            pub c_cflag: ControlFlags,
-            pub c_lflag: LocalFlags,
-            pub c_cc: [cc_t; NCCS],
-            pub c_ispeed: speed_t,
-            pub c_ospeed: speed_t
-        }
-
-        #[derive(Clone, Copy, Debug, PartialEq)]
-        pub enum BaudRate {
-            B0,
-            B50,
-            B75,
-            B110,
-            B134,
-            B150,
-            B200,
-            B300,
-            B600,
-            B1200,
-            B1800,
-            B2400,
-            B4800,
-            B9600,
-            B19200,
-            B38400,
-            B7200,
-            B14400,
-            B28800,
-            B57600,
-            B76800,
-            B115200,
-            B230400,
-            #[cfg(any(target_os = "netbsd", target_os = "freebsd"))]
-            B460800,
-            #[cfg(any(target_os = "netbsd", target_os = "freebsd"))]
-            B921600,
-        }
-
-        impl From<speed_t> for BaudRate {
-            fn from(s: speed_t) -> BaudRate {
-
-                use libc::{
-                    B0, B50, B75, B110, B134, B150,
-                    B200, B300, B600, B1200, B1800, B2400,
-                    B4800, B9600, B19200, B38400,
-                    B7200, B14400, B28800, B57600,
-                    B76800, B115200, B230400};
-
-                #[cfg(target_os = "freebsd")]
-                use libc::{B460800, B921600};
-
-                match s {
-                    B0 => BaudRate::B0,
-                    B50 => BaudRate::B50,
-                    B75 => BaudRate::B75,
-                    B110 => BaudRate::B110,
-                    B134 => BaudRate::B134,
-                    B150 => BaudRate::B150,
-                    B200 => BaudRate::B200,
-                    B300 => BaudRate::B300,
-                    B600 => BaudRate::B600,
-                    B1200 => BaudRate::B1200,
-                    B1800 => BaudRate::B1800,
-                    B2400 => BaudRate::B2400,
-                    B4800 => BaudRate::B4800,
-                    B9600 => BaudRate::B9600,
-                    B19200 => BaudRate::B19200,
-                    B38400 => BaudRate::B38400,
-                    B7200 => BaudRate::B7200,
-                    B14400 => BaudRate::B14400,
-                    B28800 => BaudRate::B28800,
-                    B57600 => BaudRate::B57600,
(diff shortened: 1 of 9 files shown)


Susurrus:
Sure, that definitely seems like the right play. I've already started here: rust-lang/libc#541

While I've been waiting for that PR I started trying to use those functions and I don't really know how to do this. if I just use `libc::tcsetspeed` then I get errors about the `termios` datatypes not matching up. So my question is where is the interface between `nix` and `libc` here? Do we use `libc::Termios` or do we want to use the current one? If we want to use the current one, how do I adapt that to `libc::termios::termios`?

For example if I implement it like this:
```diff
diff --git a/src/sys/termios.rs b/src/sys/termios.rs
index e8df1ed..954ec85 100644
--- a/src/sys/termios.rs
+++ b/src/sys/termios.rs
@@ -1,5 +1,5 @@
 use {Errno, Result};
-use libc::c_int;
+use libc::{self, c_int};
 use std::mem;
 use std::os::unix::io::RawFd;
 
@@ -599,6 +599,12 @@ pub fn cfgetospeed(termios: &Termios) -> BaudRate {
     }
 }
 
+pub fn cfmakeraw(termios: &mut Termios) {
+    unsafe {
+        libc::cfmakeraw(termios);
+    }
+}
+
 pub fn cfsetispeed(termios: &mut Termios, baud: BaudRate) -> Result<()> {
     Errno::result(unsafe {
         ffi::cfsetispeed(termios, baud as speed_t)
@@ -611,6 +617,12 @@ pub fn cfsetospeed(termios: &mut Termios, baud: BaudRate) -> Result<()> {
     }).map(drop)
 }
 
+pub fn cfsetspeed(termios: &mut Termios, baud: BaudRate) -> Result<()> {
+    Errno::result(unsafe {
+        libc::cfsetspeed(termios, baud as speed_t)
+    }).map(drop)
+}
+
 pub fn tcgetattr(fd: RawFd) -> Result<Termios> {
     let mut termios = unsafe { mem::uninitialized() };
```

then I get these errors:
```rust
error[E0308]: mismatched types
   --> src/sys/termios.rs:604:25
    |
604 |         libc::cfmakeraw(termios);
    |                         ^^^^^^^ expected struct `libc::libc::termios`, found struct `sys::termios::ffi::consts::Termios`
    |
    = note: expected type `*mut libc::libc::termios`
    = note:    found type `&mut sys::termios::ffi::consts::Termios`

error[E0308]: mismatched types
   --> src/sys/termios.rs:622:26
    |
622 |         libc::cfsetspeed(termios, baud as speed_t)
    |                          ^^^^^^^ expected struct `libc::libc::termios`, found struct `sys::termios::ffi::consts::Termios`
    |
    = note: expected type `*mut libc::libc::termios`
    = note:    found type `&mut sys::termios::ffi::consts::Termios`
```

Any guidance here as to how to properly wrap libc's datatypes in nix?

posborne:
> Any guidance here as to how to properly wrap libc's datatypes in nix?

I think we probably want to use nix's type at the interface level (which provide greater safety) and ensure that proper conversion functions into the libc struct format exist.  The memory layout of the two is required to be the same at present, so we just need to (unsafely) `transmute` between those types.

Susurrus:
Is there a convention for this elsewhere in nix I can copy? I'm uncertain whether implementing From would be better than doing it manually in each location and figured there should be consensus on this in nic already.

fiveop:
Yes. Look for example at `SigSet` in `src/sys/signal.rs`. It is just an opaque wrapper for libc::sigset_t. This way, libc takes care of the different layouts on different platforms, and we don't have to have multiple definitions.

Susurrus:
So I've been making some progress on this, but I wanted to understand how high-level you wanted the API. Given the API designed for `SigSet` I imagine that all the `cf*` functions in `termios` should be methods on the `Termios` struct rather than these standalone functions. Additionally should I replace `tcgetattr` with a `from_fd` function as the `Termios` crate has? Now the question is should the `tc*` functions all be methods on the `RawFd` type instead of stand alone functions?

Susurrus:
So I have everything all set for refactoring `src/sys/termios.rs` to use `libc` and adding `cfmakeraw` and `cfsetspeed`. Now I'm not quite certain what the difference is between the `src` and `src/sys` modules, but hopefully this code is in the right place.

fiveop:
Good work. There are at least two problems with flags under OSX left.

Regarding your question about making the functions into methods of Termios, like it was done with Signal: I do not think, that we have yet discussed what style we prefer. I prefer plain functions with arguments for nix. My reason for this is, that I see libc as the place for a raw binding, nix as a place for a safe binding, as far as that's possible, and other libraries can build upon that to create idiomatic wrappers, whatever idiomatic means at that point in time.

The module structure derives from the C file header paths. So, we have `src/sys/signal.rs` because the corresponding C header gets included with `#include <sys/signal.h>`.

Susurrus:
This is actually blocking on rust-lang/libc#542 where it looks like they don't have two of the constants we need. Once that's merged that should fix this.

Susurrus:
Rebased and pushed again now that libc has been updated with the proper constants. This should succeed on OS X now whenever Travis gets around to is.

Susurrus:
Passing now on Mac. CI hasn't finished but all the necessary platforms have passed.

Susurrus:
Missed a couple of baud rate constants in my original patch, so I've re-added those for the appropriate OSes. Also I alphabetized the `#[cfg` that use `target_os` cause I was having trouble comparing them. And finally I ran things through rustfmt which is why some of the trivial formatting has changed. I assumed that was a direction this project will go (rustfmt as part of CI), so I took the initiative to do it now. If that was inappropriate let me know and I can undo it.

kamalmarhubi:
@Susurrus I'd prefer rustfmt in a separate change.

Susurrus:
@kamalmarhubi Other projects I've worked on have only allowed style changes for code that's being modified so that `git blame` gives correct results on the first try. I don't know if there's existing policy, but all the style changes I made were only in areas that I was already touching. If you still want me to separate out the style changes, you want them as a separate commit before or after my other changes, or would you like them as a separate PR, or should I just wait on style stuff in general because running `rustfmt` on `nix` shows that most of it doesn't follow the style guidelines and will need cleanup later.

Susurrus:
I should also note that by doing this the `Termios` struct losing public access to all of its members. This "breaks" the struct itself as you cannot get or set any of its fields. I'm going to modify this change to support accessors for each field. Is there any specific naming convention I should use? For example should I change the getter/setter names to be more appropriate like `set_input_flags()` instead of `set_c_iflag()` and so on? If so, then we'd end up with:

```
Termios::set_input_flags();
Termios::set_output_flags();
Termios::set_control_flags();
Termios::set_local_flags();
Termios::set_line_discipline();
Termios::set_special_characters();
Termios::set_input_speed();
Termios::set_output_speed();
```
and corresponding functions where `set` is replaced with `get`.

The problem with this however is that these setters & getters partially overlap with existing termios functionality, specifically the `cfget?speed` and `cfset?speed` functions. I don't know what the best solution here is as previous conversations have stated that higher-level APIs should be delegated downstream where libc is trying to be as minimal as a wrapper as possible for a safe API around libc.

kamalmarhubi:
> Other projects I've worked on have only allowed style changes for code that's being modified so that git blame gives correct results on the first try.

Ah, that's https://github.com/rust-lang-nursery/rustfmt/issues/434 which I should really finish some day :/

kamalmarhubi:
> I should also note that by doing this the `Termios` struct losing public access to all of its members. This "breaks" the struct itself as you cannot get or set any of its fields.

Could you add `Deref<libc::termios>` and `DerefMut<libc::termios>` impls and see if that works?

@fiveop @posborne ^^^^ potential new convention?

posborne:
> Could you add Deref<libc::termios> and DerefMut<libc::termios> impls and see if that works?

Yeah, I've seen that used and implemented that in a few spots (not in nix) before and it can work quite nicely to succinctly provide access to fields in a field that still provides control.

posborne:
Also, this change in general looks very good.  Like seeing more of this finally get moved to libc.

fiveop:
Regarding the accessors. In the cases that I can remember, only the constructor and read only accessors were necessary. So the accessors where named after the fields (e.g. `foo()` for a field `foo`). If this is not the case for Termios and there is no convention documented in CONVENTION.md, just make a suggestion. (I prefer `set_foo` and `get_foo`. What is the convention in other Rust projects (in particular in the standard library?)

Susurrus:
So it seems like there's two directions we can take, head towards setters & getters or just expose the underlying `libc::termios::termios` struct. It would be nice to use the `InputFlags` and related fields simply with the `Termios` struct, like they currently work, but then delegate all the low-level stuff to libc. I'm still not certain how this will work out, but since I have a library that uses this functionality, I'll have on it locally a bit and see what ends up being ergonomic.

Susurrus:
Okay, so I implemented `Deref` and `DerefMut` and I'm not certain I like the ergonomics. In my code I have a bunch of stuff that used to look like:
```
        use termios::{CSIZE, CS5, CS6, CS7, CS8};
        match self.termios.c_cflag & CSIZE {
            CS8 => Some(DataBits::Eight),
            CS7 => Some(DataBits::Seven),
            CS6 => Some(DataBits::Six),
            CS5 => Some(DataBits::Five),

            _ => None,
        }
```

For this I can't just put `.bits()` after every constant (necessary for the `CS*` datatypes to match the `termios.c_cflag` ones) because it's a syntax error in a match statement.

Another use case of mine would, with this change, look like:
```
        use nix::termios::{CRTSCTS, IXON, IXOFF};

        if self.termios.c_cflag & CRTSCTS.bits() != 0 {
            Some(FlowControl::Hardware)
        } else if self.termios.c_iflag & (IXON | IXOFF).bits() != 0 {
            Some(FlowControl::Software)
        } else {
            Some(FlowControl::None)
        }
```

So there's a liberal sprinkling of `.bits()` everywhere to translate between the low-level termios flag types and those provided by `nix`.

It really seems like we're straddling in between two levels of API. A low-level API roughly similar to that provided by `libc` and a higher level one that makes the bit-flags easier to work with. And I think we have to choose what we want, basically just pass through the termios flag types or actually try to encapsulate them **and** provide higher-level functions to manipulate the entire `termios` struct.

Susurrus:
@posborne @fiveop Any comments on this? I don't really know what directly to take this API and could use some help.

Susurrus:
@fiveop I'm pushed a newer version that just re-exports the `termios` struct from `libc`. This should solve our field accessor issues, though the fields won't be usable with our custom enums and types directly without a manual conversion step. Unfortunately I think this is really the best thing to do here unless we want to offer a higher-level API above the termios header functions.

I plan on adding tests to this to make sure the functionality is sound once I get someone else to take a glance over this and make sure this is the path we want to go down.

Susurrus:
Everything compiles on all platforms but is blocking on rust-lang/libc#574

posborne:
@Susurrus I looked at the changes here late but everything seems to be in good order.  Do you feel ok with moving forward on this one still (been a little bit since it was last touched).

Maybe a rebase is in order.

Susurrus:
I think this is fine to go ahead. I'll take a look at this in the next few days and rebase it and then I'll push if you have no objections.

Susurrus:
Fixed a few things and now I'm thinking it's ready for merge:
 * Use libc constants for `SetArg`
 * Added `Debug` implementations for remaining enums
 * Use `into` instead of `as` for baud rate constants. I think it was incorrect before.

Wanna take one last look @posborne?

Susurrus:
This needs a rebase (which I'll do), but I just wanted to ping you again @posborne to see if you could review this one last time?

Susurrus:
@asomers My tests that work on Linux for this fail on FreeBSD. Do you know what the deal is with `tcgetattr(0)` on FreeBSD returning `ENOTTY`? Maybe I'll need to create ptys to make this work across all platforms.

asomers:
I'm surprised that it works on Linux.  Does Travis-CI use a pty perhaps?  What happens if you run "cargo test </dev/null | cat" ?

Susurrus:
@asomers I get the same error, good catch! This was a test that I copied from somewhere else in nix I thought, so surprised this failed and not those other ones. So I should be making a pty to do these tests, yes?

asomers:
Yep.  Making a pty is definitely the way to go.

Susurrus:
👍

Susurrus:
I've modified my `serialport-rs` lib to use this branch to confirm that this implementation is sufficient and to test the ergonomics beyond the test cases I already implemented here. What I've found is that this is okay, but definitely not ideal. The usability of the `termios-rs` crate is much better. I've listed below some of the current pain points:

  1) `libc::termios` doesn't implement `PartialEq`/`Eq`, this means my code to compare the structs looks like:
  ```rust
            if actual_termios.c_iflag != termios.c_iflag ||
               actual_termios.c_oflag != termios.c_oflag ||
               actual_termios.c_lflag != termios.c_lflag ||
               actual_termios.c_cflag != termios.c_cflag {
  ```
  2) Matching on termios values can't be done because patterns can't rely on evaluating an expression, so instead of the ideal use case being
```rust
match self.termios.c_cflag & CSIZE {
    CS8 => _,
    ...
}
```

we instead have to do:
```rust
let cflag = self.termios.c_cflag & CSIZE.bits();
if cflag == CS8.bits() {
     Some(DataBits::Eight)
} else ...
```

because you can't put `CS8.bits()` in a pattern to be matched.
  3) You have to use `.bits()` everywhere to convert between the bitflags type and the underlying integer value, so you have to do things like `if self.termios.c_cflag & PARENB.bits() != 0 {` and `if self.termios.c_iflag & (IXON | IXOFF).bits() != 0 {`. This applies to clearing or setting values as well: `self.termios.c_cflag &= !CSTOPB.bits()`.

I think it'd be nice to provide a higher-level wrapper around the termios stuff, but I've run out of ideas for a lower-level API that doesn't have these kinds of warts.

Susurrus:
I found that I missed `tcgetsid` in this as well, so I've added it here, though this is blocking on rust-lang/libc#635.

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
-    }
+/// FIXME: This struct has no documentation within libc, which is very obvious when generating
+///        docs
+pub use libc::termios as Termios;
+
+/// Baud rates supported by the system
+///
+/// B0 is special and will disable the port.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum BaudRate {
```
asomers: These enums are going to have different values than the similarly named constants from libc, so implementing the `From` trait below is going to require real work at runtime.  Did you consider simply wrapping libc's constants in a enum?  That way, converting back and forth can be done by simple casts.  See `nix::sys::Signal` or `nix::sys::LioOpcode` for examples.

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
-            pub c_cc: [cc_t; NCCS],
-            pub c_ispeed: speed_t,
-            pub c_ospeed: speed_t
-        }
+/// Specify when a port configuration change should occur.
+///
+/// Used as an argument to `tcsetattr()`
+// TODO: Add TCSASOFT, which will require treating this as a bitfield.
+#[derive(Clone, Copy, Debug)]
+pub enum SetArg {
```
asomers: In C these constants are always treated as `int`s, not `ssize_t`s.  You should add a `#[repr(i32)]` here, and ditch the `as isize` for each entry.

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
-            B3000000,
-            B3500000,
-            B4000000,
-        }
+/// Specify a combination of the input and output buffers to flush
+///
+/// Used as an argument to `tcflush()`.
+#[derive(Clone, Copy, Debug)]
+// TODO: Change these over to a `libc_enum` macro
+pub enum FlushArg {
```
asomers: as with `SetArg`, you should use `#[repr(i32)]` and ditch the `as isize`.  What's a `libc_enum` macro?

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
-                    b @ _ => unreachable!("Invalid baud constant: {}", b),
-                }
-            }
-        }
+/// Specify how transmission flow should be altered
+///
+/// Used as an argument to `tcflow()`.
+#[derive(Clone, Copy, Debug)]
+// TODO: Change these over to a `libc_enum` macro
+pub enum FlowArg {
```
asomers: Same as for `SetArg`

test/sys/test_termios.rs
```diff
@@ -19,3 +24,51 @@ fn test_tcgetattr() {
...
+// Test ergonomics of using character indices
+#[test]
+fn test_char_indices() {
+    use nix::sys::termios::SpecialCharacterIndices::VEOF;
+    use nix::sys::termios::_POSIX_VDISABLE;
+
+    let master_fd = posix_openpt(O_RDWR).unwrap();
+    assert!(master_fd.as_raw_fd() > 0);
+    let mut termios = termios::tcgetattr(master_fd.as_raw_fd()).unwrap();
+    termios.c_cc[VEOF as usize] = _POSIX_VDISABLE;
```
asomers: What test coverage does this statement provide?

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
-                const IXON    = 0x00000400;
-                const IXOFF   = 0x00001000;
-                const IXANY   = 0x00000800;
-                const IMAXBEL = 0x00002000;
-                const IUTF8   = 0x00004000;
-            }
-        }
+/// Indices into the `termios.c_cc` array for special characters.
+// TODO: Change these over to a `libc_enum` macro
+pub enum SpecialCharacterIndices {
```
asomers: Same as for `SetArg`

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
-                const PARENB     = 0x00000100;
-                const PARODD     = 0x00000200;
-                const HUPCL      = 0x00000400;
-                const CLOCAL     = 0x00000800;
-                const CRTSCTS    = 0x80000000;
-                #[cfg(target_os = "android")]
-                const CBAUD      = 0o0010017;
-            }
-        }
+// FIXME: Get these types changed to all `tcflag_t` in `libc` so this can use the `libc_bitflags` macro
```
asomers: It looks like most of these constants are already defined as `tcflag_t`.  Would it be possible to use the `libc_bitflags` macro now, if you added `as libc::tcflag_t` to the stragglers?

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
+                  target_os = "netbsd",
+                  target_os = "openbsd"))]
+        NOKERNINFO,
+        PENDIN,
+        NOFLSH,
     }
 }
 
+/// Get input baud rate (see
+/// [cfgetispeed(3p)](http://man7.org/linux/man-pages/man3/cfgetispeed.3p.html)).
```
asomers: Since this function is standardized by POSIX, it would probably be better to link to the Open Group's man page, instead of Linux's.

src/sys/termios.rs
```diff
@@ -1,648 +1,782 @@
...
+pub fn cfsetspeed(termios: &mut Termios, baud: BaudRate) -> Result<()> {
+    Errno::result(unsafe { libc::cfsetspeed(termios, baud.into()) }).map(drop)
 }
 
+/// Return the configuration of a port
+/// [tcgetattr(3p)](http://man7.org/linux/man-pages/man3/tcgetattr.3p.html)).
+///
+/// `tcgetattr()` returns a Termios structure with the current configuration for a port. Modifying
+/// this structure *will not* reconfigure the port, instead the modifications should be done to
+/// the `Termios` structure and then the port should be reconfigured using `tcsetattr()`. Note
```
asomers: Shouldn't this note about returning success if any parameters were successfully set belong in the comment block for `tcsetattr` instead of `tcgetattr`?

ndusart:
@Susurrus regarding the comparison between bitflags and integer, `bitflags` have [some functions for making these comparison](https://doc.rust-lang.org/bitflags/bitflags/example_generated/struct.Flags.html).

```rust
match self.termios.c_cflag & CSIZE {
    CS8 => _,
    ...
}
```

can be written as
```rust
match self.termios.c_cflag.contains(CSIZE) {
    CS8 => _,
    ...
}
```

Look at my merge request in your serialport-rs repo. All went quite smoothly using the current master branch of nix. It just lacks `cfmakeraw`, `cfsetspeed` from this PR and as you mentioned `Termios` is not `PartialEq` nor `Debug`.

ndusart:
@Susurrus regarding looking at flags inside a bitflags, there is [some functions](https://doc.rust-lang.org/bitflags/bitflags/example_generated/struct.Flags.html) to do it.

You could write
```rust
if self.termios.c_iflag & (IXON | IXOFF).bits() != 0 {
```
like this
```rust
if self.termios.c_iflag.intersects(IXON | IXOFF) {
```

The `match` is building correctly though because normally `self.termios.c_cflag & CSIZE` is a `Flags` (using the `BitAnd` trait of the bitflags), then you can match against the bitflags constants. You should probably sync termios.rs with the current one.

Look at my merge request in your serialport-rs repository. All went quite smoothly using the current master branch of nix.
It does not build only because it lacks `cfmakeraw` and `cfmakespeed` from this PR and because `Termios` is not `PartialEq` and `Debug`.

Susurrus:
So we're missing a constant for Android, but we also have the issue that the baud rate constants are defined as `unsigned long` in `libc`. These are therefore the wrong size on 64-bit Mac OS X and iOS. I'm not sure why these failures aren't occuring on 32-bit OS X/iOS. Just pushed a test to try to figure out what's going on here as I really want to get `iOS` builds to Tier 2 as part of this.

test/sys/test_termios.rs
Susurrus: It doesn't provide test coverage. As the comment at the start of the function describes, it's merely there to check the ergonomics. I wanted to make sure my API easily supported modifying the `termios.c_cc` array using the enums and constants I defined. I wanted to make sure everything for termios is imported from `libc` and uses safe datatypes that still allow for all normal usage. It might make more sense to move this (and the other "test" I implemented) to a doc comment instead of a separate test.

Susurrus:
@ndusart had this suggestion when using this branch with `serialport-rs`:
>Done, but we have now all the problem you mentioned regarding the flags not being able to compare with the termios fields.
>
>You replace the `Termios` with the current definition from libc.
>I think we should still use a `Termios` struct in nix that present the fields as bitflags. That would add type safety when manipulating these fields and allow to make these comparisons more easily.
>I know there are multiple implementations of `termios` according to architecture but maybe we can still use the libc definition while providing bitflags for `c_*flag` by using Deref coercion.
>
>```rust
>struct Termios {
>  inner: libc::termios,
>  c_iflag: InputFlags,
>  ...
>}
>
>impl Deref for Termios {
>  type Target = libc::termios;
>
>  fn deref(&self) -> &Self::Target {
>    &self.inner
>  }
>}
>```
>
>With this, when accessing a field that is not overwritten by `Termios`, it will look in the `libc::termios` struct.
>It seems that the `c_*flag` are present in all platforms, so that should do it :)
>What do you think ?

I'm working on implementing this, but I'm not certain if this is possible since I need to copy from these outer fields into the inner fields. My initial implementation looks like:
```rust
/// Stores settings for the termios API
pub struct Termios {
    inner: libc::termios,
    /// Input mode flags
    pub input_flags: InputFlags,
    /// Output mode flags
    pub output_flags: OutputFlags,
    /// Control mode flags
    pub control_flags: ControlFlags,
    /// Local mode flags
    pub local_flags: LocalFlags,
    /// Control characters
    pub control_chars: [libc::cc_t; libc::NCCS],
}

impl Deref for Termios {
    type Target = libc::termios;

    fn deref(&self) -> &Self::Target {
        // Make sure to update the internal reference based on the latest values of the "safe"
        // wrapper fields.
        self.inner.c_iflag = self.input_flags.bits();
        self.inner.c_oflag = self.output_flags.bits();
        self.inner.c_cflag = self.control_flags.bits();
        self.inner.c_lflag = self.local_flags.bits();
        self.inner.c_cc = self.control_chars;
        &self.inner
    }
}

impl DerefMut for Termios {
    fn deref_mut(&mut self) -> &mut Self::Target {
        // Make sure to update the internal reference based on the latest values of the "safe"
        // wrapper fields.
        self.inner.c_iflag = self.input_flags.bits();
        self.inner.c_oflag = self.output_flags.bits();
        self.inner.c_cflag = self.control_flags.bits();
        self.inner.c_lflag = self.local_flags.bits();
        self.inner.c_cc = self.control_chars;
        &mut self.inner
    }
}
```

*but* I can't actually modify `self.inner` within `deref()`, since it's borrowed immutably, but it's necessary to copy back and forth from the `safe` fields when doing this. I'm going to try using a `Cell<libc::termios>` and see how far I get.

ndusart:
Right, I should have posted it here.

Could we use `Cell` to have interior mutability ?

Susurrus:
@ndusart I'm not certain we can, but I'm working it out on IRC right now in #rust-beginners.

Susurrus:
Alright! Well, that was quite a morning trying to learn how to deal with `RefCell`! Anyways, I think I have a decent implementation at this point. I've added extensive documentation that I hope will address everything here, so I'll let that speak for itself. Feedback is welcome!

Susurrus:
Still getting some intermittent failures with `errno_values` on `powerpc`. We're gonna have to figure that out at some point, or just not run them on that platform.

Susurrus:
And `test_mq::test_mq_send_and_receive` stalled on `x86_64-unknown-linux-gnu` on the beta channel.

ndusart:
Should also make the `Termios` struct implements `Copy` as it is currently the case in master branch.

Susurrus:
`libc::termios` is `Copy` and `Clone` (all structs in `libc` are, they use the `s!` macro to define them which adds it).

But `Termios` uses `RefCell`, which doesn't implement `Copy`, so it can't be `Copy`, but I've added `Clone`, which should be good enough.

Susurrus:
We have a downstream library already successfully using this branch, providing a nice use ergonomics and functionality check and it seems like we've covered everything. I am going to write a couple more tests to make sure that the copying to/from the termios struct works alright, but I just gotta figure out how to do that. And then this should be good to merge.

Susurrus:
This includes the `::read_exact()` call that's also in #660, so I'd like to get that merged first then modify this PR to use that version instead of duplicating the code.

I managed to whip up a couple of tests that check that we can make termios settings changes and they get applied properly. These should work across all platforms, but let's see.

I'm also excited that this will bring our `ios` platforms all up to Tier 2.

Susurrus:
Alright, I think this is GTM. Wrote up a couple of tests, as much as I think is reasonable to do here. I have a PR against my `serialport-rs` crate that utilizes this functionality and it seems to work correctly. Can I get a final review from someone in case I missed something?

src/sys/termios.rs
```diff
@@ -1,647 +1,894 @@
...
+//! The flags within `Termios` are defined as bitfields using the `bitflags` crate. This provides
+//! an interface for working with bitfields that is similar to working with the raw unsigned
+//! integer types but offers type safety because of the internal checking that values will always
+//! be a valid combination of the defined flags.
+//!
+//! An example showing some of the basic operations for interacting with the control flags:
+//!
+//! ```
+//! # use self::nix::sys::termios::{CS5, CSIZE, Termios};
+//! # let mut termios = unsafe { Termios::default_uninit() };
```
Susurrus: This isn't a comment, it just means that these lines are excluded from rendering when generating docs. `default_uninit()` shouldn't be used by external code, it's just to give me an object that I can use for this.

Susurrus:
Fixed your current review comments about spellings and switched around how public some of the helper functions are. For those we're waiting on `pub(crate)` to stabilize to properly limit their exposure, but for now we hide them from the docs.

src/sys/termios.rs
asomers: Ahh, that makes sense.  As you can see, I'm a doccomment nube.

test/sys/test_termios.rs
```diff
@@ -1,6 +1,29 @@
...
+use nix::pty::openpty;
+use nix::sys::termios::{self, ECHO, OPOST, OCRNL, Termios, tcgetattr};
+use nix::unistd::{read, write, close};
+
+/// Helper function analogous to std::io::Write::write_all, but for `RawFd`s
+fn write_all(f: RawFd, buf: &[u8]) {
+    let mut len = 0;
+    while len < buf.len() {
+        // get_mut would be better than split_at_mut, but it requires nightly
+        len += write(f, &buf[0..]).unwrap();
```
asomers: Are you sure this is right?  What advances `buf`?

test/sys/test_termios.rs
```diff
@@ -19,3 +42,93 @@ fn test_tcgetattr() {
...
+// Test modifying output flags
+#[test]
+fn test_output_flags() {
+    // Open one pty to get attributes for the second one
+    let mut termios = {
+        let pty = openpty(None, None).unwrap();
+        assert!(pty.master > 0);
+        assert!(pty.slave > 0);
+        let termios = tcgetattr(pty.master).unwrap();
+        close(pty.master).unwrap();
```
asomers: This is a double close.  See issue #659.  There are a few more instances of it below.

src/sys/termios.rs
```diff
@@ -1,647 +1,894 @@
...
-        const TCFLSH: c_int = 0x540b;
-        const TCSBRKP: c_int = 0x5425;
+use ::unistd::Pid;
+
+/// Stores settings for the termios API
+///
+/// This is a wrapper around the `libc::termios` struct that provides a safe interface for the
+/// standard fields. The only safe way to obtain an instance of this struct is to extract it from
+/// an open port using `tcgetattr()`.
+pub struct Termios {
```
ndusart: `Termios` is not `Clone` anymore

src/sys/termios.rs
Susurrus: No worries, there's lots of weird syntax stuff with doccomments. Also note that `#` isn't a comment in Rust :-)

Susurrus:
I think I've addressed everyone's issues for this PR, so if you could take one last look at it @ndusart and @asomers and tell bors if we're all good, I'd appreciate it!

asomers:
I don't see any solution for the double close bug in test_termios.rs.  Did you forget that one?

Susurrus:
Indeed! I think I got one but not the other. Should be fixed now.

asomers:
That fixed the double close.  But I just noticed that we don't have a CHANGELOG entry for the new functions.  Could you please add one?

Susurrus:
@asomers Indeed I did forget the changelog. I was waiting until I got the iOS builds sorted and then I forgot about it, thanks for the reminder. While I was in there I did a little cleanup of the CHANGELOG (left it as a separate commit), so it should be a little more readable now.

asomers:
Looks like we're finally done.  Congrats @Susurrus, you've worked hard on this one.

bors r+

Susurrus:
Well thanks to everyone to helped out with this PR, with all the reviews and discussions I think we actually finalized on a pretty decent API here! I will say, this is my longest PR I've ever done, since it took 4.5 months to get this thing merged! But it was well worth the effort.

PR #528, opened 2017-02-25
Switch to trust-based CI

berkowski:
An attempt to switch to using [trust](https://github.com/japaric/trust) for multi-arch CI

I selected triplets based on CI coverage in `libc` and `nix`:

- i686-unknown-linux-gnu
- x86_64-unknown-linux-gnu
- i686-unknown-linux-musl 
- x86_64-unknown-linux-musl
- i686-apple-darwin
- x86_64-apple-darwin
- i686-unknown-freebsd (`cargo test` not run)
- x86_64-unknown-freebsd (`cargo test` not run)
- x86_64-unknown-netbsd (`cargo test` not run)
- aarch64-unknown-linux-gnu
- armv7-unknown-linux-gnueabihf
- mips-unknown-linux-gnu
- mipsel-unknown-linux-gnu
- powerpc-unknown-linux-gnu
- arm-unknown-linux-gnueabi

Notable issues:
- Would be nice to get the BSD tests to run `cargo test`
- arm-linux-androideabi is not a current `trust` target 
- `.travis.yml` will need to be updated with a github token from `nix-rust/nix` instead of `berkowski/nix`

Other issues:
- Using current stable rust -- switch to a pinned version?
That being said, `trust` is catching a number of compile errors on various platforms.  See https://travis-ci.org/berkowski/nix


ci/script.sh
```diff
@@ -7,7 +7,7 @@ main() {
     cross build --target $TARGET
     cross build --target $TARGET --release
 
-    if [ -n $DISABLE_TESTS ]; then
+    if [ ! -z $DISABLE_TESTS ]; then
```
posborne: Isn't this logic identical?

ci/script.sh
```diff
@@ -2,20 +2,22 @@
...
 
+    # Run tests on debug and release targets.
     cross test --target $TARGET
     cross test --target $TARGET --release
 
-    cross run --target $TARGET
-    cross run --target $TARGET --release
+    # nix is a library -- no run target
+    # cross run --target $TARGET
+    # cross run --target $TARGET --release
```
posborne: Probably reasonable to update this commit to just delete these.

posborne:
Ok, the errors that we are getting now seem legitimate.  In order to distribute the load of fixing up the errors, I propose the following:

* Maintainer will create a new branch call something like `new-ci/master` and we will merge this branch there.
* We'll have a few PRs fixing various issues into the branch.
* Once everything is green we'll do a final PR of that branch back into master.

Sound reasonable?  Thanks @berkowski  for doing this work and thanks @japaric for putting together trust.  It looks like the integration so far has been very reasonable.

posborne:
> Using current stable rust -- switch to a pinned version?

Down the road, it would be nice to test two versions at a minimum:

* The oldest supported version (I think right now we have discussed 1.10 for this)
* Stable

posborne:
Ok, I changed the base in preparation for doing ^^^.  For now, I think I would be OK with merging as long as there are no other changes you want to make in the first pass.

From there, I think we create/assign issues for the individual platform problems and merge those in as needed (focus on either specific issues are people volunteer to get a specific platform working) and merge them into `new-ci/master`.  Once that is green, we can merge back into master.

Thoughts @nix-rust/nix-maintainers @kamalmarhubi @fiveop ?

fiveop:
Yes, please! This will make working on this project so much more enjoyable (for me :)).

ci/script.sh
berkowski: Quotes would fix it. `-n` is used in v0.1.1 of `trust` while the `-z` version is used on the HEAD of the master branch.  I went with `-z` figuring it'll be switched to that anyway at some point.

ci/script.sh
berkowski: Fair enough, I'll prune it out.

berkowski:
Once that API key gets changed over to `nix-rust` we're good to go.  Could even merge first and swap the keys after if you want to do it when the source is owned by `nix-rust`

posborne:
> Once that API key gets changed over to nix-rust we're good to go.

Sure, I'll go ahead and get this merged into the CI branch.  I can update the API key using @nix-rust-bot which is what I have done in the past for docs and the like.

posborne:
NOTE: Manual merge to CI branch performed as not all tests are passing yet (as expected).

PR #540, opened 2017-02-28
fcntl: Support getting and setting pipe size on Linux

kamalmarhubi:


kamalmarhubi:
r? posborne

posborne:
Looks good.  @homu r+

PR #541, opened 2017-02-28
fcntl: Expose FcntlArg variants at the module level

kamalmarhubi:
This allows importing them directly from `nix::fcntl` which is more
ergonomic than needing to use them via `FcntlArg`.


fiveop:
Given `pub use self::Signal::*;` in src/sys/signal.rs:

@homu r+

Issue #458, opened 2016-11-06
Clarify policy on breaking changes


chaosagent:
Theres no general policy on breaking changes in CONTRIBUTING.md; I think it'd be useful to have one.

fiveop:
I think the current consensus is, that breaking things in order to make the code follow CONVENTIONS.md or to fix some problem (not necessarily a bug, see #445 for example) is fine. When we start to think about a 1.0 version, that will change. Feel free to write a paragraph stating that as a PR, that we can then further discus. If you don't, I might get around to it next week.

posborne:
Yep, at this point we don't want to break APIs unless there is a good reason.  But if there is a good reason, then I think it still makes sense to do so (and semver will change accordingly).

kamalmarhubi:
Thanks for asking about this. We should definitely figure this out.

PR #542, opened 2017-02-28
Remove revents from PollFd::new

Susurrus:
I could've used a `0i16`here as well but I liked the better semantics of `empty()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Exposed all fcntl(2) operations at the module level, so they can be
   imported direclty instead of via `FcntlArg` enum.
   ([#541](https://github.com/nix-rust/nix/pull/541))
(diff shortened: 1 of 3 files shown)


kamalmarhubi:
@Susurrus can you explain this a tiny bit more? I am guessing that `revents` is populated by `poll` et al and so it doesn't make sense for us to set it?

kamalmarhubi:
Ah yeah, that looks to be the case:

```c
struct pollfd {
    int   fd;         /* file descriptor */
    short events;     /* requested events */
    short revents;    /* returned events */
};
```

@Susurrus could you add a tiny bit of context to the commit message, and add this to the _Changed_ section in the changelog?

Susurrus:
Re breaking changes that's already got an issue #458, which doesn't seem to ascribe any specific action for breaking changes, so I'll assume there's nothing else I need to do and defer higher-level discussion to that issue.

Susurrus:
@kamalmarhubi Should be good to go now!

fiveop:
@homu r+

PR #549, opened 2017-03-06
Don't consider 7th bit of signal to calculate signal for Linux platform.

chaosagent:
It's used for PTRACE_O_SYSGOOD, and creates a bad signal if on.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -81,7 +81,7 @@ mod status {
     }
 
     pub fn stop_signal(status: i32) -> Signal {
-        Signal::from_c_int((status & 0xFF00) >> 8).unwrap()
+        Signal::from_c_int((status & 0x7F00) >> 8).unwrap()
     }
 
     pub fn stop_additional(status: i32) -> c_int {


chaosagent:
We need a way to return whether the bit is set though

src/sys/wait.rs
```diff
@@ -81,7 +81,7 @@ mod status {
     }
 
     pub fn stop_signal(status: i32) -> Signal {
-        Signal::from_c_int((status & 0xFF00) >> 8).unwrap()
+        Signal::from_c_int((status & 0x7F00) >> 8).unwrap()
```
kamalmarhubi: Can you add a comment explaining the mask?

kamalmarhubi:
Need to double check the cross-platform-ness of this fix, but it seems reasonable. Could you add something to the changelog for this?

chaosagent:
I'll add the comments/changelog entry when I get a chance (soon).

I think the presence of a set 7th bit should be returned in the WaitStatus somehow, so that the caller can read it. What do you think would be the best way to represent this? I feel like adding it as a PtraceEvent would be the most straightforward to implement in a platform-specific way (both are exclusive to Linux). The architecture availability of this could be managed through enabling/disabling a particular value of the event enum used in PtraceEvent

kamalmarhubi:
I'm not sure yet, as I'm not really familiar with `ptrace`. I'm quite happy to take this fix now and we can file an issue to discuss how to expose it. Does that work?

Susurrus:
@chaosagent We'd like to close this issue in favor of #566. Can you see if that PR works for your needs as well and seems like the correct path forward?

Susurrus:
I believe this was addressed by #566, so I'm closing this one. Please reopen or file a new one if not.

PR #551, opened 2017-03-08
add support for `openat`

Mic92:


src/fcntl.rs
```diff
@@ -26,6 +26,15 @@ pub fn open<P: ?Sized + NixPath>(path: &P, oflag: OFlag, mode: Mode) -> Result<R
     Errno::result(fd)
 }
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn openat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P, oflag: OFlag, mode: Mode) -> Result<RawFd> {
+    let fd = try!(path.with_nix_path(|cstr| {
+        unsafe { libc::openat(dirfd, cstr.as_ptr(), oflag.bits(), mode.bits() as c_uint) }
```
kamalmarhubi: is the cast to `c_uint` necessary? The `Mode` flags type is defined as `mode_t` which should be the right type from libc.

kamalmarhubi:
Thanks for the PR! Could you also add an entry to the changelog for this?

Mic92:
done also extended platform support.

Mic92:
replaced by #552

Mic92:
replaced by #552

PR #552, opened 2017-03-08
add support for openat, fstatat, readlink, readlinkat

Mic92:


Mic92:
I cannot explain the errors happen on travis. Works fine locally.

kamalmarhubi:
Oh I'm also confused about the CI failures! Tests pass on my machine. I probably won't get to look into it in detail for a few more days. If you figure out out, please let me know what it was!

src/sys/stat.rs
```diff
@@ -109,3 +111,62 @@ pub fn fstat(fd: RawFd) -> Result<FileStat> {
 
     Ok(dst)
 }
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+mod consts {
+    use libc::c_int;
+
+    bitflags!(
+        flags FstatatFlag: c_int {
```
kamalmarhubi: Hm, not all of them use all these flags, but at least `linkat` does.

Mic92: I saw that nix unify some flags with the same name prefix, but I did not understand the reasoning here. Because some flags are only accepted by some system calls. For instance `AT_EMPTY_PATH` is not accepted by `fchownat` and even for the same function different operating systems differ in flags they accept for a system call wrapper. Would this not lead to subtle bugs, when porting applications?

Mic92:
https://github.com/rust-lang/libc/pull/549

src/fcntl.rs
```diff
@@ -53,6 +55,23 @@ pub fn openat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P, oflag: OFlag, mode: M
     Errno::result(fd)
 }
 
+pub fn readlink<'a, P: ?Sized + NixPath>(path: &P, buffer: &'a mut [u8]) -> Result<&'a OsStr> {
+    let res = try!(path.with_nix_path(|cstr| {
+        unsafe { libc::readlink(cstr.as_ptr(), buffer.as_mut_ptr() as *mut c_char, buffer.len() as size_t) }
+    }));
+
+    Errno::result(res).map(move |len| OsStr::from_bytes(&buffer[..(len as usize)]))
```
Mic92: ok, this solution is wrong: if the buffer is too small, the buffer is truncated without notifications. I see two options here:

1. Remove buffer from arguments and reallocate buffer internally until it fits the result of readlink (this is what the rust/go stdlib do)
2. return ENAMETOOLONG if len > buffer.len()

Mic92: I decided to use strategy 2 for the moment. If people want to use the first option, they could also use the rust std.

posborne:
Looks good, thanks! @homu r+

Issue #534, opened 2017-02-26
TRUST CI: powerpc-unknown-linux-gnu Compilation Errors


posborne:
https://travis-ci.org/berkowski/nix/jobs/205534698

Susurrus:
This is now fixed by work in #536.

PR #536, opened 2017-02-26
Merge New CI Infrastructure Based on Trust/Cross Into Master

posborne:
See discussion on https://github.com/nix-rust/nix/pull/528.  This PR is to track the final decision of whether we are ready to merge the new CI infrastructure (and related bug fixes) back into master.  Initially, we know that the branch is not ready to merge (as there are still failing tests).

This PR acts as a single point of reference to get the current status of how close we are to merging things back into master.  A number of issues with "TRUST CI" in the title have been created for the work that appears to be required to get us to the point where this branch can be merged.

Issue #535, opened 2017-02-26
TRUST CI: Nightly Builds for Linux/MacOS are Failing


posborne:
https://travis-ci.org/berkowski/nix/jobs/205534701
https://travis-ci.org/berkowski/nix/jobs/205534702

Susurrus:
This is resolved in #536. I'm closing this in favor of tracking that PR instead.

PR #553, opened 2017-03-08
Compile on all travis platforms

berkowski:
Addresses nix-rust/nix#534

One test failing:

```
failures:

---- sys::test_select::test_select stdout ----

	thread 'sys::test_select::test_select' panicked at 'assertion failed: `(left == right)` (left: `1`, right: `0`)', test/sys/test_select.rs:45

note: Run with `RUST_BACKTRACE=1` for a backtrace.

failures:

    sys::test_select::test_select

test result: FAILED. 57 passed; 1 failed; 0 ignored; 0 measured

error: test failed
```

This failure is consistent.  I suspect it may stem from the `nix` implementation of `fd_set`, but nothing immediate jumps out as a cause.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   implementations).
 
 ### Fixed
+- Fixed multiple issues compiling under different archetectures and OSes.
+  Now compiles on Linux/MIPS ([#538](https://github.com/nix-rust/nix/pull/538)),
+  `Linux/PPC` ([#553](https://github.com/nix-rust/nix/pull/553)), 
+  `MacOS/x86_64,i686` ([#553](https://github.com/nix-rust/nix/pull/553),
+  and `NetBSD/x64_64` ([#538](https://github.com/nix-rust/nix/pull/538).
 - Fixed multiple issues with Unix domain sockets on non-Linux OSes
   ([#474](https://github.com/nix-rust/nix/pull/415))
 - Fixed using kqueue with `EVFILT_USER` on FreeBSD
--- a/src/sys/socket/consts.rs
+++ b/src/sys/socket/consts.rs
@@ -420,7 +420,7 @@ mod test {
 
     #[cfg(target_os = "linux")]
     #[test]
-    pub fn test_linux_consts() {
+    pub fn test_general_linux_consts() {
         // TODO Figure out how to test new constants
         check_const!(
             SOL_IP,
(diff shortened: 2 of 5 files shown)


kamalmarhubi:
hrm not sure why CI didn't run on this....

I am going to trust that the constants are correct, and hopefully some day we'll actually tackle #341. Could you add something to the changelog for this, then I'll merge it. Thanks!

berkowski:
I usually enable testing on all branches and then disable before finalizing the PR.  I've re-enabled testing on all branches in `.travis.yml`.

What additions to the changelog were you thinking of?  I'm making only the minimum number of changes to get things compiling.  Switching over wholesale to `libc` defined constants is something that probably should be done but I'm trying to constrain the PR's to just what's needed at the moment.  I'm not sure how that fits nicely into a changelog line :)

berkowski:
Expanded the scope of the PR a bit to include additional changes I had lined up for #535

`nix` should now pass compile checks on all the current travis platforms.  It still fails a number of runtime tests.

@kamalmarhubi I'm happy to put some notes in the changelog here, or would you rather wait until  #536 is ready to go?

berkowski:
@kamalmarhubi Not sure if you still wanted some notes in the changelog for this or if you wanted to wait until #536 is ready.

Susurrus:
@berkowski This does look pretty good to me, especially since it fixed the powerpc stuff that we've already had other people attempt. Two things:

 * Can you remove the commit with changes to `.travis.yml`?
 * Can you add a changelog entry that says something as simple as "Fixed compilation on Linux/PowerPC, etc." where you replace "etc." with all known platforms fixed by this?
 * Can you also fix the compilation warning on test/sys/test_aio.rs:185:1 where `signaled` should be upper-case? I'd like to lump that in with this PR since it's a build fix.

So we'll defer on fixing the tests to #536.

berkowski:
I'll try to get to those changes this weekend.

berkowski:
Weekend comes early-ish.

Updated changelog, reverted `.travis.yml`, renamed the static variable in `test_aio.rs`

Susurrus:
Does the signaled test work? You changed the declaration to all-caps, but not the usages. But since `signaled` isn't a constant, let's not worry about capitalizing it. Can you drop the last commit? I think this is good to push once you've done that.

berkowski:
Ugh.. getting sloppy.  Sorry.  Removed `151f9c8`

Susurrus:
Thanks, @berkowski! Now let's see how this shakes out on Travis and we can move on to #536.

PR #556, opened 2017-03-15
Add various pty functions

Susurrus:
* grantpt
  * posix_openpt
  * ptsname/ptsname_r
  * unlockpt

I've refactored my code to use these functions and they work correctly at least in the non-error case. Only issue is what to do for failure in `ptsname` when converting strings. This should effectively never happen. I think maybe adding an `Unknown` to the `Error` type and returning that might be useful.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,8 @@ pub mod mount;
 #[cfg(target_os = "linux")]
 pub mod mqueue;
 
+pub mod pty;
+
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 pub mod poll;
 
@@ -92,6 +94,9 @@ pub type Result<T> = result::Result<T, Error>;
 pub enum Error {
     Sys(errno::Errno),
     InvalidPath,
+    /// The operation involved a conversion to Rust's native String type, which failed because the
+    /// string did not contain all valid UTF-8.
+    InvalidUtf8,
 }
 
 impl Error {
@@ -114,8 +119,9 @@ impl Error {
     /// Get the errno associated with this error
     pub fn errno(&self) -> errno::Errno {
         match *self {
-            Error::Sys(errno) => errno,
             Error::InvalidPath => errno::Errno::EINVAL,
+            Error::InvalidUtf8 => errno::Errno::UnknownErrno,
+            Error::Sys(errno) => errno,
         }
     }
 }
@@ -124,10 +130,15 @@ impl From<errno::Errno> for Error {
     fn from(errno: errno::Errno) -> Error { Error::from_errno(errno) }
 }
 
+impl From<std::string::FromUtf8Error> for Error {
+    fn from(_: std::string::FromUtf8Error) -> Error { Error::InvalidUtf8 }
+}
+
 impl error::Error for Error {
     fn description(&self) -> &str {
         match self {
             &Error::InvalidPath => "Invalid path",
+            &Error::InvalidUtf8 => "Invalid UTF-8 string",
             &Error::Sys(ref errno) => errno.desc(),
         }
     }
@@ -137,6 +148,7 @@ impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             &Error::InvalidPath => write!(f, "Invalid path"),
+            &Error::InvalidUtf8 => write!(f, "Invalid UTF-8 string"),
             &Error::Sys(errno) => write!(f, "{:?}: {}", errno, errno.desc()),
         }
     }
@@ -146,6 +158,7 @@ impl From<Error> for io::Error {
     fn from(err: Error) -> Self {
         match err {
             Error::InvalidPath => io::Error::new(io::ErrorKind::InvalidInput, err),
+            Error::InvalidUtf8 => io::Error::new(io::ErrorKind::Other, err),
             Error::Sys(errno) => io::Error::from_raw_os_error(errno as i32),
         }
     }
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -1,0 +1,164 @@
+//! Create master and slave virtual pseudo-terminals (PTYs)
+
+use std::ffi::CStr;
+use std::mem;
+use std::os::unix::prelude::*;
+
+use libc;
+
+use {Error, fcntl, Result};
+
+/// Representation of the Master device in a master/slave pty pair
+///
+/// While this datatype is a thin wrapper around `RawFd`, it enforces that the available PTY
+/// functions are given the correct file descriptor. Additionally this type implements `Drop`,
+/// so that when it's consumed or goes out of scope, it's automatically cleaned-up.
+#[derive(Debug)]
+pub struct PtyMaster(RawFd);
+
+impl AsRawFd for PtyMaster {
+    fn as_raw_fd(&self) -> RawFd {
+        self.0
+    }
+}
+
+impl IntoRawFd for PtyMaster {
+    fn into_raw_fd(self) -> RawFd {
+        let fd = self.0;
+        mem::forget(self);
+        fd
+    }
+}
+
+impl Drop for PtyMaster {
+    fn drop(&mut self) {
+        // Errors when closing are ignored because we don't actually know if the file descriptor
+        // was closed. If we retried, it's possible that descriptor was reallocated in the mean
+        // time and the wrong file descriptor could be closed.
+        let _ = ::unistd::close(self.0);
+    }
+}
+
+/// Grant access to a slave pseudoterminal (see
+/// [grantpt(3)](http://man7.org/linux/man-pages/man3/grantpt.3.html))
+///
+/// `grantpt()` changes the mode and owner of the slave pseudoterminal device corresponding to the
+/// master pseudoterminal referred to by `fd`. This is a necessary step towards opening the slave.
+#[inline]
+pub fn grantpt(fd: &PtyMaster) -> Result<()> {
+    if unsafe { libc::grantpt(fd.as_raw_fd()) } < 0 {
+        return Err(Error::last().into());
+    }
+
+    Ok(())
+}
+
+/// Open a pseudoterminal device (see
+/// [posix_openpt(3)](http://man7.org/linux/man-pages/man3/posix_openpt.3.html))
+///
+/// `posix_openpt()` returns a file descriptor to an existing unused pseuterminal master device.
+///
+/// # Examples
+///
+/// A common use case with this function is to open both a master and slave PTY pair. This can be
+/// done as follows:
+///
+/// ```
+/// use std::path::Path;
+/// use nix::fcntl::{O_RDWR, open};
+/// use nix::pty::*;
+/// use nix::sys::stat;
+///
+/// # #[allow(dead_code)]
+/// # fn run() -> nix::Result<()> {
+/// // Open a new PTY master
+/// let master_fd = posix_openpt(O_RDWR)?;
+///
+/// // Allow a slave to be generated for it
+/// grantpt(&master_fd)?;
+/// unlockpt(&master_fd)?;
+///
+/// // Get the name of the slave
+/// let slave_name = ptsname(&master_fd)?;
+///
+/// // Try to open the slave
+/// # #[allow(unused_variables)]
+/// let slave_fd = open(Path::new(&slave_name), O_RDWR, stat::Mode::empty())?;
+/// # Ok(())
+/// # }
+/// ```
+#[inline]
+pub fn posix_openpt(flags: fcntl::OFlag) -> Result<PtyMaster> {
+    let fd = unsafe {
+        libc::posix_openpt(flags.bits())
+    };
+
+    if fd < 0 {
+        return Err(Error::last().into());
+    }
+
+    Ok(PtyMaster(fd))
+}
+
+/// Get the name of the slave pseudoterminal (see
+/// [ptsname(3)](http://man7.org/linux/man-pages/man3/ptsname.3.html))
+///
+/// `ptsname()` returns the name of the slave pseudoterminal device corresponding to the master
+/// referred to by `fd`. Note that this function is *not* threadsafe. For that see `ptsname_r()`.
+///
+/// This value is useful for opening the slave pty once the master has already been opened with
+/// `posix_openpt()`.
+#[inline]
+pub fn ptsname(fd: &PtyMaster) -> Result<String> {
+    let name_ptr = unsafe { libc::ptsname(fd.as_raw_fd()) };
+    if name_ptr.is_null() {
+        return Err(Error::last().into());
+    }
+
+    let name = unsafe {
+        CStr::from_ptr(name_ptr)
+    };
+    Ok(name.to_string_lossy().into_owned())
+}
+
+/// Get the name of the slave pseudoterminal (see
+/// [ptsname(3)](http://man7.org/linux/man-pages/man3/ptsname.3.html))
+///
+/// `ptsname_r()` returns the name of the slave pseudoterminal device corresponding to the master
+/// referred to by `fd`. This is the threadsafe version of `ptsname()`, but it is not part of the
+/// POSIX standard and is instead a Linux-specific extension.
+///
+/// This value is useful for opening the slave ptty once the master has already been opened with
+/// `posix_openpt()`.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub fn ptsname_r(fd: &PtyMaster) -> Result<String> {
+    let mut name_buf = vec![0u8; 64];
+    let name_buf_ptr = name_buf.as_mut_ptr() as *mut libc::c_char;
+    if unsafe { libc::ptsname_r(fd.as_raw_fd(), name_buf_ptr, name_buf.capacity()) } != 0 {
+        return Err(Error::last().into());
+    }
+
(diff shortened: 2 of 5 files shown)


src/pty.rs
```diff
@@ -0,0 +1,72 @@
...
+        return Err(Error::last().into());
+    }
+
+    Ok(fd)
+}
+
+pub fn ptsname(fd: RawFd) -> Result<String> {
+    let name_ptr = unsafe { libc::ptsname(fd) };
+    if name_ptr.is_null() {
+        return Err(Errno::ENOTTY.into());
```
Mic92: Is Err(Error::Sys(Errno::last())) not safer here? I mean according to standard ENOTTY is the only error, but you can never be sure.

src/pty.rs
```diff
@@ -0,0 +1,72 @@
...
+            // FIXME: Unsure what error I should report here.
+            return Err(Error::last().into());
+        }
+    };
+    Ok(name)
+}
+
+pub fn unlockpt(fd: RawFd) -> Result<()> {
+    if unsafe { libc::unlockpt(fd) } < 0 {
+        return Err(Error::last().into());
```
Mic92: Nix uses as far as I know `Err(Error::Sys(Errno::last()))` to return errno related errors.

src/pty.rs
Susurrus: Yeah, that's probably the correct thing to do, though I do think they'll end up with the same functionality. I'll change it.

Susurrus:
@Mic92 Also do you have any comment on the two `FIXME`s I have in there? I didn't see a way to return a generic unknown error.

src/pty.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub fn ptsname(fd: RawFd) -> Result<String> {
+    let name_ptr = unsafe { libc::ptsname(fd) };
+    if name_ptr.is_null() {
+        return Err(Errno::ENOTTY.into());
+    }
+
+    let name_cstr = unsafe { CStr::from_ptr(name_ptr) };
+    let name = match name_cstr.to_str() {
```
Mic92: It is probably safe to ignore utf8 validation errors as we get the string from the operating system:
see https://github.com/nix-rust/nix/blob/master/src/sys/utsname.rs#L65

Mic92: If it is not safe, then it is better to return the value as it is.

Susurrus:
@Mic92 I think I've resolved both issues. Let me know what you think. I left them as separate commits so they're easier to review, but I'll squash them before they're pulled.

src/pty.rs
```diff
@@ -52,15 +47,10 @@ pub fn ptsname_r(fd: RawFd) -> Result<String> {
...
-        Ok(s) => s.to_string(),
-        Err(_) => {
-            // FIXME: Unsure what error I should report here.
-            return Err(Error::last().into());
-        }
+    let name = unsafe {
+        CStr::from_ptr(name_buf.as_ptr())
     };
-    Ok(name)
+    Ok(name.to_string_lossy().into_owned())
```
Mic92: ```....replacing any invalid UTF-8 sequences with U+FFFD REPLACEMENT CHARACTER...```
Is the unmodified value of this function not meant to be as an input for functions like open(2)?

Susurrus: We should never receive invalid UTF-8 sequences from the OS, so that should never happen. I'd prefer to use `from_utf8_unchecked` (as you mentioned in your comment), which doesn't do any error detection, but I couldn't figure out how to use it. And note that using `from_utf8_unchecked` has the exact same problem, but instead of fixing invalid UTF-8 strings it will just crash later when it's attempted to be decoded as proper UTF-8.

Basically if we use the `String` type as the output from this function, it **must** be valid UTF-8 regardless of how we convert it. So the question really is, does `ptsname*()` **always** return valid UTF-8. I think the answer is yes, but I don't really know.

Mic92: Just wanted to note that gethostname returns `CStr` so to be consistent I would either change both to `String` or both to `CStr`.

Susurrus: Yeah, I would think that `gethostname` should return a `Result<String>` instead of returning it in an output argument as a `CStr`. @posborne Any comments here?

Susurrus:
@nix-rust/nix-maintainers Any comments on the APIs I've exposed here. Unlike with `gethostname` I'm exposing a `String` type, though a `&str` might be more appropriate. I do think we should stop exposing `CStr` types in our APIs as I see `nix` as being the interface between Rust and libc and that conversion as part of it.

src/pty.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use {Error, fcntl, Result};
+
+/// Grant access to a slave pseudoterminal (see
+/// [grantpt(3)](http://man7.org/linux/man-pages/man3/grantpt.3.html))
+///
+/// `grantpt()` changezs the mode and owner of the slave pseudoterminal device corresponding to the
+/// master pseudoterminal referred to by `fd`. This is a necessary step towards opening the slave.
+#[inline]
+pub fn grantpt(fd: RawFd) -> Result<()> {
```
asomers: How about creating a Newtype like `struct PseudoTerminal(RawFd)` to provide some static type checking on these functions?  It could also close itself on `Drop`.

src/pty.rs
```diff
@@ -0,0 +1,128 @@
...
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub fn ptsname_r(fd: RawFd) -> Result<String> {
+    let mut name_buf = [0 as libc::c_char; 64];
+    if unsafe { libc::ptsname_r(fd, name_buf.as_mut_ptr(), name_buf.len()) } != 0 {
+        return Err(Error::last().into());
+    }
+
+    let name = unsafe {
+        CStr::from_ptr(name_buf.as_ptr())
```
asomers: The subsequent `Cow::into_owned()` will perform a data copy.  How about something like this instead?
```
let mut name_buf = OsString::with_capacity(64);
libc::ptsname(fd, name.to_str().as_ptr(), 64);
return name_buf.into_string()
```

src/pty.rs
```diff
@@ -0,0 +1,128 @@
...
+/// `posix_openpt()`.
+#[inline]
+pub fn ptsname(fd: RawFd) -> Result<String> {
+    let name_ptr = unsafe { libc::ptsname(fd) };
+    if name_ptr.is_null() {
+        return Err(Error::last().into());
+    }
+
+    let name = unsafe {
+        CStr::from_ptr(name_ptr)
```
asomers: How about a test that calls `ptsname` on two separate pseudoterminals, and checks that their names are not the same?  That would enforce that this function performs a data copy (as I think it probably should).

test/test_pty.rs
```diff
@@ -0,0 +1,41 @@
+use std::path::Path;
+use nix::fcntl::{O_RDWR, open};
+use nix::pty::*;
+use nix::sys::stat;
+
+/// Test equivalence of `ptsname` and `ptsname_r`
+#[test]
+#[cfg(any(target_os = "android", target_os = "linux"))]
+fn test_ptsname() {
```
asomers: The test name is not very accurate.

test/test_pty.rs
```diff
@@ -0,0 +1,41 @@
...
+    grantpt(master_fd).unwrap();
+    unlockpt(master_fd).unwrap();
+
+    // Get the name of the slave
+    let slave_name = ptsname(master_fd).unwrap();
+
+    // Open the slave device
+    let slave_fd = open(Path::new(&slave_name), O_RDWR, stat::Mode::empty()).unwrap();
+    assert!(slave_fd > 0);
+}
```
asomers: This test leaks its file descriptors.

test/test_pty.rs
```diff
@@ -0,0 +1,41 @@
...
+fn test_ptsname() {
+    // Open a new PTTY master
+    let master_fd = posix_openpt(O_RDWR).unwrap();
+    assert!(master_fd > 0);
+
+    // Get the name of the slave
+    let slave_name = ptsname(master_fd).unwrap();
+    let slave_name_r = ptsname_r(master_fd).unwrap();
+    assert_eq!(slave_name, slave_name_r);
+}
```
asomers: This test leaks its file descriptors, too.

src/pty.rs
Susurrus: Two things: Is `OsString` the right type rather than `CString`? The docs don't clarify what the difference between the two really is, so I'm not certain.

Also, that code you provided doesn't compile. because `name.to_str().as_ptr()` doesn't return a mutable pointer and I don't see a way to get one. Additionally, I'm not certain how to map the error types. I'm thinking we should add to `nix::Error` a `StringConversion` type and map the `into_string()` failure to that, otherwise I think this should be an `unwrap()` or `expect()` since I don't think this will ever fail.

src/pty.rs
Susurrus: Wouldn't you instead want to call `ptsname()` on the same ptty and check that they don't point to the same underlying buffer? I'm unclear what the test you suggested actually tests.

src/pty.rs
Susurrus: Hmm, interesting. Would this be as simple as implementing that type and also `impl From<RawFd>` for it? We'd need to make sure that this can be easily used both with `std` and other `nix` functions. Additionally, is this sound in the context of #594?

asomers: I don't think it should implement `From<RawFd>`, because that would make it too easy to accidentally convert an inappropriate value.  Conversion should always be explicit.  In fact, are there any other Rust functions, in `nix` or in `std`, that deal with pseudoterminals?

src/pty.rs
asomers: Yeah, that could do it.  I'm really just trying to think up ways to improve the coverage.

src/pty.rs
asomers: Take a look at `getcwd`in unistd.rs.  It actually uses a `Vec` as a buffer, before transforming it into an `OsString`

Susurrus: Is that what you'd recommend doing instead of what you originally suggested?

src/pty.rs
Susurrus: Yeah, I agree, maybe just `AsRawFd` is a better trait instead? But no, nothing else really deals with this. I have some code in my serialport-rs library [here](https://gitlab.com/susurrus/serialport-rs/blob/master/src/posix/tty.rs#L239) that deals with this same thing, and that's like the only other thing working with safe ptty abstractions.

src/pty.rs
asomers: My original suggestion was just a guess, a starting point.  I didn't know exactly how to make it work.  But `getcwd` presents exactly the same problem as this function, so its solution will probably work.

src/pty.rs
asomers: `AsRawFd` sounds good.

Susurrus: So in continuing to think about this, I don't think it makes sense to provide an abstraction over `RawFd`. The use case of these ttys is generally what you can see in `serialport-rs`. Most commonly you call `read` and `write`, both of which take a `RawFd`. Additionally, the other functions I'm adding here all take `RawFd`. So every call you'd be doing `fn(port.as_raw_fd())`, so it's just a needless abstraction that doesn't actually add any safety. I think it's the role of a library like `serialport-rs` to provide abstractions on top of nix should it be desired.

asomers: What if the newtype implemented `Into<RawFd>`?  Then `fn` could automatically coerce it, as long as `fn`'s argument is a `T: Into<RawFd>`

Susurrus: But if you implement `Into` you also implement `From`, that's what my comment was about before. It's also bidirectional, so we're back to where we started. And you'd still need to use `.into()` everywhere since not all functions use `T: Into<RawFd>`. Since the topic of changing function signatures came up, there are 168 functions in `libc` that have fd's as inputs (as `c_int`, not `RawFd`) and 88 functions in `nix` that have `RawFd` inputs.

asomers: Actually, `From` implies `Into`, but not the other way around.  If this newtype implements `Into<RawFd>` , that does not automatically create an implementation for `From<Newtype>` on `RawFd`.  Do you think that typing `.into()` is too much work to justify the free type checking on the pty functions?  If the newtype implements `Drop`, that could save some typing too, because you wouldn't have to automatically close the ptys.

Susurrus:
I'm still working on how to best return a `String` with the least number of data copies (gonna reach out on IRC tomorrow). But for now I've pushed an updated version that adds a datatype. I can use it in my `serialport-rs` lib pretty ergonomically.

One question I have is if errors when calling `close` on a `PtyMaster` fd should it panic or just silently ignore the error? The `signalfd` module does, but I don't know if that's how we should do things, though I'm doing the same thing here.

Susurrus:
[Here's the commit to `serialport-rs`](https://gitlab.com/susurrus/serialport-rs/commit/49eb82142deeb7406f87b8983c9ecb460da44870) that adjusts it for these changes in `nix`. In general I think it's pretty clean looking.

test/test_pty.rs
```diff
@@ -0,0 +1,90 @@
...
+use nix::fcntl::{O_RDWR, open};
+use nix::pty::*;
+use nix::sys::stat;
+use nix::unistd::close;
+
+/// Test equivalence of `ptsname` and `ptsname_r`
+#[test]
+#[cfg(any(target_os = "android", target_os = "linux"))]
+fn test_ptsname_equivalence() {
+    // Open a new PTTY master
```
asomers: s/PTTY/PTY here and below

test/test_pty.rs
```diff
@@ -0,0 +1,90 @@
...
+fn test_ptsname_copy() {
+    // Open a new PTTY master
+    let master_fd = posix_openpt(O_RDWR).unwrap();
+    assert!(master_fd.as_raw_fd() > 0);
+
+    // Get the name of the slave
+    let slave_name1 = ptsname(&master_fd).unwrap();
+    let slave_name2 = ptsname(&master_fd).unwrap();
+    assert!(slave_name1 == slave_name2);
+    assert!(slave_name1.as_ptr() != slave_name2.as_ptr());
```
asomers: This assertion probably deserves a comment.  It's not obvious why a consumer should care about this behavior.

asomers:
The new type looks nice, and it fixes all of the file descriptor leaks in the tests.  The integration into serialport-rs looks good too.

I have conflicting feelings about errors during `drop`.  On the one hand, the most likely error is an `EBADF`, which suggests a programming bug, so the user should be informed.  OTOH, since `PtyMaster` implements `AsRawFd`, the library can't really assume that the user hasn't already closed it.  So I suppose we should ignore errors.  But that raises another question: Will `PtyMaster.into_raw_fd()` invoke `drop`?  If so, then we're screwed.  I don't think so, because `std::net::TcpStream` implements both traits.  But it would be nice to verify that with a test.

Susurrus:
Yes, we need to explicitly forget the type. This is actually done in `std::net::TcpStream` via the [`into_raw()`](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/fd.rs#L50-L55) call. Additionally we ignore any errors when closing for the same reason they do [in std](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/fd.rs#L247) and I added as similar comment.

Alright, I think this really should be everything that we need.

asomers:
Did you ever figure out a way to reduce the data copies in `ptsname_r`?

Susurrus:
Whoops! I knew I was forgetting something. I'll hop on IRC again today and see if I can figure out the "right" way to do this.

Susurrus:
Popped onto IRC and found a reasonable implementation that should be pretty efficient. It's still not clear if `String` is the right type to return, but at least this will error hard if there's every a problem rather than us assuming it's safe. It did require me expanding the `Error` type, but I think it's reasonable.

src/lib.rs
```diff
@@ -146,7 +158,9 @@ impl From<Error> for io::Error {
     fn from(err: Error) -> Self {
         match err {
             Error::InvalidPath => io::Error::new(io::ErrorKind::InvalidInput, err),
+            Error::InvalidUtf8 => io::Error::new(io::ErrorKind::Other, err),
             Error::Sys(errno) => io::Error::from_raw_os_error(errno as i32),
+
```
asomers: I don't think you meant to leave a space here

src/pty.rs
```diff
@@ -0,0 +1,166 @@
...
+/// POSIX standard and is instead a Linux-specific extension.
+///
+/// This value is useful for opening the slave ptty once the master has already been opened with
+/// `posix_openpt()`.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub fn ptsname_r(fd: &PtyMaster) -> Result<String> {
+    let mut name_buf = vec![0u8; 64];
+    let name_buf_ptr = name_buf.as_mut_ptr() as *mut libc::c_char;
+    println!("{}/{}", name_buf.len(), name_buf.capacity());
```
asomers: Please remove this debugging code

src/pty.rs
```diff
@@ -0,0 +1,166 @@
...
+    if unsafe { libc::ptsname_r(fd.as_raw_fd(), name_buf_ptr, name_buf.capacity()) } != 0 {
+        return Err(Error::last().into());
+    }
+    println!("{}/{}", name_buf.len(), name_buf.capacity());
```
asomers: This one too.

Susurrus:
Sorry, I thought I did one last look over of it, must have forgot! Okay, maybe now this is looking finalized.

PR #558, opened 2017-03-17
add support for dirent related function

Mic92:
This pr is currently work in progress, as it depends on an outstanding pull request for [fdopendir](https://github.com/rust-lang/libc/pull/550) in libc. I wanted to know if this function and DIR wrapper I added  fit into the scope of the nix crate.

### DIFF
--- a/src/dirent.rs
+++ b/src/dirent.rs
@@ -1,0 +1,135 @@
+//! Directory Stream functions
+//!
+//! [Further reading and details on the C API](http://man7.org/linux/man-pages/man3/opendir.3.html)
+
+use {Result, Error, Errno, NixPath};
+use errno;
+use libc::{self, DIR, c_long};
+use std::convert::{AsRef, Into};
+use std::ffi::CStr;
+use std::mem;
+
+#[cfg(any(target_os = "linux"))]
+use libc::ino64_t;
+
+#[cfg(any(target_os = "android"))]
+use libc::ino_t as ino64_t;
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use libc::{dirent64, readdir64};
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+use libc::{dirent as dirent64, ino_t as ino64_t, readdir as readdir64};
+
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+use std::os::unix::io::RawFd;
+
+/// Directory Stream object
+pub struct DirectoryStream(*mut DIR);
+
+impl AsRef<DIR> for DirectoryStream {
+    fn as_ref(&self) -> &DIR {
+        unsafe { &*self.0 }
+    }
+}
+
+/// Consumes directory stream and return underlying directory pointer.
+///
+/// The pointer must be deallocated manually using `libc::closedir`
+impl Into<*mut DIR> for DirectoryStream {
+    fn into(self) -> *mut DIR {
+        let dirp = self.0;
+        mem::forget(self);
+        dirp
+    }
+}
+
+impl Drop for DirectoryStream {
+    fn drop(&mut self) {
+        unsafe { libc::closedir(self.0) };
+    }
+}
+
+/// A directory entry
+pub struct DirectoryEntry<'a>(&'a dirent64);
+
+impl<'a> DirectoryEntry<'a> {
+    /// File name
+    pub fn name(&self) -> &CStr {
+        unsafe{
+            CStr::from_ptr(self.0.d_name.as_ptr())
+        }
+    }
+
+    /// Inode number
+    pub fn inode(&self) -> ino64_t {
+        #[cfg(not(any(target_os = "freebsd", target_os = "netbsd", target_os="dragonfly")))]
+        return self.0.d_ino;
+        #[cfg(any(target_os = "freebsd", target_os = "netbsd", target_os="dragonfly"))]
+        return self.0.d_fileno;
+    }
+}
+
+impl<'a> AsRef<dirent64> for DirectoryEntry<'a> {
+    fn as_ref(&self) -> &dirent64 {
+        self.0
+    }
+}
+
+/// Opens a directory stream corresponding to the directory name.
+///
+/// The stream is positioned at the first entry in the directory.
+pub fn opendir<P: ?Sized + NixPath>(name: &P) -> Result<DirectoryStream> {
+    let dirp = try!(name.with_nix_path(|cstr| unsafe { libc::opendir(cstr.as_ptr()) }));
+    if dirp.is_null() {
+        Err(Error::last().into())
+    } else {
+        Ok(DirectoryStream(dirp))
+    }
+}
+
+/// Returns directory stream corresponding to the open file descriptor `fd`
+///
+/// After a successful call to this function, `fd` is used internally by
+/// the implementation, and should not otherwise be used by the application
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+pub fn fdopendir(fd: RawFd) -> Result<DirectoryStream> {
+    let dirp = unsafe { libc::fdopendir(fd) };
+    if dirp.is_null() {
+        Err(Error::last().into())
+    } else {
+        Ok(DirectoryStream(dirp))
+    }
+}
+
+/// Returns the next directory entry in the directory stream.
+///
+/// It returns `Some(None)` on reaching the end of the directory stream.
+pub fn readdir<'a>(dir: &'a mut DirectoryStream) -> Result<Option<DirectoryEntry>> {
+    let dirent = unsafe {
+        Errno::clear();
+        readdir64(dir.0)
+    };
+    if dirent.is_null() {
+        match Errno::last() {
+            errno::UnknownErrno => Ok(None),
+            _ => Err(Error::last().into()),
+        }
+    } else {
+        Ok(Some(DirectoryEntry(unsafe { &*dirent })))
+    }
+}
+
+/// Sets the location in the directory stream from which the next `readdir` call will start.
+///
+/// The `loc` argument should be a value returned by a previous call to `telldir`
+#[cfg(not(any(target_os = "android")))]
+pub fn seekdir<'a>(dir: &'a mut DirectoryStream, loc: c_long) {
+    unsafe { libc::seekdir(dir.0, loc) };
+}
+
+/// Returns the current location associated with the directory stream.
+#[cfg(not(any(target_os = "android")))]
+pub fn telldir<'a>(dir: &'a mut DirectoryStream) -> c_long {
+    unsafe { libc::telldir(dir.0) }
+}
--- a/test/test_dirent.rs
+++ b/test/test_dirent.rs
@@ -1,0 +1,50 @@
+use libc;
+use nix::dirent::{self, opendir, readdir, seekdir, telldir, DirectoryStream};
+use std::path::Path;
+use tempdir::TempDir;
+
+fn test_readdir<Open>(open_fn: Open)
+    where Open: Fn(&Path) -> DirectoryStream
+{
+    let tempdir = TempDir::new("nix-test_readdir")
+        .unwrap_or_else(|e| panic!("tempdir failed: {}", e));
+    let mut dir = open_fn(tempdir.path());
+    let first_inode = readdir(&mut dir)
+        .unwrap()
+        .unwrap()
+        .inode();
+
+    let pos = telldir(&mut dir);
+
+    let second_inode = readdir(&mut dir)
+        .unwrap()
+        .unwrap()
+        .inode();
+    seekdir(&mut dir, pos);
+
+    let second_inode_again = readdir(&mut dir)
+        .unwrap()
+        .unwrap()
+        .inode();
+
+    assert_ne!(first_inode, second_inode);
+    assert_eq!(second_inode, second_inode_again);
+
+    // end of directory
+    assert!(readdir(&mut dir).unwrap().is_none());
+
+    unsafe { libc::closedir(dir.into()) };
+}
+
+#[test]
+fn test_opendir() {
+    test_readdir(|path| opendir(path).unwrap());
+}
+
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+#[test]
+fn test_fdopendir() {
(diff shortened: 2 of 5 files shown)


src/dirent.rs
```diff
@@ -0,0 +1,40 @@
...
+use libc::{self, DIR};
+use std::os::unix::io::RawFd;
+use errno;
+
+pub struct Dir {
+    pub handle: *mut DIR,
+}
+
+impl Dir {
+    pub fn read(&mut self) -> Result<Option<&libc::dirent>> {
```
Mic92: borrowing Dir is required here as the dirent reference might get overwritten by subsequent calls.

src/dirent.rs
```diff
@@ -0,0 +1,40 @@
+use {Result, Error, Errno};
+use libc::{self, DIR};
+use std::os::unix::io::RawFd;
+use errno;
+
+pub struct Dir {
+    pub handle: *mut DIR,
```
Susurrus: I don't think this field should be public. Instead I'd propose implementing `AsRawFd` and `IntoRawFd` instead.

Mic92: Maybe `Into<*mut DIR>` because AsRawFd is for file descriptor.

Susurrus: Sorry, you're right. `Into<*mut DIR>` is what you'd want to do.

There is also `dirfd` which you could use to implement the `*RawFd` traits if you wanted to.

Cargo.toml
```diff
@@ -21,7 +21,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc" }
+libc = { git = "https://github.com/Mic92/libc", branch = "master" }
```
kamalmarhubi: Will wait till this hits libc upstream.

src/dirent.rs
```diff
@@ -0,0 +1,40 @@
...
+            Errno::clear();
+            libc::readdir(self.handle)
+        };
+        if dirent.is_null() {
+            match Errno::last() {
+                errno::UnknownErrno => Ok(None),
+                _ => Err(Error::last().into()),
+            }
+        } else {
+            Ok(Some(unsafe{&*dirent}))
```
kamalmarhubi: Could you annotate the lifetime in the signature so we're sure that this reference has the correct lifetime:
```
pub fn read<'a>(&'a mut self) -> Result<Option<&'a libc::dirent>>

File: src/dirent.rs (hunk 2 above)
Comment (Mic92) [inline src/dirent.rs]: I currently leave it private as I have implemented all functions, where the dir pointer is useful. If it turns out, that I missed a use case, it can be still added.

File: src/dirent.rs (hunk 4 above)
Comment (Mic92) [inline src/dirent.rs]: I switched back to `readdir`

**Comment by Mic92** [comment]
added seekdir

File: src/dirent.rs [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+}
+
+pub fn readdir<'a>(dir: &'a mut Dir) -> Result<Option<&'a libc::dirent>> {
+    let dirent = unsafe {
+        Errno::clear();
+        libc::readdir(dir.handle)
+    };
+    if dirent.is_null() {
+        match Errno::last() {
+            errno::UnknownErrno => Ok(None),
```
kamalmarhubi: Is this right? It seems that on error, `-1` is returned, not `NULL`.

kamalmarhubi:
Ah just noted that @Susurrus had mentioned `{As,Into}RawFd` already. Oh and there's also the changelog entry! :-)

src/dirent.rs
Mic92: from readdir(3):

```
 If the end of the directory stream is reached,  NULL  is  returned  and
 errno  is  not changed.  If an error occurs, NULL is returned and errno
 is set appropriately.  To distinguish end of stream and from an  error,
 set  errno to zero before calling readdir() and then check the value of
 errno if NULL is returned.
```

Mic92:
I would leave adding the other functions open to someone else, as I currently only add functions, I need to proceed my project.

src/dirent.rs
kamalmarhubi: Right you are! I must have had one of the other functions' man pages open.

kamalmarhubi:
@Mic92 that's fair. Could you look at the CI failures? They seem legitimate.

Mic92:
Order of `..` and `.` seems to differ depending on platform

src/dirent.rs
```diff
@@ -0,0 +1,58 @@
...
+pub fn fdopendir(fd: RawFd) -> Result<Dir> {
+    let dirp = unsafe { libc::fdopendir(fd) };
+    if dirp.is_null() {
+        Err(Error::last().into())
+    } else {
+        Ok(Dir { handle: dirp })
+    }
+}
+
+pub fn readdir<'a>(dir: &'a mut Dir) -> Result<Option<&'a libc::dirent>> {
```
Mic92: To obtain `d_name` I use:

```
let name = OsStr::from_bytes(unsafe {
    &mem::transmute::<[i8; 256], [u8; 256]>(entry.d_name)
});
```

any idea how I could how add a convenience wrapper to this library?

Mic92:
This is ready for review.

Susurrus:
@Mic92 Can you add comments for every function based on their man pages? We're trying to have complete documentation coverage for all PRs moving forward. You can see where things are lacking when you run `cargo doc --no-deps --open`

src/dirent.rs
```diff
@@ -0,0 +1,93 @@
...
+use errno;
+use libc::{self, DIR, c_long};
+use std::convert::{AsRef, Into};
+use std::ffi::CStr;
+use std::mem;
+
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+use std::os::unix::io::RawFd;
+
+pub struct Dir(*mut DIR);
```
Susurrus: Maybe name this `DirStream` so it's a little more clear?

Mic92: I prefer `Dir` as it it closer to the underlying Api.

Mic92:
depends on https://github.com/rust-lang/libc/pull/647

test/test_dirent.rs
```diff
@@ -0,0 +1,42 @@
+use nix::dirent::{self, opendir, readdir, seekdir, telldir, Dir};
+use std::path::Path;
+use tempdir::TempDir;
+
+fn test_readdir<OPEN>(open_fn: OPEN)
```
Susurrus: Please don't use all-caps here for a type name. Maybe just use `O`.

test/test_dirent.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut dir = open_fn(tempdir.path());
+    let letter1 = readdir(&mut dir)
+                       .unwrap()
+                       .unwrap()
+                       .name()
+                       .to_bytes()[0];
+    assert_eq!(letter1, '.' as u8);
+
+    let pos = telldir(&mut dir);
+    seekdir(&mut dir, pos); // no-op
```
Susurrus: Could we have a test where we seek back a directory and check that the same directory is read twice (so that we no `seekdir` worked? I'd like to have a non-nop test of it.

src/dirent.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+impl<'a> AsRef<libc::dirent> for Dirent<'a> {
+    fn as_ref(&self) -> &libc::dirent {
+        self.0
+    }
+}
+
+/// Opens a directory stream corresponding to the directory name.
```
Susurrus: Can you run `cargo doc --no-deps --open` locally with this code and make sure the first sentence shows up nicely on the page listing all the functions here? I'm certain that some of these sentences are too long and that you need an empty `///` on the next line to make them show up nicely.

src/dirent.rs
```diff
@@ -0,0 +1,116 @@
...
+impl<'a> Dirent<'a> {
+    /// File name
+    pub fn name(&self) -> &CStr {
+        unsafe{
+            CStr::from_ptr(self.0.d_name.as_ptr())
+        }
+    }
+
+    /// Inode number
+    pub fn ino(&self) -> libc::ino_t {
```
Susurrus: Any reason not to call this `inode` here for a little more clarity?

src/dirent.rs
```diff
@@ -0,0 +1,116 @@
...
+use errno;
+use libc::{self, DIR, c_long};
+use std::convert::{AsRef, Into};
+use std::ffi::CStr;
+use std::mem;
+
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+use std::os::unix::io::RawFd;
+
+/// Directory Stream object
```
Susurrus: Why don't we call this `DirectoryStream` or `DirStream` instead of `Dir`?

Susurrus:
This is looking pretty good. Mostly some minor changes here. I would love to have more comprehensive module documentation, but I think linking to the man pages should be fine.

src/dirent.rs
Mic92: also renamed `Dirent` to `DirectoryEntry` for consistency.

src/dirent.rs
```diff
@@ -0,0 +1,124 @@
...
+impl AsRef<DIR> for DirectoryStream {
+    fn as_ref(&self) -> &DIR {
+        unsafe { &*self.0 }
+    }
+}
+
+/// Consumes directory stream and return underlying directory pointer.
+///
+/// The pointer must be deallocated manually using `libc::closedir`
+impl Into<*mut DIR> for DirectoryStream {
```
Susurrus: A convention in `std` is to use a method named `into_raw` here when you're unwrapping a type. I'd suggest using that here instead of `Into<*mut DIR>`. I don't see this used anywhere, however. Can you either remove it if it's unnecessary, or if it's necessary add it to the test or use it in an example.

src/dirent.rs
```diff
@@ -0,0 +1,124 @@
...
+    } else {
+        Ok(Some(DirectoryEntry(unsafe { &*dirent })))
+    }
+}
+
+/// Sets the location in the directory stream from which the next `readdir` call will start.
+///
+/// The `loc` argument should be a value returned by a previous call to `telldir`
+#[cfg(not(any(target_os = "android")))]
+pub fn seekdir<'a>(dir: &'a mut DirectoryStream, loc: c_long) {
```
Susurrus: I'm thinking a newtype around `c_long` for use as the return value of `telldir` and the 2nd argument of `seekdir` would be good, since that's an invariant that should be embedded in the type system. I can't think of a good name for it, hopefully you can, but I think we should definitely do it.

src/dirent.rs
```diff
@@ -0,0 +1,124 @@
...
+pub fn opendir<P: ?Sized + NixPath>(name: &P) -> Result<DirectoryStream> {
+    let dirp = try!(name.with_nix_path(|cstr| unsafe { libc::opendir(cstr.as_ptr()) }));
+    if dirp.is_null() {
+        Err(Error::last().into())
+    } else {
+        Ok(DirectoryStream(dirp))
+    }
+}
+
+/// Like `opendir()`, but returns a directory stream for the directory
```
Susurrus: Please make the first line all one line. You can go out to 99 characters with your comments.

test/test_dirent.rs
```diff
@@ -0,0 +1,51 @@
...
+    assert_eq!(letter2, '.' as u8);
+
+    seekdir(&mut dir, pos); // same again
+
+    let letter3 = readdir(&mut dir)
+                       .unwrap()
+                       .unwrap()
+                       .name()
+                       .to_bytes()[0];
+    assert_eq!(letter3, '.' as u8);
```
Susurrus: This would be more clear if you compared `letter2` and `letter3` directly if that's what should be happening here.

test/test_dirent.rs
```diff
@@ -0,0 +1,51 @@
+use nix::dirent::{self, opendir, readdir, seekdir, telldir, DirectoryStream};
+use std::path::Path;
+use tempdir::TempDir;
+
+fn test_readdir<Open>(open_fn: Open)
+    where Open: Fn(&Path) -> DirectoryStream
+{
```
Susurrus: Can you check the full name of the directory? It's confusing that you're checking just the first character of the filename. It'd also be helpful if you created a directory in the tempdir you made so you have 3 directories, then you iterate through all of them checking their names, and then seek back to check that seeking works.

Susurrus:
Just saw you comment about the order of `.` and `..` varying depending on platform. I do know many platforms don't guarantee returning things in alphabetical order. Maybe you should iterate through all the directories collecting their names into a sorted list and then compare that?

test/test_dirent.rs
Mic92: There is no guarantee on the order of directory entries.

src/dirent.rs
Mic92: `into_raw` is there for compatibility reasons. It is not the suggested way to do conversion:
https://github.com/rust-lang/rfcs/blob/master/text/0529-conversion-traits.md

test/test_dirent.rs
Susurrus: I still think it's worth checking their names, we can treat it as a set, where the order doesn't matter but the contents should still be the same. We'd expect to encounter both "." and "..", but we aren't certain of the order, but that's fine, we can still do that check.

Susurrus:
@Mic92 I'd still like to get this merged, can you resolve the lingering issues and rebase this?

Mic92:
Next week - this week I am on holiday. Maybe CHANGELOG.md should not be modified by pull requests at all. It creates to much merge conflicts and could be easily added from the pull request description after a merge.

Mic92:
can you restart, the one failing tests, looks like a transient failure: https://travis-ci.org/nix-rust/nix/builds/270775924

asomers:
offending test restarted

Mic92:
if somebody is interested in `dirfd` and wants to make a pull request, feel free to cherry-pick from this branch: https://github.com/Mic92/nix/tree/dirfd

Susurrus:
Looks like there are legitimate test failures here. @Mic92 Are you interested in pushing this PR through to be merged? Based on your last comment, it sounds like you aren't interested in continuing to work on this PR. Is that correct?

Mic92:
Yes.

PR #561, opened 2017-03-25
reuse definitions from libc

Mic92:


src/sys/statfs.rs
```diff
@@ -87,30 +64,20 @@ pub mod vfs {
...
-    use sys::statfs::vfs;
-
-    extern {
-        pub fn statfs(path: * const c_char, buf: *mut vfs::Statfs) -> c_int;
-        pub fn fstatfs(fd: c_int, buf: *mut vfs::Statfs) -> c_int;
-    }
-}
-
-pub fn statfs<P: ?Sized + NixPath>(path: &P, stat: &mut vfs::Statfs) -> Result<()> {
+pub fn statfs<P: ?Sized + NixPath>(path: &P, stat: &mut libc::statfs) -> Result<()> {
```
Mic92: This also requires a changelog entry then.

src/sys/statfs.rs
```diff
@@ -1,45 +1,22 @@
...
-        pub f_bavail: u64,
-        pub f_files: u64,
-        pub f_ffree: u64,
-        pub f_fsid: u64,
-        pub f_namelen: NameLen,
-        pub f_frsize: FragmentSize,
-        pub f_spare: [SwordType; 5],
-    }
-
     pub const ADFS_SUPER_MAGIC : FsType =      0xadf5;
```
Mic92: These definitions are also available in libc. Should I remove them?

Susurrus: Yes, the goal is to get all FFI definitions into libc.

src/sys/statfs.rs
Susurrus: Yes, please add it to this PR.

Mic92: done

src/sys/statfs.rs
Mic92: done

Susurrus:
This looks pretty good to me. I think we need to resolve the re-exporting of constants and then I think this is good to merge.

Susurrus:
I'd like if instead of making people use `libc` types with `nix`, they'd use `nix` types directly. So I suggest we keep a `StatFs` type that is a re-export of `libc::sys::statfs` and use that in places. I don't want our users to have to `use libc` for anything and instead rely on `nix` types explicitly. Can you change this for the functions you modified?

Mic92:
There is not feedback from the maintainers of this project. So I will not put effort into this.

Susurrus:
@Mic92 Are you saying you aren't interested in getting this merged?

Mic92:
It has not highest priority for me at the moment. I could also drop the last commit and only get the others merged.

Susurrus:
That'd be great if you could rebase this and keep only the first 2 commits (be sure to add a CHANGELOG entry for the first commit).

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Marked `sys::mman::{ mmap, munmap, madvise, munlock, msync }` as unsafe.
   ([#559](https://github.com/nix-rust/nix/pull/559))
+- `nix::sys::statfs::{statfs,fstatfs}` uses statfs definition from `libc::statfs` instead of own type `nix::sys::Statfs`.
+  Also file system type constants like `nix::sys::statfs::ADFS_SUPER_MAGIC` were removed in favor of the libc aquivalent.
```
Susurrus: I'm fine ignoring these constants for now. Eventually I want to implement a wrapper type like this as I've done in #527 for `libc::termios`.

Could you fix the spelling of "equivalent" here?

src/sys/stat.rs
```diff
@@ -84,15 +74,15 @@ pub fn makedev(major: u64, minor: u64) -> dev_t {
...
+    let prev = unsafe { libc::umask(mode.bits() as mode_t) };
     Mode::from_bits(prev).expect("[BUG] umask returned invalid Mode")
 }
 
 pub fn stat<P: ?Sized + NixPath>(path: &P) -> Result<FileStat> {
     let mut dst = unsafe { mem::uninitialized() };
     let res = try!(path.with_nix_path(|cstr| {
         unsafe {
-            ffi::stat(cstr.as_ptr(), &mut dst as *mut FileStat)
+            libc::stat(cstr.as_ptr(), &mut dst as *mut FileStat)
```
Susurrus: Please either pull the changes to use `libc::stat/fstat` into a separate commit or change the name of this commit as this doesn't just modify `mknod`/`umask`.

Susurrus:
This needs that spelling fix and cleaning up that one commit, then this should be good to merge.

CHANGELOG.md
Mic92: The API advertised here was incorrect anyway. All the magic constants are linux specific. BSD and OS X uses string constants: https://www.freebsd.org/cgi/man.cgi?query=statfs&sektion=2

src/sys/stat.rs
Mic92: changed the title.

Susurrus:
Great, another little bit of cleanup. Thanks for fixing those commits too, I like a nice commit history!

bors r+

asomers:
Umlaut:1, Buildbot: 0

Buildbot 0.9.5 has a known problem when a commit's author contains non-ascii characters.  That's the cause of the Bors timeout.  I'm going to bypass Bors for this one.

Susurrus:
@asomers I'm amazed at how many bugs we've found in buildbot just using it for a brief amount of time. I would've thought most of this would have been ironed out. Anyways, is there a bug filed upstream for that so we could see a fix at somepoint?

Mic92:
And I thought jenkins is buggy.

asomers:
@Susurrus here's the upstream bug.  It's already fixed, but I haven't upgraded our install yet. https://github.com/buildbot/buildbot/issues/3078 .  I think this is the only real bug we've found so far.  Everything else has been a configuration error on my part.  It's still frustrating that it's so hard to come up with a good configuration, and there isn't a template for "Travis Replacement", but we can't blame BuildBot's code for those other problems.

Susurrus:
@asomers Yeah, maybe that's something you can contribute to their docs in your copious spare time! ;-)

But I thought there was also the bug about PRs from branches named `master`?

asomers:
@Susurrus I'm pretty sure that the bug about PRs from branches named master is also a configuration problem.  See Issue #652 .  I made a config change last night, and it looks like it's working now.

PR #562, opened 2017-03-26
add more *at functions.

Mic92:


Mic92:
That's all folks. I have all methods for my fuse.

kamalmarhubi:
@Mic92 this is heroic. I'm going to take a look over the weekend. Sorry for the delay!

kamalmarhubi:
I retried the failing test and it works now so that's a plus.

Mic92:
I would also add large file support to all methods in a second pull request (64bit fields on 32-bit systems), to align the with glibc/musl/rust.

posborne:
This is great @Mic92, just needs a rebase by the looks of it (the curse of the changelog).  The only thing I would add is that we have been trying to add docs to more of the functions we are adding.  I do not think this is a blocker for getting this included but would be a nice bonus for this change or a future PR.

E.g. Docs for fork: https://github.com/nix-rust/nix/blob/master/src/unistd.rs#L47

lucab:
I have a doubt on this (and on other fn taking `RawFd` in this crate): are all these functions just taking ownership of (a copy of) the fd or borrowing it? I *think* these all do an implicit copy+own of the RawFd (as it is just an alias to an int), and I fear this doesn't really guarantee that the fd has not been closed/dropped in the meanwhile by a `close()`.

I'm not sure if it is a valid concern or not, but shouldn't all these functions take a `&RawFd` (and perhaps `RawFd` be made non-`Copy`)?

kamalmarhubi:
@lucab this is a real concern, and I agree that we should do what you suggest. We cannot change `RawFd` as it's part of the standard library. But we should change all `RawFd` parameters to take references, or something at least that safe. I've opened #594 to track it. Thanks for raising this!

Mic92:
A reference does not buy you anything: https://play.rust-lang.org/?gist=1739781faf61fcec4afaf65bbaadd968&version=stable&backtrace=0
An not copyable type would destroy the semantics of some system calls like `renameat`, as it would no longer allow to pass the same directory twice.

Mic92:
Merge conflict resolved. Man page references added.

src/unistd.rs
```diff
@@ -356,6 +375,13 @@ pub fn getcwd() -> Result<PathBuf> {
     }
 }
 
+// According to the POSIX specification, -1 is used to indicate that
```
Susurrus: This comment doesn't make sense here. Probably want to put it back in the function.

src/sys/stat.rs
```diff
@@ -53,6 +55,7 @@ bitflags! {
     }
 }
 
+
```
Susurrus: This extra line should be removed.

src/unistd.rs
```diff
@@ -601,6 +672,69 @@ pub fn lseek64(fd: RawFd, offset: libc::off64_t, whence: Whence) -> Result<libc:
     Errno::result(res).map(|r| r as libc::off64_t)
 }
 
+/// call the link function to create a link to a file
```
Susurrus: Please capitalize the first word and add a period to the end of the first line of all these doc comments. If you run `cargo doc --no-deps --open` you can easily review their appearance to make sure things look good there.

src/unistd.rs
Mic92: I use this function in four different flavors of chown. Should I copied the comment to each of them? Why it does not make it sense here?

src/sys/stat.rs
```diff
@@ -134,3 +150,99 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+                tv_sec: 0,
+                tv_nsec: libc::UTIME_OMIT,
+            },
+            &UtimeSpec::Time(spec) => *spec.as_ref()
+        }
+    }
+
+    /// Change file timestamps with nanosecond precision
+    /// (see [utimensat(2)](http://man7.org/linux/man-pages/man2/utimensat.2.html))
+    pub fn utimensat<P: ?Sized + NixPath>(dirfd: RawFd,
```
Susurrus: `dirfd` and `pathname` actually combo together to form a ternary input type: one allowing a `RawFd`, another a relative path from the current working directory, and a third using an absolute path. I think we should use a ternary enum type here instead of the `dirfd` and `pathname` inputs to make this function call safer.

src/sys/stat.rs
```diff
@@ -134,3 +150,99 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+#[cfg(target_os = "linux")]
+mod linux {
+    use {Errno, Result, NixPath};
+    use std::os::unix::io::RawFd;
+    use libc;
+    use fcntl::AtFlags;
+    use sys::time::TimeSpec;
+
+    pub enum UtimeSpec {
+        Now,
```
Susurrus: This enum is public and as such needs docs.

src/unistd.rs
Susurrus: Because you're referring to `uid_t` and `gid_t` which aren't used in this function. The description to this should be more broad (not referring to those specific use cases) and then you can put comments stating why you need to use this function at the call site that refers to specific variables.

Susurrus:
There a common theme through a lot of these functions of having a `dirfd` and a `path` that is a tristate:
  * The `dirfd` is ignored if `path` is absolute
  * If `dirfd` is `AT_FDCWD` and `path` is relative it's resolved relative to the CWD
  * If `path` is relative and `dirfd` **isn't** `AT_FDCWD` then the path is resolved relative to that path.

I'm thinking an enum encompassing these states would be better instead of having two separate variables. I think an ideal solution would be to have `NixPath` have an `Absolute` and `Relative` mode or something to that effect, but for this PR I think just having an enum with different types is good enough.

test/test_stat.rs
```diff
@@ -1,12 +1,12 @@
 use std::fs::File;
-use std::os::unix::fs::symlink;
+use std::os::unix::fs::{symlink, PermissionsExt};
 use std::os::unix::prelude::AsRawFd;
+use tempfile::NamedTempFile;
```
Susurrus: Please put non-std/core imports down in their own group.

src/sys/stat.rs
```diff
@@ -179,3 +182,70 @@ pub fn fchmodat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, mode: Mode, fla
...
+    use fcntl::AtFlags;
+    use sys::time::TimeSpec;
+
+    /// A file timestamp.
+    pub enum UtimeSpec {
+        /// File timestamp is set to the current time.
+        Now,
+        /// The corresponding file timestamp is left unchanged.
+        Omit,
+        Time(TimeSpec)
```
Susurrus: This one needs a comment too.

src/sys/stat.rs
```diff
@@ -179,3 +182,70 @@ pub fn fchmodat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, mode: Mode, fla
...
+    }
+
+    fn to_timespec(time: &UtimeSpec) -> libc::timespec {
```
Susurrus: Can you add a doccomment for this function? Nix devs should know if they can or should use it elsewhere in this function.

CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#611](https://github.com/nix-rust/nix/pull/609)
 - Added `AioCb::from_boxed_slice`
   ([#582](https://github.com/nix-rust/nix/pull/582)
-- Added `nix::unistd::{openat, fstatat, readlink, readlinkat, rename, renameat, mknodat}`
-  ([#551](https://github.com/nix-rust/nix/pull/551))
+- Added `nix::unistd::{openat, fstatat, readlink, readlinkat, rename, renameat, mknodat, unlinkat}`
+  ([#497](https://github.com/nix-rust/nix/pull/551))
```
Susurrus: I don't think the PR numbering change he was intentional.

Mic92:
@Susurrus do you want to implement that enum for these functions send me a pull request? I don't see how Rust's typesystem is expressive enough to ensure a path is always relative/absolute without runtime checks.

Susurrus:
My suggestion is to have a:
```Rust
pub enum AtPath {
  AbsPath(NixPath),
  RelativeTo(RawFd, NixPath),
  RelativeCwd(NixPath),
}
```

Now we could add runtime checks through having some conversion path from `NixPath`s to this enum and I think that's the most ergonomic, which would be ideal, and if you'd like to implement something like that we can spec it out here. But at least with this enum it's more clear to the user what combinations of inputs are valid.

Susurrus:
I wanted to point out that I also preferred the arguments @bugaevc who was working on `access`/`faccessat` in #605. There they came up with a dual-state enum for use with those functions that looks like:
```Rust
pub enum AccessFlags {
  FileExists,
  FileExistsAndPermissions(PermFlags),
}
```

where `PermFlags` are the `R_OK`, etc. constants. I'd prefer having that than the regular libc-like constants.

Susurrus:
@Mic92 I wanted to circle back to this and see if we could get this merged? I'm working on some stuff in fcntl but it should get merged within a day or so. Then I'd like to get this PR merged if possible. Still willing to push this through?

Mic92:
I am back from holiday on Thursday.

lucab:
What happened to this?

Mic92:
I made multiple pull requests out of it.

PR #563, opened 2017-03-26
Use specific version of libc

Susurrus:
It should be policy for additional functionality to wait for a `libc` release anyways before `nix` can release, so these should just bump the `libc` dependency version anyways.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ preadv_pwritev = []
 signalfd = []
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc" }
+libc = "0.2.21"
 bitflags = "0.7"
 cfg-if = "0.1.0"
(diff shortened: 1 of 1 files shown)


Susurrus:
Probably #565 should be merged instead of this, though a policy should be set on whether we want to depend only on released versions for dependencies or using git directly is fine. @posborne @fiveop Any comments on this or should I bring it up on `nix-rust/rfcs`?

kamalmarhubi:
Right now we set a fixed version at release time, and otherwise leave it tracking the repository. This allows someone to add a feature to nix and libc without requiring a release of libc until we want to release nix. It's weird, but I think it's a nice way not to force lots of libc releases to make progress on nix.

kamalmarhubi:
I merged #565 so I'm going to close this. But please reopen and/or start a new discussion if you want to chat more about it!

Issue #305, opened 2016-03-10
Rededesign iovec handling


kamalmarhubi:
Since closing #263 on using `&[u8]` and `&mut [u8]` directly in `readv(2)` et al, I've been thinking of alternatives.

My current thinking:
- switch from a type parameter `IoVec<T>` to a lifetime parameter `IoVec<'a>`, effectively restricting the type to `&'a [u8]`
- add an `IoVecMut<'a>` for the writeable case, which implements `Deref<Target=IoVec<'a>>`
- implement as many conversion traits as we can

I've got a partial implementation that I started on after #304. I think some tidying up in the `socket` modules might need to happen first though.

PR #568, opened 2017-04-03
Add process_vm_readv and process_vm_writev

geofft:


geofft:
Opened too soon - I guess I should add both a platform conditional and some tests. Let me do that.

geofft:
One design-ish question: these syscalls take a `flags` argument, but currently no flags are defined (see [the manpage](https://github.com/nix-rust/nix/pull/568)). I chose to just not expose that argument in the wrapper, but should I create an empty bitflags for API future-compatibility? The system call has been around since Linux 3.2 (January 2012) and I don't think anyone has proposed adding flags since then.

kamalmarhubi:
For some reason I thought I did this back when https://github.com/rust-lang/libc/pull/283 was filed for https://github.com/jvns/ruby-stacktrace.

Anyway, thanks for doing this!

Re the `flags` argument. I could see this going either way: we can have a breaking change in the future should that change. Alternatively, we can have the empty flags arg.

I prefer the empty flags arg since the ergonomics aren't *that* bad, and these aren't super frequently used functions. The alternative may at some point down the line force a major version bump on all users just to correctly handle these infrequently used functions.

geofft:
Good point about a major-version bump for an obscure function. I've added a flags type. I wasn't really sure what to name it since no constants exist yet, but my current best guess is `CMA_*`, for "Cross-Memory Attach", the [name of the feature](https://lwn.net/Articles/451188/) these two system calls implement. But it could be `PV_` or `PVRW_` or `PRWF_` (after preadv2 and pwritev2's `RWF_`) or something. I guess the cost of getting it wrong is pretty low: if a flag ever shows up, nix's type would be a bit unusually named, but that seems fine.

Adding an empty `bitflags!` is a bit ugly; the macro uses `+` repetition instead of `*`, so I added a variant that is always compiled out, which satisfies the macro. I wonder if the upstream crate would be interested in a fix to that.

kamalmarhubi:
@alexcrichton thoughts on this for bitflags:
> Adding an empty `bitflags!` is a bit ugly; the macro uses `+` repetition instead of `*`, so I added a variant that is always compiled out, which satisfies the macro. I wonder if the upstream crate would be interested in a fix to that.

kamalmarhubi:
@geofft re the name of the flag type, I think we can go with more or less anything since we can add a type alias if / when something actually exists. What do you think?

alexcrichton:
Oh yes we'd certainly be interested in a fix!

geofft:
@kamalmarhubi Oh good call, that sounds forwards-compatible.

nix is on bitflags 0.7.0, and upstream is on 0.8.0 which has a few changes. Can we merge this as is and I'll send in a patch to bitflags + a cleanup for nix at some point in the future?

kamalmarhubi:
@geofft sorry for the delay. I'd absolutely be ok to merge this and fix up the bitflag later. Nice trick with `#[cfg(any())]` too!

kamalmarhubi:
(This needs a rebase though!)

geofft:
Tests are failing on ARM only with "qemu: Unsupported syscall: 376" (or 270 on aarch64). These are the right syscall numbers. Is there a way to mark this as a Travis-specific false positive? I would expect the code to work on actual ARM (although I suppose I should dig out a device).

If we can't ignore this from Travis easily, I'm happy to `cfg` it away until Travis gets better.

Susurrus:
We do need a system to handle when Qemu on Travis isn't able to test features. We need a way to disable them only when testing on CI (I think ideally we could detect the CI environment variable to disable it). I'd also like to have any such issues filed upstream so that we can have a comment by ever disabled test with a URL for the tracking issue so we will be able to easily turn the test on once the feature is added.

kamalmarhubi:
If #588 makes sense to at least a couple of maintainers, it'll be a way to ignore these tests in Travis. (I'm assuming they pass locally on a reals system.)

kamalmarhubi:
@geofft can you try rebasing and using the attribute mentioned in #588?

Susurrus:
@kamalmarhubi What do you think about whenever we remove a test from running on Travis that we have a link to an upstream issue if one exists?

geofft:
OK, I tried both `#[cfg_attr(and(travis, target_arch...))]` and `#[cfg(not(and(travis, target)))]` and it's still running the tests (and I've definitely rebased on top of the commit that sets $RUSTFLAGS)... what am I doing wrong?

Susurrus:
@geofft I'm working on this in #590 and found the same issue. If you want to follow things, we'll work on it there.

Susurrus:
@geofft We're actively working on getting Travis fully operational for all targets in #590, so why don't you ignore that part of this and we'll wait until that lands. Should be fairly soon.

geofft:
Yeah, I saw the discussion on #590 - I'm trying a few things to see if they work while we wait for cross to get upgraded, and I'll comment there if they do.

Susurrus:
@geofft #590 is now merged. Please update this PR if you'd like to get it merged.

Susurrus:
@geofft Are you still interested in merging this PR?

geofft:
@Susurrus - rebased, sorry! The libc crate doesn't define these functions on Android, so I've left them as normal-Linux-only for now,. (I'm guessing this is because Bionic only gained wrappers for these syscalls in Android 6.0 / API level 23, which is way newer than rustc's target of Android 2.3 / API level 9.)

I did a `#[cfg_attr(not(any(target_arch = "x86", target_arch = "x86_64")), ignore)]` on the test since it fails for all ARM and MIPS and PowerPC variants, let me know if that's not right.

src/sys/uio.rs
```diff
@@ -56,6 +59,38 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
     Errno::result(res).map(|r| r as usize)
 }
 
+// process_vm_{read,write}v currently doesn't use flags, but we define this
```
Susurrus: Flags should always be zero in my current version of Linux. I don't think forward compatibility here is a concern because as soon as they're flags we have issues with backwards compatibility on the systems themselves. Let's just leave all this extra boilerplate off for now for the sake of a cleaner API. By the time new flags are added it could be years.

src/sys/uio.rs
```diff
@@ -56,6 +59,38 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
...
+    pub struct CmaFlags: libc::c_ulong {
+        // The bitflags! macro requires at least one variant. any() evaluates
+        // to false, so this symbol never actually exists.
+        #[cfg(any())]
+        const CMA_DUMMY = 0;
+    }
+);
+
+#[cfg(any(target_os = "linux"))]
+pub fn process_vm_writev(pid: Pid, local_iov: &[IoVec<&[u8]>], remote_iov: &mut [IoVec<&mut [u8]>], flags: CmaFlags) -> Result<usize> {
```
Susurrus: These needs docs. See some of the recently closed PRs for an idea of what we're looking for. I think these are Linux-specific functions, so you should link to a manpage on man7.org in the doc comment. Please run `cargo doc --no-deps --open` and read through the generated docs to make sure they look nice as part of your testing.

src/sys/uio.rs
```diff
@@ -56,6 +59,38 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
...
+    let res = unsafe {
+        libc::process_vm_writev(pid.into(), local_iov.as_ptr() as *const libc::iovec, local_iov.len() as libc::c_ulong,
+                                            remote_iov.as_ptr() as *const libc::iovec, remote_iov.len() as libc::c_ulong, flags.bits())
+    };
+
+    Errno::result(res).map(|r| r as usize)
+}
+
+#[cfg(any(target_os = "linux"))]
+pub fn process_vm_readv(pid: Pid, local_iov: &mut [IoVec<&mut [u8]>], remote_iov: &[IoVec<&[u8]>], flags: CmaFlags) -> Result<usize> {
```
Susurrus: Also needs docs.

src/sys/uio.rs
```diff
@@ -56,6 +59,38 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
...
+pub fn process_vm_writev(pid: Pid, local_iov: &[IoVec<&[u8]>], remote_iov: &mut [IoVec<&mut [u8]>], flags: CmaFlags) -> Result<usize> {
+    let res = unsafe {
+        libc::process_vm_writev(pid.into(), local_iov.as_ptr() as *const libc::iovec, local_iov.len() as libc::c_ulong,
+                                            remote_iov.as_ptr() as *const libc::iovec, remote_iov.len() as libc::c_ulong, flags.bits())
+    };
+
+    Errno::result(res).map(|r| r as usize)
+}
+
+#[cfg(any(target_os = "linux"))]
```
Susurrus: Remove the `any(`

src/sys/uio.rs
```diff
@@ -56,6 +59,38 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
...
+bitflags!(
+    pub struct CmaFlags: libc::c_ulong {
+        // The bitflags! macro requires at least one variant. any() evaluates
+        // to false, so this symbol never actually exists.
+        #[cfg(any())]
+        const CMA_DUMMY = 0;
+    }
+);
+
+#[cfg(any(target_os = "linux"))]
```
Susurrus: Remove the `any(`

src/sys/uio.rs
```diff
@@ -56,6 +59,38 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
...
+    };
+
+    Errno::result(res).map(|r| r as usize)
+}
+
+#[cfg(any(target_os = "linux"))]
+pub fn process_vm_readv(pid: Pid, local_iov: &mut [IoVec<&mut [u8]>], remote_iov: &[IoVec<&[u8]>], flags: CmaFlags) -> Result<usize> {
+    let res = unsafe {
+        libc::process_vm_readv(pid.into(), local_iov.as_ptr() as *const libc::iovec, local_iov.len() as libc::c_ulong,
+                                           remote_iov.as_ptr() as *const libc::iovec, remote_iov.len() as libc::c_ulong, flags.bits())
```
Susurrus: This indentation is wrong, it should be aligned to `pid`.

test/sys/test_uio.rs
```diff
@@ -190,3 +190,51 @@ fn test_preadv() {
     let all = buffers.concat();
     assert_eq!(all, expected);
 }
+
+#[test]
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Remove the `android` here.

test/sys/test_uio.rs
```diff
@@ -190,3 +190,51 @@ fn test_preadv() {
...
+// FIXME: qemu-user doesn't implement process_vm_readv/writev on most arches
+#[cfg_attr(not(any(target_arch = "x86", target_arch = "x86_64")), ignore)]
+fn test_process_vm_readv() {
+    use nix::unistd::ForkResult::*;
+    use nix::sys::signal::*;
+    use nix::sys::wait::*;
+    use std::{str, slice};
+
+    let (r, w) = pipe().unwrap();
+    match fork() {
```
Susurrus: Forking is bad within the Rust test harness as it can cause deadlocks. Any other way you can implement this test? Also some comments in here explaining things for those unfamiliar with the code would be nice.

Susurrus:
@asomers Can you chime in here about the use of `fork` as well?

src/sys/uio.rs
geofft: See the earlier discussion on the PR - I'm pretty indifferent here, @kamalmarhubi expressed a preference to just add the flags argument now. I agree that probably there will never be flags.

I guess the other option is that we just break names from libc slightly (which we do in other places) and add a four-parameter `process_vm_readv_flags(pid, local_iov, remote_iov, flags)` if we ever need to, which avoids the API break.

src/sys/uio.rs
geofft: Thanks, I'll add a linebreak before local_iov. (Should I be using rustfmt or something?)

test/sys/test_uio.rs
geofft: The child calls `pause()`, which means it never continues execution past that point, so it doesn't unwind anything and shouldn't cause any deadlocks in execution. If you're having trouble with deadlocks during child _creation_ then yeah that's a problem but also that seems surprising... I guess another thread might be holding the malloc lock or something? Blah.

I can probably use `process_vm_readv(getpid()...)` with a static string.

src/sys/uio.rs
Susurrus: Let's remove flags. We can assess a breaking change or adding `*_flags` when/if this moment comes to pass in the future.

test/sys/test_uio.rs
Susurrus: I don't fully understand the problem, @asomers is our resident expert, so I'll let him chime in.

asomers: All the fork-related deadlocks that I've seen so far happen as the child returns from the test function.  So if the child truly never returns, then you should be ok.  You'll still have to grab the FORK_MTX however, or you could cause the `wait` tests to fail.

geofft:
Ugh, I had a moment of worry that using `IoVec` to refer to a foreign process's memory is UB (since IoVec wraps a `&[u8]` or `&mut [u8]`), and @ubsan is telling me on #rust that it probably is UB to construct a pointer to an address that isn't valid in the current process. Let me redo this to take a `&[(usize, usize)]` or something, one sec.

From `git grep iovec` in a man-pages checkout, I don't think anything other than these two functions uses `struct iovec` to refer to foreign memory, so I'm inclined not to make a type for it, but happy to do so if it seems useful.

test/sys/test_uio.rs
```diff
@@ -190,3 +190,51 @@ fn test_preadv() {
...
+
+            kill(child, SIGTERM).unwrap();
+            waitpid(child, None).unwrap();
+
+            assert_eq!(Ok(4), ret);
+            assert_eq!(&buf, b"test");
+        },
+        Ok(Child) => {
+            let _ = close(r);
+            let s = String::from("test");
```
asomers: Technically, you shouldn't use `String::from` in the child, because that requires allocating memory, and memory allocation isn't async signal safe.  The `format!` macro probably does allocation too.  See http://man7.org/linux/man-pages/man7/signal-safety.7.html .

Susurrus:
@kamalmarhubi This also touches on #305, which you filed. Could you comment here on the proposal here?

Susurrus:
@geofft Just wanted to ping you on this and see if we could get this wrapped up soon.

geofft:
@Susurrus Woops, I totally forgot that this hadn't landed (also I started a new job in August). What needs to be done here - just getting rid of memory allocation in the test's child process? I can update the PR to fix that, hopefully tonight.

Susurrus:
@geofft I believe all my comments have been addressed, it's just @asomers who has some comments left and I think it's just that allocation one.

And congrats on the new job, I did the same thing in August, which is why I'm just now coming back to clean things up a bit!

test/sys/test_uio.rs
geofft: @asomers Let me know if this looks better. (I'm not actually sure if it's fully safe, in the sense that I'm not sure there's a guarantee that the writes to memory actually happen, but at least it's not async-signal-unsafe.)

asomers:
LGTM

bors r+

Issue #567, opened 2017-04-02
Nix 0.8.0 fails to build on FreeBSD 10.3


alexreg:
Here's the error log.

```
   Compiling nix v0.8.0
error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:173:21
    |
173 |                     B0 => BaudRate::B0,
    |                     ^^ expected u64, found u32
    |
    = note: expected type `u64`
               found type `u32`

error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:174:21
    |
174 |                     B50 => BaudRate::B50,
    |                     ^^^ expected u64, found u32
    |
    = note: expected type `u64`
               found type `u32`

error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:175:21
    |
175 |                     B75 => BaudRate::B75,
    |                     ^^^ expected u64, found u32
    |
    = note: expected type `u64`
               found type `u32`

error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:176:21
    |
176 |                     B110 => BaudRate::B110,
    |                     ^^^^ expected u64, found u32
    |
    = note: expected type `u64`
               found type `u32`

error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:177:21
    |
177 |                     B134 => BaudRate::B134,
    |                     ^^^^ expected u64, found u32
    |
    = note: expected type `u64`
               found type `u32`

error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:178:21
    |
178 |                     B150 => BaudRate::B150,
    |                     ^^^^ expected u64, found u32
    |
    = note: expected type `u64`
               found type `u32`

error[E0308]: mismatched types
   --> /home/alexuk/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.0/src/sys/termios.rs:179:21
    |
[...]

PR #576, opened 2017-04-14
Add AioCb::from_boxed_slice

### LINKED ISSUE #575: AioCb should work with refcounted buffers
The `AioCb` structure contains a reference (actually just a PhantomData) to a slice.  This makes it impossible to use in situations where the borrow checker can't prove that the slice will outlive the `AioCb`, such as with Tokio.  There should be a way to create `AioCb`s that use reference counted buffers instead.

asomers:
The existing AioCb constructors work for simple programs where
everything is stored on the stack.  But in more complicated programs the
borrow checker can't prove that a buffer will outlive the AioCb that
references it.  Fix this problem by introducting
AioCb::from_boxed_slice, which takes a reference-counted buffer.

Fixes #575


Susurrus:
Can you rebase onto master and update this PR? The `new-ci/master` branch is only being used for developing the new CI framework.

asomers:
Nope.  Can't rebase to master because of issue #567 .  But I can wait until new-ci/master is merged.

Susurrus:
Indeed, hoping to merge it today.

asomers:
Replaced by the rebased PR #582

PR #579, opened 2017-04-15
tests should protect concurrent access to signal handlers

### LINKED ISSUE #578: tests should protect concurrent access to signal handlers
tests/sys/test_aio.rs has two separate tests that install handlers for SIGUSR2.  This was probably the root cause of #529 .  It should've been fixed by 0e3ef36, but we still sometimes see related failures.  Instead of globally restricting the thread count to 1, those tests should use a mutex to protect access to signal handlers.  So far, there are no tests outside of that file that also use SIGUSR2.

asomers:
Adds a mutex to protect access to SIGUSR2 signal handlers by the AIO
tests.

Fixes #578

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ cfg-if = "0.1.0"
 void = "1.0.2"
 
 [dev-dependencies]
+lazy_static = "0.2"
 rand = "0.3.8"
 tempdir = "0.3"
 tempfile = "2"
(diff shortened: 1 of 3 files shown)


Susurrus:
@homu retry

Susurrus:
@asomers I switched the PR branch to `new-ci/master`, as the MIPS tests don't work on `master` right now. Can you re-push your commits to re-run Travis?

asomers:
@Susurrus  why don't we just wait until after you've merged new-ci/master?  Then I'll rebase.  I don't want to hold up that work.

Susurrus:
@asomers Sure.

Susurrus:
@asomers Can you rebase this (#536 just landed)?

asomers:
Well, it didn't fix the mips errors, but I still think that this is the right thing to do.  Does anybody object to adding the dependency on lazy_static ?

posborne:
> Does anybody object to adding the dependency on lazy_static?

Nope, that should be fine.  Doesn't really drag in anything too much extra.

PR #582, opened 2017-04-15
Add AioCb::from_boxed_slice

### LINKED ISSUE #575: AioCb should work with refcounted buffers

asomers:
The existing AioCb constructors work for simple programs where
everything is stored on the stack.  But in more complicated programs the
borrow checker can't prove that a buffer will outlive the AioCb that
references it.  Fix this problem by introducting
AioCb::from_boxed_slice, which takes a reference-counted buffer.

Fixes #575

### DIFF
--- a/src/sys/aio.rs
+++ b/src/sys/aio.rs
@@ -8,6 +8,7 @@ use std::io::Write;
 use std::io::stderr;
 use std::marker::PhantomData;
 use std::mem;
+use std::rc::Rc;
 use std::ptr::{null, null_mut};
 use sys::signal::*;
 use sys::time::TimeSpec;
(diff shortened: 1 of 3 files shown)


src/sys/aio.rs
```diff
@@ -61,6 +62,17 @@ pub enum AioCancelStat {
...
 
+/// Private type used by nix to keep buffers from Drop'ing while the kernel has
+/// a pointer to them.
+#[derive(Clone, Debug)]
+enum Keeper<'a> {
+        none,
+        /// Keeps a reference to a Boxed slice
+        boxed(Rc<Box<[u8]>>),
+        /// Keeps a reference to a slice
+        phantom(PhantomData<&'a mut [u8]>)
```
posborne: nit: Indentation on this block seems off

posborne:
One super minor nit, but otherwise the change looks good to me.

asomers:
good catch.  I'll fix it and merge.

asomers:
bors +r

asomers:
thanks @posborne .  I obviously need to work on my bors-foo.

PR #588, opened 2017-04-17
ci: Allow skipping tests when running under CI

kamalmarhubi:
This adds `--cfg=travis` to builds under CI, allowing tests to be
conditionally ignored by marking them with

    #[cfg_attr(travis, ignore)]

refs #568

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,13 @@ locally.  More information is available in the [CI Readme][ci-readme].
 [travis-ci]: https://travis-ci.org/nix-rust/nix
 [ci-readme]: ci/README.md
 
+### Disabling a test in the CI environment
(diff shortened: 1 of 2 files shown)


kamalmarhubi:
r? @posborne 

Does this seem reasonable to you?

kamalmarhubi:
In particular, setting `RUSTFLAGS` passes this down to all dependencies. Our transitive test dependencies right now:
- bitflags
- cfg-if
- gcc
- kernel32-sys
- lazy_static
- libc
- rand
- rustc_version
- semver
- tempdir
- tempfile
- void
- winapi
- winapi-build

kamalmarhubi:
A lesser ARM device would probably be all right for our use cases. :-)

bors r=asomers

PR #589, opened 2017-04-17
Raise the bors timeout to four hours

notriddle:
The timeout is, by default, one hour. `nix` builds take about that long,
sometimes longer, sometimes shorter, making builds flaky.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,1 +1,15 @@
+# Gate on Travis CI
 status = ["continuous-integration/travis-ci/push"]
+
+# Set bors's timeout to 4 hours
+#
+# bors's timeout should always be twice a long as the test suite takes.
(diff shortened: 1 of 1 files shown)


Susurrus:
Thanks, @notriddle!

When changing arbitrary values to other arbitrary values I'm a big fan of excessive documentation. And since I don't even know the reasoning for why 4 hours was chosen, can you elaborate both in the `bors.toml` and also in the commit message and to why it was chosen? Whatever heuristic is used will be useful to know so that we know how to adjust it if we need to in the future as well.

notriddle:
I added a comment about the matter to bors.toml

Susurrus:
Oh yeah, love it! Thanks, @notriddle!

bors r+

PR #590, opened 2017-04-17
Skip failing tests for Linux/MIPS/PowerPC

Susurrus:
These should be issues with QEMU itself and not problems on the actual hardware, but I want to get some upstream tickets filed before merging this. I'm also not sure if this syntax is correct on the `#cfg`, so this is testing that as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,8 +80,6 @@ matrix:
     # FIXME: targets that should pass but are currently failing
     - env: TARGET=i686-unknown-linux-musl
     - env: TARGET=x86_64-unknown-linux-musl
-    - env: TARGET=mips-unknown-linux-gnu
-    - env: TARGET=mipsel-unknown-linux-gnu
 
 install:
   - sh ci/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,10 @@ language: rust
 services: docker
 sudo: required
 
-# This is the Rust channel that build jobs will use by default but can be
-# overridden on a case by case basis down below
-rust: 1.13.0
-
 env:
   global:
     - CRATE_NAME=nix
     - CARGO_TEST_THREADS=1
-
-    # default job
-    - TARGET=x86_64-unknown-linux-gnu
 
 matrix:
   # These are all the build jobs. Adjust as necessary. Comment out what you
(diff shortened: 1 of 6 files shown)


asomers:
If CI is skipping some tests on these architectures, I think that makes them Tier 2, not Tier 1.

Susurrus:
@asomers The README has details on what the tiers mean. For Tier 1 it means the platforms are built and tested in CI and failures should never happen. Tier 2 means there is no automated testing, but the code should at least compile. For Tier 3 it means that testing is done, but we don't guarantee that they pass (admittedly Tier 2 and 3 should be swapped).

Susurrus:
Ideally we don't have any platforms under Tier 3 and hopefully I can remove that distinction soon.

asomers:
@Susurrus for the three platforms in question, you're proposing that only some of the tests will be run in CI.  If they tests aren't all run, then I don't think we can call it Tier 1.  Tier 1.1 perhaps, if we want to use floats, but not Tier 1.

And currently the Tier 3 definition says that we don't even guarantee that the build will pass.  Until we change that guarantee, we shouldn't swap Tier 2 and Tier 3.

Susurrus:
It won't always be possible to run all tests in CI due to various issues with Travis or Qemu or other possible issues. I don't think a distinction between "all" and "almost all" really makes much difference. The point of stating tiers is so that users (and I guess developers as well) have expectations of the state of the code and its rough quality. Platforms in Tier 1 are well supported, Tier 2 less so, and Tier 3 the least. Platforms not supported are at your own risk so to speak.

And you bring up a good point about Tier 3, we should separate tests and builds because Tier 3 should be guaranteed to build but don't necessary have their tests run during CI.

Susurrus:
@kamalmarhubi It doesn't look like my `#[cfg_attr]` statements are working correctly, do you have any suggestions for this?

kamalmarhubi:
weird, I tried it out locally.

Susurrus:
@kamalmarhubi Yeah, it worked correctly when I tested it locally using `x86_64` instead of `powerpc` for the `target_arch`. Do you know of any limitations of `cfg_attr` or the compiler exposing `target_arch` on `powerpc` architectures?

Susurrus:
I think the problem is that cross is testing things and since it's in qemu it might not be receiving the `travis` cfg flag we set earlier.

kamalmarhubi:
Oh good point.

kamalmarhubi:
This is relevant: https://github.com/japaric/cross/pull/77

Susurrus:
@kamalmarhubi Good find. That PR looks pretty close to finished so hopefully we can push it through to the next `cross` release. But for now I'm testing my own `cross` repo to see if it does indeed fix the issue.

kamalmarhubi:
Sent https://github.com/japaric/cross/pull/91 so hopefully we'll get this functionality sooner than later.

Susurrus:
We should actually also try to upstream our `RUSTFLAGS` modification to trust/cross as that'll make upgrading ours easier.

Susurrus:
Looks like cross is a little hard to test here because of it's need for a release .zip uploaded to GitHub, which is why the most recent commit broke everything. It'll be easiest to block on this until japaric/cross#91 gets released.

Susurrus:
@asomers @kamalmarhubi Could you guys look over this? Seems good to me and I'd like to merge.

.travis.yml
```diff
@@ -6,52 +6,57 @@ language: rust
 services: docker
 sudo: required
 
-# This is the Rust channel that build jobs will use by default but can be
-# overridden on a case by case basis down below
-rust: 1.13.0
```
asomers: Is there a reason not to use a default rust value?  It makes the matrix section more duplicative.

test/sys/test_aio.rs
```diff
@@ -26,6 +26,7 @@ fn poll_aio(mut aiocb: &mut AioCb) -> Result<()> {
 // bindings.  So it's sufficient to check that AioCb.cancel returned any
 // AioCancelStat value.
 #[test]
+#[cfg_attr(all(target_env = "musl", target_arch = "x86_64"), ignore)]
```
asomers: So all of these tests fail in x86_64 musl but pass in i686 musl?  Weird.  Does cross use qemu to run i686 tests?  I would think that it wouldn't need to.

.travis.yml
Susurrus: If you use the default, you end up with an "extra" test run. You can see how there's two `x86-64-unknown-linux-gnu` runs that execute right now. This fixes that problem. Looks like I didn't specify that in the commit message, which I can do.

test/sys/test_aio.rs
Susurrus: You bring up a good point. i don't think `cross` runs things in QEMU based on [this line](https://github.com/japaric/cross/blob/master/src/main.rs#L227).

Susurrus: But I've filed an issue (japaric/cross#107) in order for this to be made more clear.

Susurrus:
This issue is due to [a bug in `cross`](https://github.com/japaric/cross/issues/108)

Susurrus:
cross is fixed.

bors r+ asomers

PR #592, opened 2017-04-19
remove test warnings

king6cong:


Susurrus:
These changes all look good to me, but I don't get warnings on any of these either locally or see any in Travis CI. What warnings are you seeing and with what toolchain?

king6cong:
yes, `cargo test` or `cross test` will not show the warnings
this will:
`cargo test -- --nocapture`

Susurrus:
Interesting, I would have thought compilation warnings would have been exposed by default running `cargo test`.

Two things:
  1. Can you also fix the extraneous debug output on line 37 of `test/test_fcntl.rs` as part of this.
  2. I think we should change from running `cargo test` to running `cargo test -- --nocapture` to try to prevent this recurring in the future. It will still return a 0 status code even with warnings. I really think this is something that should be changed in Rust proper. @king6cong would you have any interest in filing a ticket in `rust-lang/cargo` about this?

kamalmarhubi:
Oh that's interesting. Thanks for this!

Susurrus:
@kamalmarhubi Any opinion on changing our `cargo test` runs to doing `cargo test -- --nocapture` assuming `cross` will allow it?

src/sys/ioctl/mod.rs
```diff
@@ -35,7 +35,7 @@
 //! from [rust-spidev](https://github.com/posborne/rust-spidev).
 //!
 //! ```
-//! #[macro_use] extern crate nix;
```
Susurrus: You should be actually able to completely remove this line. This is automatically inserted by the doctest script.

Susurrus:
`cargo test -- --nocapture` reveals a bunch of extraneous output however, so I suggest we don't change that.

@king6cong Can you address my comment? I think after that we're good to go.

king6cong:
@Susurrus updated 😃

PR #600, opened 2017-05-02
Change sched_setaffinity's PID argument to pid_t

tokenrove:
I might be missing something as to why this argument was made `isize`, but there's nothing obvious from the commit history, and other calls that work with thread IDs in this library return `pid_t`, so hopefully this is a nit and not noise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#542](https://github.com/nix-rust/nix/pull/542))
 - Changed type signature of `sys::select::FdSet::contains` to make `self`
   immutable ([#564](https://github.com/nix-rust/nix/pull/564))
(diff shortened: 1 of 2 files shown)


Susurrus:
I'd like to get the FreeBSD build passing, which should just require you re-pushing this branch. Otherwise I think it's RTM.

tokenrove:
Is that better?

asomers:
Hm, bors isn't picking up the PR.  Does the magic comment need to be double spaced?  Let's try.

bors r+

notriddle:
Yeah, we changed it recently so that the comment needs to be on its own. The idea if to make sure that people don't trigger builds because they happen to mention the possibility of saying "bors r+".

Susurrus:
Thanks @tokenrove for your PR!

tokenrove:
Thanks!  I have a backlog of (only slightly less trivial) contributions to submit.  This is a good reminder to do so.

Susurrus:
Excellent, look forward to reviewing them!

PR #606, opened 2017-05-22
Added sys::user.

xd009642:
This PR adds the various register structures for the x86 and x64 architectures. the Regs struct in particular is what accessed and modified by PTRACE_GETREGS and PTRACE_SETREGS so is of particular use to people using ptrace with rust-nix.

Following review comments I'll fix them and fill in the PR number in the changelog if the changes are accepted

### DIFF
--- a/src/sys/user.rs
+++ b/src/sys/user.rs
@@ -3,7 +3,7 @@
 #[cfg(any(target_arch = "x86_64"))]
 use libc::{c_ushort, c_uint, c_char, c_int, c_longlong, c_ulonglong};
 #[cfg(any(target_arch = "x86"))]
-use libc::{c_ushort, c_uint, c_char, c_int, c_long, c_ulong};
+use libc::{c_ushort, c_char, c_int, c_long, c_ulong};
 
 use libc::size_t;
 
--- a/src/sys/user.rs
+++ b/src/sys/user.rs
@@ -1,0 +1,154 @@
+
+
+#[cfg(any(target_arch = "x86_64"))]
+use libc::{c_ushort, c_uint, c_char, c_int, c_longlong, c_ulonglong};
+#[cfg(any(target_arch = "x86"))]
+use libc::{c_ushort, c_uint, c_char, c_int, c_long, c_ulong};
+
+
+#[repr(C)]
+#[cfg(any(target_arch = "x86_64"))]
+pub struct FpRegs {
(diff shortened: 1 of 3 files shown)


src/sys/user.rs
```diff
@@ -0,0 +1,177 @@
...
+const EFLAG_RESUME: Eflag  =  0x1 << 16;
+const EFLAG_VIRTUAL_80086_MODE: Eflag  =  0x1 << 17;
+const EFLAG_ALIGNMENT_CHECK: Eflag  =  0x1 << 18;
+const EFLAG_VIRTUAL_INTERRUPT: Eflag  =  0x1 << 19;
+const EFLAG_VIRTUAL_INTERRUPT_PENDING: Eflag  =  0x1 << 20;
+const EFLAG_CPUID: Eflag  =  0x1 << 21;
+
+
+#[repr(C)]
+#[cfg(any(target_arch = "x86_64"))]
```
Susurrus: This should just be `#[cfg(target_arch = "x86_64")]`. You should make the same change for the rest of these in this file.

Susurrus:
Thanks for your PR! I don't think I quite understand what this work is for. There's no functions here that produce or consume any of these values. Additionally, since all of the things you declared are safe, shouldn't they all be upstreaming into `rust-lang/libc` instead of here? `nix` serves as a thin safety layer on top of libc, but the constants and types used by it are originally defined upstream. I think you should instead file an equivalent PR to these against `rust-lang/libc` and this can be closed until safe wrappers are needed around some unsafe functions.

xd009642:
That makes sense. I'll fork and submit a PR for libc.

Susurrus:
Cool, thanks @xd009642. As for the Eflag constants, you'll have to ask them. Sounds like those should go in a separate crate that defines some of those extra constants, as I imagine one exists, but I don't know which one.

Susurrus:
This is being upstreamed in rust-lang/libc#599.

xd009642:
This has now been merged into libc. Closing the pull request.

Susurrus:
Thanks, @xd009642 for getting that pushed.

xd009642:
Follow up question @susurrus, does rust-nix have a gitter or is there a means to message without opening a PR or issue?

Susurrus:
Yeah, there's a gitter. We need to add it to our README (see #597).

PR #608, opened 2017-05-26
Add UserCredentials struct as the result of PeerCredentials sockopt

### LINKED ISSUE #493: All fields in ucred struct are private
I'm using `PeerCredentials` functionality to extract the remote end of a unix domain socket, but all of the fields in the resulting `ucred` struct are private. I can work around this by doing an unsafe reinterepret cast (err, `transmute`) to a  #struct of my own definition with the same fields and layout, but would it be reasonable to just redefine `struct ucred` to have public fields?

icorderi:
(Fixes #493)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `nix::sys::statfs::{statfs,fstatfs}` uses statfs definition from `libc::statfs` instead of own linux specific type `nix::sys::Statfs`.
   Also file system type constants like `nix::sys::statfs::ADFS_SUPER_MAGIC` were removed in favor of the libc equivalent.
   ([#561](https://github.com/nix-rust/nix/pull/561))
+- Renamed `ucred` to `UserCredentials` and exposed fields through additional methods.
+  ([#493](https://github.com/nix-rust/nix/issues/493)).
 
 ### Removed
 - Removed io::Error from nix::Error and conversion from nix::Error to Errno
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -3,9 +3,11 @@
 //! [Further reading](http://man7.org/linux/man-pages/man7/socket.7.html)
 use {Error, Errno, Result};
 use features;
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
+use unistd::{Pid, Uid, Gid};
 use fcntl::{fcntl, FD_CLOEXEC, O_NONBLOCK};
 use fcntl::FcntlArg::{F_SETFD, F_SETFL};
-use libc::{c_void, c_int, socklen_t, size_t, pid_t, uid_t, gid_t};
+use libc::{c_void, c_int, socklen_t, size_t};
 use std::{mem, ptr, slice};
 use std::os::unix::io::RawFd;
 use sys::uio::IoVec;
@@ -527,12 +529,36 @@ pub struct linger {
     pub l_linger: c_int
 }
 
+/// Kernel credential structure.
+///
+/// This structure is equivalent to [libc::ucred].
+///
+/// [libc::ucred]: http://www.man7.org/linux/man-pages/man7/unix.7.html
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
 #[repr(C)]
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
-pub struct ucred {
-    pid: pid_t,
-    uid: uid_t,
-    gid: gid_t,
+pub struct UserCredentials {
+    pid: Pid,
+    uid: Uid,
+    gid: Gid,
+}
+
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
+impl UserCredentials {
+    /// Get the process ID of the sending process.
+    pub fn pid(&self) -> Pid {
+        self.pid
+    }
+
+    /// Get the user ID of the sending process.
+    pub fn user(&self) -> Uid {
+        self.uid
+    }
+
+    /// Get the group ID of the sending process.
+    pub fn group(&self) -> Gid {
+        self.gid
+    }
 }
 
 /*
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -153,7 +153,7 @@ sockopt_impl!(Both, OobInline, consts::SOL_SOCKET, consts::SO_OOBINLINE, bool);
 sockopt_impl!(GetOnly, SocketError, consts::SOL_SOCKET, consts::SO_ERROR, i32);
 sockopt_impl!(Both, KeepAlive, consts::SOL_SOCKET, consts::SO_KEEPALIVE, bool);
 #[cfg(all(target_os = "linux", not(target_arch="arm")))]
-sockopt_impl!(GetOnly, PeerCredentials, consts::SOL_SOCKET, consts::SO_PEERCRED, super::ucred);
+sockopt_impl!(GetOnly, PeerCredentials, consts::SOL_SOCKET, consts::SO_PEERCRED, super::UserCredentials);
 #[cfg(any(target_os = "macos",
           target_os = "ios"))]
 sockopt_impl!(Both, TcpKeepAlive, consts::IPPROTO_TCP, consts::TCP_KEEPALIVE, u32);
@@ -382,12 +382,13 @@ mod test {
     #[test]
     fn can_get_peercred_on_unix_socket() {
         use super::super::*;
+        use ::unistd::Pid;
 
         let (a, b) = socketpair(AddressFamily::Unix, SockType::Stream, 0, SockFlag::empty()).unwrap();
         let a_cred = getsockopt(a, super::PeerCredentials).unwrap();
         let b_cred = getsockopt(b, super::PeerCredentials).unwrap();
         assert_eq!(a_cred, b_cred);
-        assert!(a_cred.pid != 0);
+        assert!(a_cred.pid() != Pid::from_raw(0));
     }
 
     #[test]
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -20,6 +20,7 @@ pub use self::linux::*;
 ///
 /// Newtype pattern around `uid_t` (which is just alias). It prevents bugs caused by accidentally
 /// passing wrong value.
+#[repr(C)] // FIXME: switch to repr(transparent) once RFC 1758 hits stable
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub struct Uid(uid_t);
 
@@ -64,6 +65,7 @@ pub const ROOT: Uid = Uid(0);
 ///
 /// Newtype pattern around `gid_t` (which is just alias). It prevents bugs caused by accidentally
 /// passing wrong value.
+#[repr(C)] // FIXME: switch to repr(transparent) once RFC 1758 hits stable
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub struct Gid(gid_t);
 
@@ -100,6 +102,7 @@ impl fmt::Display for Gid {
 ///
 /// Newtype pattern around `pid_t` (which is just alias). It prevents bugs caused by accidentally
 /// passing wrong value.
+#[repr(C)] // FIXME: switch to repr(transparent) once RFC 1758 hits stable
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub struct Pid(pid_t);
 


Susurrus:
You've got some test failures with this implementation that need fixing. Let me know if you need help addressing any of them.

src/sys/socket/mod.rs
```diff
@@ -511,14 +513,28 @@ pub struct linger {
...
-pub struct ucred {
+pub struct UserCredentials {
     pid: pid_t,
     uid: uid_t,
     gid: gid_t,
 }
 
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
+impl UserCredentials {
+    pub fn pid(&self) -> i32 {
```
Susurrus: Why do these convert from the base type to `{i|u}32`?

icorderi: Copy/Paste from the issue. 
I'll change them to `{p|u|g}id_t`.

Susurrus:
This needs to be rebase this to include the latest changes to CI. Additionally, now that we have `Pid`, `Uid`, and `Gid` types, you should return those instead of the `*_t` types. here.

Finally, we need an entry in the CHANGELOG under `Fixed`. Please mention the struct being renamed and the fields now being accessible.

Susurrus:
You've some test failures to fix for this.

icorderi:
@Susurrus seems like we can't compare the new `Pid` type with a `pid_t`. 
Should we add that comparison or change the assert to use `Pid::from_raw`?

```diff
    #[cfg(all(target_os = "linux", not(target_arch = "arm")))]
    #[test]
    fn can_get_peercred_on_unix_socket() {
        use super::super::*;

        let (a, b) = socketpair(AddressFamily::Unix, SockType::Stream, 0, SockFlag::empty()).unwrap();
        let a_cred = getsockopt(a, super::PeerCredentials).unwrap();
        let b_cred = getsockopt(b, super::PeerCredentials).unwrap();
        assert_eq!(a_cred, b_cred);
---     assert!(a_cred.pid() != 0);
+++     assert!(a_cred.pid() != Pid::from_raw(0));
    }
```

@Kixunil since you added `Pid` on b62ab099c8aa3ae59a9298fee342cd316f14a8b1, what are your thoughts on this?

Susurrus:
Let's just change it to `from_raw`. It does bring up a good question, should the return value of `pid()` actually be an `Option<Pid>`?

icorderi:
On which case do you return `None`?
Pid 0 is a valid pid value (used for the swapper) and Pid 1 is init. This might only be relevant on the kernel side though. 

I also noticed `Pid|Uid|Gid` are missing `#[repr(C)]`, I'm not certain of what the `repr(Rust)` layout will hold in the future regarding new types, but we might want to be on the safe sire given that we are grabbing those things from **C** land.

Susurrus:
> On which case do you return None?

It would depend on the function. `pid_t` seems to be overloaded where different integer ranges can mean different things (a great example is `waitpid()`. I don't think there's a universally-good datatype here given the age and inconsistencies across all these APIs.

> I also noticed Pid|Uid|Gid are missing #[repr(C)], I'm not certain of what the repr(Rust) layout will hold in the future regarding new types, but we might want to be on the safe sire given that we are grabbing those things from C land.

This doesn't matter since we aren't using the `Pid`/etc types directly with our FFI functions, we're converting them to the proper values first. That being said we should keep an eye on `repr(transparent)` which is designed for this exact use case and should provide atiny performance win.

icorderi:
> This doesn't matter since we aren't using the Pid/etc types directly with our FFI functions

I do not thing that is correct @Susurrus.

If you follow the implementation:

```rust
sockopt_impl!(GetOnly, PeerCredentials, consts::SOL_SOCKET, consts::SO_PEERCRED, super::UserCredentials);
```

We are returning the new `UserCredentials`.  The `sockopt_impl` macro ends up calling this other macro:

```rust
macro_rules! getsockopt_impl {
    ($name:ident, $level:path, $flag:path, $ty:ty, $getter:ty) => {
        impl GetSockOpt for $name {
            type Val = $ty;

            fn get(&self, fd: RawFd) -> Result<$ty> {
                unsafe {
                    let mut getter: $getter = Get::blank();

                    let res = ffi::getsockopt(fd, $level, $flag,
                                              getter.ffi_ptr(),
                                              getter.ffi_len());
                    try!(Errno::result(res));

                    Ok(getter.unwrap())
                }
            }
        }
    }
}
```

That `$getter:ty` resolves to a `GetStruct<UserCredentials>`, which looks like this:

```rust
struct GetStruct<T> {
    len: socklen_t,
    val: T,
}

impl<T> Get<T> for GetStruct<T> {
    unsafe fn blank() -> Self {
        GetStruct {
            len: mem::size_of::<T>() as socklen_t,
            val: mem::zeroed(),
        }
    }

    unsafe fn ffi_ptr(&mut self) -> *mut c_void {
        mem::transmute(&mut self.val)
    }

    unsafe fn ffi_len(&mut self) -> *mut socklen_t {
        mem::transmute(&mut self.len)
    }

    unsafe fn unwrap(self) -> T {
        assert!(self.len as usize == mem::size_of::<T>(), "invalid getsockopt implementation");
        self.val
    }
}
```

That memory on the `UserCredentials` is being filled on the C side.

Susurrus:
You're definitely correct that it's not correct the way it is, though we actually need `repr(transparent)` here not `repr(C)` (for details [see the RFC](https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md)). I'd suggest changing to to `repr(C)` for now with a `FIXME` to change it once that RFC hits stable. Please add an extra commit with a detailed message (before all your other commits) that does this.

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,26 @@ pub struct linger {
...
 }
 
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
 #[repr(C)]
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
-pub struct ucred {
-    pid: pid_t,
-    uid: uid_t,
-    gid: gid_t,
+pub struct UserCredentials {
```
Susurrus: Needs doc comments

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,26 @@ pub struct linger {
...
+    pid: Pid,
+    uid: Uid,
+    gid: Gid,
+}
+
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
+impl UserCredentials {
+    pub fn pid(&self) -> Pid {
```
Susurrus: Needs doc comment.

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,26 @@ pub struct linger {
...
+        self.pid
+    }
+    pub fn user(&self) -> Uid {
```
Susurrus: Needs doc comment. And a blank line above.

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,26 @@ pub struct linger {
...
+        self.uid
+    }
+    pub fn group(&self) -> Gid {
```
Susurrus: Needs doc comment. And a blank line above.

icorderi:
@Susurrus added docs and rebased with master.

CHANGELOG.md
```diff
@@ -60,6 +60,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `nix::ptrace` on all Linux-kernel-based platforms
   [#624](https://github.com/nix-rust/nix/pull/624). Previously it was
   only available on x86, x86-64, and ARM, and also not on Android.
+- Fixed missing public fields on `ucred` and renamed it to `UserCredentials`
```
Susurrus: This is incorrect. Better to say "Renamed `ucred` to `UserCredentials` and exposed fields through additional methods." Also, this should be moved to the `Changed` section. This is a feature addition & API change rather than a bugfix.

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,34 @@ pub struct linger {
     pub l_linger: c_int
 }
 
+/// Kernel credential structure.
+///
+/// This structure is equivalent to the [libc::ucred].
```
Susurrus: Remove the "the". Please also link to a manpage for this on man7.org.

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,34 @@ pub struct linger {
...
-    gid: gid_t,
+pub struct UserCredentials {
+    pid: Pid,
+    uid: Uid,
+    gid: Gid,
+}
+
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
+impl UserCredentials {
+    /// Process ID of the sending process.
```
Susurrus: Rephrase to "Get the process ID of the sending process"

Kixunil:
I agree that we need `repr(transparent)`. I'd prefer using raw types on FFI boundaries and converting to Rust types, instead of `repr(C)`.

icorderi:
@Kixunil, you mean this?

```rust
#[repr(C)]
#[derive(Clone, Copy, PartialEq, Eq, Debug)]
pub struct UserCredentials {
    pid: pid_t,
    uid: uid_t,
    gid: gid_t,
}

impl UserCredentials {
    /// Get the process ID of the sending process.
    pub fn pid(&self) -> Pid {
        Pid::from_raw(self.pid)
    }
    // ...
}
```

or getting rid of the `#[repr(C)]` in `UserCredentials` entirely?

Kixunil:
@icorderi That looks like another interesting way to approach this. What I was thinking about before was:

```rust
#[repr(C)]
struct UserCredentialsRaw {
    pid: pid_t,
    uid: uid_t,
    gid: gid_t,
}

pub struct UserCredentials {
    pub pid: Pid,
    pub uid: Uid,
    pub gid: Gid,
}
```

With obvious conversion.

Susurrus:
`ucred` is already defined by `libc`, so we don't need to redeclare that struct. I think what we want is a newtype around `ucred` named `UserCredentials` and then exposing its various fields via the `pid_t` and other newtypes should work.

icorderi:
@Susurrus is this what you had in mind?

```rust
#[repr(C)] // FIXME: change to repr(transparent)
#[derive(Clone, Copy, PartialEq, Eq, Debug)] // derives wont work
pub struct UserCredentials(ucred);

impl UserCredentials {
    /// Get the process ID of the sending process.
    pub fn pid(&self) -> Pid {
        Pid::from_raw(self.0.pid)
    }
    // ...
}
```

The [ucred] in libc doesn't have `Copy/Clone/Debug/Eq` on it. So we might have to push that upstream if you want this to be done on a new type (or do all the impl's by hand). 

[ucred]: https://doc.rust-lang.org/libc/x86_64-unknown-linux-gnu/libc/struct.ucred.html

What @Kixunil is suggesting it to keep them completely separate. After looking at the code it seems to be possible by using a custom `Get`'ter.

```rust
trait Get<T> {
    unsafe fn blank() -> Self;
    unsafe fn ffi_ptr(&mut self) -> *mut c_void;
    unsafe fn ffi_len(&mut self) -> *mut socklen_t;
    unsafe fn unwrap(self) -> T;
}
```

```diff
--- sockopt_impl!(GetOnly, PeerCredentials, consts::SOL_SOCKET, consts::SO_PEERCRED, super::UserCredentials);
+++ sockopt_impl!(GetOnly, PeerCredentials, consts::SOL_SOCKET, consts::SO_PEERCRED, super::UserCredentials, ucred);
```
we would have to `impl Get<super::UserCredentials> for ucred` instead of using the default `GetStruct<T>`.

-----

Which path do we want to take?

Kixunil:
@icorderi I don't see why we would need `Get` trait. Simply write this:

```rust
impl From<ucred> for UserCredentials {
    fn from(cred: ucred) -> Self {
        UserCredentials {
            pid: cred.pid.into(),
            uid: cred.uid.into(),
            gid: cred.gid.into(),
        }
    }
}
```

Sure, there's a little performance hit but given that it involves syscall anyway and probably won't be called often (no reason to call it twice on same socket), I don't see a huge issue with it.

OT, you might also be interested in [my PR to tokio-uds](https://github.com/tokio-rs/tokio-uds/pull/13).

Susurrus:
@icorderi All traits in `libc` are `Copy` and `Clone` (they're declared within the `s!` macro which hides it, but if you generate the docs you'll see it. `Debug` and `PartialEq`/`Eq` will not be implemented for the underlying type, but those are trivial enough to implement in our wrapper object if we want to (is it just `pid`, `uid`, and `gid`?), but even if we don't are there use cases where you really want that functionality?

As for the exact path forward, I don't particularly care. I think the data copying would be great to avoid, but as @Kixunil says, this is a low-bandwidth function call so it shouldn't be an issue.

Kixunil:
Pros of public fields:

* simple, clear design
* no future surprises (nobody will change the struct without major version bump)
* no need to write and document methods

Pros of getters:

* ability to extend in the future
* ability to change internal representation
* possibly less copying (assuming compiler will optimize them)

Another suggestions:

```rust
#[repr(transparent)] //< same for Uid and Gid
struct Pid(pid_t);

#[repr(C)]
struct UserCredentials {
    pub pid: Pid,
    pub uid: Uid,
    pub gid: Gid,
}
```

`UserCredentials` would have the same layout as `ucred`, so it could be used in the syscall and in public interface interchangeably. Since we don't have `repr(transparent)` yet, we could copy the fields now and change it later (it'd be backward-compatible). Alternatively, we could use `repr(C)` and hope for the best.

Susurrus:
One other pro to the getters is that it enforces that these are read-only fields. Is that all this struct will ever be used for or would there be some real use case to modify the fields?

Additionally, if we have `repr(transparent)` we can just do:
```
struct UserCredentials(ucred);

impl UserCredentials {
   pid() -> ...
}
```

Though of course `Pid` and friends should also be made `repr(transparent)`.

Kixunil:
Getters indeed seem to be more conservative. If I'm not mistaken, it will always be fine to change it. The only concern I see now is if we implement it via getters now and decide to change it later, then there'd be a name clash (`self.pid` - field, `self.pid()` - method)

Another thing to consider: in case of Linux, it's possible to get PID of peer process, however, it's impossible on other Unices. Since it's called **User**Credentials, we might want it to contain only `uid` and `gid` and provide `ProcessCredentials` that would include `pid`, or some other way so the users would be aware of limitations. (This is essential if we want to implement peer credentials for non-Linux Unices.)

Susurrus:
Sounds like we might want to address the lack of `ucred` on other platforms as part of our planning for this, since that could affect the final API. Though we could only expose the `pid` method on systems where that is exposed, which shouldn't be a big deal.

icorderi:
@Susurrus, I'm not familiar enough with the uses of `ucred` to know if a user will have to modify or create a `ucred` object to interact with the kernel. 

As far as the current use, those fields are **read-only**.  

But assuming someone will want to modify it, or create one (at least for testing), I say we make them public fields. The downside is to expose the user to potentially changing the value of something we gave him. I can live with that. 

The other design point was on wether to use the  `Get<T>` trait or `#[repr(C)]`/`#[repr(transparent)]`. Since we don't have `#[repr(transparent)]` to do the newtypes correctly, we might as well use the `Get<T>` and pay for the extra copy. 
We can leave a **FIXME** for making `Pid` et.al. `#[repr(transparent)]` and when we do that, also leave switch away from the `Get<T>` and just add `#[repr(C)]` to `UserCredentials`.

Kixunil:
What we may want:

```rust
pub struct ProcessCredentials {
    #[cfg(linux)]
    pub pid: Pid,
    pub uid: Uid,
    pub gid: Gid,
    // This is to prevent users accidentally making non-portable code
    _private: (),
}
```

icorderi:
@Kixunil what does the `ucred` structure look like on the systems that do not support the pid? 
I don't see a `ucred` on any of the other unixes. 

All the code regarding `UserCredentials` is guarded with a `#[cfg(all(target_os = "linux", not(target_arch="arm")))]`.

src/sys/socket/mod.rs
```diff
@@ -527,12 +529,36 @@ pub struct linger {
...
+
+#[cfg(all(target_os = "linux", not(target_arch="arm")))]
+impl UserCredentials {
+    /// Get the process ID of the sending process.
+    pub fn pid(&self) -> Pid {
+        self.pid
+    }
+
+    /// Get the user ID of the sending process.
+    pub fn user(&self) -> Uid {
```
vincenthage: In terms of naming, there is first `pid`, 
but then `user` and `group`, instead of `uid` and `gid`.
Won't this be confusing for the user (who would for instance expect `user` to be a string)?

Kixunil:
Exactly. The thing is that it's still possible to provide reasonable implementation for all unices which allows users to learn peer UID and GID, but not PID. Linux is special in that it allows learning PID too.

Susurrus:
Any progress on this?

Susurrus:
This hasn't been touched in almost 5 months, so I'm going to close this.

PR #611, opened 2017-06-01
Add SigAction getters

henninglive:
I want to inspect the old SigAction after installing a new one. This adds getter methods for handler, flags and mask. Returning the old SigAction isn't very useful if we can't read its fields.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -315,7 +315,7 @@ impl AsRef<libc::sigset_t> for SigSet {
 }
 
 #[allow(unknown_lints)]
-#[derive(Clone, Copy, PartialEq)]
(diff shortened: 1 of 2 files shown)


src/sys/signal.rs
```diff
@@ -639,6 +658,25 @@ mod tests {
         assert!(SigSet::thread_get_mask().unwrap().contains(SIGUSR2));
     }
 
+    fn test_sigaction() {
```
asomers: Missing a `#[test]` decorator

henninglive:
Should be fixed now

asomers:
LGTM, thanks!

asomers:
bors +r

Issue #609, opened 2017-05-30
*Flags don't implement set-like combinators


cyphar:
One thing I ran into almost immediately when trying to nicely define a set of flags like you would in C (with something like `CLONE_NEWUSER|CLONE_NEWUTS|CLONE_NEWNS` or whatever) is that you can't do it with the `nix` structs without declaring a variable `mut` and operating on it.

```rust
let mut sigmask = signal::SigSet::empty();
sigmask.add(signal::Signal::SIGTTOU);
sigmask.add(signal::Signal::SIGTTIN);
// etc
```

What if we made it so you could do this instead?

```rust
let sigmask = signal::SigSet::empty()
                             .add(signal::Signal::SIGTTOU)
                             .add(signal::Signal::SIGTTIN);
```

Which allows you to write it (IMO) more concisely but also doesn't mean that you have to make the variable mutable. It also better matches the `Iterator` combinators.

Susurrus:
You should be able to do
```
let sigmask = signal::Signal::SIGTTOU | signal::Signal::SIGTTIN;
```

Does that not work for you?

posborne:
Closing as I'm assuming this meets or exceeds @cyphar's requirements.

cyphar:
Sorry for not responding earlier. No, this code doesn't work:

```
   Compiling initrs v0.0.0 (file:///home/cyphar/.local/src/initrs)
error[E0369]: binary operation `|` cannot be applied to type `nix::sys::signal::Signal`
   --> src/main.rs:117:19
    |
117 |     let sigmask = signal::Signal::SIGTTOU | signal::Signal::SIGTTIN;
    |                   ^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: an implementation of `std::ops::BitOr` might be missing for `nix::sys::signal::Signal`

error: aborting due to previous error

error: Could not compile `initrs`.
```

My `Cargo.toml`:

```
[dependencies.nix]
version = "~0.8.1"
default-features = true
features = ["signalfd"]
```

cyphar:
But the solution would be to implement `BitOr` for `SigSet` and also implementing it for `Signal` such that it produces a `SigSet`.

cyphar:
I've implemented this in #615, as well as adding a `From` `impl` for `SigSet::from(Signal)`.

PR #615, opened 2017-06-03
implement signal BitOr and SigSet::from!

cyphar:
This allows for far more idiomatic initialisation of SigSet without
having to declare mut variables, as well as mirroring the vec![]
semantics (though arguably SigSet is closer to a bitfield than a
vector).

    let mut sigset = SigSet::empty();
    sigset.add(Signal::SIGINT);
    // versus
    let sigset = sigset![Signal::SIGINT];

And the BitOr* operations futher make it much easier to idiomatically
work with SigSet and Signal, without having to use sigset![] and
SigSet::extend() everywhere.

    let sigset_a = sigset![Signal::SIGINT];
    let sigset_b = sigset![Signal::SIGTERM];
    let mut sigset = SigSet::empty();
    sigset.extend(&sigset_a);
    sigset.extend(&sigset_b);
    // or
    let sigset = sigset![Signal::SIGINT, Signal::SIGTERM];
    // versus
    let sigset = Signal::SIGINT | Signal::SIGTERM;

Implements #609
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - On Linux and Android, added support for receiving `PTRACE_O_TRACESYSGOOD`
   events from `wait` and `waitpid` using `WaitStatus::PtraceSyscall`
   ([#566](https://github.com/nix-rust/nix/pull/566)).
+- Added `nix::sys::Signal::BitOr<Into<SigSet>>`,
+  `nix::sys::SigSet::BitOr<Into<SigSet>>`, and
+  `nix::sys::SigSet::BitOrAssign<Into<SigSet>>`
+  ([#615](https://github.com/nix-rust/nix/pull/615)
 
 ### Changed
 - Changed `ioctl!(write ...)` into `ioctl!(write_ptr ...)` and `ioctl!(write_int ..)` variants
@@ -57,6 +61,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   "*_buf" variants also now calculate total array size and take slice references for improved type
   safety. The documentation has also been dramatically improved.
   ([#670](https://github.com/nix-rust/nix/pull/670))
+- Added `nix::sys::SigSet::From<Signal>` and refactored some APIs to allow
+  passing `Into<SigSet>` rather than bare `SigSet`s.
+  ([#615](https://github.com/nix-rust/nix/pull/615)
 
 ### Removed
 - Removed `io::Error` from `nix::Error` and the conversion from `nix::Error` to `Errno`
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -9,6 +9,7 @@ use std::mem;
 #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
 use std::os::unix::io::RawFd;
 use std::ptr;
+use std::ops::{BitOr, BitOrAssign};
 
 // Currently there is only one definition of c_int in libc, as well as only one
 // type for signal constants.
@@ -21,11 +22,29 @@ use std::ptr;
 ///
 /// # Examples
 ///
-/// `Into` and `From` promote a `Signal` to a `SigSet` that has only the given `Signal` inside it.
+/// Note that any `Signal` can be promoted to a `SigSet` either through the use of the `|`
+/// (`BitOr`) operator, or through `std::convert::Into`.
+///
+/// `Into` and `From` promote a signal to a `SigSet` that has only the given `Signal` inside it.
 /// ```
 /// # use ::nix::sys::signal::*;
 /// let sigset_a: SigSet = Signal::SIGTERM.into();
 /// let sigset_b = SigSet::from(Signal::SIGTERM);
+/// ```
+///
+/// `SigSet`s of multiple `Signal`s can be created using the `|` (`BitOr`) operator.
+/// ```
+/// # use ::nix::sys::signal::*;
+/// let sigset = Signal::SIGTERM | Signal::SIGHUP;
+/// ```
+///
+/// Existing `SigSet`s can be extended using the `|` (`BitOr`) or `|=` (`BitOrAssign`) operator as
+/// well (note that it is also commutative).
+/// ```
+/// # use ::nix::sys::signal::*;
+/// let mut sigset = SigSet::empty();
+/// sigset |= Signal::SIGQUIT;
+/// sigset |= Signal::SIGHUP | Signal::SIGUSER;
 /// ```
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 #[repr(i32)]
@@ -207,6 +226,42 @@ impl Signal {
     }
 }
 
+impl BitOr for Signal {
+    type Output = SigSet;
+
+    fn bitor(self, rhs: Self) -> Self::Output {
+        let mut copy = SigSet::empty();
+        copy.add(self);
+        copy.add(rhs);
+        copy
+    }
+}
+
+impl BitOr<Signal> for SigSet {
+    type Output = SigSet;
+
+    fn bitor(self, rhs: Signal) -> Self::Output {
+        let mut copy = self;
+        copy.add(rhs);
+        copy
+    }
+}
+
+impl BitOr<SigSet> for Signal {
+    type Output = SigSet;
+
+    // bitor is commutative.
+    fn bitor(self, rhs: SigSet) -> Self::Output {
+        rhs | self
+    }
+}
+
+impl BitOrAssign<Signal> for SigSet {
+    fn bitor_assign(&mut self, rhs: Signal) {
+        self.add(rhs);
+    }
+}
+
 pub const SIGIOT : Signal = SIGABRT;
 pub const SIGPOLL : Signal = SIGIO;
 pub const SIGUNUSED : Signal = SIGSYS;
@@ -261,12 +316,21 @@ pub enum SigmaskHow {
     SIG_SETMASK = libc::SIG_SETMASK,
 }
 
-/// `SigSet` represents a set of `Signal`s and is a thin wrapper around the libc APIs for
-/// `sigset_t`.
+/// `SigSet` represents a set of signals and is a thin wrapper around the libc APIs for `sigset_t`.
+/// However, due to the clunky nature of the libc APIs around sigset_t (`SigSet::add`,
+/// `SigSet::extend` and so on), a more idiomatic way of constructing `SigSet`s is possible using
+/// `|` (`BitOr`) and `|=` (`BitOrAssign`).
 ///
 /// # Examples
 ///
-/// `Signal`s can be promoted to `SigSet`s through the use of `std::convert::Into`.
+/// ```
+/// # use ::nix::sys::signal::*;
+/// let mut sigset = Signal::SIGTERM | Signal::SIGHUP;
+/// sigset |= Signal::SIGKILL;
+/// sigset |= Signal::SIGUSR1 | Signal::SIGUSR2;
+/// ```
+///
+/// In addition, `Signal`s can be promoted to `SigSet`s through the use of `std::convert::Into`.
 /// ```
 /// # use ::nix::sys::signal::*;
 /// let sigset_a: SigSet = Signal::SIGTERM.into();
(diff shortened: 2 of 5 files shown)


Susurrus:
I'm not a huge fan of using `.into()` to convert a `Signal` to a `SigSet`. They're not really equivalent.

The only use case I see missing here is supporting nice initialization. What about something akin to the `vec![]` macro but for `SigSet`?

Susurrus:
Additionally I'm not a fan of using `BitOr` to OR together things that aren't really bitfields. I think it's semantically confusing.

cyphar:
> I'm not a huge fan of using .into() to convert a Signal to a SigSet. They're not really equivalent.

I will admit that I'm very new to Rust so I'm not sure what the right semantics are (and am quite willing to learn on what the "right way" of defining interfaces is in the Rust ecosystem). Is `std::convert::From::from` similarly "bad"?

> I think it's semantically confusing.

While the underyling interface isn't a bitfield (and is accessed through the super-clumsy `libc` interfaces), a `SigSet` **is** semantically a set so I'm not sure I agree that set operations shouldn't be defined on something that is a set. I don't know if you use use Python, but Python `set`s define unary operations such as `|` and so on. It's unfortunate that Rust called the trait `BitOr` because `|` is quite useful for other things :wink:.

cyphar:
> What about something akin to the vec![] macro but for SigSet?

That could also work I guess, though now users will have to use `#[macro_use]`.

Susurrus:
> I will admit that I'm very new to Rust so I'm not sure what the right semantics are (and am quite willing to learn on what the "right way" of defining interfaces is in the Rust ecosystem). Is std::convert::From::from similarly "bad"?

`From` and `Into` normally mean that the types are roughly equivalent. Like a `RawFd` and a `u32`, the datatypes are similar semantically. With a `Signal` and a `SigSet`, I don't like that you can `into()`, as they aren't equivalent. One is a container of another. I see it like how I don't think you should implement `u32::Into<Vec<u32>>` since I don't consider those equivalent. One is effectively a container and the other is an element.

Additionally the value of `Into` seems quite minimal. Your `bitor` and `bitor_assign` tests cover all use cases I can imagine and that you have provided.

One thing I'd like to see is that `SigSet` get some expanded documentation describing these various ways of initializing them like you mention in this comment thread. That should make sure that future users don't get confused on how to best generate a `SigSet`.

This is looking really good though, that's for taking this on!

cyphar:
@Susurrus Done, PTAL?

src/sys/signal.rs
```diff
@@ -218,11 +256,88 @@ pub enum SigmaskHow {
...
+/// ```
+///
+/// The above would traditionally require far more finicking with SigSet::add, SigSet::extend and
+/// so on. For further idiomatic ways to interact with SigSets see the sigset! macro.
 #[derive(Clone, Copy)]
 pub struct SigSet {
     sigset: libc::sigset_t
 }
 
+/// sigset is a macro that allows for far more idiomatic initalisation of SigSet. Note that while
```
Susurrus: I don't think we should have a `sigset![]` macro at all. I don't see what it offers that the `BitOr` operator doesn't allow. And the ergonomics with macros in Rust right now are less than that of regular functions or operators. Let's go ahead and just remove it.

src/sys/signal.rs
```diff
@@ -9,11 +9,13 @@ use std::mem;
 #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
 use std::os::unix::io::RawFd;
 use std::ptr;
+use std::ops::{BitOr, BitOrAssign};
 
 // Currently there is only one definition of c_int in libc, as well as only one
 // type for signal constants.
 // We would prefer to use the libc::c_int alias in the repr attribute. Unfortunately
 // this is not (yet) possible.
+
```
Susurrus: Please remove this line.

Also, `Signal` needs doc comments now that there's this new behavior where you can combine `Signal`s into `SigSet`s.

src/sys/signal.rs
cyphar: The downside is that it requires doing `SigSet::empty() | Signal::SIGHUP` in order to "promote" a `Signal` to a `SigSet`.

Susurrus: Yeah, that's a great point. I'm actually thinking that going back to implementing `From<Signal> for SigSet` to be the right solution. Then we can rewrite all functions that take a `SigSet` to instead take a `T: Into<SigSet>` instead.

posborne: Yeah, `From<Signal> for SigSet` is a very reasonable thing to do IMO here.  Agreed that it would be better to not have that macro unless it is really required.

cyphar: Alright, I'll go back to that. Do you want me to do the `T: Into<SigSet>` here or in a separate patchset?

Susurrus: Same patchset. It's fine to have the `From` impl and the conversion of functions to use it as another commit.

cyphar:
Alright, I've updated this to have `impl From<Signal> for SigSet`. I've also added the `Into<SigSet>`, and I fixed the `signalfd` tests and documentation (which used to not even compile).

cyphar:
I'm sad that you can't implement a trait for a generic type. It would be nice to be able to do

```rust
impl<T: Into<SigSet>> BitOr<T> for T { /* ... */ }
```

src/sys/signal.rs
```diff
@@ -9,11 +9,37 @@ use std::mem;
...
+/// of Signals.
+///
+/// # Examples
+///
+/// Note that any Signal can be promoted to a SigSet either through the use of the `|` operator, or
+/// through std::convert::Into.
+///
+/// ```
+/// # use ::nix::sys::signal::*;
+/// // Into and From promote a signal to a SigSet that has only the given signal inside it.
```
Susurrus: Please leave these comments out of the code example instead and separate these into two separate examples. It'll make it slightly easier reading. Something like:
```
/// Into and From promote a signal to a SigSet that has only the given signal inside it.
/// ```
/// # use ::nix::sys::signal::*;
/// let sigset_a: SigSet = Signal::SIGTERM.into();
/// let sigset_b = SigSet::from(Signal::SIGTERM);
///```
///
/// `SigSet`s of multiple signals can be created by using the `|` (`BitOr`) operator:
/// ```
/// # use ::nix::sys::signal::*;
/// let sigset = Signal::SIGTERM | Signal::SIGHUP;
/// ```
///
/// Existing `SigSet`s can be extended using the `|` (`BitOr`) or `|=` (`BitAssign`) operator as well:
/// ```
/// let mut sigset = SigSet::empty();
/// sigset |= Signal::SIGQUIT;
/// ```
```

src/sys/signal.rs
```diff
@@ -218,11 +263,42 @@ pub enum SigmaskHow {
     SIG_SETMASK = libc::SIG_SETMASK,
 }
 
+/// SigSet represents a set of signals and is a thin wrapper around the libc APIs for sigset_t.
```
Susurrus: Please put all types in "\`" like `sigset_t`, etc.

src/sys/signal.rs
```diff
@@ -218,11 +263,42 @@ pub enum SigmaskHow {
...
+/// However, due to the clunky nature of the libc APIs around sigset_t (SigSet::add, SigSet::extend
+/// and so on), a more idiomatic way of constructing SigSets is possible using BitOr.
+///
+/// # Examples
+///
+/// ```
+/// # use ::nix::sys::signal::*;
```
Susurrus: Please simplify this example to one `SigSet`. Maybe you should just copy/paste the relevant examples from the `Signal` docs above.

src/sys/signal.rs
```diff
@@ -218,11 +263,42 @@ pub enum SigmaskHow {
...
+/// let sigset_a = Signal::SIGTERM | Signal::SIGHUP;
+/// let sigset_b = sigset_a | Signal::SIGUSR1;
+/// let mut sigset_c = SigSet::empty();
+/// sigset_c |= Signal::SIGUSR2;
+/// sigset_c |= sigset_b;
+/// let sigset_d = sigset_c | sigset_a;
+/// ```
+///
+/// The above would traditionally require far more finicking with SigSet::add, SigSet::extend and
```
Susurrus: I think the first sentence here doesn't add anything, let's remove it.

src/sys/signalfd.rs
```diff
@@ -60,10 +60,10 @@ pub fn signalfd(fd: RawFd, mask: &SigSet, flags: SfdFlags) -> Result<RawFd> {
 /// # Examples
 ///
 /// ```
+/// use nix::sys::signal::*;
```
Susurrus: Let's make these `use` lines not appear in the docs by prepending with `#` since we're modifying this example.

test/test_signalfd.rs
```diff
@@ -11,8 +11,7 @@ use nix::unistd;
 fn main() {
     print!("test test_signalfd ... ");
 
-    let mut mask = signal::SigSet::empty();
-    mask.add(signal::SIGUSR1).unwrap();
+    let mask: signal::SigSet = signal::SIGUSR1.into();
```
Susurrus: We should allow a `Signal` to be passed directly into `SignalFd::new()` and related functions. Could you add that?

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#551](https://github.com/nix-rust/nix/pull/551))
 - Added `nix::pty::{grantpt, posix_openpt, ptsname/ptsname_r, unlockpt}`
   ([#556](https://github.com/nix-rust/nix/pull/556)
+- Added `nix::sys::SigSet::From<Signal>` and refactored some APIs to allow
```
Susurrus: Since this is a change, it should be put down in the `Changed` section.

Susurrus:
Getting close now. Some documentation cleanups and a few more functions to convert to use `Into<SigSet>` and then this should be finished.

test/test_signalfd.rs
cyphar: I tried to figure out how to do this and I've bumped up against the edges of my knowledge of Rust. Because `SignalFd::new` (and the other various methods for `SignalFd`) take `&SigSet`, we can't easily do something like this:

```
let sigmask = Signal::SIGTERM;
SignalFd::new(sigmask);
// or
SignalFd::new(&sigmask);
```

`impl From<&Signal> for &SigSet` won't work because of trait specialization not being implementated, but there's an even deeper issue. I was trying to get `impl AsRef<SigSet> for Signal` but realised that it doesn't appear to be possible (given Rust's borrowing model) to return an immutable reference to something allocated in the function that you're returning from. As a simple example, this simply doesn't work for obvious reasons:

```rust
impl AsRef<SigSet> for Signal {
    fn as_ref(&self) -> &SigSet {
        &self.clone().into()
    }
}
```

Is there a way to make that work? It seems quite silly that you can't convert `&Into<T>` into `Into<&T>` but there's probably some much harder problem here than I'm seeing.

cyphar: In particular, I'm not really convinced this particular characterbake makes sense:

```rust
pub fn signalfd<'a, T: Into<&'a SigSet>>(fd: RawFd, mask: T, flags: SfdFlags) -> Result<RawFd>
```

In particular, I'm not sure what will actually fulfill the bound `Into<&'a SigSet>` other than `&SigSet` -- making it just a waste of characters.

Susurrus: That'd a good point. My question now is why doesn't `ppoll` take a `&SigSet` instead of by-value?

Susurrus:
@cyphar You still interested in pushing this over the finish line?

cyphar:
Yes. The past few weeks have been my finals of this semester of university, so I've not been able to touch this. I'll rebase and fix up your comments in the coming days. Thanks.

Susurrus:
@cyphar I'd love to get this in the next release of nix. Will you have time coming up here to rebase this so we can get this merged?

cyphar:
@Susurrus Sorry for the delays, other projects have been getting close to release. I've rebased, but we still haven't figured out how to handle https://github.com/nix-rust/nix/pull/615#discussion_r121285145. Should I just remove my `Into<&SigSet>` changes? And what about `ppoll`?

cyphar:
Rebased again.

Susurrus:
This needs another rebase (be sure to move the CHANGELOG entries into the new unreleased section) and there are some tests failing that need to be fixed.

Susurrus:
@cyphar Can you rebase this and fix the tests that were failing previously?

Susurrus:
There's been no update in almost 5 months, so I'm going to go ahead and close this PR.

cyphar:
I didn't have a chance to rebase it because there was still the question about `&SigSet` which wasn't resolved, and I didn't see the benefit of rebasing it without resolving that problem first. If you want, I can rebase and create a new PR, but we should figure out what the right solution is for `signalfd`.

PR #618, opened 2017-06-04
Integrate the FreeBSD BuildBot

### LINKED ISSUE #603: Continuous testing for FreeBSD
I'm working on a buildbot cluster that could support continuous testing on FreeBSD for several projects.  I've got a prototype running, and you can see a PR in action at https://github.com/asomers/mio-aio/pull/2 .  The hardest question is how to secure it.  Since anybody can open a PR, that means that anybody can run arbitrary code on the buildslaves.  The potential damage is limited; each project gets its own worker, each worker runs in its own jail, and there's a timeout on each build.  I could use the firewall to prevent workers from sending email and stuff, but I can't completely isolate workers from the internet without breaking a lot of builds.  There are a few options to improve the security situation.

1. Don't automatically build a PR until a maintainer posts a specific comment.  This is what open-zfs does.  It completely eliminates unreviewed code from running on the workers.  However, it's inconvenient for people who are accustomed to Travis building stuff without needing to be asked.
2. Destroy and reclone the worker for each build.  This is what Travis does.  The worker still runs untrusted code, but not for long.  The untrusted code can't modify the filesystem in any persistent way.
3. A hybrid of the previous two.  Do a build when a maintainer give the magic comment, but also do builds automatically whenever a PR is posted by a known-good contributor.  The list of contributors would probably have to be maintained by hand, but it could be much larger than the list of maintainers.

Does anybody have any better ideas?

asomers:
* Gate Bors on the FreeBSD 11 build
* Remove the testless FreeBSD build from Travis
* Promote x86_64-unknown-freebsd to Tier 1

Fixes #603

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,5 +1,6 @@
-# Gate on Travis CI
-status = ["continuous-integration/travis-ci/push"]
+# Gate on Travis CI and Buildbot
+status = ["continuous-integration/travis-ci/push",
+          "buildbot/nix-rust/nix fbsd11"]
 
 # Set bors's timeout to 4 hours
 #
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
 
     # *BSD
     - env: TARGET=i686-unknown-freebsd DISABLE_TESTS=1
-    - env: TARGET=x86_64-unknown-freebsd DISABLE_TESTS=1
+    # - env: TARGET=x86_64-unknown-freebsd DISABLE_TESTS=1 # Uses BuildBot instead
     - env: TARGET=x86_64-unknown-netbsd DISABLE_TESTS=1
 
     # Other architectures
--- a/README.md
+++ b/README.md
@@ -49,10 +49,10 @@ Tier 1:
   * aarch64-unknown-linux-gnu
   * armv7-unknown-linux-gnueabihf
   * arm-unknown-linux-gnueabi
+  * x86_64-unknown-freebsd
 
 Tier 2:
   * i686-unknown-freebsd
-  * x86_64-unknown-freebsd
   * x86_64-unknown-netbsd
 
 Tier 3:


bors.toml
```diff
@@ -1,5 +1,6 @@
-# Gate on Travis CI
-status = ["continuous-integration/travis-ci/push"]
+# Gate on Travis CI and Buildbot
+status = ["continuous-integration/travis-ci/push",
+	  "buildbot/nix-rust/nix fbsd11"]
```
Susurrus: Minor nitpick, can you align the quotes? The off-by-two is bugging me.

Susurrus:
This seems to only address x86_64 FreeBSD, do you have plans for i686? Or was #603 really just about x86_64?

Susurrus:
After my one comment is resolved, this LGTM.

asomers:
@Susurrus it would be pretty easy to add an x86 worker.  I'd like to do it, but I haven't gotten around to it yet.  It would also be totally awesome to have an arm64 worker, which will be possible as soon as Scaleway.com's ARM64 VPSes go live.  Right now they're just in preview mode.

Susurrus:
Let's do this.

bors r+

notriddle:
Try again? [BuildBot didn't build the merge commit](https://github.com/nix-rust/nix/commits/4b78c430e0d20ca87413d875684cb8dc8cbea383) (based on clicking the X and seeing fbsd11 not listed), but [it built the newer commits done afterward](https://github.com/nix-rust/nix/commits/3be3e7007028453a165cbb5d67e0b11858973f36) (because fbsd11 is listed under the checkmark).

asomers:
I noticed that.  I changed one thing in BuildBot's configuration, but I didn't expect it to solve this particular problem.  I guess we may as well try, try again.  The worst case is that BuildBot is intermittent.  If that happens, we'll have to degate Bors on it.
bors r+

PR #620, opened 2017-06-18
Update bitflags to 0.9

equal-l2:


Susurrus:
Tests failed. Also, @asomers any reason why the freeBSD tests aren't showing up here?

Susurrus:
Looking over these changes, it actually appears to me that at least some of these `bitflags` macros should actually be changed to use the internal `libc_bitflags` macro. Since we're already touching on these values, it'd be great to do that as part of this PR. @equal-l2 could you do that?

equal-l2:
@Susurrus I'd love to. Which `bitflags` do you actually mean?

Susurrus:
Any of the bitflags that reference exclusively `libc` types, which should be most of them. `OFlag` should be assuming that all of the values exist in the `libc` crate, which you'll need to check (they may need to be conditionally-included with `#[config()]` attributes.

equal-l2:
Some of `OFlag` is not exist in `libc` yet.
I'll make a PR for them on libc.

equal-l2:
I'd like to separate the change of `bitflags` to `libc_bitflags` to another PR, since there are more constants we can hand over to `libc`.

Can't I do that?

Susurrus:
@equal-l2 That's fine. Thanks for your work on this!

bors r+

equal-l2:
I think this is not my fault...

Susurrus:
Yeah, looks like an intermittent failure as I haven't seen that before. Let's run it again.

bors r+

notriddle:
BuildBot isn't running any of the bors jobs.

notriddle:
Okay, according to the [console view](https://alan.ci/buildbot/#/console), bors got the message about the staging pushes, but no build was scheduled as a result of it. Personally, I'd turn of buildbot entirely for the master branch but always build staging (not sure how this relates to building pull requests, though).

asomers:
Let's try again

bors r+

asomers:
That did the trick.  Turns out there's a difference between "staging" and "refs/head/staging".

PR #621, opened 2017-06-18
break the build to test buildbot

asomers:


Susurrus:
So what did the problem end up being?

asomers:
I had a ChangeFilter that should've triggered a build whenever a Change had the "pull" category.  The Changes create by the GitHub change hook did have that category set.  But the ChangeFilter never matched and I don't know why.  I wrote the buildbot mailing list for help.  In the meantime, I changed the ChangeFilter to match the branchname instead, which always looks like "refs/pull/621/merge".  I didn't notice this problem during my testing, because my other ChangeFilter, intended for push events, matched Changes with branch names that I didn't intend.  So every time I created a pull request in the repository that I use for testing, I would get a build.  But that only happened because the source and target branches lay in the same repository.  I fixed that too.

PR #622, opened 2017-06-19
Separate OpenBSD and FreeBSD in fcntl.rs

equal-l2:
Those two OSes cannot be together since the following FreeBSD flags aren't available on OpenBSD.
- `O_DIRECT`
- `O_EXEC`
- `O_TTY_INIT`

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -288,7 +288,7 @@ mod consts {
     );
 }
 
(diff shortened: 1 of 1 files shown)


asomers:
LGTM.  Thanks for the fix!

bors r+

notriddle:
Going to re-r+ this one, too?

asomers:
Yeah.

bors r+

PR #623, opened 2017-06-20
Fix sendmsg on macOS when passing a zero entry cmsgs array.

kinetiknz:
On macOS, trying to call `sendmsg` with a zero entry cmsgs buffer, e.g.:

    sendmsg(fd, [IoVec::from_slice(buf)], &[], MsgFlags::empty(), None)

...fails with `EINVAL`.  This occurs due to the pointer value for zero capacity `Vec`s being 0x1 rather than 0x0, to distinguish allocated-but-zero-size from nullptr.  The [kernel validates](https://github.com/opensource-apple/xnu/blob/dc0628e187c3148723505cf1f1d35bb948d3195b/bsd/kern/uipc_syscalls.c#L1304) both the `msghdr.msg_control` and `msghdr.msg_controllen` fields and rejects `msghdr.msg_control == 0x1` as invalid.

This doesn't show up on Linux because the [kernel validates](https://github.com/torvalds/linux/blob/9705596d08ac87c18aee32cc97f2783b7d14624e/net/core/scm.c#L139) `msghdr.msg_controllen` first and ignores the value of `msghdr.msg_control` if the length was 0.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -258,14 +258,12 @@ pub fn sendmsg<'a>(fd: RawFd, iov: &[IoVec<&'a [u8]>], cmsgs: &[ControlMessage<'
         len += cmsg.len();
         capacity += cmsg.space();
     }
-    // Alignment hackery. Note that capacity is guaranteed to be a
-    // multiple of size_t. Note also that the resulting vector claims
-    // to have length == capacity, so it's presently uninitialized.
+    // Note that the resulting vector claims to have length == capacity,
+    // so it's presently uninitialized.
     let mut cmsg_buffer = unsafe {
         let mut vec = Vec::<u8>::with_capacity(len);
-        let ptr = vec.as_mut_ptr();
-        mem::forget(vec);
-        Vec::<u8>::from_raw_parts(ptr as *mut _, len, len)
+        vec.set_len(len);
+        vec
     };
     {
         let mut ptr = &mut cmsg_buffer[..];
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -277,12 +277,18 @@ pub fn sendmsg<'a>(fd: RawFd, iov: &[IoVec<&'a [u8]>], cmsgs: &[ControlMessage<'
         None => (0 as *const _, 0),
     };
 
+    let cmsg_ptr = if capacity > 0 {
+        cmsg_buffer.as_ptr() as *const c_void
+    } else {
+        ptr::null()
+    };
+
     let mhdr = msghdr {
         msg_name: name as *const c_void,
         msg_namelen: namelen,
         msg_iov: iov.as_ptr(),
         msg_iovlen: iov.len() as size_t,
-        msg_control: cmsg_buffer.as_ptr() as *const c_void,
+        msg_control: cmsg_ptr,
         msg_controllen: capacity as size_t,
         msg_flags: 0,
     };
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -131,7 +131,7 @@ pub fn test_scm_rights() {
                 panic!("unexpected cmsg");
             }
         }
-        assert_eq!(msg.flags & (MSG_TRUNC | MSG_CTRUNC), MsgFlags::empty());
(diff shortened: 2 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -279,12 +277,18 @@ pub fn sendmsg<'a>(fd: RawFd, iov: &[IoVec<&'a [u8]>], cmsgs: &[ControlMessage<'
         None => (0 as *const _, 0),
     };
 
+    let cmsg_ptr = if capacity > 0 {
```
Susurrus: Is this not supposed to be `len`?

kinetiknz: I think, if anything, cmsg_buffer should be allocated with `capacity` rather than `len`.

Susurrus: I'm not certain on this API, so I just wanted to confirm that len was correct.

Susurrus:
Since you have a simple test case, can you add it as a test to the test suite? That way this will never regress after this fix.

Susurrus:
Also we need an entry in `CHANGELOG.md` under the `Fixed` category for this.

kinetiknz:
Thanks for the feedback.  I've pushed an updated version.

> Since you have a simple test case, can you add it as a test to the test suite? That way this will never regress after this fix.

Done.



> Also we need an entry in CHANGELOG.md under the Fixed category for this.

Done.

Susurrus:
This looks pretty good to me, but right now we're blocking on a fix for #634.

Susurrus:
@kinetiknz I actually wanted to request that you add a summary of what's going on in your test, either inline or at the top. I don't understand this subsystem, so being able to roughly understand it without having to dig through the low-level code would be appreciated.

kinetiknz:
> @kinetiknz I actually wanted to request that you add a summary of what's going on in your test, either inline or at the top. I don't understand this subsystem, so being able to roughly understand it without having to dig through the low-level code would be appreciated.

I added a bit of a description to the top of the test.  Let me know if that makes sense.

I've found another couple of issues in this area of the code while updating this PR with your feedback.  I'll send another PR for those soon.

Susurrus:
Can you rebase this? We've fixed the mount tests and enabled some more platforms, but I think this is RTM otherwise.

kinetiknz:
Thanks for the update.  Rebased patch pushed.

test/sys/test_socket.rs
```diff
@@ -145,6 +145,43 @@ pub fn test_scm_rights() {
...
+    {
+        let mut buf = [0u8; 5];
+        let iov = [IoVec::from_mut_slice(&mut buf[..])];
+        let mut cmsgspace: CmsgSpace<[RawFd; 1]> = CmsgSpace::new();
+        let msg = recvmsg(fd2, &iov, Some(&mut cmsgspace), MsgFlags::empty()).unwrap();
+
+        for _ in msg.cmsgs() {
+            panic!("unexpected cmsg");
+        }
+        assert_eq!(msg.flags & (MSG_TRUNC | MSG_CTRUNC), MsgFlags::empty());
```
Susurrus: This would be more clear if it was written as:
```
assert!(!msg.flags.intersects(MSG_TRUNC | MSG_CTRUNC));

kinetiknz: Fixed there and in `test_scm_rights`, where it was copied from.

Susurrus:
This LGTM, @asomers wanna take a look over this as well and merge it if so?

asomers:
@kinetiknz please rebase your PR.

kinetiknz:
> @kinetiknz please rebase your PR.

Done.  I assume I don't have access to get bors to retry the merge.

Susurrus:
I don't think you could even if you specified me as a reviewer. Don't worry, though, I pay close attention to our massive backlog of PRs!

bors r+

Susurrus:
There's just a big backlog of PRs right now which is why bors timed out. Given that builds currently take almost two hours, I'm tempted to suggest we should double our timeout for bors again. @asomers any opinion on this?

bors r+

asomers:
@Susurrus I think that's reasonable.  I'll open a new issue for it.

PR #624, opened 2017-06-20
Enable ptrace on all Linux platforms

geofft:
Nothing that nix currently binds is architecture-specific, and Android
supports ptrace just as much as non-Android Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   `Android` ([#631](https://github.com/nix-rust/nix/pull/631)).
 - `bind` and `errno_location` now work correctly on `Android`
   ([#631](https://github.com/nix-rust/nix/pull/631))
+- Added `nix::ptrace` on all Linux-kernel-based platforms
+  [#624](https://github.com/nix-rust/nix/pull/624). Previously it was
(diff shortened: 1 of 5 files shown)


geofft:
I noticed this in the course of adding a test for #566.... but also in the course of adding a test for #566 I learned that qemu-user doesn't emulate ptrace on anything other than x86 or x86-64. So getting test coverage for this change is going to be hard.

Susurrus:
As for tests, couldn't you have a test that just makes any call and even if an `Err` is returned, it passes? This way we could at least confirm that ptrace is available on the correct platforms.

Susurrus:
Additionally we can run actual tests only on non-qemu targets.

Susurrus:
@geofft This needs a CHANGELOG entry and a rebase and it should be good to merge.

geofft:
@Susurrus rebased, added a changelog entry, and added a test that ptracing yourself fails with either EPERM (expected) or ENOSYS (qemu-user). Lemme know if you need anything else and sorry for being slow to do this!

Susurrus:
LGTM. Eventually I'd like to have a few `ptrace` tests that actuall do stuff if possible, but we'll defer on that for now. This test you added is at least good enough for this PR. Thanks for adding it!

bors r+

PR #625, opened 2017-06-20
Use libc_bitflags! for BSD in fcntl.rs

equal-l2:


Susurrus:
There's a build failure for NetBSD I don't quite understand. Hopefully you can figure it out.

equal-l2:
I have no idea why `libc` is not viable on NetBSD.

Susurrus:
You can easily test this locally by doing `rustup target install x86_64-unknown-netbsd` and then `cargo build --target=x86_64-unknown-netbsd` since it's just a compilation error. Looks like a module import is missing or something on the netbsd target. Let me know if you get stuck!

Susurrus:
@equal-l2 Are you still able to drive this forward? Would love to get this into the next 0.9 release.

src/fcntl.rs
```diff
@@ -404,33 +404,33 @@ mod consts {
 mod consts {
     use libc::c_int;
```
Susurrus: This needs to `use libc::self` here as well, same as for NetBSD. We just don't test on DragonFly. If you want to install the DragonFly target to test this, that'd be appreciated.

Susurrus:
I'd like to include a change to the `libc_bitflags` to use the same syntax as the new `bitflags` macro. This way we don't need to change `struct` to `flags` and `;` to `,` for every entry. Would you be willing to expand this PR to include a change to that macro here as well? 

Additionally, each of these `const` modules can be combined into one and you can use `#cfg[]` attributes to specify which platforms have which constants. This way we only have each variable declared once instead of multiple times for the various platforms. Should also reduce the LoC a decent amount.

equal-l2:
>I'd like to include a change to the libc_bitflags to use the same syntax as the new bitflags macro. This way we don't need to change struct to flags and ; to , for every entry. Would you be willing to expand this PR to include a change to that macro here as well?

That's too hard for me at this moment, sorry.

src/fcntl.rs
```diff
@@ -255,182 +255,53 @@ mod consts {
...
+            O_NOFOLLOW,
+            O_CREAT,
+            O_TRUNC,
+            O_EXCL,
+            O_NOCTTY,
+            O_DIRECTORY,
+            O_CLOEXEC,
+            O_FSYNC,
+            O_NDELAY,
+            #[cfg(any(target_os = "netbsd", target_os = "openbsd", target_os = "macos", target_os = "ios"))]
```
Susurrus: Please wrap lines so they're under 100 characters.

src/fcntl.rs
```diff
@@ -255,182 +255,53 @@ mod consts {
...
+            O_ALT_IO,
+            #[cfg(target_os = "netbsd")]
+            O_NOSIGPIPE,
+            #[cfg(target_os = "netbsd")]
+            O_SEARCH,
         }
     );
 
-    bitflags!(
+    libc_bitflags!(
```
Susurrus: You already have `FdFlag` declared at the top, so one of these declarations should be removed.

equal-l2: `FdFlag` exists line 242 for linux & android, and line 303 for BSDs.
It needs to merge linux's `consts` with BSDs one to unify `FdFlags`.

Susurrus: Yeah, it'd be nice to merge all of these `consts` modules together into one with `cfg` directives for every constant that needs it. It should be a small amount of work given that you've already merged the BSDs.

equal-l2: Some of linux flags are not in `libc` yet, so I can't use `libc_bitflags!` for them now.
This makes it impossible to merge all `consts` into one.

Susurrus: Yep, that makes sense. We'll defer that until upstream includes those constants.

PR #630, opened 2017-06-25
Add wrappers for sysconf(3), pathconf(2), and fpathconf(2)

asomers:


Susurrus:
I'd prefer to get the missing constants added to `libc` before merging this since there's a small and clear list of missing constants. Would you be able to do that?

asomers:
Already working on it.

Susurrus:
Awesome, 👍

Susurrus:
And otherwise this PR seems fine to me, though I'd ask that you squash these commits into one.

asomers:
Upstream PR at https://github.com/rust-lang/libc/pull/628

asomers:
All the Travis builds failed with an error like "error: `"rustup" "target" "list"` failed with exit code: Some(1)".  Lame.  I'll try forcing a rebuild in the morning.

src/unistd.rs
```diff
@@ -819,6 +821,431 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+/// # Returns
+///
+/// - `Ok(Some(x))`: the variable's limit (for limit variables) or its
+///     implementation level (for option variables).  Implementation levels are
+///     usually a decimal-coded date, such as 200112 for POSIX 2001.12
+/// - `Ok(None)`: the variable has no limit (for limit variables) or is
+///     unsupported (for option variables)
+/// - `Err(x)`: an error occurred
+///
+/// http://pubs.opengroup.org/onlinepubs/9699919799/functions/pathconf.html
```
Susurrus: It'd be good to give this a proper title for when docs are generated.

src/unistd.rs
```diff
@@ -819,6 +821,431 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+        Ok(Some(raw))
+    }
+}
+
+/// Get path-dependent configurable system variables
+///
+/// Returns the value of a path-dependent configurable system variable.  Most
+/// supported variables also have associated compile-time constants, but POSIX
+/// allows their values to change at runtime.  There are generally two types of
+/// pathconf variables: options and limits.  See pathconf(3) for more details.
```
Susurrus: Backticks around `pathconf(3)`.

src/unistd.rs
```diff
@@ -819,6 +821,431 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+/// # Returns
+///
+/// - Ok(Some(x)): the variable's limit (for limit variables) or its
+///     implementation level (for option variables).  Implementation levels are
+///     usually a decimal-coded date, such as 200112 for POSIX 2001.12
+/// - Ok(None): the variable has no limit (for limit variables) or is
+///     unsupported (for option variables)
+/// - Err(x): an error occurred
+///
+/// http://pubs.opengroup.org/onlinepubs/9699919799/functions/sysconf.html
```
Susurrus: Same as above.

src/unistd.rs
```diff
@@ -819,6 +821,431 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+pub fn sysconf(var: SysconfVar) -> Result<Option<c_long>> {
+    let raw = unsafe {
+        Errno::clear();
+        libc::sysconf(var as c_int)
+    };
+    if raw == -1 {
```
Susurrus: Can you use `Errno::result()` here to clean up this code a bit?

test/test_unistd.rs
```diff
@@ -217,3 +217,32 @@ execve_test_factory!(test_execve, execve, b"/bin/sh", b"/system/bin/sh");
...
+
+#[test]
+fn test_sysconf_limited() {
+    let open_max = sysconf(SysconfVar::OPEN_MAX);
+    assert!(open_max.expect("sysconf failed").expect("OPEN_MAX is unlimited") > 0);
+}
+
+// Test sysconf with an unsupported varible
+// Note: If future versions of FreeBSD add support for this option then the test
+// must be modified.
```
Susurrus: "should" not "must"

src/unistd.rs
```diff
@@ -819,6 +821,431 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
     Ok((fd, PathBuf::from(pathname)))
 }
 
+/// Variable names for `pathconf`
+// Note: POSIX 1003.1-2008 requires all of these symbols to be
```
Susurrus: Why not have this note be part of the doc comment? If you put a space between it and the first line, it might be a nice detail to have in the docs.

src/unistd.rs
```diff
@@ -819,6 +821,431 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+        } else {
+            Err(Error::Sys(Errno::last()))
+        }
+    } else {
+        Ok(Some(raw))
+    }
+}
+
+/// Variable names for `sysconf`
+// Note: All of these symbols are standardized by POSIX 1003.1-2008, but haven't
```
Susurrus: Same as above.

src/unistd.rs
asomers: No, because `sysconf` will return -1 with `errno` set to 0 to indicate that the requested variable is not supported.  Our `Errno::result` class can't handle that.

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+    _POSIX_PRIO_IO = libc::_PC_PRIO_IO,
+    #[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd",
+              target_os = "linux", target_os = "netbsd", target_os = "openbsd"))]
+    _POSIX_SYNC_IO = libc::_PC_SYNC_IO,
+    #[cfg(any(target_os = "dragonfly", target_os = "openbsd"))]
+    _POSIX_TIMESTAMP_RESOLUTION = libc::_PC_TIMESTAMP_RESOLUTION
+}
+
+/// Like `pathconf`, but for file descriptors instead of paths
+/// # Parameters
```
Susurrus: Needs a blank line above.

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+///
+/// - `fd`: Lookup the value of `var` for this file descriptor
```
Susurrus: This comment is more of a comment about the entire function then this specific argument.

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+}
+
+/// Variable names for `sysconf`
+///
+/// Note: All of these symbols are standardized by POSIX 1003.1-2008, but haven't
+/// been implemented by all platforms.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[repr(i32)]
+pub enum SysconfVar {
+AIO_LISTIO_MAX = libc::_SC_AIO_LISTIO_MAX,
```
Susurrus: These should all be indented.

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+_POSIX2_PBS_MESSAGE = libc::_SC_2_PBS_MESSAGE,
+#[cfg(any(target_os="dragonfly", target_os="freebsd", target_os = "ios",
+          target_os="linux", target_os = "macos", target_os="netbsd",
+          target_os="openbsd"))]
+_POSIX2_PBS_TRACK = libc::_SC_2_PBS_TRACK,
+_POSIX2_SW_DEV = libc::_SC_2_SW_DEV,
+_POSIX2_UPE = libc::_SC_2_UPE,
+_POSIX2_VERSION = libc::_SC_2_VERSION,
+PAGE_SIZE = libc::_SC_PAGE_SIZE,
+// POSIX requires an alias PAGESIZE == PAGE_SIZE, but Rust doesn't like
```
Susurrus: It'd be nice if this were able to be documented in the docs for this enum value. Maybe add a doc comment for `PAGE_SIZE` that says this as well since that way the user could search `nix` docs and find this instead of having to search the source code.

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+/// # Returns
+///
+/// - Ok(Some(x)): the variable's limit (for limit variables) or its
+///     implementation level (for option variables).  Implementation levels are
+///     usually a decimal-coded date, such as 200112 for POSIX 2001.12
+/// - Ok(None): the variable has no limit (for limit variables) or is
+///     unsupported (for option variables)
+/// - Err(x): an error occurred
+///
+/// # Referencces
```
Susurrus: Misspelled.

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+/// # Returns
+///
+/// - `Ok(Some(x))`: the variable's limit (for limit variables) or its
+///     implementation level (for option variables).  Implementation levels are
+///     usually a decimal-coded date, such as 200112 for POSIX 2001.12
+/// - `Ok(None)`: the variable has no limit (for limit variables) or is
+///     unsupported (for option variables)
+/// - `Err(x)`: an error occurred
+///
+/// # Referencces
```
Susurrus: Misspelled

src/unistd.rs
```diff
@@ -819,6 +821,441 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
```
Susurrus: Misspelled

Susurrus:
And please squash this into 1 commit in the next round.

src/unistd.rs
```diff
@@ -819,6 +821,619 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
 }
 
+/// Variable names for `pathconf`
+///
+/// Note: POSIX 1003.1-2008 standardizes all of these variables, but some OSes
+/// choose not to implement variables that cannot change at runtime.
+///
+/// # References
+///
+/// http://pubs.opengroup.org/onlinepubs/9699919799/functions/pathconf.html
```
Susurrus: Please make these bulleted, as they just show up on a single line when docs are generated. Also if you could make them links, use the title of the manpage as the title for them or providing a summary would be great.

src/unistd.rs
```diff
@@ -819,6 +821,619 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+    /// which the implementation conforms.
+    _XOPEN_VERSION = libc::_SC_XOPEN_VERSION,
+}
+
+/// Get configurable system variables
+///
+/// Returns the value of a configurable system variable.  Most supported
+/// variables also have associated compile-time constants, but POSIX
+/// allows their values to change at runtime.  There are generally two types of
+/// sysconf variables: options and limits.  See sysconf(3) for more details.
```
Susurrus: Can you turn "sysconf(3)" into a link to the man pages. You could then remove the "References" section at the end.

src/unistd.rs
```diff
@@ -819,6 +821,619 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+/// - Ok(Some(x)): the variable's limit (for limit variables) or its
+///     implementation level (for option variables).  Implementation levels are
+///     usually a decimal-coded date, such as 200112 for POSIX 2001.12
+/// - Ok(None): the variable has no limit (for limit variables) or is
+///     unsupported (for option variables)
+/// - Err(x): an error occurred
+///
+/// # References
+///
+/// http://pubs.opengroup.org/onlinepubs/9699919799/functions/sysconf.html
```
Susurrus: This doesn't turn into a path when you have it like this. Can you linkify it?

Susurrus:
Just some doc comment improvements. I found these by running `cargo doc --no-deps --open` and looking at the generated docs.

src/unistd.rs
```diff
@@ -819,6 +821,610 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+    }
+}
+
+/// Get path-dependent configurable system variables (see
+/// [pathconf(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pathconf.html))
+///
+/// Returns the value of a path-dependent configurable system variable.  Most
+/// supported variables also have associated compile-time constants, but POSIX
+/// allows their values to change at runtime.  There are generally two types of
+/// `pathconf` variables: options and limits.  See pathconf(2) for more details.
```
Susurrus: Can you turn `pathconf(2)` into a link and also put it between backticks?

src/unistd.rs
```diff
@@ -819,6 +821,610 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+    _POSIX2_PBS_MESSAGE = libc::_SC_2_PBS_MESSAGE,
+    #[cfg(any(target_os="dragonfly", target_os="freebsd", target_os = "ios",
+              target_os="linux", target_os = "macos", target_os="netbsd",
+              target_os="openbsd"))]
+    /// The implementation supports the Track Batch Job Request option.
+    _POSIX2_PBS_TRACK = libc::_SC_2_PBS_TRACK,
+    /// The implementation supports the Software Development Utilities option.
+    _POSIX2_SW_DEV = libc::_SC_2_SW_DEV,
+    /// The implementation supports the User Portability Utilities option.
+    _POSIX2_UPE = libc::_SC_2_UPE,
```
Susurrus: This shouldn't be defined here, but in `libc`. Can you upstream these into `libc` instead of defining them here? I'd like to do this for all of the constants that are missing in `libc` and then the `libc_enum!` macro can be used for defining all of these.

src/unistd.rs
```diff
@@ -819,6 +821,610 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+    /// Integer value indicating version of the Shell and Utilities volume of
+    /// POSIX.1 to which the implementation conforms.
+    _POSIX2_VERSION = libc::_SC_2_VERSION,
+    /// The size of a system page in bytes.
+    ///
+    /// POSIX also defines an alias named `PAGESIZE`, but Rust does not allow two
+    /// enum constants to have the same value, so nix omits `PAGESIZE`.
+    PAGE_SIZE = libc::_SC_PAGE_SIZE,
```
Susurrus: Why isn't this named the same thing?

src/unistd.rs
```diff
@@ -819,6 +821,610 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+    #[cfg(any(target_os="dragonfly", target_os="freebsd", target_os = "ios",
+              target_os="linux", target_os = "macos", target_os="openbsd"))]
+    /// The implementation supports the Thread Sporadic Server option.
+    _POSIX_THREAD_SPORADIC_SERVER = libc::_SC_THREAD_SPORADIC_SERVER,
+    /// The implementation supports threads.
+    _POSIX_THREADS = libc::_SC_THREADS,
+    #[cfg(any(target_os="dragonfly", target_os="freebsd", target_os = "ios",
+              target_os="linux", target_os = "macos", target_os="openbsd"))]
+    /// The implementation supports timeouts.
+    _POSIX_TIMEOUTS = libc::_SC_TIMEOUTS,
```
Susurrus: These should also be defined in `libc` and not here.

Susurrus:
This time I went through some of the enum variants and a few are missing constants in `libc`. I'd like us to stop declaring constants here and only do them in libc. I know there was just a release of `libc` and you wanted to get this into the next version of `nix`, so we either wait until a release contains those constants or we push this from the current release.

src/unistd.rs
asomers: libc contains all the definitions that it needs.  But the situation is fairly confusing.  Let me explain:

POSIX says that some "variables" can have values that are either fixed at compile time or changeable at runtime.  The variable is an abstract thing; not a C symbol.  But there are two C symbols associated with it: one to look it up at build time and one to look it up at runtime via `sysconf`.  In this case, the variable's abstract name is "_POSIX_TIMEOUTS", it's build time symbols is `_POSIX_TIMEOUTS`, and its runtime symbol is `_SC_TIMEOUTS` (the "SC" stands for "sysconf").  See http://pubs.opengroup.org/onlinepubs/9699919799/functions/sysconf.html .  The table's left column has the abstract variable names, and the right column has the sysconf names.  I used the abstract names for nix, reasoning that there's no need to namespace them with "_SC_" since Rust has real namespaces.

src/unistd.rs
asomers: In this case, I don't think I can use `libc_enum!` because the abstract variable names differ from the sysconf symbol names.  Unless, that is, I switch naming conventions and simply use the _SC_ forms for nix's public API.

src/unistd.rs
asomers: Because "PAGE_SIZE" is the abstract variable, and `_SC_PAGE_SIZE` is the `sysconf` variable name.

src/unistd.rs
Susurrus: Interesting, thanks for explaining it. I would like to see this explanation included in the docs for both of the enums so that it's clear to the users what's going on. I don't like these decisions to be completely lost in GitHub PRs.

Susurrus: Once you've done that, go ahead and r+ susurrus for me, as I think this looks good.

asomers:
Susurrus stated his conditional approval on me adding a few more comments for `SysconfVar` and `PathconfVar`, so here it is:

bors r+ susurrus

PR #631, opened 2017-06-27
Allow nix to compile on android

### LINKED ISSUE #313: arm-linux-androideabi build fails (sched, unistd, signal)
With the new test infrastructure (still being stabilized):

```
$ DOCKER_IMAGE=rust-cross-android RUST_TARGETS=arm-linux-androideabi RUST_VERSIONS=1.7.0 ./ci/run-all.sh
=======================================================
TESTING VERSION: 1.7.0, TARGET: arm-linux-androideabi
=======================================================
...
     Running `rustc /root/.multirust/toolchains/1.7.0/cargo/registry/src/github.com-88ac128001ac3a9a/rand-0.3.14/src/lib.rs --crate-name rand --crate-type lib -g -C metadata=6fab17fc670a2bc8 -C extra-filename=-6fab17fc670a2bc8 --out-dir /build/arm-linux-androideabi/debug/deps --emit=dep-info,link --target arm-linux-androideabi -C linker=arm-linux-androideabi-gcc -L dependency=/build/arm-linux-androideabi/debug/deps -L dependency=/build/arm-linux-androideabi/debug/deps --extern libc=/build/arm-linux-androideabi/debug/deps/liblibc-c1044b0a546bbfd6.rlib --cap-lints allow`
/source/src/sched.rs:29:38: 29:56 error: unresolved name `libc::CLONE_NEWUTS` [E0425]
/source/src/sched.rs:29         const CLONE_NEWUTS         = libc::CLONE_NEWUTS as c_int,
                                                             ^~~~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:8:1: 36:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:29:38: 29:56 help: run `rustc --explain E0425` to see a detailed explanation
/source/src/sched.rs:30:38: 30:56 error: unresolved name `libc::CLONE_NEWIPC` [E0425]
/source/src/sched.rs:30         const CLONE_NEWIPC         = libc::CLONE_NEWIPC as c_int,
                                                             ^~~~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:8:1: 36:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:30:38: 30:56 help: run `rustc --explain E0425` to see a detailed explanation
/source/src/sched.rs:31:38: 31:57 error: unresolved name `libc::CLONE_NEWUSER` [E0425]
/source/src/sched.rs:31         const CLONE_NEWUSER        = libc::CLONE_NEWUSER as c_int,
                                                             ^~~~~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:8:1: 36:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:31:38: 31:57 help: run `rustc --explain E0425` to see a detailed explanation
/source/src/sched.rs:32:38: 32:56 error: unresolved name `libc::CLONE_NEWPID` [E0425]
/source/src/sched.rs:32         const CLONE_NEWPID         = libc::CLONE_NEWPID as c_int,
                                                             ^~~~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:8:1: 36:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:32:38: 32:56 help: run `rustc --explain E0425` to see a detailed explanation
/source/src/sched.rs:33:38: 33:56 error: unresolved name `libc::CLONE_NEWNET` [E0425]
/source/src/sched.rs:33         const CLONE_NEWNET         = libc::CLONE_NEWNET as c_int,
                                                             ^~~~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:8:1: 36:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:33:38: 33:56 help: run `rustc --explain E0425` to see a detailed explanation
/source/src/sched.rs:34:38: 34:52 error: unresolved name `libc::CLONE_IO` [E0425]
/source/src/sched.rs:34         const CLONE_IO             = libc::CLONE_IO as c_int,
                                                             ^~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:8:1: 36:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sched.rs:34:38: 34:52 help: run `rustc --explain E0425` to see a detailed explanation
/source/src/unistd.rs:180:24: 180:41 error: unresolved name `libc::sethostname` [E0425]
/source/src/unistd.rs:180     let res = unsafe { libc::sethostname(ptr, len) };
                                                 ^~~~~~~~~~~~~~~~~
/source/src/unistd.rs:180:24: 180:41 help: run `rustc --explain E0425` to see a detailed explanation
     Running `/build/debug/build/nix-test-2ca26930e91b2627/build-script-build`
/source/src/sys/signal.rs:50:30: 50:48 error: mismatched types:
 expected `i32`,
    found `u32` [E0308]
/source/src/sys/signal.rs:50         const SA_NOCLDWAIT = libc::SA_NOCLDWAIT,
                                                          ^~~~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sys/signal.rs:47:1: 57:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sys/signal.rs:50:30: 50:48 help: run `rustc --explain E0308` to see a detailed explanation
/source/src/sys/signal.rs:52:30: 52:46 error: mismatched types:
 expected `i32`,
    found `u32` [E0308]
/source/src/sys/signal.rs:52         const SA_ONSTACK   = libc::SA_ONSTACK,
                                                          ^~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sys/signal.rs:47:1: 57:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sys/signal.rs:52:30: 52:46 help: run `rustc --explain E0308` to see a detailed explanation
/source/src/sys/signal.rs:55:30: 55:46 error: mismatched types:
 expected `i32`,
    found `u32` [E0308]
/source/src/sys/signal.rs:55         const SA_SIGINFO   = libc::SA_SIGINFO,
                                                          ^~~~~~~~~~~~~~~~
<bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sys/signal.rs:47:1: 57:2 note: in this expansion of bitflags! (defined in <bitflags macros>)
/source/src/sys/signal.rs:55:30: 55:46 help: run `rustc --explain E0308` to see a detailed explanation
error: aborting due to 10 previous errors
```

roblabla:
Fixes #313

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -198,6 +198,7 @@ pub const SIGIOT : Signal = SIGABRT;
 pub const SIGPOLL : Signal = SIGIO;
 pub const SIGUNUSED : Signal = SIGSYS;
 
+#[cfg(not(target_os = "android"))]
 libc_bitflags!{
     pub flags SaFlags: libc::c_int {
         SA_NOCLDSTOP,
@@ -207,6 +208,35 @@ libc_bitflags!{
         SA_RESETHAND,
         SA_RESTART,
         SA_SIGINFO,
+    }
+}
+
+// On 64-bit android, sa_flags is c_uint while on 32-bit android, it is
+// c_ulong.
+// FIXME: https://github.com/rust-lang/libc/pull/511
+#[cfg(all(target_os = "android", target_pointer_width = "32"))]
+libc_bitflags!{
+    pub flags SaFlags: libc::c_ulong {
+        SA_NOCLDSTOP as libc::c_ulong,
+        SA_NOCLDWAIT as libc::c_ulong,
+        SA_NODEFER as libc::c_ulong,
+        SA_ONSTACK as libc::c_ulong,
+        SA_RESETHAND as libc::c_ulong,
+        SA_RESTART as libc::c_ulong,
+        SA_SIGINFO as libc::c_ulong,
+    }
+}
+
+#[cfg(all(target_os = "android", target_pointer_width = "64"))]
+libc_bitflags!{
+    pub flags SaFlags: libc::c_uint {
+        SA_NOCLDSTOP as libc::c_uint,
+        SA_NOCLDWAIT as libc::c_uint,
+        SA_NODEFER as libc::c_uint,
+        SA_ONSTACK as libc::c_uint,
+        SA_RESETHAND as libc::c_uint,
+        SA_RESTART as libc::c_uint,
+        SA_SIGINFO as libc::c_uint,
     }
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,9 @@ matrix:
     - env: TARGET=armv7-linux-androideabi DISABLE_TESTS=1
       rust: 1.13.0
     - env: TARGET=i686-linux-android DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.18.0
     - env: TARGET=x86_64-linux-android DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.18.0
 
     # iOS
     - env: TARGET=aarch64-apple-ios DISABLE_TESTS=1
@@ -104,18 +104,6 @@ matrix:
       rust: nightly
 
   allow_failures:
-    # Android (in the process of fixing these, so they're allowed to fail for now)
-    - env: TARGET=aarch64-linux-android DISABLE_TESTS=1
-      rust: 1.13.0
-    - env: TARGET=arm-linux-androideabi DISABLE_TESTS=1
-      rust: 1.13.0
-    - env: TARGET=armv7-linux-androideabi DISABLE_TESTS=1
-      rust: 1.13.0
-    - env: TARGET=i686-linux-android DISABLE_TESTS=1
-      rust: 1.13.0
-    - env: TARGET=x86_64-linux-android DISABLE_TESTS=1
-      rust: 1.13.0
-
     # iOS is still being worked on, so for now don't block on compilation failures
     - env: TARGET=aarch64-apple-ios DISABLE_TESTS=1
       rust: 1.13.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Introduced wrapper types for gid_t, pid_t, and uid_t as Gid, Pid, and Uid
   respectively. Various functions have been changed to use these new types as
   arguments. ([#629](https://github.com/nix-rust/nix/pull/629))
+- Promoted all Android targets to Tier 2 support
 
 ### Removed
 - Removed io::Error from nix::Error and conversion from nix::Error to Errno
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   Now compiles on Linux/MIPS ([#538](https://github.com/nix-rust/nix/pull/538)),
   `Linux/PPC` ([#553](https://github.com/nix-rust/nix/pull/553)),
   `MacOS/x86_64,i686` ([#553](https://github.com/nix-rust/nix/pull/553)),
-  `NetBSD/x64_64` ([#538](https://github.com/nix-rust/nix/pull/538)), and
-  `FreeBSD/x86_64,i686` ([#536](https://github.com/nix-rust/nix/pull/536)).
+  `NetBSD/x64_64` ([#538](https://github.com/nix-rust/nix/pull/538)),
+  `FreeBSD/x86_64,i686` ([#536](https://github.com/nix-rust/nix/pull/536)), and
+  `Android` ([#631](https://github.com/nix-rust/nix/pull/631)).
+- `bind` and `errno_location` now work correctly on `Android`
+  ([#631](https://github.com/nix-rust/nix/pull/631))
 
 ## [0.8.1] 2017-04-16
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ limitations. Support for platforms is split into two tiers:
              *do not* block the inclusion of new code. Testing may be run, but
              failures in tests don't block the inclusion of new code.
 
-The following targets are all supported by nix on Rust 1.13.0 or newer:
+The following targets are all supported by nix on Rust 1.13.0 or newer (unless
+otherwise noted):
 
 Tier 1:
   * i686-unknown-linux-gnu
@@ -68,18 +69,18 @@ Tier 1:
 Tier 2:
   * i686-unknown-freebsd
   * x86_64-unknown-netbsd
+  * aarch64-linux-android
+  * arm-linux-androideabi
+  * armv7-linux-androideabi
+  * i686-linux-android (requires Rust >= 1.18)
+  * x86_64-linux-android (requires Rust >= 1.18)
 
 Tier 3:
   * aarch64-apple-ios
-  * aarch64-linux-android
-  * arm-linux-androideabi
   * armv7-apple-ios
-  * armv7-linux-androideabi
   * armv7s-apple-ios
   * i386-apple-ios
-  * i686-linux-android
   * x86_64-apple-ios
-  * x86_64-linux-android
 
 ## Usage
 
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -35,10 +35,16 @@ unsafe fn errno_location() -> *mut c_int {
     __errno()
 }
 
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(target_os = "linux")]
 unsafe fn errno_location() -> *mut c_int {
     extern { fn __errno_location() -> *mut c_int; }
     __errno_location()
+}
+
+#[cfg(target_os = "android")]
+unsafe fn errno_location() -> *mut c_int {
+    extern { fn __errno() -> *mut c_int; }
+    __errno()
 }
 
 /// Sets the platform-specific errno to no-error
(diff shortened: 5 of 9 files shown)


src/sys/signal.rs
```diff
@@ -210,6 +211,19 @@ libc_bitflags!{
     }
 }
 
+#[cfg(target_os = "android")]
```
Susurrus: These need doc comments. Also, if you search the `libc` crate you'll see these are available on Linux and all BSD platforms. We shouldn't artificially restrict them to just `android` here.

roblabla:
So there's another symbol that's apparently missing from android : __errno_location. Instead, it stores it in tread local storage, and has a function to retrieve a pointer to it called __errno : https://github.com/LineageOS/android_bionic/blob/cm-14.1/libc/bionic/__errno.cpp

errno is simply defined as dereferencing the symbol there : https://github.com/LineageOS/android_bionic/blob/cm-14.1/libc/include/errno.h#L44

When we try to compile something for android, we get a linker error saying that __errno_location is undefined.

Now, because the bionic __errno is defined as volatile in bionic, I don't know if it's a good idea to simply make __errno_location return it ? I'm pretty illiterate when it comes to what volatile does.

src/sys/signal.rs
roblabla: They are still available on other platforms, just defined as libc::c_int above ?

Maybe I should use cfg_if! to make it clearer.

roblabla: It's worth noting that this change is necessary because of https://github.com/rust-lang/libc/pull/511

Susurrus: They're defined as `c_int` for all platforms.

Since you mentioned the upstream issue, please also add a `FIXME` that links to the upstream issue and states what can be done when it's resolved. We need to track these tasks in the code, not just here in the PRs.

roblabla:
More problems, this time for the target `aarch64-linux-android`. For some reason, `libc::SO_PASSCRED`, `libc::SO_PEERCRED`, `libc::SO_SNDBUFFORCE` are not defined on android in libc (in addition to being undefined on linux arm). I believe this is a bug, so I'm pushing it upstream. In the meantime, I'll also #cfg them out.

 https://github.com/nix-rust/nix/blob/master/src/sys/socket/consts.rs#L43 is marked as `not(target_arch="arm")`.

roblabla:
It should now be possible to compile binaries using nix for both arm-linux-androideabi and aarch64-linux-android.

Should I document every `#[cfg(not(target_os = "android"))]` ? I'm not too sure what I'm supposed to put in there :/.

Susurrus:
@roblabla There are basically two categories of things here you're asking about, one is things that currently fail when testing on Android that shouldn't and things that shouldn't be tested on Android. I'd prefer to wait until upstream is fixed so that we only have cases of the latter. Those don't need to be documented. If we have things that can't currently be tested because they're blocking on upstream, those need to be documented. But like I said, I'd rather wait until upstream is fixed to merge this PR so that we only have "good" fixes here.

roblabla:
The only change that'll ever be fixed by libc is the use of 'as' in the bitflags though, along with missing symbols.

I'll try digging what symbols are missing in libc compared to the most recent version of android, so we have a clearer picture of the situation.

On June 29, 2017 6:29:12 AM GMT+02:00, Susurrus <notifications@github.com> wrote:
>@roblabla There are basically two categories of things here you're
>asking about, one is things that currently fail when testing on Android
>that shouldn't and things that shouldn't be tested on Android. I'd
>prefer to wait until upstream is fixed so that we only have cases of
>the latter. Those don't need to be documented. If we have things that
>can't currently be tested because they're blocking on upstream, those
>need to be documented. But like I said, I'd rather wait until upstream
>is fixed to merge this PR so that we only have "good" fixes here.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/nix-rust/nix/pull/631#issuecomment-311859521

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Susurrus:
Yes, well `libc` isn't going to add symbols that don't exist on Android. For things that don't exist on Android, `cfg`-ing them out of tests and builds is correct and doesn't need explicit documentation.

roblabla:
Linking https://github.com/rust-lang/libc/issues/632 since it seems to contain some items of interest.

Susurrus:
@roblabla We now have all android targets being built in CI, so if you rebase this you can see how well things will compile on those targets. Probably not worth doing so until rust-lang/libc#632 settles down, however.

ndusart:
@roblabla  rust-lang/libc#632 is fixed, we could remove some of the `#cfg[]` in the first commit now :)

roblabla:
With https://github.com/rust-lang/libc/pull/634 now merged, most of the constants are now defined. The only ones left are : 

- `SO_PASSCRED`
- `SO_PEERCRED`
- `SO_SNDBUFFORCE`

They should be defined though, as a quick grep -R reveals they are in the unified headers, under `asm-generic`. Time for a second libc PR I guess

Susurrus:
@roblabla I added those constants in rust-lang/libc#639 already as I saw they were needed for Android and I was already adding socket constants as part of #636.

Susurrus:
So give this thing a rebase and remove some of those `cfg`s and we'll see what happens!

roblabla:
https://github.com/rust-lang/libc/pull/642 is necessary now

Susurrus:
I'd also like to move Android builds to Tier 2 as part of this. Could you modify the README and also `.travis.yml` to do that as well?

roblabla:
Should be fine now

Susurrus:
Could you better document your changes to the `libc_bitflags` macro? `munch an ident` isn't very helpful to someone who doesn't use the macro system in Rust. It'd be better if you described exactly what it allowed, especially if you added an example use-case. The module comment at the top could use some additions for the two use cases you added as well.

I'd also prefer your changes to `macros.rs` be removed into a separate commit before the others if you're comfortable with `git rebase`, otherwise it's not a big deal.

roblabla:
Is this better ?

roblabla:
Right, so [this change](https://github.com/nix-rust/nix/pull/631/commits/3b031661ccb937ecc7bde10711007d5d128c73ac#diff-cd00aa2541c063d00fa0c8c6f54cdb71L43) caused the build to fail on arm because libc doesn't define those flags on arm. https://github.com/rust-lang/libc/pull/644 will fix this failure.

For the i686 and x86_64 android, there are a *bunch* of type failures. Which is surprising because there isn't much in the arch-specific files of libc (https://github.com/rust-lang/libc/blob/master/src/unix/notbsd/android/b32/arm.rs). I'll investigate.

ndusart:
The failures on Android 64 bits are due because the test case is using rustc 1.13 which has no support at all for x86_64-linux-android and for which std was broken for i686-linux-android (see [fix](https://github.com/rust-lang/rust/commit/a7add43389d435bda8769cb73ac6636237633836)), this fix was introduced in version 1.18 of rust.

So the explanation was that `std::os::raw::c_char` returned from `CStr::as_ptr` is different than `libc::c_char` for older stdlib.

I think x86_64-linux-android is not considered stable yet, libc still use the beta channel for testing it.
We should consider using at least rustc 1.18 for testing i686-linux-android.

I could also advise you to test (at least the build phase) locally on your machine with the latest rustc for the different targets. You won't have to commit just to test and you'd have seen that difference.

roblabla:
Ah I see. So I bumped the minimum version to 1.18 for i686/x86_64. I also updated the libc PR to also define `MAP_32BIT` on x86/x86_64 android targets.

I think this should be everything necessary.

Susurrus:
@roblabla Love the separate macro commit, thanks for that.

This is pretty close to being mergeable:
 * The requirement for Rust 1.18 for the x86 Android targets needs to be documented. I suggest adding it to the README like `(requires Rust >= 1.18)` to the end of the lines stating they have Tier 2 support.
 * We need some CHANGELOG entries. Under `Changed` add that Android is now a Tier 2 target and under `Fixed` `bind` and `errno_location` now work correctly on Android.

Susurrus:
Please add those CHANGELOG entries in with the corresponding commits.

If you're feeling particularly diligent, you could also reorder the 2 last commits. I like to have commits ordered such that building things will always succeed. This means fixing the errors and then requiring builds to pass for the fixed target.

roblabla:
@Susurrus actually, the commits are in the right order. Github interface is just confused, probably because of all the rebase and push --force ? https://github.com/roblabla/nix/commits/feature-androidWorks

roblabla:
Done.

Susurrus:
LGTM, thanks for all your hard work!

bors r+

PR #633, opened 2017-06-28
Skip the mount tests on kernel 4.4.0

### LINKED ISSUE #610: Failed mount tests on x86_64 Ubuntu 16.04.4
```
thread 'main' panicked at 'write failed: Value too large for defined data type (os error 75)', test/test_mount.rs:53
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:495
   7: test_mount::test_mount::test_mount_tmpfs_without_flags_allows_rwx::{{closure}}
             at ./test/test_mount.rs:53
   8: <core::result::Result<T, E>>::unwrap_or_else
             at /checkout/src/libcore/result.rs:706
   9: test_mount::test_mount::test_mount_tmpfs_without_flags_allows_rwx
             at ./test/test_mount.rs:47
  10: test_mount::main
             at ./test/test_mount.rs:217
  11: std::panicking::try::do_call
             at /checkout/src/libstd/panicking.rs:454
  12: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  13: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  14: main
  15: __libc_start_main
  16: _start
test test_mount::test_mount_tmpfs_without_flags_allows_rwx ... error: test failed
```

asomers:
Some versions of that kernel have a known bug with tmpfs in namespaces.
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1659087

Fixes #610

### DIFF
--- a/test/test_mount.rs
+++ b/test/test_mount.rs
@@ -17,6 +17,7 @@ mod test_mount {
 
     use libc::{EACCES, EROFS};
 
+    use nix::errno::Errno;
     use nix::mount::{mount, umount, MsFlags, MS_BIND, MS_RDONLY, MS_NOEXEC};
     use nix::sched::{unshare, CLONE_NEWNS, CLONE_NEWUSER};
     use nix::sys::stat::{self, S_IRWXU, S_IRWXG, S_IRWXO, S_IXUSR, S_IXGRP, S_IXOTH};
@@ -49,6 +50,23 @@ exit 23";
             .write(true)
             .mode((S_IRWXU | S_IRWXG | S_IRWXO).bits())
             .open(&test_path)
+            .or_else(|e|
+                if Errno::from_i32(e.raw_os_error().unwrap()) == Errno::EOVERFLOW {
+                    // Skip tests on certain Linux kernels which have a bug
+                    // regarding tmpfs in namespaces.
+                    // Ubuntu 14.04 and 16.04 are known to be affected; 16.10 is
+                    // not.  There is no legitimate reason for open(2) to return
+                    // EOVERFLOW here.
+                    // https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1659087
+                    let stderr = io::stderr();
+                    let mut handle = stderr.lock();
+                    writeln!(handle, "Buggy Linux kernel detected.  Skipping test.")
+                    .unwrap();
+                    process::exit(0);
+               } else {
+                   panic!("open failed: {}", e);
+               }
+            )
             .and_then(|mut f| f.write(SCRIPT_CONTENTS))
             .unwrap_or_else(|e| panic!("write failed: {}", e));
 


Susurrus:
We've got a bunch of failures because of `rustup` and then a `openpty` failure. You mentioned that `rustup` failure in another PR, I don't know what's going on there. Did you ever look into that?

asomers:
@Susurrus I'm looking at the rustup failure now, but I don't know what's going on.  It looks like it's affected every PR for the last 3 days.  Travis updated their build images 6 days ago, but we've had successful builds since then, so I don't think that's it.  Is there any way to get an interactive session in Travis?

Susurrus:
I don't think Travis has interactive sessions. I'm wondering if you could reproduce in a VM, maybe it's a Rustup bug?

Susurrus:
For the record, this fails for me on 4.11.3-202 on Fedora 25.

asomers:
Perhaps we should just screw the namespaces and require the mount tests to run as root?  We're already using sudo in Travis.

Susurrus:
That's fine. Is there anyway to test for root easily within the tests?

asomers:
Sure.  you can just use getpid() == 0.  But I don't know how to tell Cargo to use sudo for the mount tests.

asomers:
Actually, we're making this way too complicated.  There is absolutely no reason why open(2) should return EOVERFLOW.  If we see that error code, then we know we're running on a buggy kernel.  So we can delete the kernel version check.  All we have to do is skip (or pass, since Cargo doesn't know about skipped tests yet) this test whenever we see EOVERFLOW.

Susurrus:
`open()` can actually return `EOVERFLOW`, but I agree it shouldn't in these situations given the manpages for `open(2)`:

>      EOVERFLOW
>              pathname refers to a regular file that is too large to be opened.  The usual  sce‐
>              nario  here  is  that  an  application  compiled  on  a  32-bit  platform  without
>              -D_FILE_OFFSET_BITS=64 tried to open a file whose size  exceeds  (1<<31)-1  bytes;
>              see  also  O_LARGEFILE  above.  This is the error specified by POSIX.1; in kernels
>              before 2.6.24, Linux gave the error EFBIG for this case.

Susurrus:
Just tested this locally and it works on my end. So I think we might be on to something here. @henninglive Care to give this branch a spin and confirm it fixes things on your end?

asomers:
I fixed #634, so this PR is the last thing standing between us and green builds again.

Susurrus:
One thing I'd point out here is that these tests don't just fail on Linux 4.4, they also fail on my 4.11.3 box, so it's something more than that Ubuntu error you linked. I'd expand the comment to include this information as well.

asomers:
bors failed because test_aio_cancel_all failed for powerpc on linux.  I haven't seen that before, and I don't know what could cause it.  I'm going to try again, because we need this mount fix.

bors r+

PR #635, opened 2017-06-28
Fix cross

### LINKED ISSUE #634: Cross fails with error: `"rustup" "target" "list"` failed with exit code: Some(1)
On about 25-June, all of our Travis builds started to fail with the error message:

> error: `"rustup" "target" "list"` failed with exit code: Some(1)

I don't know why it started on 25-June.  It's probably either a change to Travis (they updated their builders, but they claim it was done by 21-June), or Rustup, or Cross.  We should also report this bug upstream to japaric/cross
https://travis-ci.org/nix-rust/nix/builds/246718360

asomers:
Cross crashes if there isn't a default toolchain installed, even if it's
not the toolchain we intend to use.

Fixes #634

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ dist: trusty
 language: rust
 services: docker
 sudo: required
+group: deprecated-2017Q2
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


asomers:
Issue #634 was fixed by clearing Travis's caches through the Travis control panel.  No code change is required.

PR #638, opened 2017-06-30
Make aio, chdir, and wait tests thread safe

### LINKED ISSUE #251: Investigate flaky tests: test_wait_signal, test_wait, test_execve

asomers:
Fix thread safety issues in aio, chdir, and wait tests
    
They have four problems:
    
* The chdir tests change the process's cwd, which is global.  Protect them all with a mutex.
    
* The wait tests will reap any subprocess, and several tests create subprocesses.  Protect them all with a mutex so only one subprocess-creating test will run at a time.
    
* When a multithreaded test forks, the child process can sometimes block in the stack unwinding code.  It blocks on a mutex that was held by a different thread in the parent, but that thread doesn't exist in the child, so a deadlock results.  Fix this by immediately calling `std::process:;exit` in the child processes.
    
* My previous attempt at thread safety in the aio tests didn't work, because anonymous MutexGuards drop immediately.  Fix this by naming the SIGUSR2_MTX MutexGuards.

Fixes #251

### DIFF
--- a/test/sys/test_wait.rs
+++ b/test/sys/test_wait.rs
@@ -6,6 +6,9 @@ use libc::exit;
 
 #[test]
 fn test_wait_signal() {
+    #[allow(unused_variables)]
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
     match fork() {
       Ok(Child) => pause().unwrap_or(()),
       Ok(Parent { child }) => {
@@ -19,6 +22,9 @@ fn test_wait_signal() {
 
 #[test]
 fn test_wait_exit() {
+    #[allow(unused_variables)]
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
     match fork() {
       Ok(Child) => unsafe { exit(12); },
       Ok(Parent { child }) => {
--- a/test/test_mq.rs
+++ b/test/test_mq.rs
@@ -16,6 +16,8 @@ use nix::Error::Sys;
 
 #[test]
 fn test_mq_send_and_receive() {
+    #[allow(unused_variables)]
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
 
     const MSG_SIZE: c_long =  32;
     let attr =  MqAttr::new(0, 10, MSG_SIZE, 0);
--- a/test/test_mq.rs
+++ b/test/test_mq.rs
@@ -16,45 +16,27 @@ use nix::Error::Sys;
 
 #[test]
 fn test_mq_send_and_receive() {
-    #[allow(unused_variables)]
-    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
-
     const MSG_SIZE: c_long =  32;
     let attr =  MqAttr::new(0, 10, MSG_SIZE, 0);
-    let mq_name_in_parent = &CString::new(b"/a_nix_test_queue".as_ref()).unwrap();
-    let mqd_in_parent = mq_open(mq_name_in_parent, O_CREAT | O_WRONLY, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH, Some(&attr)).unwrap();
-    let msg_to_send = "msg_1".as_bytes();
+    let mq_name= &CString::new(b"/a_nix_test_queue".as_ref()).unwrap();
 
-    mq_send(mqd_in_parent, msg_to_send, 1).unwrap();
+    let mqd0 = mq_open(mq_name, O_CREAT | O_WRONLY,
+                       S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH,
+                       Some(&attr)).unwrap();
+    let msg_to_send = "msg_1";
+    mq_send(mqd0, msg_to_send.as_bytes(), 1).unwrap();
 
-    let (reader, writer) = pipe().unwrap();
+    let mqd1 = mq_open(mq_name, O_CREAT | O_RDONLY,
+                       S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH,
+                       Some(&attr)).unwrap();
+    let mut buf = [0u8; 32];
+    let mut prio = 0u32;
+    let len = mq_receive(mqd1, &mut buf, &mut prio).unwrap();
+    assert!(prio == 1);
 
-    let pid = fork();
-    match pid {
-        Ok(Child) => {
-            let mq_name_in_child =  &CString::new(b"/a_nix_test_queue".as_ref()).unwrap();
-            let mqd_in_child = mq_open(mq_name_in_child, O_CREAT | O_RDONLY, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH, Some(&attr)).unwrap();
-            let mut buf = [0u8; 32];
-            let mut prio = 0u32;
-            mq_receive(mqd_in_child, &mut buf, &mut prio).unwrap();
-            assert!(prio == 1);
-            write(writer, &buf).unwrap();  // pipe result to parent process. Otherwise cargo does not report test failures correctly
-            mq_close(mqd_in_child).unwrap();
-      }
-      Ok(Parent { child }) => {
-          mq_close(mqd_in_parent).unwrap();
-
-          // Wait for the child to exit.
-          waitpid(child, None).unwrap();
-          // Read 1024 bytes.
-          let mut read_buf = [0u8; 32];
-          read(reader, &mut read_buf).unwrap();
-          let message_str = str::from_utf8(&read_buf).unwrap();
-          assert_eq!(&message_str[.. message_str.char_indices().nth(5).unwrap().0], "msg_1");
-      },
-      // panic, fork should never fail unless there is a serious problem with the OS
-      Err(_) => panic!("Error: Fork Failed")
-    }
+    mq_close(mqd1).unwrap();
+    mq_close(mqd0).unwrap();
+    assert_eq!(msg_to_send, str::from_utf8(&buf[0..len]).unwrap());
 }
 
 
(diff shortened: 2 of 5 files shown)


test/test_unistd.rs
```diff
@@ -8,12 +8,26 @@ use std::iter;
...
 use std::fs::File;
 use std::io::{Write, Read};
+use std::ops::Fn;
 use std::os::unix::prelude::*;
 use std::env::current_dir;
 use tempfile::tempfile;
 use tempdir::TempDir;
 use libc::off_t;
 
+fn test_in_subprocess<F: Fn()>(func: F) {
```
Susurrus: This could use a doccomment.

test/test_unistd.rs
```diff
@@ -144,40 +162,41 @@ macro_rules! execve_test_factory(
...
+        assert_eq!(getcwd().unwrap(), current_dir().unwrap());
+
+        // make path 500 chars longer so that buffer doubling in getcwd
+        // kicks in.  Note: One path cannot be longer than 255 bytes
+        // (NAME_MAX) whole path cannot be longer than PATH_MAX (usually
+        // 4096 on linux, 1024 on macos)
+        let mut inner_tmp_dir = tmp_dir.path().to_path_buf();
+        for _ in 0..5 {
+            let newdir = iter::repeat("a").take(100).collect::<String>();
+            //inner_tmp_dir = inner_tmp_dir.join(newdir).path();
```
Susurrus: Probably want to remove this.

Susurrus:
This needs a rebase to the updated CI configuration. Then it LGTM assuming it pasts all tests.

asomers:
Actually, it's not ok.  It seemed ok at first, but now I can occasionally produce deadlocks when I use 8 threads.  This PR requires more investigation.

asomers:
`test_in_subprocess`will never work.  It looks good, but it doesn't work if the process is multithreaded (which it usually is).  The child process is single threaded; only the thread that called `fork` will exist in the child.  So if some other threads in the master held locks that the test function needs, then it will never complete.  And some of the stuff in Rust's standard library is using locks under the hood, which is why I'm seeing deadlocks.  I'll try a different technique to fix these tests.

asomers:
Ok, I switched the `chdir` tests to use a Mutex for protection instead of forking.  And I also fixed some issues with the aio tests and with other tests that fork().  Now, if I disable some unrelated tests like the pty tests, I can run 10000 iterations with 8 threads on both FreeBSD and Linux.

test/test_mq.rs
```diff
@@ -53,6 +54,7 @@ fn test_mq_send_and_receive() {
       // panic, fork should never fail unless there is a serious problem with the OS
       Err(_) => panic!("Error: Fork Failed")
     }
+    drop(m);    //appease the unused_variable checker
```
Susurrus: Needs a space and capitalize "Appease"

test/test_unistd.rs
```diff
@@ -9,16 +9,16 @@ use std::ffi::CString;
...
-use libc::off_t;
+use libc::{_exit, off_t};
 
 #[test]
 fn test_fork_and_waitpid() {
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
     let pid = fork();
     match pid {
-        Ok(Child) => {} // ignore child here
+        Ok(Child) => { unsafe { _exit(0) }; }
```
Susurrus: Can you not use `std::process::exit(0)` here?

test/test_unistd.rs
```diff
@@ -29,23 +29,26 @@ fn test_fork_and_waitpid() {
                 Ok(WaitStatus::Exited(pid_t, _)) =>  assert!(pid_t == child),
 
                 // panic, must never happen
-                Ok(_) => panic!("Child still alive, should never happen"),
+                s @ Ok(_) => panic!("Child exited {:?}, should never happen", s),
```
Susurrus: I'm wondering if `unreachable!()` doesn't provide better semantics here than `panic!()`.

test/test_unistd.rs
```diff
@@ -29,23 +29,26 @@ fn test_fork_and_waitpid() {
...
                 // panic, waitpid should never fail
-                Err(_) => panic!("Error: waitpid Failed")
+                Err(s) => panic!("Error: waitpid returned Err({:?}", s)
             }
 
         },
         // panic, fork should never fail unless there is a serious problem with the OS
         Err(_) => panic!("Error: Fork Failed")
     }
+    drop(m);    // appease the unused_variable checker
```
Susurrus: Capitalize "Appease"

test/test_unistd.rs
```diff
@@ -29,23 +29,26 @@ fn test_fork_and_waitpid() {
...
 }
 
 #[test]
 fn test_wait() {
+    // grab FORK_MTX so wait doesn't reap a different test's child process
```
Susurrus: Capitalize "Grab"

test/test_unistd.rs
```diff
@@ -29,23 +29,26 @@ fn test_fork_and_waitpid() {
...
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
     let pid = fork();
     match pid {
-        Ok(Child) => {} // ignore child here
+        Ok(Child) => { unsafe { _exit(0) }; }
```
Susurrus: See earlier comment

test/test_unistd.rs
```diff
@@ -139,46 +144,47 @@ macro_rules! execve_test_factory(
...
     for _ in 0..5 {
         let newdir = iter::repeat("a").take(100).collect::<String>();
-        //inner_tmp_dir = inner_tmp_dir.join(newdir).path();
         inner_tmp_dir.push(newdir);
         assert!(mkdir(inner_tmp_dir.as_path(), stat::S_IRWXU).is_ok());
     }
     assert!(chdir(inner_tmp_dir.as_path()).is_ok());
-    assert_eq!(getcwd().unwrap(), current_dir().unwrap());
+    assert_eq!(getcwd().unwrap(), inner_tmp_dir.as_path());
+    drop(m);    // appease the unused_variable checker
```
Susurrus: I think you could also do `#[allow(unused_variable)]` on line 170 here, couldn't you? I don't know if that's any nicer, but it's got cleaner semantics.

test/test_unistd.rs
asomers: I don't think so.  The examples for `unreachable!` are for stuff that's so obviously unreachable that a smarter compiler should be able to figure it out.  But in this case, the assertion can actually be hit; for example if you SIGKILL the child.  I actually did hit this assertion frequently when my child deadlocked and had to be killed.

test/test_unistd.rs
asomers: I initially used `_exit(0)` because I didn't know exactly where the deadlock was, and `_exit(0)` skips virtually all teardown.  But I just tried it, and it looks like `std::process::exit(0)` works too.  I guess the deadlock is probably in the test harness code somewhere.  I'll switch.

test/test_unistd.rs
asomers: I didn't know you could apply that to a single variable.  Cool!  BTW, if you didn't figure out, it doesn't work to do "let _ = ..." because the anonymous variable gets dropped right away.

test/sys/test_wait.rs
```diff
@@ -6,6 +6,8 @@ use libc::exit;
 
 #[test]
 fn test_wait_signal() {
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
```
Susurrus: You forgot to change this to `allow(unused_variables)` as well. I would suggest you also add a comment for each of these sections stating why we want to have an unused variable, as someone not familiar with this might think it should be removed without one.

test/test_unistd.rs
```diff
@@ -9,16 +9,18 @@ use std::ffi::CString;
...
+use std::process::exit;
 
 #[test]
 fn test_fork_and_waitpid() {
+    #[allow(unused_variables)]
+    let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
     let pid = fork();
     match pid {
-        Ok(Child) => {} // ignore child here
+        Ok(Child) => { exit(0); }
```
Susurrus: This can just be `exit(0)` I think, without the braces or semicolon, yes?

test/test_unistd.rs
Susurrus: Okay, well if it isn't unreachable, then we definitely shouldn't use that macro!

test/test_unistd.rs
```diff
@@ -145,40 +152,40 @@ macro_rules! execve_test_factory(
...
+    let m = ::CWD_MTX.lock().expect("Mutex got poisoned by another test");
     let tmp_dir = TempDir::new("test_getcwd").unwrap();
     assert!(chdir(tmp_dir.path()).is_ok());
-    assert_eq!(getcwd().unwrap(), current_dir().unwrap());
+    assert_eq!(getcwd().unwrap(), tmp_dir.path());
 
-    // make path 500 chars longer so that buffer doubling in getcwd kicks in.
-    // Note: One path cannot be longer than 255 bytes (NAME_MAX)
-    // whole path cannot be longer than PATH_MAX (usually 4096 on linux, 1024 on macos)
+    // make path 500 chars longer so that buffer doubling in getcwd
```
Susurrus: No double-spaces before "Note" and capitalize the start of sentences. Also, please line wrap to 100 characters

Susurrus:
Also tests are failing on mac. Possibly because trust on mac doesn't use docker?

Susurrus:
I'm assuming this is finished on your end. LGTM. I think some more comments would be helpful, but I lean towards more comments rather than less, and I don't think it should hold up this PR, especially with #681 waiting to reuse the mutex functionality this sets up.

bors r+

PR #639, opened 2017-06-30
Add all Android targets as Tier 3 platforms

Susurrus:
cc @ndusart @roblabla

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,18 @@ matrix:
   # These are all the build jobs. Adjust as necessary. Comment out what you
   # don't need
   include:
+    # Android
+    - env: TARGET=aarch64-linux-android DISABLE_TESTS=1
+      rust: 1.13.0
+    - env: TARGET=arm-linux-androideabi DISABLE_TESTS=1
+      rust: 1.13.0
+    - env: TARGET=armv7-linux-androideabi DISABLE_TESTS=1
+      rust: 1.13.0
+    - env: TARGET=i686-linux-android DISABLE_TESTS=1
+      rust: 1.13.0
+    - env: TARGET=x86_64-linux-android DISABLE_TESTS=1
+      rust: 1.13.0
+
     # Linux
     - env: TARGET=aarch64-unknown-linux-gnu
       rust: 1.13.0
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,6 +42,9 @@ limitations. Support for platforms is split into two tiers:
   * Tier 2 - Builds for this target are run in CI. Failures during the build
              blocks the inclusion of new code. Tests may be run, but failures
              in tests don't block the inclusion of new code.
+  * Tier 3 - Builds for this target are un in CI. Failures during the build
```
posborne: Typo on "run" (un)

posborne:
I would like to see Android move toward being a higher tier (at least 2 but hopefully tier 1 eventually) platform over time as I will have the need to use nix on these platforms as part of my day job but this seems like the pragmatic move for now.

.travis.yml
```diff
@@ -77,6 +87,16 @@ matrix:
   # Testing nightlies on main targets. These might fail because of issues
```
asomers: This comment is out of date

Susurrus:
@asomers Skipping `x86_64` was a mistake. That being said, x86 Android is dead or dying as far as I understand, so not much of a loss anyways.

@posborne While having all of these platforms Tier 1 is the goal, they first need to get to Tier 3 and regularly part of CI as a step towards that. Additionally, there is work with trust/cross to add test support for Android, so hopefully we won't have to do much special besides continue to test those systems.

Thanks for the feedback, I've updated things accordingly, so we'll see if Travis is indeed fixed.

ndusart:
@asomers libc is testing on Android using Docker and QEMU, maybe we could try to use the same method.

posborne:
> @asomers libc is testing on Android using Docker and QEMU, maybe we could try to use the same method.

With the previous docker containers, we had this going.  I don't think the tests were passing, but it should be possible.  Getting the Android artifacts into the container is a little bit hairy but there should be plenty of prior art that exists (other docker containers doing similar stuff).

Susurrus:
Let's defer on discussion of how to move Android past Tier 1 until later. I'd like to get this merged as it's blocking #631.

bors r+

PR #640, opened 2017-06-30
Add iOS as a Tier 3 platform

Susurrus:
With this and #639, then we have at least basic testing recurring for all mobile platforms (note that this is more than the `libc` currently tests, so we may have trouble moving the non-x86 targets to Tier2 until that happens).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,23 @@ matrix:
     - env: TARGET=x86_64-linux-android DISABLE_TESTS=1
       rust: 1.13.0
 
+    # iOS
+    - env: TARGET=aarch64-apple-ios DISABLE_TESTS=1
+      rust: 1.13.0
+      os: osx
+    - env: TARGET=armv7-apple-ios DISABLE_TESTS=1
+      rust: 1.13.0
+      os: osx
+    - env: TARGET=armv7s-apple-ios DISABLE_TESTS=1
+      rust: 1.13.0
+      os: osx
+    - env: TARGET=i386-apple-ios DISABLE_TESTS=1
+      rust: 1.13.0
+      os: osx
+    - env: TARGET=x86_64-apple-ios DISABLE_TESTS=1
+      rust: 1.13.0
+      os: osx
+
     # Linux
     - env: TARGET=aarch64-unknown-linux-gnu
       rust: 1.13.0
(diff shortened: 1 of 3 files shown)


asomers:
ios isn't listed as a supported target at https://github.com/japaric/cross .  Is the documentation just out of date?  You should try cross building cargo for ios.  If it works, then you can update cross's supported target list.  If not, then I don't think we can trust cross to build nix for that platform.

Also, the build error looks ominous: `error[E0463]: can't find crate for std`

Susurrus:
It isn't listed, but I'm working on the `trust` side of things in japaric/trust#72.

Susurrus:
So this now works for building all 5 iOS targets. I don't think this is too controversial, as nobody's chimed in here, so I'll go ahead and merge this in a bit if someone doesn't beat me to it.

Susurrus:
Was an intermittent failure from one of the `errno` tests. I think we'll need to create a new subprocess for those tests.

bors r+

Susurrus:
`test::sys::test_aio::test_aio_cancel_all()` failed this most recent time. The solution might be to run these tests in separate processes or just not run the tests in QEMU.

Let's try this again...

bors r+

asomers:
@Susurrus I saw that test fail recently for powerpc.  Which arch failed it just now?

Susurrus:
@asomers  `powerpc-unknown-linux-gnu`

Susurrus:
I've definitely seen it fail before. I'm betting most of those tests need to be disabled on powerpc as well because of weird QEMU issues.

Susurrus:
LOL, back to `test_errno_values` not finishing. Have you seen that fail as well?

asomers:
I haven't seen any problems with test_errno_values

Susurrus:
I'm going to try running tests using a single thread and see if that does anything. I can't imagine why `test_errno_values` stalls out as it doesn't have any looping and it's a fairly simple test. This is my first attempt to debug it.

asomers:
I thought it was sufficient to put `CARGO_TEST_THREADS=1` in `.travis.yml`.  I guess we'll see.

Susurrus:
It wouldn't be, because `cross` gets rid of all environment variables but it allows a whitelist of environment variables.

Susurrus:
Well that worked, I'm going to move that testing commit to another branch, and we'll see if it works there.

Susurrus:
And of course, now all the tests pass...

bors r+

Susurrus:
I guess I could just merge this without relying on bors, since we're a pretty low-traffic repo, but I'll do this the "right" way here.

PR #645, opened 2017-07-03
Use single threads for testing

Susurrus:
I actually couldn't find a reference to CARGO_TEST_THREADS in the
current cargo codebase, but RUST_TEST_THREADS is definitely correct.
This should reduce some spurious failures on some platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ sudo: required
 env:
   global:
     - CRATE_NAME=nix
-    - CARGO_TEST_THREADS=1
+    - RUST_TEST_THREADS=1
 
 matrix:
   # These are all the build jobs. Adjust as necessary. Comment out what you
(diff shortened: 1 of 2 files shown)


Susurrus:
The reason for `Cross.toml` is that `Cross` runs everything (but not on Mac) in a docker instance, so it uses a whitelist to passthrough environment variables.

Susurrus:
I'm actually just realizing that I don't think that `trust` has released since that feature was added, so this probably still doesn't do anything on non-Mac platforms.

asomers:
Weird.  It looks like @posborne set that variable correctly on his branch, but somehow it didn't get merged correctly.
Compare https://github.com/posborne/rust-nix/commit/0e3ef36c7406f92483bfc75a37a382a1509fdb54
to a8cf6cbb9de5c4a4f65687c424933220289e5333

bors r+

Issue #244, opened 2016-01-23
SOCK_NONBLOCK should re-export libc::SOCK_NONBLOCK


jamperry:
I believe there is a bug for SOCK_NONBLOCK as the value is not the same for all architectures so will break if it's used on a non-x86 architecture - like MIPS. For example the flag value is 128 on MIPS.

jamperry:
I just noticed that you have a dependency to `libc`. I submitted a PR to add SOCK_NONBLOCK and SOCK_CLOEXEC flags to `libc`  so you can potentially reference them and don't have to split your code into different architectures.

kamalmarhubi:
Thanks for taking a look. Bringing in the value from `libc` is probably the right thing to do.

jamperry:
Alex has already merged the change into `libc` so I can submit a PR to you - hopefully tonight if I get time.

kamalmarhubi:
There's no release out with it yet, from what I could see.

kamalmarhubi:
@jamperry I just checked, both your commits made it into libc 0.2.6 which is out.

Susurrus:
@jamperry Any interest in filing a PR for this? Seems pretty straightforward since the constant is properly defined in libc.

Susurrus:
I believe this is resolved as of f067e73.

kristate:
@kamalmarhubi @Susurrus
My apologies, it seems that `SOCK_NONBLOCK` does not exist in darwin.
I will stick to using `fcntl()` and friends, thanks.

kristate:
@kamalmarhubi @Susurrus
Now I'm curious: `nix::sys::socket::socket` provides a fourth parameter that calls `fcntl()` but seemingly does not support macos. Opened an Issue (#861) for clarification. thanks.

PR #646, opened 2017-07-05
merge socket constants

ndusart:
> Refactor the constant declarations such that all constants are only declared once with a #[cfg] that only enables the constant on the correct platforms.

#637

PR #647, opened 2017-07-05
merge socket constants

### LINKED ISSUE #637: Upstream and use socket constants from libc
Currently `src/sys/socket/consts.rs` relies on many hard-coded constants. I've recently gone and updated many of these in #636, but there's more work to do here:

 * Get all constants that are still numbers upstreamed in `libc` for all supported `nix` platforms
 * Adapt `src/sys/socket/consts.rs` to use `libc` values
 * Refactor the constant declarations such that all constants are only declared once with a `#[cfg]` that only enables the constant on the correct platforms.

To determine what constants are available on what platforms, you can search through the source (see [here](https://github.com/rust-lang/libc/pull/625#issue-237947160) for a list of sources of various platforms).

ndusart:
> Refactor the constant declarations such that all constants are only declared once with a #[cfg] that only enables the constant on the correct platforms.

Closes #637 

(same PR as #646 but from another branch, to see if buildbot has a problem with PR made from master branch)


src/sys/socket/consts.rs
```diff
@@ -93,11 +277,16 @@ mod os {
         pub flags MsgFlags: libc::c_int {
             MSG_OOB,
             MSG_PEEK,
+            MSG_DONTWAIT,
+            #[cfg(not(target_os = "dragonfly"))]
             MSG_CTRUNC,
```
Susurrus: `MSG_CTRUNC`, `MSG_EOR`, and `MSG_TRUNC` are all available in `libc` for DragonflyBSD.

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,243 @@
...
     pub const SO_TYPE: c_int = libc::SO_TYPE;
-    pub const SO_BUSY_POLL: c_int = libc::SO_BUSY_POLL;
-    #[cfg(target_os = "linux")]
-    pub const SO_ORIGINAL_DST: c_int = 80;
+    #[cfg(any(target_os = "macos",
+              target_os = "freebsd",
+              target_os = "ios",
+              target_os = "openbsd",
+              target_os = "netbsd"))]
+    pub const SO_USELOOPBACK: c_int = libc::SO_USELOOPBACK;
```
Susurrus: This exists on DragonflyBSD as well.

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,243 @@
...
+              target_os = "netbsd"))]
+    pub const SO_NKE: c_int = 0x1021;
+    #[cfg(target_os = "dragonfly")]
+    pub const SO_NOSIGPIPE: c_int = libc::SO_NOSIGPIPE;
+    #[cfg(any(target_os = "macos",
+              target_os = "freebsd",
+              target_os = "ios",
+              target_os = "openbsd",
+              target_os = "netbsd"))]
+    pub const SO_NOSIGPIPE: c_int = 0x1022;
```
Susurrus: This is defined for `macos` and `ios`

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,243 @@
...
+    #[cfg(any(target_os = "macos",
+              target_os = "freebsd",
+              target_os = "ios",
+              target_os = "openbsd",
+              target_os = "netbsd"))]
+    pub const SO_NOADDRERR: c_int = 0x1023;
```
Susurrus: Defined for macos/ios

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,243 @@
...
+              target_os = "ios",
+              target_os = "openbsd",
+              target_os = "netbsd"))]
+    pub const SO_NP_EXTENSIONS: c_int = 0x1083;
+    #[cfg(any(target_os = "macos",
+              target_os = "freebsd",
+              target_os = "ios",
+              target_os = "openbsd",
+              target_os = "netbsd"))]
+    pub const SO_NWRITE: c_int = 0x1024;
```
Susurrus: defined for macos/ios

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,243 @@
...
     pub const SO_OOBINLINE: c_int = libc::SO_OOBINLINE;
+    #[cfg(target_os = "linux")]
+    pub const SO_ORIGINAL_DST: c_int = 80;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
     pub const SO_PASSCRED: c_int = libc::SO_PASSCRED;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
     pub const SO_PEEK_OFF: c_int = libc::SO_PEEK_OFF;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
     pub const SO_PEERCRED: c_int = libc::SO_PEERCRED;
```
Susurrus: Also supported by openbsd, bitrig, and haiku.

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,243 @@
...
+    pub const AF_SYSTEM: c_int = libc::AF_SYSTEM;
+    #[cfg(any(target_os = "macos", target_os = "ios"))]
+    pub const AF_SYS_CONTROL: c_int = 2;
 
     pub const SOCK_STREAM: c_int = libc::SOCK_STREAM;
     pub const SOCK_DGRAM: c_int = libc::SOCK_DGRAM;
     pub const SOCK_SEQPACKET: c_int = libc::SOCK_SEQPACKET;
     pub const SOCK_RAW: c_int = libc::SOCK_RAW;
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    pub const SOCK_RDM: c_int = libc::SOCK_RDM;
```
Susurrus: Supported on netbsd, freebsd, dragonfly, mac, ios.

Susurrus:
There are also 27 constants here that need to be upstreamed for all supported platforms. I'd love to have that work be done and merged in with this PR so we can have all of these constants defined correctly in `libc` and reused here. Would you be willing to do that work as part of this or is that too much to ask?

ndusart:
I could find the time for this :-)

Just a bit confused where I could find `sys/` headers for BSD platforms. For other I have running machines where I can find them.

Susurrus:
Excellent, we'd very much appreciate it. We have a huge backlog of FFI constants and functions that aren't correct across all platforms, and it'd be awesome to get better cross-platform support by upstreaming proper values.

I documented this in an old PR of mine: rust-lang/libc#625. I keep meaning to add that to the libc docs so that new users can more easily add missing constants, but I haven't had the time yet.

src/sys/socket/consts.rs
ndusart: I wrote the `#[cfg]` in an exclusive manner as it's shorter (there is a `not()`).
I'll use an inclusive form as it seems to match other `#[cfg]`

Susurrus: Ahh, yes, missed that, sorry. But it is a good idea to use a whitelist and not a blacklist, as we'll be adding new platforms moving forward and they'll be more robust to these additions (though unfortunately likely more verbose).

asomers:
@ndusart you'll have to go straight to the source.  libc is built against the most recent releases of each of the BSDs.  Here are some links you might find helpful:
* https://svnweb.freebsd.org/base/releng/11.0/ (alternate https://github.com/freebsd/freebsd/tree/releng/11.0)
* https://github.com/NetBSD/src/tree/netbsd-7
* https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/ (alternate https://github.com/openbsd/src)
* https://gitweb.dragonflybsd.org/dragonfly.git/commit/2e8b430233bef891dff30e79d4ba53eb68670574

ndusart:
Thanks for the links.

@Susurrus I made the proposed change, thanks for the listing.

I added some marker comments beside the symbols I'd like to check. It'll be cleaned, it's just for helping me tracking the progress.

src/sys/socket/consts.rs
```diff
@@ -278,15 +288,10 @@ mod os {
             MSG_OOB,
             MSG_PEEK,
             MSG_DONTWAIT,
-            #[cfg(not(target_os = "dragonfly"))]
             MSG_CTRUNC,
-            #[cfg(not(target_os = "dragonfly"))]
             MSG_TRUNC,
-            #[cfg(not(target_os = "dragonfly"))]
             MSG_EOR,
-            #[cfg(any(target_os = "linux", target_os = "android"))]
```
ndusart: Damn, was reading as a blacklist there, and seeing the symbols defined for them I removed it. Should definitely use the same form for all `#[cfg]` ^^

ndusart:
Worked a bit on this now.

I can find `SO_DONTTRUNC` reference only in MacOS source, it's not present on other BSD platforms. Currently we present this constant for all BSD except netbsd and dragonfly. That match the fact that this constant is defined in libc only for Apple architectures.

`SO_LABEL` seems to be only defined in freebsd, macos and ios. As of now, we present it for all BSD except dragonfly.

`SO_NREAD` and `SO_NKE` seem to only be defined in macos and ios. Now, we present it for all BSD except dragonfly.

`SO_NOSIGPIPE` does not seem to be defined for openbsd and netbsd. It appears in some source files, but I think this files are written with cross-platform compatibility in mind and there is some `#ifdef` somewhere to not include that part of the code on this architectures.

`SO_NOADDRERR` does not seem to be defined for openbsd, netbsd and freebsd.

That's how far I get for now.

I'd like to know how to handle these cases, should we remove them or keep them for not breaking the API ?
Maybe I'm wrong and I didn't look correctly, please correct me if it is the case.

Susurrus:
@ndusart One of the reasons we're upstreaming constants is because most of `nix`'s declarations were not correct across all supported platforms, especially as it's been supporting more platforms. So it doesn't surprise me that his many constants are improperly declared in `nix`. I have no concern of breaking the API in `nix` when it's correct to do so because at the end of the day if someone used those constants on platforms they didn't exist on and get undefined behavior, it'd eventually come back as a bug report to `nix`.

ndusart:
I have some difficulties with three constants for macos and ios:

- `SO_RESTRICTIONS`
- `SO_RESTRICT_DENYIN` (became `SO_RESTRICT_DENY_IN` since Mac OS 10.9)
- `SO_RESTRICT_DENYOUT` (became `SO_RESTRICT_DENY_OUT` since Mac OS 10.9)

There are defined here: https://github.com/opensource-apple/xnu/blob/dc0628e187c3148723505cf1f1d35bb948d3195b/bsd/sys/socket.h#L188-L190

But cannot make libc CI find these constants. The header was well included and `PRIVATE` was defined but no luck.

Should we keep them hardcoded ? The values are correct. (Just `SO_RESTRICT_DENYSET` which was removed since 10.9)

Susurrus:
Those constants don't actually exist on Mac as of 10.11.6 (I have one locally). I did `grep SO_RESTRICT -r /usr/include` and turned up nothing. Additionally I looked through `/usr/include/sys/socket.h` and those values are gone. They shouldn't be added to `libc` for `macos` and `ios` and be excluded in `nix` as well.

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,253 @@
 pub use self::os::*;
 
-#[cfg(any(target_os = "linux", target_os = "android"))]
 mod os {
```
Susurrus: I don't think we need this `os` module anymore since all constants are now included individually.

src/sys/socket/consts.rs
```diff
@@ -1,75 +1,253 @@
...
     use libc::{self, c_int, uint8_t};
 
     pub const AF_UNIX: c_int  = libc::AF_UNIX;
```
Susurrus: Does it make sense to group any of these values together into enums?

src/sys/socket/consts.rs
ndusart: Sure, I'll remove this

src/sys/socket/consts.rs
ndusart: Is not that already done in the `nix::sys::socket::addr` module ?
Specifically `AF_*` are grouped in the `AddressFamily` enum.

ndusart: I'm wondering then if this `consts` module should not be removed completely in favor of libc ?

Susurrus: It definitely should. All constants should be declared in `libc` and we should just wrap them up into nicer types. It looks like that's already done for the `AF_*` constants, but it should be done for these other ones as well.

ndusart:
`AF_*` constants are in `AddressFamily` enum.
`SOCK_*` constants are defined in `SockType` enum.
`SOL_*` and `IPPROTO_*` are defined in `SockLevel` enum. (`IPPROTO_*` seem to be preferred over `SOL_*`, `SockLevel` correctly use these when applicable)

The constants that are not being defined in an enum are the **socket options** constants: `SO_*`, `IP_*` or `TCP_*` (maybe other depending on the protocol level of the option), but structs are built around them in `sockopts.rs` so we have sort of an "higher-level type" for these constants.

My last commit remove the `consts` module and tests are passing on linux (x86_64) when I use [my branch of libc](https://github.com/ndusart/libc/tree/socket-consts) where I added the missing definitions (rust-lang/libc/pull/656).

src/sys/socket/addr.rs
```diff
@@ -702,7 +702,7 @@ pub mod sys_control {
             let addr = sockaddr_ctl {
                 sc_len: mem::size_of::<sockaddr_ctl>() as c_uchar,
                 sc_family: AddressFamily::System as c_uchar,
-                ss_sysaddr: consts::AF_SYS_CONTROL as uint16_t,
```
Susurrus: This is defined as a `c_int`, is `sockaddr_ctl` correct or is `libc` wrong on the datatype?

ndusart: `sockaddr_ctl` is correctly defined. It's defined in macos [here](https://github.com/opensource-apple/xnu/blob/53c5e2e62fc4182595609153d4b99648da577c39/bsd/sys/kern_control.h#L137-L144), `ss_sysaddr` is indeed `u_int16_t`.

I do not know if `AF_SYS_CONTROL` is used anywhere else.
We could still make that change in libc as it has not been released yet.

Susurrus: Can you file a PR with upstream then? Seems like `AF_SYS_CONTROL` should be a `u16`, do you agree?

ndusart: I definitely agree. I'll be able to do that tomorrow.

Susurrus: Looking at `https://github.com/opensource-apple/xnu/blob/dc0628e187c3148723505cf1f1d35bb948d3195b/bsd/sys/_types/_sa_family_t.h#L570` it looks like the value is used as multiple datatypes. This might not be super straightforward.

ndusart: I don't think `AF_SYS_CONTROL` should be considered as the same type as other `SA_*` constants. You'll not use it in `socket` function for example, there you should use `AF_SYSTEM`. `AF_SYS_CONTROL` seems like a subtype to `AF_SYSTEM` that only appear in `sockaddr_ctl` when connecting to the socket.
But you are right, `AF_*` constants are used sometimes as `int` (in `socket` function) or as `sa_family_t` (in `sockaddr` struct).
I guess the problem is not applicable to `AF_SYS_CONTROL` but I cannot find good documentation about its use though :/

Susurrus: So I think this is resolved as a WONTFIX on our end then.

ndusart: Not sure. `AF_SYS_CONTROL` does not seem to be used in other way than as `u16` but I never used these `AF_SYSTEM` addresses, so in doubt, probably it's better to let it as `c_int` ?

src/sys/socket/sockopt.rs
```diff
@@ -130,52 +131,52 @@ macro_rules! sockopt_impl {
...
           target_os = "linux",
           target_os = "android",
           target_os = "nacl"))]
-sockopt_impl!(Both, TcpKeepIdle, consts::IPPROTO_TCP, consts::TCP_KEEPIDLE, u32);
-sockopt_impl!(Both, RcvBuf, consts::SOL_SOCKET, consts::SO_RCVBUF, usize);
-sockopt_impl!(Both, SndBuf, consts::SOL_SOCKET, consts::SO_SNDBUF, usize);
+sockopt_impl!(Both, TcpKeepIdle, libc::IPPROTO_TCP, libc::TCP_KEEPIDLE, u32);
+sockopt_impl!(Both, RcvBuf, libc::SOL_SOCKET, libc::SO_RCVBUF, usize);
+sockopt_impl!(Both, SndBuf, libc::SOL_SOCKET, libc::SO_SNDBUF, usize);
 #[cfg(target_os = "linux")]
```
Susurrus: I think this and `SndBufForce` can be implemented on Android as well.

src/sys/socket/sockopt.rs
```diff
@@ -130,52 +131,52 @@ macro_rules! sockopt_impl {
...
-sockopt_impl!(SetOnly, RcvBufForce, consts::SOL_SOCKET, consts::SO_RCVBUFFORCE, usize);
+sockopt_impl!(SetOnly, RcvBufForce, libc::SOL_SOCKET, libc::SO_RCVBUFFORCE, usize);
 #[cfg(all(target_os = "linux", not(target_arch="arm")))]
-sockopt_impl!(SetOnly, SndBufForce, consts::SOL_SOCKET, consts::SO_SNDBUFFORCE, usize);
-sockopt_impl!(GetOnly, SockType, consts::SOL_SOCKET, consts::SO_TYPE, super::SockType);
+sockopt_impl!(SetOnly, SndBufForce, libc::SOL_SOCKET, libc::SO_SNDBUFFORCE, usize);
+sockopt_impl!(GetOnly, SockType, libc::SOL_SOCKET, libc::SO_TYPE, super::SockType);
 #[cfg(any(target_os = "freebsd",
```
Susurrus: I think this is available on all targets, why don't you just remove the `cfg` attribute and we'll see if everything passes.

src/sys/socket/mod.rs
```diff
@@ -70,6 +69,22 @@ bitflags!(
     }
 );
 
+// Flags for send/recv and their relatives
+libc_bitflags!{
+    pub flags MsgFlags: libc::c_int {
+        MSG_OOB,
```
Susurrus: Can you add doc comments for each of these? They're documented [here](http://man7.org/linux/man-pages/man2/send.2.html).

src/sys/socket/addr.rs
```diff
@@ -23,15 +23,15 @@ use ::sys::socket::addr::sys_control::SysControlAddr;
...
+    Inet6 = libc::AF_INET6,
     #[cfg(any(target_os = "linux", target_os = "android"))]
-    Netlink = consts::AF_NETLINK,
+    Netlink = libc::AF_NETLINK,
     #[cfg(any(target_os = "linux", target_os = "android"))]
-    Packet = consts::AF_PACKET,
+    Packet = libc::AF_PACKET,
     #[cfg(any(target_os = "macos", target_os = "ios"))]
-    System = consts::AF_SYSTEM,
+    System = libc::AF_SYSTEM,
```
Susurrus: There are actually a ton more `AF_` constants defined, at least on Linux. If you search through the `libc` source, you'll find a bunch, would you be willing to add them for the appropriate platforms?

src/sys/socket/sockopt.rs
ndusart: You're right and `SndBufForce` if defined in Linux ARM too.

src/sys/socket/addr.rs
ndusart: Ok, I'll go through the constants defined in `libc` and try to merge all them in here.

src/sys/socket/sockopt.rs
ndusart: Done

ndusart:
Travis is compiling `libc` 0.2.26 instead if the upstream master branch. Then it's complaining about missing constants that are not released yet.
Do you know why it keeps this version ?

Wolvereness:
@ndusart This commit is being merged: 5527ce8f00d5c993f39c57f5af5956d0d6cd89af, which changes the version in the `Cargo.toml` to https://github.com/rust-lang/libc/commit/fa23a9bc42194f354e3743e228a385198b9862c1, and it doesn't include the changes you need. You need to rebase on current master, because 3912a202255d9fee797703966998babbc79fc49b (your upstream) is before the previously mentioned commit, and then change the dependency back.

ndusart:
Thanks @Wolvereness.
I thought CI was testing against the head of the source branch of the PR. (It seems it's the case for buildbot as it fetch the master branch of libc)

Looking in Travis CI logs, I see that it fetch the source with
```sh
git fetch origin +refs/pull/647/merge
```
which is the PR merged to master.

I'll rebase then with `master` so I can modify `libc` dependency with a new commit.

ndusart:
`SYSPROTO_CONTROL` was not defined anymore since the removal of the `consts` module.

I added a type `SockProtocol` for constants expected in the `protocol` argument of `socket` and `socketpair`.
I'll add some constants found in man pages if you find this type correct.
But a lot are not present in libc (like [protocols for NETLINK sockets](http://man7.org/linux/man-pages/man7/netlink.7.html)) so I wonder if we can postpone the add of the other protocols to another PR ?

test/sys/test_sockopt.rs
```diff
@@ -1,9 +1,9 @@
 use rand::{thread_rng, Rng};
-use nix::sys::socket::{socket, sockopt, getsockopt, setsockopt, AddressFamily, SockType, SockFlag, SockLevel};
+use nix::sys::socket::{socket, sockopt, getsockopt, setsockopt, AddressFamily, SockType, SockFlag, SockProtocol};
 
 #[test]
 fn test_so_buf() {
-    let fd = socket(AddressFamily::Inet, SockType::Datagram, SockFlag::empty(), SockLevel::Udp as i32).unwrap();
+    let fd = socket(AddressFamily::Inet, SockType::Datagram, SockFlag::empty(), SockProtocol::Udp).unwrap();
```
ndusart: Maybe we should use `SockProtocol::Default` (0) here as well as it seems that `SOCK_DGRAM` only expect the protocol to be zero ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))

src/sys/socket/mod.rs
```diff
@@ -61,6 +61,16 @@ pub enum SockType {
     Rdm = libc::SOCK_RDM,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(i32)]
+pub enum SockProtocol {
+    Default = 0,
```
Susurrus: Seems like this shouldn't exist and really `socket()` should take an `Option<SockProtocol>`.

Also this enum and all it's variants needs docs. Be sure to link to relevant manpages!

src/sys/socket/mod.rs
```diff
@@ -48,18 +48,27 @@ pub use self::multicast::{
...
+    Tcp = libc::IPPROTO_TCP,
+    Udp = libc::IPPROTO_UDP,
+    #[cfg(any(target_os = "macos", target_os = "ios"))]
+    SysControl = libc::SYSPROTO_CONTROL,
 }
 
 // Extra flags - Supported by Linux 2.6.27, normalized on other platforms
```
Susurrus: Can you replace this with `libc_bitflags!` as well? These constants should be available already in `libc` so this shouldn't be a problem. This would close #244.

ndusart: Using `libc_enum!` as we cannot mix these values.

src/sys/socket/mod.rs
ndusart: Good point. Fixed.

src/sys/socket/addr.rs
```diff
@@ -252,7 +413,7 @@ impl Ipv4Addr {
     }
 
     pub fn any() -> Ipv4Addr {
-        Ipv4Addr(libc::in_addr { s_addr: consts::INADDR_ANY })
+        Ipv4Addr(libc::in_addr { s_addr: libc::INADDR_ANY })
```
ndusart: Waiting for rust-lang/libc#698 for types to match.

src/sys/socket/mod.rs
Susurrus: Ahh, yes, that's what you should use. Good catch!

src/sys/socket/addr.rs
```diff
@@ -23,15 +23,176 @@ use ::sys::socket::addr::sys_control::SysControlAddr;
 #[repr(i32)]
```
Susurrus: Can you also delete the comments above and the `TODO` (which I think is actually resolved also)? It'd be good to clean those up since you're already in here.

src/sys/socket/addr.rs
```diff
@@ -23,15 +23,176 @@ use ::sys::socket::addr::sys_control::SysControlAddr;
...
 #[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
 pub enum AddressFamily {
-    Unix = consts::AF_UNIX,
-    Inet = consts::AF_INET,
-    Inet6 = consts::AF_INET6,
+    Unix = libc::AF_UNIX,
+    Inet = libc::AF_INET,
+    Inet6 = libc::AF_INET6,
     #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: The `target_os`'s should all be alphabetized through this file.

src/sys/socket/mod.rs
```diff
@@ -48,18 +48,31 @@ pub use self::multicast::{
...
+    Stream = libc::SOCK_STREAM,
+    Datagram = libc::SOCK_DGRAM,
+    SeqPacket = libc::SOCK_SEQPACKET,
+    Raw = libc::SOCK_RAW,
+    Rdm = libc::SOCK_RDM,
+}
+
+libc_enum!{
+    #[repr(i32)]
+    pub enum SockProtocol {
```
Susurrus: This needs a comment for the entire enum here as well.

Susurrus: Is there a reason we didn't prettify the names here like for all of the other types? This really seems like it should have 3 variants: `Tcp`, `Udp`, and `Kext`.

src/sys/socket/mod.rs
```diff
@@ -48,18 +48,31 @@ pub use self::multicast::{
...
+        /// UDP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
+        IPPROTO_UDP,
+        /// Allows applications to configure and control a KEXT
+        /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
+        #[cfg(any(target_os = "macos", target_os = "ios"))]
+        SYSPROTO_CONTROL,
+    }
 }
 
 // Extra flags - Supported by Linux 2.6.27, normalized on other platforms
```
Susurrus: This should be switched to use `libc_bitflags` and the upstream declarations of these values.

src/sys/socket/mod.rs
```diff
@@ -330,9 +396,20 @@ pub fn recvmsg<'a, T>(fd: RawFd, iov: &[IoVec<&mut [u8]>], cmsg_buffer: Option<&
...
+/// The `protocol` specifies a particular protocol to be used with the
+/// socket.  Normally only a single protocol exists to support a
+/// particular socket type within a given protocol family, in which case
+/// protocol can be specified as `None`.  However, it is possible that many
+/// protocols may exist, in which case a particular protocol must be
+/// specified in this manner.
+///
 /// [Further reading](http://man7.org/linux/man-pages/man2/socket.2.html)
-pub fn socket(domain: AddressFamily, ty: SockType, flags: SockFlag, protocol: c_int) -> Result<RawFd> {
+pub fn socket(domain: AddressFamily, ty: SockType, flags: SockFlag, protocol: Option<SockProtocol>) -> Result<RawFd> {
```
Susurrus: Sorry, I should've mentioned this in my last review. We should actually change this to:
```
pub fn socket<T: Into<Option<SockProtocol>>>(domain: AddressFamily, ty: SockType, flags: SockFlag, protocol: T) -> Result<RawFd> {
```

And then within the function you just change usages of `protocol` to `protocol.into()`.

src/sys/socket/mod.rs
```diff
@@ -358,9 +435,13 @@ pub fn socket(domain: AddressFamily, ty: SockType, flags: SockFlag, protocol: c_
 /// Create a pair of connected sockets
 ///
 /// [Further reading](http://man7.org/linux/man-pages/man2/socketpair.2.html)
-pub fn socketpair(domain: AddressFamily, ty: SockType, protocol: c_int,
+pub fn socketpair(domain: AddressFamily, ty: SockType, protocol: Option<SockProtocol>,
```
Susurrus: Same change to the function arguments as above.

test/sys/test_sockopt.rs
```diff
@@ -1,9 +1,11 @@
 use rand::{thread_rng, Rng};
-use nix::sys::socket::{socket, sockopt, getsockopt, setsockopt, AddressFamily, SockType, SockFlag, SockLevel};
+use nix::sys::socket::{socket, sockopt, getsockopt, setsockopt, AddressFamily, SockType, SockFlag, SockProtocol};
 
 #[test]
 fn test_so_buf() {
-    let fd = socket(AddressFamily::Inet, SockType::Datagram, SockFlag::empty(), SockLevel::Udp as i32).unwrap();
+    let fd = socket(AddressFamily::Inet, SockType::Datagram,
+                    SockFlag::empty(), Some(SockProtocol::IPPROTO_UDP))
```
Susurrus: With the above function argument changes you can get rid of the `Some(` here.

test/sys/test_socket.rs
```diff
@@ -223,11 +220,12 @@ pub fn test_unixdomain() {
...
 pub fn test_syscontrol() {
     use nix::{Errno, Error};
-    use nix::sys::socket::{AddressFamily, SockType, SockFlag};
+    use nix::sys::socket::{AddressFamily, SockType, SockFlag, SockProtocol};
     use nix::sys::socket::{socket, SockAddr};
-    use nix::sys::socket::SYSPROTO_CONTROL;
 
-    let fd = socket(AddressFamily::System, SockType::Datagram, SockFlag::empty(), SYSPROTO_CONTROL).expect("socket failed");
+    let fd = socket(AddressFamily::System, SockType::Datagram,
+                    SockFlag::empty(), Some(SockProtocol::SYSPROTO_CONTROL))
```
Susurrus: Get rid of the `Some(` here.

test/sys/test_socket.rs
```diff
@@ -223,11 +220,12 @@ pub fn test_unixdomain() {
 #[test]
 pub fn test_syscontrol() {
     use nix::{Errno, Error};
-    use nix::sys::socket::{AddressFamily, SockType, SockFlag};
+    use nix::sys::socket::{AddressFamily, SockType, SockFlag, SockProtocol};
```
Susurrus: Please combine this with the following line.

Susurrus:
This is looking really good, thanks for doing all this work! The `AddressFamily` and `SockType` enums could use docs as well as they're the only ones missing them now. It'd be really nice to add those as part of this work!

Susurrus:
One other thing that should be done here, since we're switching to already-tested `libc` constants, is that our tests in `libc-test/src/const.c` should be modified to not test these constants again. I'm pretty certain all of the constants there under `===== SOCKET OPTIONS =====` can be removed. Can you double-check that and remove them if so?

src/sys/socket/mod.rs
ndusart: There were prettify but you asked to used `libc_bitflags!` for these :P  

We cannot have both I think, we have to choose ^^

src/sys/socket/mod.rs
ndusart: I used `libc_enum!` because we should not use a combination of theses constants. They are not flags. (`IPPROTO_TCP` is 6 for example)

src/sys/socket/mod.rs
ndusart: That would simplify the calls indeed !

src/sys/socket/addr.rs
ndusart: Alright, I'll sort them.

src/sys/socket/addr.rs
ndusart: It's is fixed upstream.

src/sys/socket/mod.rs
Susurrus: Yes, you're right, `libc_enum!` is correct here.

src/sys/socket/mod.rs
Susurrus: That's true, sorry about that! I think we should either use all the same constants or prettify all of them. I like the pretty version, so let's make all of these pretty.

ndusart: Ok, I'll do it.

Sorry, I seem responsive but I'm not able to make these changes right now. I'll be able to do it tomorrow.

Susurrus: No worries, there's no deadline on doing any of this stuff unless it's been a few months with no action. This all runs on your own schedule.

src/sys/socket/addr.rs
ndusart: Done

ndusart:
I added doc to `AddressFamily` and `SockType`.

Regarding the test `nix-text/src/const.c`, are these tests ever run ? Because tests are passing here while constants are not defined anymore.
And yes, they can be all removed as they are no socket constant definition in this PR.

Susurrus:
@ndusart Yes, the tests are run. When you run `cargo test` it compiles and runs `nix-test` as part of that. It's only if there are tests in `/test` that depend on stuff implemented in `nix-test` will you see anything happen. Currently that's not for very much right now.

ndusart:
Well, I think is quite done now. :)
Waiting rust-lang/libc#701 to fix the test failure for s390x, and it should be ok.

Do you have any remark on the current state of this PR ?

ndusart:
Tests should pass now.

src/sys/socket/mod.rs
```diff
@@ -70,6 +101,59 @@ bitflags!(
     }
 );
 
+// Flags for send/recv and their relatives
```
Susurrus: The doc comment for this should be the first line inside the macro.

ndusart:
Done, and convert these comments into doc comments

src/sys/socket/addr.rs
```diff
@@ -642,10 +809,10 @@ pub mod netlink {
     }
 }
 
-#[cfg(any(target_os = "macos", target_os = "ios"))]
+#[cfg(any(target_os = "ios", target_os = "macos"))]
 pub mod sys_control {
-    use ::sys::socket::consts;
     use ::sys::socket::addr::{AddressFamily};
+    use libc;
```
Susurrus: This can be fixed by modifying the following line to be `use libc::{self, ...`

src/sys/socket/mod.rs
```diff
@@ -9,9 +9,9 @@ use libc::{c_void, c_int, socklen_t, size_t, pid_t, uid_t, gid_t};
 use std::{mem, ptr, slice};
 use std::os::unix::io::RawFd;
 use sys::uio::IoVec;
+use libc;
```
Susurrus: Same with this `use`, combine it with the previous one.

src/sys/socket/sockopt.rs
```diff
@@ -1,9 +1,8 @@
-use super::{ffi, consts, GetSockOpt, SetSockOpt};
+use super::{ffi, GetSockOpt, SetSockOpt};
+use libc;
```
Susurrus: Same thing here.

ndusart:
Are these CHANGELOGS entry good ?

I merged the use statements.

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - On Linux and Android, added support for receiving `PTRACE_O_TRACESYSGOOD`
   events from `wait` and `waitpid` using `WaitStatus::PtraceSyscall`
   ([#566](https://github.com/nix-rust/nix/pull/566)).
+- Added protocol families in `AddressFamily` enum.
```
Susurrus: These need to be moved above into the "Unreleased" section.

CHANGELOG.md
```diff
@@ -64,6 +66,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   "*_buf" variants also now calculate total array size and take slice references for improved type
   safety. The documentation has also been dramatically improved.
   ([#670](https://github.com/nix-rust/nix/pull/670))
+- Changed function signature of `socket()` and `socketpair()`. The `protocol` argument
```
Susurrus: Same as above.

Susurrus:
This needs a rebase, and you need to move the CHANGELOG entries when you do that. I'd also prefer that the last commit be merged into the other commits where appropriate instead of being a separate one, but this doesn't need to block merging this.

ndusart:
@Susurrus rebased on master

Susurrus:
Awesome, thanks @ndusart!

bors r+

PR #648, opened 2017-07-05
 Fix `struct msghdr` types on non-Linux platforms.

### LINKED ISSUE #748: Migrate all FFI declarations to libc
We should remove all FFI declarations within `nix` in favor of types defined in the `libc` crate. Much of this work has already been done recently, but there's still some left. And since we don't let any new PRs be merged until `libc` support has been added, we can resolve this problem relatively easily.

Left to address:
 * [ ] `src/sys/socket/mod.rs` (~~#746~~, #648)
 * [x] `src/sys/memfd.rs` (#794)
 * [x] `src/fcntl.rs` (see #745)
 * [x] `src/sched.rs` (see #747)
 * [x] `src/sys/quota.rs` (see #750)
 * [x] `src/sys/ptrace.rs` (see #749)
 * [x] `src/sys/syscall.rs` (see #747)
 * [x] `src/errno.rs` (see #798, #684, #807)

kinetiknz:
Type `struct msghdr` and `struct cmsghdr` types defined in `sys/socket/ffi.rs` match the Linux headers, but differ from the standard header (and other OSes): http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html

This PR fixes a memory corruption/unsafety issue on non-Linux machines when using `recvmsg` with certain parameters.

While fixing this, I wondered what the reason was for nix not using libc's definition of these structures.

Closes #748.


Susurrus:
I'd like to take the policy that we never fix FFI stuff directly within `nix`, instead only fixing in `libc` and switching our types to use those. All the types defined in `socket::ffi` actually already exist in libc. Can you just remove that module and use the `libc` types directly instead?

Susurrus:
Whoops, should have read your entire post. The goal is for `nix` to exclusively use `libc` datatypes for all platforms, the only reason not to would be if they aren't fixing their end.

kinetiknz:
> Can you just remove that module and use the libc types directly instead?

Will do.  Updated PR coming soon.

CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#623](https://github.com/nix-rust/nix/pull/623))
 - Multiple constants related to the termios API have now been properly defined for
   all supported platforms. ([#527](https://github.com/nix-rust/nix/pull/527))
+- Fixed field types of `sys::socket::ffi::{msghdr, cmsghdr}` on
+  non-Linux platforms.  ([#](https://...))
```
Susurrus: You can fill this in at this point. It won't change between now and when it's committed.

src/sys/socket/ffi.rs
```diff
@@ -33,9 +48,9 @@ pub struct msghdr<'a> {
     pub msg_name: *const c_void,
```
Susurrus: Both `msghdr` and `cmsghdr` here should be removed in favor of the declarations in `libc`.

kinetiknz:
Apologies for the delay, I've been busy with other stuff.  I'll try to get the PR updated this week with the removal of socket/ffi.rs and various fixes for this area of the code.  Most of the work is already complete, it just needs tidying up.

Susurrus:
@kinetiknz I actually did some work in this area (see #748). Would you be willing to push this through? This is the last remaining FFI declarations in nix.

kinetiknz:
Sure thing, apologies for letting this sit so long!

src/sys/socket/mod.rs
```diff
@@ -198,8 +203,10 @@ use self::ffi::{cmsghdr, msghdr, type_of_cmsg_data, type_of_msg_iovlen, type_of_
 /// To make room for multiple messages, nest the type parameter with
 /// tuples, e.g.
 /// `let cmsg: CmsgSpace<([RawFd; 3], CmsgSpace<[RawFd; 2]>)> = CmsgSpace::new();`
```
Susurrus: I think we should make this a doccomment so it can be compile & run tested.

src/sys/socket/mod.rs
```diff
@@ -380,7 +388,7 @@ pub enum ControlMessage<'a> {
 pub struct UnknownCmsg<'a>(&'a cmsghdr, &'a [u8]);
 
 fn cmsg_align(len: usize) -> usize {
```
Susurrus: I'd suggest we add an `#[inline]` here as it's a small convenience function.

Susurrus: Could you also add a comment here summarizing this function? I'm having trouble following this code and having a comment for this helper function would be nice.

Susurrus:
@kinetiknz Just pinging you again on this as it's the last nugget before we have all FFI types upstreamed.

kinetiknz:
> @kinetiknz Just pinging you again on this as it's the last nugget before we have all FFI types upstreamed.

Sorry! I'll push my changes tomorrow.

src/sys/socket/mod.rs
```diff
@@ -197,9 +202,16 @@ use self::ffi::{cmsghdr, msghdr, type_of_cmsg_data, type_of_msg_iovlen, type_of_
 ///
 /// To make room for multiple messages, nest the type parameter with
 /// tuples, e.g.
```
Susurrus: Remove the ", e.g." and add ":"

src/sys/socket/mod.rs
```diff
@@ -405,17 +423,17 @@ impl<'a> ControlMessage<'a> {
...
     unsafe fn encode_into<'b>(&self, buf: &mut &'b mut [u8]) {
         match *self {
             ControlMessage::ScmRights(fds) => {
-                let cmsg = cmsghdr {
-                    cmsg_len: self.len() as type_of_cmsg_len,
-                    cmsg_level: libc::SOL_SOCKET,
-                    cmsg_type: libc::SCM_RIGHTS,
-                    cmsg_data: [],
+                let cmsg = {
+                  let mut cmsg: cmsghdr = mem::uninitialized();
```
Susurrus: Shouldn't think be `mem::zero()` so we make sure all memory is in a known state?

src/sys/socket/mod.rs
```diff
@@ -430,11 +448,12 @@ impl<'a> ControlMessage<'a> {
                 copy_bytes(fds, buf);
             },
             ControlMessage::ScmTimestamp(t) => {
-                let cmsg = cmsghdr {
-                    cmsg_len: self.len() as type_of_cmsg_len,
-                    cmsg_level: libc::SOL_SOCKET,
-                    cmsg_type: libc::SCM_TIMESTAMP,
-                    cmsg_data: [],
+                let cmsg = {
+                  let mut cmsg: cmsghdr = mem::uninitialized();
```
Susurrus: Shouldn't think be `mem::zero()` so we make sure all memory is in a known state?

src/sys/socket/mod.rs
```diff
@@ -494,16 +513,18 @@ pub fn sendmsg<'a>(fd: RawFd, iov: &[IoVec<&'a [u8]>], cmsgs: &[ControlMessage<'
...
-    let mhdr = msghdr {
-        msg_name: name as *const c_void,
-        msg_namelen: namelen,
-        msg_iov: iov.as_ptr(),
-        msg_iovlen: iov.len() as type_of_msg_iovlen,
-        msg_control: cmsg_ptr,
-        msg_controllen: capacity as type_of_cmsg_len,
-        msg_flags: 0,
+    let mhdr = unsafe {
+        let mut mhdr: msghdr = mem::uninitialized();
```
Susurrus: Shouldn't think be `mem::zero()` so we make sure all memory is in a known state?

src/sys/socket/mod.rs
```diff
@@ -517,16 +538,18 @@ pub fn recvmsg<'a, T>(fd: RawFd, iov: &[IoVec<&mut [u8]>], cmsg_buffer: Option<&
...
-    let mut mhdr = msghdr {
-        msg_name: &mut address as *const _ as *const c_void,
-        msg_namelen: mem::size_of::<sockaddr_storage>() as socklen_t,
-        msg_iov: iov.as_ptr() as *const IoVec<&[u8]>, // safe cast to add const-ness
-        msg_iovlen: iov.len() as type_of_msg_iovlen,
-        msg_control: msg_control as *const c_void,
-        msg_controllen: msg_controllen as type_of_cmsg_len,
-        msg_flags: 0,
+    let mut mhdr = unsafe {
+        let mut mhdr: msghdr = mem::uninitialized();
```
Susurrus: Shouldn't think be `mem::zero()` so we make sure all memory is in a known state?

src/sys/socket/mod.rs
```diff
@@ -197,9 +202,16 @@ use self::ffi::{cmsghdr, msghdr, type_of_cmsg_data, type_of_msg_iovlen, type_of_
...
-/// `let cmsg: CmsgSpace<([RawFd; 3], CmsgSpace<[RawFd; 2]>)> = CmsgSpace::new();`
+///
+/// ```
+/// use std::os::unix::io::RawFd;
+/// use nix::sys::socket::CmsgSpace;
+/// let cmsg: CmsgSpace<([RawFd; 3], CmsgSpace<[RawFd; 2]>)> = CmsgSpace::new();
+/// ```
+#[repr(C)]
 pub struct CmsgSpace<T> {
     _hdr: cmsghdr,
```
Susurrus: I think both `_hdr` and `_data` should be changed to just `hdr` and `data`, yeah?

Susurrus:
Also, could you rebase and squash all the commits into one since this is really a single big change?

src/sys/socket/mod.rs
```diff
@@ -197,9 +202,16 @@ use self::ffi::{cmsghdr, msghdr, type_of_cmsg_data, type_of_msg_iovlen, type_of_
 ///
 /// To make room for multiple messages, nest the type parameter with
 /// tuples, e.g.
-/// `let cmsg: CmsgSpace<([RawFd; 3], CmsgSpace<[RawFd; 2]>)> = CmsgSpace::new();`
+///
+/// ```
+/// use std::os::unix::io::RawFd;
+/// use nix::sys::socket::CmsgSpace;
+/// let cmsg: CmsgSpace<([RawFd; 3], CmsgSpace<[RawFd; 2]>)> = CmsgSpace::new();
```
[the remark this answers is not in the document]
kinetiknz: There are a few bugs remaining in this code, I have tests and fixes for them that I can either include with this PR or as a new PR.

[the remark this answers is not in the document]
Susurrus: Let's keep them in this PR I think. Probably best to keep these fixes as follow up commits depending on how tightly coupled they all are.

src/sys/socket/mod.rs
kinetiknz: I'm fine with making this change in all of the places suggested, but the prevailing style seems to be towards uninitialized() and zeroed() potentially has a small additional cost.

kinetiknz:
> Also, could you rebase and squash all the commits into one since this is really a single big change?

Will do, but I figured it'd be easier to review as a series of small commits.

src/sys/socket/mod.rs
kinetiknz: Sure.  The code never actually accesses the fields directly either way, so not sure which is clearer.

src/sys/socket/mod.rs
```diff
@@ -187,8 +188,12 @@ unsafe fn copy_bytes<'a, 'b, T: ?Sized>(src: &T, dst: &'a mut &'b mut [u8]) {
     mem::swap(dst, &mut remainder);
 }
 
+// OSX always aligns struct cmsghdr as if it were a 32-bit OS
+#[cfg(target_os = "macos")]
```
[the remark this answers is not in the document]
kinetiknz: Looks like it, thanks.

src/sys/socket/mod.rs
Susurrus: If these are never accessed, then I think this is Rust-y.

src/sys/socket/mod.rs
Susurrus: I was just worried that something might not have gotten initialized. But given that we don't access any of the uninitialized fields, and there's no way for this datatype to escape our wrapper, this makes sense.

Susurrus:
@kinetiknz I think there are just a few small things to finish to wrap this up in a nice pretty bow. You have some time soon to finish this off?

kinetiknz:
Sure, I should have time early next week.

Susurrus:
@kinetiknz Will you have time to do some work on this this week? This is the last piece of `nix` that still relies on its own C types rather than `libc`s.

kinetiknz:
> @kinetiknz Will you have time to do some work on this this week? This is the last piece of nix that still relies on its own C types rather than libcs.

Sorry about the delay, rebased and squashed with the last set of comments addressed.  Also adjusted the definition of `align_of_cmsg_data` for DragonFly BSD based on 709cbdf12cacb6ecf42cca9000d1d46380b6ba62.

src/sys/socket/mod.rs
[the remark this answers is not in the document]
kinetiknz: So I think it'd be better to merge this PR as-is (once you're happy with it, of course) and I'll submit additional tests and fixes in a later PR.  That way my lack of time won't block removing ffi.rs, at least.

src/sys/socket/mod.rs
Susurrus: This would be more clear if rewritten to use struct update syntax:

```rust
    let mut cmsg = cmsghdr {
        cmsg_len: self.len() as _,
        cmsg_level: libc::SOL_SOCKET,
        cmsg_type: libc::SCM_RIGHTS,
        ..unsafe { mem::uninitialized() }
    };
```

src/sys/socket/mod.rs
```diff
@@ -538,11 +556,12 @@ impl<'a> ControlMessage<'a> {
                 copy_bytes(fds, buf);
             },
             ControlMessage::ScmTimestamp(t) => {
-                let cmsg = cmsghdr {
-                    cmsg_len: self.len() as type_of_cmsg_len,
-                    cmsg_level: libc::SOL_SOCKET,
-                    cmsg_type: libc::SCM_TIMESTAMP,
-                    cmsg_data: [],
+                let cmsg = {
```
Susurrus: Same as above.

src/sys/socket/mod.rs
```diff
@@ -602,16 +621,18 @@ pub fn sendmsg<'a>(fd: RawFd, iov: &[IoVec<&'a [u8]>], cmsgs: &[ControlMessage<'
...
-    let mhdr = msghdr {
-        msg_name: name as *const c_void,
-        msg_namelen: namelen,
-        msg_iov: iov.as_ptr(),
-        msg_iovlen: iov.len() as type_of_msg_iovlen,
-        msg_control: cmsg_ptr,
-        msg_controllen: capacity as type_of_cmsg_len,
-        msg_flags: 0,
+    let mhdr = unsafe {
+        let mut mhdr: msghdr = mem::uninitialized();
```
Susurrus: Same as above.

src/sys/socket/mod.rs
```diff
@@ -625,16 +646,18 @@ pub fn recvmsg<'a, T>(fd: RawFd, iov: &[IoVec<&mut [u8]>], cmsg_buffer: Option<&
...
-    let mut mhdr = msghdr {
-        msg_name: &mut address as *const _ as *const c_void,
-        msg_namelen: mem::size_of::<sockaddr_storage>() as socklen_t,
-        msg_iov: iov.as_ptr() as *const IoVec<&[u8]>, // safe cast to add const-ness
-        msg_iovlen: iov.len() as type_of_msg_iovlen,
-        msg_control: msg_control as *const c_void,
-        msg_controllen: msg_controllen as type_of_cmsg_len,
-        msg_flags: 0,
+    let mut mhdr = unsafe {
+        let mut mhdr: msghdr = mem::uninitialized();
```
Susurrus: Same as above.

src/sys/socket/mod.rs
```diff
@@ -295,19 +296,30 @@ unsafe fn copy_bytes<'a, 'b, T: ?Sized>(src: &T, dst: &'a mut &'b mut [u8]) {
     mem::swap(dst, &mut remainder);
 }
 
+// Darwin and DragonFly BSD always aligns struct cmsghdr as if it were a 32-bit OS
+#[cfg(any(target_os = "ios", target_os = "macos", target_os = "dragonfly"))]
```
Susurrus: Can you alphabetize this list?

kinetiknz:
No problem, let me know if there's anything else!

src/sys/socket/mod.rs
kinetiknz: I agree that's nicer, but it looks like the musl target prevents this due to private fields. 😞

kinetiknz: It's only a problem for `msghdr`, so I'll keep the `cmsghdr` changes and revert the rest.

kinetiknz:
Repushed with musl and iOS target fixes.

src/sys/socket/mod.rs
```diff
@@ -513,17 +531,16 @@ impl<'a> ControlMessage<'a> {
...
+    // the provided slice; panics if the buffer is too small.
     unsafe fn encode_into<'b>(&self, buf: &mut &'b mut [u8]) {
         match *self {
             ControlMessage::ScmRights(fds) => {
                 let cmsg = cmsghdr {
-                    cmsg_len: self.len() as type_of_cmsg_len,
-                    cmsg_level: libc::SOL_SOCKET,
-                    cmsg_type: libc::SCM_RIGHTS,
-                    cmsg_data: [],
+                  cmsg_len: self.len() as _,
```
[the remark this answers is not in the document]
kinetiknz: Not sure what happened there, but fixed version pushed.

Susurrus:
Just realized this is more than fixing the FFI stuff, it actually fixes things on non-Linux platforms, so it's a bug fix. Can you add a CHANGELOG entry under the Fixed section for this?

kinetiknz:
> Just realized this is more than fixing the FFI stuff, it actually fixes things on non-Linux platforms, so it's a bug fix. Can you add a CHANGELOG entry under the Fixed section for this?

Added something, suggestions welcome if that's not clear enough.

Susurrus:
Awesome, thanks for following through on this, @kinetiknz. And when/if you have time any additional tests and refinements are always appreciated!

PR #651, opened 2017-07-07
Improve Wait Documentation

### LINKED ISSUE #643: Document nix::sys::wait Module
I'm trying to use `waitpid` in Ion, but I cannot figure out what each of the `WaitStatus` variants and their fields represent.

mmstick:
As the title states, improves the documentation for the wait module by providing documentation where there is currently no documentation. Also, changed the signature of `waitpid` to do as the PID parameter currently does (`Into<Option<T>>`), so that you don't have to always type `Some` every time you want to supply a flag -- but you can still provide a `None` value.

Closes #643.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -7,6 +7,8 @@
 //!
 //! ```rust
 //! use nix::sys::wait::*;
+//! use nix::unistd::Pid;
+//!
 //! let pid = Pid::from_raw(17563);
 //! loop {
 //!     match waitpid(PidGroup::ProcessGroupID(pid), WUNTRACED) {
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -7,6 +7,7 @@
 //!
 //! ```rust
 //! use nix::sys::wait::*;
+//! let pid = Pid::from_raw(17563);
 //! loop {
 //!     match waitpid(PidGroup::ProcessGroupID(pid), WUNTRACED) {
 //!         Ok(WaitStatus::Exited(pid, status)) => {
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -14,7 +14,7 @@
 //!             break
 //!         },
 //!         Ok(WaitStatus::Stopped(pid, signal)) => {
-//!             println!("Process '{}' stopped with signal '{}'", pid, signal);
+//!             println!("Process '{}' stopped with signal '{:?}'", pid, signal);
 //!         },
 //!         Ok(WaitStatus::Continued(pid)) => {
 //!             println!("Process '{}' continued", pid);
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -48,15 +48,20 @@ libc_bitflags!(
         /// process
         WUNTRACED,
         /// Waits for children that have terminated.
-        #[cfg(any(target_os = "android", target_os = "linux"))]
+        #[cfg(any(target_os = "android",
+                  target_os = "freebsd",
+                  target_os = "linux",
+                  target_os = "netbsd"))]
         WEXITED,
         /// Report the status of any continued child process specified by pid whose status has not
         /// been reported since it continued from a job control stop.
-        #[cfg(any(target_os = "android", target_os = "linux"))]
         WCONTINUED,
         /// Leave the child in a waitable state; a later wait call can be used to again retrieve
         /// the child status information.
-        #[cfg(any(target_os = "android", target_os = "linux"))]
+        #[cfg(any(target_os = "android",
+                  target_os = "freebsd",
+                  target_os = "linux",
+                  target_os = "netbsd"))]
         WNOWAIT,
         /// Don't wait on children of other threads in this group
         #[cfg(any(target_os = "android", target_os = "linux"))]
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -1,3 +1,5 @@
+//! This module contains the `wait()` and `waitpid()` functions, which are used
+
 use libc::{self, c_int};
 use {Errno, Result};
 use unistd::Pid;
@@ -14,40 +16,33 @@ libc_bitflags!(
         /// has not yet been reported since they stopped, shall also be reported to the requesting
         /// process
         WUNTRACED,
-        #[cfg(any(target_os = "linux",
-                  target_os = "android"))]
         /// Waits for children that have terminated.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         WEXITED,
-        #[cfg(any(target_os = "linux",
-                  target_os = "android"))]
         /// Report the status of any continued child process specified by pid whose status has not
         /// been reported since it continued from a job control stop.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         WCONTINUED,
-        #[cfg(any(target_os = "linux",
-                  target_os = "android"))]
         /// Leave the child in a waitable state; a later wait call can be used to again retrieve
         /// the child status information.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         WNOWAIT,
-        #[cfg(any(target_os = "linux",
-                  target_os = "android"))]
         /// Don't wait on children of other threads in this group
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         __WNOTHREAD,
-        #[cfg(any(target_os = "linux",
-                  target_os = "android"))]
         /// Wait for all children, regardless of type (clone or non-clone)
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         __WALL,
-        #[cfg(any(target_os = "linux",
-                  target_os = "android"))]
         /// Wait for "clone" children only. If omitted then wait for "non-clone" children only.
         /// (A "clone" child is one which delivers no signal, or a signal other than `SIGCHLD` to
         /// its parent upon termination.) This option is ignored if `__WALL` is also specified.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         __WCLONE,
 
     }
 );
 
-#[cfg(any(target_os = "linux",
-          target_os = "android"))]
+#[cfg(any(target_os = "linux",target_os = "android"))]
 const WSTOPPED: WaitPidFlag = WUNTRACED;
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug)]
@@ -61,18 +56,17 @@ pub enum WaitStatus {
     /// Signifies that the process was stopped by a signal, providing the PID and associated
     /// signal.
     Stopped(Pid, Signal),
-    #[cfg(any(target_os = "linux", target_os = "android"))]
     /// Signifies that the process was stopped due to a ptrace event, providing the PID, the
     /// associated signal, and an integer that represents the status of the event.
+    #[cfg(any(target_os = "linux", target_os = "android"))]
     PtraceEvent(Pid, Signal, c_int),
     /// Signifies that the process received a `SIGCONT` signal, and thus continued.
     Continued(Pid),
-    /// if `WNOHANG` was set, this value is returned when no children have changed state.
+    /// If `WNOHANG` was set, this value is returned when no children have changed state.
     StillAlive
 }
 
-#[cfg(any(target_os = "linux",
-          target_os = "android"))]
+#[cfg(any(target_os = "linux", target_os = "android"))]
 mod status {
     use sys::signal::Signal;
     use libc::c_int;
@@ -114,8 +108,7 @@ mod status {
     }
 }
 
-#[cfg(any(target_os = "macos",
-          target_os = "ios"))]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
 mod status {
     use sys::signal::{Signal,SIGCONT};
 
@@ -236,9 +229,9 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 /// Designates whether the supplied `Pid` value is a process ID, process group ID,
 /// specifies any child of the current process's group ID, or any child of the current process.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub enum PidGroup {
     ProcessID(u32),
     ProcessGroupID(u32),
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -4,48 +4,41 @@ use unistd::Pid;
 
 use sys::signal::Signal;
 
-mod ffi {
-    use libc::{pid_t, c_int};
-
-    extern {
-        pub fn waitpid(pid: pid_t, status: *mut c_int, options: c_int) -> pid_t;
-    }
-}
-
-#[cfg(not(any(target_os = "linux",
-              target_os = "android")))]
 libc_bitflags!(
+    /// Defines optional flags for the `waitpid` function.
     pub flags WaitPidFlag: c_int {
         /// Returns immediately if no child has exited
         WNOHANG,
         /// Returns if a child has been stopped, but isn't being traced by ptrace.
         WUNTRACED,
-    }
-);
-
-#[cfg(any(target_os = "linux",
-          target_os = "android"))]
-libc_bitflags!(
-    pub flags WaitPidFlag: c_int {
-        /// Returns immediately if no child has exited
-        WNOHANG,
-        /// Returns if a child has been stopped, but isn't being traced by ptrace.
-        WUNTRACED,
+        #[cfg(any(target_os = "linux",
+                  target_os = "android"))]
         /// Waits for children that have terminated.
         WEXITED,
+        #[cfg(any(target_os = "linux",
+                  target_os = "android"))]
         /// Waits for previously stopped children that have been resumed with `SIGCONT`.
         WCONTINUED,
+        #[cfg(any(target_os = "linux",
+                  target_os = "android"))]
         /// Leave the child in a waitable state; a later wait call can be used to again retrieve
         /// the child status information.
         WNOWAIT,
+        #[cfg(any(target_os = "linux",
+                  target_os = "android"))]
         /// Don't wait on children of other threads in this group
         __WNOTHREAD,
+        #[cfg(any(target_os = "linux",
+                  target_os = "android"))]
         /// Wait for all children, regardless of type (clone or non-clone)
         __WALL,
+        #[cfg(any(target_os = "linux",
+                  target_os = "android"))]
         /// Wait for "clone" children only. If omitted then wait for "non-clone" children only.
         /// (A "clone" child is one which delivers no signal, or a signal other than `SIGCHLD` to
         /// its parent upon termination.) This option is ignored if `__WALL` is also specified.
         __WCLONE,
+
     }
 );
 
@@ -54,6 +47,7 @@ libc_bitflags!(
 const WSTOPPED: WaitPidFlag = WUNTRACED;
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug)]
+/// Contains the status returned by the `wait` and `waitpid` functions.
 pub enum WaitStatus {
     /// Signifies that the process has exited, providing the PID and associated exit status.
     Exited(Pid, i8),
@@ -65,6 +59,7 @@ pub enum WaitStatus {
     PtraceEvent(Pid, Signal, c_int),
     /// Signifies that the process received a `SIGCONT` signal, and thus continued.
     Continued(Pid),
+    /// if `WNOHANG` was set, this value is returned when no children have changed state.
     StillAlive
 }
 
@@ -232,6 +227,55 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+/// Designates whether the supplied `Pid` value is a process ID, process group ID,
+/// specifies any child of the current process's group ID, or any child of the current process.
+pub enum PidGroup {
+    ProcessID(u32),
+    ProcessGroupID(u32),
+    AnyGroupChild,
+    AnyChild,
+}
+
+impl From<Pid> for PidGroup {
+    fn from(pid: Pid) -> PidGroup {
+        if pid > Pid::from_raw(0) {
+            PidGroup::ProcessID(i32::from(pid) as u32)
+        } else if pid < Pid::from_raw(-1) {
+            PidGroup::ProcessGroupID(-i32::from(pid) as u32)
+        } else if pid == Pid::from_raw(0) {
+            PidGroup::AnyGroupChild
+        } else {
+            PidGroup::AnyChild
+        }
+    }
+}
+
+impl From<i32> for PidGroup {
+    fn from(pid: i32) -> PidGroup {
+        if pid > 0 {
+            PidGroup::ProcessID(pid as u32)
+        } else if pid < -1 {
+            PidGroup::ProcessGroupID(-pid as u32)
+        } else if pid == 0 {
+            PidGroup::AnyGroupChild
+        } else {
+            PidGroup::AnyChild
+        }
+    }
+}
+
+impl Into<i32> for PidGroup {
+    fn into(self) -> i32 {
+        match self {
+            PidGroup::ProcessID(pid)      => pid as i32,
+            PidGroup::ProcessGroupID(pid) => -(pid as i32),
+            PidGroup::AnyGroupChild       => 0,
+            PidGroup::AnyChild            => -1
+        }
+    }
+}
+
 /// Waits for and returns events that are received from the given supplied process or process group
 /// ID, and associated options.
 ///
@@ -254,31 +298,23 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
 ///     `SIG_IGN`.
 /// - **EINTR**: `WNOHANG` was not set and either an unblocked signal or a `SIGCHLD` was caught.
 /// - **EINVAL**: The supplied options were invalid.
-pub fn waitpid<P,O>(pid: P, options: O) -> Result<WaitStatus>
-    where P: Into<Option<Pid>>,
-          O: Into<Option<WaitPidFlag>>
+pub fn waitpid<O>(pid: PidGroup, options: O) -> Result<WaitStatus>
+    where O: Into<Option<WaitPidFlag>>
 {
     use self::WaitStatus::*;
 
-    let mut status: i32 = 0;
+    let mut status = 0;
+    let options = options.into().map_or(0, |o| o.bits());
 
-    let option_bits = options.into().map_or(0, |bits| bits.bits());
-
-    let res = unsafe {
-        ffi::waitpid(pid.into().unwrap_or (
-            Pid::from_raw(-1)).into(),
-            &mut status as *mut c_int,
-            option_bits
-        )
-    };
+    let res = unsafe { libc::waitpid(pid.into(), &mut status as *mut c_int, options) };
 
     Errno::result(res).map(|res| match res {
-        0 => StillAlive,
+        0   => StillAlive,
         res => decode(Pid::from_raw(res), status),
     })
 }
 
-/// Equivalent to `waitpid(-1, None)`, which waits on any child process of the current process.
+/// Waits on any child of the current process.
 pub fn wait() -> Result<WaitStatus> {
-    waitpid(None, None)
+    waitpid(PidGroup::AnyChild, None)
 }
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -1,4 +1,32 @@
-//! This module contains the `wait()` and `waitpid()` functions, which are used
+//! This module contains the `wait()` and `waitpid()` functions, which are used to wait on and
+//! obtain status information from child processes. These provide more granular control over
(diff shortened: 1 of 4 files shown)


src/sys/wait.rs
```diff
@@ -215,24 +232,53 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+    {
     use self::WaitStatus::*;
 
     let mut status: i32 = 0;
 
-    let option_bits = match options {
-        Some(bits) => bits.bits(),
-        None => 0
-    };
+    let option_bits = options.into().map_or(0, |bits| bits.bits());
```
Susurrus: `bits.bits()` reads weird, can you just shorten it to `b.bits()`?

src/sys/wait.rs
```diff
@@ -215,24 +232,53 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
 
-    let res = unsafe { ffi::waitpid(pid.into().unwrap_or(Pid::from_raw(-1)).into(), &mut status as *mut c_int, option_bits) };
+    let res = unsafe {
+        ffi::waitpid(pid.into().unwrap_or (
```
Susurrus: No space before the last `(`

src/sys/wait.rs
```diff
@@ -215,24 +232,53 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+    };
 
-    Ok(match try!(Errno::result(res)) {
+    Errno::result(res).map(|res| match res {
         0 => StillAlive,
         res => decode(Pid::from_raw(res), status),
     })
 }
 
+/// Equivalent to `waitpid(-1, None)`, which waits on any child process of the current process.
```
Susurrus: This is confusing because it's actually equivalent to `waitpid(None, None)`.

src/sys/wait.rs
```diff
@@ -42,11 +55,15 @@ const WSTOPPED: WaitPidFlag = WUNTRACED;
...
     Exited(Pid, i8),
+    /// Signifies that the process was killed by a signal, providing the PID and associated signal.
     Signaled(Pid, Signal, bool),
+    /// Signifies that the process was stopped by a signal, providing the PID and associated signal.
     Stopped(Pid, Signal),
     #[cfg(any(target_os = "linux", target_os = "android"))]
     PtraceEvent(Pid, Signal, c_int),
+    /// Signifies that the process received a `SIGCONT` signal, and thus continued.
     Continued(Pid),
     StillAlive
```
Susurrus: I'd like to get some doc comments on `StillAlive` and `PtraceEvent` since we're here. Are you stuck on understanding those or did you just forget to add them?

src/sys/wait.rs
```diff
@@ -42,11 +55,15 @@ const WSTOPPED: WaitPidFlag = WUNTRACED;
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug)]
```
Susurrus: It'd be good to have a comment for the entire enum.

src/sys/wait.rs
```diff
@@ -25,13 +27,24 @@ libc_bitflags!(
           target_os = "android"))]
```
Susurrus: Rather than duplicating this `libc_bitflags`, could you condense these into one? Before each item you can put the proper `#[cfg` so that it still compiles properly. This way we aren't duplicating declarations and comments.

src/sys/wait.rs
```diff
@@ -25,13 +27,24 @@ libc_bitflags!(
...
 libc_bitflags!(
```
Susurrus: I'd like to see a doccomment for `WaitPidFlag` here since you've already documented (nearly) every item.

src/sys/wait.rs
mmstick: I couldn't find any mention of `StillAlive` or `PtraceEvents` mentioned inthe documentation for `waitpid`, so I'm not sure what they do.

src/sys/wait.rs
mmstick: The `None` actually converts to `-1` inside the function, so it's better to state exactly what `None` means, especially given that the original C function doesn't support algebraic data types like `None`.

src/sys/wait.rs
```diff
@@ -215,24 +232,53 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+///
+/// # Usage Notes
+///
+/// - If the value of PID is greater than `0`, it will wait on the child with that PID.
+/// - If the value of the PID is less than `-1`, it will wait on any child that
+///   belongs to the process group with the absolute value of the supplied PID.
+/// - If the value of the PID is `0`, it will wait on any child process that has the same
+/// group ID as the current process.
+/// - If the value of the PID is `-1`, it will wait on any child process of the current process.
+/// - If the value of the PID is `None`, the value of PID is set to `-1`.
```
Susurrus: Combine this line with the above. But then, what's the point of making `Pid` and Option type rather than leaving it numeric if there are two `None` cases? In reality, this should take an enum of these different groups, which are then converted to a numeric `pid_t` internally. Like:

```
pub enum PidGroup {
    ChildPid(Pid),
    ProcessGroupChild(Pid),
    GroupChild,
    CurrentProcessChild,
}
```

Then the user uses one of those to specify the option. What do you think about that?

Susurrus:
Awesome, great work starting this! We always could use more docs, so thanks for the PR!

I've got a few general comments beyond what I left about your code:
  * Could you add a module-level comment at the top? These start with `//!` and have a one-line summary at the beginning and then a space and a more detailed summary below as necessary.
  * Could you remove the `ffi` module and instead use `libc::waitpid()`
  * Could you add to the CHANGELOG under the `Changed` section that the function arguments to `waitpid` have changed (see my comment about the function arguments before doing this).

Susurrus:
@asomers Another FreeBSD build that failed, and the merge branch here is also `master`, so I think that's definitely the problem.

src/sys/wait.rs
mmstick: I think that'd be a much better option, although it would require a version bump due to breaking the existing API. I think the best would be to have the parameter value as `Into<PidGroup>`, where `PidGroup` could also implement `From<i32>`, so that we could continue using `i32` values as in C in addition.

Susurrus: We already have breaking changes in this release, and I'm not afraid to have breaking changes when it results in improved APIs. The 0.9 release that will come out soon will have breaking changes, and I don't see that as a problem as long as they're well documented.

asomers:
@Susurrus I opened an issue for the master branch thing: #652 .  I think it should be fixed now; we'll see as soon as mmstick pushes again.

src/sys/wait.rs
Susurrus: It's still confusing because the code doesn't match the comment until you read it a little more thoroughly. I'm a fan of the "principle of least surprise": I was surprised to read that comment and then see that the code didn't match that, though I later realized they're equivalent. I think it'd be better here to just remove the `waitpid...` part and just have a description of what it does. A more descriptive comment below could explain that it's equivalent, but it's definitely not necessary on the title line.

mmstick:
I'll make a push within 30 mins with all the changes. What do you think though? Should the `pid` parameter of `waitpid` be `Into<pid_t>`, where either a `pid_t` or `PidGroup` value could be supplied; or `Into<PidGroup>`, that converts a given `pid_t` into a `PidGroup`, or takes a `PidGroup`, and then the function turns the `PidGroup` back into a `pid_t` internally?

src/sys/wait.rs
```diff
@@ -42,11 +55,15 @@ const WSTOPPED: WaitPidFlag = WUNTRACED;
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug)]
 pub enum WaitStatus {
+    /// Signifies that the process has exited, providing the PID and associated exit status.
     Exited(Pid, i8),
+    /// Signifies that the process was killed by a signal, providing the PID and associated signal.
```
asomers: Could you also please document the bool member?

mmstick: Once I figure out what the `bool` parameter signifies.

Susurrus:
No rush on the changes, whenever you've got time is fine. I'm just at a computer, so you're getting nice quick responses! :-)

As to your question, it might take some test code that plays with the API to see what's most ergonomic, but I'm inclined to think taking a `PidGroup` and converting it internally makes the most sense. There's no need for `Into` if we use that enum. Since `PidGroup` group can wrap `Pid` types, there's no need for `Into<PidGroup>`.

src/sys/wait.rs
```diff
@@ -41,13 +51,23 @@ libc_bitflags!(
...
+    /// Signifies that the process was stopped by a signal, providing the PID and associated
+    /// signal.
     Stopped(Pid, Signal),
     #[cfg(any(target_os = "linux", target_os = "android"))]
+    /// Signifies that the process was stopped due to a ptrace event, providing the PID, the
+    /// associated signal, and an integer that represents the status of the event.
     PtraceEvent(Pid, Signal, c_int),
+    /// Signifies that the process received a `SIGCONT` signal, and thus continued.
     Continued(Pid),
+    /// if `WNOHANG` was set, this value is returned when no children have changed state.
```
Susurrus: Please capitalize `if` here.

src/sys/wait.rs
```diff
@@ -215,24 +236,94 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
-pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
-    use self::WaitStatus::*;
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+/// Designates whether the supplied `Pid` value is a process ID, process group ID,
+/// specifies any child of the current process's group ID, or any child of the current process.
+pub enum PidGroup {
+    ProcessID(u32),
```
Susurrus: I imagine this would be a lot cleaner if you used the `Pid` type here directly instead of `u32`. Any reason you didn't?

src/sys/wait.rs
```diff
@@ -215,24 +236,94 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
-pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
-    use self::WaitStatus::*;
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+/// Designates whether the supplied `Pid` value is a process ID, process group ID,
```
Susurrus: Please put the comments above `#cfg` attributes.

src/sys/wait.rs
```diff
@@ -4,35 +4,45 @@ use unistd::Pid;
...
+    /// Defines optional flags for the `waitpid` function.
     pub flags WaitPidFlag: c_int {
+        /// Do not suspend execution of the calling thread if the status is not immediately
+        /// available for one of the child processes specified by pid.
         WNOHANG,
+        /// The status of any child processes specified by pid that are stopped, and whose status
+        /// has not yet been reported since they stopped, shall also be reported to the requesting
+        /// process
         WUNTRACED,
+        #[cfg(any(target_os = "linux",
```
Susurrus: Please put `cfg` statements on one line if they're under 100 characters.

src/sys/wait.rs
```diff
@@ -215,24 +236,94 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+/// Designates whether the supplied `Pid` value is a process ID, process group ID,
+/// specifies any child of the current process's group ID, or any child of the current process.
+pub enum PidGroup {
+    ProcessID(u32),
+    ProcessGroupID(u32),
+    AnyGroupChild,
+    AnyChild,
+}
+
+impl From<Pid> for PidGroup {
```
Susurrus: The `From`s for `PidGroup` should be removed in favor on embedded `Pid`s directly into the specific enum types.

Susurrus:
Looking pretty good, the tests didn't get updated. Please fix those on your next update.

src/sys/wait.rs
mmstick: It's kind of odd to use something other than `u32`, as the purpose of separating the variants is to eliminate the need to express meaning through negative values, as `Pid` is basically `i32`. A negative value indicates a process group ID, where as a positive value indicates a process ID. If the process ID and group ID are going to get their own variants, it wouldn't make sense to require them to retain their positive and negative values. If you retained the `Pid` value, then there'd be less meaning to having both a PID and PGID variant when they'd always get the same value and that value would be treated the same by functions.

Especially when I would, for example, create a `PidGroup` manually in Ion by just supplying my existing `u32(Linux/BSD)/usize(Redox)` values that I have stored in the shell to the `ProcessGroupID` variant, rather than having to do as I do now and convert the `u32` into an `i32`, and then set the negative bit to signify to the function that I'm wanting to wait/signal a process group. Being able to just designate the PGID with a positive value is more idiomatic.

Susurrus: I'm not certain we're quite on the same page with understanding this issue based on your comment here: I'm trying to develop a solution where you never need to negate the `pid` as well. My understanding that when using `waitpid()` that you will already have a process ID, the question is what does it mean. That extra meaning is why there's the enum type, so even though `ProcessID(pid)` and `ProcessGroupId(pid)` have the same process id within them, they mean different things. And internally the values would be mapped as the `libc::waitpid` function expected. So the point is that the user shouldn't be doing the negation just as you suggested, the value used for `ProcessId` and `ProcessGroupId` is the same raw pid value.

What we do need to make sure is that the raw value is "correct". For example, the `pid` should be positive and non-zero for both of those types, so that should be checked. We can either embed this into the enum type (which I don't know a good implementation for) or embed it within `waitpid()`, which is the path I'd suggest.

Is that more clear? Because I think we agree we're just losing something in translation.

mmstick: It seems, then, that it'd be better if `Pid` was mapped to an unsigned integer rather than a signed integer that can be constructed from a negative value, if we need to guarantee that a negative value cannot be mapped. Otherwise, there'd be a cost in having to always check if the value of `ProcessGroupID` is negative, and then setting the negative bit if it's positive, else use the value as is. If it was an unsigned integer, then you'd always know that the PGID variant just needs to be converted into a signed integer and have it's negative bit set before passing it to the function. Process IDs are never negative values to begin with.

Susurrus: `pid_t` is the underlying type and that's defined by the platform, we don't change that. And yes, there will be a runtime cost of checking these values, but we'd need to check them even if they're unsigned since `0` doesn't make sense for these enum types either.

mmstick:
I believe that I have all concerns addressed now.

src/sys/wait.rs
```diff
@@ -267,13 +253,14 @@ impl From<i32> for PidGroup {
     }
 }
 
-impl Into<i32> for PidGroup {
-    fn into(self) -> i32 {
+impl Into<pid_t> for PidGroup {
```
Susurrus: This is a public implementation that I don't think we want to support. While the first argument of `waitpid` is a `pid_t` type, it's actually not a "proper" one. As in it's different than the `pid_t` you'd get or pass to many other functions, so it's in essence a different type (it's just overloaded in C because of language limitations). I think instead we want to just hardcode this inside of `waitpid` or implement it as a private function (so it can be unit tested) instead.

src/sys/wait.rs
```diff
@@ -267,13 +253,14 @@ impl From<i32> for PidGroup {
...
+    fn into(self) -> pid_t {
         match self {
-            PidGroup::ProcessID(pid)      => pid as i32,
-            PidGroup::ProcessGroupID(pid) => -(pid as i32),
-            PidGroup::AnyGroupChild       => 0,
-            PidGroup::AnyChild            => -1
+            PidGroup::ProcessID(pid) if pid < Pid::from_raw(-1)      => -pid_t::from(pid),
```
Susurrus: Please don't align the `=>` in match arms, just have them surrounded by one space on each size.

src/sys/wait.rs
```diff
@@ -233,32 +233,18 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
 /// specifies any child of the current process's group ID, or any child of the current process.
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
```
Susurrus: Can you alphabetize the derived traits?

src/sys/wait.rs
```diff
@@ -215,24 +229,77 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+        }
+    }
+}
+
+/// Waits for and returns events that are received from the given supplied process or process group
+/// ID, and associated options.
+///
+/// # Usage Notes
+///
+/// - If the value of PID is greater than `0`, it will wait on the child with that PID.
```
Susurrus: Can you update these docs now that the pid argument type has changed?

src/sys/wait.rs
```diff
@@ -215,24 +229,77 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
-pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
+/// Designates whether the supplied `Pid` value is a process ID, process group ID,
+/// specifies any child of the current process's group ID, or any child of the current process.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
```
Susurrus: Can you alphabetize these?

src/sys/wait.rs
```diff
@@ -1,58 +1,72 @@
...
     PtraceEvent(Pid, Signal, c_int),
+    /// Signifies that the process received a `SIGCONT` signal, and thus continued.
     Continued(Pid),
+    /// If `WNOHANG` was set, this value is returned when no children have changed state.
     StillAlive
 }
 
-#[cfg(any(target_os = "linux",
-          target_os = "android"))]
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize the list of oses here and everywhere else `cfg` is used within the scope of this PR.

src/sys/wait.rs
```diff
@@ -1,58 +1,72 @@
-use libc::{self, c_int};
+//! This module contains the `wait()` and `waitpid()` functions, which are used
```
Susurrus: I think something got left out here, it's an incomplete sentence.

src/sys/wait.rs
```diff
@@ -1,58 +1,100 @@
...
+        #[cfg(any(target_os = "android", target_os = "linux"))]
         __WCLONE,
+
     }
 );
 
-#[cfg(any(target_os = "linux",
-          target_os = "android"))]
+#[cfg(any(target_os = "android", target_os = "linux"))]
 const WSTOPPED: WaitPidFlag = WUNTRACED;
```
Susurrus: This should actually be included in the `WaitPidFlag` bitflag declaration, yes?

mmstick: Can it be? The other flags are automatically generated via the `libc_bitflags!` macro.

Susurrus: We don't want to be defining `WSTOPPED` in `nix`, instead it should be defined in the `libc` crate and we include it on appropriate platforms. It isn't defined in `libc` for `android` or `linux` however, so we'll need a PR for that.

Susurrus:
@mmstick For fixing documentation errors, please test these locally first. Tests take 2 hours to run for us and you're pushing so many commits that we have a huge backlog in CI. I've already canceled a bunch of your old pushes because of it.

mmstick:
That last commit fixed everything.

src/sys/wait.rs
```diff
@@ -1,58 +1,100 @@
...
+        /// has not yet been reported since they stopped, shall also be reported to the requesting
+        /// process
         WUNTRACED,
+        /// Waits for children that have terminated.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
         WEXITED,
+        /// Report the status of any continued child process specified by pid whose status has not
+        /// been reported since it continued from a job control stop.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
         WCONTINUED,
```
myfreeweb: WCONTINUED is available pretty much everywhere. Definitely on `freebsd`, `openbsd`, `bitrig`, `dragonfly`, `netbsd` and `solaris`.

src/sys/wait.rs
```diff
@@ -1,58 +1,100 @@
...
-        WNOWAIT, // Don't reap, just poll status.
-        __WNOTHREAD, // Don't wait on children of other threads in this group
-        __WALL, // Wait on all children, regardless of type
+        /// Leave the child in a waitable state; a later wait call can be used to again retrieve
+        /// the child status information.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        WNOWAIT,
```
myfreeweb: WNOWAIT is available on `freebsd` and `netbsd`.

src/sys/wait.rs
```diff
@@ -1,58 +1,100 @@
...
+        /// Do not suspend execution of the calling thread if the status is not immediately
+        /// available for one of the child processes specified by pid.
         WNOHANG,
+        /// The status of any child processes specified by pid that are stopped, and whose status
+        /// has not yet been reported since they stopped, shall also be reported to the requesting
+        /// process
         WUNTRACED,
+        /// Waits for children that have terminated.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
         WEXITED,
```
myfreeweb: WEXITED is available on `freebsd` and `netbsd`.

src/sys/wait.rs
myfreeweb: (Does `nix` even support `solaris`? it's [waitid](https://illumos.org/man/2/waitid) on illumos instead of `waitpid`…)

myfreeweb: WCONTINUED is not mentioned in the macOS manpage, someone with macOS should check there…

Susurrus: It's defined for Mac within `libc`, so it should be fine on both `macos` and `ios`

myfreeweb: Nice. Probably should just drop the `cfg` here

Susurrus:
This also touches on changes made in #566. I think I'm going to try to get that PR merged first and then this can be rebased afterwards.

src/sys/wait.rs
```diff
@@ -215,24 +262,84 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+/// - If the value of the PID is `PidGroup::ProcessID(Pid)`, it will wait on the child
+///   with that has the specified process ID.
+/// - If the value of the PID is `PidGroup::ProcessGroupID(Pid)`, it will wait on any child that
+///   belongs to the specified process group ID.
+/// - If the value of the PID is `PidGroup::AnyGroupChild`, it will wait on any child process
+///   that has the same group ID as the current process.
+/// - If the value of the PID is `PidGroup::AnyChild`, it will wait on any child process of the
+///   current process.
+///
+/// # Possible Error Values
```
Susurrus: This should just be `# Errors` (see the [Rust special documentation section](https://doc.rust-lang.org/stable/book/first-edition/documentation.html#special-sections))

src/sys/wait.rs
```diff
@@ -215,24 +262,84 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+            PidGroup::ProcessGroupID(Pid::from_raw(pid))
+        } else if pid == 0 {
+            PidGroup::AnyGroupChild
+        } else {
+            PidGroup::AnyChild
+        }
+    }
+}
+
+/// Waits for and returns events that are received from the given supplied process or process group
```
Susurrus: This should be a short descriptive line, then a blank line, then more description. If you run `rust doc --no-deps --open` and look at the generated output, you'll see why. The first line is used on the summary page, and the rest is shown on the detailed page.

Additionally this should link to the [man page](http://pubs.opengroup.org/onlinepubs/007908799/xsh/wait.html) (see how other functions are documented).

src/sys/wait.rs
```diff
@@ -215,24 +262,84 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
-pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
+/// Designates whether the supplied `Pid` value is a process ID, process group ID,
```
Susurrus: Same here with the documentation.

src/sys/wait.rs
```diff
@@ -1,58 +1,105 @@
-use libc::{self, c_int};
+//! This module contains the `wait()` and `waitpid()` functions, which are used to wait on and
+//! obtain status information from child processes. These provide more granular control over
+//! child management than the primitives provided by Rust's standard library, and are critical
+//! in the creation of shells and managing jobs on *nix platforms.
+//!
+//! # Example
```
Susurrus: This should be `# Examples` in line with the [special sections](https://doc.rust-lang.org/stable/book/first-edition/documentation.html#special-sections) syntax

src/sys/wait.rs
```diff
@@ -1,58 +1,105 @@
-use libc::{self, c_int};
+//! This module contains the `wait()` and `waitpid()` functions, which are used to wait on and
```
Susurrus: Short descriptive first line, then a blank line, then more description. Generate the docs locally with `rust doc --no-deps --open` to see what I'm talking about.

src/sys/wait.rs
```diff
@@ -215,24 +262,84 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+/// - If the value of the PID is `PidGroup::ProcessGroupID(Pid)`, it will wait on any child that
+///   belongs to the specified process group ID.
+/// - If the value of the PID is `PidGroup::AnyGroupChild`, it will wait on any child process
+///   that has the same group ID as the current process.
+/// - If the value of the PID is `PidGroup::AnyChild`, it will wait on any child process of the
+///   current process.
+///
+/// # Possible Error Values
+///
+/// If this function returns an error, the error value will be one of the following:
```
Susurrus: You can just remove this sentence as it's superfluous.

src/sys/wait.rs
```diff
@@ -215,24 +262,84 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+    let res = unsafe { libc::waitpid(pid.into(), &mut status as *mut c_int, options) };
 
-    Ok(match try!(Errno::result(res)) {
+    Errno::result(res).map(|res| match res {
         0 => StillAlive,
         res => decode(Pid::from_raw(res), status),
     })
 }
 
+/// Waits on any child of the current process, returning on events that change the status of
```
Susurrus: Same thing here about a short descriptive first sentence. Also, it should provide a link to the [man page](http://pubs.opengroup.org/onlinepubs/007908799/xsh/wait.html) (see how other functions are documented). Also it looks like the possible errors are a subset of `waitpid`'s, so it'd be good to explicitly document them here.

Susurrus:
Looks like this also needs a rebase.

I'd love to move forward with this and let #566 catch up later as it's currently stalled on the OP.

mmstick:
Just performed a `git pull --rebase upstream master`.

mmstick:
Performed requested changes to the documentation.

src/sys/wait.rs
```diff
@@ -1,38 +1,83 @@
-use libc::{self, c_int};
+//! This module contains the `wait()` and `waitpid()` functions.
+//!
+//! These are used to wait on and obtain status information from child processes, which provide
+//! more granular control over child management than the primitives provided by Rust's standard
+//! library, and are critical in the creation of shells and managing jobs on *nix platforms.
+//!
+//! Manual Page: http://pubs.opengroup.org/onlinepubs/007908799/xsh/wait.html
+//!
+//! # Examples
```
Susurrus: Please provide a brief description of what this code does, either in text before the example or using comments within the code.

src/sys/wait.rs
```diff
@@ -1,38 +1,83 @@
...
+//!
+//! ```rust
```
Susurrus: Add `,no_run` at the end since this code example will fail when run.

src/sys/wait.rs
```diff
@@ -1,38 +1,83 @@
...
+//!         Ok(WaitStatus::Stopped(pid, signal)) => {
+//!             println!("Process '{}' stopped with signal '{:?}'", pid, signal);
+//!         },
+//!         Ok(WaitStatus::Continued(pid)) => {
+//!             println!("Process '{}' continued", pid);
+//!         },
+//!         Ok(_) => (),
+//!         Err(why) => {
+//!             println!("waitpid returned an error code: {}", why);
+//!             break
```
Susurrus: Add a semi-colon after `break`.

src/sys/wait.rs
```diff
@@ -270,24 +313,92 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+    /// Signifies that the `Pid` is supposed to represent a process group ID.
+    ProcessGroupID(Pid),
+    /// Signifies that any child process that belongs to the process group of the current process
+    /// should be selected.
+    AnyGroupChild,
+    /// Signifies that any child process that belongs to the current process should be selected.
+    AnyChild,
+}
+
+impl From<i32> for PidGroup {
```
Susurrus: This should be removed as it's unused and we shouldn't support this for our users.

Susurrus:
@mmstick Any interest in cleaning this up a bit? We're close to merging and we would love to have this documentation in `nix`!

Susurrus:
@mmstick Just pinging you again on this. Do you have time to wrap this up in the near future?

PR #653, opened 2017-07-07
Remove unused administrative code

Susurrus:
All types should come from the libc crate, which has extensive
testing on its types, so this isn't necessary anymore. Additionally
there were no docs on this test suite and it was unmaintained. Remove
it to reduce confusion.


Susurrus:
This test suite was also not run in CI as part of our test suite.

Susurrus:
I also pushed a commit to remove `deploy.sh`, which is unused as far as I could tell. https://docs.rs should be doing all our documentation generation now, so let's rely on them and not have custom scripts.

asomers:
Don't forget to also remove nix-test from the main `Cargo.toml`.  That's why buildbot failed.

Susurrus:
@asomers Thanks! I thought it was weird builbot failed on this. I did search for `nix-test`, but looks like I missed `Cargo.toml`.

Susurrus:
Hmmm, it actually seems like `nix-test` was used for a few things within `src/`. It appears that there was a lot of validating the sizes of datatypes and values of constants. These are things that should be handled by `libc` primarily. I think we can remove all this test code, but I'll need to look into this a bit more.

Susurrus:
This is turning out to be more complicated than I thought, so I'm going to close this as it's not urgent and I won't be finishing it any time soon.

PR #654, opened 2017-07-07
Document the PID argument in waitpid

marmistrz:


Susurrus:
This is already being worked on in #651 and I'd prefer to avoid conflicting PRs here. Can you add comments there if that PR doesn't include documentation that you want to see for `waitpid()`?

src/sys/wait.rs
```diff
@@ -215,6 +215,22 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
     }
 }
 
+/// Counterpart of the POSIX `waitpid` function
```
Susurrus: This should actually describe what `waitpid` does with a brief summary line at the top and maybe an extended summary below it (with a blank comment line in between). Just saying it is a wrapper around `waitpid` isn't very helpful.

Additionally can you link to [the man page](http://man7.org/linux/man-pages/man2/waitpid.2.html) in the extended description here.

src/sys/wait.rs
```diff
@@ -215,6 +215,22 @@ fn decode(pid : Pid, status: i32) -> WaitStatus {
...
+/// It's best to use `nix::unistd::Pid` for passing the PID to this function
```
Susurrus: Isn't this specified by the argument type?

Susurrus:
Let's go ahead and move this forward since #651 looks stalled.

Susurrus:
@marmistrz I wanted to ping you on this. Would you be interested in updating this PR so we can get it merged?

marmistrz:
As in #666, when I have more time.

Susurrus:
@marmistrz Any idea when you'd be able to wrap this up? I'd like to either merge this or close this issue in the near future.

asomers:
@marmistrz looks like this fell off of your radar.  Any interest in finishing it?

Issue #652, opened 2017-07-07
buildbot screws up pull requests from a "master" branch


asomers:
Buildbot is mostly building stuff fine.  But if the source branch of a pull request is named "master", then buildbot will try to build the master branch of the nix repo instead of the source repo.  But it can't find the revision in question, so the build fails.  For example, PR #651, from mmstick/nix:master, led to this broken build: https://alan.ci/buildbot/#/builders/5/builds/229 .

Susurrus:
Is this fixed or is there something more to do here?

asomers:
My buildbot config change seems to have fixed it.

PR #657, opened 2017-07-07
Add arm/musl, mips64, and s390x linux targets to Tier3

Susurrus:
I don't know why these weren't added originally. They should work with `cross`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,8 @@ matrix:
       rust: 1.13.0
     - env: TARGET=arm-unknown-linux-gnueabi
       rust: 1.13.0
-    # - env: TARGET=arm-unknown-linux-musleabi
+    - env: TARGET=arm-unknown-linux-musleabi
+      rust: 1.13.0
     - env: TARGET=armv7-unknown-linux-gnueabihf
       rust: 1.13.0
     - env: TARGET=i686-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


asomers:
LGTM

bors r+

Susurrus:
I don't know if the just released version of libc will work, since I saw a commit that inspired me to do this. Let's merge that first and then we can see if this works.

Susurrus:
Nvm, this doesn't actually make them blocking, so we can merge this.

Bors r+

asomers:
On the rust=beta channel, test_wait hung.  I've seen this on the bench for multithreaded tests, but not for single threaded tests.  I'm going to try again.

bors r+

notriddle:
BuildBot is being unresponsive again...

asomers:
I think that Bors sometimes renames staging.tmp to staging instead of doing a merge commit, and BuildBot doesn't know to build the branch rename.  I've just configured buildbot to build the staging.tmp branch as well, even though the Bors docs advise against it.  We'll see if it starts picking up this change.

bors r+

notriddle:
Nope.

asomers:
Ok, I figured it out.  It's a matter of whether the `repository` ends in `.git`.  When GitHub sends change requests, it leaves off that part.  I introduced this error as part of my fix for issue #652 .

bors r-

asomers:
Try again.

bors r+

notriddle:
By the way, there's really no reason to build staging.tmp. Bors never just moves the commit from staging.tmp to staging; it always uses the Git Data API to take the staging.tmp commit and make one in staging with the same contents but a different commit message.

PR #658, opened 2017-07-07
Prepare for the 0.9.0 release

asomers:
Bump the version to 0.9.0-pre
Update libc dependency to 0.2.25


Susurrus:
We should do this now before anything else merges to make sure we can use the newest libc release.

asomers:
try again

bors r+

PR #660, opened 2017-07-08
Fix double close bugs in test_lseek and test_lseek64

asomers:
std::fs::File closes the underlying file descriptor on Drop, without
checking for errors.  test_lseek and test_lseek64 also manually close
the file descriptor.  That works for single threaded test runs.  But for
multithreaded runs, it causes EBADF errors in other tests.  Fix the
tests by consuming the File with into_raw_fd(), so its drop method will
never be called.

### DIFF
--- a/test/test.rs
+++ b/test/test.rs
@@ -24,6 +24,18 @@ mod test_stat;
 mod test_unistd;
 
 use nixtest::assert_size_of;
+use std::os::unix::io::RawFd;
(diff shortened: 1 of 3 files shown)


Susurrus:
Please squash these commits into one and also in the commit message state that the other fix here was to use `read_exact` and `write_all` to make sure the entire buffer was read/written, which wasn't done before.

Then it's GTM.

PR #661, opened 2017-07-08
Allow doc attributes in ioctl

### LINKED ISSUE #571: Use of ioctl!() in a crate with [warn(missing_docs)]
We'd like to enable `[warn(missing_docs)]` for our crate that uses nix, but when doing so we get this warning:
```
   Compiling devicemapper v0.4.3 (file:///home/agrover/git/stratis/devicemapper-rs)
warning: missing documentation for a function
  --> src/util.rs:11:1
   |
11 | ioctl!(read blkgetsize64 with 0x12, 114; u64);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: lint level defined here
  --> src/lib.rs:43:9
   |
43 | #![warn(missing_docs)]
   |         ^^^^^^^^^^^^
   = note: this error originates in a macro outside of the current crate

    Finished dev [unoptimized + debuginfo] target(s) in 1.58 secs
```
I don't see a way in my nix-using crate to silence this (without disabling the warning, of course!) Thanks.

roblabla:
fixes #571 . Note that this is a breaking change because it also changes 

```
ioctl!(some_name with 12);
```

to

```
ioctl!(bad some_name with 12);
```

This is to work around a bug in the rust compiler whereby rules around matching idents are overly strict. See https://github.com/rust-lang/rust/issues/24189

It doesn't break anything else though.


src/sys/ioctl/mod.rs
```diff
@@ -97,6 +97,16 @@
 //! Most `ioctl`s have no or little documentation. You'll need to scrounge through
 //! the source to figure out what they do and how they should be used.
 //!
+//! How do I document the generated functions?
```
Susurrus: Let's change this to "Documenting generated ioctl functions". I'd also like to request that you add a rationale why this was added (that generated ioctls are public and should have docs and that many crates error on missing documentation).

test/sys/test_ioctl.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(dead_code)]
 
 // Simple tests to ensure macro generated fns compile
-ioctl!(do_bad with 0x1234);
+ioctl!(bad do_bad with 0x1234);
```
Susurrus: We also need a test here that uses a comment with all of the various forms of the `ioctl!` macro given that this is the functionality you're really adding here.

Susurrus:
I'd also like to see this refactored into 2 commits. First fix the `bad` version of the macro and update the tests in one commit. Then change the macros to support doccomments, update the docs, and add tests for the documented version of all forms of the `ioctl!` macro. I prefer to keep commits to a single issue as it's easier to read through version history if you do that. In this case since it will break code, I'd like a nice clear commit that we can point to for someone what shows the exact change clearly and how they can update their code.

Susurrus:
I did a bunch of minor cleanup in `ioctl` while I was there to improve the docs. I'd like to have you rebase this PR once we get that merged so this just adds the comments and tests.

roblabla:
I rebased on top of master, and made some of the requested modifications. Is this good enough ?

Susurrus:
This looks good. Can you squash this into a single commit? Then we can merge it.

roblabla:
Done.

Susurrus:
I missed this on the original PR, but this should also have added a CHANGELOG entry. @roblabla Could you add one to the `Added` section for this?

PR #662, opened 2017-07-09
Fix tier3s

Susurrus:
This handles the low-hanging fruit. rust-lang/libc#654 needs to land and there's a bit of work necessary for the syscall and ioctl errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
     - env: TARGET=arm-unknown-linux-gnueabi
       rust: 1.13.0
     - env: TARGET=arm-unknown-linux-musleabi
-      rust: 1.13.0
+      rust: 1.14.0
     - env: TARGET=armv7-unknown-linux-gnueabihf
       rust: 1.13.0
     - env: TARGET=i686-unknown-linux-gnu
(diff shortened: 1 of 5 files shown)


asomers:
LGTM.  Since these are all Tier 3, there's no harm in merging before 0.9.0, even though we won't have the relevant change to libc.

bors r+

Susurrus:
I'm not quite done with this PR. It's annoying that bors doesn't ignore PRs that start with "WIP". GitLab has the nice feature that it won't let you merge PRs that start with WIP, it'd be nice if bors did that as well.

Susurrus:
I  think this addresses everything that's not libc. Still need to figure out the `arm-unknown-linux-musleabi` problems, but not as part of this PR.

notriddle:
bors-ng can use labels for the WIP thing, if you want to use that. The syntax looks like:

```toml
block-labels = [ "WIP", "DO NOT MERGE YET" ]
```

You can also file an issue to block PRs based on their title.

Susurrus:
@asomers What do you say about adding "WIP" and "DO NOT MERGE" to the `block-labels` setting for `bors`?

asomers:
@Susurrus that sounds like a great idea.

PR #665, opened 2017-07-09
Check tier 3s

Susurrus:
All of the necessary constants should have been merged in upstream libc, so I wanted to file this PR pre-emptively to check that all constants have been accounted for.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ matrix:
       rust: 1.13.0
     - env: TARGET=arm-unknown-linux-gnueabi
       rust: 1.13.0
-    - env: TARGET=arm-unknown-linux-musleabi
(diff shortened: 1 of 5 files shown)


Susurrus:
This adds `mips64*` targets as Tier 1 platforms and `s390x-unknown-linux` and `arm-unknown-linux-musleabi` as Tier 2 platforms. We'll wait until the 0.9 before merging this, however, as I also want to re-arrange and squash some of these commits.

Susurrus:
Cool, well this adds 4 new targets (`mips64*` to Tier 1 and `arm-unknown-linux-musleabi` and `s390x-unknown-linux-gnu` to Tier 2). Now just waiting to merge after the next release.

asomers:
Looks like we're retiring Tier 3.  At least for now.

bors r+

PR #666, opened 2017-07-10
Add more ptrace_* wrappers

marmistrz:
Btw. I think that the `nix::sys::ptrace::ptrace` function should be marked as unsafe. It allows to change arbitrary registers and addresses in a child process - just as dereferencing a pointer is unsafe in Rust.

I was unable to use `libc::user_regs_struct`, the Rust compiler complains that it can't find the members.
By the way, we're missing `PTRACE_GETREGS`.

I tried to use the `user_regs_struct` in the following way:

```
extern crate libc;

fn main() {
    println!("{}", libc::user_regs_struct::rdx);
}
```
which bails out with
```
file: 'file:///home/marcin/proj/user_regs/src/main.rs'
severity: 'Error'
message: 'no associated item named `rdx` found for type `libc::user_regs_struct` in the current scope'
at: '4,22'
source: ''
```


vincenthage:
Hi,
wouldn't it be more useful to have the `Register` enum like this:
```rust
pub enum Register {
    SysArgNum,
    SysArg1,
    SysArg2,
   ...
}
```
and then use `cfg`  to make the link between the enum and the real register, either with a macro or a function:
```rust
#[cfg(all(target_os = "linux", any(target_arch = "x86_64")))]
pub fn get_reg(regs: &user_regs_struct, register: Register) {
    match register {
        SysArgNum => regs.orig_rax,
        SysArg1 => regs.rdi,
        SysArg2 => regs.rsi,
        ...
    }
}
```
See [here ](https://github.com/vincenthage/proot_rust/blob/master/src/register/regs.rs)how to use all this (though I've implemented it like a macro so that it's faster, it can be converted to functions with big matches).

marmistrz:
@vincenthage for me this could be an extra feature. I think that our wrappers should support all registers, since it's what raw `ptrace` supports. I don't want to cripple existing functionality.

How did you manage to get the `user_regs_struct` members recognized?

vincenthage:
I don't understand your question, as `user_regs_struct` is a [structure](https://github.com/rust-lang/libc/blob/472ffbb75fcbb771d59208641846db196295c021/src/unix/notbsd/linux/other/b64/x86_64.rs#L96).

Nix's wrappers should indeed support all registers, but a list of field names, where some of them wouldn't be used depending on the OS, isn't a solution either. 
The register calling conventions are different for every OS (even on a same OS, they can differ by whether the processor is in 32-bits or 64-bits), so in every case, at some point, you will have to have one `cfg` for each OS/configuration.

marmistrz:
@vincenthage see the first post. (*I tried to use...*)

And even if we did as you suggest: what if the user requests the 3rd argument when we have a syscall which only accepts 2? Or uses that after `PTRACE_SINGLESTEP`. This would automatically make this function `unsafe`.

We could have an extra wrapper to answer the question *what is the register is used for storing nth argument*, but I'd treat it as an extra feature. @Susurrus, what do you think?

I'd suggest we left this as it is and focus on another thing: do we declare `pokedata`/`peekdata` as `unsafe`? I'm not fully convinced but I think we should.

vincenthage:
Let's just start again so that we agree on the basics.
Every syscall arguments are like these:
```rust
pub enum Register {
   SysArgNum,
   SysArg1,
   SysArg2,
   SysArg3,
   SysArg4,
   SysArg5,
   SysArg6,
   SysArgResult,
   StackPointer,
   InstrPointer,
   RtldFini,
   StateFlags,
   UserArg1
}
```
Each of them corresponds to a certain field of `user_regs_struct`. Each field of this structure is a memory address, or `NULL` if unused.
For instance, `SysArg1`, the address to the first argument of any syscall, is `rdi` on x86_64 processors, and `ebx` on x86 processors.
`user_regs_struct` itself is different on every type of processors, which makes all this way more difficult. And `libc` currently has defined this structure only for x86_64 and x86. Yay.

Every time the tracee is stopped, you can get the content of an argument by 2 ways (let's say we want the 1st argument):
- by using the `ptrace` syscall to fetch only one of the registers. This is done by giving the corresponding offset of the `user_regs_struct`'s field. This is what you're trying to do, if I understood correctly. So something like `let offset_arg1 = offset(Register, RDI)` and then `ptrace(PEEKUSER, pid, offset_arg1 null_mut())`.
- by using the `ptrace` syscall to fetching all of the regs (`user_regs_struct`) at once, and then going to the field you want. So: `let regs = fetch_regs()`, and then `get_reg(regs, SysArg1)` which will return the value of the field `rdi` in the `user_regs_struct` structure.

Now, if we agree on all this:

> And even if we did as you suggest: what if the user requests the 3rd argument when we have a syscall which only accepts 2?

The `get_reg` function just redirects to one of the fields of ``. If it's not used by the syscall, it will be 0/null, no issue. And worst case, if the user only wants one argument, then they can still use the ABI (`Register` to `user_regs_struct` correspondance) to know which register (and so which offset) to use, and call `ptrace` manually with `PEEKUSER`. But this would be useless, because to get any of the arguments, you need to know which syscall this is. And so you would first need to get `SysArgNum`, and then call it again for the other argument. So, we might as well just get all the regs at the same time, it's faster.

Again, the `Register` class you wrote is redundant with `user_regs_struct`, and only corresponds to the [x86_64 version](https://github.com/rust-lang/libc/blob/472ffbb75fcbb771d59208641846db196295c021/src/unix/notbsd/linux/other/b64/x86_64.rs#L96) of it, so it will only has a meaning for this precise type of processor.  For instance, in the [x86](https://github.com/rust-lang/libc/blob/158d778258124b7def4107f3003c23dafa08c2d8/src/unix/notbsd/linux/other/b32/x86.rs#L50) version of it, `R15` doesn't even exist. I don't think writing a (redundant) wrapper for all the other versions of it is that useful.

Now, it all depends on what you mean by a _high-level_ wrapper. But in my opinion, a high-level wrapper shouldn't involve having the user manipulate offsets to find the arguments of a syscall.

vincenthage:
On a side note, there is this [offset macro](https://github.com/nix-rust/nix/blob/master/src/macros.rs#L422) that might work for your wrap, but it would require `lazy_static` to use it I think.

marmistrz:
All the members of the `user_regs_struct` are longs. Not to mention that my experience was: peeking wrong register means reading garbage which is completely nondeterministic. (I used `PEEKUSER`, no experience with `GETREGS`, but `nix::sys::ptrace` doesn't support `GETREGS` anyway)
Even if `ptrace` returned 0/NULL for all the registers which are not in use, we'd have absolutely no way of finding out if's 0 because that's the arguments - or that it's simply unused.

Moreover, I don't see a `ptrace` routine to check the number of arguments in a syscall. This would probably require some FFI work which I don't know if I have time for.

And your suggested approach with `fetch_regs` is something I won't accept - if I only need the value from 2 registers, why fetch all >= 7 (6 arguments + RAX + ...). As far as I remember assembly on x86_64 - you don't put the argument number into any register either.

The whole point of these *high-level* wrappers is that they wrap the existing, atrocious `ptrace` API. Casting integers to `const* c_void` is something that shouldn't appear in user Rust code. This should be hidden from the programmer - and the current `nix::sys::ptrace::ptrace` is just a 1-1 wrapper when it comes to the input arguments.

My idea is: add a `fn getreg(...)` which would just give the appropriate value of `enum Register` for the nth argument. We should not constrain the possibilites of `ptrace` and `sys::nix::ptrace::ptrace` should eventually get deprecated.

vincenthage:
You win. The lighter the wrap, the better anyway.

About `GETREGS`, it's not that bad actually if you zero the structure beforehand; all the unused arguments are deterministically set at 0. Though I haven't seen any advanced benchmarks comparing zero-ing a structure then call `GETREGS` vs multiple calls of `PEEKUSER`.

marmistrz:
Is this behavior documented? 

Let's get down to the other issue: what about `unsafe`? 
I think that `peekdata`, `pokedata` should definitely be `unsafe`, since they can crash the traced process. What do you think?

I asked a related question on users.rust-lang.org: https://users.rust-lang.org/t/when-should-we-mark-our-functions-unsafe/11834

vincenthage:
All `PtraceRequest` that can crash the tracee should be `unsafe` indeed.

`PtraceRequest` that can write are probably also concerned (simultaneous writes / memory unsafe / tracee crash by giving invalid arguments), so the `poke*` and `set*` families should all be `unsafe`.

It might actually be simpler to make all `ptrace` operations unsafe by default, and then whitelist the ones that we can control (`traceme` for instance).

marmistrz:
Marked that.
The register offsets are not in the `libc` crate. Do we leave them hardcoded as they are for now?

And one more thing about the types. The `sys::nix::ptrace::ptrace` returns `Result<c_long>`. I think the user API should avoid these types. What do you think about returning `Result<u64>` in `peek*`? I think it would be reasonable.

Susurrus:
> The register offsets are not in the libc crate. Do we leave them hardcoded as they are for now?

This won't be merged until we figure out if those constants can be added to libc and get them added if so. We try not to define any FFI values or functions within `nix` directly.

vincenthage:
What do you think of this:
```rust
use std::ptr;
use libc::user_regs_struct;
use nix::sys::ptrace::ptrace::PTRACE_PEEKUSER;

pub type UserRegisters = user_regs_struct;
pub type Word = c_long;

macro_rules! offset_of {
    ($ty:ty, $field:ident) => {
        &(*(0 as *const $ty)).$field as *const _ as usize
    }
}

macro_rules! peekuser {
    ($pid:expr, $reg:ident) => {
        {
            let reg_offset = unsafe {offset_of!(UserRegisters, $reg)} as i32;
            ptrace(PTRACE_PEEKUSER, reg_offset, ptr::null_mut(), ptr::null_mut())
        }
    }
}

```
which can be used like this:
```rust
#[test]
fn test_peekuser() {
    let pid = -1;
    let reg_value = peekuser!(pid, r15);

    assert!(reg_value.is_err());
}
```
whereas using non existing register field is impossible:
```rust
peekuser!(pid, r9999);
```
yield:
```rust
error[E0609]: no field `r9999` on type `libc::user_regs_struct`
   --> src/register/regs.rs:80:64
    |
80  |             let reg_offset = unsafe {offset_of!(UserRegisters, $reg)} as i32;
    |                                                                ^^^^ did you mean `r9`?
...
100 |     let reg_value = peekuser!(pid, r9999);
    |                     --------------------- in this macro invocation

error: aborting due to previous error(s)
```

vincenthage:
Also, @marmistrz, from what I gathered the value of `c_long` can change depending on the processor. It's 8 bytes in 64bits, but 4 in 32bits. If so, we can't replace it by a fixed type like `u64`.

Though, I agree that a light wrapper (like `Word` above) this would be useful.
We could also define it as `struct Word(c_long)` and implemented useful functions for it (`has_value`, `is_null`, `to_ptr`, `to_value`).

marmistrz:
Yeah, I like it! :) This one's cross platform, the API is nice. Should I try adding the offsets to libc nevertheless?
`ptrace` returns a long: `man ptrace`

```
       long ptrace(enum __ptrace_request request, pid_t pid,
                   void *addr, void *data);
```

I'll ask on the Rust forum about the return types, I don't really like the idea of leaking C types to Rust.

vincenthage:
There are safe versions of the `offset_of` macro, like [here](https://stackoverflow.com/questions/41823321/how-to-get-pointer-offset-of-an-enum-member-in-bytes).
And according from the comments it compiles down to constants so it's quick.

About replacing `c_ulong`, wouldn't `usize` do the job? It's equivalent to `u32` on x86 processors and `u64` on 64bits.

vincenthage:
Version 2:
```rust
macro_rules! offset_of {
    ($ty:tt, $field:ident) => {
        {
            let base: $ty = (unsafe { ::std::mem::uninitialized() });
            let offset = &(*(&base as *const $ty)).$field as *const _ as usize;
            ::std::mem::forget(base);
            offset
        }
    }
}
```
This one uses an instance of the structure to calculate the offset.
So instead of doing:
`0.field - 0` equivalent to `0.field`
it does:
`base.field - base`.

marmistrz:
Does `std::mem::uninitialized` does heap or stack allocation? It seems to be optimized: https://godbolt.org/g/i4DqKx
Asked about the C types here: https://users.rust-lang.org/t/leaking-c-types-in-rust-interfaces/11898

marmistrz:
I noticed another problem: depending on the context we may want to put a signed or unsigned value into the register. I'm not so sure if a cast to `usize` will always do what we want. (it could possibly modify the values in some way or panic)

marmistrz:
First of all, I didn't add `PTRACE_GETREGS`. I added other wrappers.
Second of all - do you have an idea why the automated tests fail? On Android `use` of `self::ptrace::*` fails, but I just added an `unsafe` here.

In my new test it complains about `Word` not recognized, but this works on my system.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+    DS = 23 * 8,
+    ES = 24 * 8,
+    FS = 25 * 8,
+    GS = 26 * 8,
+}
+
+type Word = usize;
+
+/// Makes the `PTRACE_SYSCALL` request to ptrace
+pub fn syscall(pid: Pid) -> Result<()> {
```
Susurrus: This should be called `ptrace_syscall`, we're prefixing our helper functions with `ptrace_` to make it more clear. `syscall` is already a POSIX function, so we don't want any confusion here.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+    unsafe {
+        ptrace(ptrace::PTRACE_SYSCALL, pid, ptr::null_mut(), ptr::null_mut()).map(|_| ()) // ignore the useless return value
+    }
+}
+
+/// Makes the `PTRACE_PEEKUSER` request to ptrace
+pub fn peekuser(pid: Pid, reg: Register) -> Result<Word> {
```
Susurrus: Same here, prefix with `ptrace_` like the other wrapper functions.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+    ).map(|_| ()) // ignore the useless return value
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_types() {
```
Susurrus: Please document why this test is necessary.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+
+/// Makes the `PTRACE_PEEKDATA` request to ptrace
+/// 
+/// This function allows to access arbitrary data in the traced process
+/// and may crash the inferior if used incorrectly and is thus marked `unsafe`.
+pub unsafe fn peekdata(pid: Pid, addr: usize) -> Result<Word> {
+    ptrace(ptrace::PTRACE_PEEKDATA, pid, addr as *mut c_void, ptr::null_mut()).map(|r| r as Word)
+}
+
+/// Makes the `PTRACE_PEEKDATA` request to ptrace
```
Susurrus: This comment is wrong.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+///
+/// This function allows to access arbitrary data in the traced process
```
Susurrus: Put this under a `# Safety` section, which is a [special section understood by rustdoc](https://doc.rust-lang.org/stable/book/first-edition/documentation.html#special-sections)

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+/// 
+/// When incorrectly used, may change the registers to bad values,
+/// causing e.g. memory being corrupted by a syscall, thus is marked unsafe
+pub unsafe fn pokeuser(pid: Pid, reg: Register, val: Word) -> Result<()> {
+    let reg_arg = (reg as u64) as *mut c_void;
+    ptrace(ptrace::PTRACE_POKEUSER, pid, reg_arg, val as *mut c_void).map(|_| ()) // ignore the useless return value
+}
+
+/// Sets the process as traceable with `PTRACE_TRACEME`
+pub fn traceme() -> Result<()> {
```
Susurrus: Again, should be `ptrace_traceme`

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+    unsafe {
+        ptrace(
```
Susurrus: Please put first argument on first line and align subsequent arguments with it only if they take up more than 99 columns of space.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
         Err(e) => Err(e),
     }
 }
+
+//-------------------------- Second part: a high-level wrapper for ptrace ----------------------//
```
Susurrus: These comments need to be removed.

src/sys/ptrace.rs
```diff
@@ -140,3 +150,110 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+
+#[cfg(target_arch = "x86_64")]
```
Susurrus: This is why your tests are failing, you need to either limit `peekuser()` to x86_64, since that's the only play this register is defined for, or expand `Register` to more platforms.

Susurrus:
> First of all, I didn't add PTRACE_GETREGS.

You can edit the title as well, feel free to change it to something more appropriate. The previous title didn't help me know which issue this was when reading it, so I changed it.

> do you have an idea why the automated tests fail? On Android use of self::ptrace::* fails, but I just added an unsafe here.

It's as the error says, `Register` is undefined for no-x86_64 targets.

src/sys/ptrace.rs
marmistrz: If it has to, I'll change it. But IMHO what I do now is cleaner. With my approach you just `use nix::sys::ptrace` and call it like `ptrace::syscall()`, with the suggested approach: `use nix::sys::ptrace::{ /* an endless list of imports */ }` or you have to use a star, which is not a great idea in general.

src/sys/ptrace.rs
marmistrz: What's wrong with them? They clearly separate the sections

src/sys/ptrace.rs
Susurrus: That's a good point as these are all namespaced unlike in C. Could you have a separate commit then that renames the existing `ptrace_*` wrappers? That'll have to be documented in the CHANGELOG entry as well.

marmistrz:
> That's a good point as these are all namespaced unlike in C. Could you 
> have a separate commit then that renames the existing |ptrace_*| 
> wrappers? That'll have to be documented in the CHANGELOG entry as well.
>
Should I do it before this one gets merged?

src/sys/ptrace.rs
Susurrus: They're ugly mostly, the code has a sane ordering and these comments are mostly clutter. Generally source files are written so that low level stuff is at the top and high-level stuff is at the bottom. This is a convention from the C days when declarations had to be in order. It's not necessary in Rust, but it's still followed generally.

Susurrus:
Yes, because otherwise it's a consistency issue. You should have 3 commits here, the first is renaming the existing functions, the second is all of your changes, and the third is updating the CHANGELOG.

src/sys/ptrace.rs
marmistrz: I ran `rustfmt` on the code (commit 2a295c6) and it's how rustfmt does it.

src/sys/ptrace.rs
Susurrus: FYI I just meant a separate commit within this PR, not a new PR. But that's fine too.

src/sys/ptrace.rs
```diff
@@ -140,3 +149,124 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
         Err(e) => Err(e),
     }
 }
+
+#[cfg(target_arch = "x86_64")]
+// We're going to export it anyway
```
Susurrus: Comments come before attributes.

src/sys/ptrace.rs
```diff
@@ -68,9 +73,13 @@ mod ffi {
     }
 }
 
-/// Performs a ptrace request. If the request in question is provided by a specialised function
+/// A low-level wrapper for `ptrace`. If available, the higher-level wrappers should be considered instead.
```
Susurrus: This should be a short one-line description with a blank line and then more description below.

src/sys/ptrace.rs
```diff
@@ -140,3 +149,124 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+            ptr::null_mut(),
+        ).map(|_| ()) // ignore the useless return value
+    }
+}
+
+/// Makes the `PTRACE_PEEKDATA` request to ptrace
+///
+/// # Safety
+/// This function allows to access arbitrary data in the traced process
+/// and may crash the inferior if used incorrectly and is thus marked `unsafe`.
```
Susurrus: What is "the inferior" here? That doesn't read right to me.

src/sys/ptrace.rs
```diff
@@ -140,3 +149,124 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+            ptr::null_mut(),
+            ptr::null_mut(),
+        ).map(|_| ()) // ignore the useless return value
+    }
+}
+
+/// Makes the `PTRACE_PEEKDATA` request to ptrace
+///
+/// # Safety
+/// This function allows to access arbitrary data in the traced process
```
Susurrus: Should be "allows for accessing"

src/sys/ptrace.rs
```diff
@@ -140,3 +149,124 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+        pid,
+        addr as *mut c_void,
+        ptr::null_mut(),
+    ).map(|r| r as Word)
+}
+
+/// Makes the `PTRACE_POKEDATA` request to ptrace
+///
+/// # Safety
+/// This function allows to access arbitrary data in the traced process
```
Susurrus: Same grammar nitpicks as above.

src/sys/ptrace.rs
```diff
@@ -140,3 +149,124 @@ pub fn ptrace_setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+/// and may crash the inferior or introduce race conditions if used
+/// incorrectly and is thus marked `unsafe`.
+pub unsafe fn pokedata(pid: Pid, addr: usize, val: Word) -> Result<()> {
```
Susurrus: I'd like to see some tests for these if possible, even if they just return errors, if for no other reason than to test the ergonomics of our wrappers. For example here I don't know how you'd generate `addr` since it's supposed to be a raw pointer type.

marmistrz:
For some reason the doc tests are not run by `cargo test`

marmistrz:
Why do we hard forbid the users use the low-level wrapper? If they need it, why forbid them?

src/sys/ptrace.rs
marmistrz: Well, gdb always talks about the inferior - the process that's spawned by gdb and traced.

Susurrus: Is this the term used on the man pages? That's the reference docs we link to, so we should use that terminology. I believe they use "tracer" and "tracee".

Susurrus:
> For some reason the new doc tests are not run by cargo test. Do you see why?

The current parser fails to find some doctests. If you run with a current version of Rust you'll see a warning saying that some tests will begin to be run when we switch to a new parser which will be coming by the end of the year.

> Should I show the whole workflow in the doc tests?

However much you think is helpful to have in there. I think it's most important to show how our Rust version differs from the C version.

> Why do we hard forbid the users use the low-level wrapper? If they need it, why forbid them?

Because they don't need it. There's no behavior we're preventing with the high-level wrapper and we offer much better type safety with them. That's the whole point of `nix`. Eventually we'll move towards where `ptrace()` won't exist because the higher-level wrappers will do all the work.

marmistrz:
Yeah, but it's a clutter to maintain. I currently have to add at least
```
PTRACE_CONT | PTRACE_PEEKUSER | PTRACE_PEEKTEXT | PTRACE_PEEKDATA
```
and the same for other kinds of routines. I'd rather just mark it deprecated, Rust will issue a warning on its own

marmistrz:
What about the new `enum Register`? Is it ok?

marmistrz:
What about dropping support for Rust 1.13? I added a trait which is a shortcut for a very common ptrace usage pattern: (I was using it almost everywhere) 
https://github.com/marmistrz/nix/blob/fd64d367fa71b9fefa5af76771a28d62a0a88753/src/sys/ptrace.rs#L308
but `before_exec` was unstable in 1.13 (it's stable now)

asomers:
Looks like `before_exec` requires 1.15.  Is there an easy way to do without?  I'd rather not raise the minimum supported version if we don't have to.

marmistrz:
This is probably blocked by https://github.com/rust-lang/libc/pull/699

Susurrus:
You need to do `super::Word` instead as `Word` is one module down from `test`.

And let's see how rust-lang/libc#699 shakes out.

In the mean time can you get this code all proper again (you commented out a ton of stuff in your last update) and also squash everything into one commit that's rebased on master? And finally this will need a CHANGELOG entry under "Added"

marmistrz:
But I'm doing a `use super::*`, so `Word` should be imported.

Susurrus:
You should be doing your testing locally on 1.13 to debug these things. `rustup` makes it quite easy because one you have the toolchain installed you can test like `rustup run 1.13.0 cargo test` and debug.

So I did that and if you change your glob-import to `use super::Word` then it works, not certain why. You shouldn't have been using a glob import to begin with, but I don't know why this fixes it.

marmistrz:
As I was told on #rust-beginners, it's a bug, fixed in 1.15.0: https://github.com/rust-lang/rust/issues/23157
The glob import is how the Rust book does it: https://doc.rust-lang.org/book/second-edition/ch11-01-writing-tests.html

marmistrz:
Oh, we'll need to add the same registers to Android x86: https://github.com/rust-lang/libc/pull/699#issuecomment-319096656

Maybe let's disable this for musl and android until this gets settled upstream?

Susurrus:
@marmistrz I'd prefer to wait until rust-lang/libc#699 is resolved so we can ge this feature enabled for all relevant platforms on the first pass.

marmistrz:
Writing any tests here will require #709 to be merged first. (If the current commit passes the tests, we'll only need a changelog entry and some extra tests)

Susurrus:
@marmistrz With #709 merged, could you rebase this and add tests?

marmistrz:
Yes, I'll do that tomorrow

marmistrz:
Great, I'll have to add some tests for the new functions.

Susurrus:
Is this ready for review? I still see some commented code in there.

marmistrz:
No, not yet, I said that I have to add the tests. I'm going to do this today.

I'm going to let you know when it is.

marmistrz:
Do you see why the syscalls are not caught by the tracer? It's basically an excerpt from my previous program and it looks prefectly fine to me.

src/sys/ptrace.rs
```diff
@@ -204,3 +199,188 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg!(1), nix::sys::ptrace::Register::RDI);
+/// # }
+#[cfg(target_arch = "x86_64")]
+#[macro_export]
+macro_rules! syscall_arg {
+    (0) => ($crate::sys::ptrace::Register::RAX);
```
vincenthage: For x86_64, `rax` corresponds to the sys arg result.
It's `orig_rax` that corresponds to the syscall number.

src/sys/ptrace.rs
```diff
@@ -204,3 +199,188 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg!(1), nix::sys::ptrace::Register::RDI);
+/// # }
+#[cfg(target_arch = "x86")]
+#[macro_export]
+macro_rules! syscall_arg {
+    (0) => ($crate::sys::ptrace::Register::EAX);
```
vincenthage: Same here.
For x86, `eax` corresponds to the sys arg result.
`orig_eax` corresponds to the syscall number.

src/sys/ptrace.rs
```diff
@@ -204,3 +199,188 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+/// # Examples
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg!(1), nix::sys::ptrace::Register::RDI);
+/// # }
+#[cfg(target_arch = "x86_64")]
+#[macro_export]
+macro_rules! syscall_arg {
```
vincenthage: These mappings only work for 64bits programs.
They are different for 32bits programs run by a x86_64 processor.

src/sys/ptrace.rs
vincenthage: One simple test is to cancel a syscall by changing its syscall number to 0 during the enter stage, and then reverting it during the exit stage.

src/sys/ptrace.rs
marmistrz: I don't know if there's much we can do with it. Every 64-bit program will link with a 64-bit libc, use 64-bit syscall table from libc, have 64-bit `user_regs_struct`.

Susurrus:
What's the current state of this? I see some more work has been done and @vincenthage has been adding some comments.

marmistrz:
As for me, the PR is ready, modulo the comment about x86 programs on x86_64. But, as I said, I don't know what we could do with that.

vincenthage:
And that's exactly were the fun begins, because whether or not the current traced program is 32 or 64bits does not depend on a fixed `cfg` configuration, but on the registers themselves at the time when they are fetched.

- Mappings for x86_64 processors and 64bits programs:
```rust
macro_rules! syscall_arg {
    (0) => ($crate::sys::ptrace::Register::ORIG_RAX);
    (1) => ($crate::sys::ptrace::Register::RDI);
    (2) => ($crate::sys::ptrace::Register::RSI);
    (3) => ($crate::sys::ptrace::Register::RDX);
    (4) => ($crate::sys::ptrace::Register::R10);
    (5) => ($crate::sys::ptrace::Register::R8);
    (6) => ($crate::sys::ptrace::Register::R9);
```
- Mappings for x86_64 processors and 32bits programs:
```rust
macro_rules! syscall_arg {
    (0) => ($crate::sys::ptrace::Register::ORIG_RAX);
    (1) => ($crate::sys::ptrace::Register::RDX);
    (2) => ($crate::sys::ptrace::Register::RCX);
    (3) => ($crate::sys::ptrace::Register::RDX);
    (4) => ($crate::sys::ptrace::Register::RSI);
    (5) => ($crate::sys::ptrace::Register::RDI);
    (6) => ($crate::sys::ptrace::Register::RBP);
}
```
You could select one of the two if you had a `is32bits()` function, but this function itself depends on the CS register:
```rust
fn is32bits(regs: &user_regs_struct) -> bool {
    regs.cs == 0x23	
}
```
Hence, if you want your mappings to be exact, you have 4 ways of fixing this issue:
* Fetch the value of the CS register everytime `syscall_arg` is called.
* Add an argument to the `syscall` macro, so that the user can themselves indicate if it's a 32 or 64bits; it will be their responsability to cache the value somewhere.
* Modify the mappings so that they only work with a complete `user_regs_struct` obtained with `GETREGS`.
* Remove all mappings.

I personally use the solution 3) (see my posts at the beginning of the thread), but I recommend the 4) if you want to keep this PR simple.
Though the unit test will have to get the registers manually, but that shouldn't be an issue.

marmistrz:
What about adding a separate macro on x86_64, `syscall32` and leave the checks to the user. The user will usually have to check the value of the CS register only at the start of each traced process.
1 and 3 are bad for performance, so I don't really want to have them here

vincenthage:
> What about adding a separate macro on x86_64, syscall32 and leave the checks to the user.

Doesn't sound like a bad idea, let's go with this.
Just a couple remaining points:
- Are you sure about naming the macro `syscall`? This macro doesn't do a syscall, it returns the offset of the registers. A more appropriate name should be chosen so that it doesn't conflict with other crates macros. Maybe `sysarg`, `sysarg_offset` or even `register_offset`.
- The macro only returns the syscall number and the arguments' offsets. What about the return values, the stack pointer, etc... (I have the mappings if you want to complete the macros).
- Now that the offsets are statically determined, isn't there a way to get the offsets in an enum at compile-time, without having to use a macro?

marmistrz:
Oh, yeah, I meant syscall_arg32.
But it may be indeed better to make it an enum

vincenthage:
Yep, though before that, I wonder if there is a way to create a mapping macro that would work both for the offset (so that would map `SysNum` to `$crate::sys::ptrace::Register::ORIG_RAX`) and the `user_regs_struct` (so that would map `SysNum, regs` to `regs.orix_rax`).
This would make the mapping double more useful, and it could be use by users that use `GETREGS` too.

I see one way of doing it, but it would require having a macro that translate uppercase fields into lowercase fields (`ORIG_RAX` -> `orig_rax`) so that we can reuse the same mapping macros). But it's possible.

marmistrz:
@vincenthage I'd like to close this pull request relatively quickly, because I'll need to rebase a project to use the upstreamed changes before I go on a short vacation. It's a PoC, which won't take any use of extra things mentioned here.

Enum has this problem that can only be mapped to `isize`. Not to another enum

My suggestion: let's stick to a macro and think about a better way of doing it in a separate issue. I may even ask on the the Rust forum for an idiomatic way of doing such mappings.

vincenthage:
As you want.
We can leave the rest of the mappings for later then.

Susurrus:
I haven't been following because you guys seemed to be making progress on this. What's the current state of this PR, have you guys got it sorted mostly? A quick glance at this shows that it needs a CHANGELOG entry and the commits to be squashed. You have a couple of changes that aren't functionality, so if you could leave those in a separate commit (1 is fine) it'd be apprecaited.

marmistrz:
@Susurrus it's ready. I'll rebase, squash and add the changelog entry.
I'll remove the trait for now and make a separate PR out of it, waiting for buildbot update.

marmistrz:
Done, please review.

src/sys/ptrace.rs
```diff
@@ -134,8 +134,8 @@ pub unsafe fn ptrace(request: Request, pid: Pid, addr: *mut c_void, data: *mut c
     }
 }
 
-fn ptrace_peek(request: Request, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
-    let ret = unsafe {
+unsafe fn ptrace_peek(request: Request, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
```
Susurrus: Please separate this into a separate commit and also add a separate changelog entry for it under CHANGED.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,195 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
     }
 }
 
+#[cfg(target_arch = "x86_64")]
+#[allow(non_camel_case_types)]
+#[derive(Debug, PartialEq)]
+pub enum Register {
```
Susurrus: This needs docs. Every variant should be documented as well.

test/sys/test_ptrace.rs
```diff
@@ -89,3 +199,4 @@ fn test_ptrace_cont() {
         Err(_) => panic!("Error: Fork Failed")
     }
 }
+
```
Susurrus: Please remove this extra newline.

test/sys/test_ptrace.rs
```diff
@@ -55,6 +162,9 @@ fn test_ptrace_cont() {
     use nix::unistd::fork;
     use nix::unistd::ForkResult::*;
 
+    let _m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+
```
Susurrus: This only needs 1 newline.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,195 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+/// # Examples
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg32!(1), nix::sys::ptrace::Register::RBX);
+/// # }
+#[cfg(target_arch = "x86_64")]
+#[macro_export]
+macro_rules! syscall_arg32 {
```
Susurrus: Why are these named differently if they're the same thing just for different platforms? I'm confusd by `syscall_arg` is not on user-API just with different implementations for x86 versus x86_64.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,195 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+    (0) => ($crate::sys::ptrace::Register::ORIG_EAX);
+    (1) => ($crate::sys::ptrace::Register::EBX);
+    (2) => ($crate::sys::ptrace::Register::ECX);
+    (3) => ($crate::sys::ptrace::Register::EDX);
+    (4) => ($crate::sys::ptrace::Register::ESI);
+    (5) => ($crate::sys::ptrace::Register::EDI);
+    (6) => ($crate::sys::ptrace::Register::EBP);
+}
+
+type Word = usize;
```
Susurrus: This needs a doc comment.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,195 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+
+/// Makes the `PTRACE_PEEKUSER` request to ptrace
```
Susurrus: This should provide more details as to what that means especially with regards to the function arguments and return value. The same should be done for the other ptrace wrappers in this PR.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,195 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+        val as *mut c_void,
+    ).map(|_| ()) // ignore the useless return value
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Word;
+
+    #[test]
+    fn test_types() {
```
Susurrus: If `Word` is an alias for `usize`, then these tests don't do anything, so let's just remove them. If they do something useful, please document this more clearly because I'm not certain what it is.

marmistrz:
is it ok now?

src/sys/ptrace.rs
Susurrus: Indeed you're correct assuming you meant it *is* a part of our private API.

src/sys/ptrace.rs
```diff
@@ -134,8 +134,8 @@ pub unsafe fn ptrace(request: Request, pid: Pid, addr: *mut c_void, data: *mut c
     }
 }
 
-fn ptrace_peek(request: Request, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
-    let ret = unsafe {
+unsafe fn ptrace_peek(request: Request, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
+    let ret = {
```
Susurrus: There's no reason to have this in braces, just make these two separate lines instead.

src/sys/ptrace.rs
marmistrz: What's the point of announcing our implementation details?

src/sys/ptrace.rs
```diff
@@ -238,12 +236,7 @@ pub fn syscall(pid: Pid) -> Result<()> {
 /// Attaches to the process specified in pid, making it a tracee of the calling process.
 pub fn attach(pid: Pid) -> Result<()> {
     unsafe {
-        ptrace_other(
-            Request::PTRACE_ATTACH,
-            pid,
-            ptr::null_mut(),
-            ptr::null_mut(),
-        ).map(|_| ()) // ignore the useless return value
+        ptrace_other(Request::PTRACE_ATTACH, pid, ptr::null_mut(), ptr::null_mut()).map(|_| ()) // ignore the useless return value
```
Susurrus: This line's too long, but it's obvious what the final `map()` is doing, so just remove the comment.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,207 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
     }
 }
 
+#[cfg(target_arch = "x86_64")]
+#[allow(non_camel_case_types)]
+#[derive(Debug, PartialEq)]
+/// Represents all possible ptrace-accessible registers on x86_64
```
Susurrus: Please move this comment before the `#[cfg...`

src/sys/ptrace.rs
```diff
@@ -275,3 +268,207 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+pub enum Register {
+    R15 = 8 * ::libc::R15 as isize,
```
Susurrus: I'd prefer to have all these registers documented, which I think I requested in my earlier review. Could you do this?

src/sys/ptrace.rs
marmistrz: fixed that and reformatted to avoid exceeding 80 columns of length.

src/sys/ptrace.rs
```diff
@@ -275,3 +268,207 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+    EFL = 4 * ::libc::EFL as isize,
+    UESP = 4 * ::libc::UESP as isize,
+    SS = 4 * ::libc::SS as isize,
+}
+
+/// Returns the register containing nth register argument.
+///
+/// 0th argument is considered to be the syscall number.
+/// Please note that these mappings are only valid for 64-bit programs.
+/// Use syscall_arg32 for tracing 32-bit programs instead.
```
Susurrus: Please link to `syscall_arg32` and put it in backticks. You can see how we link to other types by looking at `src/sys/wait.rs:57`.

src/sys/ptrace.rs
```diff
@@ -275,3 +277,207 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+/// # Examples
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg!(1), nix::sys::ptrace::Register::RDI);
+/// # }
+#[cfg(target_arch = "x86")]
+#[macro_export]
+macro_rules! syscall_arg {
```
Susurrus: All 3 of these macros should be somehow tied to peek/pokeuser in their docs or the docs to those ptrace wrappers. I'm actually inclined to say this would be a more coherent API if these macros were actually functions on the Register type instead of being independent macros.

src/sys/ptrace.rs
```diff
@@ -275,3 +277,207 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+/// Returns the register containing nth register argument.
+///
+/// 0th argument is considered to be the syscall number.
+///
+/// # Examples
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg!(1), nix::sys::ptrace::Register::RDI);
```
Susurrus: This isn't a good example because a) it uses an assert (this should be a test if anything) and b) it doesn't demonstrate actual usage. No one is going to write code like this. How would you actually use `syscall_arg!` in practice? As an argument to `pokeuser`, right? This applies to all 3 of these macros.

test/sys/test_ptrace.rs
```diff
@@ -1,9 +1,14 @@
+#![cfg(all(target_os = "linux", any(target_arch = "x86",
```
Susurrus: This is already specified in `test/sys/mod.rs`, why are you respecifying it here? You probably want to restrict just your new tests a little more rather than the whole module.

src/sys/ptrace.rs
marmistrz: Please, R15 is just an... R15 register. If someone's using ptrace, it they have to know what the register does anyway.

I don't see any added value in documenting it as 
```
R15 = 8 * ::libc::R15 as isize, /// the R15 register
```

People using `ptrace` are not kids.

src/sys/ptrace.rs
marmistrz: It's a macro so as to allow compile-type checks.

test/sys/test_ptrace.rs
marmistrz: Ok, I don't remember. Fixed that.

src/sys/ptrace.rs
marmistrz: done

src/sys/ptrace.rs
marmistrz: done

src/sys/ptrace.rs
marmistrz: done

marmistrz:
I think that having this compile-time checked is a superior benefit. 

Could we please get it merged before Friday? I want to rebase a project onto the upstreamed functions before I go on a short vacation.

src/sys/ptrace.rs
```diff
@@ -275,3 +280,209 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
     }
 }
 
+/// Represents all possible ptrace-accessible registers on x86_64
+#[cfg(target_arch = "x86_64")]
+#[allow(non_camel_case_types)]
+#[derive(Debug, PartialEq)]
+pub enum Register {
```
Susurrus: Is there a logical ordering to these registers? I would at least expect R10..R15 to be in order and they're not. I'm not certain how the docs for these are generally arranged, but maybe we can do a better grouping here?

src/sys/ptrace.rs
Susurrus: I have no issues with the boilerplate, I'm suggesting that examples be a demonstration to the user of how this is used. While yes, you show the syntax to use with this macro, we can do better by showing it in a broader context, where it's used as an argument to one of the `ptrace` wrapper functions. So I suggest you copy one of the lines where this is used from the test code (like as an argument to `pokedata`) and stick it here and make this code block `no_run` so it will at least be compile-checked.

src/sys/ptrace.rs
Susurrus: This should be stated somewhere in the docs I believe, as the overhead of our `nix` wrappers constantly comes up. It'd be good to state for the users that the use of the macro is to result in compile-time lookups instead of run-time lookups to provide the most minimal overhead possible in these APIs.

Susurrus:
Yep we can shoot for Friday and I think it's doable given we're mostly fixing little things at this point and I think the API is solid.

src/sys/ptrace.rs
```diff
@@ -275,3 +280,209 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+    RBX = 8 * ::libc::RBX as isize,
+    R11 = 8 * ::libc::R11 as isize,
+    R10 = 8 * ::libc::R10 as isize,
+    R9 = 8 * ::libc::R9 as isize,
+    R8 = 8 * ::libc::R8 as isize,
+    RAX = 8 * ::libc::RAX as isize,
+    RCX = 8 * ::libc::RCX as isize,
+    RDX = 8 * ::libc::RDX as isize,
+    RSI = 8 * ::libc::RSI as isize,
+    RDI = 8 * ::libc::RDI as isize,
```
Susurrus: I think we can definitely document these registers. For example `RDI` is the "Destination Index register. Used as a pointer to a destination in stream operations" (taken from [here](https://en.wikibooks.org/wiki/X86_Assembly/X86_Architecture)). Even for R15 you can write "General purpose register 15", which is more descriptive than the name itself.

I think it's important to be kind to our users by a) not assuming too much about their knowledge and b) to lead them in the right direction if they want to know more.

marmistrz: `General purpose register 15` says nothing. You need to know what the ABI is to do anything with ptrace. And if they need to search, they'll be better off searching for the 3-letter names, since it's what the whole C API uses and what everything uses.

Being kind is one thing, but we can't teach them everything. There's some learning curve for using ptrace and knowing registers is just a thing. The process has to be stopped, or weird errors will arise. It's out of scope for nix to provide a ptrace tutorial.

The thing about learning curve is that it protects users who don't know enough from shooting themselves in the foot by using something they don't understand.

If you insist, we can add a link to the wikibook in the docs.

src/sys/ptrace.rs
marmistrz: Yes, that's the order from `user_regs_struct`. Please, don't make nix an operating systems kindergarten. As said in another comment, there's a lot of other things that you need to know, to be able to use `ptrace`.

src/sys/ptrace.rs
marmistrz: More compile-time error checks. We decided with @vincenthage that it would be the best to make it an `enum` but it's non-trivial and I don't really have time to dwell on it.

Should I mention that it's a macro for the sake of _compile-time error checking_?

src/sys/ptrace.rs
marmistrz: This needs a valid `struct Pid`, which has to be obtained from somewhere (a user will wonder what the heck is `pid` here). What about just mentioning that the purpose is to achieve arch-independent syscall argument lookups?

marmistrz:
@Susurrus, please remember I'm UTC+2. I'd like to have it merged before Friday morning UTC+2.

test/sys/test_ptrace.rs
```diff
@@ -55,6 +158,8 @@ fn test_ptrace_cont() {
     use nix::unistd::fork;
     use nix::unistd::ForkResult::*;
 
+    let _m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
```
Susurrus: This is unrelated to all the other changes in this commit. Can you break this out into a separate commit?

src/sys/ptrace.rs
Susurrus: @marmistrz Any comment on this?

src/sys/ptrace.rs
Susurrus: Yes, that's perfect.

src/sys/ptrace.rs
Susurrus: You can leave the boilerplate commented-out here as well. Do what you need to do to get it to compile, but having something like `ptrace::peekuser(child, syscall_arg!(0))` As the example would be better in my opinion.

test/sys/test_ptrace.rs
```diff
@@ -46,6 +47,108 @@ fn test_ptrace_setsiginfo() {
...
+
+            ptrace::setoptions(child, ptrace::PTRACE_O_TRACESYSGOOD).unwrap();
+
+            loop {
+                ptrace::syscall(child).unwrap();
+                match waitpid(child, None).unwrap() {
+                    WaitStatus::PtraceSyscall(child) => {
+                        match syscall_no {
+                            None => {
+                                let no = ptrace::peekuser(child, syscall_arg!(0)).unwrap();
```
Susurrus: Is it possible to run the tests as a 32-bit process on a 64-bit architecture and so these tests will fail because they use `syscall_arg!`? If so we should either fix the tests to work regardless or disable the tests.

src/sys/ptrace.rs
Susurrus: If you feel so strongly about this, then let's drop this as it's mostly a nice-to-have. Eventually I'm going to enable a requirement that everything have docs, but since we're not there yet this isn't too important.

Susurrus:
@marmistrz Looked like we were pretty close on this. Sorry we missed your original target date, but any interest in polishing this off for merging?

marmistrz:
Interest yes, but unfortunately severe lack of time. This will have to wait for another week or two at least.

marmistrz:
is this PR still relevant?

Susurrus:
@marmistrz Who is supposed to answer that question? As you proposed this PR, I guess I'd ask you if it's still relevant: do you want it in nix and are you willing to finish polishing it off to merge? If no to either of those, let's go ahead and close this.

marmistrz:
@Susurrus more precisely, I'm asking if the changes accepted during the last half a year have not obsoleted this PR. In the meantime the features could have been partially implemeneted or the API may have gone in a completely different direction.

Susurrus:
I don't know what's happened with the API, thought the CHANGELOG is comprehensive and I'd suggest you check there first, though `git log` when you specify a file will tell you what commits touched it, so that'd be easy to run too. Once you've determined that, please rebase this API if there's anything still worth merging.

test/sys/test_ptrace.rs
marmistrz: I simply removed that for now. This can be added once the changes are squashed.

marmistrz:
Fixed the things mentioned in the last review. Should I now squash it all into one commit

> `ptrace::peekuser(child, syscall_arg!(0))` As the example would be better in my opinion.
this can't be easily asserted, so I'd leave it as it is.

CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#739](https://github.com/nix-rust/nix/pull/739))
 - Expose `signalfd` module on Android as well.
   ([#739](https://github.com/nix-rust/nix/pull/739))
+- Added nix::sys::ptrace::detach.
```
Susurrus: This should be removed, I assume it was a rebase error.

CHANGELOG.md
```diff
@@ -88,6 +89,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added socket option variant that enables the timestamp socket
   control message: `nix::sys::socket::sockopt::ReceiveTimestamp`
   ([#663](https://github.com/nix-rust/nix/pull/663))
+- Added specialized wrappers: `sys::ptrace::{peek, poke}{user, data}`
+  and macros: `syscall_arg`, `syscall_arg32` for register-to-argument
+  mappings. Using the matching routines
+  with `sys::ptrace::ptrace` is now deprecated.
```
Susurrus: There's some weird indentation going on with this line.

src/sys/ptrace.rs
Susurrus: If you need valid values and can't get them, just mark this as only a compilation test by using \```no_run` to start the example. The point of an example isn't to test the code it's to show the user how they might use the code and provide a little additional context. The example I provided above, when annotated with a little description, would suit both of those goals.

src/sys/ptrace.rs
```diff
@@ -280,7 +285,214 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+/// Please note that these mappings are only valid for 64-bit programs.
+/// Use [`syscall_arg32`] for tracing 32-bit programs instead.
+///
+/// [`syscall_arg32`]: macro.syscall_arg32.html
+/// # Examples
+///
+/// ```
+/// # #[macro_use] extern crate nix;
+/// # fn main() {
+/// assert_eq!(syscall_arg!(1), nix::sys::ptrace::Register::RDI);
```
Susurrus: Please improve this example in line with my comments about `syscall_arg32!`.

src/sys/ptrace.rs
```diff
@@ -1,4 +1,7 @@
-//! For detailed description of the ptrace requests, consult `man ptrace`.
+//! Interface for `ptrace`
+//!
+//! For detailed description of the ptrace requests, consult [`ptrace`(2)].
+//! [`ptrace`(2)]: http://man7.org/linux/man-pages/man2/ptrace.2.html
```
Susurrus: This doesn't render correctly on Rust 1.25.0.

test/sys/test_ptrace.rs
Susurrus: Can you address my question here?

marmistrz:
Let's close it. This PR is over two years old and the API has substantially change since this PR was opened.

PR #667, opened 2017-07-10
Add support for getifaddrs.

### LINKED ISSUE #650: missing getifaddrs
I implemented a trivial wrapper around `getifaddrs`, which I'd like to polish and get into the nix crate, if that's appropriate. Any specific requirements regarding the interface? Otherwise I'd prepare something based on an `Iterator` over the entries returned.

mwanner:
Here's a first attempt at a rustian interface for `getifaddrs`, please review.

Changes for the changelog:

- Added `nix::ifaddrs::{getifaddrs, InterfaceAddressIterator,
  InterfaceAddress and InterfaceFlags}`

Closes #650.
Closes #764.

### DIFF
--- a/src/ifaddrs.rs
+++ b/src/ifaddrs.rs
@@ -1,0 +1,153 @@
+//! Query network interface addresses
+//!
+//! Uses the Linux and/or BSD specific function `getifaddrs` to query the list
+//! of interfaces and their associated addresses.
+
+use std::ffi;
+use std::fmt;
+use std::iter::Iterator;
+use std::mem;
+use std::option::Option;
+
+use libc;
+
+use {Result, Errno};
+use sys::socket::SockAddr;
+use net::if_::*;
+
+/// Describes a single address for an interface as returned by `getifaddrs`.
+#[derive(Clone, Eq, Hash, PartialEq)]
+pub struct InterfaceAddress {
+    /// Name of the network interface
+    pub interface_name: String,
+    /// Flags as from `SIOCGIFFLAGS` ioctl
+    pub flags: InterfaceFlags,
+    /// Network address of this interface
+    pub address: Option<SockAddr>,
+    /// Netmask of this interface
+    pub netmask: Option<SockAddr>,
+    /// Broadcast address of this interface, if applicable
+    pub broadcast: Option<SockAddr>,
+    /// Point-to-point destination address
+    pub destination: Option<SockAddr>,
+}
+
+impl fmt::Debug for InterfaceAddress {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "InterfaceAddress ({:?})", self.interface_name)
+    }
+}
+
+cfg_if! {
+    if #[cfg(any(target_os = "emscripten", target_os = "fuchsia", target_os = "linux"))] {
+        fn get_ifu_from_sockaddr(info: &libc::ifaddrs) -> *const libc::sockaddr {
+            info.ifa_ifu
+        }
+    } else {
+        fn get_ifu_from_sockaddr(info: &libc::ifaddrs) -> *const libc::sockaddr {
+            info.ifa_dstaddr
+        }
+    }
+}
+
+impl InterfaceAddress {
+    /// Create an `InterfaceAddress` from the libc struct.
+    fn from_libc_ifaddrs(info: &libc::ifaddrs) -> InterfaceAddress {
+        let ifname = unsafe { ffi::CStr::from_ptr(info.ifa_name) };
+        let address = unsafe { SockAddr::from_libc_sockaddr(info.ifa_addr) };
+        let netmask = unsafe { SockAddr::from_libc_sockaddr(info.ifa_netmask) };
+        let mut addr = InterfaceAddress {
+            interface_name: ifname.to_string_lossy().to_string(),
+            flags: InterfaceFlags::from_bits_truncate(info.ifa_flags as i32),
+            address: address,
+            netmask: netmask,
+            broadcast: None,
+            destination: None,
+        };
+
+        let ifu = get_ifu_from_sockaddr(info);
+        if addr.flags.contains(InterfaceFlags::IFF_POINTOPOINT) {
+            addr.destination = unsafe { SockAddr::from_libc_sockaddr(ifu) };
+        } else if addr.flags.contains(InterfaceFlags::IFF_BROADCAST) {
+            addr.broadcast = unsafe { SockAddr::from_libc_sockaddr(ifu) };
+        }
+
+        addr
+    }
+}
+
+/// Holds the results of `getifaddrs`.
+///
+/// Use the function `getifaddrs` to create this Iterator. Note that the
+/// actual list of interfaces can be iterated once and will be freed as
+/// soon as the Iterator goes out of scope.
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub struct InterfaceAddressIterator {
+    base: *mut libc::ifaddrs,
+    next: *mut libc::ifaddrs,
+}
+
+impl Drop for InterfaceAddressIterator {
+    fn drop(&mut self) {
+        unsafe { libc::freeifaddrs(self.base) };
+    }
+}
+
+impl Iterator for InterfaceAddressIterator {
+    type Item = InterfaceAddress;
+    fn next(&mut self) -> Option<<Self as Iterator>::Item> {
+        match unsafe { self.next.as_ref() } {
+            Some(ifaddr) => {
+                self.next = ifaddr.ifa_next;
+                Some(InterfaceAddress::from_libc_ifaddrs(ifaddr))
+            }
+            None => None,
+        }
+    }
+}
+
+/// Get interface addresses using libc's `getifaddrs`
+///
+/// Note that the underlying implementation differs between OSes. Only the
+/// most common address families are supported by the nix crate (due to
+/// lack of time and complexity of testing). The address family is encoded
+/// in the specific variant of `SockAddr` returned for the fields `address`,
+/// `netmask`, `broadcast`, and `destination`. For any entry not supported,
+/// the returned list will contain a `None` entry.
+///
+/// # Example
+/// ```
+/// let addrs = nix::ifaddrs::getifaddrs().unwrap();
+/// for ifaddr in addrs {
+///   match ifaddr.address {
+///     Some(address) => {
+///       println!("interface {} address {}",
+///                ifaddr.interface_name, address);
+///     },
+///     None => {
+///       println!("interface {} with unsupported address family",
+///                ifaddr.interface_name);
+///     }
+///   }
+/// }
+/// ```
+pub fn getifaddrs() -> Result<InterfaceAddressIterator> {
+    let mut addrs: *mut libc::ifaddrs = unsafe { mem::uninitialized() };
+    Errno::result(unsafe { libc::getifaddrs(&mut addrs) }).map(|_| {
+        InterfaceAddressIterator {
+            base: addrs,
+            next: addrs,
+        }
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // Only checks if `getifaddrs` can be invoked without panicking.
+    #[test]
+    fn test_getifaddrs() {
+        let _ = getifaddrs();
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,6 +43,15 @@ pub mod poll;
 
 pub mod net;
 
+#[cfg(any(target_os = "dragonfly",
+          target_os = "freebsd",
+          target_os = "ios",
+          target_os = "linux",
+          target_os = "macos",
+          target_os = "netbsd",
+          target_os = "openbsd"))]
+pub mod ifaddrs;
+
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub mod sched;
 
--- a/src/net/if_.rs
+++ b/src/net/if_.rs
@@ -17,3 +17,244 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
         Ok(if_index)
     }
 }
+
+libc_bitflags!(
+    /// Standard interface flags, used by `getifaddrs`
+    pub struct InterfaceFlags: libc::c_int {
+        /// Interface is running. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_UP;
+        /// Valid broadcast address set. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_BROADCAST;
+        /// Internal debugging flag. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_DEBUG;
+        /// Interface is a loopback interface. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_LOOPBACK;
+        /// Interface is a point-to-point link. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_POINTOPOINT;
+        /// Avoid use of trailers. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android",
+                  target_os = "fuchsia",
+                  target_os = "ios",
+                  target_os = "linux",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "solaris"))]
+        IFF_NOTRAILERS;
+        /// Interface manages own routes.
+        #[cfg(any(target_os = "dragonfly"))]
+        IFF_SMART;
+        /// Resources allocated. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android",
+                  target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "fuchsia",
+                  target_os = "ios",
+                  target_os = "linux",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "solaris"))]
+        IFF_RUNNING;
+        /// No arp protocol, L2 destination address not set. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_NOARP;
+        /// Interface is in promiscuous mode. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_PROMISC;
+        /// Receive all multicast packets. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_ALLMULTI;
+        /// Master of a load balancing bundle. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+        IFF_MASTER;
+        /// transmission in progress, tx hardware queue is full
+        #[cfg(any(target_os = "freebsd",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "ios"))]
+        IFF_OACTIVE;
+        /// Protocol code on board.
+        #[cfg(target_os = "solaris")]
+        IFF_INTELLIGENT;
+        /// Slave of a load balancing bundle. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+        IFF_SLAVE;
+        /// Can't hear own transmissions.
+        #[cfg(any(target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "osx"))]
+        IFF_SIMPLEX;
+        /// Supports multicast. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_MULTICAST;
+        /// Per link layer defined bit.
+        #[cfg(any(target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "ios"))]
+        IFF_LINK0;
+        /// Multicast using broadcast.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_MULTI_BCAST;
+        /// Is able to select media type via ifmap. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+        IFF_PORTSEL;
+        /// Per link layer defined bit.
+        #[cfg(any(target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "ios"))]
+        IFF_LINK1;
+        /// Non-unique address.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_UNNUMBERED;
+        /// Auto media selection active. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+        IFF_AUTOMEDIA;
+        /// Per link layer defined bit.
+        #[cfg(any(target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd",
+                  target_os = "ios"))]
+        IFF_LINK2;
+        /// Use alternate physical connection.
+        #[cfg(any(target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "macos",
+                  target_os = "ios"))]
+        IFF_ALTPHYS;
+        /// DHCP controlls interface.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_DHCPRUNNING;
+        /// The addresses are lost when the interface goes down. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+        IFF_DYNAMIC;
+        /// Do not advertise.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_PRIVATE;
+        /// Driver signals L1 up. Volatile.
+        #[cfg(any(target_os = "fuchsia", target_os = "linux"))]
+        IFF_LOWER_UP;
+        /// Interface is in polling mode.
+        #[cfg(any(target_os = "dragonfly"))]
+        IFF_POLLING_COMPAT;
+        /// Unconfigurable using ioctl(2).
+        #[cfg(any(target_os = "freebsd"))]
+        IFF_CANTCONFIG;
+        /// Do not transmit packets.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_NOXMIT;
+        /// Driver signals dormant. Volatile.
+        #[cfg(any(target_os = "fuchsia", target_os = "linux"))]
+        IFF_DORMANT;
+        /// User-requested promisc mode.
+        #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
+        IFF_PPROMISC;
+        /// Just on-link subnet.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_NOLOCAL;
+        /// Echo sent packets. Volatile.
+        #[cfg(any(target_os = "fuchsia", target_os = "linux"))]
+        IFF_ECHO;
+        /// User-requested monitor mode.
+        #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
+        IFF_MONITOR;
+        /// Address is deprecated.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_DEPRECATED;
+        /// Static ARP.
+        #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
+        IFF_STATICARP;
+        /// Address from stateless addrconf.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_ADDRCONF;
+        /// Interface is in polling mode.
+        #[cfg(any(target_os = "dragonfly"))]
+        IFF_NPOLLING;
+        /// Router on interface.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_ROUTER;
+        /// Interface is in polling mode.
+        #[cfg(any(target_os = "dragonfly"))]
+        IFF_IDIRECT;
+        /// Interface is winding down
+        #[cfg(any(target_os = "freebsd"))]
+        IFF_DYING;
+        /// No NUD on interface.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_NONUD;
+        /// Interface is being renamed
+        #[cfg(any(target_os = "freebsd"))]
+        IFF_RENAMING;
+        /// Anycast address.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_ANYCAST;
+        /// Don't exchange routing info.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_NORTEXCH;
+        /// IPv4 interface.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_IPV4;
+        /// IPv6 interface.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_IPV6;
+        /// in.mpathd test address
+        #[cfg(any(target_os = "solaris"))]
+        IFF_NOFAILOVER;
+        /// Interface has failed
+        #[cfg(any(target_os = "solaris"))]
+        IFF_FAILED;
+        /// Interface is a hot-spare
+        #[cfg(any(target_os = "solaris"))]
+        IFF_STANDBY;
+        /// Functioning but not used
+        #[cfg(any(target_os = "solaris"))]
+        IFF_INACTIVE;
+        /// Interface is offline
+        #[cfg(any(target_os = "solaris"))]
+        IFF_OFFLINE;
+        #[cfg(any(target_os = "solaris"))]
+        IFF_COS_ENABLED;
+        /// Prefer as source addr.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_PREFERRED;
+        /// RFC3041
+        #[cfg(any(target_os = "solaris"))]
+        IFF_TEMPORARY;
+        /// MTU set with SIOCSLIFMTU
+        #[cfg(any(target_os = "solaris"))]
+        IFF_FIXEDMTU;
+        /// Cannot send / receive packets
+        #[cfg(any(target_os = "solaris"))]
+        IFF_VIRTUAL;
+        /// Local address in use
+        #[cfg(any(target_os = "solaris"))]
+        IFF_DUPLICATE;
+        /// IPMP IP interface
+        #[cfg(any(target_os = "solaris"))]
+        IFF_IPMP;
+    }
+);
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -201,6 +201,26 @@ pub enum AddressFamily {
     Natm = libc::AF_NATM,
 }
 
+impl AddressFamily {
+    /// Create a new `AddressFamily` from an integer value retrieved from `libc`, usually from
+    /// the `sa_family` field of a `sockaddr`.
+    ///
+    /// Currently only supports these address families: Unix, Inet (v4 & v6), Netlink
+    /// and System. Returns None for unsupported or unknown address families.
+    pub fn from_i32(family: i32) -> Option<AddressFamily> {
+        match family {
+            libc::AF_UNIX => Some(AddressFamily::Unix),
+            libc::AF_INET => Some(AddressFamily::Inet),
+            libc::AF_INET6 => Some(AddressFamily::Inet6),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            libc::AF_NETLINK => Some(AddressFamily::Netlink),
+            #[cfg(any(target_os = "macos", target_os = "macos"))]
+            libc::AF_SYSTEM => Some(AddressFamily::System),
+            _ => None
+        }
+    }
+}
+
 #[derive(Copy)]
 pub enum InetAddr {
     V4(libc::sockaddr_in),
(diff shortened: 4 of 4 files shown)


src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+use std::iter::Iterator;
+use std::mem;
+use std::option::Option;
+
+use libc;
+use {Result, Error};
+
+use sys::socket::InetAddr;
+
+#[cfg(any(target_os = "linux", target_os = "emscripten", target_os = "fuchsia"))]
```
Susurrus: There's no need to use a module here. Please use a single `libc_bitflags` instance and annotate each entry with a `cfg` attribute with their platform support.

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
+//! Query network interface addresses
+//!
+//! Uses the Linux and/or BSD specific function getifaddrs to query the list
+//! of interfaces and their associated addresses.
+
+use std::ffi;
```
Susurrus: Please group the use statements into three groups: stdlib, third-party, and internal. Put a blank line between each group.

Susurrus: Please group the use statements into three groups: stdlib, third-party, and internal. Put a blank line between each group.

Susurrus: Please group the use statements into three groups: stdlib, third-party, and internal. Put a blank line between each group.

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+use libc;
+use {Result, Error};
+
+use sys::socket::InetAddr;
+
+#[cfg(any(target_os = "linux", target_os = "emscripten", target_os = "fuchsia"))]
+mod flags {
+    use libc;
+    libc_bitflags!(
+        /// Flags for `InterfaceAddress` entries returned by `getifaddrs`
```
Susurrus: That comment doesn't really provide any detail. Can you summarize what these values actually refer to?

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+            IFF_LOOPBACK,
+        }
+    );
+}
+
+pub use self::flags::*;
+
+/// Describes a single address for an interface as returned by `getifaddrs`.
+pub struct InterfaceAddress {
+    /// name of the interface
```
Susurrus: Please capitalize the first character of all of these comments

mwanner:
One area I'm not quite sure about is the use of the nix specific `InetAddr` struct. Due to it supporting only IPv4 and IPv6, but not AF_PACKET, for example, lots of entries returned from `getifaddrs` remain empty.

This works for me, ATM. However, I think nix should better provide some way to represent those entries as well.

src/ifaddrs.rs
Susurrus: I also think Android supports all these as well (you can find this out easily by searching the libc repo).

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+                Some(InterfaceAddress::from_libc_ifaddrs(ifaddr))
+            }
+            None => None,
+        }
+    }
+}
+
+/// Get interface addresses using libc's `getifaddrs`
+///
+/// Note that the underlying implementation differs between implementations and
```
Susurrus: "the underlying implementation differs between implementations", that's pretty redundant. Can you rephrase this to be a little more clear?

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+///             },
+///             None => {}
+///         }
+///     },
+///     Err(err) => {
+///         println!("error getting interface addresses: {}", err)
+///     }
+/// }
+/// ```
+#[cfg(not(any(target_os = "android", target_os = "haiku")))]
```
Susurrus: Can you check that this isn't on Android? According to [this](https://android.googlesource.com/platform/prebuilts/ndk/+/dev/platform/sysroot/usr/include/ifaddrs.h) it should.

src/sys/socket/addr.rs
```diff
@@ -64,6 +64,24 @@ impl InetAddr {
         }
     }
 
+    /// Creates an `IpAddr` struct from libc's sockaddr.
```
asomers: s/IpAddr/InetAddr/

src/sys/socket/addr.rs
```diff
@@ -64,6 +64,24 @@ impl InetAddr {
...
+    ///
+    /// Note that this supports only IPv4 and IPv6 address families.
+    pub unsafe fn from_libc_sockaddr(addr: *mut libc::sockaddr) -> Option<InetAddr> {
```
Susurrus: Why is this unsafe and the other `from_libc` isn't? What's the reason for providing this API?

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub netmask: Option<InetAddr>,
+    /// broadcast address of interface, if applicable
+    pub broadcast: Option<InetAddr>,
+    /// point-to-point destination address
+    pub destination: Option<InetAddr>,
+}
+
+impl InterfaceAddress {
+    /// Create an `InterfaceAddress` from the libc struct.
+    fn from_libc_ifaddrs(info: &libc::ifaddrs) -> InterfaceAddress {
```
Susurrus: Why is this an exported function?

src/sys/socket/addr.rs
```diff
@@ -64,6 +64,24 @@ impl InetAddr {
...
+        if addr.is_null() {
+            None
+        } else {
+            let family = (*addr).sa_family as u32;
+            match mem::transmute::<u32, AddressFamily>(family) {
```
asomers: Instead of using transmute here, it's better to provide an `AddressFamily::from_u32` method.

Susurrus:
Thanks for the PR! There are a few things to address from this, notably a lack of tests besides the doctest that you added (which is great BTW).

src/ifaddrs.rs
asomers: nix doesn't support emscripten or fuschia, so you don't need to mention them here.  It doesn't hurt, though, so you can leave it in if you want to.  But why not enable it for the BSDs?  They support all this stuff.  If there are a few Linux specific flags, you can `#[cfg()]` individual flags instead of the whole thing.

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+            } else if addr.flags.contains(IFF_BROADCAST) {
+                addr.broadcast = ifo;
+            }
+        }
+
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "freebsd",
+                    target_os = "dragonfly", target_os = "openbsd", target_os = "netbsd",
+                    target_os = "bitrig", target_os = "solaris"))]
+        {
+            addr.destination = unsafe { InetAddr::from_libc_sockaddr(info.ifa_dstaddr) };
```
asomers: Looks like you forgot to set the broadcast for non-Linux platforms.  You can do it the same way as on Linux.  Depending on the flags, either `ifa_dstaddr` or `ifa_broadaddr` is set.  The only difference is that on BSD, one of those is a macro for the other, whereas on Linux they're a union.

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// Get interface addresses using libc's `getifaddrs`
+///
+/// Note that the underlying implementation differs between implementations and
+/// the nix crate doesn't support `AF_PACKET` entries. Therefore, the returned
+/// list often contains empty entries.
+///
+/// # Example
+/// ```
+/// match nix::ifaddrs::getifaddrs() {
```
asomers: For the example code, it would probably be clearer to use `unwrap` instead of a `match` block

src/ifaddrs.rs
asomers: And the BSDs, too.  They all have it.

src/ifaddrs.rs
```diff
@@ -0,0 +1,180 @@
...
+///     },
+///     Err(err) => {
+///         println!("error getting interface addresses: {}", err)
+///     }
+/// }
+/// ```
+#[cfg(not(any(target_os = "android", target_os = "haiku")))]
+pub fn getifaddrs() -> Result<InterfaceAddressIterator> {
+    let mut addrs: *mut libc::ifaddrs = unsafe { mem::zeroed() };
+    match unsafe { libc::getifaddrs(&mut addrs) } {
```
asomers: You can probably simplify this logic to something like:
`Errno::result(unsafe { libc::getifaddrs(&mut addrs) }).map(|_| Ok(InterfaceAddressIterator{base: addrs, next:addrs}))`

src/ifaddrs.rs
Susurrus: Indeed, it's already defined in `libc` for all BSDs and Solaris as well.

src/ifaddrs.rs
mwanner: Android doesn't support `getifaddrs`, that's probably why I excluded it. But you're right, the flags themselves are known to Android as well.

Maybe I should move `InterfaceFlags` to `net/if_.rs`, as those constants are defined in `/usr/include/net/if.h` and not in `ifaddrs.h`.

Susurrus: [Android does](https://android.googlesource.com/platform/prebuilts/ndk/+/dev/platform/sysroot/usr/include/ifaddrs.h), but libc does need to have it added.

I think it's fine having them here for now.

Susurrus:
@asomers [mockers](https://crates.io/crates/mockers) looks promising for a general-purpose mocking library.

src/ifaddrs.rs
mwanner: Oh, and: I used the module to reduce the amount of `#[cfg()]` conditions. I only used the module as I didn't find any good way to write this without repeating the conditions many times. I'm fine if you prefer the latter and changed things accordingly.

src/ifaddrs.rs
mwanner: I checked against rust's libc bindings, where `src/unix/notbsd/android/mod.rs` doesn't define `getifaddrs` (at least not in 3f903a1d from master as of Jul 11).

And: Huh? `not(any(target_os = "android", target_os = "haiku"))` already includes the BSDs.

src/ifaddrs.rs
mwanner: Exported? It's not public and used internally, so I fear I don't understand what you're asking for.

src/sys/socket/addr.rs
mwanner: I marked this one unsafe because it takes a raw pointer, which could point anywhere and yield a segfault. Again, this is not a public function, but only used internally (by that other `InterfaceAddress::from_libc_ifaddrs` function)

src/sys/socket/addr.rs
mwanner: Right, that's a last-minute typo. I didn't notice `SockAddr` before and currently think the `getifaddrs` API shoud rather use `SockAddr`.

src/ifaddrs.rs
mwanner: Ah, it's the same field for BSDs as well? I did it that way due to rust's libc bindings, where there is not `ifa_broadaddr` or `ifa_broadcastaddr` defined. I'll check the C headers from FreeBSD and adjust.

Given the BSDs also lack `IFF_BROADCAST` and `IFF_POINTOPOINT`, how do they distinguish between broadcast vs destination address?

src/ifaddrs.rs
mwanner: An unwrap is very likely to fail due to the many `address = None` cases you'll encounter. It certainly makes `cargo test` fail on my box (Debian, with many AF_PACKET entries that cannot be represented ATM).

src/ifaddrs.rs
mwanner: I duplicated the comment from the C header, now.

mwanner:
@susurrus and @asomers: Thanks a lot for your review. I adjusted some things and commented on others.

I still need to check how destination vs broadcast addresses work for BSDs. And I'd like to change InetAddr for SockAddr, maybe adding better support for `AF_PACKET` entries.

mwanner:
Proper support for BSDs depends on corrections in libc, see rust-lang/libc#682.

I'm currently in favour of disabling `getifaddrs` for those OSes for which libc is buggy and at least add the `getifaddrs` support for the others (including Linux and OS X, IIRC). We can later enable full support when libc gets corrected. (BTW, can rust handle code dependent on versions of external crates?)

src/net/if_.rs
```diff
@@ -17,3 +17,40 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
         Ok(if_index)
     }
 }
+
+libc_bitflags!(
+    /// Standard interface flags, used by `getifaddrs`
+    pub flags InterfaceFlags: libc::c_int {
+        #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: The `target_os` list should be alphabetized.

src/net/if_.rs
```diff
@@ -17,3 +17,40 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
         Ok(if_index)
     }
 }
+
+libc_bitflags!(
```
Susurrus: Each of these constants should have a doccomment explaining them.

Susurrus:
I'd prefer to wait on this until rust-lang/libc#682 gets addressed assuming you're willing to take that on. Were you planning on doing that?

mwanner:
> Were you planning on doing that?

Not really, sorry. The itch I was trying to scratch originally was `getifaddrs` on Linux. I fear I have neither enough BSDish boxes to test this nor enough spare time to hack on rust-lang/libc#682. Therefore my suggestion to at least provide support for OSes (correctly) supported by libc.

Susurrus:
@mwanner I'm fine with that as long as the code documents why we aren't exposing it everywhere and it references a bug filed with `libc`.

posborne:
@mwanner @asomers I happened to be working on a problem today where I needed just this functionality.  What are the next step on this one at this point?  I'm assuming the libc issue and related CI hole is still the gate at this point?

Susurrus:
@posborne I think there are two options, one is deliver this just for Linux or the second is get someone working on resolving the `libc` blocker and then enable it on all supported platforms. Since we don't have anyone stepping up to do the latter, I'd suggest going with the former. Strip out anything non-Linux, and put a big TODO in there along with a reference to the `libc` bug.

asomers:
Actually, the libc definition is only wrong for NetBSD.  FreeBSD is fine.

src/ifaddrs.rs
asomers: I just checked, and libc does define both `IFF_BROADCAST` and `IFF_POINTOPOINT` for most, if not all, platforms.

src/net/if_.rs
asomers: libc defines most of these constants for most platforms, not just Linux.  You should expand the `#[cfg()]` directives appropriately.

src/sys/socket/addr.rs
asomers: If it's not a public function, you should remove the `pub`

mwanner:
Thanks for the heads up, @asomers. Looks like `libc` has been updated, because the former `cfg` conditions were based on reading its source, too.

I made an attempt to update this PR to match current libc sources.

mwanner:
@asomers wrote:
> If it's not a public function, you should remove the pub

Given `libc::sockaddr` is public, users of `nix` might well want to obtain a `nix::sys::socket::SockAddr` from a `libc::sockaddr`. At least I found myself missing that constructor, before.

posborne:
I can take a look at Android support as well.  I happen to need to target Android aarch64 myself and bionic has had getifaddrs for some time but the libc bindings appear to be missing that on the rust side (explicitly).  No need to gate this on that work though, as it will definitely require libc changes.

Susurrus:
@posborne It looks like `getifaddrs()` just needs to be moved up into `src/notbsd/mod.rs` and removed from `linux/mod.rs` and `emscripten/mod.rs`. If you want to do this and help run any tests that come out of this on Android, then I think it wouldn't be hard to include that support as part of this PR.

src/ifaddrs.rs
```diff
@@ -0,0 +1,137 @@
...
+///             None => {}
+///         }
+///     },
+///     Err(err) => {
+///         println!("error getting interface addresses: {}", err)
+///     }
+/// }
+/// ```
+pub fn getifaddrs() -> Result<InterfaceAddressIterator> {
+    let mut addrs: *mut libc::ifaddrs = unsafe { mem::zeroed() };
```
Susurrus: Can this not be `mem::uninitialized()` instead?

src/net/if_.rs
```diff
@@ -17,3 +17,32 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+    /// Standard interface flags, used by `getifaddrs`
+    pub struct InterfaceFlags: libc::c_int {
+        IFF_UP;
```
Susurrus: Please add doccomments for each variant.

src/net/if_.rs
```diff
@@ -17,3 +17,32 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+        IFF_BROADCAST;
+        IFF_DEBUG;
+        IFF_LOOPBACK;
+        IFF_POINTOPOINT;
+        #[cfg(not(any(target_os = "freebsd", target_os = "dragonfly", target_os = "haiku")))]
```
Susurrus: Please alphabetize all OSes listed in `#[cfg`s. Additionally please use a whitelist instead of a blacklist for targets for consistency.

src/sys/socket/addr.rs
```diff
@@ -201,6 +201,23 @@ pub enum AddressFamily {
     Natm = libc::AF_NATM,
 }
 
+impl AddressFamily {
+    pub fn from_i32(family: i32) -> Option<AddressFamily> {
```
Susurrus: Please add a doccomment for this function as it's public.

src/sys/socket/addr.rs
```diff
@@ -201,6 +201,23 @@ pub enum AddressFamily {
...
+        match family {
+            libc::AF_UNIX => Some(AddressFamily::Unix),
+            libc::AF_INET => Some(AddressFamily::Inet),
+            libc::AF_INET6 => Some(AddressFamily::Inet6),
+            #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize the OSes.

src/sys/socket/addr.rs
```diff
@@ -688,6 +705,30 @@ impl SockAddr {
         format!("{}", self)
     }
 
+    /// Creates a `SockAddr` struct from libc's sockaddr.
+    pub unsafe fn from_libc_sockaddr(addr: *mut libc::sockaddr) -> Option<SockAddr> {
```
Susurrus: Do you actually need a `*mut` here?

src/lib.rs
```diff
@@ -43,6 +43,9 @@ pub mod poll;
 
 pub mod net;
 
+#[cfg(not(any(target_os = "android", target_os = "haiku")))]
```
Susurrus: We prefer to avoid blocklists with `#cfg`s, can you instead whitelist all appropriate platforms?

src/ifaddrs.rs
```diff
@@ -0,0 +1,137 @@
...
+            flags: InterfaceFlags::from_bits_truncate(info.ifa_flags as i32),
+            address: address,
+            netmask: netmask,
+            broadcast: None,
+            destination: None,
+        };
+
+        let ifu : *mut libc::sockaddr;
+
+        #[cfg(any(target_os = "linux", target_os = "emscripten", target_os = "fuchsia"))]
```
Susurrus: Probably want to use `cfg_if!` here instead.

src/ifaddrs.rs
```diff
@@ -0,0 +1,137 @@
...
+        let mut addr = InterfaceAddress {
+            interface_name: ifname.to_string_lossy().to_string(),
+            flags: InterfaceFlags::from_bits_truncate(info.ifa_flags as i32),
+            address: address,
+            netmask: netmask,
+            broadcast: None,
+            destination: None,
+        };
+
+        let ifu : *mut libc::sockaddr;
```
Susurrus: Remove the space before the colon.

src/ifaddrs.rs
```diff
@@ -0,0 +1,137 @@
+//! Query network interface addresses
+//!
+//! Uses the Linux and/or BSD specific function getifaddrs to query the list
```
Susurrus: Please put backticks around `getifaddrs` so it renders nicely with `rustdoc`.

src/sys/socket/addr.rs
```diff
@@ -688,6 +705,30 @@ impl SockAddr {
...
+                    InetAddr::V4(*(addr as *mut libc::sockaddr_in)))),
+                Some(AddressFamily::Inet6) => Some(SockAddr::Inet(
+                    InetAddr::V6(*(addr as *mut libc::sockaddr_in6)))),
+                #[cfg(any(target_os = "android", target_os = "linux"))]
+                Some(AddressFamily::Netlink) => Some(SockAddr::Netlink(
+                    NetlinkAddr(*(addr as *mut libc::sockaddr_nl)))),
+                #[cfg(any(target_os = "ios", target_os = "macos"))]
+                Some(AddressFamily::System) => Some(SockAddr::SysControl(
+                    SysControlAddr(*(addr as *mut sys_control::sockaddr_ctl)))),
+                // No other address families are supported, including netlink and syscontrol.
```
Susurrus: Is this a `TODO` or `FIXME` or just impossible? Please clarify this comment further.

src/ifaddrs.rs
mwanner: I didn't manage to use a `cfg_if!` for just these two statements, but extracted a small private function, now. Seems cleaner to me that way, but YMMV.

mwanner:
@Susurrus thanks for your review. I think I covered all points.

src/ifaddrs.rs
```diff
@@ -0,0 +1,142 @@
...
+            }
+            None => None,
+        }
+    }
+}
+
+/// Get interface addresses using libc's `getifaddrs`
+///
+/// Note that the underlying implementation differs between OSes and the nix
+/// crate doesn't support `AF_PACKET` entries. Therefore, the returned list
```
posborne: Would be a nice change in the future to support `AF_PACKET`/`AF_LINK` in order to be able to build something out for getting MAC addresses (useful for, among other things, getting a node id for UUID V1 generation).  Not gating for the first pass.

src/ifaddrs.rs
```diff
@@ -0,0 +1,142 @@
...
+use std::option::Option;
+
+use libc;
+
+use {Result, Errno};
+use sys::socket::SockAddr;
+use net::if_::*;
+
+/// Describes a single address for an interface as returned by `getifaddrs`.
+pub struct InterfaceAddress {
```
posborne: Can we implement or derive `Debug`/`Display` for this.  Also, I think deriving `Clone`, `PartialEq`, and `Hash` would be good.

src/sys/socket/addr.rs
```diff
@@ -201,6 +201,28 @@ pub enum AddressFamily {
     Natm = libc::AF_NATM,
 }
 
+impl AddressFamily {
+    /// Create a new `AddressFamily` from an integer value retrieved from `libc`, usually from
+    /// the `sa_family` field of a `sockaddr`.
+    ///
+    /// Currently only supports these address families: Unix, Inet (v4 & v6), Packet, Netlink
+    /// and System. Returns None for unsupported or unknown address families.
+    pub fn from_i32(family: i32) -> Option<AddressFamily> {
```
Susurrus: Would it not make more sense for this to be From<i32> instead of a custom method?

src/net/if_.rs
```diff
@@ -17,3 +17,68 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+        /// Interface is a loopback interface. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_LOOPBACK;
+        /// Interface is a point-to-point link. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_POINTOPOINT;
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "ios",
+                  target_os = "linux", target_os = "macos", target_os = "netbsd",
+                  target_os = "openbsd"))]
+        /// Avoid use of trailers. (see
```
Susurrus: Please put the doccomments for items above any `#[cfg`s for those items. Applies in several places here.

src/lib.rs
```diff
@@ -43,6 +43,10 @@ pub mod poll;
 
 pub mod net;
 
+#[cfg(any(target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd",
```
Susurrus: The `linux` is out of alphabetical order, and could you list these 1-per-line? When `cfg`'s get longer than a line we like to wrap with each target on its own line.

src/ifaddrs.rs
Susurrus: I think this comment could be made more clear as to why this isn't implemented. Was it because we were lazy or because the OSes differed enough it'd be quite difficult to expose a nice API here?

src/ifaddrs.rs
Susurrus: I agree with this as well, should definitely have `Clone`, `Copy`, `Debug`, `Eq`, `Hash`, and `PartialEq`. We shouldn't implement `Display` here and I don't think we should implement `Default` here either.

src/ifaddrs.rs
```diff
@@ -0,0 +1,142 @@
...
+        addr
+    }
+}
+
+/// Holds the results of `getifaddrs`.
+///
+/// Use the function `getifaddrs` to create this Iterator. Note that the
+/// actual list of interfaces can be iterated once and will be freed as
+/// soon as the Iterator goes out of scope.
+pub struct InterfaceAddressIterator {
```
Susurrus: We should definitely derive `Debug` here, but does it make sense to derive any of the other standard ones (`Clone`, `Copy`, `Eq`, `Hash`, and `PartialEq`)?

src/net/if_.rs
```diff
@@ -17,3 +17,68 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_NOTRAILERS;
+        #[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "fuchsia",
```
Susurrus: Please list targets one-per-line when wrapping these `cfg`s.

src/ifaddrs.rs
mwanner: In commit ef40c9c, I added `derive`s for `Clone`, `Eq`, `Hash`, and `PartialEq`, but not `Copy`. The struct uses `String`s, so I don't this that marker is applicable.

src/ifaddrs.rs
mwanner: `Clone` certainly doesn`t make much sense, as the `InterfaceAddressIterator` is used to govern the memory allocated by `getifaddrs`. I added `Debug`, `Eq`, `Hash`, and `PartialEq`.

src/sys/socket/addr.rs
mwanner: It only returns an `Option`, so it could be a [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) at best. But that's nightly-only...

src/ifaddrs.rs
mwanner: Oh, and `Debug`. I wasn't quite sure how much or in what format to return from `fn fmt`. Also note that `SockAddr` doesn't implement `Debug`, but `Display`. Is this really consistent?

src/ifaddrs.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+impl fmt::Debug for InterfaceAddress {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "InterfaceAddress ({:?})", self.interface_name)
+    }
+}
+
+cfg_if! {
+    if #[cfg(any(target_os = "emscripten", target_os = "fuchsia", target_os="linux"))] {
```
Susurrus: Needs spacing around the `=` before `linux`.

src/ifaddrs.rs
```diff
@@ -0,0 +1,152 @@
...
+            None => None,
+        }
+    }
+}
+
+/// Get interface addresses using libc's `getifaddrs`
+///
+/// Note that the underlying implementation differs between OSes. Only the
+/// most common address families are supported by the nix crate (due to
+/// lack of time and complexity of testing). For any entry not supported,
```
Susurrus: If I was looking at this as a user, my first thought would be oh, what are those specific address families? I think this information is embedded in `InterfaceFlags`, yes? If so, that should be mentioned in this comment.

src/ifaddrs.rs
```diff
@@ -0,0 +1,152 @@
...
+    } else {
+        fn get_ifu_from_sockaddr(info: &libc::ifaddrs) -> *const libc::sockaddr {
+            info.ifa_dstaddr
+        }
+    }
+}
+
+impl InterfaceAddress {
+    /// Create an `InterfaceAddress` from the libc struct.
+    fn from_libc_ifaddrs(info: &libc::ifaddrs) -> InterfaceAddress {
```
Susurrus: Shouldn't thinks consume the `libc::ifaddrs` by taking `ifaddrs` or `&mut ifaddrs`?

Susurrus:
Also, could you squash this down into 1 commit? Makes viewing the history of `nix` a lot easier and better supports `git bisect`.

src/ifaddrs.rs
mwanner: Oops, corrected.

src/ifaddrs.rs
mwanner: The entire struct is freed by calling `libc::freeifaddrs`, so you cannot really "consume" individual `ifaddrs`. I fail to see the benefits of requiring `&mut ifaddrs`. It doesn't need to be mutable in `from_libc_ifaddrs`.

src/ifaddrs.rs
mwanner: Maybe some of the `InterfaceFlags` are specific per address family, yes. However, it's the returned `SockAddr` type(s) that directly map to an `AddressFamily`, see `from_libc_sockaddr`. I tried to improve the comment.

mwanner:
Squashed into commit:3711dea4 and rebased onto `master`.

Susurrus:
LGTM. @asomers?

src/sys/socket/addr.rs
```diff
@@ -201,6 +201,28 @@ pub enum AddressFamily {
...
+        match family {
+            libc::AF_UNIX => Some(AddressFamily::Unix),
+            libc::AF_INET => Some(AddressFamily::Inet),
+            libc::AF_INET6 => Some(AddressFamily::Inet6),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            libc::AF_NETLINK => Some(AddressFamily::Netlink),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            libc::AF_PACKET => Some(AddressFamily::Packet),
```
asomers: If you're going to have `AF_PACKET` here, then you should have `AF_LINK` too.  It means the same thing.

src/net/if_.rs
```diff
@@ -17,3 +17,77 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
         Ok(if_index)
     }
 }
+
+libc_bitflags!(
+    /// Standard interface flags, used by `getifaddrs`
+    pub struct InterfaceFlags: libc::c_int {
```
asomers: It looks like you only included Linux's interface flags.  Could you please add, at a minimum, FreeBSD's?  That would be `IFF_RUNNING` and `IFF_OACTIVE` (yes, they're defined in FreeBSD, but currently missing from libc), as well as `IFF_SIMPLEX`, `IFF_CANTCONFIG`, `IFF_PPROMISC`, `IFF_MONITOR`, `IFF_STATICARP`, `IFF_DYING`, `IFF_RENAMING`, `IFF_LINK0`, `IFF_LINK1`, and `IFF_LINK2`?

mwanner: Hm.. fails on FreeBSD.

@asomers Want to wait for a `libc` release, or why did you ask for stuff you know isn't in `libc`, yet?

asomers: No need to wait for a release since nix's master branch depends on libc's master branch.  Just open a pull request to add the missing constants to libc.  We have to do it all the time for nix.

src/net/if_.rs
```diff
@@ -21,73 +21,307 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+        // bitmask: 0x10
         /// Interface is a point-to-point link. (see
         /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
         IFF_POINTOPOINT;
+
+        // bitmask: 0x20
         /// Avoid use of trailers. (see
         /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
         #[cfg(any(target_os = "android",
+                  target_os = "bitrig",
```
asomers: There's no need to worry about bitrig anymore.  A future nix PR will probably remove all references.  For now, it doesn't really matter whether or not you leave it in.
https://www.phoronix.com/scan.php?page=news_item&px=Bitrig-Fork-Dead

src/net/if_.rs
```diff
@@ -17,3 +17,311 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+        IFF_LINK0;
+        /// Multicast using broadcast.
+        #[cfg(any(target_os = "solaris"))]
+        IFF_MULTI_BCAST;
+
+        // bitmask: 0x2000
+        /// Is able to select media type via ifmap. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "fuchsia", target_os = "linux"))]
+        IFF_PORTSEL;
```
asomers: This is what caused your build failure.  `IFF_PORTSEL` is not defined on FreeBSD.

src/net/if_.rs
```diff
@@ -21,73 +21,307 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
                   target_os = "linux",
                   target_os = "macos",
                   target_os = "netbsd",
-                  target_os = "openbsd"))]
+                  target_os = "openbsd",
+                  target_os = "solaris"))]
         IFF_RUNNING;
+        /// Resources allocated.
+        #[cfg(any(target_os = "freebsd"))]
+        IFF_DRV_RUNNING;
```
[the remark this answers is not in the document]
mwanner: `libc` defines only `IFF_DRV_RUNNING` and `IFF_DRV_OACTIVE` for FreeBSD, so that's what I used. Given I don't really use - let alone know - FreeBSD, I don't quite feel in the position to open a PR on `libc`.

Especially since FreeBSD seems to prefer the `DRV` variant since version 8.1:
https://www.freebsd.org/cgi/man.cgi?query=ifnet

src/net/if_.rs
mwanner: Okay, thanks for the hint.

src/net/if_.rs
mwanner: Ah, I see. That's clearly wrong. Corrected.

src/net/if_.rs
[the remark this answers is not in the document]
asomers: The distinction between the `DRV` variants and the non-`DRV` variants only matters for kernel code.  You can read the gory details at https://svnweb.freebsd.org/base?view=revision&revision=148886 .  It's perfectly fine to use the non-`DRV` variants in userland code, especially since their values are identical.  For a cross-platform library like nix, the non-`DRV` variants are better, because they're portable.  The current definitions are at https://svnweb.freebsd.org/base/head/sys/net/if.h?view=markup#l170 .  You can go ahead and submit a PR to libc based on those definitions, even if you don't have a FreeBSD VM to test on.  If you like, CC me as a reviewer on the libc PR.

src/net/if_.rs
```diff
@@ -17,3 +17,325 @@ pub fn if_nametoindex<P: ?Sized + NixPath>(name: &P) -> Result<c_uint> {
...
+        /// No arp protocol, L2 destination address not set. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_NOARP;
+
+        // bitmask: 0x0100
+        /// Interface is in promiscuous mode. (see
+        /// [`netdevice(7)`](http://man7.org/linux/man-pages/man7/netdevice.7.html))
+        IFF_PROMISC;
+
+        // bitmask: 0x0200
```
Susurrus: Please remove all these "bitmask: 0xXXX" constants and the blank line between all enum entries here.

mwanner: Those greatly simplified assembling the list of `IFF_` constants and will help maintaining it in the future. Can you please suggest a viable alternative? Dropping that information is not a good option, IMO.

Susurrus: These are all listed in `/usr/include/linux/if.h`, so was this helpful for the other platforms? I don't know how this will help make things more maintainable except be incorrect or out of date possible since these values will never likely be updated. I think it's best we remove them.

mwanner: @Susurrus Yes, because I could map to the other OSes' header files by bitmask and ensure I got the complete list covered (these flags aren't alphabetically ordered, but by value). If you want to take over and keep the list in sync with libc, I'm fine with having the bitmasks removed. However, I'm certainly not ever gonna do it again w/o the information these hints provide.

That being said, maybe there's a better way to "encode" them that better conforms to the style guide applicable for nix.

lucab:
@Susurrus @asomers what's the status and differences between this PR and https://github.com/nix-rust/nix/pull/764? I'm also in the need of support for ifaddrs features and willing to give it a try, but I'm not sure which of the two PRs to look at.

src/net/if_.rs
Susurrus: I understand how it helped you to compile the list, but it won't be helpful moving forward in maintaining it as I mentioned above. That being said, I'm not going to throw too much more of a fight over it as I don't think this is that important. Please do, however, remove the blank lines in between the variants.

Susurrus:
@lucab It's basically whichever gets there first at this point. I would recommend we focus our efforts on this PR since #764 hasn't been touched in months and I don't expect it to ever resolve.

@mwanner Please remove the blank lines from `InterfaceFlags` and squash all commits into 1, and then let's get @asomers and @lucab to take a pass at this and we'll be golden!

src/ifaddrs.rs
```diff
@@ -0,0 +1,154 @@
...
+            None => None,
+        }
+    }
+}
+
+/// Get interface addresses using libc's `getifaddrs`
+///
+/// Note that the underlying implementation differs between OSes. Only the
+/// most common address families are supported by the nix crate (due to
+/// lack of time and complexity of testing). The address familiy is encoded
```
lucab: s/familiy/family/

src/ifaddrs.rs
```diff
@@ -0,0 +1,154 @@
...
+///         match ifaddr.address {
+///             Some(address) => {
+///                 println!("interface {} address {}",
+///                     ifaddr.interface_name, address);
+///             },
+///             None => {}
+///         }
+///     },
+///     Err(err) => {
+///         println!("error getting interface addresses: {}", err)
```
lucab: I would simplify this example in order to show just the `getifaddrs`, by doing:
 * a `try!` / `unwrap` / `expect` on the `getifaddrs`
 * a `while let` on the interator

The resulting example would be less indented and more compact:
```
let addrs = getifaddrs()?;
while let Ok(a) = addrs.next() {
  println(...)
}
```

lucab:
I agree it would be nice to have `AF_PACKET` as it is pretty ubiquitous, but personally I don't it right now. I'd be fine having a TODO line in place for that and a followup ticket detailing what's left to be done.

Except for typos and very verbose example, this PR looks nice at a first glance. Thanks for this!

src/ifaddrs.rs
mwanner: Good point, I stripped the error handling.

Note however, that the iterator is not over a `Result<..>`, but over multiple `InterfaceAddress` structs, where always get an `interface_name`, but for which we do not (yet) support the address type. Therefore, neither `while let Ok(..)` nor `while let Some(..)` work and the simple `for` loop likely is the easiest. I tried to clarify the example in that regard.

mwanner:
I guess this branch currently fails due to d374a1ec from the `master` branch. @Susurrus any visibility change I need to be aware of? Or why is this suddenly not finding IFF_BROADCAST, anymore.

lucab:
@mwanner #801 changed the scope of bit-constants, see https://github.com/nix-rust/nix/pull/801#issuecomment-346256220. Those are now `InterfaceFlags::IFF_*`. You'd see this with a `cargo update`.

mwanner:
@Susurrus, @asomers, @lucab: I think I covered all concerns and the result now passes CI tests as well.

As two of you voted against the bitmask comments and the corresponding grouping by empty lines, I removed those. Given the lack of that information, I wonder if alphabetical ordering of the `IFF_*` flags would ease future maintenance. Opinions?

To track progress on the requested support of `AF_PACKET`, I opened #809 (which I started hacking on, BTW, nothing ready to push, though.)

Given @lucab's simplified example wasn't correct, I fear the `getifaddrs` interfaces is less than obvious. IMO that justifies the verbosity of the example. You could even argue that the `netmask`, `broadcast`, and `destination` addresses are missing completely...

Compared to #764, this variant is closer to the libc interface, not providing a `HashMap` collecting all entries per interface, but (hopefully) working without any additional memory allocation. That might make the interface of `getifaddrs` look a bit clumsy, though.

Susurrus:
> As two of you voted against the bitmask comments and the corresponding grouping by empty lines, I removed those. Given the lack of that information, I wonder if alphabetical ordering of the IFF_* flags would ease future maintenance. Opinions?

The way maintenance of this will work is that someone will find a constant that's missing. So they'll just add them. There really isn't going to be someone doing a full audit of these per-platform, so I think we're fine here. I think you're overly concerned with the maintenance burden of that bitmask.

> To track progress on the requested support of AF_PACKET, I opened #809 (which I started hacking on, BTW, nothing ready to push, though.)

Perfect, thanks.

>Given @lucab's simplified example wasn't correct, I fear the getifaddrs interfaces is less than obvious. IMO that justifies the verbosity of the example. You could even argue that the netmask, broadcast, and destination addresses are missing completely...

I don't think we need a complete solution on the first pass, I think this one is pretty good. And I agree with keeping a verbose example. My only question here would be how to make sure that *incorrect* usage of this API fails at compile time. I think @lucab's example would have failed at compile time, so I think concerns about confusion about this API might be unfounded.

> Compared to #764, this variant is closer to the libc interface, not providing a HashMap collecting all entries per interface, but (hopefully) working without any additional memory allocation. That might make the interface of getifaddrs look a bit clumsy, though.

That's a good thing. `nix` aims to be as thin of a wrapper as possible around `libc`, so it won't always deliver the *most* ergonomic APIs. Then again, allocating is really only a big problem if you're doing high-iterations of things. For example, my implementation of the Termios API in `nix` does have some overhead, but considering that is rarely dealt with at high frequency, it's not an issue. We use a relatively subjective measurement for how "close to the metal" `nix` APIs should be.

Overall, I think this looks pretty good. I suggest we merge it. I'll wait until @asomers responds (@lucab feel free to chime in here as well if you have anything to add).

asomers:
Yep, LGTM.

bors r+

LeoTindall:
Thanks for doing this, and sorry my PR died. I over estimated my ability to do open source while in school, but I'm glad to see that the functionality has been added!

posborne:
Great to see this land! 🎉   Thanks for the contribution @mwanner and @Susurrus / @asomers for the diligent reviews.

PR #668, opened 2017-07-10
Add waitpid flags for *BSD operating systems

myfreeweb:
Works on FreeBSD 11, FreeBSD -CURRENT and NetBSD 7. OpenBSD -CURRENT and DragonFly 4.8 don't compile for other reasons, I should probably fix that stuff too…

Prompted by [ion's use of WCONTINUED](https://github.com/redox-os/ion/commit/8fc7729400b7ba8b2bad8e5248be61b9a39472eb#diff-0902043a20bccae7d1c9e1e03e51ff59R163) :)

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -13,9 +13,48 @@ mod ffi {
 }
 
 #[cfg(not(any(target_os = "linux",
-              target_os = "android")))]
+              target_os = "android",
+              target_os = "openbsd",
+              target_os = "bitrig",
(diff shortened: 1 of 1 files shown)


Susurrus:
@myfreeweb Thanks for the PR! This work overlaps heavily with #651. Could you comment on that PR with what BSD platforms support the various `WaitPidFlags` constants?

myfreeweb:
Oh I completely missed that one. Sure.

PR #670, opened 2017-07-11
Improve ioctl docs

### LINKED ISSUE #573: Additional examples for ioctl
Some additional examples for ioctl would be very helpful. It's taken me a bit to figure out how I'm supposed to do BLKSZGET. Thanks.

### LINKED ISSUE #641: Improve documentation for ioctl
I'm working on converting my crate to use `nix` instead of `ioctl-rs` as `nix` is better maintained. The big issue here is that the documentation for using `ioctl`s with `nix` is not good from a user's point of view. Here are some issues I ran into:

  * The difference between using `ioctl!` and using `ior!`, `iow!`, `iorw!`, and `ioc` is not listed clearly. I'm sure one is preferred over the other, but I'm not certain which. The example only uses `ioctl!`.
  * It should be clarified that ioctl constants should be upstreamed into `libc` and then those should be used as inputs/outputs. The example doesn't do that, so we'll have tons of people reimplementing everything per-crate.
  * The `ioctl!` example only uses `read` and `write`. Looking at the macro there are 8 forms of it, but the documentation doesn't actually describe any of them.
  * It's unclear how to use `ioc_dir` and friends. There are no docs and they don't take clear types, but there are constants that have similar names in the module, so maybe those are related?
  * There are various constants that are also a part of the `ioctl` module, but those don't seem attached to these functions.
  * `nix` doesn't provide helpful wrapper functions for ioctls even if they're quite common (`TIOCEXCL` would be an example) and well-defined. I'd suggest that wrapper functions be provided for well-defined ioctls. This may be controversial, as I think it's come up before, because ioctl's are unbounded and defined by kernel drivers. So maybe there is a role for an ioctl crate that does this on top of `nix`.

Susurrus:
Integrates suggestions and comments made in #641. Basically we hid a lot of the internal workings of things and also revised the docs to make it more clear the exact API that `nix` is exposing.

There is a small amount of code cleanup I did in the macros. I also fixed the `bad` version of the `ioctl!` macro and also added a `bad none` version for use with no-data, hardcoded-number `ioctl`s.

Would appreciate any and all feedback. Please especially fetch this code locally and generate the pretty docs for it (`cargo doc --no-deps --open`) so you can see what our users will see.

Closes #641.
Closes #573.

cc @cmr @roblabla


Susurrus:
I need to see if the changes I made broke anything across all of our platforms and I also want to have a go at using this within my `serialport-rs` crate.

@posborne if you'd like to check that this refactoring doesn't break things for you in `rust-spidev`, that'd also be appreciated.

Susurrus:
Alright, I removed a few more constants that were unnecessary and also cleaned out two ctypes that were exported in the root for some reason. The generated docs are looking better and better!

Susurrus:
There is also no `bad read` version of the `ioctl!` macro, so the datatype is always a `*mut, even if it's a reading `ioctl` and should be `*const`. A minor annoyance, but still annoying.

Susurrus:
And the datatype for `bad` is always a `*u8` when it should really be a proper datatype.

src/sys/ioctl/mod.rs
```diff
@@ -24,79 +22,119 @@
...
+//! The `TIOCEXCL` `ioctl` that's part of the termios API can be implemented as:
+//!
+//! ```
+//! # #[macro_use] extern crate nix;
+//! # use nix::libc::TIOCEXCL as TIOCEXCL;
+//! ioctl!(bad none tiocexcl with TIOCEXCL);
+//! # fn main() {}
+//! ```
+//!
+//! More examples on using `ioctl!` can be found in the [rust-spidev crate](https://github.com/posborne/rust-spidev).
```
posborne: Minor, this has moved over to https://github.com/rust-embedded/rust-spidev

src/sys/ioctl/mod.rs
```diff
@@ -24,79 +22,119 @@
...
+//!
+//! This library provides the `ioctl!` macro, for binding `ioctl`s. This macro generates unsafe
+//! functions that can then be used for calling the ioctl. This macro has a few different ways it
+//! can be used depending on the specific ioctl you're working with.
 //!
-//! This library provides the `ioctl!` macro, for binding `ioctl`s.
-//! Here's a few examples of how that can work for SPI under Linux
-//! from [rust-spidev](https://github.com/posborne/rust-spidev).
+//! A simple `ioctl` is `SPI_IOC_RD_MODE`. This ioctl works with the SPI interface on Linux. This
+//! specific `ioctl` reads the mode of the SPI mode (one of 4 values, so 2 bits each) as a `u8`.
```
posborne: reads the mode of the SPI ~~mode~~ device

Susurrus:
@posborne Thanks for the review. both are now fixed in the latest revision.

Susurrus:
Alright, I think this is as much work as needs to happen to really clean this up. @roblabla and @cmr would appreciate any feedback from you guys on this.

posborne:
Yep, changes look good.  I'll leave this open for now in case otherwise want to review.  Otherwise r=me is fine.

Susurrus:
Thanks @posborne. I also wanted to get a downstream repo using `ioctl!` onboard to make sure these changes were sound. It looks like `rust-spidev` hasn't had any work in a long time, as in it's using some pretty old dependencies. Is it still used? If so, would you want to try to update it to use the revised `ioctl!` macro?

src/sys/ioctl/mod.rs
```diff
@@ -1,102 +1,175 @@
...
+//!   * Do whatever else the device driver creator thought made most sense.
+//!
+//! `ioctl`s are synchronous system calls and are similar to read and write calls in that regard.
+//! They operate on file descriptors and have an identifier that specifies what the ioctl is.
+//! Additionally they may read or write data and therefore need to pass along a data pointer.
+//! Besides the semantics of the ioctls being confusing, the generation of this identifer can also
+//! be difficult.
+//!
+//! Historically `ioctl` numbers were arbitrary hard-coded values. This changed to a more-ordered
+//! system where the ioctl numbers had various subcomponents:
```
roblabla: *has* various subcomponents ? Also, AFAIK, this is actually linux-specific, and the change was made a *loooong* time ago (before 2.6).

Maybe say "In Linux, most ioctl numbers are made of various subcomponents" instead ?

Maybe add a link to http://elixir.free-electrons.com/linux/latest/source/Documentation/ioctl/ioctl-number.txt for more information ?

roblabla: (As it turns out, FreeBSD also has _IOR/_IOW, so I guess it got adopted by the wider unix community ? I can't find them referenced anywhere in POSIX though)

src/sys/ioctl/mod.rs
```diff
@@ -1,102 +1,175 @@
...
+//! pub unsafe fn spi_read_mode(fd: c_int, data: *mut u8) -> Result<c_int> {
+//!     let res = libc::ioctl(fd, ior!(SPI_IOC_MAGIC, 1, mem::size_of::<u8>()), data);
+//!     Errno::result(res)
 //! }
+//! # fn main() {}
+//! ```
+//!
+//! The return value for `ioctl` functions generated by the `ioctl!` macro are assumed to return -1
+//! on error and everything else is a valid return value. If the return value needs to be changed,
+//! you can use `Result::map` on the return value in a helper function.
```
roblabla: Maybe specify that when the ioctl function returns -1, we will use `errno` to build our Error value.

src/sys/ioctl/mod.rs
```diff
@@ -1,102 +1,175 @@
...
 //!
-//! // doctest workaround
-//! fn main() {}
+//! There are equivalent forms for `write`, `none` (no data in or out), and `readwrite`. The mode
+//! for a given `ioctl` should be clear from the documentation if it has good documentation.
+//! Otherwise it will be clear based on the macro used to generate the `ioctl` number where
+//! `_IOC`, _IOR`, `_IOW`, and `_IORW` map to "none", "read", "write", and "readwrite"
```
roblabla: `_IO` maps to "none", not `_IOC`. `_IOC` is a macro that puts all the components together to create a single IOCTL number.

src/sys/ioctl/mod.rs
```diff
@@ -1,102 +1,175 @@
...
 //!
-//! Most `ioctl`s have no or little documentation. You'll need to scrounge through
-//! the source to figure out what they do and how they should be used.
+//! Some `ioctl`s work with entire arrays of elements. These are supported by the `buf` suffix in
+//! the `ioctl!` macro which can be used by specifying `read buf`, `write buf`, and
+//! `readwrite buf`. Note that there are no "bad" versions for working with buffers. The generated
+//! functions include a `len` argument to specify the number of elements (where the type of each
+//! element is specified in the macro).
 //!
+//! Again looking to the SPI `ioctl`s on Linux for an example, `SPI_IOC_NR_TRANSFER` is an `ioctl`
```
roblabla: Linux doesn't specify `SPI_IOC_NR_TRANSFER` AFAIK. I guess this comes from `rust-spidev` ? Linux specifies `SPI_IOC_MESSAGE`, which is a macro that takes one argument : 

```c
#define SPI_IOC_MESSAGE(N) _IOW(SPI_IOC_MAGIC, 0, char[SPI_MSGSIZE(N)])
```

You can find this here : http://elixir.free-electrons.com/linux/latest/source/include/uapi/linux/spi/spidev.h#L116

src/sys/ioctl/mod.rs
```diff
@@ -1,102 +1,175 @@
...
+//!     Errno::result(res)
+//! }
+//! # fn main() {}
+//! ```
+//!
+//! Finding ioctl documentation
+//! ---------------------------
+//!
+//! For Linux, look at your system's headers. For example, `/usr/include/linux/input.h` has a lot
+//! of lines defining macros which use `_IOR`, `_IOW`, `_IOC`, and `_IORW`. Some `ioctl`s are
```
roblabla: and `_IO`. `_IOC` seems to be mostly used for arrays BTW.

src/sys/ioctl/mod.rs
Susurrus: Yes, I knew that when I added this, but I needed some form of a name for this ioctl, so I just reused the same name that spidev used. Know of a good way to rephrase that that doesn't use `SPI_IOC_NR_TRANSFER`?

Susurrus:
@roblabla Thanks for taking a look! I do have concerns that someone could pass in the wrong size in `len` for the `ioctl!(* buf, ...)` case. I was thinking like about an `assert_ne!(len, len * ::std::mem::size_of::<$ty>())` to the `buf` `ioctl`s, though I wonder if that would ban some valid use cases. This change I have in here is a breaking change that won't break at compile time, but subtly at runtime more likely.

roblabla:
I don't think it's a good idea to do that (what about if I want to pass an array of 1 `u8` ?).

OK, maybe a stupid idea, but maybe the `buf` cases could actually take a slice ? I doubt there are many cases of `buf` where a NULL pointer would be sent (Please correct me if I'm wrong ?), and it'd avoid degrading back to C ergonomics.

Susurrus:
@roblabla Yeah, I was thinking having it as a slice would be good. This is actually how `rust-spidev` uses these ioctls, it's all slices. But again, ioctls is such a long list, I'm not certain if NULL is necessary and I don't want to break stuff. In theory people who upgrade will have failures in some of their `ioctl!` macro calls and will need to go through and fix some, so it's unlikely this wouldn't be caught by that change.

I think it's reasonable to go ahead and make `* buf` versions take a buf slice, so I'll go and change that. Also, I should have addressed all of your other comments in the new version I just pushed. Thanks for those edits!

src/sys/ioctl/mod.rs
roblabla: Wouldn't just doing 

```
Again looking to the SPI `ioctl`s on linux for an example, [`SPI_IOC_MESSAGE(n)`](http://elixir.free-electrons.com/linux/v4.12.2/source/include/uapi/linux/spi/spidev.h#L116) is an `ioctl` that queues up...
```

work ? (Also, maybe adding links to the kernel headers for various documentation bits would be nice, so readers can actually understand what symbols we're talking about).

src/sys/ioctl/mod.rs
```diff
@@ -108,27 +107,64 @@
...
+//!
+//! Again looking to the SPI `ioctl`s on Linux for an example, `SPI_IOC_NR_TRANSFER` is an `ioctl`
+//! that queues up multiple SPI messages, so it writes an entire array of `spi_ioc_transfer`
+//! structs. This can be implemented like:
 //!
 //! ```
 //! # #[macro_use] extern crate nix;
-//! # use nix::libc::TIOCEXCL as TIOCEXCL;
-//! ioctl!(bad none tiocexcl with TIOCEXCL);
+//! # const SPI_IOC_MAGIC: nix::libc::c_int = 'k' as nix::libc::c_int;
```
roblabla: Another thing that just popped up. Maybe we shouldn't hide this ? I fear that reading the docs, it's not immediately obvious that SPI_IOC_MAGIC is just another expression if we hide this.

Also, instead of 'k' as `nix::libc::c_int`, I think `b'k'` works better :)

Susurrus: `SPI_IOC_MAGIC` is actually defined in `/usr/include/linux/spi/spidev.h`, so I could expose it and just add a comment for it. `SPI_IOC_NR_TRANSFER` isn't defined, however, but I think I should expose that as well.

Susurrus:
@roblabla I tried to be more explicit about the various `SPI` constants used. Hopefully this is a lot more clear. Let me know what you think!

roblabla:
Looks great ! I find this much clearer than the older ones :)

Susurrus:
@oposborne This has changed a bit since you r+'d this, you want to have another look over it? If not, just let bors know.

Susurrus:
Hmmm, this looks wrong for `ioctl!(write...)` given that #626 changed it. Will need to look into this a bit more.

Susurrus:
I've been looking over #626 to understand why that change was made, and while it looks like the docs for `ioctl` all suggest that the third argument is a pointer (though they argue over what kind of pointer it is), it looks like some `ioctl`s actually shove an integer value in there instead of a valid pointer. So that's cool.

I think the question now is how do we want to support that? I know that both forms are needed, so maybe we're just bikeshedding on names? Maybe split `write` into `write_val` and `write_ptr` or maybe `write_int` and `write`, since this is supposed to be a pointer, the default should stay for the pointer type.

cc @bkchr as you originally reported #626 and would like to get your feedback here.

bkchr:
As I understood the api correctly, before my change we had 'write' and 'write buf'. Both took a *const pointer as argument. After my change, 'write" takes the value by ref and 'write buf' still as *const pointer. That should be enough to cover all cases?

Susurrus:
@bkchr In this PR I changed the `... buf` versions to take slice references instead of raw pointers. So with this PR there's no way to pass just a straight number as a pointer with `ioctl!`. I did that to expose some more type safety to users of the functions. I haven't changed the standard variants to take references, instead reverting them to take pointers, as I think there might be some `ioctl`s that need to take a NULL.

Therefore I'm trying to figure out a new variant to add that will support passing an int straight to the ioctl instead of a pointer. I think having different versions for all of the different `ioctl!` semantics should be nicer as it's more explicit and we can have better type safety through more confined arguments.

bkchr:
Ohh sorry for that! Did not looked over your exact changes. More type safety is always better! :)

Then I would recommend to use write_val and write_ptr, that should really help the user to understand what the function does.

Susurrus:
@bkchr No worries, this stuff's a little complicated to. So I'm trying to figure out if `write_val` should support any datatype or just `c_int`. I'm pretty certain only `c_int` makes sense given what I've seen (man ioctl_list shows all pointers except for a range of ones that take int).

bkchr:
Hmm I would make it configurable, so don't make it explicitly int. You never know what the future brings ^^

Susurrus:
This also makes me think we should change the `ioctl(_ buf` variants to be `_buf` instead as one work to better match the `write_int` and `write_ptr` variants we have now. Seems a little weird not to do that.

bkchr:
I don't understand what you want to say with your last comment ^^ 

And if that was not clear, I would propose to use write_val and write_ptr.

Susurrus:
@bkchr I was saying that I didn't like keeping `buf` as a suffix, so I changed those variants to be `read_buf`, `write_buf`, `readwrite_buf` instead. This better matches the `write_ptr` and `write_int` types that we added.

I also kept the name `write_int` and hard-coded the type to `libc::c_int` because I don't want to sacrifice usability and sane semantics now in the off chance that `ioctls` are changed later. We can cross that bridge when we come to it so-to-speak.

Hopefully this finally addresses all use cases for this refactoring as it's been quite a ride!

bkchr:
Okay :) I hope that your changes get merged :)

Susurrus:
Just updated because of conflicts from some merges last night.

@roblabla @posborne Any last comments on these changes before I merge them?

src/sys/ioctl/mod.rs
```diff
@@ -134,63 +225,96 @@ macro_rules! convert_ioctl_res {
...
-        pub unsafe fn $name(fd: $crate::sys::ioctl::libc::c_int,
-                            val: *mut $ty,
-                            len: usize)
-                            -> $crate::Result<$crate::sys::ioctl::libc::c_int> {
-            convert_ioctl_res!($crate::sys::ioctl::ioctl(fd, ior!($ioty, $nr, len) as $crate::sys::ioctl::libc::c_ulong, val))
+    (read_buf $name:ident with $ioty:expr, $nr:expr; $ty:ty) => (
+        pub unsafe fn $name(fd: $crate::libc::c_int,
+                            data: &mut [$ty])
+                            -> $crate::Result<$crate::libc::c_int> {
+            convert_ioctl_res!($crate::libc::ioctl(fd, ior!($ioty, $nr, data.len() * ::std::mem::size_of::<$ty>()) as $crate::sys::ioctl::ioctl_num_type, data))
```
asomers: Is the size of `data` fixed when the macro is invoked, or can different length slices be passed into the generated function on different calls?

src/sys/ioctl/platform/linux.rs
```diff
@@ -1,12 +1,27 @@
-pub const NRBITS: u32 = 8;
-pub const TYPEBITS: u32 = 8;
+/// The datatype used for the ioctl number
+#[cfg(any(target_os = "android", target_env = "musl"))]
+#[doc(hidden)]
+pub type ioctl_num_type = ::libc::c_int;
```
asomers: Really?  Musl defines the request as an int, even when it's running on Linux?  Baffling!

src/sys/ioctl/mod.rs
Susurrus: Different length slices can be passed in on different calls. The ioctl for `SPI_IOC_MESSAGE` that I give as an example is used like that (you'd use it like `ioctl(file, SPI_IOC_MESSAGE(2), xfer)` where `xfer` is a `spi_ioc_transfer[2]`). You tell it how many messages and that's actually part of the calculated ioctl number. Would adding the C-equivalent help make this more clear or should I explain it better in the docs somehow?

src/sys/ioctl/platform/linux.rs
Susurrus: I'm not certain if this is sarcastic or not, but yes, [musl defines it as an int](https://github.com/bminor/musl/blob/e738b8cbe64b6dd3ed9f47b6d4cd7eb2c422b38d/include/sys/ioctl.h#L9). Why are you surprised by that?

Susurrus:
Thanks for the review, @asomers! Regarding your specific questions:

> How do you use ioctl! for ioctls that have more than 3 arguments?

I've never seen an example or docs that suggest this is a thing on POSIX systems, but I'm not an expert on `ioctl`s, so I could definitely be wrong here. While I always see it defined like `int ioctl(int fd, unsigned long request, ...);`, with the third argument unspecified, the docs always say it's a pointer and never refer to any additional arguments.

That being said, I would consider it out-of-scope for this change to handle more than 1 `request` argument given that the previous code didn't either and this was **supposed** to be simply a documentation PR.

> How do you use `ioctl!` for read or write ioctls that take a pointer to struct argument? I think I understand, but it would be helpful to add such an example to the compile tests near the top of test_ioctl.rs

You use `ioctl!(read ..)` and `ioctl!(write_ptr ...)` for this case. I provide an example working with reading using a struct pointer: `ioctl!(bad read tcgets with TCGETS; termios);`. Using `write_ptr` will look similar to the example I give in the "Working with arrays" section: `ioctl!(write_buf spi_transfer with SPI_IOC_MAGIC, SPI_IOC_TYPE_MESSAGE; spi_ioc_transfer);`, where instead of a primitive you provide the datatype name. I wouldn't think that would need very much explicit documentation because it follows exactly from how primitives are used.

If instead you're referring to how to *use* the generated function, there are two examples I have in here that show the generated function which both take pointers as their second arguments, so I don't provide a specific usage example of it.

Susurrus:
@asomers I just pushed a new commit that adds simple compile-tests of the generated functions for all the `ioctl`s that it was easy to find examples for. So I skipped `bad readwrite`, `read_buf`, or `readwrite_buf`, as I couldn't find any clear examples of these. If anyone knows of any off the top of their head, please chime in and I'll add them.

Though I do think that `bad readwrite` can probably be removed as it's the same as `bad write`, I just added it for consistency. It doesn't offer any type safety over `bad write` either. The write versus readwrite distinction is only used for newer ioctls that use a different ioctl number generation for both.

Susurrus:
@asomers I'd love to get these tests working on FreeBSD as well since we claim to support more than just Linux/Android for ioctls. Do you have any suggestions for some `ioctl`s that I could use?

Susurrus:
We're getting runtime failures because the ioctls numbers are being rejected by the kernel. I think these should work across all Linux platforms. The `ioctl`s that our macros are generating, however, vary per target, which I'm not certain is correct.

So I decided to try something a little ambitious here; I'm going to try to use a build script to generate the ioctl numbers for each platform by compiling C functions that return various ioctl numbers. Then we can call into those functions from our Rust ioctl test code. This should get rid of those annoying hard-coded values and also make sure things are correct when we add new platforms. Since these calculations don't rely on any specific `ioctl`s being available on the platform, just generating the numbers, they should be perfectly portable.

asomers:
I can't find any examples of ioctls that take more than 3 arguments either, despite it technically being allowed.  And you're right, the man page suggests that it shouldn't happen.  So let's forget about supporting that weird use case.  And the `tcgets` example looks good to me.

As for FreeBSD testing, most of the terminal control ioctls are somewhat portable, and most should work on ptys.  Here's a partial list of ioctls that I found defined on both FreeBSD and Linux: `TIOCEXCL TIOCNXCL TIOCSCTTY TIOCGPGRP TIOCSPGRP TIOCOUTQ TIOCSTI TIOCGWINSZ TIOCSWINSZ TIOCMGET TIOCCONS TIOCPKT TIOCNOTTY TIOCSETD TIOCGETD TIOCSBRK TIOCCBRK TIOCGSID TIOCGPTN TIOCSIG`.

As for your test failures, I think it's untrue that all ioctls have the same value across Linux platforms.  Some platforms, like mips, are unnecessarily divergent.

Shouldn't gcc be a dev-dependency instead of a build-dependency?

Susurrus:
> As for FreeBSD testing, most of the terminal control ioctls are somewhat portable, and most should work on ptys. Here's a partial list of ioctls that I found defined on both FreeBSD and Linux: TIOCEXCL TIOCNXCL TIOCSCTTY TIOCGPGRP TIOCSPGRP TIOCOUTQ TIOCSTI TIOCGWINSZ TIOCSWINSZ TIOCMGET TIOCCONS TIOCPKT TIOCNOTTY TIOCSETD TIOCGETD TIOCSBRK TIOCCBRK TIOCGSID TIOCGPTN TIOCSIG.

Cool, I'll expose all of the TTY ioctls I'm using for testing on FreeBSD as well.

> Shouldn't gcc be a dev-dependency instead of a build-dependency?

`build-dependencies` is for dependencies for the build-scripts themselves (so the `build.rs` build, not for compiling the actual library). `dev-dependencies` are not available in `build.rs`.

asomers:
But build-dependencies are going to be pulled in every time nix is compiled by a dependent crate, right?  It would be a shame to make all of our dependent crates depend on gcc, when it's only our tests that need to build a C extension.

Susurrus:
I think the solution is to integrate this into `nix-test`. I think that's why it actually exists, to work around this. If we go down this road, I'll integrate this testing into `nix-test` so it won't be a dependency for building.

cmr:
@Susurrus see also https://github.com/cmr/ioctl/tree/master/etc, the original source of this macro. I have scripts there for parsing the ioctl numbers out of the C headers. https://github.com/cmr/ioctl/blob/master/src/platform/linux-generated-x86_64.rs has all of them, but most of them are commented out for lacking the structs.

Susurrus:
@cmr Thanks, but I think this solution will work out. I'd prefer to not manually parse and instead run the C code directly. As an aside, why did you deprecate your `ioctl` crate? It would have been nice to have it around to provide these implementations for my `serialport-rs` crate and other crates would likely use it.

But even with these checks passing on all targets, it appears that the ioctl number isn't liked by the kernel. I wonder if those ioctl's really don't exist on those platforms. I'll have to dig into this further, but I think I'm going to reduce this PR down a bit and we can look into improving ioctl testing in another one.

Susurrus:
Okay, so I've removed the comparison of generated ioctl numbers and made `ENOSYS` be valid for the runtime calls for now. This should pass CI at this point and then we can merge!

@asomers The runtime ioctl calls are passing for FreeBSD, so that's cool!

asomers:
Cool!  I think we have a winner.

bors r+

posborne:
Thanks for pushing this through @Susurrus and everyone else.  I've been a bit busy with other things lately, but I am appreciative of the diligent work you guys have been doing.

Susurrus:
Thanks, @posborne! And I appreciate you chiming in with your edits. Definitely helps!

PR #673, opened 2017-07-12
From<io::Error> for nix::Error

vincenthage:
Allows implicit conversions from `std::io::Error` to `nix::Error`.
Now you can do this:
```rust
use std::fs::FileType;
use nix::Result;

fn get_file_type(path: &Path) -> Result<FileType> {
    // `symlink_metadata` returns an `std::io::Error`
    let metadata = path.symlink_metadata()?;

    metadata.file_type()
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,7 @@ use std::path::{Path, PathBuf};
 use std::os::unix::ffi::OsStrExt;
 use std::fmt;
 use std::error;
+use std::io::{Error as IOError, ErrorKind as IOErrorKind};
 use libc::PATH_MAX;
 
 /// Nix Result Type
@@ -92,6 +93,8 @@ pub type Result<T> = result::Result<T, Error>;
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum Error {
     Sys(errno::Errno),
+    /// Error received from one of std::io's functions; it does not have a corresponding errno.
+    IOError(IOErrorKind),
     InvalidPath,
     /// The operation involved a conversion to Rust's native String type, which failed because the
     /// string did not contain all valid UTF-8.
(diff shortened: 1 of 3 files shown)


Susurrus:
There's no real code that uses this. And nothing outside of this crate will have a use for this. I'd rather see this as part of a larger PR that requires this functionality, otherwise this is just adding dead code.

vincenthage:
There might be code outside of `nix` that uses it though. And because neither the `From` trait nor `NixError` are owned outside of `nix`, `From<io::Error> for Error` can't be defined by users outside of `nix`.

If the goal is to have `NixError` encompass all errors, wouldn't it be useful to have `From` defined at least for the main kinds of errors?
`NixPath` is defined for `Path` and `PathBuf`, and both of these have functions that return `io::Error`, so it might not be that rare of a situation.

Susurrus:
> There might be code outside of nix that uses it though. And because neither the From trait nor NixError are owned outside of nix, From<io::Error> for Error can't be defined by users outside of nix.

No there won't, code outside of `nix` should not be generating `NixError` for any functions. 

> If the goal is to have NixError encompass all errors, wouldn't it be useful to have From defined at lest for the main kinds of errors? NixPath is defined for Path and PathBuf, and both of these have functions that return io::Error, so it might not be that rare of a situation, wouldn't it?

The goal isn't for *other* libraries to use `NixError` to encapsulate all errors in their own functions, but for `nix` itself to use. You asked if this is useful, and the answer is currently no, as I have not seen any code in `nix` that would benefit from this. If there is code in `nix` that could benefit from this, please refactor that code to use this and then we'll have somewhere to start. But currently there's no code internal to `nix` that needs this and no code outside of `nix` will ever need this, so there's no reason to merge this.

vincenthage:
Ah indeed, I misunderstood what you meant by `all encompassing error`.
If it's restricted to `nix`, nevermind then.

Susurrus:
Generally libraries generate their own error types and split them up as it makes sense internally. In our case we use `NixError` for everything internally, but it's not for use outside the lib. Other libraries should make their own error types as well, since the types are so dependent on the specifics of each library it'd be hard to make an "all encompassing error" type that was actually easier to use than a specific one.

Susurrus:
I thought your question last night was in the context of an error type for `nix`, and that you had found code that will generate `io::Error` so was wondering if that case should be handled. I didn't realize you were talking in the general case, so sorry about the confusion!

PR #677, opened 2017-07-15
PtyMaster::drop should panic on EBADF

### LINKED ISSUE #659: Potential double close in PtyMaster
`PtyMaster` implements `Drop` to close its `RawFd`.  However, it ignores errors, because the caller may have manually closed the `RawFd`.  This is bad, because it can lead to double closes, like this:
```
{
    let m = posix_openpt(O_RDWR);		# Creates a pty with file descriptor 3
    close(m.master);				# Close file descriptor 3
    let f = std::fs::File::create("foo");	# Creates a file with file descriptor 3
}						# PtyMaster::Drop closes file descriptor 3
f.write("whatever");				# fails with EBADF
```

There are three possible solutions:
1) Always check for errors in `PtyMaster::Drop`
2) Don't implement `PtyMaster::Drop`; make the caller responsible for closing it.  If we go this route, we should also add a `PtyMaster::close` method.
3) Change `PtyMaster` to wrap an `Option<RawFd>` and have the `close` method set it to `None`.  That way, the `drop` method will know whether to call `close`.

I prefer option 1, but I could go with any of these solutions.

asomers:
Fixes #659

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,7 @@ test = true
 name = "test-mount"
 path = "test/test_mount.rs"
 harness = false
+
+[[test]]
+name = "test-ptymaster-drop"
+path = "test/test_ptymaster_drop.rs"
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -45,10 +45,17 @@ impl IntoRawFd for PtyMaster {
 
 impl Drop for PtyMaster {
     fn drop(&mut self) {
-        // Errors when closing are ignored because we don't actually know if the file descriptor
-        // was closed. If we retried, it's possible that descriptor was reallocated in the mean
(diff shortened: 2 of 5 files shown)


Susurrus:
LGTM. This bug was never in a released version of nix, so we're fine with no CHANGELOG entry.

Is it worth adding any doccomments somewhere that elaborate on this case so people understand what could happen? I'm not certain there is, but I wanted to make sure you thought about it.

asomers:
I have no idea why i686-unknown-linux-musl keeps failing.  It passed 10,000 times in a row on my system.  QEMU is not to blame; i686 doesn't use it.  @Susurrus any ideas?

Susurrus:
Please squash these into a single commit. And then it's r+ by me.

PR #680, opened 2017-07-16
Fix some bugs in the termios tests

### LINKED ISSUE #154: Unexpected ENOTTY in sys::test_termios::test_tcgetattr
I occasionally see this failure if I run cargo test over and over:

```
failures:

---- sys::test_termios::test_tcgetattr stdout ----
    thread 'sys::test_termios::test_tcgetattr' panicked at 'assertion failed: `(left == right)` (left: `Some(Sys(ENOTTY))`, right: `Some(Sys(EBADF))`)', test/sys/test_termios.rs:15



failures:
    sys::test_termios::test_tcgetattr

test result: FAILED. 22 passed; 1 failed; 0 ignored; 0 measured

thread '<main>' panicked at 'Some tests failed', /home/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libtest/lib.rs:255
```

asomers:
* Make test_tcgetattr deterministic.  The old version behaved
  differently depending on what file descriptors were opened by the
  harness or by other tests, and could race against other tests.

* Close some file descriptor leaks

Fixes #154

### DIFF
--- a/test/sys/test_termios.rs
+++ b/test/sys/test_termios.rs
@@ -1,6 +1,7 @@
 use std::os::unix::prelude::*;
+use tempfile::tempfile;
 
-use nix::{Error, fcntl, unistd};
+use nix::{Error, fcntl};
 use nix::errno::Errno;
 use nix::pty::openpty;
 use nix::sys::termios::{self, ECHO, OPOST, OCRNL, Termios, tcgetattr};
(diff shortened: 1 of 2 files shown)


test/sys/test_termios.rs
```diff
@@ -14,22 +15,18 @@ fn write_all(f: RawFd, buf: &[u8]) {
...
-        }
-    }
+    let pty = openpty(None, None).unwrap();
+    assert!(termios::tcgetattr(pty.master).is_ok());
+    close(pty.master);
+    close(pty.slave);
+    let file = tempfile().unwrap();
+    assert_eq!(termios::tcgetattr(file.as_raw_fd()).err(),
+               Some(Error::Sys(Errno::ENOTTY)));
+    assert_eq!(termios::tcgetattr(1000000).err(),
```
Susurrus: Can you add a comment for what this is testing?

test/sys/test_termios.rs
```diff
@@ -14,22 +15,18 @@ fn write_all(f: RawFd, buf: &[u8]) {
...
-            // Otherwise it should return any error
-            _ => assert!(termios.is_err())
-        }
-    }
+    let pty = openpty(None, None).unwrap();
+    assert!(termios::tcgetattr(pty.master).is_ok());
+    close(pty.master);
+    close(pty.slave);
+    let file = tempfile().unwrap();
+    assert_eq!(termios::tcgetattr(file.as_raw_fd()).err(),
```
Susurrus: Can you add a comment for what you're testing here too.

test/sys/test_termios.rs
```diff
@@ -14,22 +15,18 @@ fn write_all(f: RawFd, buf: &[u8]) {
     }
 }
 
+// Test test_tcgetattr with multiple file descriptor types
```
Susurrus: Since this describes multiple file descriptor types, can you enumerate each type within the test? I'd actually suggest breakaing this test into 3 different ones for each type instead of one big one.

Susurrus:
Can you update the `openpt` docs to state that the `fd`s returned from it should be manually closed. I think we should be more explicit about this even though the user *should* know this. Primarily this is because some of the `nix` datatypes implement `Drop` and some don't, so it's more important that everything's documented. Once that's down it's r+ susurrus.

asomers:
as he said:

bors r+ susurrus

PR #681, opened 2017-07-17
Remove feature flags

### LINKED ISSUE #98: Add a conditional compilation system to allow features available only on some platforms.
I removed a number of bindings that were not generally portable in order to get everything compiling on Rust 1.0 beta. These should be brought back.

### LINKED ISSUE #206: Documentation lacks modules that depend on compile-time features.
The documentation linked from the readme does not include the modules `sys::signalfd` or `sys::eventfd`, both of which are present in the source (with `#[cfd(feature = "...")]` flags).

### LINKED ISSUE #306: Build fails with execvpe feature enabled
Build errors:
- missing `ffi` module in `unistd`
- use of undeclared type `Error::Sys` at `src/unistd.rs:396`

### LINKED ISSUE #308: Optional features are not build tested
Currently none of the available optional features are built by Travis, therefore:
- patches for the optional parts are not tested by Travis (or Homu)
- patches for other parts can silently break the optional parts (see #306, #307)

Susurrus:
These are vestiges of the initial push to get this working on Rust 1.0. These feature flags are undocumented and so hard to discover (only learned about them today!), prevent functions being included that should be and this also affects documentation on docs.rs, and none of the features are tested in CI and the `execvpe` has been broken for forever.

The solution is to conditionally compile everything supported for a given platform and do away completely with the feature flags. The `execvpe` function is completely removed as it's not available for *nix platforms in libc and is already broken, so no loss removing it. We'll add it back once it's back in libc (rust-lang/libc#670).

Closes #98.
Closes #206.
Closes #306.
Closes #308.


asomers:
Coincidentally, I first learned of those feature flags today, too.  I agree with getting rid of them.  Could you also change test_signalfd.rs into a normal test file?  And there are also some feature gates in test_unistd.rs and sys/test_uio.rs to remove.

Susurrus:
I'm working on fixing the tests but I get a "main function not found" when I "fix" things. You have this problem before?

Susurrus:
Nvm, figured it out. Weird we have mount and signalfd tests using a custom test harness instead of the main one. This is some old code..

Susurrus:
And this needs a CHANGELOG before it can be merged, but I'm working on getting CI passing first.

Susurrus:
Android failures are because of missing libc items (see rust-lang/libc#671), but I'll probably just disable signalfd on android so this can be part of the current release.

Susurrus:
There are `ECHILD` errors from running the tests, which I can also produce locally, so I'll need to figure those out as well.

Susurrus:
@asomers Want to take a look-see over this for me, as I think the new `signalfd` test should be good but I'm not too comfortable with `fork` and would appreciate a second pair of eyes on it? Should be ready to go if it looks good to you.

test/sys/test_signalfd.rs
```diff
@@ -0,0 +1,52 @@
+#[cfg(target_os = "linux")]
```
asomers: Isn't this line redundant with test/sys/mod.rs:5 ?

test/sys/test_signalfd.rs
```diff
@@ -0,0 +1,52 @@
...
+    use nix::sys::signal::*;
+    use nix::sys::wait::*;
+
+    // Set up a signalfd to listen for SIGUSR1 on this thread
+    let mut mask = SigSet::empty();
+    mask.add(SIGUSR1);
+    mask.thread_block().unwrap();
+
+    // Start a child process
+    match unistd::fork() {
```
asomers: I don't think you should fork here.  For one thing, cargo's test harness has deadlocks when a multithreaded test program forks.  That's why I had to add std::process:exit(0) all over the place.  For another thing, it's not necessary.  You can just create the eventfd and send the signal all from the same thread, just like the old test did.

test/sys/test_signalfd.rs
```diff
@@ -0,0 +1,52 @@
...
+    use nix::sys::signal;
+    use nix::unistd;
+    use nix::unistd::*;
+    use nix::unistd::ForkResult::*;
+    use nix::sys::signal::*;
+    use nix::sys::wait::*;
+
+    // Set up a signalfd to listen for SIGUSR1 on this thread
+    let mut mask = SigSet::empty();
+    mask.add(SIGUSR1);
```
asomers: `SIGUSR1` is a global resource, so this test will conflict with any other test that does anything with it.  I dealt with that in PR #638 by creating a mutex to protect `SIGUSR2`.  You should probably do the same for `SIGUSR1`

test/sys/test_signalfd.rs
Susurrus: I remember why I was doing this, because if I didn't I saw errors like "process didn't exit successfully (signal: 10)" even just running my test by itself. This is `ECHILD`, which states there are no child processes. Would you have another suggestion rather than `fork`ing?

asomers: `ECHILD` is an errno, not a signal.  Signal 10, on Linux, is `SIGUSR1`.  It sounds like you weren't correctly blocking the signal.  I can take a look at it, if you want.

test/sys/test_signalfd.rs
Susurrus: I was going to use `SIGUSR2`, but I actually think this implementation isn't quite right. According to [this](https://stackoverflow.com/questions/11679568/signal-handling-with-multiple-threads-in-linux#11679770] signal disposition is shared across all threads in a process. So I think we should actually have a `SIGNAL_MTX` and not one specific to each signal.

Susurrus:
Thanks for the review, @asomers, everything should be addressed now!

Note that I b roadened the `SIGUSR2_MTX` to a general `SIGNAL_MTX` because depending on how signal handling is modified, it might not be updated in one test in a way that doesn't clobber the signal mask in another test. So instead of worrying about that, just make it general to all signal handling. This also makes it more in line with the rest of the mutexes which are pretty broad.

Also, I wanted to note that this includes a commit to fix unused imports introduced by you @asomers in 2cd4420279. You might be running an older version of Rust that doesn't output warnings about unused variables in tests by default, which is why this was easy to miss. Might be worth updating Rust depending on your specific dev needs!

test/sys/test_signalfd.rs
```diff
@@ -0,0 +1,26 @@
...
+    let mut mask = SigSet::empty();
+    mask.add(signal::SIGUSR1);
+    mask.thread_block().unwrap();
+
+    let mut fd = SignalFd::new(&mask).unwrap();
+
+    // Send a SIGUSR1 signal to the current process. Note that this uses `raise` instead of `kill`
+    // because `kill` with `getpid` isn't correct during multi-threaded execution like during a
+    // cargo test session. Instead use `raise` which does the correct thing by default.
+    raise(signal::SIGUSR1).ok().expect("Error: Kill Failed");
```
asomers: s/Kill/raise

Susurrus:
Cool, thanks for reviewing it!

As for supporting preadv/pwritev on more platforms, can you open an issue for it? It'd be great to have more mentored/good-first-bug issues.

bors r+ asomers

asomers:
We hit the `test_double_close` timeout on powerpc again.  But we also had failures in the `epoll` tests on x86_64.  I haven't personally seen that one yet.

Susurrus:
I'm thinking it's time to deprecate powerpc to Tier 2 as you suggested, and I propose we do that for the 0.9 release so we aren't promising too much to our customers.

Susurrus:
I'm seeing those epoll failures on #686 as well....Hmmm.

Susurrus:
I'm re-running master by itself to see what's going on: https://travis-ci.org/nix-rust/nix/builds/254708956

PR #684, opened 2017-07-17
Fix nix for Dragonfly

mneumann:


Cargo.toml
```diff
@@ -27,6 +27,9 @@ bitflags = "0.9"
 cfg-if = "0.1.0"
 void = "1.0.2"
 
+[target.'cfg(target_os = "dragonfly")'.dependencies]
+errno-dragonfly = "0.1"
```
asomers: We shouldn't need an extra crate just for this.  The basic symbols should all be defined in libc, and nix's won src/errno.rs defines `errno_location`.  Can you merge whatever you need into nix and libc so we don't depend on the extra crate?

src/sys/socket/ffi.rs
```diff
@@ -20,7 +20,10 @@ pub type type_of_cmsg_len = socklen_t;
 #[cfg(target_os = "macos")]
 pub type type_of_cmsg_data = c_uint;
 
-#[cfg(not(target_os = "macos"))]
+#[cfg(target_os = "dragonfly")]
```
asomers: I think this should be `#[cfg(any(target_os = "dragonfly", target_os = "macos"))]`

src/sys/wait.rs
```diff
@@ -145,27 +145,34 @@ mod status {
           target_os = "netbsd"))]
 mod status {
     use sys::signal::Signal;
+    use libc::c_int;
```
asomers: Why is it necessary to change all of these types?

Susurrus:
There's big changes coming to the FFI declarations in the `socket` module in #647. I suggest you test that PR first and provide edits there and keep this PR rebased on top of it so it can be merged afterwards.

Cargo.toml
mneumann: No, there is no other possibility!!! I talked with @alexcrichton to make the thread_local attribute attribute stable, and he actually suggested me to write a C extension like errno-dragonfly. ```libc``` does not expose the errno variable directly AFAIK. 

The problem is that in DragonFly, the errno variable is not a global, but a thread locale variable, so a simple "extern __errno" as used by the other OSes simply does not work. In libstd, errno can be accessed by using "#[thread_local]", as unstable features are allowed for libstd.

If you want to support DragonFly there really isn't any other option to go via C atm.

src/sys/socket/ffi.rs
mneumann: No, this is correct! ```type_of_cmsg_data``` is already defined above as c_uint. Here it is defined for DragonFly as c_int. Below it is defined for all other OSes.

src/sys/wait.rs
mneumann: I don't understand why you use ```i32``` here instead of ```c_int```. The latter is the correct type for those APIs, as waitpid returns a c_int, not a i32. On most platforms these types are identical, but c_int should be the correct one.

mneumann:
Ok, I will wait for #647 to land, then rebase on top of that.

Cargo.toml
asomers: That's ... sad.  I took a stab at this myself.  It doesn't look as though we can use `#[thread_local]` unless we enable it for _all_ builds, not just Dragonfly's.  Still, what would you say to moving the C extension into Nix instead of making it its own crate?  Cargo just seems like an awfully heavyweight way to manage such a tiny amount of code.

src/sys/socket/ffi.rs
asomers: Oh, you're right.  I didn't notice the "u".

src/sys/wait.rs
```diff
@@ -145,27 +145,34 @@ mod status {
...
 
-    pub fn signaled(status: i32) -> bool {
+    #[cfg(target_os = "dragonfly")]
+    pub fn signaled(status: c_int) -> bool {
+        wstatus(status) != WSTOPPED && wstatus(status) != 0
+    }
+
+    #[cfg(not(target_os = "dragonfly"))]
+    pub fn signaled(status: c_int) -> bool {
         wstatus(status) != WSTOPPED && wstatus(status) != 0 && status != 0x13
```
asomers: What does the magic number 0x13 mean?  This sounds like a constant that should be in libc.

src/sys/wait.rs
asomers: Sure, if you want to be pedantically correct.  But it's got nothing to do with DragonFly.  It you're going to change those types, please do it in a separate commit, not comingled with DragonFly specific changes.

Susurrus:
@mneumann Do you want to rebase this and address @asomers review now that #647 landed?

Susurrus:
@mneumann I'd like to go ahead and get these changes except the errno-dragonfly parts merged. We can open that as an issue and defer on resolving it for now. Do you have time coming up here to address this?

Cargo.toml
Susurrus: I don't have a problem with this crate staying external, though I'd much rather rely on a crate that had proper semver (i.e. was 1.0+).

src/sys/wait.rs
Susurrus: I agree with @asomers here, please remove this changes from this PR.

Susurrus:
@mneumann I think I'd be fine with errno-dragonfly being a dependency of `nix`, as I'd like to get this merged. Please remove some of the unnecessary `int` -> `i32` changes you have and rebase and let's get this merged!

mneumann:
Closing in favor of https://github.com/nix-rust/nix/pull/799

PR #689, opened 2017-07-19
Fix testing

Susurrus:
Move `powerpc-unknown-linux-gnu` to Tier 2. And attempt a fix for the epoll tests that seem to fail reliably now.

cc @asomers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ matrix:
       rust: 1.13.0
     - env: TARGET=mipsel-unknown-linux-gnu
       rust: 1.13.0
-    - env: TARGET=powerpc-unknown-linux-gnu
+    - env: TARGET=powerpc-unknown-linux-gnu DISABLE_TESTS=1
       rust: 1.13.0
     - env: TARGET=powerpc64-unknown-linux-gnu
       rust: 1.13.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,6 +97,11 @@ matrix:
       rust: beta
 
   allow_failures:
+    # We allow beta to fail here because a bug might crop up in rust that causes it. We will
+    # however be on the lookout for this and file those bugs with upstream.
+    - env: TARGET=x86_64-unknown-linux-gnu
+      rust: beta
+
     # Planning to add these targets, but they can fail for now
     - env: TARGET=mips64-unknown-linux-gnuabi64
       rust: 1.13.0
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ Tier 1:
   * i686-unknown-linux-musl
   * mips-unknown-linux-gnu
   * mipsel-unknown-linux-gnu
-  * powerpc-unknown-linux-gnu
   * powerpc64-unknown-linux-gnu
   * powerpc64le-unknown-linux-gnu
   * x86_64-apple-darwin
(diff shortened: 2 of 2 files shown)


Susurrus:
Note that this looks like a big patch series because it's based on #681 because it needs the signalfd fixes there, but only the newest two commits are part of this series.

test/sys/test_epoll.rs
```diff
@@ -1,24 +1,39 @@
...
 use nix::sys::epoll::{epoll_create1, epoll_ctl};
-use nix::{Error, Errno};
+use nix::sys::signal::SigSet;
+use nix::sys::signalfd::SignalFd;
+use nix::unistd::close;
 
 #[test]
 pub fn test_epoll_errno() {
+    // Create a new signalfd to use for testing. This is better than using stdin or other such fd.
+    let sfd = SignalFd::new(&SigSet::empty()).unwrap();
```
asomers: is a `SignalFd` really the simplest kind of file descriptor that works with `epoll`?  What about plain old files?

.travis.yml
```diff
@@ -97,6 +97,9 @@ matrix:
       rust: beta
 
   allow_failures:
+    # There are too many inconsistent failures on this arch to make it Tier 1.
```
asomers: Frankly, I would rather just add `DISABLE_TESTS=1` to powerpc.  I don't think we're likely to learn much by running them, especially since we can't trust any of their failures.

test/sys/test_epoll.rs
Susurrus: Does it matter? Is `signalfd` overkill or would you just rather reduce cross-dependencies between tests? [This](https://stackoverflow.com/questions/8057892/epoll-on-regular-files#8057973) seems to suggest you can use it on regular files, so let me give that a shot.

.travis.yml
Susurrus: I agree, allowing failures would also make this Tier 2 not Tier 3, so good catch!

Susurrus:
Well this failed with signalfd anyways...I'm wondering if this has anything to do with us removing `sudo` from our CI builds...

test/sys/test_epoll.rs
asomers: Yes, exactly.  I would rather use a plainer file descriptor type, so that this test doesn't depend on `signalfd`.  Sockets would work too.

asomers:
I can reproduce the epoll test failures, but only when using the nightly compiler and only in release mode.  The failure is due to unwrapping an `Sys(UnknownErrno)`.  However, it happens after `libc::epoll_ctl` returns 0.  So `epoll_ctl` ought to be returning `Ok()`.  This looks like a compiler bug to me.

Susurrus:
Man, didn't even realize it was the beta that was failing and stable was passing. This passes on nightly (7-16 and 7-18) for me on x86_64-linux-gnu however. I also installed the same beta Rust version on the machine that failed and it succeeded for me. Hmmm...

I'm thinking with should put the `beta` under `allow_failures`.

Susurrus:
To elaborate on that, we have builds from yesterday that were still using `rustc 1.19.0-beta.4 (deb2ab60b 2017-07-14)`, but the latest ones use `rustc 1.20.0-beta.1 (e93aa3aa8 2017-07-18)` and that fails. Can you look into this a bit since you can reproduce it and file a bug against rust? For now I'm going to change beta to be allowed to fail.

Susurrus:
@asomers What do you think? Should we merge it (BTW we can just merge this without bors I think).

asomers:
Yes, we should.  I also think that you should hop in Gitter so we can talk about this Rust bug.

bors r+

PR #691, opened 2017-07-20
Fix thread safety issues in pty and termios tests

asomers:
ptsname(3) returns a pointer to a global variable, so it isn't
thread-safe.  Protect it with a mutex.

### DIFF
--- a/test/sys/test_termios.rs
+++ b/test/sys/test_termios.rs
@@ -18,10 +18,14 @@ fn write_all(f: RawFd, buf: &[u8]) {
 // Test tcgetattr on a terminal
 #[test]
 fn test_tcgetattr_pty() {
-    let pty = openpty(None, None).unwrap();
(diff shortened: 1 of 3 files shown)


Susurrus:
Have you seen this be an actual issue? My understanding is that this is just like the `errno` strings, where the strings are static in global memory, but they never change, so you should be able to pass around pointers to them willy-nilly.

asomers:
Yep.  If I run the tests with --test-threads=8, I get failures due to `ptsname` within seconds.  This is the very reason that `ptsname_r` exists.  It's like the difference between `strerror` and `strerror_r`

Susurrus:
Hmm, I've never seen those fail locally, but this does seem like the right fix.

LGTM

bors r+

asomers:
Since bors is stuck, I'm going to manually merge

bors r-

PR #692, opened 2017-07-20
Rename the public ptrace_* functions.

marmistrz:
Unlike in C, we have namespacing in Rust. Renaming the functions allows us to avoid a `use nix::sys::ptrace::*` in favor of `use nix::sys::ptrace` and then calling, for example, `ptrace::traceme()`

I'm wondering if we should rename the private functions too...


Susurrus:
If you think the private functions could have better names, I'm all for it. You'll also need to re-name the tests (if you run `cargo test` before pushing you'll generally save yourself some time with this PRs).

Also, could you add this usage to the module documentation at the top? It'd be nice if we explicitly told people that's how they should do this versus just letting them figure it out. In theory they should know cause it's how a lot of Rust works, but it might be worth mentioning. What do you think? Too basic?

marmistrz:
I think I'll add the documentation in #666

test/sys/test_ptrace.rs
```diff
@@ -1,6 +1,7 @@
 use nix::Error;
 use nix::errno::*;
 use nix::unistd::*;
+use nix::sys::ptrace;
 use nix::sys::ptrace::*;
```
Susurrus: This change doesn't seem so useful if you still need to glob import a ton of things. Can we get rid of those here or does that make things really cluttered?

marmistrz:
I can't find a better name for the private functions and I'd leave them as they are. A function called `other` looks weird.

test/sys/test_ptrace.rs
```diff
@@ -2,22 +2,22 @@ use nix::Error;
...
     // FIXME: qemu-user doesn't implement ptrace on all arches, so permit ENOSYS
-    let err = ptrace(PTRACE_ATTACH, getpid(), ptr::null_mut(), ptr::null_mut()).unwrap_err();
+    let err = ptrace::ptrace(PTRACE_ATTACH, getpid(), ptr::null_mut(), ptr::null_mut()).unwrap_err();
     assert!(err == Error::Sys(Errno::EPERM) || err == Error::Sys(Errno::ENOSYS));
 }
 
 // Just make sure ptrace_setoptions can be called at all, for now.
 #[test]
 fn test_ptrace_setoptions() {
+    use nix::sys::ptrace::ptrace::*; // for PTRACE_O_TRACESYSGOOD
```
Susurrus: Why aren't you just importing the single constant here then? No need to bulk import. Same for `PTRACE_ATTACH` above.

Man, we really need to get `PtraceRequest`, `PtraceEvent`, and `PtraceOptions` into their own enums/bitflags. Having a `ptrace` module within a `ptrace` module is just horrible ergonomics.

marmistrz: Oh, I probably was too tired back then ;) I'll fix that tomorrow.

marmistrz: And I totally agree about the enums

marmistrz:
Anything else needed to merge this?

Susurrus:
Yes, this needs a CHANGELOG entry (you should wait for #706 to land and rebase on that).

marmistrz:
Is it expected that rebasing on current master generated a big list of commits mentioned in the PR? (see the whole list after your comment)

test/sys/test_wait.rs
```diff
@@ -51,7 +51,8 @@ mod ptrace {
     use libc::_exit;
 
     fn ptrace_child() -> ! {
-        let _ = ptrace(PTRACE_TRACEME, Pid::from_raw(0), ptr::null_mut(), ptr::null_mut());
+        // TODO ptrace::traceme will be added in #666
```
Susurrus: There's no need for these comments.

test/sys/test_wait.rs
```diff
@@ -62,16 +63,18 @@ mod ptrace {
...
+        assert!(ptrace::setoptions(child, PTRACE_O_TRACESYSGOOD | PTRACE_O_TRACEEXIT).is_ok());
 
         // First, stop on the next system call, which will be exit()
-        assert!(ptrace(PTRACE_SYSCALL, child, ptr::null_mut(), ptr::null_mut()).is_ok());
+        // TODO ptrace::syscall will be added in #666
+        assert!(ptrace::ptrace(PTRACE_SYSCALL, child, ptr::null_mut(), ptr::null_mut()).is_ok());
         assert_eq!(waitpid(child, None), Ok(WaitStatus::PtraceSyscall(child)));
         // Then get the ptrace event for the process exiting
-        assert!(ptrace(PTRACE_CONT, child, ptr::null_mut(), ptr::null_mut()).is_ok());
+        // TODO ptrace::cont will be added in #666
```
Susurrus: Same here.

test/sys/test_wait.rs
```diff
@@ -62,16 +63,18 @@ mod ptrace {
         // Wait for the raised SIGTRAP
         assert_eq!(waitpid(child, None), Ok(WaitStatus::Stopped(child, SIGTRAP)));
         // We want to test a syscall stop and a PTRACE_EVENT stop
-        assert!(ptrace_setoptions(child, PTRACE_O_TRACESYSGOOD | PTRACE_O_TRACEEXIT).is_ok());
+        assert!(ptrace::setoptions(child, PTRACE_O_TRACESYSGOOD | PTRACE_O_TRACEEXIT).is_ok());
 
         // First, stop on the next system call, which will be exit()
-        assert!(ptrace(PTRACE_SYSCALL, child, ptr::null_mut(), ptr::null_mut()).is_ok());
+        // TODO ptrace::syscall will be added in #666
```
Susurrus: And here.

src/sys/ptrace.rs
```diff
@@ -108,7 +108,7 @@ fn ptrace_other(request: ptrace::PtraceRequest, pid: Pid, addr: *mut c_void, dat
 }
 
 /// Set options, as with `ptrace(PTRACE_SETOPTIONS,...)`.
-pub fn ptrace_setoptions(pid: Pid, options: ptrace::PtraceOptions) -> Result<()> {
+pub fn setoptions(pid: Pid, options: ptrace::PtraceOptions) -> Result<()> {
```
Susurrus: There's no module-level documentation for the ptrace module, but it'd be good to add one showing an example usage that demonstrates the namespacing we recomment. We don't *need* it per-say, but it'd help users use this crate very easily. It'll also help us see where the imports aren't very ergonomic so we can improve them later.

marmistrz: I'll add the documentation when I'm done with all the ptrace changes, ok?

Besides, I really think about splitting up #666 into a couple smaller PRs because I'm starting to lose control of the tests.

marmistrz:
Removed the comments.

PR #693, opened 2017-07-20
fix some tests for Android

ndusart:
I tested the crate on an Android emulator and quite all the tests passed. Only 6 were failing.

> failures:
> 
> ---- sys::test_pthread::test_pthread_self stdout ----
> 	thread 'sys::test_pthread::test_pthread_self' panicked at 'assertion failed: tid > 0', test/sys/test_pthread.rs:6
> note: Run with `RUST_BACKTRACE=1` for a backtrace.
> 
> ---- sys::test_socket::test_getsockname stdout ----
> 	thread 'sys::test_socket::test_getsockname' panicked at 'bind failed: Sys(EACCES)', /checkout/src/libcore/result.rs:859
> 
> ---- sys::test_socket::test_unixdomain stdout ----
> 	thread 'sys::test_socket::test_unixdomain' panicked at 'bind failed: Sys(EACCES)', /checkout/src/libcore/result.rs:859
> 
> ---- sys::test_termios::test_local_flags stdout ----
> 	thread 'sys::test_termios::test_local_flags' panicked at 'called `Option::unwrap()` on a `None` value', /checkout/src/libcore/option.rs:329
> 
> ---- test_net::test_if_nametoindex stdout ----
> 	thread 'test_net::test_if_nametoindex' panicked at 'assertion failed: if_nametoindex(&LOOPBACK[..]).is_ok()', test/test_net.rs:11
> 
> ---- test_unistd::test_mkstemp stdout ----
> 	thread 'test_unistd::test_mkstemp' panicked at 'mkstemp failed: ENOENT: No such file or directory', test/test_unistd.rs:73

This PR fixes 3 of those:
 - `test_pthread_self`: pthread_t is unsigned, so it can be negative (and it is often the case)
 - `test_if_nametoindex`: constant `LOOPBACK` is the same than on Linux
 - `test_mkstemp`: directory `/tmp` does not exist in Android

Two are still failing (`test_unixdomain` and `test_getsockname`) because we need some special permissions on Android for playing with sockets. On a rooted physical device, they passed. So, if tests for Android are integrated in CI, we should try to embed the tests to run in an APK with requested permissions or find a way to give the permission to a native program.

`test_local_flags` is still failing also because `O_LARGEFILE` is hardcoded and is not the right value in Android. Then `fcntl::OFlag::from_bits(flags).unwrap()` fails because this flag is then not recognised.
I made a PR in `libc` so that we can rely on libc definitions for all `O_*` flags. (rust-lang/libc#683)

Do you prefer to wait for this one to be fixed for this PR to merged as well ?

### DIFF
--- a/test/sys/test_pthread.rs
+++ b/test/sys/test_pthread.rs
@@ -3,5 +3,5 @@ use nix::sys::pthread::*;
 #[test]
 fn test_pthread_self() {
     let tid = pthread_self();
-    assert!(tid > 0);
+    assert!(tid != 0);
 }
--- a/test/test_net.rs
+++ b/test/test_net.rs
@@ -1,9 +1,9 @@
 use nix::net::if_::*;
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "android", target_os = "linux"))]
 const LOOPBACK: &'static [u8] = b"lo";
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(not(any(target_os = "android", target_os = "linux")))]
 const LOOPBACK: &'static [u8] = b"lo0";
 
 #[test]
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -4,7 +4,7 @@ use nix::unistd::*;
 use nix::unistd::ForkResult::*;
 use nix::sys::wait::*;
 use nix::sys::stat;
-use std::iter;
+use std::{env, iter};
 use std::ffi::CString;
 use std::fs::File;
 use std::io::Write;
@@ -64,12 +64,10 @@ fn test_wait() {
 
 #[test]
 fn test_mkstemp() {
-    #[cfg(target_os = "android")]
-    let tmp = "/data/local/tmp/";
-    #[cfg(not(target_os = "android"))]
-    let tmp = "/tmp/";
+    let mut path = env::temp_dir();
+    path.push("nix_tempfile.XXXXXX");
 
-    let result = mkstemp((tmp.to_owned() + "nix_tempfile.XXXXXX").as_str());
+    let result = mkstemp(&path);
     match result {
         Ok((fd, path)) => {
             close(fd).unwrap();
@@ -77,14 +75,12 @@ fn test_mkstemp() {
         },
         Err(e) => panic!("mkstemp failed: {}", e)
     }
+}
 
-    let result = mkstemp(tmp);
-    match result {
-        Ok(_) => {
-            panic!("mkstemp succeeded even though it should fail (provided a directory)");
-        },
-        Err(_) => {}
-    }
+#[test]
+fn test_mkstemp_directory() {
+    // mkstemp should fail if a directory is given
+    assert!(mkstemp(&env::temp_dir()).is_err());
 }
 
 #[test]
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -64,7 +64,12 @@ fn test_wait() {
 
 #[test]
 fn test_mkstemp() {
-    let result = mkstemp("/tmp/nix_tempfile.XXXXXX");
+    #[cfg(target_os = "android")]
+    let tmp = "/data/local/tmp/";
+    #[cfg(not(target_os = "android"))]
+    let tmp = "/tmp/";
+
+    let result = mkstemp((tmp.to_owned() + "nix_tempfile.XXXXXX").as_str());
     match result {
         Ok((fd, path)) => {
             close(fd).unwrap();
@@ -73,7 +78,7 @@ fn test_mkstemp() {
         Err(e) => panic!("mkstemp failed: {}", e)
     }
 
-    let result = mkstemp("/tmp/");
+    let result = mkstemp(tmp);
     match result {
         Ok(_) => {
             panic!("mkstemp succeeded even though it should fail (provided a directory)");


test/sys/test_pthread.rs
```diff
@@ -3,5 +3,5 @@ use nix::sys::pthread::*;
 #[test]
 fn test_pthread_self() {
     let tid = pthread_self();
-    assert!(tid > 0);
+    assert!(tid != 0);
```
Susurrus: Actually the man page for `pthread_self` says we should only use `pthread_equals` for comparing pthreads. If anything we shouldn't even check this value as [it's non-portable](https://stackoverflow.com/questions/16599463/what-is-the-value-range-of-thread-and-process-id#16599528).

test/test_unistd.rs
```diff
@@ -64,7 +64,12 @@ fn test_wait() {
 
 #[test]
 fn test_mkstemp() {
-    let result = mkstemp("/tmp/nix_tempfile.XXXXXX");
+    #[cfg(target_os = "android")]
```
Susurrus: Shouldn't this use [`std::env::temp_dir`](https://doc.rust-lang.org/std/env/fn.temp_dir.html) instead?

Susurrus: And actually, this would be cleaner if we separated this into two tests, since they are two separate tests just rolled into one. Could you do that here as well?

test/test_net.rs
```diff
@@ -1,9 +1,9 @@
 use nix::net::if_::*;
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: The conditions should be alphabetized, so "android" should come first here. Same for the following `#cfg` attribute.

Susurrus:
This is a great start, thanks @ndusart!

>Do you prefer to wait for this one to be fixed for this PR to merged as well?

While fixing all the tests is ideal, I'd like to limit this PR to those tests which can work with the current release of `libc` and also don't add any new hard-coded values. We'll wait on the rest once 0.9 is released and we're back to using `libc-git`. You can either raise a new issue or a separate PR to track that work; your call.

Regarding some points you brought up:
> Two are still failing (test_unixdomain and test_getsockname) because we need some special permissions on Android for playing with sockets. On a rooted physical device, they passed. So, if tests for Android are integrated in CI, we should try to embed the tests to run in an APK with requested permissions or find a way to give the permission to a native program.

I get the feeling we'll run into tons of permission issues with Android as we continue to develop nix. So I'm fine with needing to figure out a rooted app to run tests on. It'll be up to the users of `nix` to figure out if Android supports that functionality or not for their own use cases.

test/sys/test_pthread.rs
ndusart: I was wondering that when I read that test.

So should we remove it ?

test/test_unistd.rs
ndusart: Didn't know about `std::env::temp_dir`, we definitely should use it instead.

ndusart: The tests are splitted and use `std::env::temp_dir` now :)

test/sys/test_pthread.rs
Susurrus: Yeah, remove it, but leave it as a separate commit with a description of why it was removed.

test/test_unistd.rs
```diff
@@ -64,22 +64,28 @@ fn test_wait() {
...
 
-    let result = mkstemp("/tmp/");
-    match result {
-        Ok(_) => {
-            panic!("mkstemp succeeded even though it should fail (provided a directory)");
-        },
-        Err(_) => {}
-    }
+#[test]
+#[should_panic]
```
Susurrus: Our test shouldn't panic, you should just not use `unwrap()` here. You should use assertions to check the return value.

ndusart: Yes, that's right, I tried to tied to the behavior of the test as it was written, it seemed logical at the moment.

test/test_unistd.rs
```diff
@@ -64,22 +64,27 @@ fn test_wait() {
 
 #[test]
 fn test_mkstemp() {
-    let result = mkstemp("/tmp/nix_tempfile.XXXXXX");
+    use std::env;
```
asomers: stylistically, I prefer to put all `use` statements at the top of the file, unless there's a good reason not to.

test/sys/test_pthread.rs
Susurrus: 0 still isn't valid here, we should really fix this at some point, but this is fine for now.

test/test_unistd.rs
Susurrus: I agree with this, especially since this is repeated in the following test. My rule is that if there are a lot of little things that need to be imported in a function/module or if they're conditionally imported, then they're fine within the function/module. Otherwise they should be left out.

test/sys/test_pthread.rs
ndusart: Do I just remove the test leaving the `test_pthread.rs` file for a future test to be inserted later or should I completely remove the file ?

Concerning the validity of the test. Maybe we have to wait `pthread_create` is defined in nix, so that we could make a test where we spin a thread and test that `pthread_self` returns different `pthread_t` in the different threads.
And we shouldn't even test for an "invalid" return value as `pthread_self` is not meant to fail, man page says it always succeed.

Susurrus: Leaving this test as-is is fine.

Susurrus:
@asomers Wanna take another look at this? Looks like small solid fixes to me!

PR #694, opened 2017-07-20
Remove workaround for `pub extern crate`

### LINKED ISSUE #655: `pub extern crate libc` can probably be used now.
https://github.com/nix-rust/nix/blob/master/src/lib.rs#L26
```
// In rust 1.8+ this should be `pub extern crate libc` but prior
// to https://github.com/rust-lang/rust/issues/26775 being resolved
// it is necessary to get a little creative.
pub mod libc {
    extern crate libc;
    pub use self::libc::*;
}
```

https://github.com/rust-lang/rust/issues/26775 has already been closed.

jonas-schievink:
On semi-recent Rust versions (I think 1.8+) this now works properly.

Closes #655

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,13 +23,7 @@ extern crate nix_test as nixtest;
 
 #[macro_use] mod macros;
 
-// In rust 1.8+ this should be `pub extern crate libc` but prior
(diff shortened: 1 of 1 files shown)


asomers:
This looks like a temporary failure in rustup.  We can try remerging.

bors r+ susurrus

Issue #586, opened 2017-04-17
Rethinkging he interface for `fork`


kamalmarhubi:
I was writing a shell as an example for nix, and I realised that #555 makes a really good point: the code that runs in the child after `fork(2)` must not allocate if there is more than one thread. I don't know if `fork` should necessarily be made unsafe, but certainly we need to document better what is ok to do afterwards.

This is somewhat related to #90 in that there's an execution context where only limited operations should be performed. Ideally nix would force only correct usage, but at the same time there's a lot of value of staying close to the libc definitions of functions. (#190 seems connected here too.)

jonas-schievink:
Also see http://www.linuxprogrammingblog.com/threads-and-fork-think-twice-before-using-them

And the [man page](http://man7.org/linux/man-pages/man2/fork.2.html) says:
```
After a fork() in a multithreaded program, the child can safely
call only async-signal-safe functions (see signal-safety(7)) until
such time as it calls execve(2).
```

This makes it sound like `fork` *really really* should be made `unsafe`: Spawning threads and calling non-signal-safe function is possible in safe Rust, therefore there must not be a safe way to do so after a `fork`. Instead, `nix` could provide safe utility functions that do common patterns like `fork + execve` in a safe way.

jonas-schievink:
This has come up in the standard library, too: https://github.com/rust-lang/rust/issues/39575

asomers:
You're right that `fork` is unsafe.  But is it `unsafe` in the Rust sense?  I think we should follow whatever decision rust-lang makes in the issue you linked.

Susurrus:
@asomers I read through some of the discussion there. The argument is that UB should be assumed to include memory unsafety by default. Specifically for `fork()` there are some more reasonings that should make this more likely.

asomers:
Is there any undefined behavior associated with calling async-signal-unsafe functions after forking?  I can't find any.  [fork(2)](http://pubs.opengroup.org/onlinepubs/9699919799/) doesn't mention any undefined behavior.  Instead, it suggests not calling async-signal-unsafe functions in order to prevent deadlocks and the like.

Susurrus:
@jonas-schievink @asomers Did this get cleaned resolved with the merging of #695?

jonas-schievink:
I believe the conclusion was that fork is safe but (eg.) allocating afterwards can cause dadlocks, which is still a footgun.

[...]

PR #695, opened 2017-07-20
Document safety of `fork()` and fix tests

jonas-schievink:
Some tests were invoking non-async-signal-safe functions from the child
process after a `fork`. Since they might be invoked in parallel, this
could lead to problems.

cc #586

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -191,6 +191,18 @@ impl ForkResult {
 /// Continuing execution in parent process, new child has pid: 1234
 /// I'm a new child process
 /// ```
+///
+/// # Safety
+///
+/// In a multithreaded program, only [async-signal-safe] functions like `pause`
+/// and `_exit` may be called by the child (the parent isn't restricted). Note
+/// that memory allocation may **not** be async-signal-safe and thus must be
+/// prevented.
+///
+/// Those functions are only a small subset of your operating system's API, so
+/// special care must be taken to only invoke code you can control and audit.
+///
+/// [async-signal-safe]: http://man7.org/linux/man-pages/man7/signal-safety.7.html
 #[inline]
 pub fn fork() -> Result<ForkResult> {
     use self::ForkResult::*;
@@ -687,7 +699,7 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
 /// use nix::unistd::close;
 ///
 /// fn main() {
-///     let mut f = tempfile::tempfile().unwrap();
+///     let f = tempfile::tempfile().unwrap();
 ///     close(f.as_raw_fd()).unwrap();   // Bad!  f will also close on drop!
 /// }
 /// ```
@@ -700,7 +712,7 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
 /// use nix::unistd::close;
 ///
 /// fn main() {
-///     let mut f = tempfile::tempfile().unwrap();
+///     let f = tempfile::tempfile().unwrap();
 ///     close(f.into_raw_fd()).unwrap(); // Good.  into_raw_fd consumes f
 /// }
 /// ```
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -11,16 +11,17 @@ use std::io::Write;
 use std::os::unix::prelude::*;
 use tempfile::tempfile;
 use tempdir::TempDir;
-use libc::off_t;
-use std::process::exit;
+use libc::{_exit, off_t};
 
 #[test]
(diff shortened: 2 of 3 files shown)


test/test_unistd.rs
```diff
@@ -11,16 +11,17 @@ use std::io::Write;
...
 
 #[test]
 fn test_fork_and_waitpid() {
     #[allow(unused_variables)]
     let m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
-    let pid = fork();
+
+    // Safe: Child only calls `_exit`, which is signal-safe
+    let pid = unsafe { fork() };
     match pid {
```
Susurrus: Let's get rid of the `pid` variable completely here for consistency with other tests and since it isn't used anywhere else.

src/unistd.rs
```diff
@@ -191,10 +191,21 @@ impl ForkResult {
 /// Continuing execution in parent process, new child has pid: 1234
 /// I'm a new child process
 /// ```
+///
+/// # Safety
+///
+/// In a multithreaded program, only [async-signal-safe] functions may be called
```
Susurrus: Your comments talk about `libc::exit` and `libc::pause` being safe to use here. Let's explicitly list those here. We should also state that `std::process:exit` is not safe.

test/test_unistd.rs
```diff
@@ -111,7 +114,10 @@ macro_rules! execve_test_factory(
         // data from `reader`.
         let (reader, writer) = pipe().unwrap();
 
-        match fork().unwrap() {
+        // Safe: Child calls `exit`, `dup` and the provided `exec*` family function.
```
asomers: If you're going to list them all, you should add `close`

src/unistd.rs
jonas-schievink: `libc::exit` actually isn't safe, only `_exit` is

Susurrus: Sorry, that's what I meant. We should document a few of the safe and unsafe things for reference. The big thing about this is that how would the user figure out what is safe or not. I don't know if we have any guidance on that end, however.

jonas-schievink:
> Nobody has yet come up with a POC that violates Rust's memory safety rules by forking. Until that happens, I don't think libstd is going to change.

And this is a very bad thing IMO. Maybe it's hard to come up with a POC on the currently supported configurations, but what about other allocator implementations, other libcs, other operating systems, other architectures where this is a more serious issue? Even if those aren't supported by Rust today, even if those configurations don't even *exist* today - the fact that POSIX says "don't do this it's unsafe" should be more than enough of a reason to proceed with extreme caution.

A language that aims to **guarantee** memory safety shouldn't do things like this.

asomers:
POSIX says it's "unsafe", but the libstd guys say it's not `unsafe`.  `unsafe` means something very particular in Rust.  You can easily do things that are "unsafe" without using any `unsafe` code, like leak memory, trigger deadlocks, etc.  In order to convince libstd to change, we'll probably have to come up with a POC that violates one of Rust's narrow memory safety rules, for example by overrunning an array or using after free.

jonas-schievink:
Yeah, seems like POSIX leaves the meaning of "safe" completely undefined. Depending on that, `fork` might indeed by safe according to Rust's definition of safety. However, common sense makes me doubt that POSIX, which defines a C API, would call something "unsafe" that Rust would deem safe.

jonas-schievink:
As an aside, from the [docs of `pthread_atfork`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_atfork.html):

> If a fork() call in a multi-threaded process leads to a child fork handler calling any function that is not async-signal-safe, the behavior is undefined.

I would be *very* surprised if "the behaviour is undefined" means something different than UB in C in this context, and the child handler is basically just code that is executed by the child after a fork, so there isn't much of a difference to what we're dealing with. (UB is strictly worse than memory unsafety, after all)

asomers:
I don't want to argue with you about the safety of `fork`, because I don't completely disagree with you.  But nix needs to maintain consistency with libc, and since libc is far more important than nix, we need to be having the conversation over there.  For now, please revert the changes to `fork` and let's just merge the changes to the tests and to `fork`'s documentation.

Susurrus:
I'd like to echo what @asomers says here. Rust has a pretty narrow definition of "unsafe" and whatever `libc` goes with we'll stick with for convention. That said, it is worth bringing up the invariants for these functions so that users can be aware.

asomers:
PR #566 just added a new `fork` call in a test.  So when you rebase, please replace its `sys::process::exit` with `_exit`

asomers:
Thanks for the PR @jonas-schievink .  Looks like we can sneak this one into 0.9.0

bors r+

PR #696, opened 2017-07-20
Stop reexporting `Errno` variants

### LINKED ISSUE #664: Errnos are available through multiple paths in nix
The current problem is the errnos (like `EAGAIN`) are exposed at `nix::Errno::EAGAIN` and `nix::errno::EAGAIN`. We should choose one "right" path and only expose these constants there. I'd probably keep `Errno` at `nix::errno::Errno` instead of re-exporting it in the root.

jonas-schievink:
Closes #664 (unsure if this is everything needed)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,7 +25,7 @@ extern crate nix_test as nixtest;
 
 pub extern crate libc;
 
-pub use errno::Errno;
+use errno::Errno;
 
 pub mod errno;
 pub mod features;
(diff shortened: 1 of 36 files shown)


src/lib.rs
```diff
@@ -104,24 +104,24 @@ pub enum Error {
...
     /// Get the current errno and convert it to a nix Error
     pub fn last() -> Error {
-        Error::Sys(errno::Errno::last())
+        Error::Sys(Errno::last())
     }
 
     /// Create a new invalid argument error (`EINVAL`)
     pub fn invalid_argument() -> Error {
-        Error::Sys(errno::EINVAL)
+        Error::Sys(Errno::EINVAL)
```
Susurrus: I'm not certain that I like `Errno::EINVAL` and `Errno::last()`. I don't think it makes sense for `last()` to be in the same place in the heirarchy as `EINVAL`.

jonas-schievink: Are you sure about that? This is definitely the most "rustic" way to do it.

Susurrus:
This still actually has Errno visible from the root when I think it should stay within `nix::errno`. Can you fix that?

Susurrus:
LGTM.

@asomers You want to give this a lookover at all?

Susurrus:
@jonas-schievink if you could squash this into a single commit, I'll go ahead and merge this. Thanks!

Susurrus:
Perfect, thanks @jonas-schievink!

bors r+

PR #697, opened 2017-07-20
Remove `homepage` from Cargo.toml

jonas-schievink:
It's just a (redundant) link to the repo, not a real homepage. According
to the [API
guidelines](https://github.com/brson/rust-api-guidelines#cargotoml-includes-all-common-metadata-c-metadata),
this shouldn't be set in this case.


asomers:
Good catch.

bors r+

asomers:
Shoot.  Bors is really stuck on this one.  I'm going to go ahead and merge it the old-fashioned way.

Issue #373, opened 2016-05-20
Stamp out transmute except in certain cases


kamalmarhubi:
nix is currently a bit too liberal with `mem::transmute`, using it in situations where some casts would suffice. I'd like to reduce these to a minimum because [`transmute` is a scarily big hammer](https://doc.rust-lang.org/nightly/nomicon/transmutes.html).

fiveop:
All occurences. We can discuss each case and tick them off.

```
git grep -n transmute | while read line; do echo -n $line |  sed -r 's/([a-z\/\.]+\:[0-9]+:).*/\1/g' | sed -r 's/([^\:]+)\:([^\:]+)\:/\- \[ \] https\:\/\/github\.com\/nix-rust\/nix\/blob\/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8\/\1#L\2\n/g'; done
```

- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sched.rs#L195
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sched.rs#L210
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/quota.rs#L116
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/quota.rs#L124
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/signal.rs#L155
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/signal.rs#L301
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/signal.rs#L302
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/signal.rs#L303
- [ ] https://github.com/nix-rust/nix/blob/9c6c71a9f5fa4a64d8e5c2996988a93ff5f724c8/src/sys/signal.rs#L304
[...]

PR #700, opened 2017-07-20
Get rid of a lot of transmutes

jonas-schievink:
Most could be replaced by simple raw pointer casts (or even perfectly
safe coercions!).

cc #373

### DIFF
--- a/src/sys/quota.rs
+++ b/src/sys/quota.rs
@@ -111,16 +111,10 @@ pub fn quotactl_sync<P: ?Sized + NixPath>(which: quota::QuotaType, special: Opti
 }
 
 pub fn quotactl_get<P: ?Sized + NixPath>(which: quota::QuotaType, special: &P, id: c_int, dqblk: &mut quota::Dqblk) -> Result<()> {
-    use std::mem;
-    unsafe {
-        quotactl(quota::QuotaCmd(quota::Q_GETQUOTA, which), Some(special), id, mem::transmute(dqblk))
-    }
+    quotactl(quota::QuotaCmd(quota::Q_GETQUOTA, which), Some(special), id, dqblk as *mut _ as *mut c_char)
 }
 
 pub fn quotactl_set<P: ?Sized + NixPath>(which: quota::QuotaType, special: &P, id: c_int, dqblk: &quota::Dqblk) -> Result<()> {
-    use std::mem;
     let mut dqblk_copy = *dqblk;
-    unsafe {
-        quotactl(quota::QuotaCmd(quota::Q_SETQUOTA, which), Some(special), id, mem::transmute(&mut dqblk_copy))
-    }
+    quotactl(quota::QuotaCmd(quota::Q_SETQUOTA, which), Some(special), id, &mut dqblk_copy as *mut _ as *mut c_char)
 }
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -227,7 +227,7 @@ impl<'a> Iterator for CmsgIterator<'a> {
         if self.buf.len() < sizeof_cmsghdr {
             return None;
         }
-        let cmsg: &cmsghdr = unsafe { mem::transmute(self.buf.as_ptr()) };
+        let cmsg: &'a cmsghdr = unsafe { &*(self.buf.as_ptr() as *const cmsghdr) };
 
         // This check is only in the glibc implementation of CMSG_NXTHDR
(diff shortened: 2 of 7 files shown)


src/sys/quota.rs
```diff
@@ -111,16 +111,10 @@ pub fn quotactl_sync<P: ?Sized + NixPath>(which: quota::QuotaType, special: Opti
 }
 
 pub fn quotactl_get<P: ?Sized + NixPath>(which: quota::QuotaType, special: &P, id: c_int, dqblk: &mut quota::Dqblk) -> Result<()> {
-    use std::mem;
-    unsafe {
-        quotactl(quota::QuotaCmd(quota::Q_GETQUOTA, which), Some(special), id, mem::transmute(dqblk))
-    }
+    quotactl(quota::QuotaCmd(quota::Q_GETQUOTA, which), Some(special), id, dqblk as *mut _ as *mut c_char)
```
asomers: Why do you need to cast twice?

src/sys/socket/mod.rs
```diff
@@ -139,7 +139,7 @@ impl<'a> Iterator for CmsgIterator<'a> {
         if buf.len() < sizeof_cmsghdr {
             return None;
         }
-        let cmsg: &cmsghdr = unsafe { mem::transmute(buf.as_ptr()) };
+        let cmsg: &'a cmsghdr = unsafe { &*(buf.as_ptr() as *const cmsghdr) };
```
asomers: why the `&*` ?

src/sys/socket/sockopt.rs
```diff
@@ -210,11 +210,11 @@ impl<T> Get<T> for GetStruct<T> {
     }
 
     unsafe fn ffi_ptr(&mut self) -> *mut c_void {
-        mem::transmute(&mut self.val)
+        &mut self.val as *mut T as *mut c_void
     }
 
     unsafe fn ffi_len(&mut self) -> *mut socklen_t {
-        mem::transmute(&mut self.len)
+        &mut self.len
```
asomers: teehee, this one was silly

src/sys/quota.rs
jonas-schievink: Because you can only cast a reference to a raw pointer to the exact same type

src/sys/socket/mod.rs
jonas-schievink: To dereference the raw pointer, then get a reference to the value behind (the dereference doesn't copy the value out)

src/sys/socket/sockopt.rs
```diff
@@ -368,7 +368,7 @@ impl<'a> Set<'a, usize> for SetUsize {
     }
 
     unsafe fn ffi_ptr(&self) -> *const c_void {
```
Susurrus: This shouldn't actually be `unsafe`, since they are creating raw pointers from safe references. Creating raw pointers is safe, dereferencing raw pointers is unsafe. I suggest we also change all of these functions that do that to not be unsafe.

jonas-schievink: It would be more accurate in terms of safety to make the `Get` and `Set` traits `unsafe` instead of those functions.

Susurrus: That would be required, yes. Could you change those in a separate commit after the current changes? I'm pretty certain the other function in `Get` are also safe, but I don't feel like getting familiar with it enough here to suggest we change those as well.

Susurrus: Oh, sorry, I misread. You suggest the traits themselves be unsafe? The docs say:

> If unsafe code cannot reasonably expect to defend against a bad implementation of the trait, then marking the trait unsafe is a reasonable choice.

Given this and other topics around unsafe, it does appear that the traits should be unsafe and all the functions should be safe.

jonas-schievink: `unwrap` and `blank` should probably stay unsafe: The caller must uphold safety for those, not the callee or implementor.

Susurrus:
This looks ready-to-go to me. There are no public API changes, so no CHANGELOG entry is needed. 

@asomers Any last comments? Didn't know if you got all your questions resolved yet?

Susurrus:
@jonas-schievink Can you rebase this? Then I think it's ready for merge.

Susurrus:
@asomers I think this is ready for merge, are you alright with it?

asomers:
Yep.

bors r+

Issue #679, opened 2017-07-16
Intermittent test failures in CI


asomers:
The following tests have been observed to fail occasionally in CI.  If you find a fix for one of them, check it off:

- [x] `sys::test_aio::test_aio_cancel_all` times out on powerpc-unknown-linux-gnu (https://travis-ci.org/nix-rust/nix/jobs/252267360 https://travis-ci.org/nix-rust/nix/jobs/251497766 https://travis-ci.org/nix-rust/nix/jobs/253133267)
- [x] `errno::test::test_errno_values` times out on powerpc-unknown-linux-gnu (https://travis-ci.org/nix-rust/nix/jobs/252157518 https://travis-ci.org/nix-rust/nix/jobs/249711100)
- [ ] `test_unistd::test_wait` times out on x86_64-unknown-linux-gnu (https://travis-ci.org/nix-rust/nix/jobs/251440414 PR #638 might fix this one)
- [x] `sys::test_aio::test_aio_cancel_all` panics in thread_info.rs on powerpc-unknown-linux-gnu (https://travis-ci.org/nix-rust/nix/jobs/249502626)
- [x] `test_double_close` times out on powerpc-unknown-linux-gnu (https://travis-ci.org/nix-rust/nix/jobs/254249145 https://travis-ci.org/nix-rust/nix/jobs/255050223)
- [x] `sys::test_epoll::test_epoll_ctl` panics when unwrapping `Error::Sys(UnknownErrno)` on `x86_64-unknown-linux-gnu` (https://travis-ci.org/nix-rust/nix/jobs/254999411)
- [x] `sys::test_epoll::test_epoll_errno` panics when unwrapping `Error::Sys(UnknownErrno)` on `x86_64-unknown-linux-gnu` (https://travis-ci.org/nix-rust/nix/jobs/254999411)
- [ ] `error: component download failed for rust-std-armv7-unknown-linux-gnueabihf` on `armv7-unknown-linux-gnueabihf` (https://travis-ci.org/nix-rust/nix/jobs/254708971)

Susurrus:
I'm going to work on the `epoll` ones. I think I know how to solve them, you can create a `signalfd` to then call epoll on and we should be able to consistent results.

Susurrus:
See #689 for a potential fix to the `epoll` errors.

asomers:
[...]

PR #701, opened 2017-07-20
Calculate `nfds` parameter for `select`

jonas-schievink:
Doing this behind the scenes makes the API less error-prone and easier
to use. It should also fix my issue in https://github.com/nix-rust/nix/issues/679#issuecomment-316838148


src/sys/select.rs
```diff
@@ -53,6 +53,20 @@ impl FdSet {
             *bits = 0
         }
     }
+
+    /// Finds the largest file descriptor in the set.
+    ///
+    /// Returns `None` if the set is empty.
+    pub fn largest(&self) -> Option<RawFd> {
```
jonas-schievink: I had to make this function public in order to add a test. For some reason putting a `#[test]` function in this module didn't run the test.

test/sys/test_select.rs
```diff
@@ -31,6 +31,20 @@ fn test_fdset() {
     }
 }
 
+#[test]
```
Susurrus: This isn't needed here.

jonas-schievink: It is, otherwise rustc never calls this method (I just checked by putting a `panic!` inside). Maybe you're thinking of some other test file without harness?

src/sys/select.rs
jonas-schievink: Ah, nevermind, it does work. I probably typo'd `#[cfg(test)]` before. But since the other tests for this type are inside `test_select.rs`, I'm going to leave it this way unless you object.

jonas-schievink:
Note that this is technically not zero-cost, but it is pretty close since the array is small. `select`s performance isn't that great even if it were, though :)

src/sys/select.rs
```diff
@@ -53,6 +53,20 @@ impl FdSet {
...
+        for (i, &block) in self.bits.iter().enumerate().rev() {
+            if block != 0 {
+                // highest bit is located at `n.trailing_zeros()`
+                return Some((i * BITS + block.trailing_zeros() as usize) as RawFd);
```
jonas-schievink: Ah, there's a bug in here.

test/sys/test_select.rs
Susurrus: Sorry, you're right!

src/sys/select.rs
Susurrus: I'd rather have `largest()` be private. And in general unit tests should be in the same file as the functions themselves, it's only the larger integration tests that should be in `/tests`. Let's more the tests into this file at the bottom in a `test` module.

src/sys/select.rs
```diff
@@ -68,11 +87,18 @@ mod ffi {
     }
 }
 
-pub fn select(nfds: c_int,
-              readfds: Option<&mut FdSet>,
+pub fn select(readfds: Option<&mut FdSet>,
```
Susurrus: If you're feeling particularly ambitious, this would be a good time to change the definiton of select to something like `pub fn select<T: Into<Option<&mut FdSet>>>(readfds: T ...` such that all the `Option` types have those kinds of signatures. Then you can use `.into()` within the function and the users of the functions don't need to wrap arguments in `Some()`. It's really ergonomic!

src/sys/select.rs
```diff
@@ -68,11 +87,18 @@ mod ffi {
...
               writefds: Option<&mut FdSet>,
               errorfds: Option<&mut FdSet>,
               timeout: Option<&mut TimeVal>) -> Result<c_int> {
+    // calculate nfds parameter, the largest fd we want to select on
+    let nfds = readfds.iter()
+        .chain(writefds.iter())
```
Susurrus: Can you indent these to align the "."s vertically? Much more readable then.

jonas-schievink:
@asomers It's not O(n), though. The loop runs a maximum of `FD_SETSIZE / BITS` iterations, which is a constant, so it's still O(1). The array only contains 16 or 32 integers for the defined platforms, which should be completely unnoticeable (although I haven't tested). Performance of the calculation even improves with more fds in the set (or higher-numbered ones).

So, what should I do about `largest`? I can make it `pub(crate)` at best, not completely private.

src/sys/select.rs
jonas-schievink: Will do, but please consider moving to the proposed RFC style. Most editors make it surprisingly hard to indent code like this.

jonas-schievink:
Okay, I just tried to re-enable the test on mips and powerpc, but it looks like it still fails.

src/sys/select.rs
```diff
@@ -85,3 +122,84 @@ pub fn select(nfds: c_int,
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use sys::time::{TimeVal, TimeValLike};
+    use unistd::{write, pipe};
+
+    #[test]
+    fn test_fdset() {
```
Susurrus: This should probably be broken into a few smaller tests that make it more clear as to what's being tested. Like break up the adding an `Fd` into a test, then test adding & removing, then test clear in a separate test, etc. It makes debugging failures much easier if tests are grouped into a small amount of related tests. This has all the tests in one function.

asomers:
I still disagree with automatically calculating `nfds`.  Rust is all about zero-cost abstractions, and nix is no different.  But I like the idea of exposing `largest` to the user.

jonas-schievink:
Okay, I'll make that optional

asomers:
Interesting making `nfds` optional.  I guess that gives the best of both worlds.  Everything here looks good to me, though it still needs to be squashed.  Also, please add a doccomment for `select`, and add an entry to CHANGELOG documenting the API change.

src/sys/select.rs
```diff
@@ -53,6 +53,25 @@ impl FdSet {
             *bits = 0
         }
     }
+
+    /// Finds the largest file descriptor in the set.
```
Susurrus: I don't think `largest` is an appropriate term here, as `RawFd`s don't have a size. They have a numbering, however, so `highest` might be appropriate here. Also, since this basically exists exclusively for use with `select()`, please add a simple example showing how you might use it here.

jonas-schievink:
@Susurrus @asomers Anything left to do here?

Susurrus:
You need to rebase this and move your entries into the CHANGELOG into the new Unreleased section since this no longer applies to the 0.9 release.

I'm not a fan of leaving the `nfds` computation up to the user and would prefer to have it handled exclusively by `nix` given that it is O(1) time. We have non-zero-cost abstractions around other APIs (like termios) that I don't consider a problem because the ergonomics/API-safety win outweighs the performance. Users always have an easy escape-hatch from `nix`, and that's to use `libc` directly so I'm not too worried about that. But @asomers seems to not like that idea and I'll defer to him on this.

asomers:
LGTM.  Thanks for all the hard work.

bors r+

asomers:
I'm satisfied, but for the CHANGELOG issue.

asomers:
bors, wait for the changelog to be fixed.

bors r-

jonas-schievink:
Ah, good catch about the changelog, didn't notice that I broke it during rebase. Should be fixed now.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   has changed type from `c_int` to `SockProtocol`.
   It accepts a `None` value for default protocol that was specified with zero using `c_int`.
   ([#647](https://github.com/nix-rust/nix/pull/647))
+- Made `select` easier to use, adding the ability to automatically calculate the `nfds` parameter using the new
+  `FdSet::largest` ([#701](https://github.com/nix-rust/nix/pull/701))
```
Susurrus: This should be "highest"

jonas-schievink:
@Susurrus Fixed the changelog and updated the docs with links. I can also add another test if you tell me how it should look, but the current logic is so simple that I think a test is not warranted.

Susurrus:
The point of the tests is as much about regressions and ergonomics testing as it is about logic. While the logic may be simple, there are 3 discrete ways to call `select()` and all 3 should have tests. This is important because it makes sure the API that we advertise to our users (3 different ways) always works since we have tests for all 3 ways. Please add the remaining tests.

src/sys/select.rs
```diff
@@ -85,3 +162,141 @@ pub fn select(nfds: c_int,
...
+        let (r1, w1) = pipe().unwrap();
+        write(w1, b"hi!").unwrap();
+        let (r2, _w2) = pipe().unwrap();
+
+        let mut fd_set = FdSet::new();
+        fd_set.insert(r1);
+        fd_set.insert(r2);
+
+        let mut timeout = TimeVal::seconds(10);
+        assert_eq!(1, select(fd_set.highest(),
```
Susurrus: Isn't this not correct? Shouldn't it be `fd_set.highest() + 1`?

jonas-schievink: Oh, good catch!

Susurrus: It does concern me however that this test didn't fail when ran, however. Shouldn't it have?

jonas-schievink: Probably the fd that fired the select was the one with the lower number.

Susurrus: There any way to guarantee that's not the case so this test will always fail?

jonas-schievink: Well, that's why I opened the PR is the first place - the existing test wasn't threadsafe and could fail because of this. I guess you could put a mutex around every test that allocates fds...

asomers:
@jonas-schievink I think we're all ready to commit this, but you'll need to rebase first to fix a trivial merge conflict in the CHANGELOG.

PR #709, opened 2017-07-26
Mark nix::sys::ptrace::ptrace as unsafe, add safe variants of some routines

marmistrz:
These include:
* PTRACE_TRACEME
* PTRACE_CONT
* PTRACE_ATTACH
* PTRACE_SYSCALL

This is a part of #666, which is being split up into a couple smaller PRs.


src/sys/ptrace.rs
```diff
@@ -140,3 +140,51 @@ pub fn setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+            ptrace::PTRACE_ATTACH,
+            pid,
+            ptr::null_mut(),
+            ptr::null_mut(),
+        ).map(|_| ()) // ignore the useless return value
+    }
+}
+
+/// Makes the `PTRACE_CONT` request to ptrace
+pub fn cont(pid: Pid) -> Result<()> {
```
Susurrus: Call this "continue" instead.

And provide better docs here.

src/sys/ptrace.rs
```diff
@@ -70,7 +70,7 @@ mod ffi {
 
 /// Performs a ptrace request. If the request in question is provided by a specialised function
 /// this function will return an unsupported operation error.
-pub fn ptrace(request: ptrace::PtraceRequest, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
+pub unsafe fn ptrace(request: ptrace::PtraceRequest, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
```
Susurrus: This needs to be expanded to report `UnsupportedOperation` for all 4 added specialized functions.

Also, what are the specific usecases of `ptrace` that will generate unsafety as Rust defines it (crashing isn't necessarily unsafe)? Since you pulled this out of another PR, I don't know where that discussion went.

src/sys/ptrace.rs
```diff
@@ -140,3 +140,51 @@ pub fn setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+        ptrace(
+            ptrace::PTRACE_TRACEME,
+            Pid::from_raw(0),
+            ptr::null_mut(),
+            ptr::null_mut(),
+        ).map(|_| ()) // ignore the useless return value
+    }
+}
+
+/// Makes the `PTRACE_SYSCALL` request to ptrace
```
Susurrus: This comment should be more descriptive on the first sentence and include a really good extended description explaining what it does.

src/sys/ptrace.rs
```diff
@@ -140,3 +140,51 @@ pub fn setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
...
+        ptrace(
+            ptrace::PTRACE_SYSCALL,
+            pid,
+            ptr::null_mut(),
+            ptr::null_mut(),
+        ).map(|_| ()) // ignore the useless return value
+    }
+}
+
+/// Makes the `PTRACE_ATTACH` request to ptrace
```
Susurrus: Better docs, like above.

src/sys/ptrace.rs
```diff
@@ -140,3 +140,51 @@ pub fn setsiginfo(pid: Pid, sig: &siginfo_t) -> Result<()> {
         Err(e) => Err(e),
     }
 }
+
+/// Sets the process as traceable with `PTRACE_TRACEME`
```
Susurrus: Needs an extended description that elaborates a bit here.

Susurrus:
Please don't break up PRs unless we ask you too as it makes it hard for us to follow the discussion (see my comments about safety). It also generates a lot of email traffic for a lot of us.

src/sys/ptrace.rs
marmistrz: I wanted to keep consistent with the C API: `PTRACE_TRACEME => ptrace::traceme`, so `PTRACE_CONT => ptrace::cont`

src/sys/ptrace.rs
marmistrz: Should I copy-paste the ptrace manual? I thought that the user should be familiar with what ptrace does.

src/sys/ptrace.rs
marmistrz: And thus break the wrapper: `ptrace::syscall` calls `ptrace::ptrace` with `PTRACE_SYSCALL`.

`ptrace` can dereference arbitrary pointer in the child process memory.

src/sys/ptrace.rs
Susurrus: Not the whole thing, but providing a little more detail would be nice. `nix` documentation can also be used as a form of discovery for users who aren't familiar with the POSIX APIs. While I don't expect `nix` to replace the man pages or other sources of documentation, it should allow people unfamiliar with the concepts to understand a little of the functionality.

Susurrus:
@marmistrz Let's refocus #666 on just adding the register ptrace wrapper since that's what most of it is about. Closing it we lose the history of the discussion, which has mostly been around that single feature, so let's keep it around.

src/sys/ptrace.rs
marmistrz: Sorry, `continue` will not work:
```
error: expected identifier, found keyword `continue`
```
Leaving this as it is.

marmistrz:
I'm wondering about the `ptrace::cont`. From manpage:
```
Restart the stopped tracee process.  If data is nonzero, it is interpreted as the number of a signal to be delivered to the tracee; otherwise, no signal is delivered.  Thus, for example,  the
              tracer can control whether a signal sent to the tracee is delivered or not.  (addr is ignored.)
```
Currently `ptrace::cont` uses `data = 0`, so delivers no signal. We could solve it in two ways:

1. Change the signature to 
```
pub fn cont(pid: Pid, sig: Option<Signal>) -> Result<()> {
```

2. Create two functions:
```
pub fn cont(pid: Pid) -> Result<()> {
pub fn cont_signal(pid: Pid, sig: Signal) -> Result<()> {
```

I really think about settling with solution 2 because the most common use case is not to deliver any signal. What do you think?

Susurrus:
Let's use:
```
pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()>;
```

I don't see the need for two functions just because it's too hard to add a `None` argument. It's a goal of nix to stay as close as is reasonable to the C APIs.

marmistrz:
It looks like we've got some problems with qemu. Should I disable the test on these arm/ppc?

Susurrus:
@marmistrz If you think it's a QEMU failure, just document it and disable them for it.

marmistrz:
The tests have succeeded but Github still shows them as not yet finished: https://travis-ci.org/nix-rust/nix/builds/259406655?utm_source=github_status&utm_medium=notification

test/sys/test_ptrace.rs
```diff
@@ -58,6 +58,7 @@ fn test_ptrace_cont() {
 
     // FIXME: qemu-user doesn't implement ptrace on all arches and gives ENOSYS then.
     // use it to filter out the affected platforms
```
Susurrus: Both of these comments could use proper punctuation and capitalization.

marmistrz:
Is it ok now?

Susurrus:
@marmistrz Not yet. You haven't expanded the list of ptrace operations that return and `Unsupported` error. This is also missing a Changed entry in the CHANGELOG. Should be something like "Using `ptrace()` with `PTRACE_TRACEME`, ... is no longer supported. Instead use the specialized functions `traceme()`, ..."

marmistrz:
@Susurrus, but this will break traceme, which uses ptrace internally.

Susurrus:
@marmistrz You're correct, you should change these functions to use `libc::ptrace` or a helper function. This is how the existing specialized wrapper functions work.

marmistrz:
So you're asking me to add 4 more matches now, 6 more in #666... That's poor for readability, maintanance, that's ~3 rows of listing enum variants! And there are about 30 ptrace routines!

Man, it's breaking other people's code without a single, smallest warning, without a single compile-time message. If a program starts tracing another using `PTRACE_SEIZE`, it may not realize something's going on until the program enters an unfortunate branch!! And it's so easy to `cargo update`.

Let's just mark it as deprecated:
```
#[deprecated(since="...", note="please use specialized wrappers instead or contribute one if such doesn't exist")]
```

Susurrus:
Yes I am. This isn't that onerous and is not a large maintenance concern. Eventually `ptrace()` will be completely removed from our external API and only the specialized variants will be supported. Until then, we'll be making each variant return `Unsupported` errors as they're implemented. `#deprecated` doesn't quite have the right semantics, since the function as a whole isn't deprecated, just some variants of it.

As to the breaking people's code piece, yes we are. Every release will likely have breaking changes as we improve or stabilize our APIs. `nix` is pre-1.0 for a reason and breakages should be expected. You should also be wary of `cargo update` if you use any pre-1.0 crates for exactly this reason.

marmistrz:
@Susurrus It's one thing to break the API so it doesn't compile and another to introduce runtime error in the most unexpected place. API changes should be clear from the developers perspective and this, in any case, is not. And I can't stress how sneaky the change is - you may realized it weeks after updating the dependency because you simply used the supported operations.

It's just as though Rust decided that `try!` will panic instead of returning a result, and you've never realized it because you never had any errors before. Changing the interface is one thing, changing the semantics - another. 

Moreover, Rust favors compile-time checking and falls back to runtime errors only when it's absolutely necessary.

Deprecating a 1/10 of an API because we have a 1/10 of a new API is just insane. It's just as though you had to use methods `.minus()`, `.multiply()` to multiply complex numbers but `.add()` threw a runtime error out of nowhere because you should a `+` operator.

I suggest we annotate the current state of the API in the documentation. The general `ptrace` function should be simply removed when all the wrappers are ready, and since we're `0.x` - we can simply remove it.

Susurrus:
@marmistrz I understand your viewpoint on this, but I don't see it as as big of an issue as you suggest.

That being said, I think we should go ahead and add `#[deprecated(since="0.10.0", note="usages of `ptrace()` should be replaced with the specialized helper functions instead")]`.

Additionally I'm fine with you not expanding the `UnsupportedOperation` part of `ptrace()`.

marmistrz:
It's great we've found a compromise! I think we should document the internal functions `ptrace_other` and `ptrace_peek` (possibly in a separate pull request)

As far as I can see, `ptrace_other` takes care of those routines which do not return anything interesting and `ptrace_peek` - those who do (so peeks). But why does `ptrace_peek` match against `Err(Error::Sys(Errno::UnknownErrno))`?

I think the name `ptrace_other` could be clearer. I had an idea to create a secon d function, `fn ptrace_noretval(...) -> Result<()>`, which would save us the `.map(|_| ())` in the specialized wrappers. And remove `ptrace_other` just when we remove `ptrace`. What do you think?

Susurrus:
@marmistrz I think you're on the right track for cleaning up these helper functions. I don't think you need feedback from me, so why don't you go ahead and clean those up. You can put it as the first commit in this series or file a new PR, whatever you want.

As to why `UnknownErrno` is checked for, I'm not certain. It was there in the first implementation added in a0fc607.

marmistrz:
Sure. I'll do this in a separate PR. Let's finally have it merged :)

Susurrus:
Thanks for your hard work pushing on this @marmistrz! Love that we have more tests for this too.

bors r+

PR #672, opened 2017-07-12
add poll module in Android

ndusart:
`poll` functions are defined in Android as well.

libc is missing some constant, but once rust-lang/libc#663 is merged, it'll be good to merge here.

Closes #711

PR #710, opened 2017-07-29
Let freebsd and openbsd access 'poll'

berkowski:
Gave freebsd and openbsd access to 'poll'.  Don't see a reason why other bsd's shouldn't be added as well.  Might make more sense to add exclusive `#cfg(not(...)` statements around a few OS's instead of long strings of acceptable targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
-- Enabled `nix::poll` for 'freebsd' and 'netbsd' targets
 
 ### Changed
+- Enabled `nix::poll` for 'freebsd' and 'netbsd' targets. ([#710](https://github.com/nix-rus/nix/pull/710))
 - Renamed existing `ptrace` wrappers to encourage namespacing ([#692](https://github.com/nix-rust/nix/pull/692))
 
 ## [0.9.0] 2017-07-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Enabled `nix::poll` for 'freebsd' and 'netbsd' targets
 
 ### Changed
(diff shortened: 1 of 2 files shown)


asomers:
Thanks for fixing this.  Could you please add a CHANGELOG entry too?

berkowski:
Updated changelog.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Enabled `nix::poll` for 'freebsd' and 'netbsd' targets
```
asomers: This should have the same format as other CHANGELOG messages.  Move it into the `Changed` section, and add a link to the PR

berkowski: Yup, my mistake.  Sorry for the noise

berkowski:
Added #711 as this PR doesn't fix the issue for other BSD's

asomers:
Is there some reason not to add NetBSD and Dragonfly support right now?  Do they have `poll` definitions in libc?

berkowski:
Only reason at the moment is that I can't easily check locally by using `cross` to easily cross-compile with an appropriate bsd toolchain (only `x86_64-unknown-freebsd` and `-netbsd` are available as targets in `rustup`).

I could go through `libc`, see how things are supported by other BSD's, and try to add support; just can't really test it.  Maybe as a separate PR?

asomers:
You should inspect the code for libc and enable `poll` on all platforms that have libc support.  Better yet would be to inspect the C libraries for each of the BSDs, and add libc `poll` bindings for any platforms that lack it.  Our CI system will catch any mistakes for FreeBSD or NetBSD.  OpenBSD and Dragonfly aren't officially supported by Nix, so nobody will be upset with you if you break the build for one of those platforms.

berkowski:
Fair enough.  I'll expand the PR to the other BSD's supported by [rust-lang/libc](https://github.com/rust-lang/libc)

Susurrus:
This is already covered in #672 which is close to merging.

berkowski:
Great.  Well I just made a small fuss about nothing then.  Sorry for that, didn't catch it in the active issues or PR's

PR #712, opened 2017-07-29
Change the FreeBSD builder's name to "amd64_fbsd11"

asomers:


asomers:
I already made this change on the buildbot server, so we need to merge this PR before bors will accept any others.

bors r+

PR #715, opened 2017-07-31
Fix multiple issues with POSIX AIO

### LINKED ISSUE #714: AIOs that fail to enqueue should not be marked as `in_progress`
When enqueuing an AIO, such as `read`, `write`, and `fsync`, the `AioCb` is marked as `in_progress` before the system call is made.

However, if the system call returns an error, such as `EAGAIN` or `ENOSYS`, etc. then the AIO is not actually in progress, and the subsequent `drop` will panic in its attempt at doing an orderly termination of the AIO:

```
WARNING: dropped an in-progress AioCbthread 'writes_must_be_observable_after_completion' panicked at 'Inconsistent AioCb.in_progress value', /Users/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.8.1/src/sys/aio.rs:305
stack backtrace:
   0:        0x10fb43ca3 - std::sys::imp::backtrace::tracing::imp::unwind_backtrace::h884a721e113c3303
   1:        0x10fb45b33 - std::panicking::default_hook::{{closure}}::h7a7d734b2824d103
   2:        0x10fb4577c - std::panicking::default_hook::h3eb11bd6cbfdc331
   3:        0x10fb47ca7 - std::panicking::rust_panic_with_hook::h8b9b25777425677b
   4:        0x10fb01125 - std::panicking::begin_panic::he221b30de772fec8
   5:        0x10fb03bf5 - <nix::sys::aio::AioCb<'a> as core::ops::Drop>::drop::hfbac5799b913ac7f
   6:        0x10fafee34 - core::ptr::drop_in_place::hd03fe87c6acd00d7
   7:        0x10fafedd7 - core::ptr::drop_in_place::hc23ed067f74bc487
```

Also note that the `WARNING` printed above is missing a line-break, and thus messes up the output.

The above was produced on MacOS Sierra, but I think it's the same on any relevant platform.

asomers:
Fixed error handling in `AioCb::fsync`, `AioCb::read`, and `AioCb::write`
    
    Previously, the `AioCb`'s `in_progress` field would erroneously be set
    to `true`, even if the syscall had an error
    
    Fixes #714

AioCb::Drop will now panic for in-progress AioCb

    Printing a warning message to stderr isn't really appropriate, because
    there's no way to guarantee that stderr is even valid.  Nor is
    aio_suspend necessarily an appropriate action to take.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 # Fixed
 - Fix compilation and tests for OpenBSD targets
   ([#688](https://github.com/nix-rust/nix/pull/688))
+- Fixed error handling in `AioCb::fsync`, `AioCb::read`, and `AioCb::write`.
+  It is no longer an error to drop an `AioCb` that failed to enqueue in the OS.
+  ([#715](https://github.com/nix-rust/nix/pull/715))
 
 # Removed
 - The syscall module has been removed. This only exposed enough functionality for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#744](https://github.com/nix-rust/nix/pull/744))
 - Moved constants ptrace request, event and options to enums and updated ptrace functions and argument types accordingly. 
   ([#749](https://github.com/nix-rust/nix/pull/749))
+- `AioCb::Drop` will now panic if the `AioCb` is still in-progress ([#715](https://github.com/nix-rust/nix/pull/715))
 
 # Fixed
 - Fix compilation and tests for OpenBSD targets
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,10 @@ name = "test"
 path = "test/test.rs"
 
 [[test]]
+name = "test-aio-drop"
+path = "test/sys/test_aio_drop.rs"
+
+[[test]]
 name = "test-mount"
 path = "test/test_mount.rs"
 harness = false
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Renamed existing `ptrace` wrappers to encourage namespacing ([#692](https://github.com/nix-rust/nix/pull/692))
+- `AioCb::Drop` will now panic if the `AioCb` is still in-progress ([#715](https://github.com/nix-rust/nix/pull/715))
+
+### Fixed
+- Fixed error handling in `AioCb::fsync`, `AioCb::read`, and `AioCb::write` ([#715](https://github.com/nix-rust/nix/pull/715))
```
Susurrus: I think adding the specific thing that was fixed here would be good rather than having people see this and then having to reference the PR.

Susurrus:
@asomers That's a more clear CHANGELOG entry, thanks. But I must ask why are these tests only ran on FreeBSD?

Susurrus:
But there are still errors on Mac and a musl target that will need to be fixed.

asomers:
It seems that on Linux, `aio_write` and `aio_read` return success even if the arguments are invalid, contrary to the man page.  I don't know what's happening on OSX.  Hopefully I'll know after this latest test run.

asomers:
Somehow I seriously screwed up the squash.  I'll try to fix it.

asomers:
It looks like bors is ignoring me because chrisvest doesn't have the authority to approve a merge?  I thought that everyhhing following "bors r+" was ignored, but I guess not.

bors r+

notriddle:
It wasn't ignoring you. It was just being slow.

PR #716, opened 2017-07-31
Fix changelog from bad merge

Susurrus:
Missed this when merging #672.


Susurrus:
Hmm, bors never got around to finishing this. Let's try again.

bors r+ asomers

PR #718, opened 2017-08-02
Add a proc_alive helper function.

marmistrz:


src/sys/signal.rs
```diff
@@ -722,6 +735,25 @@ mod tests {
         assert_eq!(action_ign.handler(), SigHandler::SigIgn);
     }
 
+    #[test]
+    fn test_procalive() {
+        use sys::wait::waitpid;
+        use unistd::fork;
+        use unistd::ForkResult::*;
+
+        match fork() {
```
asomers: Calling `fork` conflicts with other tests that call `wait`.  So you must grab the `FORK_MTX` at the top of the test.

src/sys/signal.rs
```diff
@@ -722,6 +735,25 @@ mod tests {
...
+            Ok(Child) => loop {},
+            Ok(Parent { child }) => {
+                assert_eq!(proc_alive(child).unwrap(), true);
+                kill(child, SIGTERM).unwrap();
```
asomers: Signal handlers are inherited across a fork, so any test that installs a `SIGTERM` handler will screw up this test.  You should therefore grab the `SIGNAL_MTX` at the top of this test.  If you grab multiple mutexes, make sure to do it in alphabetical order to prevent deadlocks.

asomers:
What is the intended use case for this function?  The obvious uses are all racy, since the process could exit before `proc_alive` returns.

src/sys/signal.rs
marmistrz: Good point.

Susurrus:
Sounds like this should just closed then.

PR #720, opened 2017-08-04
[Do not merge] This PR is only to test if current master passes on Android targets

marmistrz:


marmistrz:
This should be fixed by rust-lang/libc#713

PR #721, opened 2017-08-07
Enable setresuid/setresgid on FreeBSD and OpenBSD

### LINKED ISSUE #717: Enable setresuid and setresgid for more platforms
`unistd::setresuid` and `unistd::setresgid` are currently wrapped in a module that only is only included on Linux and Android. The functions appear to be present on other systems as well:

- [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=setresuid&apropos=0&sektion=0&manpath=FreeBSD+11.1-RELEASE+and+Ports&arch=default&format=html)
- [OpenBSD](http://man.openbsd.org/setresuid)

NetBSD and macOS don't appear to have them.

wezm:
Fixes #717 

This PR enables `setresuid` and `setresgid` on FreeBSD and OpenBSD. These functions are not present on NetBSD, macOS, and iOS as far as I can tell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   has changed type from `c_int` to `SockProtocol`.
   It accepts a `None` value for default protocol that was specified with zero using `c_int`.
   ([#647](https://github.com/nix-rust/nix/pull/647))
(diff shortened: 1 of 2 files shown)


asomers:
Looks good to me.  Just squash the two commits into one and I'll merge it.

wezm:
Done 👍

PR #722, opened 2017-08-07
Add a convenience method .pid() for WaitStatus.

marmistrz:


test/sys/test_wait.rs
```diff
@@ -37,6 +37,19 @@ fn test_wait_exit() {
     }
 }
 
+#[test]
+fn test_waitstatus_pid() {
+    let _m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+    match fork().unwrap() {
+        Child => {},
```
asomers: Don't allow the child to return.  It's vulnerable to deadlocks in the test harness code.  Instead, call `_exit(0)`.

marmistrz: You mean that the test harness executes the function as a normal function, so the forked process will duplicate what test harness does?

Why `_exit` and not `std::process::exit`? The former is `unsafe`.

asomers: Yeah, the forked process duplicates the entire test harness.  But it doesn't duplicate any threads other than the one that called `fork`.  So if one of those other threads held a mutex that is required by this thread during its teardown phase, then this thread will deadlock.  The reason to use `_exit` instead of `sys::process::exit` is subtler, and it's because `_exit` doesn't do any C-level teardown.  See [fork(2)](http://man7.org/linux/man-pages/man2/fork.2.html).

marmistrz: You mean the reasons like this: https://stackoverflow.com/questions/5422831/what-is-the-difference-between-using-exit-exit-in-a-conventional-linux-fo
(double buffer flushing or double execution of `atexit` handlers?)

asomers: Like that, but worse because the parent is a multi-threaded process.  In a multi-threaded process, even memory allocation could theoretically deadlock, so until the child execs, it is limited to only calling async-signal-safe functions, which are a pretty restrictive set.

asomers:
This all looks good to me, though you still need to squash.  I'll give @Susurrus a day or so to comment before I merge it.

marmistrz:
@asomers Can you squash while merging? The GitHub web UI allows it.

asomers:
@marmistrz I can't squash while merging, because nix doesn't use the GitHub web UI.  We use bors instead.  bors prevents test failures caused by merge conflicts, but unfortunately can't automatically squash.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#672](https://github.com/nix-rust/nix/pull/672))
 - Added protocol families in `AddressFamily` enum.
   ([#647](https://github.com/nix-rust/nix/pull/647))
+- Added a convenience method `.pid()` to the `enum WaitStatus`
```
Susurrus: Instead "Added the `pid()` method to `WaitStatus` for extracting the PID."

src/sys/wait.rs
```diff
@@ -93,6 +93,25 @@ pub enum WaitStatus {
     StillAlive
 }
 
+impl WaitStatus {
+    /// Extracts the PID from the WaitStatus if the status is not StillAlive.
+    pub fn pid(&self) -> Option<Pid> {
+        use self::WaitStatus::*;
+        match self {
```
Susurrus: I'm not certain if this [StackOverflow comment](https://stackoverflow.com/questions/36590549/matching-on-a-reference-to-an-enum) is correct, but it would read nicer if you did `match *self` to get rid of the `&` on every match arm.

marmistrz:
Squashed.

PR #725, opened 2017-08-09
Match syntax of libc_bitflags! with bitflags!

Susurrus:
Also update a couple of constant declarations while we're at it.

### DIFF
--- a/src/sys/select.rs
+++ b/src/sys/select.rs
@@ -1,57 +1,37 @@
-use std::ptr::null_mut;
+use std::mem;
 use std::os::unix::io::RawFd;
-use libc::{c_int, timeval};
+use std::ptr::null_mut;
+use libc::{self, c_int};
 use {Errno, Result};
 use sys::time::TimeVal;
 
-pub const FD_SETSIZE: RawFd = 1024;
+pub use libc::FD_SETSIZE;
 
-#[cfg(any(target_os = "macos", target_os = "ios"))]
+// FIXME: Change to repr(transparent) once it's stable
 #[repr(C)]
-#[derive(Clone)]
-pub struct FdSet {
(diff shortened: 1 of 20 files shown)


Susurrus:
@asomers Want to give this a quick glance-over for me? It should be pretty straight-forward.

src/sys/select.rs
```diff
@@ -1,70 +1,38 @@
...
-const BITS: usize = 64;
+pub struct FdSet(libc::fd_set);
 
 impl FdSet {
     pub fn new() -> FdSet {
-        FdSet {
-            bits: [0; FD_SETSIZE as usize / BITS]
-        }
+        let mut fdset = unsafe { mem::uninitialized() };
+        unsafe { libc::FD_ZERO(&mut fdset) };
```
asomers: This shouldn't require `unsafe`.  `FD_ZERO` is implemented in pure Rust.  Same for `FD_CLR`, `FD_SET`, and `FD_ISSET`

Susurrus: Same thing as `WIFCONTINUED`, this is how they're defined in `libc`:
```rust
pub unsafe extern "C" fn FD_SET(fd: c_int, set: *mut fd_set);
```

Susurrus:
@asomers Want to give this another look over now? Should be RFM.

asomers:
LGTM.  But can you squash the last commit?

Susurrus:
@asomers Squash that last commit into which other commit? I left it separate because it didn't seem appropriate to merge it with any other changes in this patchset.

asomers:
Sorry, I thought that the extra `AtFlags` definition was also introduced by this PR.  But now I see that it wasn't.

bors r+

Issue #594, opened 2017-04-24
Taking RawFd by value is not sound


kamalmarhubi:
[Raised by @lucab on #562 ](https://github.com/nix-rust/nix/pull/562#issuecomment-296684917)
>  I *think* these all do an implicit copy+own of the RawFd (as it is just an alias to an int), and I fear this doesn't really guarantee that the fd has not been closed/dropped in the meanwhile by a `close()`.

This is a problem. A `RawFd` value may be kept long enough to end up referring to a different file. We may wish to provide a non-copy wrapper type similar to `mio`'s `EventedFd` that allows enforcing the borrowing relationship. From [the docs](https://docs.rs/mio/0.6.6/mio/unix/struct.EventedFd.html):

> Note that `EventedFd` takes a `&RawFd`. This is because `EventedFd` **does not** take ownership of the FD. Specifically, it will not manage any lifecycle related operations, such as closing the FD on drop.

lucab:
Do you know if this concern has been raised within stdlib before? I think `RawFd` may benefit from being `Clone` via `dup(2)`, but I miss the point of `Copy` (at least on linux).

kamalmarhubi:
`dup(2)` can fail, eg if a file descriptor limit is reached. I don't think it makes sense to have a fallible `Clone` impl. And I don't think the stdlib would be interested in this change. The wording `Raw` indicates moving outside of Rust's guarantees. [Most every occurrence of `raw` in the stdlib](https://doc.rust-lang.org/stable/std/rc/struct.Rc.html?search=raw) has that connotation. For this case, note that the [`FromRawFd` trait](https://doc.rust-lang.org/stable/std/os/unix/io/trait.FromRawFd.html)'s method is unsafe.

More likely we should introduce a nix-specific wrapper that enforces safe usage. This needs a bit of thought!

Kixunil:
> More likely we should introduce a nix-specific wrapper that enforces safe usage. This needs a bit of thought!

I was thinking exactly like this.

* Keep `RawFd` for interfacing with other crates and raw C libraries
* Provide `Fd` which is a newtype for `RawFd` that impls `Drop` and `{From,Into,As}Fd` traits.
* Implement all other wrappers for `Fd` (files, sockets, ...). This has nice benefit of not having to impl `Drop` on all other types
[...]

PR #727, opened 2017-08-09
unistd: add fexecve()

lucab:
This adds fexecve(). It is available in libc since 0.2.29.

Ref: http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -593,6 +593,37 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
     Err(Error::Sys(Errno::last()))
 }
 
+/// Replace the current process image with a new one (see
+/// [fexecve(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html)).
+///
+/// The `fexecve` function allows for another process to be "called" which will
+/// replace the current process image.  That is, this process becomes the new
+/// command that is run. On success, this function will not return. Instead,
+/// the new program will run until it exits.
+///
+/// This function is similar to `execve`, except that the program to be executed
+/// is referenced as a file descriptor instead of a path.
+///
+/// # Errors
+///
+/// If an error occurs, this function will return with an indication of the
+/// cause of failure.  See
+/// [fexecve(2)#errors](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html#tag_16_111_05)
+/// for a list of error conditions.
+#[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd",
+          target_os = "netbsd", target_os = "openbsd", target_os = "linux"))]
+#[inline]
+pub fn fexecve(fd: RawFd, args: &[CString], env: &[CString]) -> Result<Void> {
+    let args_p = to_exec_array(args);
+    let env_p = to_exec_array(env);
+
+    unsafe {
+        libc::fexecve(fd, args_p.as_ptr(), env_p.as_ptr())
+    };
+
+    Err(Error::Sys(Errno::last()))
+}
+
 /// Daemonize this process by detaching from the controlling terminal (see
 /// [daemon(3)](http://man7.org/linux/man-pages/man3/daemon.3.html)).
 ///
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -105,7 +105,7 @@ mod linux_android {
 }
 
 macro_rules! execve_test_factory(
-    ($test_name:ident, $syscall:ident, $unix_sh:expr, $android_sh:expr) => (
+    ($test_name:ident, $syscall:ident, $exe: expr) => (
     #[test]
     fn $test_name() {
         #[allow(unused_variables)]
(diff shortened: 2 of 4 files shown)


src/unistd.rs
```diff
@@ -593,6 +593,33 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
...
+/// [fexecve(3)](http://man7.org/linux/man-pages/man3/fexecve.3.html)).
+///
+/// The `fexecve` function allows for another process to be "called" which will
+/// replace the current process image.  That is, this process becomes the new
+/// command that is run. On success, this function will not return. Instead,
+/// the new program will run until it exits.
+///
+/// If an error occurs, this function will return with an indication of the
+/// cause of failure.  See
+/// [fexecve(3)#errors](http://man7.org/linux/man-pages/man3/fexecve.3.html#ERRORS)
```
asomers: Since `fexecve` is not linux-specific, could you please link to the Open Group man page instead?

Susurrus:
One thing that hasn't been addressed with this functions is their use of `CString` as arguments. This is partially touched on in #358.

lucab:
@asomers amended after review, PTAL.

@Susurrus right. For reference this is also affected by https://github.com/nix-rust/nix/issues/594. I don't have good answers/proposals for any of those. I would refrain from tackling those here, I can personally accept a later breaking change around this.

src/unistd.rs
```diff
@@ -593,6 +593,33 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
...
 
+/// Replace the current process image with a new one (see
+/// [fexecve(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html)).
+///
+/// The `fexecve` function allows for another process to be "called" which will
+/// replace the current process image.  That is, this process becomes the new
+/// command that is run. On success, this function will not return. Instead,
+/// the new program will run until it exits.
+///
+/// If an error occurs, this function will return with an indication of the
```
Susurrus: This should be in an "Errors" section (see [the book](https://doc.rust-lang.org/book/first-edition/documentation.html#special-sections)).

src/unistd.rs
```diff
@@ -593,6 +593,33 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
...
+/// cause of failure.  See
+/// [fexecve(2)#errors](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html#tag_16_111_05)
+/// for a list of error conditions.
+///
+/// This function is similar to `execve`, except that the program to be executed
```
Susurrus: Let's move this above the errors section.

lucab:
@Susurrus docstring amended.

Susurrus:
@asomers Can you confirm that your review changes have been addressed?

@lucab Any way we can implement some tests for this? Or maybe even some code examples?

lucab:
@Susurrus I have a full example/demo at https://github.com/lucab/fxe-rs. Not sure if you want to extrapolate anything from there, as it makes more sense as part of the whole explanation.

Regarding tests, I guess I can try adapting this [test macro(https://github.com/nix-rust/nix/blob/master/test/test_unistd.rs#L107) to receive the first exec parameter from the outside.

Let me know if you want me too add more stuff in here, or if it is fine already fine having an external demo/consumer for the moment.

Susurrus:
@lucab That'd be great if you could get a test working for this! The examples aren't as important, I just want something in-tree that uses this code, and having a test is the best, an example is second-best, and we're trying to avoid adding anything that doesn't include tests.

lucab:
@Susurrus pushed one more commit with the test (CI is mostly green by now but still ongoing). PTAL.

Susurrus:
LGTM. @asomers can you wrap up your review? Then I think we're ready to merge!

Susurrus:
There are build failures on mac because `fexecve` is declared there in `libc`, but doesn't actually exist. Can you remove it from `macos` and `ios` for this PR? I've already filed some work with upstream to be `libc` cleaned up (rust-lang/libc#732).

lucab:
@Susurrus I feel that bad that I didn't catch this when wiring-in the libc bits. Thanks for spotting it and promptly fixing there. I've amended to exclude macos/ios, CI is greeen now.

src/unistd.rs
```diff
@@ -593,6 +593,36 @@ pub fn execvp(filename: &CString, args: &[CString]) -> Result<Void> {
...
+/// is referenced as a file descriptor instead of a path.
+///
+/// # Errors
+///
+/// If an error occurs, this function will return with an indication of the
+/// cause of failure.  See
+/// [fexecve(2)#errors](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html#tag_16_111_05)
+/// for a list of error conditions.
+#[cfg(not(any(target_os = "macos", target_os = "ios")))]
```
Susurrus: We prefer to use white-lists here instead of black-lists as it makes it easier to support new platforms. Can you change this and the other ones below?

test/test_unistd.rs
```diff
@@ -156,6 +150,15 @@ macro_rules! execve_test_factory(
     )
 );
 
+#[cfg(not(target_os = "android"))]
```
Susurrus: This will be cleaner if you use `cfg_if!` and group the platform-specific code together.

Susurrus:
@lucab No worries. I have a Mac, so i wanted to check that support for it was in fact missing as I assumed you didn't have access to one (most people don't). Not a big deal at all!

asomers:
Don't forget to squash your commits before we merge.

lucab:
@Susurrus I've switched `to cfg_if!` and turned all conditionals into explicit whitelists, PTAL. (I am however not really sure this is the aesthetic result you were suggesting; if you had something different in mind, feel free to explicitly detail how you wanted this to look like)

Susurrus:
Awesome, this looks pretty solid, thanks @lucab!

bors r+

Issue #724, opened 2017-08-07
Incorrect value for PTRACE_EVENT_SECCOMP


xd009642:
PTRACE_EVENT_SECCOMP and PTRACE_EVENT_EXIT are both given the value of 6. Instead the seccomp event should have the value 7. Source: https://github.com/torvalds/linux/blob/master/include/uapi/linux/ptrace.h

Susurrus:
Indeed. All constants in `src/sys/ptrace.rs` should be replaced with their libc equivalents. All the `PTRACE_EVENT_*` constants are missing, however, so those should be submitted to upstream for all libc/nix supported platforms. A partial PR regarding these constants is in #461. If you'd like to get those constants upstreamed and then take over #461, it'd be greatly appreciated as I'm not certain @chaosagent is working on it anymore.

Susurrus:
Whoops, misspoke on this. All `PTRACE_EVENT_*` constants have been merged into libc.

@xd009642 Would you be interested in cleaning up the constants in `src/sys/ptrace.rs`? It'd be mostly what's already done in #461 but with the addition of using `PTRACE_EVENT_*` constants from `libc`. Using the `libc_enum!` macro makes this pretty easy.

xd009642:
I am slightly busy at the moment I'll try and fit sometime in this weekend or maybe later this week. Guess it just depends how quickly you'd like this to move along

Susurrus:
@xd009642 I like to move this along as fast you can, whatever that speed is. We just don't have enough people who work on nix directly to take care of many of these types of issues, instead relying on external contributors to scratch their own itch. So if you weren't willing to take this on, I don't know when this would get resolved.

xd009642:
That's fair enough, I'll probably start looking at it tonight or tomorrow.

xd009642:
I've started work on it today, so I'll use this issue to ask questions/float ideas until I've submitted something I'm somewhat happy with if that's okay? 

I might not use the libc_enum macro because there look to be some type (signed versus unsigned) discrepancies and things available in some OS's but not others. I'm still working it out so it's early days though. Also, libc removed PTRACE_EVENT_STOP as it is defined in glibc 2.26. I may define the event enum with the macro so i can add the stop event field for completeness

xd009642:
Yeah I'll raise the PR when I've got something that's actually worth showing in some form then continue discussion there. And I'll look into the source examples

xd009642:
Just going to close this issue as it was resolved in #749 merged today

PR #728, opened 2017-08-10
Use upstream libc ptrace FFI

Susurrus:
This might fail on Mac because of how libc defines ptrace on that platform, but we'll see!

### DIFF
--- a/src/sys/ptrace.rs
+++ b/src/sys/ptrace.rs
@@ -2,14 +2,21 @@
 
 use std::{mem, ptr};
 use {Errno, Error, Result};
-use libc::{c_void, c_long, siginfo_t};
+use libc::{self, c_void, c_long, siginfo_t};
 use ::unistd::Pid;
 use sys::signal::Signal;
 
 pub mod ptrace {
     use libc::c_int;
 
-    pub type PtraceRequest = c_int;
+    cfg_if! {
+        if #[cfg(any(all(target_os = "linux", arch = "s390x"),
+                     all(target_os = "linux", target_env = "gnu")))] {
+            pub type PtraceRequest = ::libc::c_uint;
+        } else {
+            pub type PtraceRequest = c_int;
+        }
+    }
 
     pub const PTRACE_TRACEME:     PtraceRequest = 0;
     pub const PTRACE_PEEKTEXT:    PtraceRequest = 1;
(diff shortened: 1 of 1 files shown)


Susurrus:
We don't expose this on mac, which is good because that API is considerably different. So this should be good to merge.
=

Susurrus:
@marmistrz Wanna give this a quick look-over since it pertains to the ptrace work you've been doing?

src/sys/ptrace.rs
```diff
@@ -1,12 +1,23 @@
...
+    /// The datatype used for the `request` argument to `ptrace`
+    #[cfg(any(all(target_os = "linux", arch = "s390x"),
+              all(target_os = "linux", target_env = "gnu")))]
+    #[doc(hidden)]
+    pub type ptrace_request_type = ::libc::c_uint;
+
+    #[cfg(not(any(all(target_os = "linux", arch = "s390x"),
+                  all(target_os = "linux", target_env = "gnu"))))]
+    #[doc(hidden)]
+    pub type ptrace_request_type = ::libc::c_int;
```
marmistrz: `__ptrace_request` is an enum in C. Won't we be cafe with an int?

src/sys/ptrace.rs
```diff
@@ -83,7 +86,7 @@ pub fn ptrace(request: ptrace::PtraceRequest, pid: Pid, addr: *mut c_void, data:
 fn ptrace_peek(request: ptrace::PtraceRequest, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
     let ret = unsafe {
         Errno::clear();
-        ffi::ptrace(request, pid.into(), addr, data)
+        libc::ptrace(request, Into::<libc::pid_t>::into(pid), addr, data)
```
marmistrz: `libc::pid_t::from(pid)` would be cleaner. same in several other places.

src/sys/ptrace.rs
```diff
@@ -1,12 +1,23 @@
...
 
 pub mod ptrace {
     use libc::c_int;
 
-    pub type PtraceRequest = c_int;
+    /// The datatype used for the `request` argument to `ptrace`
+    #[cfg(any(all(target_os = "linux", arch = "s390x"),
+              all(target_os = "linux", target_env = "gnu")))]
+    #[doc(hidden)]
+    pub type ptrace_request_type = ::libc::c_uint;
```
marmistrz: You're importing `libc::self`, so the leading `::` is not needed.

src/sys/ptrace.rs
Susurrus: Enums are `c_int` by default in C. But I'm following the argument type dictated by `libc`'s definition of `ptrace()` across all supported platforms.

src/sys/ptrace.rs
Susurrus: Actually `libc::self` is not imported into this module, so the leading `::` is necessary here.

Susurrus:
@marmistrz Thanks for the review. This should clean things up a bit nicely.

src/sys/ptrace.rs
marmistrz: Ah, indeed, it's a submodule.

Susurrus: Yeah, we'll remove it when we refactor all these constants as part of #724.

PR #729, opened 2017-08-10
Refactor statvfs module

Susurrus:
* Use upstream libc FFI declarations and constants
 * Use more Rust-y interface of returning entire struct versus passing it as an
   argument to (f)statvfs.
 * Replace field accesses with accessor methods
 * Flatten vfs module into parent
 * Remove all non-libc-based interfaces for working with the statvfs struct


asomers:
Everything here looks great to me.  Nice improvement to the API too.  But you need a CHANGELOG entry for the API change.

Issue #676, opened 2017-07-15
Improve `unistd::Whence`


Susurrus:
Needs a few things here:
 * [ ] Whence needs to be documented
 * [ ] Hardcoded constants for `SEEK_DATA` and `SEEK_HOLE` need to be converted to use `libc` types.
 * [ ] `SeekData` and `SeekHole` should only be available on supported platforms.
 * [ ] Instead of having different enum values and libc values, just use the libc values directly for each enum variant. Maybe improve the `libc_enum!` to support variable renames to make this easy.

ncaracci:
Is it okay to use the descriptions found on linux's lseek manpage?

asomers:
For Linux-specific constants, yes.  But for standardized constants, it's usually better to use the OpenGroup's descriptions.  You can find them at http://pubs.opengroup.org/onlinepubs/9699919799/functions/lseek.html .

ncaracci:
Would this be okay?
```
/// Directive that tells lseek how to reposition the offset of the file.
pub enum Whence {
    /// The file offset shall be set to offset bytes.
    SeekSet,
    /// The file offset shall be set to its current location plus offset.
    SeekCur,
    /// The file offset shall be set to the size of the file plus offset.
    SeekEnd,
    /// Adjust the file offset to the next location in the file greater than
    /// or equal to offset containing data.  If offset points to data, then 
    /// the file offset is set to offset.
    SeekData,
    /// Adjust the file offset to the next hole in the file greater than or
    /// equal to offset.  If offset points into the middle of a hole, then 
    /// the  file  offset is  set to offset.  If there is no hole past offset,
    /// then the file offset is adjusted to the end of the file (i.e., there
    /// is an implicit hole at the end of any file).
    SeekHole
}
```

Susurrus:
@ncaracci Can you put this in a PR? It's easier to comment on these when I can put comments on specific lines. And you're definitely on the right track, so you shouldn't worry about this getting merged.

ncaracci:
Ok, here's the PR #730.

Susurrus:
Closed by 3d7d7ea5445ea0e2626822bd397dd3dad45ac8e6.

Thanks for your contribution @ncaracci!

PR #730, opened 2017-08-10
Documentation for Whence.

ncaracci:


src/unistd.rs
```diff
@@ -737,11 +737,23 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
     Errno::result(res).map(|r| r as usize)
 }
 
+/// Directive that tells lseek how to reposition the offset of the file.
 pub enum Whence {
+    /// The file offset shall be set to offset bytes.
```
Susurrus: "Specify an offset relative to the start of the file"

src/unistd.rs
```diff
@@ -737,11 +737,23 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
...
     SeekSet,
+    /// The file offset shall be set to its current location plus offset.
```
Susurrus: "Specify an offset relative to the current file location"

src/unistd.rs
```diff
@@ -737,11 +737,23 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
...
     SeekCur,
+    /// The file offset shall be set to the size of the file plus offset.
```
Susurrus: "Specify an offset relative to the end of the file"

src/unistd.rs
```diff
@@ -737,11 +737,23 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
...
     SeekEnd,
+    /// Adjust the file offset to the next location in the file greater than
```
Susurrus: "Specify a file offset relative to the next location in the file greater than or equal to offset containing data.  If offset points to data, then the file offset is set to offset."

That being said I really don't understand what "data" is in this context and how "offset" relates to it.

src/unistd.rs
```diff
@@ -737,11 +737,23 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
...
+    /// or equal to offset containing data.  If offset points to data, then 
+    /// the file offset is set to offset.
     SeekData,
+    /// Adjust the file offset to the next hole in the file greater than or
```
Susurrus: Again, this only "specifies", it doesn't "adjust". The `*seek()` functions adjust, these constants are just specifying a position.

src/unistd.rs
```diff
@@ -737,11 +737,23 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
     Errno::result(res).map(|r| r as usize)
 }
 
+/// Directive that tells lseek how to reposition the offset of the file.
```
Susurrus: Please reference `lseek` and `lseek64` and specify them as Markdown links to the documentation. See [here](https://doc.rust-lang.org/src/core/num/mod.rs.html#2635) for an example. You'll need to figure out the exact paths for lseek/lseek64, but the premise is the same.

src/unistd.rs
ncaracci: Data is supposed to mean a part of the file that's been written to. Maybe it should be called 'some data'?

src/unistd.rs
```diff
@@ -737,11 +737,25 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
     Errno::result(res).map(|r| r as usize)
 }
 
+/// Directive that tells [`lseek`] and [`lseek64`] how to reposition the offset of the file.
```
Susurrus: "Directive for [`lseek`] and [`lseek64`] that specifies what the `offset` is relative to."

Susurrus:
If you're planning on tackling all of #676, we can keep that as part of this same PR as additional commits.

Susurrus:
Can you squash these commits all into one (lookup the `git rebase` command if you're unfamiliar with doing this)?

Otherwise are you planning on adding the other parts of #676 to this PR or we just going to merge the documentation part in this one?

ncaracci:
I was thinking on tackling all of #676. I have a question with the conditional availabilty part, SEEK_DATA and SEEK_HOLE are defined on libc only on the linux module, so should I consider that the only platform where they are available?

Susurrus:
You'll want to check all applicable platforms. They should at least be added to DragonFlyBSD ([declared here](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/725edadf86d63f56a584adf23265845c8590d734/sys/sys/unistd.h#L126)). [FreeBSD also has them](https://github.com/freebsd/freebsd/blob/f5d95e1f8d32db4ccccfd5ad9cecb21ed07a695d/sys/sys/unistd.h). It looks like it's not in uclibc or linux/musl. So I think those platforms are the only two missing it. I checked NetBSD, OpenBSD, and Mac and they don't have either of them, so that's fine.

Susurrus:
@ncaracci You'll need to add these constants to libc before this will work. You should file a PR against [libc](https://github.com/rust-lang/libc) adding these constants [here](https://github.com/rust-lang/libc/blob/master/src/unix/bsd/freebsdlike/mod.rs#L935). Then your changes will work.

src/unistd.rs
```diff
@@ -751,8 +767,10 @@ impl Whence {
             &Whence::SeekSet => libc::SEEK_SET,
             &Whence::SeekCur => libc::SEEK_CUR,
             &Whence::SeekEnd => libc::SEEK_END,
-            &Whence::SeekData => 3,
-            &Whence::SeekHole => 4
+            #[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd"))]
```
Susurrus: Don't do this here, delete this whole `impl` and just assign values within the enum like `SeekSet = libc::SEEK_SET,`. You'll also want to say that it's a `i32` datatype. Look at how it's implemented for [`Signal`](https://github.com/nix-rust/nix/blob/master/src/sys/signal.rs#L19).

src/unistd.rs
```diff
@@ -737,36 +737,40 @@ pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
...
+    /// Specify an offset relative to the start of the file.
+    SeekSet = libc::SEEK_SET,
+    /// Specify an offset relative to the current file location.
+    SeekCur = libc::SEEK_CUR,
+    /// Specify an offset relative to the end of the file.
+    SeekEnd = libc::SEEK_END,
+    /// Specify an offset relative to the next location in the file greater than or 
+    /// equal to offset that contains some data. If offset points to
+    /// some data, then the file offset is set to offset.
+    #[cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd"))]
```
Susurrus: Please alphabetize the OS names here and below.

ncaracci:
Would adding `not(target_env="musl",` target_arch="mips")` to the cfg fix the build errors?

Susurrus:
Use a whitelist for Linux instead: `#[cfg(any(target_os = "dragonfly", target_os = "freebsd", all(target_os = "linux", not(any(target_env = "musl", target_arch = "mips"))))]`

Susurrus:
You'll need to exclude arch mip64 as well.

asomers:
Thanks for the improvements, @ncaracci 

bors r+

asomers:
bors timed out because of a known bug in buildbot, triggered by the non-ascii character in the author's name.  I'll manually merge it.

PR #731, opened 2017-08-10
Use upstream libc definitions in mman module

Susurrus:


src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
+        MAP_SHARED,
+        /// Create a private copy-on-write mapping. Mutually exclusive with `MAP_SHARED`.
+        MAP_PRIVATE,
+        /// Place the mapping at exactly the address specified in `addr`.
+        MAP_FIXED,
+        /// Synonym for `MAP_ANONYMOUS`. Deprecated.
+        MAP_ANON,
+        /// The mapping is not backed by any file.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        MAP_ANONYMOUS,
```
asomers: FreeBSD has this one too, though it seems nobody has added it to libc yet.

src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
+    pub flags MapFlags: c_int {
+        /// Compatibility flag. Ignored.
+        MAP_FILE,
+        /// Share this mapping. Mutually exclusive with `MAP_PRIVATE`.
+        MAP_SHARED,
+        /// Create a private copy-on-write mapping. Mutually exclusive with `MAP_SHARED`.
+        MAP_PRIVATE,
+        /// Place the mapping at exactly the address specified in `addr`.
+        MAP_FIXED,
+        /// Synonym for `MAP_ANONYMOUS`. Deprecated.
```
asomers: `MAP_ANON` can't very well be deprecated on platforms where its synonym is undefined, can it?

src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
+        MAP_ANON,
+        /// The mapping is not backed by any file.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        MAP_ANONYMOUS,
+        /// Put the mapping the first 2GB of the process address space.
```
asomers: s/mapping/mapping into/

src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+        MAP_32BIT,
```
asomers: Present on FreeBSD also.

src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        MAP_DENYWRITE,
+        /// Compatibility flag. Ignored.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        MAP_EXECUTABLE,
+        /// Mark the mmaped region to be locked in the same way as `mlock(2)`.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        MAP_LOCKED,
+        /// Do not reserve swap space for this mapping.
+        MAP_NORESERVE,
```
asomers: `MAP_NORESERVE` and `MAP_RENAME` were dropped in FreeBSD 11.  For compatibility, they're still supported in older binaries.  But there are no older binaries using these symbols from Nix.  So we should drop the symbols too, at least on FreeBSD.

src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
+        /// Expect sequential page references.
+        MADV_SEQUENTIAL,
+        /// Expect access in the near future.
+        MADV_WILLNEED,
+        /// Do not expect access in the near future.
+        MADV_DONTNEED,
+        /// Free up a given range of pages and its associated backing store.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        MADV_REMOVE,
+        /// Do not make pages in this rage available to the child after a fork(2).
```
asomers: s/rage/range/

src/sys/mman.rs
```diff
@@ -1,241 +1,251 @@
...
-    #[allow(improper_ctypes)]
-    extern {
-        pub fn shm_open(name: *const c_char, oflag: c_int, mode: mode_t) -> c_int;
-        pub fn shm_unlink(name: *const c_char) -> c_int;
-        pub fn mlock(addr: *const c_void, len: size_t) -> c_int;
-        pub fn munlock(addr: *const c_void, len: size_t) -> c_int;
-        pub fn madvise (addr: *const c_void, len: size_t, advice: c_int) -> c_int;
-        pub fn msync (addr: *const c_void, len: size_t, flags: c_int) -> c_int;
+libc_bitflags!{
+    /// Configuration flags for `msync'.
```
asomers: s/'/`/

Cargo.toml
```diff
@@ -15,7 +15,7 @@ exclude     = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc" }
+libc = { git = "https://github.com/susurrus/libc", branch = "mman_constants" }
```
asomers: Don't forget to undo this part before merging.

src/sys/mman.rs
Susurrus: Probably only deprecated on Linux platforms. In fact it appears that on FreeBSD `MAP_ANON` is the "correct" constant and `MAP_ANONYMOUS` is just there for compatibility. I'll just remove the deprecation notice.

Susurrus:
@asomers I think everything should be ready for merge now that `libc` has all necessary constants.

PR #733, opened 2017-08-11
unistd: Add getgroups, setgroups, getgrouplist, initgroups

JayH5:


src/unistd.rs
```diff
@@ -942,6 +942,42 @@ pub fn setgid(gid: Gid) -> Result<()> {
 }
 
 #[inline]
```
Susurrus: Remove `#[inline]` for both functions. Also need documentation.

src/unistd.rs
```diff
@@ -942,6 +942,42 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+pub fn setgroups(groups: &[Gid]) -> Result<()> {
+    cfg_if! {
+        if #[cfg(any(target_os = "macos",
```
Susurrus: Alphabetize the target_os listing. This should be done everywhere.

src/unistd.rs
```diff
@@ -942,6 +942,42 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+                     target_os = "freebsd",
+                     target_os = "dragonfly",
+                     target_os = "openbsd",
+                     target_os = "netbsd"))] {
+            type setgroups_ngroups_t = c_int;
+        } else {
+            type setgroups_ngroups_t = size_t;
+        }
+    }
+    let gids: Vec<gid_t> = groups.iter().cloned().map(Into::into).collect();
```
asomers: This looks like it does a data copy.  Can you eliminate it?  `sys::aio::lio_listio` does something similar.

src/unistd.rs
```diff
@@ -942,6 +942,42 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    let res = unsafe { libc::setgroups(groups.len() as setgroups_ngroups_t, gids.as_ptr()) };
+
+    Errno::result(res).map(drop)
+}
+
+#[cfg(any(target_os = "linux", target_os = "android", target_os = "macos", target_os = "ios"))]
```
asomers: FreeBSD also has `initgroups`

src/unistd.rs
JayH5: I'm pretty new to Rust and had no idea you could do this. Not entirely comfortable doing so before there are tests, though.

src/unistd.rs
JayH5: Yes, but it's unfortunately not implemented in `rust-lang/libc` for FreeBSD yet.

src/unistd.rs
JayH5: Thanks, was just copying without knowing what I was doing 😅. Docs added.

src/unistd.rs
```diff
@@ -941,6 +941,25 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    // Now actually get the groups
+    let mut groups = Vec::with_capacity(size as usize);
+    let ret = unsafe { libc::getgroups(size, groups.as_mut_ptr()) };
+
+    Errno::result(ret).map(|s| {
+        // Use the size returned from the second getgroups call: the user could have been removed
+        // from a group between the two calls and we don't want to incorrectly set the length of
+        // the Vec and expose uninitialized memory.
+        unsafe { groups.set_len(s as usize) };
+        groups.iter().cloned().map(|gid| Gid::from_raw(gid)).collect()
```
asomers: Can you avoid the data copy here?  You may have to `mem::transmute` `gid_t`s into `Gid`s.

JayH5:
@asomers I don't know much about the test infrastructure, but `setgroups()` (and by extension, `initgroups()`) require root on most platforms. Is that possible?

I've added a `getgroups()` implementation for now, in case it is possible.

JayH5:
Thank you both for the reviews 😄

src/unistd.rs
JayH5: I can sure give it a go. Like I said with the other case, I'd like to have tests first, if possible.

asomers:
Rust's unit test framework lacks a skip method.  So you should do a UID check.  If non-zero, simply return 0.

src/unistd.rs
```diff
@@ -941,6 +941,76 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+
+    Errno::result(res).map(drop)
+}
+
+/// Initialize the supplementary group access list. Sets the supplementary
+/// group IDs for the calling process using all groups that `user` is a member
+/// of. The additional group `group` is also added to the list.
+///
+/// [Further reading](http://man7.org/linux/man-pages/man3/initgroups.3.html)
+// TODO: Get initgroups binding added for FreeBSD in libc
```
Susurrus: We should have this done before we merge this PR. Can you file a PR against libc adding the necessary FFI?

JayH5: Created rust-lang/libc#726. Also added `getgrouplist()` as it'd help for testing `initgroups()`, at least on some platforms.

src/unistd.rs
```diff
@@ -941,6 +941,94 @@ pub fn setgid(gid: Gid) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Get the list of supplementary group IDs of the calling process.
+///
+/// *Note:* On macOS, `getgroups()` behavior differs somewhat from other Unix
+/// platforms. It returns the current group access list for the user associated
+/// with the effective user id of the process; the group access list may change
+/// over the lifetime of the process, and it is not affected by calls to
+/// `setgroups()`.
```
JayH5: These notes are borrowed from the Python documentation: https://docs.python.org/3.6/library/os.html#os.getgroups

JayH5:
I can also implement `getgrouplist()`. I've done that [here](https://github.com/JayH5/nix/commit/56c0167116449f7cb8533a12fcade7f56208095d), but it is fairly complex and I'm not sure if it's OK to grow this PR any more.

Susurrus:
I'm find with adding `getgrouplist()` into this PR since it's all closely related and the tests are different with it added.

Susurrus:
I think we can also squash this into a single commit, or a commit per function rather than the 14 it currently stands at.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// will only ever return the complete list or else an error.
+///
+/// [Further reading](http://man7.org/linux/man-pages/man3/getgrouplist.3.html)
+pub fn getgrouplist(user: &CString, group: Gid) -> Result<Vec<Gid>> {
+    let ngroups_max = match sysconf(SysconfVar::NGROUPS_MAX) {
+        Ok(Some(n)) => n as c_int,
+        Ok(None) | Err(_) => <c_int>::max_value(),
+    };
+    use std::cmp::min;
+    let mut ngroups = min(ngroups_max, 8);
```
JayH5: 8 is pretty arbitrarily chosen. `NGROUPS_MAX` can vary quite a bit:
* macOS: 16
* Debian (Linux/GNU): 65536
* Alpine (Linux/musl): 32

JayH5:
I've been on leave before this week so I have less time now. My git rebasing skills are... a little rusty. Not sure when I'll get to cleaning up these commits but may only be on the weekend.

The code is all here for review for now at least.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    // We can coerce a pointer to some `Gid`s as a pointer to some `gid_t`s as
+    // they have the same representation in memory.
+    let res = unsafe {
+        libc::setgroups(groups.len() as setgroups_ngroups_t, groups.as_ptr() as *const gid_t)
+    };
+
+    Errno::result(res).map(drop)
+}
+
+/// Calculate the supplementary group access list. Gets the group IDs of all
```
Susurrus: Make the 2nd and third sentence a new paragraph.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// groups that `user` is a member of. The additional group `group` is also
+/// added to the list.
+///
+/// *Note:* Although the `getgrouplist()` call does not return any specific
```
Susurrus: Put this under an `# Errors` section (see [the book](https://doc.rust-lang.org/stable/book/first-edition/documentation.html#special-sections))

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// errors on any known platforms, this implementation will return a system
+/// error of `EINVAL` if the number of groups to be fetched exceeds the
+/// `NGROUPS_MAX` sysconf value. This mimics the behaviour of `getgroups()`
+/// and `setgroups()`. Additionally, while some implementations will return a
+/// partial list of groups when `NGROUPS_MAX` is exceeded, this implementation
+/// will only ever return the complete list or else an error.
+///
+/// [Further reading](http://man7.org/linux/man-pages/man3/getgrouplist.3.html)
```
Susurrus: Put this above the errors section.

src/unistd.rs
Susurrus: What's the point of capping the number of groups?

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+        // BSD systems only return 0 or -1, Linux returns ngroups on success.
+        if ret >= 0 {
+            unsafe { groups.set_len(ngroups as usize) };
+            break
+        }
+
+        // Returns -1 if ngroups is too small, but does not set errno.
+        // BSD systems will still fill the groups buffer with as many groups
+        // as possible, but Linux manpages do not mention this behavior.
+        if ret == -1 {
```
Susurrus: This should be an `else if` to be more readable.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+                               gid as getgrouplist_group_t,
+                               groups.as_mut_ptr() as *mut getgrouplist_group_t,
+                               &mut ngroups)
+        };
+        // BSD systems only return 0 or -1, Linux returns ngroups on success.
+        if ret >= 0 {
+            unsafe { groups.set_len(ngroups as usize) };
+            break
+        }
+
```
Susurrus: Remove this blank line.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+        // Returns -1 if ngroups is too small, but does not set errno.
+        // BSD systems will still fill the groups buffer with as many groups
+        // as possible, but Linux manpages do not mention this behavior.
+        if ret == -1 {
+            let cap = groups.capacity();
+            if cap >= ngroups_max as usize {
+                // We already have the largest capacity we can, give up
+                return Err(Error::invalid_argument())
```
Susurrus: This should have a semicolon at the end.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+        if #[cfg(any(target_os = "ios", target_os = "macos"))] {
+            type initgroups_group_t = c_int;
+        } else {
+            type initgroups_group_t = gid_t;
+        }
+    }
+    let gid: gid_t = group.into();
+    let res = unsafe { libc::initgroups(user.as_ptr(), gid as initgroups_group_t) };
+
+    Errno::result(res).map(drop)
```
Susurrus: Instead of `map(drop` do `map(|_| ())`. It's more semantically clear what it's doing.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    // First get the number of groups so we can size our Vec
+    use std::ptr;
+    let ret = unsafe { libc::getgroups(0, ptr::null_mut()) };
+    let mut size = try!(Errno::result(ret));
+
+    // Now actually get the groups
+    let mut groups = Vec::<Gid>::with_capacity(size as usize);
+    loop {
+        // We can coerce a pointer to some `Gid`s as a pointer to some `gid_t`s
+        // as they have the same representation in memory.
```
Susurrus: This is actually not technically true. newtype's are not guaranteed to have the same memory layout as their containing structs. `repr(transparent)` has been merged as [an RFC](https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md) to rectify this. We've been making this assumption within nix because I think for all platforms we actually target it's true, but we need to be careful with this.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// with the effective user id of the process; the group access list may change
+/// over the lifetime of the process, and it is not affected by calls to
+/// `setgroups()`.
+///
+/// [Further reading](http://pubs.opengroup.org/onlinepubs/009695399/functions/getgroups.html)
+pub fn getgroups() -> Result<Vec<Gid>> {
+    // First get the number of groups so we can size our Vec
+    use std::ptr;
+    let ret = unsafe { libc::getgroups(0, ptr::null_mut()) };
+    let mut size = try!(Errno::result(ret));
```
Susurrus: Use the `?` operator.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    let mut groups = Vec::<Gid>::with_capacity(size as usize);
+    loop {
+        // We can coerce a pointer to some `Gid`s as a pointer to some `gid_t`s
+        // as they have the same representation in memory.
+        let ret = unsafe { libc::getgroups(size, groups.as_mut_ptr() as *mut gid_t) };
+
+        match Errno::result(ret) {
+            Ok(s) => {
+                unsafe { groups.set_len(s as usize) };
+                break
```
Susurrus: It'd be more clear here if you explicitly returned instead of breaking.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// `setgroups()`.
+///
+/// [Further reading](http://pubs.opengroup.org/onlinepubs/009695399/functions/getgroups.html)
+pub fn getgroups() -> Result<Vec<Gid>> {
+    // First get the number of groups so we can size our Vec
+    use std::ptr;
+    let ret = unsafe { libc::getgroups(0, ptr::null_mut()) };
+    let mut size = try!(Errno::result(ret));
+
+    // Now actually get the groups
```
Susurrus: This comment should be expanded to explain why we need to loop here.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+                // resize of the groups Vec and try again...
+                let cap = groups.capacity();
+                unsafe { groups.set_len(cap) };
+                groups.reserve(1);
+                size = groups.capacity() as c_int;
+            },
+            Err(e) => return Err(e)
+        }
+    }
+    Ok(groups)
```
Susurrus: This should be replaced with an `unreachable!`.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+            type setgroups_ngroups_t = size_t;
+        }
+    }
+    // We can coerce a pointer to some `Gid`s as a pointer to some `gid_t`s as
+    // they have the same representation in memory.
+    let res = unsafe {
+        libc::setgroups(groups.len() as setgroups_ngroups_t, groups.as_ptr() as *const gid_t)
+    };
+
+    Errno::result(res).map(drop)
```
Susurrus: `map(|_| ())` as well.

src/unistd.rs
```diff
@@ -941,6 +941,158 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+                     target_os = "macos",
+                     target_os = "netbsd",
+                     target_os = "openbsd"))] {
+            type setgroups_ngroups_t = c_int;
+        } else {
+            type setgroups_ngroups_t = size_t;
+        }
+    }
+    // We can coerce a pointer to some `Gid`s as a pointer to some `gid_t`s as
+    // they have the same representation in memory.
```
Susurrus: Again, not actually true.

test/test_unistd.rs
```diff
@@ -104,6 +104,57 @@ mod linux_android {
     }
 }
 
+#[test]
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
```
Susurrus: Why aren't these run on these targets?

src/unistd.rs
JayH5: Is it safer to do a `mem::transmute`? Difficult to find recent and complete documentation on this stuff.

src/unistd.rs
JayH5: Not sure I understand? I'm choosing 8 as a good size to start the `groups` buffer at. If `groups` is bigger than `NGROUPS_MAX` then the `getgrouplist()` call will error on most platforms.

test/test_unistd.rs
JayH5: Because the `getgroups()` and `setgroups()` calls don't work as expected on Apple platforms, as discussed in the docs.

JayH5:
@Susurrus I guess the usual use for `initgroup` (or at least what I want to use it for) is when switching user to drop privileges. That usually works with `passwd` and `group` stuff that isn't in nix yet (and would probably be my next PR).

Hypothetically speaking, something like:
```rust
let user = "daemon";
let passwd = Passwd::for_name(&user);
initgroups(passwd.name, passwd.gid)?;
setgid(passwd.gid)?;
setuid(passwd.uid)?;
```
Guess I can try figure out an example without the passwd stuff though. Like, I said not a _lot_ of free time right now but should get there...

src/unistd.rs
Susurrus: Yeah, re-read the docs and it makes sense now.

Susurrus:
That looks to be a reasonable example then, let's add it as one.

src/unistd.rs
```diff
@@ -941,11 +941,162 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+            },
+            Err(e) => return Err(e)
+        }
+    }
+}
+
+/// Set the list of supplementary group IDs for the calling process.
+///
+/// *Note:* On macOS, `getgroups()` may not return the same group list set by
+/// calling `setgroups()`. Apple discourages the use of `setgroups()`.
```
Susurrus: What does Apple suggest one use instead? It's frustrating when this is stated and then there's no suggest alternative.

test/test_unistd.rs
```diff
@@ -104,6 +104,57 @@ mod linux_android {
...
+
+    // Set some new made up groups
+    let groups = [Gid::from_raw(123), Gid::from_raw(456)];
+    setgroups(&groups).unwrap();
+
+    let new_groups = getgroups().unwrap();
+    assert_eq!(new_groups, groups);
+
+    // Revert back to the old groups
+    setgroups(&old_groups).unwrap();
```
Susurrus: Shouldn't this have one last assertion following this to check that they were properly reverted?

test/test_unistd.rs
```diff
@@ -104,6 +104,57 @@ mod linux_android {
...
+}
+
+#[test]
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+fn test_initgroups() {
+    if !Uid::current().is_root() {
+        // initgroups(), setgroups() require root
```
Susurrus: Move this comment above the `if` and rephrase to something like "Skip this test when not run as `root` as `initgroups()` and `setgroups()` both require root."

test/test_unistd.rs
```diff
@@ -104,6 +104,57 @@ mod linux_android {
...
+    let group_list = getgrouplist(&user, group).unwrap();
+    assert!(group_list.contains(&group));
+
+    initgroups(&user, group).unwrap();
+
+    let new_groups = getgroups().unwrap();
+    assert_eq!(new_groups, group_list);
+
+    // Revert back to the old groups
+    setgroups(&old_groups).unwrap();
```
Susurrus: Again, shouldn't there be one last assertion following this?

test/test_unistd.rs
Susurrus: Please leave that as a comment in the code here for both this and the following test.

test/test_unistd.rs
JayH5: The `getgroups()` and `setgroups()` calls at the start and end of these tests aren't really part of the tests, they're just a small attempt to restore the state of the current process to what it was before the test ran. 😕

Susurrus:
Alright, can you squash everything into a single commit and add a CHANGELOG entry for this?

@asomers Can you approve your review if everything's been addressed?

Once that's all done we should be good to merge!

JayH5:
@Susurrus I've rebased to fewer, clearer commits. If you want just one commit then GitHub can do that when we merge?

Thanks for the reviewing 😄

Susurrus:
The number of commits are fine now, you didn't need to segregate them all, but it's fine.

You have failures in CI that need resolving before we can merge.

Also, @asomers can you resolve your review here? I want to make sure everything's fine with you (please click the "Approve review" or whatever link down towards the bottom of the page).

JayH5:
Looks like rust-lang/libc#742 broke `initgroups` on Linux platforms. I can try fix, but may only get to it next weekend.

test/test_unistd.rs
```diff
@@ -104,6 +104,60 @@ mod linux_android {
     }
 }
 
+#[test]
+// `getgroups()` and `setgroups()` do not behave as expected on Apple platforms
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+fn test_setgroups() {
+    // Skip this test when not run as root as `setgroups()` requires root.
+    if !Uid::current().is_root() {
+        return
```
asomers: Until Cargo adds a way to skip tests, you should print something here to let the user know that the test was skipped.  Otherwise it will rot, because people won't realize that the test isn't run.  See tests/test_mount.rs.  Ditto for line 137.

test/test_unistd.rs
```diff
@@ -104,6 +104,60 @@ mod linux_android {
     }
 }
 
+#[test]
+// `getgroups()` and `setgroups()` do not behave as expected on Apple platforms
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+fn test_setgroups() {
```
asomers: These two tests race against each other.  You must add a mutex to prevent that.  grep for `CWD_MTX` for an example.

JayH5:
I've (hopefully) fixed the libc `initgroups` call on Linux, rebased, and made the test changes @asomers requested.

JayH5:
@asomers @Susurrus ping on this.

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// *Note:* On macOS, `getgroups()` behavior differs somewhat from other Unix
+/// platforms. It returns the current group access list for the user associated
+/// with the effective user id of the process; the group access list may change
+/// over the lifetime of the process, and it is not affected by calls to
+/// `setgroups()`.
+///
+/// [Further reading](http://pubs.opengroup.org/onlinepubs/009695399/functions/getgroups.html)
+pub fn getgroups() -> Result<Vec<Gid>> {
+    // First get the number of groups so we can size our Vec
+    use std::ptr;
```
Susurrus: Remove this and just use the full path below. Since this import is only used once it's a little cleaner that way.

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    let ret = unsafe { libc::getgroups(0, ptr::null_mut()) };
+    let mut size = Errno::result(ret)?;
+
+    // Now actually get the groups. We try multiple times in case the number of
+    // groups has changed since the first call to getgroups() and the buffer is
+    // now too small
```
Susurrus: Missing "."

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+            Ok(s) => {
+                unsafe { groups.set_len(s as usize) };
+                return Ok(groups);
+            },
+            Err(Error::Sys(Errno::EINVAL)) => {
+                // EINVAL indicates that size was too small, so trigger a
+                // resize of the groups Vec and try again...
+                let cap = groups.capacity();
+                unsafe { groups.set_len(cap) };
+                groups.reserve(1);
```
Susurrus: I would suggest we double the size of the array rather than keep increasing it by 1. Are there any heuristics we might be able to use here?

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+
+    // Now actually get the groups. We try multiple times in case the number of
+    // groups has changed since the first call to getgroups() and the buffer is
+    // now too small
+    let mut groups = Vec::<Gid>::with_capacity(size as usize);
+    loop {
+        // FIXME: On the platforms we currently support, the `Gid` struct has
+        // the same representation in memory as a bare `gid_t`. This is not
+        // necessarily the case on all Rust platforms, though. See RFC 1785.
+        let ret = unsafe { libc::getgroups(size, groups.as_mut_ptr() as *mut gid_t) };
```
Susurrus: I'd rather we lose the `size` variable and instead use `groups.capacity()` (I think that's what we'd want) everywhere. This way we don't have temp variables that are being modified through every loop. Would that work?

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+            Err(e) => return Err(e)
+        }
+    }
+}
+
+/// Set the list of supplementary group IDs for the calling process.
+///
+/// *Note:* On macOS, `getgroups()` may not return the same group list set by
+/// calling `setgroups()`. The use of `setgroups()` on macOS is 'highly
+/// discouraged' by Apple. Developers are referred to the `opendirectoryd`
```
Susurrus: Can you link to these sets of APIs?

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// daemon and its set of APIs.
+///
+/// [Further reading](http://man7.org/linux/man-pages/man2/getgroups.2.html)
+///
+/// # Examples
+///
+/// `setgroups` can be used when dropping privileges from the root user to a
+/// specific user and group. For example, given the user `www-data` with UID
+/// `33` and the group `backup` with the GID `34`, one could switch user as
```
Susurrus: "switch the user"

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    // same representation in memory as a bare `gid_t`. This is not necessarily
+    // the case on all Rust platforms, though. See RFC 1785.
+    let res = unsafe {
+        libc::setgroups(groups.len() as setgroups_ngroups_t, groups.as_ptr() as *const gid_t)
+    };
+
+    Errno::result(res).map(|_| ())
+}
+
+/// Calculate the supplementary group access list. Gets the group IDs of all
```
Susurrus: There should be a blank line between the short summary and any extra details. The short summary is a short line of text. Please see the Rustdoc documentation conventions for details here.

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    }
+    let gid: gid_t = group.into();
+    loop {
+        let ret = unsafe {
+            libc::getgrouplist(user.as_ptr(),
+                               gid as getgrouplist_group_t,
+                               groups.as_mut_ptr() as *mut getgrouplist_group_t,
+                               &mut ngroups)
+        };
+        // BSD systems only return 0 or -1, Linux returns ngroups on success.
```
Susurrus: Put a space above this comment to more clearly separate it from the above. Whitespace helps the readability here.

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+            groups.reserve(ngroups as usize);
+
+            // Even if the buffer gets resized to bigger than ngroups_max,
+            // don't ever ask for more than ngroups_max groups
+            ngroups = min(ngroups_max, groups.capacity() as c_int);
+        }
+    }
+}
+
+/// Initialize the supplementary group access list. Sets the supplementary
```
Susurrus: Again, blank comment line after the first sentence. Run rustdoc on this to see how it affects the output.

src/unistd.rs
```diff
@@ -982,6 +982,193 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// respectively, one could switch user as follows:
+/// ```
+/// let user = CString::new("www-data").unwrap();
+/// let uid = Uid::from_raw(33);
+/// let gid = Gid::from_raw(33);
+/// initgroups(&user, gid)?;
+/// setgid(gid)?;
+/// setuid(uid)?;
+/// ```
+pub fn initgroups(user: &CString, group: Gid) -> Result<()> {
```
Susurrus: We've had discussions for APIs where we'd like to use the regular Rust `String` type rather than `CString` on these APIs. I prefer it, but understand there's a performance penalty for doing that. You or @asomers do either of you have an opinion? Right now we kind of have a mix within `nix`.

test/test_unistd.rs
```diff
@@ -104,6 +104,74 @@ mod linux_android {
...
+    // Skip this test when not run as root as `setgroups()` requires root.
+    if !Uid::current().is_root() {
+        use std::io;
+        let stderr = io::stderr();
+        let mut handle = stderr.lock();
+        writeln!(handle, "test_setgroups requires root privileges. Skipping test.").unwrap();
+        return
```
Susurrus: This looks odd without the semi-colon. Please add one.

test/test_unistd.rs
```diff
@@ -104,6 +104,74 @@ mod linux_android {
...
+
+    let new_groups = getgroups().unwrap();
+    assert_eq!(new_groups, groups);
+
+    // Revert back to the old groups
+    setgroups(&old_groups).unwrap();
+}
+
+#[test]
+// `getgroups()` and `setgroups()` do not behave as expected on Apple platforms
```
Susurrus: I'd think it'd still be good to test the mac/ios functionality if we could. Do you understand how they work enough to write some tests for them, even if they don't cover all functionality? I don't like that we document how Apple's platforms are different and then just don't test them even though it's clear we should know how they do work well enough to test them.

test/test_unistd.rs
```diff
@@ -104,6 +104,74 @@ mod linux_android {
...
+    #[allow(unused_variables)]
+    let m = ::GROUPS_MTX.lock().expect("Mutex got poisoned by another test");
+
+    // Save the existing groups
+    let old_groups = getgroups().unwrap();
+
+    // It doesn't matter if the root user is not called "root" or if a user
+    // called "root" doesn't exist. We are just checking that the extra,
+    // made-up group, `123`, is set.
+    // FIXME: This only tests half of initgroups' functionality.
```
Susurrus: What's the other half and is it easy to test?

src/unistd.rs
asomers: I don't have a strong opinion.  But I would weakly suggesting sticking with `CString`, because it will entail less conversion work at runtime.  And the correct type is `&CStr`, not `&CString`.  `CString` should only be used when the value must be owned.

src/unistd.rs
JayH5: This should be triggering the internal resizing mechanism of `Vec`. Calling `.reserve(1)` on a full `Vec` should reserve at a _minimum_ 1 extra space, but probably more (don't know what `Vec` does but probably grows exponentially?).

I copied this from elsewhere in nix. Perhaps I need a better comment?

src/unistd.rs
JayH5: Google can't seem to give me anything more than raw manpage source.. like [here](https://opensource.apple.com/source/Libinfo/Libinfo-330.3/membership.subproj/mbr_uid_to_uuid.3.auto.html). Not sure what the canonical online source of documentation for macOS is? Seems hard to find up-to-date documentation.

test/test_unistd.rs
JayH5: The half being tested is that `initgroups()` sets the made-up group `123`. Whether it fetches & sets the groups that the user is a member of is untested, because to do that we'd have to control which groups a user is a part of...which I don't actually know how to do from a libc API 🤔

Susurrus: Please document in that FIXME this.

asomers:
Everything looks fine to me, except for the final squash.  @Susurrus do you approve too?

Susurrus:
I have a few comments I made a while ago that didn't get a response and should get one before I'm okay with this PR. I don't have any other comments after those are all resolved.

Susurrus:
@JayH5 Pinging you on this as we'd love to get it merged.

src/unistd.rs
Susurrus: Yes, a comment specifying exactly that would be awesome.

JayH5:
@Susurrus sorry for the slow reply, I was on holiday.

I think I've responded to all your comments now, except:
> I'd think it'd still be good to test the mac/ios functionality if we could. Do you understand how they work enough to write some tests for them, even if they don't cover all functionality? I don't like that we document how Apple's platforms are different and then just don't test them even though it's clear we should know how they do work well enough to test them.

I'm honestly not sure what to do about this. I tried adding a test for Apple platforms based on the following note in the `getgroups` manpage:
> Calling initgroups(3) to opt-in for supplementary groups will cause getgroups() to return a single entry, the GID that was passed to initgroups(3)

But I don't actually see that behaviour on my machine running High Sierra (`getgroups` returns more than one entry after a call to `initgroups`).

So, honestly, no, I don't know enough about the way these calls work on Apple platforms to write tests for them. I could experiment and test what the behaviour is but I'm not sure I have the time/patience to do that. Perhaps these functions could just be disabled on Apple platforms?

Susurrus:
And I agree with your final assessment above, that this should be disabled on Apple platforms since it's unclear what the expected behavior should be. Let's just make it clear in the documentation why this is disabled on Apple targets.

JayH5:
>  [...]this should be disabled on Apple platforms since it's unclear what the expected behavior should be. Let's just make it clear in the documentation why this is disabled on Apple targets.

@Susurrus I've tried to do this now. Not sure if these functions should go inside a submodule, like with [`unistd::setres`](https://github.com/nix-rust/nix/blob/08c41c853b092f20cad6f3a3ff9153b3b34564f2/src/unistd.rs#L1740-L1778)?

Susurrus:
@JayH5 There's no need to put them in a submodule (I'd actually suggest we take those functions out of a submodule). But this LGTM. Thanks for your PR!

bors r+

Issue #669, opened 2017-07-11
TravisCI is too slow


asomers:
We have 34 builds in Travis, and it takes over an hour for all of them to finish.  It's causing Bors to timeout.  At a minimum, we need to increase bors's timeout.  Other ways to speed up the builds include:

1. Move i686-unknown-freebsd into BuildBot
2. remove `sudo: required`.  It causes Travis to use a full VM instead of a container, and I don't think we need it.
3. Only test beta and nightly builds on one architecture.
4. Enable parallel tests.  This will likely have minimal benefit, because setting up the environment is slower than running tests.
5. Use a custom Docker image so we won't have to download the QEMU stuff all the time
6. Set Travis's Auto Cancel Pull Requests feature

Any other ideas?

Susurrus:
I agree with 1, 2, 3, and 6. For 3, I'd suggest we just drop nightly, as there's little benefit to us testing on it. I like that these are all pretty low-hanging fruit. We can deal with a custom docker image later if this continues to be problematic.

Susurrus:
I've already handled 6 and turned on auto cancel for pull requests. I left it on for branches as it isn't necessary there.

Susurrus:
With #674 closed, we've knocked out 2 & 3. What do you say @asomers about doing 1? I think it'd be nice to get Buildbot doing a bit more since it runs so quickly!

asomers:
FreeBSD 11.1 is scheduled to be released on July 26.  I'm going to wait until then so I can kill 2 birds with one stone.

Susurrus:
Sounds good. Also, do you want to turn this issue's numbered list into checkboxes so we can check off ones that are done?

Susurrus:
BTW: Good work on #674. An 18% reduction in CI build times is nothing to sneeze at!

asomers:
Progress report:

- [ ] Move i686-unknown-freebsd into BuildBot (waiting for 11.1-RELEASE)
- [x] remove sudo: required. It causes Travis to use a full VM instead of a container, and I don't think we need it.
- [x] Only test beta and nightly builds on one architecture.
- [ ] Enable parallel tests. This will likely have minimal benefit, because setting up the environment is slower than running tests.
- [ ] Use a custom Docker image so we won't have to download the QEMU stuff all the time
- [x] Set Travis's Auto Cancel Pull Requests feature

Susurrus:
[...]

PR #738, opened 2017-08-17
Reorder builds to improve build times

Susurrus:
The main issue is that all iOS builds are in a block and there are only 5
simultaneous builds allowed by our Travis CI plan. So if there's a backlog
of iOS builders, then the entire build process stalls. So instead of having
them all clumped together, distribute the Mac and iOS builds throughout the
build pipeline.

Addresses part of #669.


Susurrus:
This is a pretty straight-forward change. Let's do it and see how PRs go.

bors r+

PR #739, opened 2017-08-17
Support features across more platforms

### LINKED ISSUE #607: Enable utsname on MacOS
I'm doing a really quick experiment to try to invoke the `uname` function on macOS. Following [the documentation](https://docs.rs/nix/0.8.1/nix/sys/utsname/fn.uname.html), I wrote the code below.

```rust
extern crate nix;

fn main() {
    println!("Hello, world!");
    let name = nix::sys::utsname::uname();
    println!("
      sysname = {}
      nodename = {}
      release = {}
      version = {}
      machine = {}
    ", name.sysname(), name.nodename(), name.release(), name.version(), name.machine());
}
```

My `Cargo.toml` file points to version 0.8.1 of `nix` as a dependency.  When I try to compile this program with `cargo build` I get the following error:

```
error[E0433]: failed to resolve. Could not find `utsname` in `sys`
 --> src/main.rs:5:16
  |
5 |     let name = nix::sys::utsname::uname();
  |                ^^^^^^^^^^^^^^^^^^^^^^^^ Could not find `utsname` in `sys`
```

I'm running this on macOS 10.12.4 with rustc 1.17.0.

Susurrus:
* Exposes `utsname` on all platforms
 * Expose `signalfd` on Android

Closes #607.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#722](https://github.com/nix-rust/nix/pull/722))
 - Added `nix::unistd:fexecve`.
   ([#727](https://github.com/nix-rust/nix/pull/727))
+- Expose `uname()` on all platforms.
+  ([#739](https://github.com/nix-rust/nix/pull/739))
+- Expose `signalfd` module on Android as well.
+  ([#739](https://github.com/nix-rust/nix/pull/739))
 
 ### Changed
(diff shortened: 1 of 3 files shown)


asomers:
sendfile has a different signature on FreeBSD than on Linux, so the build is failing.

Susurrus:
@asomers Yep, I should've labeled this as WIP as I just wanted to run these through CI to see where things would break. I'll reduce this PR to things that require almost no changes and then file issues for the rest so we can track them. This should end up being a rather small PR at the end of it.

PR #742, opened 2017-08-23
Provide and emulate ptsname_r on macOS and iOS by using the TIOCPTYGNAME syscall.

nelsonjchen:
This provides a thread-safe alternative to ptsname() and compatibility for existing projects using `ptsname_r()` on macOS.

So, I'm not sure if this belongs in `nix`. Would this fall under the "but to unify what can" aspect?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#739](https://github.com/nix-rust/nix/pull/739))
 - Expose `signalfd` module on Android as well.
   ([#739](https://github.com/nix-rust/nix/pull/739))
+- Added the `ptsname_r` function for macOS and iOS that emulates Linux's `ptsname_r` behavior.
+  ([#742](https://github.com/nix-rust/nix/pull/742))
 
 ### Changed
 - Renamed existing `ptrace` wrappers to encourage namespacing ([#692](https://github.com/nix-rust/nix/pull/692))
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -169,6 +169,41 @@ pub fn ptsname_r(fd: &PtyMaster) -> Result<String> {
     Ok(name)
 }
 
+/// (emulated on macOS) Get the name of the slave pseudoterminal (see
+/// [ptsname(3)](http://man7.org/linux/man-pages/man3/ptsname.3.html))
+///
+/// returns the name of the slave pseudoterminal device corresponding to the master
+/// referred to by `fd`. This is the threadsafe version of `ptsname()`, but it is not part of the
+/// POSIX standard and is instead a Linux-specific extension.
+///
+/// This value is useful for opening the slave ptty once the master has already been opened with
+/// `posix_openpt()`.
+///
+/// As `ptsname_r()` is Linux-specific, this implementation emulates `ptsname_r()` through
+/// the `TIOCPTYGNAME` syscall on macOS.
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+#[inline]
+pub fn ptsname_r(fd: &PtyMaster) -> Result<String> {
+    // This is based on
+    // https://blog.tarq.io/ptsname-on-osx-with-rust/
+    // and its derivative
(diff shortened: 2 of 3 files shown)


Susurrus:
@nelsonjchen This isn't something that's currently done in `nix` as far as I know, emulating functionality, and I'm not certain where I stand on this quite yet. But one thing that would help would be hearing what your specific use case is for this? Given that this functionality isn't provided by mac directly, and that mac also supports `posix_openpt` I'm not certain how much of a use there is for this and whether `nix` should bear the support costs of providing it.

nelsonjchen:
@Susurrus

I'm not sure how the availability of `posix_openpt` support on the Mac relates here. It's great it is and I'm using it but I think the addition of a thread-safe `ptsname` or just a straight `ptsname_r` that's thread safe would be a great help to anyone wanting to doing work with master and slave terminals in macOS and to keep the `nix` API feeling Rusty.

My use case is that I would like to write a Rust application that simulates a device on the other end of a serial port in macOS and Linux. If I want to use the "UI" of this simulated device, I would need to know what the corresponding slave terminal device is and would need some form of `ptsname` functionality to be able to point whatever at the slave terminal device aka. emulated serial port. I would like this application to have minimal to no change between the two OSes. I greatly prefer to use thread-safe APIs wherever possible as surprising things ~~can~~ will happen when a non-thread-safe API is used if I was offered a choice. On macOS, I can only use `ptsname` from `nix` which is not thread safe and not `ptsname_r` which is thread safe.

With `ptsname`, I guess I could implement my own static mutex lock around `ptsname` and copy the buffer to another `String` while that lock is happening to get past that issue. But, should there even be these kinds of concerns or workarounds about memory access when using Rust outside of `unsafe`? Runtime error and panic stuff with `Rc`'s and `Cell`s are one thing but memory concurrency issues are another.  The Rust developers and community go on and on about "fearless concurrency" and using `ptsname` as-is in `nix` sure makes me fear even though it's outside of `unsafe`. The `String` underneath can change on you! Mentioning that it's not thread safe in the comment doesn't feel like a very Rusty API or approach and I think `nix` falls a tad bit short there. Use of `ptsname` in `nix` just doesn't feel safe nor Rusty to me.

I too have some doubts about this PR and I couldn't find any precedent so I figure it might be good to just use this issue to set some precedent for `nix` and to refer others to this approach or a possible externalized functionality filler crate if it doesn't get in.

I think I'm at least the third person using Rust to want a simple `ptsname_r`-like functionality on macOS that behaves much like the one on Linux.

As cited in the code, this is derived from two other examples of this kind of code out there:

* This [guy](https://blog.tarq.io/ptsname-on-osx-with-rust/), who came up with most of this approach, was using it for what I'm assuming is something work related at some large hosting provider. I'm not sure what purpose he was exactly using it for but he had a need for `ptsname_r` functionality.  He's not using `nix`, but I can see that he `cfg` gates his similar `nix`-like wrappers to the OS. I'm guessing that he also wants to maintain a similar API between Linux and macOS. This version was the basis for the following reference.

* My other [reference](https://github.com/philippkeller/rexpect/blob/a71dd02/src/process.rs#L67) is doing a pexpect clone. To do a pexpect clone, he has to open the slave terminal by its name which he had to find through `ptsname_r` and then assign them to the correct descriptors before calling exec in the child. Unlike the first reference, he uses `nix`, but polyfills in pretty much this implementation of `ptsname_r` for macOS where `nix`/macOS falls short. Like me, and the previous example, he also fills in the gap with a similar API for macOS that emulates the Linux version.

asomers:
My feeling is that this wrapper is outside of the scope of nix.  However, you raise an interesting point.  Since mutable global variables are always considered `unsafe` by Rust, nix's `ptsname` function should also be `unsafe`.  We can't simply add a mutex in nix's `ptsname` wrapper, because a Rust program could access both `nix::ptsname` and `libc::ptsname` directly.

Susurrus:
I'm also in agreement that `ptsname_r` shouldn't be emulated by `nix`. This is in line with our stance that syscalls shouldn't be supported by `nix` directly. I'm closing in this in favor of an external library covering this use case.

nelsonjchen:
I've since moved this into [its own external library](https://crates.io/crates/nix-ptsname_r-shim).

PR #745, opened 2017-08-25
Use upstream libc definitions for fcntl FFI

Susurrus:


Susurrus:
The modification to the test is because `O_LARGEFILES` is 0 on some systems, which breaks `from_bits`. Instead we use `from_bits_truncate` to properly ignore that bitfields on the necessary systems. I think this should be good to merge.

src/unistd.rs
```diff
@@ -2,7 +2,7 @@
 
 use errno;
 use {Errno, Error, Result, NixPath};
-use fcntl::{fcntl, OFlag, O_CLOEXEC, FD_CLOEXEC};
```
asomers: Why move these use statements?

Susurrus:
@asomers Fixed.

PR #746, opened 2017-08-26
Replace socket FFI with libc versions

Susurrus:
Getting this up there to get some CI run on it. Replacing the `msghdr`, `cmsghdr`, and other FFI function declarations will be a little more involved so I wanted to run this through first.


Susurrus:
There's more work to do here, but I don't know when I'll be able to get to it, so let's start with this and get it merged.

PR #747, opened 2017-08-26
Use libc types for sched and syscall FFI

Susurrus:


Susurrus:
This also includes removing the `src/sys/syscall.rs` module completely. This may be controversial, but a) the types were already all exposed by libc and b) the two syscalls exposed here already have safe wrapper functions exposed in nix.

Susurrus:
Blocking on rust-lang/libc#743.

PR #762, opened 2017-09-04
test/test_stat.rs: test correct stat function

Mic92:


Susurrus:
@mic92 good catch! This LGTM, but I wanted to ask if you could fix the temp folder name for the `test_fstatat` function to be different from the above test as well. Looks like a copy/paste error when it was added. Once that's done here, I'm r+.

Mic92:
Changed via the github web editor (which is also writeable as reviewer and create to fix unrelated nitpicks).

Susurrus:
@Mic92 Thanks for adding that.

Alright, this is minor and looks fine to me.

bors r+

asomers:
Again, the timeout was due to the submitter's umlaut.  I'll manually merge it.

Mic92:
python and unicode..

PR #764, opened 2017-09-09
Add getifaddrs/ifaddrs wrapper

LeoTindall:
This code is from my `getaddrs` crate, which I think fits quite well into this crate. However, I am aware that there are a number of issues with integrating this code with `nix`, and would really appreciate some help dealing with those issues.

- [ ] Error handling is basically nonexistent
- [ ] I may not be doing memory management right? This is my first time writing any significant unsafe Rust.
- [ ] This may be a more heavyweight wrapper than other `nix`-provided wrappers
- [ ] Not properly feature gated, causing build failures (I think? OTOH all the other PRs are failing)

Many thanks for assistance.

### DIFF
--- a/src/net/ifaddrs/iff_flags.rs
+++ b/src/net/ifaddrs/iff_flags.rs
@@ -48,4 +48,3 @@ bitflags! {
         const IFF_ECHO = 1<<18;
     }
 }
-
--- a/src/net/ifaddrs/iff_flags.rs
+++ b/src/net/ifaddrs/iff_flags.rs
@@ -1,50 +1,50 @@
 use libc;
 use libc::c_uint;
 
-bitflags! {
+libc_bitflags! {
     pub struct IffFlags: c_uint {
         /// Interface is running.
-        const IFF_UP = libc::IFF_UP as c_uint;
+        IFF_UP as c_uint;
         /// Valid broadcast address set.
-        const IFF_BROADCAST = libc::IFF_BROADCAST as c_uint;
+        IFF_BROADCAST as c_uint;
         /// Internal debugging flag.
-        const IFF_DEBUG = libc::IFF_DEBUG as c_uint;
+        IFF_DEBUG as c_uint;
         /// Interface is a loopback interface.
-        const IFF_LOOPBACK = libc::IFF_LOOPBACK as c_uint;
+        IFF_LOOPBACK as c_uint;
         /// Interface is a point-to-point link.
-        const IFF_POINTOPOINT = libc::IFF_POINTOPOINT as c_uint;
+        IFF_POINTOPOINT as c_uint;
         /// Resources allocated.
-        const IFF_NOTRAILERS = libc::IFF_NOTRAILERS as c_uint;
+        IFF_NOTRAILERS as c_uint;
         /// No arp protocol, L2 destination address not set.
-        const IFF_RUNNING = libc::IFF_RUNNING as c_uint;
+        IFF_RUNNING as c_uint;
         /// Interface is in promiscuous mode.
-        const IFF_NOARP = libc::IFF_NOARP as c_uint;
+        IFF_NOARP as c_uint;
         /// Avoid use of trailers.
-        const IFF_PROMISC = libc::IFF_PROMISC as c_uint;
+        IFF_PROMISC as c_uint;
         /// Receive all multicast packets.
-        const IFF_ALLMULTI = libc::IFF_ALLMULTI as c_uint;
+        IFF_ALLMULTI as c_uint;
         /// Master of a load balancing bundle.
-        const IFF_MASTER = libc::IFF_MASTER as c_uint;
+        IFF_MASTER as c_uint;
         /// Slave of a load balancing bundle.
-        const IFF_SLAVE = libc::IFF_SLAVE as c_uint;
+        IFF_SLAVE as c_uint;
         /// Supports multicast
-        const IFF_MULTICAST = libc::IFF_MULTICAST as c_uint;
+        IFF_MULTICAST as c_uint;
         /// Is able to select media type via ifmap.
-        const IFF_PORTSEL = libc::IFF_PORTSEL as c_uint;
+        IFF_PORTSEL as c_uint;
         /// Auto media selection active.
-        const IFF_AUTOMEDIA = libc::IFF_AUTOMEDIA as c_uint;
+        IFF_AUTOMEDIA as c_uint;
         /// The addresses are lost when the interface goes down.
-        const IFF_DYNAMIC = libc::IFF_DYNAMIC as c_uint;
+        IFF_DYNAMIC as c_uint;
 
         // These flags are available on modern Linuxes
         #[cfg(any(target_os = "linux", target_os = "android"))]
         /// Driver signals L1 up (since Linux 2.6.17)
-        const IFF_LOWER_UP = 1<<16;
+        IFF_LOWER_UP as c_uint;
         #[cfg(any(target_os = "linux", target_os = "android"))]
         /// Driver signals dormant (since Linux 2.6.17)
-        const IFF_DORMANT = 1<<17;
+        IFF_DORMANT as c_uint;
         #[cfg(any(target_os = "linux", target_os = "android"))]
         /// Echo sent packets (since Linux 2.6.25)
-        const IFF_ECHO = 1<<18;
+        IFF_ECHO as c_uint;
     }
 }
--- a/src/net/ifaddrs/iff_flags.rs
+++ b/src/net/ifaddrs/iff_flags.rs
@@ -2,7 +2,7 @@ use libc;
 use libc::c_uint;
 
 bitflags! {
-    pub struct InterfaceFlags: c_uint {
+    pub struct IffFlags: c_uint {
         /// Interface is running.
         const IFF_UP = libc::IFF_UP as c_uint;
         /// Valid broadcast address set.
@@ -35,10 +35,15 @@ bitflags! {
         const IFF_AUTOMEDIA = libc::IFF_AUTOMEDIA as c_uint;
         /// The addresses are lost when the interface goes down.
         const IFF_DYNAMIC = libc::IFF_DYNAMIC as c_uint;
+
+        // These flags are available on modern Linuxes
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         /// Driver signals L1 up (since Linux 2.6.17)
         const IFF_LOWER_UP = 1<<16;
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         /// Driver signals dormant (since Linux 2.6.17)
         const IFF_DORMANT = 1<<17;
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         /// Echo sent packets (since Linux 2.6.25)
         const IFF_ECHO = 1<<18;
     }
--- a/src/net/ifaddrs/mod.rs
+++ b/src/net/ifaddrs/mod.rs
@@ -63,7 +63,6 @@
 //!
 
 use libc;
-use std::net::IpAddr;
 use std::ptr::null_mut;
 use std::ffi::CStr;
 use std::collections::HashMap;
@@ -75,7 +74,7 @@ pub mod iff_flags;
 use self::iff_flags::IffFlags;
 
 mod sockaddr;
-use self::sockaddr::sockaddr_to_ipaddr;
+use self::sockaddr::{IfAddrValue, sockaddr_to_ifaddrvalue};
 
 pub type InterfaceMap = HashMap<String, Vec<InterfaceAddr>>;
 
@@ -149,12 +148,12 @@ pub struct InterfaceAddr {
     /// The address assigned to the interface for this protocol.
     /// A value of `None` means the libc reported a type of address that
     /// `std::net` doesn't understand.
-    pub address: Option<IpAddr>,
+    pub address: Option<IfAddrValue>,
 
     /// The netmasks assigned to the interface for this protocol.
     /// A value of `None` means the libc reported a type of address that
     /// `std::net` doesn't understand.
-    pub netmask: Option<IpAddr>,
+    pub netmask: Option<IfAddrValue>,
 
     /// The ifu assigned to the interface for this protocol.
     /// A value of `{Broadcast, Destination}Addr(None)` means the libc reported
@@ -171,8 +170,8 @@ pub struct InterfaceAddr {
 /// point-to-point destination address.
 #[derive(Debug, Clone)]
 pub enum InterfaceIfu {
-    BroadcastAddr(Option<IpAddr>),
-    DestinationAddr(Option<IpAddr>),
+    BroadcastAddr(Option<IfAddrValue>),
+    DestinationAddr(Option<IfAddrValue>),
     Neither,
 }
 
@@ -208,22 +207,22 @@ impl Iterator for InterfaceAddrs {
         // Interpret the flags field into a typed version of those flags
         let flags = IffFlags::from_bits_truncate(p.ifa_flags);
 
-        // Get std::net::IpAddr representations of the address and netmask
-        // UNSAFETY: sockaddr_to_ipaddr requires valid pointer.
-        let address = unsafe { sockaddr_to_ipaddr(p.ifa_addr) };
-        // UNSAFETY: sockaddr_to_ipaddr requires valid pointer.
-        let netmask = unsafe { sockaddr_to_ipaddr(p.ifa_netmask) };
+        // Get IfAddrValue representations of the address and netmask
+        // UNSAFETY: sockaddr_to_ifaddrvalue requires valid pointer.
+        let address = unsafe { sockaddr_to_ifaddrvalue(p.ifa_addr) };
+        // UNSAFETY: sockaddr_to_ifaddrvalue requires valid pointer.
+        let netmask = unsafe { sockaddr_to_ifaddrvalue(p.ifa_netmask) };
 
         // Figure out which ifu type is needed and create it
         let ifu = if flags.contains(iff_flags::IFF_POINTOPOINT) {
             // Point to point destination address
-            // UNSAFETY: sockaddr_to_ipaddr requires valid pointer.
-            let ifu_addr = unsafe { sockaddr_to_ipaddr(p.ifa_ifu) };
+            // UNSAFETY: sockaddr_to_ifaddrvalue requires valid pointer.
+            let ifu_addr = unsafe { sockaddr_to_ifaddrvalue(p.ifa_ifu) };
             InterfaceIfu::DestinationAddr(ifu_addr)
         } else if flags.contains(iff_flags::IFF_BROADCAST) {
             // Broadcast address
-            // UNSAFETY: sockaddr_to_ipaddr requires valid pointer.
-            let ifu_addr = unsafe { sockaddr_to_ipaddr(p.ifa_ifu) };
+            // UNSAFETY: sockaddr_to_ifaddrvalue requires valid pointer.
+            let ifu_addr = unsafe { sockaddr_to_ifaddrvalue(p.ifa_ifu) };
             InterfaceIfu::BroadcastAddr(ifu_addr)
         } else {
             InterfaceIfu::Neither
--- a/src/net/ifaddrs/mod.rs
+++ b/src/net/ifaddrs/mod.rs
@@ -102,7 +102,7 @@ impl<'a> InterfaceAddrs<'a> {
     ///
     /// # Errors
     /// `Err(())` will be returned if `p` was void.
-    pub unsafe fn from_raw(p: *mut libc::ifaddrs) -> ::std::result::Result<Self, ()> {
+    pub unsafe fn from_raw(p: *mut libc::ifaddrs) -> ::std::result::Result<InterfaceAddrs<'a>, ()> {
         match p.as_ref() {
             Some(r) => Ok(Self {
                 inner: p,
--- a/src/net/ifaddrs/mod.rs
+++ b/src/net/ifaddrs/mod.rs
@@ -1,0 +1,250 @@
+//! `ifaddrs` provides a safe interface for the system's network interface data.
+//!
+//! The `InterfaceAddrs` struct provides access to the system's network 
+//! interface data. You can either iterate over it or consume it and convert 
+//! it into an `InterfaceMap` (a `HashMap<String, Vec<InterfaceAddr>>`) for 
+//! more convenient access by interface name.
+//!
+//! # Examples
+//!
+//! You can access the basic information of the system in a few lines.
+//! The following program prints all the known addresses.
+//!
+//! ```
+//! use nix::net::ifaddrs::InterfaceAddrs;
+//!
+//! let addrs = InterfaceAddrs::query_system()
+//!     .expect("System has no network interfaces.");
+//!
+//! for addr in addrs {
+//!     println!("{}: {:?}", addr.name, addr.address);
+//! }
+//! ```
+//!
+//! The `InterfaceFlags` struct provides access to info about the 
+//! state of an interface. This program prints the addresses of only
+//! interfaces which are up.
+//!
+//! ```
+//! use nix::net::ifaddrs::{InterfaceAddrs, if_flags};
+//!
+//! let addrs = InterfaceAddrs::query_system()
+//!     .expect("System has no network interfaces.");
+//!
+//! for addr in addrs {
+//!     if addr.flags.contains(if_flags::IFF_UP) {
+//!         println!("{}: {:?}", addr.name, addr.address);
+//!     }
+//! }
+//! ```
+//!
+//! You can convert the `InterfaceAddrs` struct into a `HashMap` easily.
+//! `InterfaceMap` is an alias for `HashMap<String, Vec<InterfaceAddr>>` for
+//! easier reference.
+//!
+//! ```
+//! use nix::net::ifaddrs::{InterfaceAddrs, InterfaceAddr, InterfaceMap};
+//! use std::collections::HashMap;
+//!
+//! let interfaces: InterfaceMap = 
+//!     InterfaceAddrs::query_system()
+//!     .expect("System has no network interfaces.")
+//!     .into(); // Convert to a hash map
+//!
+//! // Print all the addresses of the loopback interface
+//! if let Some(addrs) = interfaces.get("lo") {
+//!    println!("Loopback addresses:");
+//!    for addr in addrs {
+//!         println!("\t{:?}", addr);
+//!    }
+//! }
+//!     
+//! ```
+//!
+
+use libc;
+use std::net::{IpAddr};
+use std::ptr::null_mut;
+use std::ffi::CStr;
+use std::collections::HashMap;
+
+pub mod if_flags;
+use self::if_flags::InterfaceFlags;
+
+mod sockaddr;
+use self::sockaddr::sockaddr_to_ipaddr;
+
+pub type InterfaceMap = HashMap<String, Vec<InterfaceAddr>>;
+
+/// Represents a handle into the operating system's knowledge about network 
+/// interfaces present on the system. Allows the user to iterate over 
+/// interface configurations.
+pub struct InterfaceAddrs {
+    inner: *mut libc::ifaddrs,
+    current: Option<&'static libc::ifaddrs>,
+}
+
+impl InterfaceAddrs {
+    /// Produce an `InterfaceAddrs` from the system's information. Returns `None`
+    /// if there are no interfaces to be inspected.
+    pub fn query_system() -> Option<Self> {
+       let mut p = null_mut();
+
+       // UNSAFETY: Calling libc FFI function, which allocates memory and
+       // fills it with info about interfaces.
+       unsafe { libc::getifaddrs(&mut p); } 
+        
+       // UNSAFETY: *mut -> &'static mut. This is known to be either in valid memory
+       // or null based on the guarantees of getifaddrs()
+       return unsafe{ p.as_ref() }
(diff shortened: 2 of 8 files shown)


src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+            ifu: ifu,
+            flags: flags,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn tests_get_if_addrs() {
```
asomers: s/get_if_addrs/getifaddrs/

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        let ifs = super::InterfaceAddrs::query_system().unwrap();
+        for i in ifs {
+            println!("{}:", i.name);
```
asomers: This isn't much of a test, because it doesn't verify anything.  It's really an example, and should go in the examples directory.  Making a proper test would be hard.  You'd probably have to compare against the output of `ifconfig`.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+/// interface configurations.
+pub struct InterfaceAddrs {
+    inner: *mut libc::ifaddrs,
+    current: Option<&'static libc::ifaddrs>,
+}
+
+impl InterfaceAddrs {
+    /// Produce an `InterfaceAddrs` from the system's information. Returns `None`
+    /// if there are no interfaces to be inspected.
+    pub fn query_system() -> Option<Self> {
```
asomers: May as well name this function `getifaddrs`.  That's the nix convention.  We usually name functions the same as the underlying C function, even though the API is rustier.

src/net/ifaddrs/iff_flags.rs
```diff
@@ -0,0 +1,51 @@
...
+        const IFF_PORTSEL = libc::IFF_PORTSEL as c_uint;
+        /// Auto media selection active.
+        const IFF_AUTOMEDIA = libc::IFF_AUTOMEDIA as c_uint;
+        /// The addresses are lost when the interface goes down.
+        const IFF_DYNAMIC = libc::IFF_DYNAMIC as c_uint;
+
+        // These flags are available on modern Linuxes
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        /// Driver signals L1 up (since Linux 2.6.17)
+        const IFF_LOWER_UP = 1<<16;
```
Susurrus: All constants should be defined in libc if appropriate. Please upstream these constants and then switch to the `libc_bitflags` macro that's defined in nix (see src/macros).

Susurrus:
First off, thanks for submitting this PR. Most of our functionality is contributed by others rather than us maintainers. Looking briefly through the code, I think it'd be appropriate to include in nix, and @asomers hasn't commented otherwise. We'll just need to do some minor  cleanups to the code to bring it inline with nix's conventions.

One thing that this PR is missing is a changelog entry describing the addition of this API.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub flags: IffFlags,
+}
+
+/// Represents the ifu of an interface: either its broadcast address or
+/// point-to-point destination address. 
+#[derive(Debug, Clone)]
+pub enum InterfaceIfu {
+    BroadcastAddr(Option<IpAddr>),
+    DestinationAddr(Option<IpAddr>),
+    Neither,
```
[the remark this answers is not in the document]
LeoTindall: Actually, that comment on 153 refers to a `BroadcastAddr(None)` or `DestinationAddr(None)`. I'll make that more clear.

src/net/ifaddrs/sockaddr.rs
```diff
@@ -0,0 +1,47 @@
...
+                // UNSAFETY: Transmuting a sockaddr into a sockaddr_in. 
+                // They're the same thing.
+                let data_v4: &libc::sockaddr_in = unsafe { transmute(sa) };
+
+                // UNSAFETY: Transmuting a u32 into a [u8; 4] because 
+                // the address is in network byte order.
+                let s_addr_v4: [u8; 4];
+                unsafe { s_addr_v4 = transmute(data_v4.sin_addr.s_addr); }
+                Some(IpAddr::V4(
+                   Ipv4Addr::from(s_addr_v4)
```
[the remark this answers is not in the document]
LeoTindall: I have tried this and it results in having backwards addresses (`lo` with the address `1.0.0.127`, et cetera).

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+use self::sockaddr::sockaddr_to_ipaddr;
+
+pub type InterfaceMap = HashMap<String, Vec<InterfaceAddr>>;
+
+/// Represents a handle into the operating system's knowledge about network 
+/// interfaces present on the system. Allows the user to iterate over 
+/// interface configurations.
+pub struct InterfaceAddrs {
+    inner: *mut libc::ifaddrs,
+    current: Option<&'static libc::ifaddrs>,
```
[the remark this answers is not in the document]
LeoTindall: It works just as well using either *mut or &'static; I was told on users.rustlang that it's more "rusty" to do it this way, but I'm happy to revert it if that's better style for the nix crate.

LeoTindall:
Thanks to both of you for commenting so extensively! Many of @asomers concerns are addressed in a pending commit (which will be pushed after this comment). Remaining:

- [ ] Error handling
- [ ] Upstream `IFF_LOWER_UP`, `IFF_DORMANT`, and `IFF_ECHO`
- [ ] Convert `ifaddrs::iff_flags` to use `libc_bitflags!` macro
- [ ] Implement `AF_LINK` addresses (does `std::net` have a representation for these?)
- [ ] Maybe switch from `&'static` back to `*mut` inside the struct
- [ ] Make a changelog entry

I'll do these over the weekend as much as possible, once my homework is done (darn college getting in the way of open source)

src/net/ifaddrs/sockaddr.rs
```diff
@@ -1,33 +1,64 @@
...
+        IfAddrValue::MacAddr(mac)
+    }
+}
+
+/// Converts a `libc::sockaddr` into an `Option<IfAddrValue>`.
 ///
-/// It returns `None` if the libc reports a type of address that `std::net`
-/// doesn't understand, or if the given `sockaddr_input` was null.
+/// It returns `None` if the libc reports a type of address other than
+/// IPv4, IPv6, or EUI-84 MAC, or if the given `sockaddr_input` was null.
```
asomers: s/84/48/

src/net/ifaddrs/sockaddr.rs
```diff
@@ -1,33 +1,64 @@
...
-                let data_v6: &libc::sockaddr_in6;
-                // UNSAFETY: Transmuting a sockaddr into a sockaddr_in6.
-                // They're the same thing.
-                data_v6 = transmute(sa);
-                Some(IpAddr::V6(Ipv6Addr::from(data_v6.sin6_addr.s6_addr)))
+                let data_v6: &libc::sockaddr_in6 = transmute(sa);
+                Some(IpAddr::V6(Ipv6Addr::from(data_v6.sin6_addr.s6_addr)).into())
+            }
+            libc::AF_PACKET => {
+                let data_mac: &libc::sockaddr_ll = transmute(sa);
```
asomers: On FreeBSD, you should use `sockaddr_dl`

src/net/ifaddrs/sockaddr.rs
```diff
@@ -1,33 +1,64 @@
...
             libc::AF_INET6 => {
-                let data_v6: &libc::sockaddr_in6;
-                // UNSAFETY: Transmuting a sockaddr into a sockaddr_in6.
-                // They're the same thing.
-                data_v6 = transmute(sa);
-                Some(IpAddr::V6(Ipv6Addr::from(data_v6.sin6_addr.s6_addr)))
+                let data_v6: &libc::sockaddr_in6 = transmute(sa);
+                Some(IpAddr::V6(Ipv6Addr::from(data_v6.sin6_addr.s6_addr)).into())
+            }
+            libc::AF_PACKET => {
```
asomers: Is `AF_PACKET` a Linux thing?  In the BSD world, it's `AF_LINK`.

src/net/ifaddrs/sockaddr.rs
```diff
@@ -1,33 +1,64 @@
...
+                let data_mac: &libc::sockaddr_ll = transmute(sa);
+                if data_mac.sll_halen != 6 {
```
asomers: FYI, Infiniband addresses are longer than 6 bytes.  I think it's fine for Nix to ignore them, at least for now.

src/net/ifaddrs/sockaddr.rs
```diff
@@ -1,33 +1,64 @@
...
+                    None // If the length of the hardware address (halen) isn't
+                        // 6, it's not EUI48 and we can't handle it.
+                } else {
+                    let a = data_mac.sll_addr;
+                    let eui: Eui48 = [a[0], a[1], a[2], a[3], a[4], a[5]];
```
asomers: Can this be expressed as an array operation?

src/net/ifaddrs/mod.rs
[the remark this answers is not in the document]
asomers: I'm very skeptical.  The only reason why your patch compiles now is because you cast that lifetime within an `unsafe` block.  Could you please show me the post where somebody recommended using a static lifetime?

src/net/ifaddrs/sockaddr.rs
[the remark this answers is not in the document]
asomers: This sounds like a bug in the standard library, then.  If you can reproduce it in a standalone test program, please report it upstream.

asomers:
We still need some tests here.  Fortunately, your code that calls `libc::getifaddrs` is well isolated from the code that interprets its output.  Could you write a test that takes canned `libc::getifaddrs` output and converts it to a `HashMap`?

src/net/ifaddrs/sockaddr.rs
LeoTindall: Yes, this is a Linux thing. Unfortunately I don't have a BSD box to test on, but I'll try to write conditionally compiled code for this.

src/net/ifaddrs/sockaddr.rs
LeoTindall: Not as far as I can tell, unfortunately; that is, there's no way to go from `data_mac.sll_addr[0..6]` (`&[u8]`) to `Eui48` (which is `[u8; 6]`) except explicitly, because there's no guarantee that `&[u8]` is 6 bytes long.

src/net/ifaddrs/sockaddr.rs
LeoTindall: Yes, this is intended to be implemented later.

src/net/ifaddrs/mod.rs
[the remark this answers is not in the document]
LeoTindall: Here's the post: https://users.rust-lang.org/t/code-review-for-my-first-libc-wrapper/12743/9?u=silverwingedseraph

As I said, I'm happy to switch it back.

src/net/ifaddrs/sockaddr.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl From<MacAddress> for IfAddrValue {
+    fn from(mac: MacAddress) -> IfAddrValue {
+        IfAddrValue::MacAddr(mac)
+    }
+}
+
+/// Converts a `libc::sockaddr` into an `Option<IfAddrValue>`.
+///
+/// It returns `None` if the libc reports a type of address other than
```
asomers: I've been chewing on this, and I've concluded that we're doing it wrong.  There are two problems:
1) Nix consumers can't extend this function to handle weird stuff like `AF_IPX`.
2) Handling link addresses is too platform specific.  That functionality probably doesn't belong in Nix at all.

Both of these problems share a common solution.  For unknown address families, `sockaddr_to_ifaddrvalue` should return `Other(&libc::sockaddr)` instead of `None`.  That way Nix consumers can do whatever they need to handle `AF_LINK` or anything else.

src/net/ifaddrs/mod.rs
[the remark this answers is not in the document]
asomers: I think you should.

src/net/ifaddrs/sockaddr.rs
LeoTindall: I like that solution a lot, actually. I'll go ahead and implement that.

LeoTindall:
Note that a consequence of the change to returning `Other(&sockaddr)` is that the reference inside `InterfaceAddrs` now has a sane lifetime.

LeoTindall:
OK, my checklist is done. I'd love to have a code review of the current state of this at your leisure @asomers

src/net/ifaddrs/mod.rs
```diff
@@ -76,17 +76,43 @@ use self::iff_flags::IffFlags;
...
+    ///
+    /// # Unsafety
+    /// The caller is responsible for making sure the given pointer is not
+    /// in invalid memory.
+    ///
+    /// # Errors
+    /// `Err(())` will be returned if `p` was void.
+    pub unsafe fn from_raw(p: *mut libc::ifaddrs) -> ::std::result::Result<Self, ()> {
+        match p.as_ref() {
+            Some(r) => Ok(Self {
```
asomers: This fails to compile.  You should change `Self` to `InterfaceAddrs`

src/net/ifaddrs/tests.rs
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+    let created = unsafe { InterfaceAddrs::from_raw(&mut test_lo_ipv4) }.unwrap();
+
+    let hm: InterfaceMap = created.into();
+
+    assert_eq!(hm.len(), 2, "Expected 2 interfaces, found {}.", hm.len());
+
+    assert!(hm.contains_key("test_lo"));
+    assert!(hm.contains_key("test_ext"));
```
asomers: How about assertions for the addresses in `hm`?  How about some IPv6 addresses and AF_LINK addresses in the mix?

src/net/ifaddrs/mod.rs
LeoTindall: It compiles for me? Something must be screwy with my Rustup, I'll change it. Thanks.

src/net/ifaddrs/mod.rs
```diff
@@ -76,17 +76,43 @@ use self::iff_flags::IffFlags;
...
+    /// The destructor will not attempt to free memory on an InterfaceAddrs
+    /// created in this way.
+    ///
+    /// # Unsafety
+    /// The caller is responsible for making sure the given pointer is not
+    /// in invalid memory.
+    ///
+    /// # Errors
+    /// `Err(())` will be returned if `p` was void.
+    pub unsafe fn from_raw(p: *mut libc::ifaddrs) -> ::std::result::Result<Self, ()> {
```
asomers: Since this function is only used by test code, it should not be `pub`

src/net/ifaddrs/mod.rs
```diff
@@ -76,17 +76,43 @@ use self::iff_flags::IffFlags;
...
 
 /// Represents a handle into the operating system's knowledge about network
 /// interfaces present on the system. Allows the user to iterate over
 /// interface configurations.
-pub struct InterfaceAddrs {
+pub struct InterfaceAddrs<'a> {
     inner: *mut libc::ifaddrs,
-    current: Option<&'static libc::ifaddrs>,
+    current: Option<&'a libc::ifaddrs>,
+    do_free: bool,
```
asomers: It looks like `do_free` only exists to support your test code.  You can eliminate it by using `mem::forget` in the test code to prevent `drop` from being run.

src/net/ifaddrs/mod.rs
```diff
@@ -101,16 +127,17 @@ impl InterfaceAddrs {
             Some(r) => Ok(Self {
```
asomers: Once you eliminate `do_free`, this section can turn into a call to `from_raw`.

src/net/ifaddrs/mod.rs
asomers: You fixed the wrong line.  Just check the Travis build log to see the error I'm talking about.  If it works for you, that's probably because you're using a newer version of Rust than Travis.  We're using 1.13.0 in Travis.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+//!
+//! The `InterfaceAddrs` struct provides access to the system's network
+//! interface data. You can either iterate over it or consume it and convert
+//! it into an `InterfaceMap` (a `HashMap<String, Vec<InterfaceAddr>>`) for
+//! more convenient access by interface name.
+//!
+//! # Examples
+//!
+//! You can access the basic information of the system in a few lines.
+//! The following program prints all the known addresses.
```
Susurrus: Let's get a colon at the end of this line.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+//!     .into(); // Convert to a hash map
+//!
+//! // Print all the addresses of the loopback interface
+//! if let Some(addrs) = interfaces.get("lo") {
+//!    println!("Loopback addresses:");
+//!    for addr in addrs {
+//!         println!("\t{:?}", addr);
+//!    }
+//! }
+//!
```
Susurrus: Remove this blank line.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+//! ```
+//!
```
Susurrus: And this one.

src/net/ifaddrs/iff_flags.rs
```diff
@@ -0,0 +1,50 @@
...
+        IFF_SLAVE as c_uint;
+        /// Supports multicast
+        IFF_MULTICAST as c_uint;
+        /// Is able to select media type via ifmap.
+        IFF_PORTSEL as c_uint;
+        /// Auto media selection active.
+        IFF_AUTOMEDIA as c_uint;
+        /// The addresses are lost when the interface goes down.
+        IFF_DYNAMIC as c_uint;
+
```
Susurrus: Please remove this blank line.

src/net/ifaddrs/iff_flags.rs
```diff
@@ -0,0 +1,50 @@
...
+        // These flags are available on modern Linuxes
+        #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: The OSes here should be listed alphabetically.

src/net/ifaddrs/iff_flags.rs
```diff
@@ -0,0 +1,50 @@
...
+        /// Driver signals L1 up (since Linux 2.6.17)
+        IFF_LOWER_UP as c_uint;
+        #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Same here.

src/net/ifaddrs/iff_flags.rs
```diff
@@ -0,0 +1,50 @@
...
+        /// Driver signals dormant (since Linux 2.6.17)
+        IFF_DORMANT as c_uint;
+        #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Same here.

src/net/ifaddrs/iff_flags.rs
```diff
@@ -0,0 +1,50 @@
+use libc;
```
Susurrus: Combine these into `use libc::{self, c_uint};`

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    /// Flags regarding the interface's behaviour and state
+    pub flags: IffFlags,
+}
+
+/// Represents the ifu of an interface: either its broadcast address or
+/// point-to-point destination address.
+#[derive(Debug, Clone)]
+pub enum InterfaceIfu<'a> {
+    BroadcastAddr(Option<IfAddrValue<'a>>),
```
Susurrus: All variants here need doc comments.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    DestinationAddr(Option<IfAddrValue<'a>>),
+    Neither,
+}
+
+
```
Susurrus: Remove the extra blank line here.

src/net/ifaddrs/tests.rs
```diff
@@ -0,0 +1,76 @@
+#![cfg(test)]
+
+use std::ptr::null_mut;
+use std::mem::transmute;
+
+use std::ffi::CString;
```
Susurrus: Please group all `use std ...` lines together and alphabetize them (so `std::ffi ...` should be first, etc.

src/net/ifaddrs/tests.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use libc;
```
Susurrus: These three uses can all be grouped together.

src/net/ifaddrs/tests.rs
```diff
@@ -0,0 +1,76 @@
...
+use libc::{ifaddrs, sockaddr, sockaddr_in};
+use libc::in_addr;
+
+use super::iff_flags::*;
+use super::InterfaceAddrs;
```
Susurrus: This and the one below it can be grouped together.

src/net/ifaddrs/tests.rs
```diff
@@ -0,0 +1,76 @@
...
+        sin_port: 0,
+        sin_addr: convert_v4_addr(b0, b1, b2, b3),
+        sin_zero: [0; 8],
+    };
+
+    unsafe { transmute(sin) }
+}
+
+#[test]
+fn test_ifaddrs() {
```
Susurrus: An overall summary of this test would be useful as a comment above it (put it above the `#[test]` part)

src/net/mod.rs
```diff
@@ -1,3 +1,5 @@
 // To avoid clashing with the keyword "if", we use "if_" as the module name.
 // The original header is called "net/if.h".
 pub mod if_;
+
```
Susurrus: Remove this blank line.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+        // Workaround for the borrow checker being overzealous
+        // (without ptr_temp, self.current would technically
+        // "still be in use" when the loop ends, meaning we
+        // couldn't advance to the next struct)
+        let ptr_temp = self.current.clone();
+        let p = ptr_temp.as_ref().unwrap();
+
+        // Get a pointer to the interface's name
+        let name_ptr = p.ifa_name;
+        // Check that name_ptr isn't null.
```
Susurrus: I think this check might be more semantically clear if you used `assert!` instead of an `if` here. @asomers any opinion on this?

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+        if name_ptr.is_null() {
+            panic!("getifaddrs() gave an ifaddrs struct with a null ifa_name");
+        }
+
+        // UNSAFETY: Constructing CStr from pointer. If this pointer is
```
Susurrus: Since the pointer being null is checked above with a panic if it's NULL at this point, it's a compiler error or something's super wrong with the CPU. I think this comment can be cleaned up a big to clarify that. Same for the one below.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// # Errors
+    /// `Err(())` will be returned if `p` was void.
+    pub unsafe fn from_raw(p: *mut libc::ifaddrs) -> ::std::result::Result<InterfaceAddrs<'a>, ()> {
+        match p.as_ref() {
+            Some(r) => Ok(Self {
+                inner: p,
+                current: Some(r),
+                do_free: false,
+            }),
+            None => Err(()),
```
Susurrus: Can't an Invalid Argument error be provided here instead which would be a bit clearer?

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+
+mod tests;
+
+pub type InterfaceMap<'a> = HashMap<String, Vec<InterfaceAddr<'a>>>;
+
+/// Represents a handle into the operating system's knowledge about network
+/// interfaces present on the system. Allows the user to iterate over
+/// interface configurations.
+pub struct InterfaceAddrs<'a> {
+    inner: *mut libc::ifaddrs,
```
Susurrus: Even those these inner members are not public, it's useful for developers to get some docs on these. Can you provide doc comments for each element here?

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+//!    println!("Loopback addresses:");
+//!    for addr in addrs {
+//!         println!("\t{:?}", addr);
+//!    }
+//! }
+//!
+//! ```
+//!
+
+use libc;
```
Susurrus: The grouping here should be `std`, then third-party, then local library modules each separated by a blank line. Additionally imports should be alphabetized.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+/// interface configurations.
+pub struct InterfaceAddrs<'a> {
+    inner: *mut libc::ifaddrs,
+    current: Option<&'a libc::ifaddrs>,
+    do_free: bool,
+}
+
+impl<'a> InterfaceAddrs<'a> {
+    /// Creates an `InterfaceAddrs` from a raw pointer, without calling into
+    /// the `libc`.
```
Susurrus: I don't think we need the "the" here.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// The destructor will not attempt to free memory on an InterfaceAddrs
+    /// created in this way.
+    ///
+    /// # Unsafety
+    /// The caller is responsible for making sure the given pointer is not
+    /// in invalid memory.
+    ///
+    /// # Errors
+    /// `Err(())` will be returned if `p` was void.
+    pub unsafe fn from_raw(p: *mut libc::ifaddrs) -> ::std::result::Result<InterfaceAddrs<'a>, ()> {
```
Susurrus: I'm trying to understand the use case for having this in our API? Is this really necessary for the user to do everything they want to do with `ifaddrs`? Where would the user get a struct like this rather than just have our own home-grown constructor that hides the struct creation from the user?

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+
+impl<'a> Drop for InterfaceAddrs<'a> {
+    fn drop(&mut self) {
+        if self.do_free {
+            // UNSAFETY: Calling libc FFI function which frees previously allocated
+            // memory.
+            unsafe {
+                // Ask the libc to drop free the memory it allocated when
```
Susurrus: "to free the memory" I think is what this should read.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+
+
+/// Represents the configuration and state of a network interface.
+/// Interfaces are uniquely identified by name, and each interface is likely
+/// to be referred to multiple times, e.g. one for IPv4 and one for IPv6.
+#[derive(Debug, Clone)]
+pub struct InterfaceAddr<'a> {
+    /// The name of the interface
+    pub name: String,
+
```
Susurrus: Please remove these blank lines from between the elements in this struct.

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+                // the struct was created.
+                libc::freeifaddrs(self.inner as *mut libc::ifaddrs);
+            }
+        }
+    }
+}
+
+
+/// Represents the configuration and state of a network interface.
+/// Interfaces are uniquely identified by name, and each interface is likely
```
Susurrus: Needs a blank line above for the generated docs to look pretty (run rustdoc to see what I'm talking about).

src/net/ifaddrs/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+            unsafe {
+                // Ask the libc to drop free the memory it allocated when
+                // the struct was created.
+                libc::freeifaddrs(self.inner as *mut libc::ifaddrs);
+            }
+        }
+    }
+}
+
+
```
Susurrus: Remove this blank line. Should only be 1 between things in a file.

src/net/ifaddrs/sockaddr.rs
```diff
@@ -0,0 +1,65 @@
...
+/// refers to valid memory.
+pub unsafe fn sockaddr_to_ifaddrvalue<'a>(
+    sockaddr_input: *mut libc::sockaddr,
+) -> Option<IfAddrValue<'a>> {
+    if let Some(sa) = sockaddr_input.as_ref() {
+        // Only IPv4 and IPv6 are supported.
+        match sa.sa_family as i32 {
+            libc::AF_INET => {
+                let data_v4: &libc::sockaddr_in = transmute(sa);
+                // Transmuting a u32 into a [u8; 4] because
```
Susurrus: This should be wrapped to the same number of columns as the code. We should be wrapping everything to 99 characters (which is the Rust standard).

src/net/ifaddrs/mod.rs
asomers: Yeah, `assert!` would be better.

PR #765, opened 2017-09-10
added docs for read/write

MarkusJais:
just some docs for read and write calls

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -756,27 +756,29 @@ pub fn close(fd: RawFd) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-/// reads from a raw file descriptor.
-/// reads at most as many bytes as will fit into the provided buffer
-/// 
-/// returns the number of bytes read or an error
+/// Reads from a raw file descriptor.
+/// Reads at most as many bytes as will fit into the provided buffer.
+///
+/// Returns the number of bytes successfully read.
+///
 /// For details see:
-/// [read(2)](http://man7.org/linux/man-pages/man2/read.2.html)
+/// [read(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/read.html)
 pub fn read(fd: RawFd, buf: &mut [u8]) -> Result<usize> {
     let res = unsafe { libc::read(fd, buf.as_mut_ptr() as *mut c_void, buf.len() as size_t) };
 
     Errno::result(res).map(|r| r as usize)
 }
 
-/// writes to a raw file descriptor
-/// writes at most as many bytes as there are in the provided buffer
+/// Writes to a raw file descriptor.
+/// Writes at most as many bytes as there are in the provided buffer.
 ///
-/// insufficient space on the underlying medium or settings for RLIMIT_FSIZE my limit
-/// the maximum number of bytes being written
+/// Insufficient space on the underlying medium or settings for RLIMIT_FSIZE may limit
+/// the maximum number of bytes being written.
 ///
-/// returns the number of bytes written or an error
+/// Returns the number of bytes successfully written.
+///
 /// For details see:
-/// [write(2)](http://man7.org/linux/man-pages/man2/write.2.html)
+/// [write(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html)
 pub fn write(fd: RawFd, buf: &[u8]) -> Result<usize> {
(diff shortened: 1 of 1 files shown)


asomers:
Thanks for the help!  Fleshing out documentation is always good.  Could you please change the links to point to the POSIX man pages instead of the Linux man pages?  Nix is a cross-platform crate, so POSIX is more applicable.
http://pubs.opengroup.org/onlinepubs/9699919799/functions/read.html
http://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html

src/unistd.rs
```diff
@@ -756,12 +756,27 @@ pub fn close(fd: RawFd) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// reads from a raw file descriptor.
```
Susurrus: s/reads/Reads/

src/unistd.rs
```diff
@@ -756,12 +756,27 @@ pub fn close(fd: RawFd) -> Result<()> {
...
+/// reads at most as many bytes as will fit into the provided buffer
```
Susurrus: Blank line before this one.

src/unistd.rs
```diff
@@ -756,12 +756,27 @@ pub fn close(fd: RawFd) -> Result<()> {
...
+/// 
+/// returns the number of bytes read or an error
```
Susurrus: The Error is pretty implied by the return value, I'd suggest instead "Returns the number of bytes successfully read". Also needs a period and capitalized first letter.

src/unistd.rs
```diff
@@ -756,12 +756,27 @@ pub fn close(fd: RawFd) -> Result<()> {
...
+/// For details see:
+/// [read(2)](http://man7.org/linux/man-pages/man2/read.2.html)
 pub fn read(fd: RawFd, buf: &mut [u8]) -> Result<usize> {
     let res = unsafe { libc::read(fd, buf.as_mut_ptr() as *mut c_void, buf.len() as size_t) };
 
     Errno::result(res).map(|r| r as usize)
 }
 
+/// writes to a raw file descriptor
```
Susurrus: Same guidelines as the above.

MarkusJais:
Fixed the typos and formatting problems. Added links to Posix docs instead of Linux man pages.

src/unistd.rs
asomers: You put the blank line in the wrong place.  Susurrus wanted the blank to be the second line of the comment block, so the first line would stand alone.

Susurrus: This is important because the first line is a brief summary and then there should be a blank line before additional information. I'd suggest reading the [documentation section](https://doc.rust-lang.org/stable/book/first-edition/documentation.html) of the Rust book, since many people don't seem to know the various conventions used by rustdoc.

src/unistd.rs
```diff
@@ -756,12 +756,29 @@ pub fn close(fd: RawFd) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Reads from a raw file descriptor.
+/// Reads at most as many bytes as will fit into the provided buffer.
```
Susurrus: You still need a blank line here after the first line in the docs:
rendered doc comment.

Please add that blank comment line and confirm that only the "Reads from a raw file descriptor" part renders on the `unistd` doc page. You can do this with `cargo doc --no-deps --open`. This also applies to `write()`.

src/unistd.rs
```diff
@@ -756,12 +756,29 @@ pub fn close(fd: RawFd) -> Result<()> {
...
+///
+/// Returns the number of bytes successfully read.
```
Susurrus: Put this in the same paragraph as the above sentence and please rewrite it so it flows a little better. Something like "The number of bytes actually read is returned."

MarkusJais: "The number of bytes actually read is returned." does not flow better in my opinion. Passive voice not necessary here.

src/unistd.rs
Susurrus: Still need a blank line here.

src/unistd.rs
Susurrus: Sure, I'm not too opinionated about this.

Susurrus:
This also needs a rebase.

MarkusJais:
Ok, I'll fix the blank lines. Might be a week or two due to limited time and problems with internet connection.
Once done I'll add more docs.

MarkusJais:
no longer needed, some other docs already present.

Issue #596, opened 2017-04-28
Implement fallocate(2)


agrover:
See man 2 fallocate. I think this should be relatively straightforward (at least I hope so.)

Susurrus:
@agrover Would you be willing to submit a PR adding this functionality? I'd be willing to mentor it through the process.

There's also `posix_fallocate(3)` for a more-portable version, but I don't know how similar those are.

SanchayanMaity:
@Susurrus As a beginner in rust, is this something I can take up? Would you mentor me and point me in the right direction to get started?

Susurrus:
Yeah, id be happy to mentor and get you started. The first step is for you to read over the CONTRIBUTING.md document which should cover the basics. Additionally you should read the man pages for fallocate and make sure you understand the API. After those things Please post back here if you have any fundamental questions. Otherwise just go ahead and open a PR and we can iterate there.

agrover:
fixed by #768, thanks!!

PR #768, opened 2017-09-14
Add support for 'fallocate'

SanchayanMaity:
For #596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#663](https://github.com/nix-rust/nix/pull/663))
 - Added more accessor methods for `AioCb`
   ([#773](https://github.com/nix-rust/nix/pull/773))
+- Add nix::sys::fallocate
+  ([#768](https:://github.com/nix-rust/nix/pull/768))
 
 ### Changed
 - Renamed existing `ptrace` wrappers to encourage namespacing ([#692](https://github.com/nix-rust/nix/pull/692))
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -341,3 +341,43 @@ pub fn vmsplice(fd: RawFd, iov: &[IoVec<&[u8]>], flags: SpliceFFlags) -> Result<
     Errno::result(ret).map(|r| r as usize)
 }
 
+#[cfg(any(target_os = "linux"))]
+libc_bitflags!(
+    /// Mode argument flags for fallocate determining operation performed on a given range.
+    pub struct FallocateFlags: libc::c_int {
+        /// File size is not changed.
+        ///
+        /// offset + len can be greater than file size.
+        FALLOC_FL_KEEP_SIZE;
+        /// Deallocates space by creating a hole.
+        ///
+        /// Must be ORed with FALLOC_FL_KEEP_SIZE. Byte range starts at offset and continues for len bytes.
+        FALLOC_FL_PUNCH_HOLE;
+        /// Removes byte range from a file without leaving a hole.
+        ///
+        /// Byte range to collapse starts at offset and continues for len bytes.
+        FALLOC_FL_COLLAPSE_RANGE;
+        /// Zeroes space in specified byte range.
+        ///
+        /// Byte range starts at offset and continues for len bytes.
+        FALLOC_FL_ZERO_RANGE;
+        /// Increases file space by inserting a hole within the file size.
+        ///
+        /// Does not overwrite existing data. Hole starts at offset and continues for len bytes.
+        FALLOC_FL_INSERT_RANGE;
+        /// Shared file data extants are made private to the file.
+        ///
+        /// Gaurantees that a subsequent write will not fail due to lack of space.
+        FALLOC_FL_UNSHARE_RANGE;
+    }
+);
+
+/// Manipulates file space.
+///
+/// Allows the caller to directly manipulate the allocated disk space for the
+/// file referred to by fd.
+#[cfg(any(target_os = "linux"))]
+pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc::off_t) -> Result<c_int> {
+    let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
+    Errno::result(res)
+}
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -54,11 +54,11 @@ mod linux_android {
 
     use libc::loff_t;
 
-    use nix::fcntl::{SpliceFFlags, splice, tee, vmsplice};
+    use nix::fcntl::{SpliceFFlags, FallocateFlags, fallocate, splice, tee, vmsplice};
     use nix::sys::uio::IoVec;
     use nix::unistd::{close, pipe, read, write};
 
-    use tempfile::tempfile;
+    use tempfile::{tempfile, NamedTempFile};
 
     #[test]
     fn test_splice() {
@@ -131,4 +131,15 @@ mod linux_android {
         close(wr).unwrap();
     }
 
+    #[test]
+    fn test_fallocate() {
+        let tmp = NamedTempFile::new().unwrap();
+
+        let fd = tmp.as_raw_fd();
+        fallocate(fd, FallocateFlags::empty(), 0, 100).unwrap();
+
+        // Check if we read exactly 100 bytes
+        let mut buf = [0u8; 200];
+        assert_eq!(100, read(fd, &mut buf).unwrap());
+    }
 }


SanchayanMaity:
@Susurrus Does this look alright? I believe tests are also required? Can you help me in that area? I was looking [here](https://github.com/linux-test-project/ltp/tree/master/testcases/kernel/syscalls/fallocate) for example tests.

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added socket option variant that enables the timestamp socket
   control message: `nix::sys::socket::sockopt::ReceiveTimestamp`
   ([#663](https://github.com/nix-rust/nix/pull/663))
+- Add nix::sys::fallocate
+  ([#596](https:://github.com/nix-rust/nix/pull/596))
```
Susurrus: This should be `768` and not `596`. You can either guess on the PR number or just modify the PR after you submit adding this in the future.

src/sys/fallocate.rs
```diff
@@ -0,0 +1,20 @@
+use libc;
+use libc::{off_t, c_int};
+use std::os::unix::io::RawFd;
+use {Errno, Result};
+
+libc_bitflags!(
+    pub struct FallocateFlags: libc::c_int {
+        FALLOC_FL_KEEP_SIZE;
```
Susurrus: These all need docs.

src/sys/fallocate.rs
```diff
@@ -0,0 +1,20 @@
...
+        FALLOC_FL_PUNCH_HOLE;
+        FALLOC_FL_COLLAPSE_RANGE;
+        FALLOC_FL_ZERO_RANGE;
+        FALLOC_FL_INSERT_RANGE;
+        FALLOC_FL_UNSHARE_RANGE;
+    }
+);
+
+pub fn fallocate(fd: RawFd, mode: c_int, offset: off_t, len: off_t) -> Result<c_int> {
```
Susurrus: Also needs docs, see the other PRs or the source for how they're commonly written.

Susurrus: The datatypes for this should not be the raw libc types. For example, `mode` should be `FallocateFlags`. I don't know what we should do for `off_t`, but it's likely we'd want to use `i64`s here instead. The main thing here is that we want to restrict the number of possible invalid inputs by using the type system which gives us compile-time errors. Like reading through the docs you can see that there are invalid combinations of `FALLOC_*` flags. We might want to figure out a way to encode that (I haven't thought about it too hard, but maybe)...

SanchayanMaity: Thanks for the feedback. I will implement the requested changes and get back in case of a query.

SanchayanMaity: @Susurrus How can I leverage the type system to restrict the number of possible valid inputs? I asked a friend of mine on how to do this, and he said that I would have to leverage traits.

asomers: @SanchayanMaity no need to use traits.  In this case, all you have to do is change `mode`'s type to `FallocateFlags`, like we do for `SigAction::new`.  I don't have any good ideas about how to restrict the combinations of flags besides declaring an `enum` with every valid combination.  But that would be awkward, because the values of the `enum` would no longer match anything in the `fallocate(2)` man page.

SanchayanMaity: @asomers Alright.
@Susurrus  The diff is what I have now. Does this look good now? Had one doubt. I have send patches before, but, for this pull request do I do a force push to the same branch of my fork or create a new branch and then create another pull request?

```C
diff --git a/CHANGELOG.md b/CHANGELOG.md
index d8be190..24c2007 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added socket option variant that enables the timestamp socket
   control message: `nix::sys::socket::sockopt::ReceiveTimestamp`
+- Add nix::sys::fallocate
+  ([#768](https:://github.com/nix-rust/nix/pull/768))
 
 ### Changed
 - Renamed existing `ptrace` wrappers to encourage namespacing ([#692](https://github.com/nix-rust/nix/pull/692))
diff --git a/src/fcntl.rs b/src/fcntl.rs
index 003c316..4ccbf12 100644
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -341,3 +341,28 @@ pub fn vmsplice(fd: RawFd, iov: &[IoVec<&[u8]>], flags: SpliceFFlags) -> Result<
     Errno::result(ret).map(|r| r as usize)
 }
 
+libc_bitflags!(
+    pub struct FallocateFlags: libc::c_int {
+        /// File size will not be changed even if offset + len is greater than file size.
+        FALLOC_FL_KEEP_SIZE;
+        /// Deallocates space in byte range starting at offset and continuing for len bytes.
+        FALLOC_FL_PUNCH_HOLE;
+        /// Removes a byte range from a file without leaving a hole. The byte range to
+        /// collapse starts at offset and continues for len bytes.
+        FALLOC_FL_COLLAPSE_RANGE;
+        /// Zeroes space in byte range starting at offset and continuing for len bytes.
+        FALLOC_FL_ZERO_RANGE;
+        /// Increases file space by inserting a hole within the file size and does not
+        /// overwrite existing data. Hole starts at offset and continues for len bytes.
+        FALLOC_FL_INSERT_RANGE;
+        /// Shared file data extants will be made private to the file to guarantee that a
+        /// subsequent write will not fail due to lack of space.
+        FALLOC_FL_UNSHARE_RANGE;
+    }
+);
+
+pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: i64, len: i64) -> Result<c_int> {
+    let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
+    Errno::result(res)
+}
```

agrover:
@SanchayanMaity just wanted to say thanks for working on this 😃

src/sys/fallocate.rs
Susurrus: That's definitely better, and you can do whatever style you want, I don't care, but I will ask to have your commit squashed into a single one at the end.

One thing to note about doc comments is that they should be written with a summary sentence on it's own line, then a blank line, then more details. I really encourage you to read all of the [documentation chapter of the Rust book](https://doc.rust-lang.org/stable/book/first-edition/documentation.html) to get familiar with the general usage of doc comments.

You still need a doccomment here for the `fallocate` function and the `FallocateFlags` struct as a whole (see the `src/macros.rs` file for where it should go.

Susurrus: Also, I think this is still supposed to be in the `fcntl.rs` file as that's the same header that defines it in `libc`. If you confirm that's the case, then please move it there.

SanchayanMaity: @Susurrus You are referring to rust-libc right? fallocate seems to be defined in [mod.rs](https://github.com/rust-lang/libc/blob/master/src/unix/notbsd/linux/mod.rs#L1115). What do you mean by "You still need a doccomment here for the fallocate function and the FallocateFlags struct as a whole (see the src/macros.rs file for where it should go."?

Susurrus: You need a comment describing the `FallocateFlags` type. You have comments for every variant, but not for the overall type. If you run `cargo doc --no-deps --open` and then look at the `fcntl.rs` page you'll see that the `FallocateFlags` type won't have any documentation. That should be fixed.

Mic92:
Is `posix_fallocate` not more common these days?

Mic92:
I needed this feature and added a test, feel free to add it:

https://github.com/Mic92/nix/commit/ba2929c4fac96658ca934d5c8df7444e751736c5#diff-b4dfe2398c5000381cff49cbbc0efa87

SanchayanMaity:
@Mic92 Thanks for the test.

test/test_fcntl.rs
```diff
@@ -132,3 +133,15 @@ mod linux_android {
     }
 
 }
+
+#[test]
+fn test_fallocate() {
```
Mic92: This needs to be moved into the module above as I did, so it only runs on linux/android.

SanchayanMaity: Argh.. my bad. Just saw the build failures. Had missed that. Fixed it now. Thank you very much.

SanchayanMaity:
I am confuse as to why the test ran for bsd platforms?!. The test for fallocate is scoped for Linux & Android.

Mic92:
no it is the `FallocateFlags` struct you have defined in `src/fcntl.rs`, which pulls in constant not available on bsd.

src/fcntl.rs
```diff
@@ -341,3 +341,39 @@ pub fn vmsplice(fd: RawFd, iov: &[IoVec<&[u8]>], flags: SpliceFFlags) -> Result<
...
+        FALLOC_FL_INSERT_RANGE;
+        /// Shared file data extants are made private to the file.
+        ///
+        /// Gaurantees that a subsequent write will not fail due to lack of space.
+        FALLOC_FL_UNSHARE_RANGE;
+    }
+);
+
+#[cfg(any(target_os = "linux"))]
+pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc::off_t) -> Result<c_int> {
```
asomers: Can we have a doc comment for this function?

SanchayanMaity: @asomers I added the doc comment and also rebased on top of master.

asomers: Please put two newlines between the two sentences of your doc comment.  Rustdoc uses such a gap to distinguish between brief and detailed docs.

SanchayanMaity: Modified it now.

SanchayanMaity: @asomers One of the Travis CI test seems to fail due to some TLS handshake timeout. Anything I need to do at my end? Do I need to force push again to inititate a build again and have all pass?

asomers: Nothing you need to do.  I restarted the build.

SanchayanMaity: @asomers @Mic92 @Susurrus Thanks for all the feedback and helping me through this.

asomers:
Try again.

bors r+

PR #771, opened 2017-09-24
ptrace: add PTRACE_O_EXITKILL option

bpowers:
It is a somewhat newer option -- it requires Linux 3.8. Is there a
more precise way to specify that?

### DIFF
--- a/src/sys/ptrace.rs
+++ b/src/sys/ptrace.rs
@@ -116,6 +116,10 @@ libc_bitflags! {
         /// Stop tracee when a SECCOMP_RET_TRACE rule is triggered. See `man seccomp` for more
         /// details.
         PTRACE_O_TRACESECCOMP;
+        /// Send a SIGKILL to the tracee if the tracer exits.  This is useful
+        /// for ptrace jailers to prevent tracees from escaping their control.
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        PTRACE_O_EXITKILL;
     }
 }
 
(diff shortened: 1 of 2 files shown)


asomers:
@bpowers no, there's no way to conditionally build stuff based on kernel version.  It's up to the application to check for `EINVAL`.  Also, could you please add a `CHANGELOG` entry?

src/sys/ptrace.rs
```diff
@@ -116,6 +116,10 @@ libc_bitflags! {
         /// Stop tracee when a SECCOMP_RET_TRACE rule is triggered. See `man seccomp` for more
         /// details.
         PTRACE_O_TRACESECCOMP;
+        /// Send a SIGKILL to the tracee if the tracer exits.  This is useful
+        /// for ptrace jailers to prevent tracees from escaping their control.
+        #[cfg(target_os = "linux")]
```
Susurrus: This should be available on Android as well. Please keep the OSes listed in alphabetical order here.

bpowers: good catch!  updated diff.

bpowers:
@asomers added a changelog entry, and enabled under android as @susurrus recommended.

bpowers:
that errored Travis test is unrelated to this -- seems to be some NetBSD problem?

bpowers:
awesome, thanks!!

Susurrus:
Thank you for your contribution to nix!

Issue #602, opened 2017-05-09
Request for safe mkfifo function


sector-f:
As far as I know, `mkfifo(3)` is just a wrapper around `mknod(3)`. Regardless, a FIFO can be created with `mknod`. 

I think a wrapper function (placed in `stat.rs`) would look something like this:

````rust
pub fn mkfifo<P: ?Sized + NixPath>(path: &P, perm: Mode) -> Result<()> {
    let res = try!(path.with_nix_path(|cstr| {
        unsafe {
            ffi::mknod(cstr.as_ptr(), nix::sys::stat::S_IFIFO.bits() as mode_t | perm.bits() as mode_t, 0)
        }
    }));

    Errno::result(res).map(drop)
}
````

Alternatively, you could also wrap `nix`'s existing safe `mknod` function like I did in this code:

````rust
use nix;
use nix::sys::stat::{mknod, Mode, S_IFIFO};
use nix::NixPath;

pub fn mkfifo<P: ?Sized + NixPath>(path: &P, perm: Mode) -> nix::Result<()> {
    mknod(path, S_IFIFO, perm, 0)
}
````

asomers:
Only on Linux is `mkfifo` a wrapper around `mknod`.  Your proposed implementation would be unnecessarily unportable.  What's wrong with using the regular `mkfifo` function?  That one is standardized.

sector-f:
> Only on Linux is mkfifo a wrapper around mknod.

Come to think of it, that does sound familiar. My mistake. 

> What's wrong with using the regular mkfifo function? That one is standardized.

Probably nothing; I'm just being forgetful and/or ignorant.

Susurrus:
@sector-f Would you be willing to contribute a PR adding a safe wrapper for mkfifo? `libc` already has the FFI definition for it, so all it takes is adding a function for it here in `nix`. I'd be happy to mentor you through it.

Susurrus:
Closed by #774

PR #774, opened 2017-10-04
#602 added unistd::mkfifo

jpastuszek:
Since libc has mkfifo already I have just copied mkdir and adapted to mkfifo. I have tested that on MacOS only.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -438,6 +438,49 @@ pub fn mkdir<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Creates new fifo special file (named pipe) with path `path` and access rights `mode`.
+///
+/// # Errors
+///
+/// There are several situations where mkfifo might fail:
+///
+/// - current user has insufficient rights in the parent directory
+/// - the path already exists
+/// - the path name is too long (longer than `PATH_MAX`, usually 4096 on linux, 1024 on OS X)
+///
+/// For a full list consult
+/// [posix specification](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mkfifo.html)
+///
+/// # Example
+///
+/// ```rust
+/// extern crate tempdir;
+/// extern crate nix;
+///
+/// use nix::unistd;
+/// use nix::sys::stat;
+/// use tempdir::TempDir;
+///
+/// fn main() {
+///     let tmp_dir = TempDir::new("test_fifo").unwrap();
+///     let fifo_path = tmp_dir.path().join("foo.pipe");
+/// 
+///     // create new fifo and give read, write and execute rights to the owner
+///     match unistd::mkfifo(&fifo_path, stat::S_IRWXU) {
+///        Ok(_) => println!("created {:?}", fifo_path),
+///        Err(err) => println!("Error creating fifo: {}", err),
+///     }
+/// }
+/// ```
+#[inline]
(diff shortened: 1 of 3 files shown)


src/unistd.rs
```diff
@@ -441,6 +441,49 @@ pub fn mkdir<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
...
+/// # Errors
+///
+/// There are several situations where mkfifo might fail:
+///
+/// - current user has insufficient rights in the parent directory
+/// - the path already exists
+/// - the path name is too long (longer than `PATH_MAX`, usually 4096 on linux, 1024 on OS X)
+///
+/// For a full list consult
+/// [man mkfifo(3)](http://man7.org/linux/man-pages/man3/mkfifo.3.html)
```
asomers: Odd that you apparently develop on OSX but reference Linux man pages.  However, nix is cross-platform, so the best man pages to use are opengroup's.  Could you please change the link to this?
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/mkfifo.html

jpastuszek: Well not that odd considering all other man page links in that file are pointing to http://man7.org/linux.

asomers: Yeah, we've been slowly moving them as we go, but we haven't mass-converted.

asomers:
Could you also please add an entry to the CHANGELOG?

Mic92:
you can probably adapt the tests from the code I made here: https://github.com/nix-rust/nix/pull/751/files#diff-899c5a4ff7ee773df222af8874b4761eR115

jpastuszek:
Is there anything else needed for this PR?

asomers:
Just a squash, then I'll merge it.

jpastuszek:
Squashed. Thank you.

asomers:
Sorry, but I just noticed one more problem: `mkfifo` belongs in `src/sys/stat.rs`, not `src/unistd.rs`, because its C prototype is located in `sys/stat.h`.

asomers:
Damn my out-of-date browser windows!  I guess this one slipped by.

jpastuszek:
Sorry, my bad. I should have read the comment on the top of the file.
My reasoning was that `mkfifo` is like `mkdir` or other kind of function that crates objects in the file system so I just placed it beside it... 

You may also want to consider moving `mkdir` so others won't make this mistake in the future:
```
NAME
     mkdir, mkdirat -- make a directory file

SYNOPSIS
     #include <sys/stat.h>

     int
     mkdir(const char *path, mode_t mode);
```

```
NAME
     mkfifo -- make a fifo file

SYNOPSIS
     #include <sys/types.h>
     #include <sys/stat.h>

     int
     mkfifo(const char *path, mode_t mode);
```

From http://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/stat.h.html:
> The following shall be declared as functions and may also be defined as macros. Function prototypes shall be provided.
```c
int    chmod(const char *, mode_t);
int    fchmod(int, mode_t);
int    fstat(int, struct stat *);
int    lstat(const char *restrict, struct stat *restrict);
int    mkdir(const char *, mode_t);
int    mkfifo(const char *, mode_t);
[XSI][Option Start]
int    mknod(const char *, mode_t, dev_t);
[Option End]
int    stat(const char *restrict, struct stat *restrict);
mode_t umask(mode_t);
```
And indeed the http://pubs.opengroup.org/onlinepubs/7908799/xsh/unistd.h.html does not list `mkfifo` or `mkdir`.

Do you want me to do something about this?

PR #777, opened 2017-10-08
Use the real pipe2(2) on most targets

asomers:
FreeBSD added pipe2(2) in version 10.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#800](https://github.com/nix-rust/nix/pull/800))
 
 ### Changed
+- Use native `pipe2` on all BSD targets.  Users should notice no difference.
+  ([#777](https://github.com/nix-rust/nix/pull/777))
 - Renamed existing `ptrace` wrappers to encourage namespacing ([#692](https://github.com/nix-rust/nix/pull/692))
 - Marked `sys::ptrace::ptrace` as `unsafe`.
 - Changed function signature of `socket()` and `socketpair()`. The `protocol` argument
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -876,10 +876,22 @@ pub fn pipe() -> Result<(RawFd, RawFd)> {
     }
 }
 
-// libc only defines `pipe2` in `libc::notbsd`.
-#[cfg(any(target_os = "linux",
-          target_os = "android",
-          target_os = "emscripten"))]
+/// Like `pipe`, but allows setting certain file descriptor flags.
+///
+/// The following flags are supported, and will be set atomically as the pipe is
+/// created:
+///
+/// `O_CLOEXEC`:    Set the close-on-exec flag for the new file descriptors.
+/// `O_NONBLOCK`:   Set the non-blocking flag for the ends of the pipe.
+///
+/// See also [pipe(2)](http://man7.org/linux/man-pages/man2/pipe.2.html)
+#[cfg(any(target_os = "android",
+          target_os = "dragonfly",
+          target_os = "emscripten",
+          target_os = "freebsd",
+          target_os = "linux",
+          target_os = "netbsd",
+          target_os = "openbsd"))]
 pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
     let mut fds: [c_int; 2] = unsafe { mem::uninitialized() };
 
@@ -890,9 +902,18 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
     Ok((fds[0], fds[1]))
 }
 
-#[cfg(not(any(target_os = "linux",
-              target_os = "android",
-              target_os = "emscripten")))]
+/// Like `pipe`, but allows setting certain file descriptor flags.
+///
+/// The following flags are supported, and will be set after the pipe is
+/// created:
+///
+/// `O_CLOEXEC`:    Set the close-on-exec flag for the new file descriptors.
+/// `O_NONBLOCK`:   Set the non-blocking flag for the ends of the pipe.
+#[cfg(any(target_os = "ios", target_os = "macos"))]
+#[deprecated(
+    since="0.10.0",
+    note="pipe2(2) is not actually atomic on these platforms.  Use pipe(2) and fcntl(2) instead"
+)]
 pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
     let mut fds: [c_int; 2] = unsafe { mem::uninitialized() };
 
@@ -905,9 +926,7 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
     Ok((fds[0], fds[1]))
 }
 
-#[cfg(not(any(target_os = "linux",
-              target_os = "android",
-              target_os = "emscripten")))]
+#[cfg(any(target_os = "ios", target_os = "macos"))]
 fn pipe2_setflags(fd1: RawFd, fd2: RawFd, flags: OFlag) -> Result<()> {
     use fcntl::FdFlag;
     use fcntl::FcntlArg::F_SETFL;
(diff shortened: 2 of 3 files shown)


Susurrus:
This needs an `Added` CHANGELOG entry and could also use comments for the tests.

src/unistd.rs
```diff
@@ -827,7 +827,8 @@ pub fn pipe() -> Result<(RawFd, RawFd)> {
 // libc only defines `pipe2` in `libc::notbsd`.
 #[cfg(any(target_os = "linux",
           target_os = "android",
-          target_os = "emscripten"))]
+          target_os = "emscripten",
+          target_os = "freebsd"))]
```
Susurrus: Since you're here, can you reorder the OSes to be alphabetical?

src/unistd.rs
```diff
@@ -840,7 +841,8 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
 
 #[cfg(not(any(target_os = "linux",
               target_os = "android",
-              target_os = "emscripten")))]
+              target_os = "emscripten",
+              target_os = "freebsd")))]
```
Susurrus: Same here.

src/unistd.rs
```diff
@@ -855,7 +857,8 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
...
```
Susurrus: And here.

src/unistd.rs
```diff
@@ -827,7 +827,8 @@ pub fn pipe() -> Result<(RawFd, RawFd)> {
 // libc only defines `pipe2` in `libc::notbsd`.
```
Susurrus: Can you clean up this comment? Seems out of date since freebsd support is being added.

Additionally, it appears that this is supported by NetBSD and OpenBSD according to `libc`, so let's add support for bind2 on those platforms as well.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#663](https://github.com/nix-rust/nix/pull/663))
 
 ### Changed
+- On FreeBSD, bind directly to `pipe2` instead of emulating it.  Users should
```
Susurrus: Since we tell the user they'll notice no difference, I think it'd be good to say why we're doing this. I'm not certain why we were emulating this in the first place, so that could be clarified here.

Also, this should be updated for NetBSD and OpenBSD also.

src/unistd.rs
asomers: I didn't check those platforms.  It looks like pretty much everything supports it now, even Illumos.

Susurrus:
Looks like it needs a rebase and I'm not certain why testing failed, let's give it another run.

Susurrus:
Needs a rebase then I'm r+ on it.

Susurrus:
Looks like Darwin doesn't support it or it's just not implemented in `libc` yet, which is why testing fails.

Also, it'd be good if you added doccomments for `pipe2` as part of this work since you're touching the function already and it's missing that.

Susurrus:
Note that `pipe2` doesn't exist on Mac, so this will need to be restricted from those targets.

asomers:
Oops, I forgot to update doc comments.  I'll do that shortly.

test/test_unistd.rs
```diff
@@ -291,3 +292,26 @@ fn test_sysconf_unsupported() {
...
+    assert_eq!(m1, stat::S_IFIFO);
+}
+
+// pipe2(2) is the same as pipe(2), except it allows setting some flags.  Check
+// that we can set a flag.
+#[test]
+fn test_pipe2() {
+    let (fd0, fd1) = pipe2(fcntl::O_CLOEXEC).unwrap();
+    let f0 = fcntl::FdFlag::from_bits_truncate(fcntl::fcntl(fd0, fcntl::F_GETFD).unwrap());
+    assert!(f0.contains(fcntl::FD_CLOEXEC));
```
Susurrus: Shouldn't you check that both of these `fd`s don't have this flag to begin with?

src/unistd.rs
```diff
@@ -869,9 +881,15 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
...
+/// Like `pipe`, but allows setting certain file descriptor flags.
+///
+/// The following flags are supported, and will be set after the pipe is
+/// created:
+///
+/// `O_CLOEXEC`:    Set the close-on-exec flag for the new file descriptors.
+/// `O_NONBLOCK`:   Set the non-blocking flag for the ends of the pipe.
+#[cfg(any(target_os = "ios",
+          target_os = "macos"))]
 pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
```
Susurrus: Does it make sense to keep this emulation for just ios/macos? We've removed emulated functionality before, and I'd like to suggest we remove it here as well.

test/test_unistd.rs
asomers: What do you mean "to begin with"?  `pipe2` sets that flag atomically as the pipe is created.

src/unistd.rs
asomers: Removing it would make me happy.  But perhaps we should mark it as deprecated for a release cycle first?

test/test_unistd.rs
Susurrus: I ask because in the case that `FD_CLOEXEC` is "default", this test could still pass even though the desired behavior didn't necessarily happen. I wanted to confirm that this assumption is correct.

Susurrus: I ask because in the case that `FD_CLOEXEC` is "default", this test could still pass even though the desired behavior didn't necessarily happen. I wanted to confirm that this assumption is correct.

src/unistd.rs
Susurrus: Sure, let's mark it deprecated for 0.10 and then remove for the next release.

test/test_unistd.rs
Susurrus: You're correct, ignore this comment.

asomers:
This is unusual.  The build failures in Travis were caused by a merge error.  Travis's merge error.  Travis pulled the master branch then tried to apply my patch, and it looks like it duped a line.  I'll rebase.

Susurrus:
Ahh, yes. Tests need to be updated because #801 was merged.

src/unistd.rs
```diff
@@ -905,9 +927,8 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
     Ok((fds[0], fds[1]))
 }
 
-#[cfg(not(any(target_os = "linux",
-              target_os = "android",
-              target_os = "emscripten")))]
+#[cfg(any(target_os = "ios",
```
Susurrus: This should be just 1 line.

src/unistd.rs
```diff
@@ -890,9 +902,19 @@ pub fn pipe2(flags: OFlag) -> Result<(RawFd, RawFd)> {
...
-              target_os = "android",
-              target_os = "emscripten")))]
+/// Like `pipe`, but allows setting certain file descriptor flags.
+///
+/// The following flags are supported, and will be set after the pipe is
+/// created:
+///
+/// `O_CLOEXEC`:    Set the close-on-exec flag for the new file descriptors.
+/// `O_NONBLOCK`:   Set the non-blocking flag for the ends of the pipe.
+#[cfg(any(target_os = "ios",
```
Susurrus: Should just be 1 line also.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#800](https://github.com/nix-rust/nix/pull/800))
 
 ### Changed
+- Use native `pipe` on all BSD targets.  Users should notice no difference.
```
Susurrus: Don't you actually mean `pipe2` here?

Susurrus:
Probably should have added that we deprecated our wrapper on other platforms in the CHANGELOG as well.

PR #778, opened 2017-10-08
Replace most Linux man page links with Open Group man page links

asomers:
But leave in place Linux links for non-standard functions.  Also, add
brief docs for some functions that were lacking them.

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -63,6 +63,9 @@ impl MqAttr {
 }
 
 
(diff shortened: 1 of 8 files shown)


posborne:
Excellent, thanks for making this change.  I'm probably to blame for the pervasive use of man7 (which is generally OK but linux specific in some cases).  Agreed that the man pages on open group are probably better as a standard reference in this case.

PR #779, opened 2017-10-09
 Add support for 'posix_fallocate' 

SanchayanMaity:


src/fcntl.rs
```diff
@@ -381,3 +381,13 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
     let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
     Errno::result(res)
 }
+
+/// Allocate file space.
+///
+/// Ensure disk space is allocated for the file referred to by the file descriptor
+/// fd for the bytes in the range starting at offset and continuing for len bytes.
+#[cfg(any(target_os = "linux"))]
```
asomers: This `#[cfg()]` statement is too narrow.  In addition to Linux, libc already defines `posix_fallocate` for FreeBSD, Dragonfly, and Android.  NetBSD has it too, though libc lacks the definition for NetBSD.  Please broaden the definition.  You will also need a matching `#[cfg()]` statement on the tests, or else nix's tests won't build on platforms that don't define this function.

Susurrus:
@SanchayanMaity You still planning on working on this?

SanchayanMaity:
@Susurrus Yes, I plan to work on this. Got busy with a few things last week. Will work on it today.

SanchayanMaity:
How to fix the build for darwin?

asomers:
The build failed on Darwin because libc doesn't define `posix_fallocate` for that platform.  Either the definition is missing from libc or Darwin really doesn't have `posix_fallocate`.  You need to figure out which.  In the former case, send a PR to libc.  In the latter, you just need to amend your `#[cfg()]` guard in this PR.

Susurrus:
@asomers @SanchayanMaity I just checked on my mac running 10.11.6 and there is no definition of `posix_fallocate`, so `macos` and `ios` targets should definitely be skipped.

Also, this needs a rebase so that it'll run on all the current targets `nix` supports.

src/fcntl.rs
```diff
@@ -381,3 +381,13 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
     let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
     Errno::result(res)
 }
+
+/// Allocate file space.
+///
+/// Ensure disk space is allocated for the file referred to by the file descriptor
+/// fd for the bytes in the range starting at offset and continuing for len bytes.
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "android", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd"))]
```
asomers: Please wrap long lines.

src/fcntl.rs
```diff
@@ -381,3 +381,13 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
     let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
     Errno::result(res)
 }
+
+/// Allocate file space.
+///
+/// Ensure disk space is allocated for the file referred to by the file descriptor
+/// fd for the bytes in the range starting at offset and continuing for len bytes.
```
asomers: Please add a link to http://pubs.opengroup.org/onlinepubs/9699919799/ .

test/test_fcntl.rs
```diff
@@ -143,3 +144,16 @@ mod linux_android {
...
     }
 }
+
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "android", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd"))]
+#[test]
+fn test_posix_fallocate() {
+    let tmp = NamedTempFile::new().unwrap();
+
+    let fd = tmp.as_raw_fd();
+    posix_fallocate(fd, 0, 100).unwrap();
```
asomers: This test will fail if /tmp's filesystem doesn't support `posix_fallocate`.  ZFS, for example, does not.  POSIX stupidly requires that `posix_fallocate` return `EINVAL` in that case.  So you should skip the rest of the test if you get that errno.

Susurrus:
@SanchayanMaity Any time to wrap up this PR? It's pretty close to being able to be merged.

SanchayanMaity:
@Susurrus Will pick this up again. Sorry for the delay. Was in between jobs.

src/fcntl.rs
```diff
@@ -380,3 +380,15 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
     let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
     Errno::result(res)
 }
+
+/// Allocate file space.
+///
+/// Ensure disk space is allocated for the file referred to by file
+/// descriptor fd for the bytes in the range starting at offset and
```
asomers: parameter names like `fd` and `offset` should be in backticks.

src/fcntl.rs
```diff
@@ -380,3 +380,15 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
...
+/// continuing for len bytes.
+/// http://pubs.opengroup.org/onlinepubs/9699919799/
```
asomers: The link is wrong.  I think you probably linked to the wrong iframe.  Also, for consistency with other Nix functions, you should format the link like this:
```rust
/// # References
///
/// [`posix_fallocate`](http://whatever)
```

src/fcntl.rs
```diff
@@ -380,3 +380,15 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
...
+#[cfg(any(target_os = "linux", target_os = "android", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd"))]
```
asomers: You still need to wrap long lines here and elsewhere.

Susurrus: This list should also be alphabetized.

test/test_fcntl.rs
```diff
@@ -143,3 +146,27 @@ mod linux_android {
...
+#[test]
+fn test_posix_fallocate() {
+    use nix::fcntl::{posix_fallocate};
+    let tmp = NamedTempFile::new().unwrap();
+
+    let fd = tmp.as_raw_fd();
+    let ret = posix_fallocate(fd, 0, 100).unwrap();
+
+    match Errno::result(ret) {
+       Err(Error::Sys(Errno::EINVAL)) => {
```
Susurrus: Clippy will get mad about this, a `match` where only 1 arm is checked. Please rewrite this using `if ret.is_some()`, but keep the comment there so it's clear what's going on.

Long-term it'd be better to actually check somehow before the test is run if this will work and skip the test otherwise. Adding a `TODO` here would be helpful for that.

test/test_fcntl.rs
```diff
@@ -143,3 +146,27 @@ mod linux_android {
         assert_eq!(100, read(fd, &mut buf).unwrap());
     }
 }
+
+#[cfg(any(target_os = "linux", target_os = "android", target_os = "freebsd", target_os = "dragonfly", target_os = "openbsd"))]
```
Susurrus: Please alphabetize this and line wrap it also.

test/test_fcntl.rs
```diff
@@ -1,10 +1,13 @@
+use nix::errno::{Errno};
```
Susurrus: Please remove unnecessary braces here and elsewhere.

Susurrus:
@SanchayanMaity Do you have time to come back and finish up this PR in the near future? It's been a while since you last updated it.

Susurrus:
You'll need to correct your CHANGELOG entries such that they end up in the UNRELEASED section.

src/fcntl.rs
```diff
@@ -380,3 +380,20 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
     let res = unsafe { libc::fallocate(fd, mode.bits(), offset, len) };
     Errno::result(res)
 }
+
+/// Allocate file space.
+///
+/// Ensure disk space is allocated for the file referred to by file
```
Susurrus: "to by the file"

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Add nix::sys::posix_fallocate
```
Susurrus: This should be under the "Added" section and rephrased like "Added `nix::sys::posix_fallocate`". Please also list the platforms that it was added for since it wasn't all platforms `nix` supports.

src/fcntl.rs
```diff
@@ -380,3 +380,20 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
...
+/// descriptor `fd` for the bytes in the range starting at `offset` and
+/// continuing for `len` bytes.
+///
+/// # References
+///
+/// [`posix_fallocate`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_fallocate.html)
+#[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd",
```
Susurrus: `posix_fallocate` is not defined in `libc` for DragonFlyBSD and OpenBSD and I couldn't find it in the public sources for those OSes either. I think this and the tests should be removed for those platforms.

src/fcntl.rs
```diff
@@ -380,3 +380,20 @@ pub fn fallocate(fd: RawFd, mode: FallocateFlags, offset: libc::off_t, len: libc
...
+          target_os = "linux", target_os = "openbsd"))]
+pub fn posix_fallocate(fd: RawFd, offset: libc::off_t,
```
Susurrus: Please annotate this as `#[inline]` since it's a very small wrapper around the bare `libc` call.

Susurrus:
@SanchayanMaity Want to take care of these comments so we can get this merged?

SanchayanMaity:
@Susurrus Yes, of course. Been busy lately. I will send out a patch tomorrow addressing your feedback.

Susurrus:
Please change the CHANGELOG entry to be under the current UNRELEASED section, as this didn't make it into the 0.10 release.

Then it LGTM. @asomers if you want another look-over, now's the time as I'll merge it as soon as this gets updated.

Susurrus:
Please move your changelog entry under the "Added" section within the "[UNRELEASED]" section and squash these commits into 1. Then we'll merge!

test/test_fcntl.rs
```diff
@@ -143,3 +144,23 @@ mod linux_android {
...
+#[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+#[test]
+fn test_posix_fallocate() {
+    use nix::fcntl::posix_fallocate;
+    let tmp = NamedTempFile::new().unwrap();
+
+    let fd = tmp.as_raw_fd();
+    let ret = posix_fallocate(fd, 0, 100);
+    if ret.is_ok() {
+        // The test will fail if /tmp's filesystem doesn't support
```
asomers: You shouldn't ignore _all_ errors, just `EINVAL`.

asomers:
Superseded by #1105

PR #780, opened 2017-10-12
Add a test for ppoll

asomers:


test/test_poll.rs
```diff
@@ -16,3 +18,22 @@ fn test_poll() {
     assert_eq!(nfds, 1);
     assert!(fds[0].revents().unwrap().contains(POLLIN));
 }
+
+#[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd",
```
Susurrus: It'd be great to have comments either inline or a big summary one that explains what this is doing and any gotchas that exist in this test. With a project as diverse as nix and the complexity of some of the APIs it exposes I think it's helpful to provide comments.

Susurrus:
Looks like our Rust 1.14 problems are solved.

Also, thanks a lot for those comments. I think they're helpful and hopefully weren't too much of a burden to write. Where's the gold star emoji!?

bors r+

asomers:
The failure looks related to Travis's build cache

bors r+ susurrus

asomers:
The build failed due to an ABI change at libc.  I suspect that the ABI change was unintentional.  If so, we should just wait for libc to revert it.  If not, we'll have to make a matching update to nix.
https://github.com/rust-lang/libc/commit/9029cfb9eb76647cf9122a47252f1129eb18a562

Susurrus:
This is already being resolved at least partially at rust-lang/libc#813. See the comment discussion I've been having over there. If you're willing to finish off the fix there, it'd be awesome.

PR #783, opened 2017-10-15
Promote i686-unknown-freebsd to Tier1 in the README

asomers:
We effectively made it Tier1 back in
ec6fe0629df87d8355cbdba8b7a0855934a5ac9e, but never updated the README.


Susurrus:
1.14 was having problems before, which is what we're seeing here, but it should have been fixed yesterday. Let's just give it a shot again.

bors r+

Issue #743, opened 2017-08-24
Statvfs b_fsize overflows in i686 (32bit) linux system


ixeft:
Hello,

I'm not a rust programmer, but trying to compile
https://github.com/greshake/i3status-rust
and particularly this file : 
https://github.com/greshake/i3status-rust/blob/master/src/blocks/disk_space.rs 
I noticed that Statvfs::b_fsize where u32 instead of u64 creating a compilation error 

using "as u64" I got the code compiling but now, with path bigger than 4Gio, I got an overflow
getting wrong value of statvfs.f_bfree / statvfs.f_bsize (crazy big values) 

Exemple of code that should demontrate the bug with 32bit system
```
extern crate nix;
use self::nix::sys::statvfs::vfs::Statvfs;
let statvfs = Statvfs::for_path(Path::new("a/path/with/size/bigger/than/4Gio"))
```
Using glibc in C/C++ I would define _LARGEFILE_SOURCE and _LARGEFILE64_SOURCE
to solve this problem. not sure how to do so here.

Regards,

Susurrus:
We just wrap the statvfs strict provided by libc. This is likely an issue with the type as it's declared there. Please open and issue there to discuss this and link back to this one.

ixeft:
yes, libc provide a flag to modify the type of those variable and allowing usage of more than 2Gio files (not 4Gio as I mentioned) by setting 
``` 
#define _LARGEFILE_SOURCE
#define _LARGEFILE64_SOURCE
```
or even better : 
```
#define _FILE_OFFSET_BITS
```
http://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html

ixeft:
Ok,
I just realized you wanted me to report a bug in i3status-rust, not in glibc :)

I just did that 
https://github.com/greshake/i3status-rust/issues/88

Thanks

asomers:
Actually, susurrus meant that you should report a bug to Rust's libc, here: https://github.com/rust-lang/libc .  That crate defines the FFI wrappers for all standard C structures.

Susurrus:
I believe the solution here is to expose the `stat*64` functions. for i3status to consume. I've therefore retitled this issue accordingly.

joshtriplett:
I'd love to have a safe wrapper for `statvfs64`. Should I open a fresh issue (or PR) for that?

asomers:
Sure.  If you're motivated, create a PR for it.

PR #784, opened 2017-10-19
adjusting statvfs struct to have 32-bit values for 32-bit machines

noyez:
This PR fixes the statvfs struct on 32-bit machines (my test bed was an R-PI). By default the statvfs struct uses mostly u64 values on all platforms. This PR uses u32 on 32-bit platforms and u64 on 64-bit platforms. Below is a short demonstration of the bug.

Given the following Rust program:

    extern crate nix;
    fn main() {
        let stat = nix::sys::statvfs::vfs::Statvfs::for_path("/").unwrap();
        println!("Total blocks: {}", stat.f_blocks);
        println!("Total blocks (as 32): {}", stat.f_blocks as u32);
    }

On my R-PI output:

    Total blocks: 836247313342446
    Total blocks (as 32): 942062

Consider the following python command: 

    python -c "import os;  statvfs = os.statvfs('/'); print(statvfs.f_blocks)"

On my R-PI:

    942062

### DIFF
--- a/src/sys/statvfs.rs
+++ b/src/sys/statvfs.rs
@@ -51,6 +51,40 @@ pub mod vfs {
     /// http://linux.die.net/man/2/statvfs
     #[repr(C)]
     #[derive(Debug,Copy,Clone)]
+    #[cfg(target_pointer_width = "32")]
+    pub struct Statvfs {
+        /// Filesystem block size. This is the value that will lead to
+        /// most efficient use of the filesystem
+        pub f_bsize: c_ulong,
+        /// Fragment Size -- actual minimum unit of allocation on this
+        /// filesystem
+        pub f_frsize: c_ulong,
+        /// Total number of blocks on the filesystem
+        pub f_blocks: u32,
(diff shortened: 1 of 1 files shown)


Susurrus:
This actually shouldn't be something `nix` does, declaring `libc` datatypes. If you search the libc repo you'll see that `statvfs` and `statvfs64` is already declared for us, we should just make `nix` use a Newtype around those structs instead of declaring them directly. See the work linked to from #748.

Additionally, this may be related to #743, which is about getting the 64-bit version implemented.

Susurrus:
Actually, this is already fixed in git, doing exactly what I said, so I don't think this PR is relevant anymore. Can you check the latest code and confirm that this is resolved for you?

Susurrus:
See 6ae3f319b for the commit that already fixed this.

noyez:
Yup, i just tested master on my platform, works as expected.

I swear i tried master before i created the PR, but it didn't compile at that instant (for some unmemorable reason). Eitherway, it works now.

PR #785, opened 2017-10-26
sys/socket: add UnixAddr abstract name getter

lucab:
This introduces an `as_abstract()` getter to `UnixAddr` in order to
retrieve the name of an abstract unix socket.
This also adds tests around abstract addresses and clarify docs,
adding explicit semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `nix::sys::uio::{process_vm_readv, process_vm_writev}` on Linux
   ([#568](https://github.com/nix-rust/nix/pull/568))
 - Added `nix::unistd::{getgroups, setgroups, getgrouplist, initgroups}`. ([#733](https://github.com/nix-rust/nix/pull/733))
+- Added `nix::sys::socket::UnixAddr::as_abstract` on Linux and Android.
+  ([#785](https://github.com/nix-rust/nix/pull/785))
 
 ### Changed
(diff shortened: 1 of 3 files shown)


lucab:
It looks like master is currently broken due to this libc change: https://github.com/rust-lang/libc/pull/787. /cc @vojtechkral

I also realized I forgot the relevant CHANGELOG entry, I'll add it on the next rebase once master is fixed. Reviews and comments still welcome in the meanwhile.

vojtechkral:
Hi @lucab 
To fix this, you probably just need to change `#[repr(i32)]` of the `EventFilter` to `#[repr(u32)]`. (OTOH if you believe my fix was incorrect or should not have been merged, please feel free to comment in the libc issue thread.)

lucab:
PTAL on second iteration:
 * rebased on a green master
 * added CHANGELOG entries
 * restricted abstract addresses methods to Linux only

@asomers I did as you suggested but I am not a fan of it, because `UnixAddr` is just a container for a typed address. I don't see why it should conditionally change depending on where you build. Perhaps a `LinuxExt` trait?

asomers:
@lucab The `#[cfg(target_os = "linux")]` guard doesn't care about where you build, only where you target.  And abstract sockets only work on Linux*.  So it doesn't make sense to enable that code for any other OS.  And I don't see what value a `LinuxExt` trait would add.  Could you please explain?

* It would be worth checking if Android, which is considered a separate OS by Rust,  supports them too.

lucab:
@asomers I'm fine with your feedback, I don't want to keep bikeshedding this further.

I've double-checked Android docs and then added support for it as well, PTAL.

asomers:
Thanks for your submission.

bors r+

PR #772, opened 2017-09-25
haiku: add platform support.

jessicah:

PR #787, opened 2017-10-29
Omit invalid waitpid flags on OpenBSD

worr:
OpenBSD doesn't have `WEXITED`, `WSTOPPED`, or `WNOWAIT`, so omit those
from that platform.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -8,10 +8,31 @@ libc_bitflags!(
     pub struct WaitPidFlag: c_int {
         WNOHANG;
         WUNTRACED;
+        #[cfg(any(target_os = "android",
+                  target_os = "freebsd",
+                  target_os = "haiku",
+                  target_os = "ios",
+                  target_os = "linux",
+                  target_os = "macos",
+                  target_os = "netbsd"))]
         WEXITED;
         WCONTINUED;
+        #[cfg(any(target_os = "android",
+                  target_os = "freebsd",
+                  target_os = "haiku",
+                  target_os = "ios",
+                  target_os = "linux",
+                  target_os = "macos",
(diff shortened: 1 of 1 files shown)


Susurrus:
We don't like to blacklist OS targets and instead follow a whitelist strategy. Please correct this to a whitelist for appropriate platforms.

The test failure is due to an upstream change in libc that needs to be fixed.

worr:
I've updated the PR to be a whitelist of targets.

Susurrus:
A few issues here:

- Please order the `target_os` OSes alphabetically.
- You have unrelated style changes wrapped up with functionality changes. This is widely considered bad-form with version control. Please separate this into two commits (their order doesn't matter) with your real code changes in one and style changes in the other.

Susurrus:
Additionally, please rebase now that #790 was merged.

worr:
Sorry about that, I messed up! I just pushed the latest rebased diffs.

Susurrus:
No worries. But looking at this further, this seems to actually restrict these constants beyond OpenBSD. These three constants are available on haiku, ios, macos, and netbsd as well. Please re-add those platforms for those constants. You can search for them in the libc crate to see exactly which platforms support what for future reference.

worr:
Whoops, I didn't think to check libc. Instead I checked the manpages, and it looks like I checked some out-of-date manpages.

I had no idea that y'all supported Haiku 😲

Susurrus:
@worr We don't technically support Haiiku OS (though there is work in #772 to do just that), but I like to support things on all platforms that libc is supporting.

Anyways, thanks for cleaning this up. Was this all that was necessary for OpenBSD to build again?

bors r+

PR #789, opened 2017-11-05
Add unistd::getlogin

scvalex:
Add wrapper for getlogin(3).

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -716,7 +716,9 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
 /// null-terminate in this case, but the nix implementation will ensure that the
 /// buffer is null terminated in this case.
 ///
-/// ```no_run
+/// # Example
+///
+/// ```
 /// use nix::unistd;
 ///
 /// let mut buf = [0u8; 64];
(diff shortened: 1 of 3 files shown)


scvalex:
It looks like nix is failing to build on `x86_64-unknown-netbsd`, but it doesn't look like the breakage was introduced in this pull request.

src/unistd.rs
```diff
@@ -735,6 +735,27 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
...
+/// Get a CStr containing the name of the user logged in on the
+/// controlling terminal of the process
+/// ([getlogin(3)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/getlogin.html)).
+///
+/// The posix specification says that "the string is statically
+/// allocated and might be invalidated or overwritten on subsequent
+/// calls to this function", so you may want to copy the value before
+/// using it.
+///
+/// ```no_run
```
asomers: Why the `no_run`?  This looks like a fine test case to me.  Are there some circumstances where it won't work?

scvalex:
Removed the `no_run`.  I had copied that from `gethostname` above.

src/unistd.rs
```diff
@@ -735,6 +735,27 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
...
+/// Get a CStr containing the name of the user logged in on the
+/// controlling terminal of the process
+/// ([getlogin(3)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/getlogin.html)).
+///
+/// The posix specification says that "the string is statically
+/// allocated and might be invalidated or overwritten on subsequent
+/// calls to this function", so you may want to copy the value before
+/// using it.
+///
+/// ```
```
Susurrus: This should go in an Examples section. See the [Rust docs about this](https://doc.rust-lang.org/book/first-edition/documentation.html#special-sections).

scvalex: Added it to "Example" because that was more common in the rest of the file, and removed the `no_run` from `gethostname` because we might as well be consistent.

src/unistd.rs
```diff
@@ -735,6 +737,29 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
     })
 }
 
+/// Get a CStr containing the name of the user logged in on the
```
Susurrus: Mentioning `CStr` is redundant here, let's shorten this a little to "Get the name of the user logged in on the controlling terminal of the process".

scvalex:
Done.  Thanks for your patiance requesting all the changes.

Susurrus:
@asomers Can you take a look at this especially in regards to safety? I think this is fine, but would appreciate an extra set of eyes?

asomers:
`libc::getlogin` is not thread-safe.  For Rust it doesn't make much sense to expose both `getlogin` and `getlogin_r`, so can we use `getlogin_r` internally?

asomers:
Your latest test run shows why you used `no_run` in the first place.  It seems that Linux only checks stdin, and Travis must close that or something.  But reading the POSIX docs, there are many other reasons why `getlogin` could fail in a CI environment too.  Probably best to restore `no_run`

Susurrus:
@scvalex Will you have time to get this merged soon? It's been around a month and I'd either like to move this forward or close this PR.

Susurrus:
It's been two months since my last ping and three since the last comment by the OP. Thanks for the PR, but I'm going to close this PR due to inaction.

PR #790, opened 2017-11-05
Fix netbsd kevent for breakage introduced by libc

Susurrus:


asomers:
So to be clear, the problem is that libc switched kevent.filter from i32 to u32?  In addition to fixing Nix for 0.10.0, should we release a 0.9.1 that contains a maximum allowed version of libc in Cargo.toml to prevent people from running into this problem?

Susurrus:
Probably. What we should definitely do during a release is to use a locked version of libc instead of specifying a minimum version. Because their backwards compatibility guarantees are different from ours.

asomers:
Supposedly libc always guarantees backwards compatibility.  In the past, they've even refused to let me fix buggy definitions because it would break backwards compatibility.  This kevent business is the first time I've seen otherwise.

Susurrus:
@asomers They only guarantee backwards compatibility for certain targets (Tier 1 I think), not all targets, since that doesn't match our target list, we have this issue.

src/sys/event.rs
```diff
@@ -30,51 +30,44 @@ type type_of_udata = intptr_t;
...
-    EVFILT_WRITE = libc::EVFILT_WRITE,
-    EVFILT_AIO = libc::EVFILT_AIO,
-    EVFILT_VNODE = libc::EVFILT_VNODE,
-    EVFILT_PROC = libc::EVFILT_PROC,
-    EVFILT_SIGNAL = libc::EVFILT_SIGNAL,
-    EVFILT_TIMER = libc::EVFILT_TIMER,
+libc_enum!{
+    #[cfg_attr(target_os = "netbsd", repr(u32))]
+    #[cfg_attr(not(target_os = "netbsd"), repr(i16))]
+    #[derive(Clone, Copy, Debug, PartialEq)]
```
asomers: Looks like the build is breaking because all of these `derive`d things are also `derive`d by `libc_enum!`.

src/sys/event.rs
```diff
@@ -30,51 +30,44 @@ type type_of_udata = intptr_t;
...
-    #[cfg(any(target_os = "macos",
-              target_os = "ios",
-              target_os = "dragonfly",
-              target_os = "freebsd"))]
-    EVFILT_USER = libc::EVFILT_USER,
-    #[cfg(any(target_os = "macos", target_os = "ios"))]
-    EVFILT_VM = libc::EVFILT_VM,
-    EVFILT_VNODE = libc::EVFILT_VNODE,
-    EVFILT_WRITE = libc::EVFILT_WRITE,
+cfg_if!{
```
asomers: Why use a `cfg_if!` instead of just plain `#[cfg()]` like the old code did?

src/sys/event.rs
Susurrus: Yes, WIP since I don't know how to crosscompile to these platforms from Linux.

src/sys/event.rs
Susurrus: Remnant of my first revision. This was supposed to make things clearer, but can you can't put the `cfg_attr(..., repr(..))` inside it, so it's less useful.

src/sys/event.rs
asomers: Cross compiling is blissfully easy.  It's the best thing about FFI.  If you use rustup, you can do something like `rustup target add i686-unknown-freebsd` then `cargo build --target=i686-unknown-freebsd`.

Susurrus:
@asomers I think this should be GTM, please review.

asomers:
Everything LGTM

bors r+

PR #792, opened 2017-11-05
Use libc_enum! where possible

### LINKED ISSUE #254: Use enums for enumeration constants

wginolas:
Some enums which use different names for values than libc still set the
discriminators manually.

closes #254

### DIFF
--- a/src/sys/aio.rs
+++ b/src/sys/aio.rs
@@ -19,9 +19,12 @@ libc_enum! {
         /// do it like `fsync`
         O_SYNC,
         /// on supported operating systems only, do it like `fdatasync`
-        #[cfg(any(target_os = "openbsd", target_os = "bitrig",
-                target_os = "netbsd", target_os = "macos", target_os = "ios",
-                target_os = "linux"))]
+        #[cfg(any(target_os = "bitrig",
+                  target_os = "ios",
+                  target_os = "linux",
+                  target_os = "macos",
+                  target_os = "netbsd",
+                  target_os = "openbsd"))]
         O_DSYNC
     }
 }
--- a/src/sys/aio.rs
+++ b/src/sys/aio.rs
@@ -13,38 +13,41 @@ use sys::time::TimeSpec;
 
 /// Mode for `AioCb::fsync`.  Controls whether only data or both data and
 /// metadata are synced.
-#[repr(i32)]
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub enum AioFsyncMode {
-    /// do it like `fsync`
-    O_SYNC = libc::O_SYNC,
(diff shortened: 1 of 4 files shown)


wginolas:
My editor used rustfmt, so this change is bigger than necessary. Should I undo the stuff rustfmt changed?

Susurrus:
There's no problem with those changes being the PR, though I don't agree with all of the changes. In general when contributing PRs, however, you should separate changes into multiple commits to make it easier to review. In this case, I'd suggest one commit for the functional code changes and one commit for all the remaining style changes.

wginolas:
OK, I removed the formatting changes for now.

src/sys/aio.rs
```diff
@@ -13,38 +13,41 @@ use sys::time::TimeSpec;
...
-              target_os = "netbsd", target_os = "macos", target_os = "ios",
-              target_os = "linux"))]
-    O_DSYNC = libc::O_DSYNC
+libc_enum! {
+    #[repr(i32)]
+    pub enum AioFsyncMode {
+        /// do it like `fsync`
+        O_SYNC,
+        /// on supported operating systems only, do it like `fdatasync`
+        #[cfg(any(target_os = "openbsd", target_os = "bitrig",
```
Susurrus: Since you're touching these lines I'd ask that you also alphabetize the target OSes and when they're longer than a line and need to be wrapped that you wrap them such that there's 1 OS per line. This applies across this whole PR.

Susurrus:
You'll need to rebase now that #790 was merged.

src/sys/aio.rs
```diff
@@ -19,9 +19,13 @@ libc_enum! {
         /// do it like `fsync`
         O_SYNC,
         /// on supported operating systems only, do it like `fdatasync`
-        #[cfg(any(target_os = "openbsd", target_os = "bitrig",
-                target_os = "netbsd", target_os = "macos", target_os = "ios",
-                target_os = "linux"))]
+        #[cfg(any(
```
Susurrus: Can you put the first target OS on the same line as `#[cfg(any(`? This is consistent with our styling elsewhere.

wginolas:
OK, I will try to take care of this in ~10 hours.

Susurrus:
Please squash your last two commits into one as they're both part of the same change. I believe the test failures on Darwin targets were changes in Cross that have since been resolved, but I'm testing now. So with those things fixed, we can merge this.

PR #794, opened 2017-11-05
Migrate memfd constants to libc

Susurrus:
This will fail because rust-lang/libc#836 is not merged yet, but just getting it staged for when it is.


Susurrus:
@asomers So it looks like some of the memfd constants aren't on musl targets, so when I filed rust-lang/libc#836 I left it off those platforms. But I'd think that even if it isn't offered by musl, it should be available on all linux platforms as it's a kernel interface, but I'm not certain if that's correct. I've pinged the libc people about it so we'll see what they say.

PR #795, opened 2017-11-13
Use sh as interpreter

dereckson:
No bash-specific feature is used, so deploy script will work
in every POSIX OS, even when bash isn't installed.


asomers:
Thanks for the fix.  We don't actually use this script anymore since we switched to docs.rs.  But it's still better to have it working than not.

bors r+

PR #796, opened 2017-11-13
Bulletproof the deploy script

dereckson:
If the target/doc folder can't be reached, the script should exit
without trying to continue.

This issue has been found using [ShellCheck](https://github.com/koalaman/shellcheck):

```
In deploy.sh line 5:                   
cd target/doc                          
^-- SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.    
```

Reference: https://github.com/koalaman/shellcheck/wiki/SC2164

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -2,7 +2,7 @@
 
 rev=$(git rev-parse --short HEAD)
 
-cd target/doc
+cd target/doc || exit 72
 
 git init
 git config user.name "Carl Lerche"


deploy.sh
```diff
@@ -2,7 +2,7 @@
 
 rev=$(git rev-parse --short HEAD)
 
-cd target/doc
+cd target/doc || exit 72
```
asomers: What's magic about "72"?

dereckson: Should be 66 actually.

72 means "a system file hasn't been found", 66 means "a non system file hasn't been found"

That follows libc standard exit codes (and as there is no other standard specific for bash, it's a good way to explain what happens).

Susurrus:
I believe we can just delete this shell script. I think I've brought it up before but I think now we should actually do it.

deploy.sh
asomers: Are you talking about sysexits.h?  That's not part of libc.  It actually comes from Sendmail.  libc only defines `EXIT_SUCCESS` and `EXIT_FAILURE`.  sysexits.h seems to have fallen out of favor years ago, at least in the BSD world.

http://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdlib.h.html

Susurrus:
Closing this in favor of #797.

@dereckson Thanks for bringing up the use of this script tho, glad we can clean up more outdated code in our codebase.

PR #800, opened 2017-11-19
unistd: add execveat() on Linux and Android

lucab:
This adds execveat() to `nix::unistd`. It uses the execveat(2) Linux
kernel syscall, which is available since 3.19.
This is a Linux-specific extension which is not covered by POSIX and
does not have any userland libc wrapper.

Ref: http://man7.org/linux/man-pages/man2/execveat.2.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `nix::unistd::{getgroups, setgroups, getgrouplist, initgroups}`. ([#733](https://github.com/nix-rust/nix/pull/733))
 - Added `nix::sys::socket::UnixAddr::as_abstract` on Linux and Android.
   ([#785](https://github.com/nix-rust/nix/pull/785))
+- Added `nix::unistd::execveat` on Linux and Android.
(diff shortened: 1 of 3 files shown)


lucab:
I had to constrain Linux tests to x86 and amd64 only, as travis was failing with `qemu: unsupported syscall` everywhere else.

asomers:
Please s/Linus/Linux in the commit message.
Were you able to run the tests on Android?  Our CI system only builds for Android, it doesn't test.

lucab:
Woops for the typo. I rebased to fix that and to resolve a conflict on the changelog, PTAL.

No, I didn't test explicitly test this on Android as I don't have at hand an environment for that.

asomers:
I don't have an Android test environment either.  But it looks correct and Android is only Tier 2, so it's ok to merge.

lucab:
Ack. May I then get a review and merge? Or is there anything else I can do here?

Susurrus:
@lucab @asomers was likely waiting for one of the other maintainers to have a chance to look this over in case he missed something. We do that pretty commonly, so there wasn't anything for you to do but wait!

LGTM as well.

bors r+

PR #801, opened 2017-11-22
Upgrade to Bitflags 1.0

### LINKED ISSUE #766: Upgrade to Bitflags 1.0
This isn't straightforward because something about 944ee530cd04712021ee21a60f1ef59d2ab6d489 broke our `libc_bitflags!` macro with the following error:

```rust
   Compiling bitflags v0.9.1
   Compiling nix v0.10.0-pre
error: no rules expected the token `cfg(any(target_os = "android", target_os = "linux"))`
   --> src/macros.rs:129:19
    |
129 |                   #[$attr]
    |                     ^^^^^
    | 
   ::: src/fcntl.rs
    |
11  | / libc_bitflags!{
12  | |     pub struct AtFlags: c_int {
13  | |         AT_SYMLINK_NOFOLLOW;
14  | |         #[cfg(any(target_os = "android", target_os = "linux"))]
...   |
18  | |     }
19  | | }
    | |_- in this macro invocation

error: Could not compile `nix`.

To learn more, run the command again with --verbose.
```

So that should get figured out so we get nice constant scoping in our docs. Note that this will require use to require 1.20 as our minimum Rust version, but I think it's worth it for us to switch to that because of the significant documentation win.

Susurrus:
The libc_bitflags! macro was replaced with a non-recursive one supporting
only public structs. I could not figure out how to make the old macro work
with the upgrade, so I reworked part of the bitflags! macro directly to suit
our needs, much as the original recursive macro was made. There are no uses
of this macro for non-public structs, so this is not a problem for internal code.

Closes #766.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,67 +18,67 @@ matrix:
   include:
     # Android
     - env: TARGET=aarch64-linux-android DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.20.0
     - env: TARGET=arm-linux-androideabi DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.20.0
     - env: TARGET=armv7-linux-androideabi DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.20.0
     - env: TARGET=aarch64-apple-ios DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.20.0
       os: osx
     - env: TARGET=i686-linux-android DISABLE_TESTS=1
-      rust: 1.18.0
+      rust: 1.20.0
     - env: TARGET=x86_64-linux-android DISABLE_TESTS=1
-      rust: 1.18.0
+      rust: 1.20.0
 
     # Linux
     - env: TARGET=aarch64-unknown-linux-gnu
-      rust: 1.13.0
+      rust: 1.20.0
     - env: TARGET=armv7-apple-ios DISABLE_TESTS=1
-      rust: 1.13.0
+      rust: 1.20.0
       os: osx
     - env: TARGET=arm-unknown-linux-gnueabi
-      rust: 1.13.0
+      rust: 1.20.0
     - env: TARGET=arm-unknown-linux-musleabi DISABLE_TESTS=1
-      rust: 1.14.0
+      rust: 1.20.0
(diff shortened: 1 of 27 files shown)


Susurrus:
This should probably also include a CHANGELOG entry since it does change how constants are scoped.

Susurrus:
This makes Rust 1.20 our base requirement. I've updated Travis, but the buildbots will need to be updated separately for this to pass there.

On the 23rd Rust 1.22 will be released, so once that's done requiring this newer version of Rust will fall in line with our N-2 minimum supported Rust standard.

Susurrus:
@asomers With this change we're now seeing epoll failures during testing as the only error (besides your FreeBSD buildbots not being Rust 1.20). Any idea what could be causing this?

asomers:
@Susurrus this is due to the Rust upgrade.  I saw this months ago in the Beta channel.  At first I suspected a code generation bug in Rust itself, but now I'm not so sure.  It could be that something in nix's epoll code is overrunning a buffer or something.  I never did get to the bottom of it.  But this patch makes the problem a little more obvious.
https://gist.github.com/asomers/198774d364042e85ffcbdb7eaeaa256f

Susurrus:
Hmmm, maybe we should test on stable in CI instead of beta. I didn't realize we weren't doing that at all anymore.

Susurrus:
Blocking on #805.

Susurrus:
With #805 merged, this is just waiting on an update of the FreeBSD buildbots. @asomers when would you be able to update those?

asomers:
@Susurrus The amd64 buildbot should be good to go.  But I'm out of space on my VM :( so it will take a little while longer to upgrade the i386 buildbot.

Susurrus:
@asomers Great, thanks! I've restarted this on the amd64 one and we'll see what happens. As for the i386 I guess we'll need to block on that for this to do things "the right way". What would be a rough ETA on that?

Susurrus:
Bam! It worked, so I'm betting i386 will work as well.

The downside to upgrading the buildbot outside of this PR is now other PRs will be testing on 1.20 there. So I'd love to get this merged ASAP to keep things in phase.

asomers:
The buildbots should all be working now.  I've restarted the failed i386 build.

PR #802, opened 2017-11-27
WIP: Fix epoll bug

### LINKED ISSUE #793: Resolve CI failures for beta-x86_64-unknown-linux-gnu
I've just run `cargo test` using both the `beta` and `nightly` toolchains, and neither failed. I'm guessing this is a CI failure caused by caching. With beta failing, it's not doing us any good because we always ignore it. I suggest we work to resolve this failure and change our CI tests so that failures running the beta toolchain will fail the whole build.

Susurrus:
Applies epoll debugging code from [asomers](https://gist.githubusercontent.com/asomers/198774d364042e85ffcbdb7eaeaa256f/raw/a499d23d684a1dc1d54fad7c340833389b2c3780/epoll_bug.diff) to see what failures are like in CI since they can't be reproduced locally. Hoping to get to the bottom of them.

Closes #793.


Susurrus:
Closing this as it triggers no failures. Testing on this will continue in #801.

PR #806, opened 2017-12-02
Cleanup all doc warnings

Susurrus:
With the impending switch to Pulldown as the default doc generator, warnings
have been enabled for incompatible syntax. This fixes all of said warnings.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1124,12 +1124,24 @@ pub fn getgroups() -> Result<Vec<Gid>> {
 /// specific user and group. For example, given the user `www-data` with UID
 /// `33` and the group `backup` with the GID `34`, one could switch the user as
 /// follows:
-/// ```
+///
+/// ```rust,no_run
+/// # use std::error::Error;
+/// # use nix::unistd::*;
+/// #
+/// # fn try_main() -> Result<(), Box<Error>> {
 /// let uid = Uid::from_raw(33);
 /// let gid = Gid::from_raw(34);
 /// setgroups(&[gid])?;
 /// setgid(gid)?;
 /// setuid(uid)?;
+/// #
+/// #     Ok(())
+/// # }
+/// #
+/// # fn main() {
+/// #     try_main().unwrap();
+/// # }
 /// ```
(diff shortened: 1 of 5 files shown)


asomers:
What switch to Pulldown?  Is this a new feature of rustc?

Susurrus:
rustdoc is switching from whatever it's using now to pulldown-cmark. This is good because it moves us to use CommonMark, which is standardized, than whatever the previous thing was doing. And it's a Rust-only lib. But there are some syntax changes. There's been warnings for this on nightly for a while and I believe Rust 1.22 includes these warnings as well.

Susurrus:
@asomers For reference here's the tracking bug: rust-lang/rust#38400

And as of 1.18, you can enable the new rendering with

Susurrus:
@asomers want to give this a review or are you fine with me merging it as-is?

src/unistd.rs
```diff
@@ -1124,12 +1125,24 @@ pub fn getgroups() -> Result<Vec<Gid>> {
 /// specific user and group. For example, given the user `www-data` with UID
 /// `33` and the group `backup` with the GID `34`, one could switch the user as
 /// follows:
-/// ```
+///
+/// ```rust,no_run
```
asomers: Why are you adding `no_run` ?  If this was an unreliable doc test, then you should fix it in a separate commit than the commit that fixes the doc warnings.

Susurrus: It was never run before, that's the biggest part of this change. There were doctests that weren't being found and executed, so this doctest has never been run. And turns out it fails when run.

Susurrus: I can put both of these changes in a separate commit that explains all this.

asomers: Do you understand why they weren't being found?  Because I don't.

Susurrus: The previous parser didn't find them, so they weren't run. I was always curious why, but I never cared enough to actually read up on it. If you ran our tests right now with the latest Rust, it'd explain this in the compiler warnings, though not give too many specifics.

asomers:
Ok, perhaps a parser bug or something.  Probably not worth investigating further.

bors r+

PR #807, opened 2017-12-03
Switch to libc definitons of errno constants

### LINKED ISSUE #487: `#[derive(PartialEq, Eq)]` for `nix::Errno`?
```rust
match nix::unistd::read(*self.stdio.stdin.get_ref().0, self.buf) {
    Err(nix::Error::Sys(nix::errno::EWOULDBLOCK)) => {
    ...
    },
    ...
}
```

produces:

```
src/lib.rs|84 col 37 warning| to use a constant of type `nix::Errno` in a pattern, `nix::Errno` must be annotated with `#[derive(PartialEq, Eq)]`, #[warn(illegal_struct_or_e
num_constant_pattern)] on by default                                                                                                                                         
|| this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!                                                    
|| for more information, see RFC 1445 <https://github.com/rust-lang/rfcs/pull/1445>  
```

I'm not exactly sure if I do the matching the best way.

Susurrus:
Still need to do OpenBSD and NetBSD. Primary goal with this first pass is to see what doesn't exist properly in `libc`.

Closes #487.


Susurrus:
I think with this change all the constant checks done at the bottom of `src/errno.rs` can be safely removed also. Unless there's an objection, once I get all platforms passing with this change I'll add that as another commit as well.

Susurrus:
This should be ready to merge once the FreeBSD tests pass. Note that there are more improvements that could be made to the errno code, but I wanted to restrict this PR to just using libc for constants and removing any unnecessary testing for it.

Susurrus:
Note that the check of the size of `IoVec<&[u8]>` matching `iovec` is the only check that *might* be worth keeping (and is removed by this PR), but I'm worried in how fragile our `IoVec` API is if it relies on the compiler making these the same size. Probably worth a glance at that code to make sure we aren't doing anything too onerous there.

PR #808, opened 2017-12-03
include LICENSE to nix-test

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


asomers:
I don't understand.  What is the purpose of this commit?  What cares about that file?

ignatenkobrain:
@asomers nix-test is dependency of nix (dev one) hence it is published on crates.io. And MIT license requires text license to be present along with source. So if you download `.crate` from crates.io for nix-test it doesn't have license text which is violating license terms.

Thanks for understanding.

Susurrus:
@ignatenkobrain Actually nix-test isn't regularly published to `crates.io`. I see there's a version up there, but it should actually be removed.

ignatenkobrain:
@Susurrus I don't think that crates with missing dependencies can be published on crates.io

Susurrus:
I'm pretty certain we don't need this dependency published on crates.io. Therefore instead of updating this, we should actually pull that project.

Susurrus:
@ignatenkobrain Thanks for bringing this to our attention, but rather than accept this patch, I'm going to close this in favor of #807, which will be removing this crate in its entirety as there will be no more reason for its existance once that's merged. I've also contacted Carl Lerche via email about pulling all versions of that crate and removing the project in its entirety.

PR #811, opened 2017-12-05
impl `std::fmt::Display` for `InterfaceFlags`

LuoZijun:


Susurrus:
@LuoZijun Can you provide a usecase here for this code. You didn't offer a description, so I'm uncertain why you submitted this. We almost never implement `Display` because it's for user-facing text. And since `nix` is a library, we don't have direct users. I'm inclined to close this, but I'll wait for your response.

LuoZijun:
okay, i think you are right. i will close this PR.

PR #812, opened 2017-12-05
Impl  `nix::ifaddrs::InterfaceAddress`

LuoZijun:
Impl `fmt::Debug` and `fmt::Display` for `InterfaceAddress` .

Note: need merge https://github.com/nix-rust/nix/pull/811 first.

### DIFF
--- a/src/ifaddrs.rs
+++ b/src/ifaddrs.rs
@@ -32,9 +32,30 @@ pub struct InterfaceAddress {
     pub destination: Option<SockAddr>,
 }
 
+// NOTE: see https://github.com/nix-rust/nix/issues/810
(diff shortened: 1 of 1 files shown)


Susurrus:
Mirroring what I said in #811, we don't implement `Display` for any types because it's not appropriate. There is no "user output" that makes sense for these types.

As to the `Debug` for InterfaceAddress, it would be nice to improve the output, yes. But I believe the correct solution is to implement `Debug` for the underlying types and then derive it for all the higher-level ones.

LuoZijun:
@Susurrus okay, i will make an new PR(just impl `Debug`).

PR #814, opened 2017-12-06
Misc cleanup

Susurrus:
Add more traits to the various datatypes in `nix`. Also try to utilize more `libc` types where appropriate.

This is still WIP because:
  * Looking for feedback on ba2d85bc. Does it make sense for bitflags structs to have `Ord` & `PartialOrd` derived for them?
  * Need to implement a newtype wrapper around `libc::linger` with both a `new()` constructor and getters. Additionally this needs manual implementations of `Eq`, `PartialEq`, `Debug` (I've been skipping manually implementing `Hash` for now.
  * In 2b3fb11a46 I need to implement newtype wrappers around `ip_mreq` and `ipv6_mreq` that have `new()` constructors and field getters. Additionally they need manual implementations of `Eq`, `PartialEq`, `Debug` (I've been skipping manually implementing `Hash` for now. This commit also needs a CHANGELOG entry.
  * ed79cfbc needs a CHANGELOG entry because its variants names have changed with the switch to using `libc_bitflags!`.


Susurrus:
For example, I'm not certain if we should be using `ucred` on FreeBSD or should be using `xucred` instead.

Susurrus:
For the BSDs, it's looks like NetBSD uses the `sockcred` struct, which is missing from `libc` and FreeBSD and DragonFlyBSD uses the `cmsgcred` struct, also missing from libc. I don't believe OpenBSD has this interface as `SCM_CREDS` is missing from their `socket.h`.

Next steps are to add those structs to libc. I'm not certain how I want to abstract this, if I want to hide all these structs behind the new `UserCredentials` datatype with appropriate accessor methods or have separate types for each different platform. I'm leaning towards abstrating it all away behind `UserCredentials`, but we'll see how the implementation goes.

asomers:
@Susurrus The answer lies [here](https://www.freebsd.org/cgi/man.cgi?query=unix&apropos=0&sektion=0&manpath=FreeBSD+11.1-RELEASE+and+Ports&arch=default&format=html).  Despite the doc comments, it looks like Nix is using `UnixCredentials` with Linux's `SO_PEERCRED` socket option.  The analogous socket option on FreeBSD is `LOCAL_PEERCRED`, which returns a `struct xucred`.  `struct xucred`, however, has different fields than Linux's `struct ucred`, so we would need to conditionally compile the various accessors.  Linux also has an `SCM_CREDENTIALS` message type for use with `sendmsg`.  FreeBSD uses a `struct cmsgcred` for that.  However, Nix doesn't seem to support that yet.

src/sys/socket/mod.rs
```diff
@@ -173,6 +170,47 @@ libc_bitflags!{
     }
 }
 
+/// Unix credentials of the sending process.
+///
+/// This struct is used with the `SCM_CREDENTIALS` ancillary message for UNIX sockets.
```
asomers: This message, while true, is red herring because Nix doesn't support `SCM_CREDENTIALS`.  It should read "This struct is used with the `SO_PEERCRED`/`LOCAL_PEERCRED` socket options for UNIX sockets.

src/sys/socket/mod.rs
```diff
@@ -173,6 +170,47 @@ libc_bitflags!{
...
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct UnixCredentials(libc::ucred);
```
asomers: should be `libc::xucred` on FreeBSD.

src/sys/socket/sockopt.rs
```diff
@@ -151,7 +152,7 @@ sockopt_impl!(Both, OobInline, libc::SOL_SOCKET, libc::SO_OOBINLINE, bool);
 sockopt_impl!(GetOnly, SocketError, libc::SOL_SOCKET, libc::SO_ERROR, i32);
 sockopt_impl!(Both, KeepAlive, libc::SOL_SOCKET, libc::SO_KEEPALIVE, bool);
 #[cfg(all(target_os = "linux", not(target_arch="arm")))]
-sockopt_impl!(GetOnly, PeerCredentials, libc::SOL_SOCKET, libc::SO_PEERCRED, super::ucred);
+sockopt_impl!(GetOnly, PeerCredentials, libc::SOL_SOCKET, libc::SO_PEERCRED, super::UnixCredentials);
```
asomers: On FreeBSD, this should be `LOCAL_PEERCRED` instead of `SO_PEERCRED`.

src/sys/socket/mod.rs
Susurrus: This is a very different looking struct, I'm really surprised that this is correct. I would have thought it'd be the struct I mentioned above, `cmsgcred`, which does have pid, uid, and gid fields.

asomers: Except there are two problems.  If `UnixCredentials` were actually used with `SCM_CREDENTIALS`, then you would be correct.  `struct cmsgcred` would be the struct to wrap.  However, the `SCM_CREDENTIALS` comment is red herring.  `UnixCredentials` is actually used as a socket option, with `SO_PEERCRED`.  That's why I say that it should wrap `xucred`.

Susurrus: From [this SO question](https://stackoverflow.com/questions/10037086/so-peercred-vs-scm-credentials-why-there-are-both-of-them#11213530) it seems like `SCM_CREDENTIALS` should be what we offer as an API and we abandon the `SO_PEERCRED` implementation, since one looks to be a newer/better version of the other.

Susurrus:
Note that all Linux MIPS build jobs will fail until rust-lang/libc#867 is merged. Once that's merged, this can go in.

I'm going to back off expanding APIs to any other platforms as it's just too much work. I'd really like to get all this `libc` cleanup stuff wrapper up, and it's just taking way too long! I keep finding new constants or types we have squirreled away that should be in `libc` instead. Hopefully this gets us pretty damn close to the end of it.

Susurrus:
Should go green once this runs through. @asomers Let me know if you have any feedback here, hopefully these changes are uncontroversial at this point.

Susurrus:
@asomers Can I get a final check with you on this before I merge it?

CHANGELOG.md
```diff
@@ -83,6 +85,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - The `ucred` struct has been removed in favor of a `UserCredentials` struct that
   contains only getters for its fields.
   ([#814](https://github.com/nix-rust/nix/pull/814))
+- Both `ip_mreq` and `ipv6_mreq` have been replaced with `IpMembershipRequest` and
```
asomers: This bit should be squashed with a9022b5

CHANGELOG.md
```diff
@@ -98,6 +103,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#747](https://github.com/nix-rust/nix/pull/747))
 - The `Errno` variants are no longer reexported from the `errno` module. `Errno` itself is no longer reexported from the
   crate root and instead must be accessed using the `errno` module. ([#696](https://github.com/nix-rust/nix/pull/696))
+- Removed `MS_VERBOSE`, `MS_NOSEC`, and `MS_BORN` from `MsFlags`. These
```
asomers: This part should be squashed with 5837c6d

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added the `from_raw()` method to `WaitStatus` for converting raw status values
   to `WaitStatus` independent of syscalls.
   ([#741](https://github.com/nix-rust/nix/pull/741))
+- Added more standard trait implementations for various types.
```
asomers: This bit should be squashed with e59c72b and 310d205

CHANGELOG.md
```diff
@@ -75,11 +77,17 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
-- Moved constants ptrace request, event and options to enums and updated ptrace functions and argument types accordingly. 
+- Moved constants ptrace request, event and options to enums and updated ptrace functions and argument types accordingly.
   ([#749](https://github.com/nix-rust/nix/pull/749))
 - `AioCb::Drop` will now panic if the `AioCb` is still in-progress ([#715](https://github.com/nix-rust/nix/pull/715))
 - Restricted `nix::sys::socket::UnixAddr::new_abstract` to Linux and Android only.
   ([#785](https://github.com/nix-rust/nix/pull/785))
+- The `ucred` struct has been removed in favor of a `UserCredentials` struct that
+  contains only getters for its fields.
+  ([#814](https://github.com/nix-rust/nix/pull/814))
+- Both `ip_mreq` and `ipv6_mreq` have been replaced with `IpMembershipRequest` and
```
asomers: This bit should be squashed wit a9022b5

CHANGELOG.md
```diff
@@ -75,11 +77,17 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#731](https://github.com/nix-rust/nix/pull/731))
 - Marked `pty::ptsname` function as `unsafe`
   ([#744](https://github.com/nix-rust/nix/pull/744))
-- Moved constants ptrace request, event and options to enums and updated ptrace functions and argument types accordingly. 
+- Moved constants ptrace request, event and options to enums and updated ptrace functions and argument types accordingly.
   ([#749](https://github.com/nix-rust/nix/pull/749))
 - `AioCb::Drop` will now panic if the `AioCb` is still in-progress ([#715](https://github.com/nix-rust/nix/pull/715))
 - Restricted `nix::sys::socket::UnixAddr::new_abstract` to Linux and Android only.
   ([#785](https://github.com/nix-rust/nix/pull/785))
+- The `ucred` struct has been removed in favor of a `UserCredentials` struct that
```
asomers: This bit should be squashed with c25f551

Susurrus:
You're right, I was just being lazy about it.

Bors r+ asomers

PR #815, opened 2017-12-08
Reduce ios builds

### LINKED ISSUE #669: TravisCI is too slow

Susurrus:
Closes #669

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,7 +100,7 @@ matrix:
 before_install: set -e
 
 install:
-  - sh ci/install.sh
+  - bash ci/install.sh
   - source ~/.cargo/env || true
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,24 @@ matrix:
   # likes to have a big backlog on builds on those machines. This way at least
   # all of the other jobs can finish while waiting on those builds.
   include:
+    # iOS builds
+    # These are all done on a single machine because Travis rations their OS X
+    # builders so heavily that we otherwise can't merge PRs during the work week.
+    # Additionally they're moved to the front of the line to get them in the Travis
+    # OS X build queue first.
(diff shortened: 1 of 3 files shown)


asomers:
Since we're not running tests on ios anyway, would it be possible to cross-build the ios targets from a Linux host?

Susurrus:
Yeah, I'll test it locally and we'll switch to that if so.

Susurrus:
```
$ rustup target install aarch64-apple-ios
$ cargo build --target aarch64-apple-ios
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names --target aarch64-apple-ios --crate-type bin --crate-type rlib` (exit code: 101)
--- stderr
error: Error loading target specification: Could not find specification for target "aarch64-apple-ios"
  |
  = help: Use `--print target-list` for a list of built-in targets
```

This is the error I get. I'd think cross-compiling would be possible, but I don't know what to do with this error. Any ideas?

asomers:
I think you need to use `--target=foo`, not `--target foo`

Susurrus:
That would've given me a different error. Both of those forms are equivalent.

Susurrus:
This should be pretty uncontroversial.

bors r+

asomers:
I get the same error.  It seems to me like rustup is installing std for aarch64-apple-ios, but not installing rustc.

Susurrus:
See rust-lang/rustup.rs#1303

PR #816, opened 2017-12-10
unistd: add {get,set}domainname

lucab:
This implements `getdomainname(2)` and `setdomainname(2)`, which are
Linux-specific extensions. References:
 * http://man7.org/linux/man-pages/man2/getdomainname.2.html
 * http://man7.org/linux/man-pages/man2/setdomainname.2.html


src/unistd.rs
```diff
@@ -756,6 +756,36 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
...
+
+/// Get the domain name and store it in the provided buffer, returning a pointer
+/// the CStr in that buffer on success (see
+/// [getdomainname(2)](http://man7.org/linux/man-pages/man2/getdomainname.2.html)).
+///
+/// This function call attempts to get the NIS domain name for the running system
+/// and store it in a provided buffer.  The buffer will be populated with bytes up
+/// to the length of the provided slice including a NUL terminating byte.
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn getdomainname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
```
lucab: I'm not super-happy about the ergonomics of this signature, but I just followed `gethostname` path. Can we rework these to allocate internally and return an owning `CString`?

src/unistd.rs
```diff
@@ -756,6 +756,36 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
...
+
+/// Get the domain name and store it in the provided buffer, returning a pointer
+/// the CStr in that buffer on success (see
+/// [getdomainname(2)](http://man7.org/linux/man-pages/man2/getdomainname.2.html)).
+///
+/// This function call attempts to get the NIS domain name for the running system
+/// and store it in a provided buffer.  The buffer will be populated with bytes up
+/// to the length of the provided slice including a NUL terminating byte.
+#[cfg(target_os = "linux")]
+pub fn getdomainname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
```
lucab: I'm not super-happy about the ergonomics of this signature, but I just followed gethostname path. Can we rework these to allocate internally and return an owning CString?

asomers: I agree.  I too would rather allocate internally.

src/unistd.rs
```diff
@@ -756,6 +756,36 @@ pub fn gethostname<'a>(buffer: &'a mut [u8]) -> Result<&'a CStr> {
     })
 }
 
+/// Set the NIS domain name (see
+/// [setdomainname(2)](http://man7.org/linux/man-pages/man2/setdomainname.2.html)).
+#[cfg(target_os = "linux")]
```
asomers: Both of these functions are available on FreeBSD too.  And while I haven't checked, I would bet that they're available on all of the BSDs based on their age.

Susurrus: They're also declared on emscripten, so I'd add both on those platforms as well.

test/test_unistd.rs
```diff
@@ -404,3 +404,22 @@ fn test_pipe2() {
     let f1 = FdFlag::from_bits_truncate(fcntl(fd1, FcntlArg::F_GETFD).unwrap());
     assert!(f1.contains(FdFlag::FD_CLOEXEC));
 }
+
+#[test]
+fn test_gethostname() {
```
Susurrus: Please put the addition into a separate commit before the rest of these changes.

src/unistd.rs
lucab: I did a quick check and indeed they seem to be available on all BSD (plus ES) too.

lucab: It looks like libc does not expose them, so https://github.com/rust-lang/libc/issues/868 needs to be fixed first.

lucab:
Rebased, PTAL. In this iteration:
 * split gethostname test to its own commit
 * changed gethostname and getdomainname to allocate their own buffer
 * added emscripten, reported https://github.com/rust-lang/libc/issues/868 for BSDs (I can tackle this later if nobody else does)

asomers:
libc's maintainers probably aren't going to add the definitions themselves, even for reported issues.  Like Nix, they mostly rely on pull requests from their users.  Could you please open a PR for libc?

lucab:
This got stale due to lack of time on my side, but I'll eventually get back to this (just not high priority at the moment). In the meanwhile, @gnzlbg is pushing forward the BSD libc fix, kudos to them.

lucab:
Rebased to take care of BSD and Apple, PTAL.

src/unistd.rs
```diff
@@ -728,34 +728,33 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
-/// let hostname = hostname_cstr.to_str().expect("Hostname wasn't valid UTF-8");
+/// let hostname_cstr = unistd::gethostname().expect("Failed getting hostname");
+/// let hostname = hostname_cstr.into_string().expect("Hostname wasn't valid UTF-8");
 /// println!("Hostname: {}", hostname);
 /// ```
-pub fn gethostname(buffer: &mut [u8]) -> Result<&CStr> {
-    let ptr = buffer.as_mut_ptr() as *mut c_char;
-    let len = buffer.len() as size_t;
+pub fn gethostname() -> Result<CString> {
+    const MAXHOSTNAMELEN: usize = 256;
```
Susurrus: We should document why we use the value we use here. Is there some reference example you pulled this from?

src/unistd.rs
```diff
@@ -728,34 +728,33 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+    let ptr = unsafe { CString::from_vec_unchecked(buf).into_raw() };
 
-    let res = unsafe { libc::gethostname(ptr, len) };
+    let res = unsafe { libc::gethostname(ptr, MAXHOSTNAMELEN) };
     Errno::result(res).map(|_| {
-        buffer[len - 1] = 0; // ensure always null-terminated
-        unsafe { CStr::from_ptr(buffer.as_ptr() as *const c_char) }
+        unsafe {
+            // Ensure returned string is always null-terminated
+            *ptr.offset((MAXHOSTNAMELEN - 1) as isize) = 0;
```
Susurrus: Couldn't you instead use a `MAXHOSTNAMELEN` of `255` here and be able to skip this step?

src/unistd.rs
lucab: This is `_POSIX_HOST_NAME_MAX` plus the trailing null, from http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html.

I'll rename, add a reference, and leave the trailing byte out of this.

src/unistd.rs
lucab: Yes, that's probably better, I'll do.

lucab:
@Susurrus I adjusted and documented `POSIX_HOST_NAME_MAX` as suggested, PTAL.

lucab:
Gentle ping for a review.

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+/// let hostname = hostname_cstr.into_string().expect("Hostname wasn't valid UTF-8");
 /// println!("Hostname: {}", hostname);
 /// ```
-pub fn gethostname(buffer: &mut [u8]) -> Result<&CStr> {
-    let ptr = buffer.as_mut_ptr() as *mut c_char;
-    let len = buffer.len() as size_t;
+pub fn gethostname() -> Result<CString> {
+    // Hostname maximum length as defined by POSIX,
+    // http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html
+    const POSIX_HOST_NAME_MAX: usize = 255;
```
Susurrus: According to this link the max host name length is 64 on Linux, and that's confirmed on my system.  `_POSIX_HOST_NAME_MAX` on my system is 255, however. We should figure out the appropriate values here and get them into `libc` to use them here.

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+
+/// Get the NIS domain name, returning a `CString` on success (see
+/// [getdomainname(2)](http://man7.org/linux/man-pages/man2/getdomainname.2.html)).
+///
+/// This function call attempts to get the NIS domain name for the running system
+/// and return it as a NUL-terminated string.
+///
+/// # Examples
+///
+/// ```no_run
```
Susurrus: Can this not run in CI?

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+                 target_os = "ios", target_os = "macos"))] {
+        type namelen_t = c_int;
+    } else if #[cfg(not(target_os = "android"))] {
+        type namelen_t = size_t;
+    }
+}
+
+/// Set the NIS domain name (see
+/// [setdomainname(2)](http://man7.org/linux/man-pages/man2/setdomainname.2.html)).
+#[cfg(not(target_os = "android"))]
```
Susurrus: We prefer to use whitelists instead of blacklists for listing targets. Please do so here and for `getdomainname`.

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+    // http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html
+    const POSIX_HOST_NAME_MAX: usize = 255;
+    let buf = vec![0; POSIX_HOST_NAME_MAX + 1];
+    let ptr = unsafe { CString::from_vec_unchecked(buf).into_raw() };
+
+    let res = unsafe { libc::gethostname(ptr, POSIX_HOST_NAME_MAX + 1) };
+    Errno::result(res).map(|_| {
+        unsafe {
+            // Ensure returned string is always null-terminated
+            *ptr.offset(POSIX_HOST_NAME_MAX as isize) = 0;
```
Susurrus: You shouldn't need to do this because `gethostname` won't overrun the buffer. Same for the other functions.

test/test_unistd.rs
```diff
@@ -400,3 +400,20 @@ fn test_pipe2() {
...
+    assert!(hn.is_ok(), "Error: {:?}", hn.unwrap_err());
+    let hn_string = hn.unwrap().into_string();
+    assert!(hn_string.is_ok(), "Error: {:?}", hn_string.unwrap_err());
+}
+
+#[test]
+#[cfg(not(target_os = "android"))]
+fn test_getdomainname() {
+    let dn = getdomainname();
+    assert!(dn.is_ok(), "Error: {:?}", dn.unwrap_err());
```
Susurrus: Replace this with `dn.expect("")` on the above line.

test/test_unistd.rs
```diff
@@ -400,3 +400,20 @@ fn test_pipe2() {
     let f1 = FdFlag::from_bits_truncate(fcntl(fd1, FcntlArg::F_GETFD).unwrap());
     assert!(f1.contains(FdFlag::FD_CLOEXEC));
 }
+
+#[test]
+fn test_gethostname() {
+    let hn = gethostname();
+    assert!(hn.is_ok(), "Error: {:?}", hn.unwrap_err());
```
Susurrus: Replace this with `hn.expect("")` on the above line.

CHANGELOG.md
```diff
@@ -6,10 +6,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Added `nix::unistd::{getdomainname, setdomainname}`.
```
Susurrus: Please specify which platforms this applies to.

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+        type namelen_t = size_t;
+    }
+}
+
+/// Set the NIS domain name (see
+/// [setdomainname(2)](http://man7.org/linux/man-pages/man2/setdomainname.2.html)).
+#[cfg(not(target_os = "android"))]
+pub fn setdomainname<S: AsRef<OsStr>>(name: S) -> Result<()> {
+    let ptr = name.as_ref().as_bytes().as_ptr() as *const c_char;
+    let len = name.as_ref().len() as namelen_t;
```
asomers: I think it would probably be more pedantically correct to do `len = name.as_ref().as_bytes().len()`.

src/unistd.rs
asomers: Why use a `CString` here but an `OsStr` for `setdomainname`?  We should be consistent in our choice of C vs OS strings.

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+/// ```no_run
+/// use nix::unistd;
+///
+/// let domainname_cstr = unistd::getdomainname().expect("Failed getting domain name");
+/// let domainname = domainname_cstr.into_string().expect("Domain name wasn't valid UTF-8");
+/// println!("Domain name: {}", domainname);
+/// ```
+#[cfg(not(target_os = "android"))]
+pub fn getdomainname() -> Result<CString> {
+    const MAXDOMAINNAMELEN: usize = 256;
```
asomers: According to the man page, `MAXHOSTNAMELEN` applies to the domain name, too.

Susurrus:
@lucab Are you planning on coming back to this?

lucab:
@Susurrus yes but I need to figure out the story with `HOST_NAME_MAX` Linux and POSIX, and forward to libc.

Susurrus:
@lucab Did the `HOST_NAME_MAX` get sorted yet?

asomers:
Yes, sysconf sounds like the safest approach.

src/unistd.rs
lucab: @asomers This follows the split currently in master for the hostname, see `gethostname` vs `sethostname` types. Perhaps it can be left out of this PR? Otherwise, in which direction do you want to resolve this?

lucab:
Travis is red due to android missing POSIX constants in libc crate, see https://github.com/rust-lang/libc/pull/990.

lucab:
Rebased, this is now green. I think I've addressed all comments in reviews so far, but at this point this PR is quite different from its initial version. PTAL.

src/unistd.rs
```diff
@@ -739,34 +739,97 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+pub fn gethostname() -> Result<OsString> {
+    // Minimum hostname maximum length as defined by POSIX,
+    // http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html
+    const _POSIX_HOST_NAME_MAX: c_long = 255;
+    let buf_len = sysconf(SysconfVar::HOST_NAME_MAX)
+        .unwrap_or(None).unwrap_or(_POSIX_HOST_NAME_MAX + 1);
+    let buf = vec![0; buf_len as usize];
+    let ptr = unsafe { CString::from_vec_unchecked(buf).into_raw() };
+
+    let res = unsafe { libc::gethostname(ptr, buf_len as usize) };
```
asomers: You don't need to convert the buffer to a `CString` first.  You can just do `libc::gethostname(buf.as_mut() as *mut c_char, buf_len as usize`

lucab:
@asomers I was a bit unsure about the safety of your comments so I added some idempotency checks to the tests. You are indeed right, so I pushed the rework and the additional testing logic. Everything is green, PTAL.

asomers:
Ok, everything looks good to me!  It only needs a squash.  @Susurrus do you have any comments?

lucab:
@asomers ack. I squashed one commit out. I'm keeping two commits for the two separate changelog entries. Let me know if you prefer everything together.

src/unistd.rs
```diff
@@ -769,6 +769,60 @@ pub fn gethostname() -> Result<OsString> {
...
+/// Get the NIS domain name, returning a `OsString` on success (see
+/// [getdomainname(2)](http://man7.org/linux/man-pages/man2/getdomainname.2.html)).
+///
+/// This function call attempts to get the NIS domain name for the
+/// running system.
+///
+/// # Examples
+///
+/// ```
+/// use nix::unistd;
```
Susurrus: Please prefix this line with a `#` so they don't render in the final help. You can then also delete the blank line below.

src/unistd.rs
```diff
@@ -769,6 +769,60 @@ pub fn gethostname() -> Result<OsString> {
...
+///
+/// let domainname_os = unistd::getdomainname().expect("Failed getting domain name");
```
Susurrus: You can reuse the name `domainname` here because of Rust's variable shadowing. Doing this is idiomatic. Please rename this variable to `domainname`.

src/unistd.rs
```diff
@@ -769,6 +769,60 @@ pub fn gethostname() -> Result<OsString> {
...
+/// let domainname = domainname_os.into_string().expect("Domain name wasn't valid UTF-8");
+/// println!("Domain name: {}", domainname);
+/// ```
+#[cfg(any(target_os = "dragonfly", target_os = "freebsd", target_os = "ios",
+          target_os = "linux", target_os = "macos", target_os = "netbsd",
+          target_os = "openbsd"))]
+pub fn getdomainname() -> Result<OsString> {
+    // Minimum hostname maximum length as defined by POSIX,
+    // http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html
+    const _POSIX_HOST_NAME_MAX: c_long = 255;
```
Susurrus: Why isn't this constant pulled from `libc`? I didn't see anywhere in the discussion a comment touching on this.

src/unistd.rs
```diff
@@ -728,34 +728,85 @@ pub fn sethostname<S: AsRef<OsStr>>(name: S) -> Result<()> {
...
+    let ptr = unsafe { CString::from_vec_unchecked(buf).into_raw() };
+
+    let res = unsafe { libc::gethostname(ptr, POSIX_HOST_NAME_MAX + 1) };
+    Errno::result(res).map(|_| {
+        unsafe {
+            // Ensure returned string is always null-terminated
+            *ptr.offset(POSIX_HOST_NAME_MAX as isize) = 0;
+            CString::from_raw(ptr)
+        }
+    })
```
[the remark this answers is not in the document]
Susurrus: @asomers Does this comment still stand or is it not relevant anymore?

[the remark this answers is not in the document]
asomers: Unfortunately GitHub has forgotten the context of that comment.  But I think my objection was because he tried to handle that error condition and didn't do it correctly.  Now, he isn't handling any error conditions, which is better.

CrumblyLiquid:
I came looking exactly for this! I'm not very knowledgeable about POSIX but I'm happy to help this get merged if needed. Are there any major blocks that need to be resolved?

PR #817, opened 2017-12-11
Implement nix wrapper for libc::signal

### LINKED ISSUE #476: No libc::signal
There doesn't seem to be any function in nix that calls `libc::signal` but there is a signal module.  Is this deliberate or has it just not been added yet.  If the latter than should I add it to `nix::sys::signal` ?

rgardner:
This implements `nix::sys::signal::signal` and adds corresponding documentation and tests.

Closes #476

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -542,6 +542,79 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
     Errno::result(res).map(|_| SigAction { sigaction: oldact })
 }
 
+/// Signal management (see [signal(3p)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/signal.html))
+///
+/// Installs `handler` for the given `signal`, returning the previous signal
+/// handler. `signal` should only be used following another call to `signal` or
+/// if the current handler is the default. The return value of `signal` is
+/// undefined after setting the handler with [`sigaction`][SigActionFn].
+///
+/// # Safety
+///
+/// If the pointer to the previous signal handler is invalid, undefined
+/// behavior could be invoked when casting it back to a [`SigAction`][SigActionStruct].
+///
+/// # Examples
+///
+/// Ignore `SIGINT`:
+///
+/// ```no_run
+/// # use nix::sys::signal::{self, Signal, SigHandler};
+/// unsafe { signal::signal(Signal::SIGINT, SigHandler::SigIgn) }.unwrap();
+/// ```
+///
+/// Use a signal handler to set a flag variable:
+///
+/// ```no_run
+/// # #[macro_use] extern crate lazy_static;
+/// # extern crate libc;
+/// # extern crate nix;
+/// # use std::sync::atomic::{AtomicBool, Ordering};
+/// # use nix::sys::signal::{self, Signal, SigHandler};
+/// lazy_static! {
+///    static ref SIGNALED: AtomicBool = AtomicBool::new(false);
+/// }
+///
+/// extern fn handle_sigint(signal: libc::c_int) {
+///     let signal = Signal::from_c_int(signal).unwrap();
+///     SIGNALED.store(signal == Signal::SIGINT, Ordering::Relaxed);
+/// }
+///
+/// fn main() {
+///     let handler = SigHandler::Handler(handle_sigint);
+///     unsafe { signal::signal(Signal::SIGINT, handler) }.unwrap();
+/// }
+/// ```
+///
+/// # Errors
+///
+/// Returns [`Error::UnsupportedOperation`] if `handler` is
+/// [`SigAction`][SigActionStruct]. Use [`sigaction`][SigActionFn] instead.
+///
+/// `signal` also returns any error from `libc::signal`, such as when an attempt
+/// is made to catch a signal that cannot be caught or to ignore a signal that
+/// cannot be ignored.
+///
+/// [`Error::UnsupportedOperation`]: ../../enum.Error.html#variant.UnsupportedOperation
+/// [SigActionStruct]: struct.SigAction.html
+/// [sigactionFn]: fn.sigaction.html
+pub unsafe fn signal(signal: Signal, handler: SigHandler) -> Result<SigHandler> {
+    let signal = signal as libc::c_int;
+    let res = match handler {
+        SigHandler::SigDfl => libc::signal(signal, libc::SIG_DFL),
+        SigHandler::SigIgn => libc::signal(signal, libc::SIG_IGN),
+        SigHandler::Handler(handler) => libc::signal(signal, handler as libc::sighandler_t),
+        SigHandler::SigAction(_) => return Err(Error::UnsupportedOperation),
+    };
+    Errno::result(res).map(|oldhandler| {
+        match oldhandler {
+            libc::SIG_DFL => SigHandler::SigDfl,
+            libc::SIG_IGN => SigHandler::SigIgn,
+            f => SigHandler::Handler(mem::transmute(f)),
+        }
+    })
+}
+
 /// Manages the signal mask (set of blocked signals) for the calling thread.
 ///
 /// If the `set` parameter is `Some(..)`, then the signal mask will be updated with the signal set.
--- a/test/sys/test_signal.rs
+++ b/test/sys/test_signal.rs
@@ -1,5 +1,8 @@
-use nix::unistd::*;
+use libc;
+use nix::Error;
 use nix::sys::signal::*;
+use nix::unistd::*;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 #[test]
 fn test_kill_none() {
@@ -59,4 +62,37 @@ fn test_sigprocmask() {
     // Reset the signal.
     sigprocmask(SigmaskHow::SIG_UNBLOCK, Some(&signal_set), None)
         .expect("expect to be able to block signals");
+}
+
+lazy_static! {
+    static ref SIGNALED: AtomicBool = AtomicBool::new(false);
+}
+
+extern fn test_sigaction_handler(signal: libc::c_int) {
+    let signal = Signal::from_c_int(signal).unwrap();
+    SIGNALED.store(signal == Signal::SIGINT, Ordering::Relaxed);
+}
+
+extern fn test_sigaction_action(_: libc::c_int, _: *mut libc::siginfo_t, _: *mut libc::c_void) {
+}
+
+#[test]
+fn test_signal() {
+    let _m = ::SIGNAL_MTX.lock().expect("Mutex got poisoned by another test");
+
+    unsafe { signal(Signal::SIGINT, SigHandler::SigIgn) }.unwrap();
+    raise(Signal::SIGINT).unwrap();
+    assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(), SigHandler::SigIgn);
+
+    let handler = SigHandler::Handler(test_sigaction_handler);
+    assert_eq!(unsafe { signal(Signal::SIGINT, handler) }.unwrap(), SigHandler::SigDfl);
+    raise(Signal::SIGINT).unwrap();
+    assert!(SIGNALED.load(Ordering::Relaxed));
+    assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(), handler);
+
+    let action_handler = SigHandler::SigAction(test_sigaction_action);
+    assert_eq!(unsafe { signal(Signal::SIGINT, action_handler) }.unwrap_err(), Error::UnsupportedOperation);
+
+    // Restore default signal handler
+    unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap();
 }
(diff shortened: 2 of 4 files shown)


src/sys/signal.rs
```diff
@@ -410,6 +410,49 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+///     let result = signal::signal(Signal::SIGINT, action_handler);
+///     assert_eq!(result.unwrap_err(), nix::Error::UnsupportedOperation);
+/// }
+/// ```
+pub unsafe fn signal(signal: Signal, handler: SigHandler) -> Result<SigHandler> {
+    let signal = signal as libc::c_int;
+    let res = match handler {
+        SigHandler::SigDfl => libc::signal(signal, libc::SIG_DFL),
+        SigHandler::SigIgn => libc::signal(signal, libc::SIG_IGN),
+        _ => return Err(Error::UnsupportedOperation),
```
asomers: Why not support arbitrary signal handlers?

src/sys/signal.rs
```diff
@@ -739,6 +782,15 @@ mod tests {
         assert_eq!(action_ign.handler(), SigHandler::SigIgn);
     }
 
+    #[test]
+    fn test_signal() {
+        unsafe {
```
asomers: To prevent races against other tests, please add to the top of this function:
```rust
let m = ::SIGNAL_MTX.lock().expect("Mutex got poisoned by another test");
```

rgardner: I added the mutex guard and moved the test to `test/sys/test_signal.rs` (given it's an integration test and to access `SIGNAL_MTX`)

src/sys/signal.rs
```diff
@@ -410,6 +410,51 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+        SigHandler::SigDfl => libc::signal(signal, libc::SIG_DFL),
+        SigHandler::SigIgn => libc::signal(signal, libc::SIG_IGN),
+        SigHandler::Handler(handler) => libc::signal(signal, handler as libc::sighandler_t),
+        SigHandler::SigAction(handler) => libc::signal(signal, handler as libc::sighandler_t),
+    };
+    Errno::result(res).map(|oldhandler| {
+        match oldhandler {
+            libc::SIG_DFL => Some(SigHandler::SigDfl),
+            libc::SIG_IGN => Some(SigHandler::SigIgn),
+            _ => None, // cannot safely convert custom handler to SigHandler
```
rgardner: any suggestions here? I don't think we can get the type information of the pointer to know if it was a `SigHandler::Handler` function or a `SigHandler::SigAction` function. Another option is to return `libc::sighandler_t` here, but my thinking was that callers who needed that level of detail could just work with the  raw `libc::signal` API instead.

src/sys/signal.rs
```diff
@@ -410,6 +410,51 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+///     assert!(signal(Signal::SIGSTOP, SigHandler::SigIgn).is_err());
+/// }
+/// ```
+pub unsafe fn signal(signal: Signal, handler: SigHandler) -> Result<Option<SigHandler>> {
+    let signal = signal as libc::c_int;
+    let res = match handler {
+        SigHandler::SigDfl => libc::signal(signal, libc::SIG_DFL),
+        SigHandler::SigIgn => libc::signal(signal, libc::SIG_IGN),
+        SigHandler::Handler(handler) => libc::signal(signal, handler as libc::sighandler_t),
+        SigHandler::SigAction(handler) => libc::signal(signal, handler as libc::sighandler_t),
```
asomers: Is it valid to pass a `sa_sigaction` function pointer to `signal(3)`?  It may work on some OSes, but the man page doesn't say anything about it.  I think we should force people to use `sigaction(2)` if they want to do that.

src/sys/signal.rs
asomers: I think we should always treat it like a `SigHandler::Handler`, because that should be the only type of handler that can be set with `signal(2)`.  If the program is setting the handler with `sigaction`, then it should also use `sigaction` to check the old handler.

src/sys/signal.rs
rgardner: you're totally right, signal(3) specifies `sighandler_t`

rgardner:
@asomers, could you review this again when you have some time? No rush - and happy holidays!

test/sys/test_signal.rs
```diff
@@ -1,7 +1,39 @@
...
+
+#[test]
+fn test_signal() {
+    let _m = ::SIGNAL_MTX.lock().expect("Mutex got poisoned by another test");
+    unsafe {
+        signal(Signal::SIGINT, SigHandler::SigIgn).unwrap();
+        raise(Signal::SIGINT).unwrap();
+        assert_eq!(signal(Signal::SIGINT, SigHandler::SigDfl).unwrap(), SigHandler::SigIgn);
+
+        static mut RECEIVED_SIGINT: bool = false;
```
asomers: `static mut` variables are `unsafe`.  The only reason the compiler allowed it here is because you put the entire test case into an `unsafe` block.  Why don't you use `unsafe` only where you need it?  I think the only reason that `signal` is `unsafe` is because of `mem::transmute`, right?

In any case, the safe way to do a static global is demonstrated by the `SIGNALED` variable in test/sys/test_aio.rs.

rgardner: Yes, the `mem::transmute` call + calling the `libc::signal` API. Should `signal` be marked safe?

asomers: `signal` should only be marked safe if you can guarantee that the `mem::transmute` operation will always be safe.  In this case, I don't think you can, because `f` might be an `sa_sigaction` function previously set by a call to `sigaction`.

asomers:
This looks good; it just needs a squash.

src/sys/signal.rs
```diff
@@ -410,6 +410,50 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+///
+/// assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigIgn) }.unwrap(), SigHandler::SigDfl);
+/// assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(), SigHandler::SigIgn);
+/// ```
+///
+/// # Errors
+/// 
+/// Returns [`Error::UnsupportedOperation`] if `handler` is
+/// [`SigAction`][SigActionStruct]. Use [`sigaction`][SigActionFn] instead.
+/// `signal` also returns any error from `libc::signal`. such as when an attempt
```
Susurrus: Can you put a blank line above this one? These should be separate paragraphs since they're distinct comments.

src/sys/signal.rs
```diff
@@ -416,6 +416,51 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+/// assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigIgn) }.unwrap(), SigHandler::SigDfl);
+/// assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(), SigHandler::SigIgn);
+/// ```
+///
+/// # Errors
+///
+/// Returns [`Error::UnsupportedOperation`] if `handler` is
+/// [`SigAction`][SigActionStruct]. Use [`sigaction`][SigActionFn] instead.
+///
+/// `signal` also returns any error from `libc::signal`. such as when an attempt
```
Susurrus: Let's change that period to a comma.

src/sys/signal.rs
```diff
@@ -416,6 +416,51 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+/// is made to catch a signal that cannot be caught or ignore a signal that
```
Susurrus: "or ignore" -> "or to ignore"

Susurrus:
I'm also not the biggest fan of the example since it's not a realistic one, it looks more like a test. Can you turn it into a more realistic example? Like something a user might actually write? If not, let's actually remove it. @asomers any ideas on a better example here?

src/sys/signal.rs
```diff
@@ -416,6 +416,51 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+/// cannot be ignored.
+///
+/// [`Error::UnsupportedOperation`]: ../../enum.Error.html#variant.UnsupportedOperation
+/// [SigActionStruct]: struct.SigAction.html
+/// [sigactionFn]: fn.sigaction.html
+pub unsafe fn signal(signal: Signal, handler: SigHandler) -> Result<SigHandler> {
```
asomers: Please add a section to the doc comments called "Safety" explaining why this function is unsafe and what the caller must do to use it safely.

Susurrus:
@rgardner You planning to come back to this? We'd like to get this merged if you're willing. Otherwise I'd like to close this PR.

rgardner:
Yes, I too would like to get this merged. I'm unsure of exactly what to write in the safety section, I mostly chose unsafe out of inertia with the existing `sigaction` function above. Of the four [unsafe superpowers](https://doc.rust-lang.org/book/second-edition/ch19-01-unsafe-rust.html#unsafe-superpowers) called out in the book, "calling an unsafe function or method" applies. The transmuting of the previous signal handler is unsafe as the previous signal handler could be invalid - I'll start with that.

I'm open to suggestions for the safety section. Otherwise, I don't think this should block the PR, someone can add it in a follow up.

Susurrus:
This needs the commits to be squashed to one commit, a rebase onto the latest master, and a re-run through CI to pass before it can be merged. Otherwise it looks good to me.

@asomers any comments on the Safety section?

src/sys/signal.rs
```diff
@@ -416,6 +416,78 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+/// undefined after setting the handler with [`sigaction`][SigActionFn].
+///
+/// # Safety
+///
+/// If the pointer to the previous signal handler is invalid, undefined
+/// behavior could be invoked when casting it back to a [`SigAction`][SigActionStruct].
+///
+/// # Examples
+///
+/// Ignore SIGINT:
```
asomers: `SIGINT` should be in backticks.

src/sys/signal.rs
```diff
@@ -416,6 +416,78 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
...
+///
+/// ```no_run
+/// use nix::sys::signal::{self, Signal, SigHandler};
+/// unsafe { signal::signal(Signal::SIGINT, SigHandler::SigIgn) }.unwrap();
+/// ```
+///
+/// Use a signal handler to set a flag variable:
+///
+/// ```no_run
+/// #[macro_use]
```
asomers: In the examples, you can prefix lines with "# " to suppress them from the documentation.  Typically, "extern crate", "use", and "fn main" lines are suppressed.

test/sys/test_signal.rs
```diff
@@ -61,3 +66,34 @@ fn test_sigprocmask() {
...
+    let _m = ::SIGNAL_MTX.lock().expect("Mutex got poisoned by another test");
+
+    unsafe { signal(Signal::SIGINT, SigHandler::SigIgn) }.unwrap();
+    raise(Signal::SIGINT).unwrap();
+    assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(), SigHandler::SigIgn);
+
+    let handler = SigHandler::Handler(test_sigaction_handler);
+    assert_eq!(unsafe { signal(Signal::SIGINT, handler) }.unwrap(), SigHandler::SigDfl);
+    raise(Signal::SIGINT).unwrap();
+    thread::sleep(Duration::from_millis(5));
```
asomers: The sleep shouldn't be necessary.  You can remove it.

test/sys/test_signal.rs
```diff
@@ -61,3 +66,34 @@ fn test_sigprocmask() {
...
+    assert!(SIGNALED.load(Ordering::Relaxed));
+    assert_eq!(unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(), handler);
+
+    let action_handler = SigHandler::SigAction(test_sigaction_action);
+    assert_eq!(unsafe { signal(Signal::SIGINT, action_handler) }.unwrap_err(), Error::UnsupportedOperation);
+}
```
asomers: Please restore the default handler at the end of the test.

rgardner:
@asomers - friendly ping, could you take another look at this PR when you get a chance? Thank you for all the feedback.

asomers:
Sorry that this fell off of my radar, @rgardner .  It LGTM now.  But it will need a rebase before it can be merged.  Be careful when rebasing CHANGELOG to get your entry in the right section.

rgardner:
No worries, fell of my radar too. I've rebased and ensured that the CHANGELOG entry is at the bottom of the Added section under [Unreleased].

PR #818, opened 2017-12-16
Clippy cleanup

Susurrus:
Various fixes for errors and warnings pointed out by clippy.

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -675,7 +675,6 @@ mod consts {
         use self::Errno::*;
 
         match e {
-            0 => UnknownErrno,
             libc::EPERM => EPERM,
             libc::ENOENT => ENOENT,
             libc::ESRCH => ESRCH,
@@ -940,7 +939,6 @@ mod consts {
         use self::Errno::*;
 
         match e {
-            0 => UnknownErrno,
             libc::EPERM => EPERM,
             libc::ENOENT => ENOENT,
             libc::ESRCH => ESRCH,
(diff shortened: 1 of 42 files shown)


Susurrus:
@asomers I think there's only one controversial change in this PR, 7d733d4b6c2f2cd0be864417c6a07f75362a9b78. Based on the comments I think these should have been `const` rather than standard `let` variables.

Susurrus:
@asomers I'd like to merge this, any chance to look over the commit I mentioned above?

asomers:
I should be able to get to it sometime tonight.

asomers:
I think the reason for the `'static` declarations that you removed in 7d733d4 was because originally those variables were `const` instead of `let`.  And as you say in one of your other commits, constants had to be explictly `'static` until 1.17.0.  I'm not sure why I removed the `const` before I committed.

asomers:
The commit message for 263ed2c is incorrect.  It should say "elide" or "remove implied" instead of "remove elided".

Susurrus:
@asomers The tests still seem to pass, do you think the tests are correct to begin with?

asomers:
Once I finish converting aio from using `Rc<[u8]>` to `Bytes`, a81dba7 will be completely irrelevant.  But it's fine to commit it anyway.

asomers:
Yes, the tests are correct.  They don't rely on `wbuf` being `const`.

asomers:
7f3ae6b and 5568128 are the same thing.  It would make sense to squash those two together.

asomers:
Apart from that one commit message, everything looks acceptable to me.  Just fix that, then bors +asomers .

Susurrus:
Awesome, thanks @asomers! I merged those two commits and gave that other one a better title.

Let's wait to merge this until #648 is merged as I really want to get that one in. It's the last of the FFI issues we have in `nix`, been a long time trying to knock all of those out.

PR #820, opened 2017-12-21
Change AioCb to primarily use Bytes instead of Rc<[u8]>

asomers:
`Rc<[u8]>` isn't a very good buffer type to use for aio.  For one thing, it lacks interior mutability.  For another, a single `Rc<[u8]>` can't be carved up into smaller buffers of the same type.  `Bytes` and `BytesMut` fix both problems.  This PR removes the ability to construct an `AioCb` from `Rc<[u8]>` and adds the ability to construct one from `Bytes`, `BytesMut`, or raw pointers (for consumers who need even more flexibility).  At this stage, the PR has the following warts:

1. A hack is necessary to force small `Bytes` buffers to allocate on the heap.  I plan to fix this with an enhancement to the bytes crate.
2. The `AioCb::buffer` method is necessary due to a deficiency in the tokio-core crate.  Once I fix that, then only `AioCb::into_buffer`will need to be public.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Added `AioCb::from_ptr` and `AioCb::from_mut_ptr`
+  ([#820](https://github.com/nix-rust/nix/pull/820))
 - Added specialized wrappers: `sys::ptrace::{traceme, syscall, cont, attach}`. Using the matching routines
   with `sys::ptrace::ptrace` is now deprecated.
 - Added `nix::poll` module for all platforms
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#648](https://github.com/nix-rust/nix/pull/648))
 
 ### Removed
+- `AioCb::from_boxed_slice` has been removed.  It was never actually safe.  Use
+  `from_bytes` or `from_bytes_mut` instead.
+  ([#820](https://github.com/nix-rust/nix/pull/820))
 - The syscall module has been removed. This only exposed enough functionality for
   `memfd_create()` and `pivot_root()`, which are still exposed as separate functions.
   ([#747](https://github.com/nix-rust/nix/pull/747))
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,11 @@ bitflags = "1.0"
 cfg-if = "0.1.0"
 void = "1.0.2"
 
+[dependencies.bytes]
+version = "0.4.5"
+# Don't include the optional serde feature
+default-features = false
+
 [target.'cfg(target_os = "dragonfly")'.build-dependencies]
 gcc = "0.3"
 
--- a/src/sys/aio.rs
+++ b/src/sys/aio.rs
@@ -178,6 +178,73 @@ impl<'a> AioCb<'a> {
         }
     }
 
+    /// Constructs a new `AioCb` from a mutable raw pointer
+    ///
+    /// * `fd`  File descriptor.  Required for all aio functions.
+    /// * `offs` File offset
+    /// * `buf` Pointer to the memory buffer
+    /// * `len` Length of the buffer pointed to by `buf`
+    /// * `prio` If POSIX Prioritized IO is supported, then the operation will
+    /// be prioritized at the process's priority level minus `prio`
+    /// * `sigev_notify` Determines how you will be notified of event
+    /// completion.
+    /// * `opcode` This field is only used for `lio_listio`.  It determines
+    /// which operation to use for this individual aiocb
+    ///
+    /// # Safety
+    ///
+    /// Unsafe because using this `AioCb` will cause `libc::aio_read` or
+    /// `libc::aio_write` to dereference a raw pointer, without type, bounds, or
+    /// lifetime checking.
+    pub unsafe fn from_mut_ptr(fd: RawFd, offs: off_t,
+                           buf: *mut c_void, len: usize,
+                           prio: libc::c_int, sigev_notify: SigevNotify,
+                           opcode: LioOpcode) -> AioCb<'a> {
+        let mut a = AioCb::common_init(fd, prio, sigev_notify);
+        a.aio_offset = offs;
+        a.aio_nbytes = len;
+        a.aio_buf = buf;
+        a.aio_lio_opcode = opcode as libc::c_int;
+
+        let aiocb = AioCb { aiocb: a, mutable: true, in_progress: false,
+                            keeper: Keeper::none};
+        aiocb
+    }
+
+    /// Constructs a new `AioCb` from a raw pointer
+    ///
+    /// * `fd`  File descriptor.  Required for all aio functions.
+    /// * `offs` File offset
+    /// * `buf` Pointer to the memory buffer
+    /// * `len` Length of the buffer pointed to by `buf`
+    /// * `prio` If POSIX Prioritized IO is supported, then the operation will
+    /// be prioritized at the process's priority level minus `prio`
+    /// * `sigev_notify` Determines how you will be notified of event
+    /// completion.
+    /// * `opcode` This field is only used for `lio_listio`.  It determines
+    /// which operation to use for this individual aiocb
+    ///
+    /// # Safety
+    ///
+    /// Unsafe because using this `AioCb` will cause `libc::aio_write` to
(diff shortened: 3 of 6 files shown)


src/sys/aio.rs
```diff
@@ -86,10 +118,21 @@ pub struct AioCb<'a> {
...
-    keeper: Keeper<'a>
+    buffer: Buffer<'a>
 }
 
 impl<'a> AioCb<'a> {
+    /// Remove the inner `Buffer` and return it
+    ///
+    /// It is an error to call this method while the `AioCb` is still in
+    /// progress.
+    pub fn buffer(&mut self) -> Buffer<'a> {
```
Susurrus: Does it make sense to make this return a `Result` type instead of having failing if the in progress invariant is broken?

Susurrus:
Any movement on this?

asomers:
Yeah, I'm 99% done modifying the downstream crates.  But I can't be confident that my new API will work until I'm 100% done.

Susurrus:
Got an ETA on this? We're at 6 months since our last release, and I think it's smart that we get this out the door instead of waiting. How would you feel about releasing 0.10 now?

src/sys/aio.rs
asomers: I think not.  If we do that, then `AioCb::into_buffer` will need to be able to handle it, such as by returning `None`.  However, that method can already return `Buffer::none`, which means something different.  It would be confusing to return two different None types.

In my experience, your program's flow should make it obvious whether any given `AioCb` is still in progress or not.  Treating an in-progress `AioCb` as though it has already finished is almost always going to be a programming bug.  So I think `assert`ing is appropriate.

asomers:
I'm all satisfied with this change now.  Review at your leisure.

src/sys/aio.rs
```diff
@@ -66,17 +67,48 @@ pub enum AioCancelStat {
...
+/// the kernel has a pointer to it.
 #[derive(Clone, Debug)]
-enum Keeper<'a> {
+pub enum Buffer<'a> {
     none,
-    /// Keeps a reference to a Boxed slice
-    boxed(Rc<Box<[u8]>>),
+    /// Immutable shared ownership `Bytes` object
+    // Must use out-of-line allocation so the address of the data will be
+    // stable.  Bytes and BytesMut sometimes dynamically allocate a buffer, and
```
Susurrus: Backticks around Bytes and BytesMut

src/sys/aio.rs
```diff
@@ -66,17 +67,48 @@ pub enum AioCancelStat {
...
 }
 
-/// Private type used by nix to keep buffers from Drop'ing while the kernel has
-/// a pointer to them.
+/// Owns (uniquely or shared) a memory buffer to keep it from `Drop`ing while
+/// the kernel has a pointer to it.
 #[derive(Clone, Debug)]
-enum Keeper<'a> {
+pub enum Buffer<'a> {
     none,
```
Susurrus: Can you add a doccomment for this variant?

Additionally, why is this lowercase `none`? I'd expect it to be `None`.

CHANGELOG.md
```diff
@@ -107,6 +105,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#648](https://github.com/nix-rust/nix/pull/648))
 
 ### Removed
+- `AioCb::from_boxed_slice` has been removed.  It was never actually safe.
```
Susurrus: What would be the replacement for this kind of functionality? If there's a short answer, it should be said here, otherwise it should be stated to "read the docs for more details" or "see the PR for more details". We should make it as easy as possible for downstream users of our library to update their code, this means pointing them in the right direction when we break it.

Cargo.toml
```diff
@@ -18,6 +18,11 @@ bitflags = "1.0"
 cfg-if = "0.1.0"
 void = "1.0.2"
 
+# Don't include the optional serde feature
```
Susurrus: Can you move this comment to right about the `default-features` line so it's more clear that's what it's pertaining to?

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Added `AioCb::from_ptr` and `AioCb::from_mut_ptr`
-  ([#TODO](https://github.com/nix-rust/nix/pull/TODO))
+  ([#820](https://github.com/nix-rust/nix/pull/820))
```
Susurrus: Please fix the initial commit to have this number from the start.

CHANGELOG.md
```diff
@@ -58,8 +58,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Use native `pipe2` on all BSD targets.  Users should notice no difference.
   ([#777](https://github.com/nix-rust/nix/pull/777))
-- `AioCb::from_boxed_slice` no longer allows `aio_read`
```
Susurrus: Is there a reason why we have this step fixing up `from_boxed_slice` only to remove it in the following commit?

asomers: Not really.  I should squash them together.

src/sys/aio.rs
asomers: Not for any good reason.  I'll change it back.  However, I'm wondering if I should make the Buffer type private.  If I do that, then I'll have to replace `AioCb::into_buffer` with `AioCb::into_bytes -> Option<Bytes>` and `AioCb::into_bytes_mut -> Option<BytesMut>`.  Do you think that would be a better API?

asomers: Nevermind.  I forgot that I can't do that due to a limitation of tokio-core.  In order to satisfy tokio-core, I need way to get the `Buffer` without consuming the `AioCb`.

asomers:
@Susurrus I think I responded to all of your concerns except the one about squashing.  I left everything unsquashed for your viewing pleasure.

src/sys/aio.rs
```diff
@@ -1,4 +1,27 @@
...
+//! [`fsync`](struct.AioCb.html#method.fsync) operations.  Completion
+//! notifications can optionally be delivered via
+//! [signals](../signal/enum.SigevNotify.html#variant.SigevSignal), via the
+//! [`aio_suspend`](fn.aio_suspend.html) function, or via polling.  Some
+//! platforms support other completion
+//! notifications, such as
+//! [kevent](../signal/enum.SigevNotify.html#variant.SigevKevent).
+//!
+//! Multiple operations may be submitted in a batch with
+//! [`lio_listio`](fn.lio_listio.html), thought the standard does not guarantee
```
Susurrus: thought -> though

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Added `AioCb::from_ptr` and `AioCb::from_mut_ptr`
+  ([#820](https://github.com/nix-rust/nix/pull/TODO))
```
Susurrus: This first link should be removed.

src/sys/aio.rs
```diff
@@ -231,6 +617,46 @@ impl<'a> AioCb<'a> {
...
 
     /// Cancels an outstanding AIO request.
+    ///
+    /// The operating system is not required to implement cancellation for all
+    /// file and device types.  Even if it does, there is no guarantee that the
+    /// operation has not already completed.  So the caller must check the
+    /// result and handle operations that were not canceled or that have already
+    /// completed.
+    ///
+    /// See also
```
Susurrus: I would rephrase to indicate this is an external link like "See the man pages for `aio_cancel` for more details." This applies for all the uses of "See also LINK".

asomers:
Travis is back up now.  I'm going to go ahead and merge this even though the last OSX builds never ran, because the last changes I made were just formatting stuff that shouldn't have broken OSX.

bors r+ susurrus

Susurrus:
🤞

asomers:
Try try again

bors r+ susurrus

asomers:
If at third you don't succeed, try try again.

bors r+ susurrus

Susurrus:
Travis is still having a tough time with Mac builds. Let's wait a couple of days until the backlog dies down a bit and we can try again.

asomers:
I see that libc was able to run OSX builds on Travis.  Maybe we can too.

bors r+ susurrus

Susurrus:
`libc` has a paid Travis subscription, so they can almost always run jobs. Note that the Open Source Mac build queue is almost 3000; this isn't going to finish.

Susurrus:
Mac OS build queue down to 1100, so maybe this could actually happen now!

bors r+

PR #822, opened 2018-01-01
bump rand to 0.4

ignatenkobrain:


Susurrus:
I don't see any bug fixes in the CHANGELOG for `rand`, any particular reason for doing this (you didn't provide a description of why you submitted this PR, which would have been helpful)?

ignatenkobrain:
@Susurrus is there reason not to? ;)

Relying on latest versions is always good. Also in Fedora (where I package nix) we try to stick with latest versions.

Susurrus:
I didn't see `nix` packaged for Fedora (it's the repo I use), but that's cool.

As for bumping dependencies, `rand` is used by various packages within the dependency graph for `nix`, and none of those use 0.4, so bumping the dependency will reduce compilation times as `rand-0.4` and `rand-0.3` are now both compiled instead of just `0.3`.

Ignoring that, upgrades do carry a small risk, and if there is no technical benefit to an upgrade, I'm inclined to not do so. Does this make your packaging job harder, or will it once `nix-0.10.0` comes out?

ignatenkobrain:
> Ignoring that, upgrades do carry a small risk, and if there is no technical benefit to an upgrade, I'm inclined to not do so. Does this make your packaging job harder, or will it once nix-0.10.0 comes out?

To be honest, having tons of libraries depending on nix 0.8.x instead of 0.9.x creates much more problems ;) Since this doesn't need any patching in source code, I don't have problems maintaining downstream patch.

ignatenkobrain:
> I didn't see nix packaged for Fedora (it's the repo I use), but that's cool.

https://koji.fedoraproject.org/koji/packageinfo?packageID=25930

Susurrus:
I'm definitely interested in making lives easier for packagers, so if this makes your job easier, that's a benefit that I see outweighing the potential risk.

As I use Fedora, I'm really curious how packaging Rust libraries works in practice. Is there any documentation about this effort? Could I use Rust libraries installed from the Fedora archives in my own project development or if they're merely available in order to build other Rust software. Since there's no defined Rust ABI, and `nix` doesn't export a C ABI, I'm curious what the use case is here for packaging it.

bors r+

ignatenkobrain:
> As I use Fedora, I'm really curious how packaging Rust libraries works in practice. Is there any documentation about this effort? Could I use Rust libraries installed from the Fedora archives in my own project development or if they're merely available in order to build other Rust software. Since there's no defined Rust ABI, and nix doesn't export a C ABI, I'm curious what the use case is here for packaging it.

https://fedoraproject.org/wiki/Packaging:Rust
https://fedoraproject.org/wiki/Changes/Packaging_Rust_applications_and_libraries

So we package source-code just in order to not bundle sources of each library into each applications. So it still links statically, but it's not such big problem as bundling source code of dependencies into binary.

Unfortunately even you have all those rust-*-devel installed, you can't automagically use them by cargo, it requires replacing crates.io repository with local one.. See https://pagure.io/fedora-rust/rust2rpm/blob/master/f/data/macros.cargo#_10-35 how it is done in practice.

I really need to file RFE for cargo to allow mixing crates.io archive & local repository as it is done in pip/gem/whatever.

Susurrus:
Spurious failure for cargo on `aarch64-unknown-linux-gnu`, retrying.

bors r+

Susurrus:
I don't know what's going on here, but I'm betting it's a Travis or rustup problem, so I'm going to let this sit for a bit.

PR #823, opened 2018-01-02
implement FromRawFd for PtyMaster

Mic92:
This is needed if file descriptor is received from a unix socket

### DIFF
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -45,6 +45,12 @@ impl IntoRawFd for PtyMaster {
     }
 }
 
+impl FromRawFd for PtyMaster {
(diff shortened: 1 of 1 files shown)


Susurrus:
Can you explain the use case for this a bit more? Ideally you can provide a doc test for `PtyMaster` illustrating the use of this functionality.

Additionally there should be a CHANGELOG entry for this.

Mic92:
The doc test would be longer then the code itself. What do you want me to test?

Mic92:
I don't like the way this project is maintained. I decided to continue to use my fork.

PR #828, opened 2018-01-06
Add `alarm` and `cancel_alarm`

Thomasdezeeuw:


src/unistd.rs
```diff
@@ -1326,6 +1327,45 @@ pub fn pause() -> Result<()> {
...
+///
+/// See also [alarm](http://pubs.opengroup.org/onlinepubs/9699919799/functions/alarm.html).
+///
+/// # Note
+///
+/// Alarm only allows for times with a second granularity, anything smaller then
+/// a second will be ignored (not rounded up). E.g. 1 second and 800 miliseconds
+/// will become 1 second.
+///
+/// Also, Strictly conforming the POSIX spec, duriations longer then `65535`
```
asomers: IMHO the warning in the POSIX specification is dumb.  It just warns you against using constants greater than `UINT_MAX`, which is something your compiler can already warn you about.  However, this Rust wrapper introduces a new problem: `secs` is a `u64`, which can overflow at runtime.  You should probably `assert` that `secs <= libc::c_uint::max_value()`

src/unistd.rs
```diff
@@ -1326,6 +1327,45 @@ pub fn pause() -> Result<()> {
...
+    }
+}
+
+/// Cancels an alarm signal.
+///
+/// The alarm signal must be previously set by [`alarm`], and will be returned
+/// if any is canceld.
+///
+/// [`alarm`]: fn.alarm.html
+pub fn cancel_alarm() -> Option<Duration> {
```
asomers: Is this function really necessary?  Can't we accomplish the same thing just by making `alarm`'s argument an `Option<Duration>`?

test/test_unistd.rs
```diff
@@ -400,3 +403,56 @@ fn test_pipe2() {
...
+    let m = ::SIGNAL_MTX.lock().expect("Mutex got poisoned by another test");
+
+    // Set an alarm.
+    assert_eq!(alarm(Duration::from_secs(60)), None);
+
+    // Overwriting an alarm should return the old alarm.
+    assert_eq!(alarm(Duration::from_secs(1)), Some(Duration::from_secs(60)));
+    let start = Instant::now();
+
+    let handler = SigHandler::Handler(alarm_signal_handler);
```
asomers: You should set up the hander _before_ calling `alarm`

Thomasdezeeuw: I know, however this test takes at least a second so I though I add the handler after so it removes some of the overhead, but I'll before it before the call.

src/unistd.rs
Thomasdezeeuw: I will change it.

src/unistd.rs
Thomasdezeeuw: You want just an assert, or a silent convertion into the max value with a warning in the doc?

asomers: I think an assert would be better than a saturating conversion.  After all, why would somebody ever set an alarm for more than 4 billion seconds into the future?  Such a thing is probably a bug.

Thomasdezeeuw:
I addressed all feedback and squashed the commits.

test/test_unistd.rs
```diff
@@ -400,3 +403,55 @@ fn test_pipe2() {
...
+    // Reset the signal.
+    unsafe {
+        sigaction(Signal::SIGALRM, &old_handler)
+            .expect("unable to set signal handler for alarm");
+    }
+}
+
+#[test]
+fn test_canceling_alarm() {
+    #[allow(unused_variables)]
```
asomers: I believe you can eliminate this directive if you change the variable's name from `m` to `_m`.  That didn't work with older Rust releases, but it should be possible now that the minimum Rust is 1.20.0

test/test_unistd.rs
```diff
@@ -400,3 +403,55 @@ fn test_pipe2() {
     let f1 = FdFlag::from_bits_truncate(fcntl(fd1, FcntlArg::F_GETFD).unwrap());
     assert!(f1.contains(FdFlag::FD_CLOEXEC));
 }
+
+// Used in `test_alarm`.
+static mut ALARM_CALLED: bool = false;
+
+// Used in `test_alarm`.
+#[no_mangle]
```
asomers: What's the purpose of `no_mangle`?  Using `extern` should be sufficient to pass the function pointer to a C function.

test/test_unistd.rs
Thomasdezeeuw: Shall I do it for all uses while I'm at it?

test/test_unistd.rs
Thomasdezeeuw: I just copied a handle from my code, which (if I can remeber correctly) needed this. Will see if I can remove it.

test/test_unistd.rs
asomers: If you do, make it a separate commit.

Susurrus:
I'm not a big fan of using `Duration` with these functions given that there are limitations with it's range that cannot be checked at compile time. What's wrong with making it take a `c_uint` directly? Anyone using a `Duration` can convert it easily.

test/test_unistd.rs
Thomasdezeeuw: Then I'll make a seperate PR for that, including this line.

Thomasdezeeuw:
@Susurrus I don't really mind either way, I just though to make the API little more Rust-like.

Thomasdezeeuw:
Chages the parameter and return value to `libc::c_uint`.

test/test_unistd.rs
Thomasdezeeuw: You're right it wasn't needed.

asomers:
The tests are failing because Buildbot can't figure out how to clone your repo.  Probably something to do with your most recent force push.  But it looks like you need to rebase anyway, to resolve the conflicts.  Could you please do that?

Thomasdezeeuw:
This should be good now, trying to resolve the conflict on the GitHub website messed up the branch for me. But I reset based on the new master and it should be good.

Susurrus:
Looking at this diff on GitHub shows it as adding all of `test_unistd.rs` and `unistd.rs`. I think something's still wrong with this branch.

Thomasdezeeuw:
I'll make another pr tomorrow from a fresh repo, although I'm not sure what when wrong.

Thomasdezeeuw:
New pr in #830.

Issue #827, opened 2018-01-06
Remove return value from `pause`


Thomasdezeeuw:
Currently `nix::unistd::pause` returns an `nix::Error`, however it can only return `EINTR` as an error, which is the point of pause. It pauses until a signal is received. So I suggest to remove the return value completely and document that `pause` only returns if the process receives a signal.

If you want to be really safe maybe an assert can be added to check if the error is really `EINTR`, but then the OS/libc is vialting the spec.

Man page: http://pubs.opengroup.org/onlinepubs/9699919799/functions/pause.html.

Susurrus:
Yes, that seems like a nice ergonomics improvement! Care to submit a PR for it?

Thomasdezeeuw:
Closed by #836.

PR #829, opened 2018-01-07
Remove return value from `pause`

Thomasdezeeuw:
Closes issue #827.

### DIFF
--- a/test/sys/test_wait.rs
+++ b/test/sys/test_wait.rs
@@ -12,7 +12,7 @@ fn test_wait_signal() {
 
     // Safe: The child only calls `pause` and/or `_exit`, which are async-signal-safe.
     match fork().expect("Error: Fork Failed") {
-      Child => pause().unwrap_or_else(|_| unsafe { _exit(123) }),
+      Child => pause(),
       Parent { child } => {
           kill(child, Some(SIGKILL)).expect("Error: Kill Failed");
           assert_eq!(waitpid(child, None), Ok(WaitStatus::Signaled(child, SIGKILL, false)));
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
I'll redo this as #828, since it might create another conflict when adding to the changelog.

test/sys/test_wait.rs
```diff
@@ -12,7 +12,7 @@ fn test_wait_signal() {
 
     // Safe: The child only calls `pause` and/or `_exit`, which are async-signal-safe.
     match fork().expect("Error: Fork Failed") {
-      Child => pause().unwrap_or_else(|_| unsafe { _exit(123) }),
+      Child => pause(),
```
Susurrus: I believe that you have to call `_exit()` instead of letting the thread naturally exit because of how Rust's standard cleanup isn't signal safe or something. @asomers Can you verify this?

asomers: Yep.  The standard cleanup may try to access shared resources that were exclusively held by other threads at the time that `fork` happened.  But those other threads don't exist in the child, so your thread will never get the resources, and deadlock will result.  That's why we call `_exit` in child threads.

Thomasdezeeuw: Maybe I'm misunderstanding something but, isn't `unwrap_or_else` only called if `pause` fails and returns an error. Since that can't happend anymore `_exit` can never be called, and would have never be called even without the change.

Thomasdezeeuw: I've fixed this in #836, for some reason this pull request is now "from unknown repository". I don't know how to fix that, so instead I created an new pr.

PR #830, opened 2018-01-07
Add alarm

Thomasdezeeuw:
Fixed #828.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added `alarm`. ([#830](https://github.com/nix-rust/nix/pull/830))
+
 ### Changed
 
 ### Fixed
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1324,6 +1324,85 @@ pub fn pause() {
     unsafe { libc::pause() };
 }
 
+pub mod alarm {
+    //! Alarm signal scheduling.
+    //!
+    //! Scheduling an alarm will trigger a `SIGALRM` signal when the time has
+    //! elapsed, which has to be caught, because the default action for the
+    //! signal is to terminate the program. This signal also can't be ignored
+    //! because the system calls like `pause` will not be interrupted, see the
+    //! second example below.
+    //!
+    //! # Examples
+    //!
+    //! Canceling an alarm:
+    //!
+    //! ```
+    //! use nix::unistd::alarm;
+    //!
+    //! // Set an alarm for 60 seconds from now.
+    //! alarm::set(60);
+    //!
+    //! // Cancel the above set alarm, which returns the number of seconds left
+    //! // of the previously set alarm.
+    //! assert_eq!(alarm::cancel(), Some(60));
+    //! ```
+    //!
+    //! Scheduling an alarm and waiting for the signal:
+    //!
+    //! ```
+    //! use std::time::{Duration, Instant};
+    //!
+    //! use nix::unistd::{alarm, pause};
+    //! use nix::sys::signal::*;
+    //!
+    //! // We need to setup an empty signal handler to catch the alarm signal,
+    //! // otherwise the program will be terminated once the signal is delivered.
+    //! extern fn signal_handler(_: nix::libc::c_int) { }
+    //! unsafe { sigaction(Signal::SIGALRM, &SigAction::new(SigHandler::Handler(signal_handler), SaFlags::empty(), SigSet::empty())); }
+    //!
+    //! // Set an alarm for 1 second from now.
+    //! alarm::set(1);
+    //!
+    //! let start = Instant::now();
+    //! // Pause the process until the alarm signal is received.
+    //! pause();
+    //!
+    //! assert!(start.elapsed() >= Duration::from_secs(1));
+    //! ```
+    //!
+    //! # References
+    //!
+    //! See also [alarm(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/alarm.html).
+
+    use libc;
+
+    /// Schedule an alarm signal.
+    ///
+    /// This will cause the system to generate a `SIGALRM` signal for the
+    /// process after the specified number of seconds have elapsed.
+    ///
+    /// Returns the leftover time of a previously set alarm if there was one.
+    pub fn set(secs: libc::c_uint) -> Option<libc::c_uint> {
+        assert!(secs != 0, "passing 0 to `alarm::set` is not allowed, to cancel an alarm use `alarm::cancel`");
+        alarm(secs)
+    }
+
+    /// Cancel an previously set alarm signal.
+    ///
+    /// Returns the leftover time of a previously set alarm if there was one.
+    pub fn cancel() -> Option<libc::c_uint> {
+        alarm(0)
+    }
+
+    fn alarm(secs: libc::c_uint) -> Option<libc::c_uint> {
+        match unsafe { libc::alarm(secs) } {
+            0 => None,
+            secs => Some(secs),
+        }
+    }
+}
+
 /// Suspend execution for an interval of time
 ///
 /// See also [sleep(2)](http://pubs.opengroup.org/onlinepubs/009695399/functions/sleep.html#tag_03_705_05)
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -3,6 +3,7 @@ extern crate tempdir;
 use nix::fcntl::{fcntl, FcntlArg, FdFlag, OFlag};
 use nix::unistd::*;
 use nix::unistd::ForkResult::*;
+use nix::sys::signal::{SaFlags, SigAction, SigHandler, SigSet, Signal, sigaction};
 use nix::sys::wait::*;
 use nix::sys::stat::{self, Mode, SFlag};
 use std::{self, env, iter};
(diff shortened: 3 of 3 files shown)


asomers:
Travis failed because of a temporary network failure.  Try again.

bors r+

Thomasdezeeuw:
Bors timeout out, but the build passed. Does that mean it's ok to merge?

asomers:
The timeout was probably buildbot's fault.  Buildbot has suddenly started to run much slower than it used to.  There's probably some kind of problem with the VM host.  We should just try again.  However, you'll have to rebase to fix the conflict in CHANGELOG.md.

Thomasdezeeuw:
@asomers This should be good to merge now.

Susurrus:
You'll need to rebase this to update the CHANGELOG now that 0.10.0 was released.

Thomasdezeeuw:
done.

Susurrus:
In looking this over again I realized that this is another one of those functions that overloads the intneger range with different functionality. I'm not a fan of this and generally try to alter the API to avoid this. For example in this case there are really two functions that should exist, one to set an alarm to a given number of seconds and another to clear any existing alarm. I'd much prefer using an `enum` datatype for the arguments so you'd do something like `alarm(Seconds(15))` to set the alarm and `alarm(Clear)` to clear it. That reads better to me than `alarm(15)` and `alarm(0)`.

Thomasdezeeuw:
@Surrus I've change the argument to `SetAlarm`, with the variants `Cancel` and `Set` to be more inline with the documentation on Opengroup (they use cancel rather then clear).

src/unistd.rs
```diff
@@ -1324,6 +1324,42 @@ pub fn pause() {
...
+///
+/// See also [alarm](http://pubs.opengroup.org/onlinepubs/9699919799/functions/alarm.html).
+///
+/// # Example
+///
+/// ```
+/// use nix::unistd::{alarm, SetAlarm};
+///
+/// // Set an alarm for 60 seconds from now.
+/// alarm(SetAlarm::Set(60));
```
Susurrus: I forgot that you need to specify the enum name when you initialize a variant of it, so this doesn't read as well as I hoped. I was originally going to suggest that instead of an enum we split `alarm` into two functions, `alarm_set(c_uint)` and `alarm_cancel()`. I'm now thinking that would be a more ergonomic API. The only downside to that is discoverability, but I think the docs will be enough there. And if a user searches the docs for "alarm", I'd expect them to find these two functions and be able to understand their usage.

@Thomasdezeeuw What do you think?

Thomasdezeeuw: In my original pr (https://github.com/nix-rust/nix/pull/828) I had two those function exactly, but @asomers was against that.

Susurrus: I didn't realize this was a spin off of an earlier PR, sorry about that. Read through it quickly and it provides a little more context. I didn't get a chance to respond early in the last PR, but I think if we have functions that start with `alarm_` it's a nice API. So let's go ahead and switch to that type of API. @asomers Care to chime in here?

Susurrus: I'm also actually thinking rather than name these functions that are split off from a single `libc` function by prefixing them within their function name, instead prefix them within a module. So you'd have `alarm::set(c_uint)` and `alarm::cancel()`. That should really help with the end-user documentation.

asomers: Actually, I like the module idea.  It's less verbose than using enums, but still Rustier than using `alarm_set` and `alarm_cancel`.

Thomasdezeeuw: I've put the function into their own module and also expanded the documentation.

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
...
+    //!
+    //! Scheduling an alarm will trigger a `SIGALRM` signal when the the time
+    //! has elapsed, which has to be caught, because the default action for the
+    //! signal is to terminate the program. This signal also can't be ignored
+    //! because the system calls like `pause` will not be interrupted, see the
+    //! second example below.
+    //!
+    //! # Examples
+    //!
+    //! Canceling an alarm.
```
Susurrus: : instead of .

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
...
 }
 
+pub mod alarm {
+    //! Alarm signal scheduling.
+    //!
+    //! See also [alarm](http://pubs.opengroup.org/onlinepubs/9699919799/functions/alarm.html).
+    //!
+    //! # Notes
+    //!
+    //! Scheduling an alarm will trigger a `SIGALRM` signal when the the time
```
Susurrus: "the the"

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
     unsafe { libc::pause() };
 }
 
+pub mod alarm {
+    //! Alarm signal scheduling.
+    //!
+    //! See also [alarm](http://pubs.opengroup.org/onlinepubs/9699919799/functions/alarm.html).
+    //!
+    //! # Notes
```
Susurrus: I would just remove this "Notes" header

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
...
+    //!
+    //! // Set an alarm for 60 seconds from now.
+    //! alarm::set(60);
+    //!
+    //! // Cancel the above set alarm, which returns the number of seconds left
+    //! // of the previously set alarm.
+    //! assert_eq!(alarm::cancel(), 60);
+    //! ```
+    //!
+    //! Scheduling an alarm and waiting for the signal.
```
Susurrus: : instead of .

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
...
+    //!
+    //! assert!(start.elapsed() >= Duration::from_secs(1));
+    //! ```
+
+    use libc;
+
+    /// Schedule an alarm signal.
+    ///
+    /// This will cause the system to generate a `SIGALRM` signal for the
+    /// process after the number of `secs` have elapsed.
```
Susurrus: "after the specified number"

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
...
+    ///
+    /// Returns the leftover time of a previously set alarm, or `0` if no alarm
```
Susurrus: Do you think an enum return type would be more helpful here than just an int?

src/unistd.rs
```diff
@@ -1324,6 +1324,79 @@ pub fn pause() {
...
+    /// was set.
+    pub fn set(secs: libc::c_uint) -> libc::c_uint {
```
Susurrus: Should we check for 0 in the input argument with an assertion? I think if people are doing that they should really be using `cancel` instead of `set`.

src/unistd.rs
Thomasdezeeuw: I was thinking about that as well, maybe `Option<libc::c_uint>` with 0 mapped to `None`.

src/unistd.rs
Thomasdezeeuw: I could add a simple assertion.

Thomasdezeeuw: What about this:
```
assert!(secs != 0, "passing 0 to `alarm::set` is not allowed, to cancel an alarm use `alarm::cancel`");
```

Susurrus: Sure, that'd be fine.

src/unistd.rs
Susurrus: I think using the `Option` type here makes a lot of sense!

src/unistd.rs
```diff
@@ -1324,6 +1324,85 @@ pub fn pause() {
...
+    //! ```
+
+    use libc;
+
+    /// Schedule an alarm signal.
+    ///
+    /// This will cause the system to generate a `SIGALRM` signal for the
+    /// process after the specified number of seconds have elapsed.
+    ///
+    /// Returns the leftover time of a previously set alarm, or `0` if no alarm
```
Susurrus: Change to "Returns the leftover time of a previously set alarm if there was one" since now this returns an `Option`.

src/unistd.rs
```diff
@@ -1324,6 +1324,85 @@ pub fn pause() {
...
+    /// was set.
+    pub fn set(secs: libc::c_uint) -> Option<libc::c_uint> {
+        assert!(secs != 0, "passing 0 to `alarm::set` is not allowed, to cancel an alarm use `alarm::cancel`");
+        alarm(secs)
+    }
+
+    /// Cancel an previously set alarm signal.
+    ///
+    /// Returns the leftover time of a previously set alarm, or 0 if no alarm
```
Susurrus: Same as the above.

src/unistd.rs
```diff
@@ -1324,6 +1324,85 @@ pub fn pause() {
     unsafe { libc::pause() };
 }
 
+pub mod alarm {
+    //! Alarm signal scheduling.
+    //!
+    //! See also [alarm](http://pubs.opengroup.org/onlinepubs/9699919799/functions/alarm.html).
```
Susurrus: Change the link to "alarm(2)" and move this into its own "# References" section after the Examples section.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added `alarm`.
```
Susurrus: Just make this one line.

Susurrus:
LGTM. Please squash into a single commit and then we'll merge.

@asomers Care to take another gander?

Thomasdezeeuw:
@Susurrus done.

Issue #824, opened 2018-01-03
ioctl write_int doesn't support passing long integers


jethrogb:
The Linux kernel API always passes the ioctl parameter as an `unsigned long`. However, the `ioctl!` macro always uses `c_int`. It should probably use `c_long`/`c_ulong`/`usize` instead, in order to pass all possible values.

Susurrus:
Where is this specified for Linux? And if this is correct, then why wouldn't `c_ulong` be the most appropriate choice out of those 3 options? Only one should be used, but you didn't clarify which is the best option in your statement.

Susurrus:
@jethrogb I don't understand enough to fix this, so I need some more information from you. Specifically a test case that you know is failing currently.

Linux should actually be using a `c_ulong` type for this according to this code in `src/sys/ioctl/platform/linux.rs`:
```rust
#[cfg(not(any(target_os = "android", target_env = "musl")))]
#[doc(hidden)]
pub type ioctl_num_type = ::libc::c_ulong;
```

Susurrus:
Sorry, actually looking at the code I do see the size of a `c_int` is used in generating the ioctl request code, so maybe that's the error you're indicating? I'm going to wait for more details from you before digging into this further.

jethrogb:
`ioctl_num_type` is the type used for the ioctl number, not the parameter. It might be ok to reuse it for the parameter? Not sure if it's consistent across platforms.

The problem is with this line https://github.com/Susurrus/nix/blob/7eb8dd1ccf5ec8281076e9aa8f1598e0e8bcbd68/src/sys/ioctl/mod.rs#L544

Susurrus:
I still don't see any documentation suggesting what youre saying here. The passed integer must be equal in size to a pointer type because of the way ioctl is defined. We'll need to sort out what's appropriate for each platform, and we'll need docs for that. Can you link to any?

jethrogb:
https://github.com/torvalds/linux/blob/5d38f04/include/linux/fs.h#L1702-L1703

Susurrus:
I couldn't quickly find the declaration of `ioctl` on Android to find the `ioctl_num_type`, but I'd assume it's `c_int`. And `musl` shouldn't matter since this is a Linux kernel API which is independent of the `libc` being used. So I agree with you there. I wonder why it was ever like this to begin with?

I'll fix this and add also specify the parameter as `c_ulong` as well in #833.

[...]

PR #825, opened 2018-01-03
FreeBSD: cfmakesane, EVFILT_*

myfreeweb:
Depends on: https://github.com/rust-lang/libc/pull/887

PR #833, opened 2018-01-08
Refactor the ioctl API and documentation

Susurrus:
I still need to flesh out the docs for the different `ioctl_*!` variants that now exist. I separated them into different macros so they can have their own docs, which should help discoverability. And when macro namespacing comes around this will be a pretty neatly documented API I think (though we'll likely want to rename these macros again at that point).

 * Split `ioctl!` into separate macros. This makes documentation easier to read.
 * Expose `request_code_*!` in the documentation to make the `ioctl_*_bad` macros easier to use.
 * Reorganize the file hierarchy to be simpler

cc @gabrielesvelto @posborne @jethrogb


Susurrus:
I think I've gotten this into its final form. By clarifying in every ioctl wrapper generator macro what the macro arguments are, what the generated function will look like, and an example if I've found one, we end up with I think some very solid docs!

Right now I'm missing examples (and tests) for `ioctl_readwrite_buf!`, `ioctl_read_buf!`, and `ioctl_readwrite_bad!`, so I'd also appreciate if anyone knows of any ioctls like those that I could add.

I'd encourage anyone viewing this to pull this branch and `cargo doc --open`, as I'd very much appreciate some double-checks on the final formatted output.

posborne:
Read through everything and I think this is a significant improvement on the documentation front (including very helpful examples).  I also like the design of the macros versus the overloaded macros we have previously with the unconventional "with" syntax.

I think the biggest decision here is whether it makes sense to just drop the old macro or if we should provide a mapping from the old macros to the new ones and mark the old `ioctl!` macro as deprecated.

Susurrus:
Great! I thought this makes things a lot better too, so I'm glad someone else read through and confirmed that!

Regarding maintaining the old macro through a deprecation period, I would be onboard except that I'd need to whip up a ton of new documentation for the ioctl! macro to make sure it was properly documented, or would you say punt on the documentation and just say, use these other functions instead?

src/sys/ioctl/mod.rs
```diff
@@ -185,64 +202,71 @@
...
 //! # fn main() {}
 //! ```
 //!
-//! Finding ioctl documentation
-//! ---------------------------
+//! Finding `ioctl` Documentation
+//! -----------------------------
 //!
 //! For Linux, look at your system's headers. For example, `/usr/include/linux/input.h` has a lot
 //! of lines defining macros which use `_IO`, `_IOR`, `_IOW`, `_IOC`, and `_IORW`. Some `ioctl`s are
```
jethrogb: `_IORW` is actually `_IOWR`

src/sys/ioctl/mod.rs
```diff
@@ -107,14 +107,12 @@
 //! [`ioctl_list` man page](http://man7.org/linux/man-pages/man2/ioctl_list.2.html) describes a
```
jethrogb: `_IORW` is actually `_IOWR`

src/sys/ioctl/mod.rs
```diff
@@ -138,17 +136,36 @@
 //! pub unsafe fn tcgets(fd: c_int, data: *mut termios) -> Result<c_int>;
 //! ```
 //!
-//! There is also a `bad none`, `bad write_int`/`bad write_ptr`, and `bad readwrite` variant that work
-//! similar to the standard `none`, `write_int`/`write_ptr`, and `readwrite` variants.
+//! The `ioctl_*_bad` variants should also be used when `ioctl` numbers are generated using the
+//! wrong macro. For example, the `KVM_CREATE_VM` ioctl is defined as follows in C:
```
jethrogb: In this example, the ioctl is **not** using the wrong macro. The R/W in `_IOR`,  `_IOW`, `_IOWR` means that in addition to the parameter passed to ioctl, user memory is to be read/written. Since this ioctl doesn't pass a pointer, no additional user memory is involved, so there is nothing to be read or written.

myfreeweb:
From #825: writing integers should be of type VOID on FreeBSD and DragonFly:

https://github.com/freebsd/freebsd/blob/7982612fd5e8ef311a2a58178893f54e6a0601a4/sys/sys/ioccom.h#L59

https://gitweb.dragonflybsd.org/dragonfly.git/blob/c3c55883c35bc4a6ebdf01eab8725c83884536c9:/sys/sys/ioccom.h#l57

example of usage:

https://github.com/freebsd/freebsd/blob/d8596f6f687a64b994b065f3058155405dfc39db/sys/sys/consio.h#L357-L363

Susurrus:
@myfreeweb or @asomers Could either of you provide correct values for `_IOWINT('v', 4)` and `_IOWINT('p', 2)` on both x64 and i686 FreeBSD? I want to write some tests to make sure our calculations are correct on both.

Susurrus:
@asomers and @myfreeweb I've added a commit that should rectify the `ioctl_write_int` problem, but the tests will fail until I get the proper values from either of you to test against.

I still need to fix #824 here and also jethrog's comment above.

myfreeweb:
`0x20047604` and `0x20047002` on `amd64` and `arm64`.

I don't have an actual 32 bit machine right now, but just cross compiling with the `-m32` flag produces the same values. (I haven't seen any 32/64 bit conditionals in the ioctl headers either…)

src/sys/ioctl/mod.rs
Susurrus: So the mapping should look like:
no argument (implicit 0) -> `_IO`
ulong argument -> `_IO`
pointer argument -> `_IOR`/`_IOW`/`_IOWR`

I believe that's what you're saying?

jethrogb: Yes. This seems to also be the case for (some of the) BSDs as @myfreeweb pointed out.

Susurrus:
Okay, did another pass through this PR and I think I've fixed all outstanding issues. Since the last time I added a CHANGELOG entry for the last commit and removed the section about "bad" ioctl's applying to ioctl's that used the wrong macros as I don't have a concrete example.

posborne:
@Susurrus Looks great to me.

bors r+

Susurrus:
Just realized this is missing a CHANGELOG entry to identify that the `ioctl!` macro has changed to the `ioctl_*` macros. But I can do that as a separate commit.

PR #834, opened 2018-01-08
Expose mqueue functions for all supported OSes

asomers:


Susurrus:
This is actually defined for `fuchsia` as well, should probably add that since we're already doing this.

Otherwise this LGTM.

asomers:
I enabled it for fuschia.  Since we don't do builds on fushia, I'll merge it without waiting for the builds to complete.

bors r+ susurrus

PR #837, opened 2018-01-14
More cleanups and enable some more compiler lints

Susurrus:
Gonna run this through CI and see what breaks on non-Linux-x64 platforms.

These changes are minor cleanups, but improve our doc story and should help with future submitted PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#837](https://github.com/nix-rust/nix/pull/837))
 
 ### Fixed
+- Properly exposed 460800 and 921600 baud rates on NetBSD
+  ([#837](https://github.com/nix-rust/nix/pull/837))
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 ### Changed
+- Display and Debug for SysControlAddr now includes all fields.
+  ([#837](https://github.com/nix-rust/nix/pull/837))
 
 ### Fixed
 
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -205,6 +205,7 @@ libc_bitflags!(
     }
 );
 
+#[allow(missing_debug_implementations)]
 pub enum FcntlArg<'a> {
     F_DUPFD(RawFd),
     F_DUPFD_CLOEXEC(RawFd),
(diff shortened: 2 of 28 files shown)


src/fcntl.rs
```diff
@@ -205,6 +205,7 @@ libc_bitflags!(
     }
 );
 
+#[allow(missing_debug_implementations)]
```
asomers: Why do you allow missing debug impls for certain types?

Susurrus: No `libc` types implement `Debug`, so `derive(Debug)` doesn't work for them. I've decided to defer these implementations for now in order to get these checks working for PRs in general. I'd like to come back and add `impl Debug` for all types as a future PR.

Susurrus:
The latest version relies on rust-lang/libc#898 being merged, but that's passing test so should happen soon.

Susurrus:
@asomers This is now passing on all platforms and should be good to merge once the OS X backlog dies down a bit on Travis. Can you take a look-see in the meantime? I only expect 429af5e1f190cae22a2bb7acf81e89f54eebabe1 to possibly be controversial since it adds `Copy`/`Clone` to a bunch of times, maybe some where it doesn't make sense to have those implemented, specifically `RemoteIoVec`, `SignalIterator`, and `OpenptyResult`.

CHANGELOG.md
```diff
@@ -96,6 +96,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#814](https://github.com/nix-rust/nix/pull/814))
 - Removed return type from `pause`.
   ([#829](https://github.com/nix-rust/nix/pull/829))
+- Display and Debug for SysControlAddr now includes all fields.
+  ([#829](https://github.com/nix-rust/nix/pull/829))
```
asomers: This PR number is wrong.

src/features.rs
```diff
@@ -82,6 +83,7 @@ mod os {
         }
     }
 
+    /// Check if the OS supports atmomic close-on-exec for sockets
```
asomers: s/atmomic/atomic here and on line 99

src/sys/aio.rs
```diff
@@ -86,7 +86,7 @@ pub struct AioCb<'a> {
     /// Could this `AioCb` potentially have any in-kernel state?
     in_progress: bool,
     /// Used to keep buffers from Drop'ing
-    keeper: Keeper<'a>
+    _keeper: Keeper<'a>
```
asomers: These changes will be unnecessary after and will conflict with #820 .  You should revert the changes to this file.

Susurrus:
I've updated based on your first 2 comments. I'll block this one on the merging of #820 and update this once that's merged.

Susurrus:
@asomers All good?

PR #841, opened 2018-01-21
Fix libc at 0.2.36 in preparation for release

asomers:


Susurrus:
Let's just do this before Travis is dumb again.

bors r+

Issue #577, opened 2017-04-14
Ability to add custom socket options


code-ape:
I was looking to do the following code except using nix instead of libc:

```rust
use libc::{c_ushort, c_int, c_void, socklen_t, SOL_SOCKET, setsockopt};

const SO_ATTACH_FILTER: c_int = 26;

pub fn attach_filter(fd: RawFd, prog: Prog) -> Result<(), Error> {
    match unsafe {
              setsockopt(fd as c_int,
                         SOL_SOCKET,
                         SO_ATTACH_FILTER,
                         &prog as *const _ as *const c_void,
                         size_of_val(&prog) as socklen_t)
          } {
        0 => Ok(()),
        _ => Err(Error::last_os_error()),
    }
}
```

However nix currently doesn't support this, which is fine. However, nix doesn't expose an easy way to use it's internal macros to construct the SO_ATTACH_FILTER command. I wanted to file an issue to suggest exposing this. I'll also look into contributing my use case to nix so it's added t the library for future users.

Code source: [https://github.com/jedisct1/rust-bpf/blob/0122d0a95e6012abab0475b14cef20dbfa481704/src/bpf_linux.rs#L58](https://github.com/jedisct1/rust-bpf/blob/0122d0a95e6012abab0475b14cef20dbfa481704/src/bpf_linux.rs#L58)

Susurrus:
I don't quite understand the issue here as I'm not familiar with `setsockopt`. Do you have a proposal for how we might integrate this code example that would fit in with the rest of `nix`? Would exposing `sockopt_impl!` be enough for you?

mexus:
Sorry for bringing up an old issue, but it happened so I need the same feature :)

Basically the issue is that the macro `sockopt_impl!` is not exposed, which will help a lot.
But it's not the only thing.

For example, for `SO_ATTACH_FILTER` we need to pass an object of a special type, which obviously does not implement [Set](https://docs.rs/nix/0.9.0/src/nix/sys/socket/sockopt.rs.html#193). But the `Set` trait is not exposed as well :)

So I guess exposing `sockopt_impl!` macro and `Set` and `Get` traits will help a lot.

Susurrus:
Is there a reason we can't provide a `sockopt_impl!` within `nix` that exposes `SO_ATTACH_FILTER`? This has been the need for both you and @code-ape, so I'd suggest we start going down this route. We'd welcome a PR for it.

[...]

PR #846, opened 2018-01-27
Export socket options related macros

mexus:
In this PR I've added docs to that macros and exported them.
I've also made related structs/traits to be public as well.
Hope I didn't forget anything :)

Reference issue: #577.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -6,6 +6,28 @@ use libc::{self, c_int, uint8_t, c_void, socklen_t};
 use std::mem;
 use std::os::unix::io::RawFd;
 
+/// Helper for implementing `SetSockOpt` for a given socket option. See
+/// [`::sys::socket::SetSockOpt`](sys/socket/trait.SetSockOpt.html).
+///
+/// This macro aims to help implementing `SetSockOpt` for different socket options that accept
+/// different kinds of data to be used with `setsockopt`.
+///
+/// Instead of using this macro directly consider using `sockopt_impl!`, especially if the option
+/// you are implementing represents a simple type.
+///
+/// # Arguments
+///
+/// * `$name:ident`: name of the type you want to implement `SetSockOpt` for.
+/// * `$level:path` : socket layer, or a `protocol level`: could be *raw sockets*
+///    (`lic::SOL_SOCKET`), *ip protocol* (libc::IPPROTO_IP), *tcp protocol* (`libc::IPPROTO_TCP`),
+///    and more. Please refer to your system manual for more options. Will be passed as the second
+///    argument (`level`) to the `setsockopt` call.
+/// * `$flag:path`: a flag name to set. Some examples: `libc::SO_REUSEADDR`, `libc::TCP_NODELAY`,
+///    `libc::IP_ADD_MEMBERSHIP` and others. Will be passed as the third argument (`option_name`)
+///    to the `setsockopt` call.
+/// * Type of the value that you are going to set.
+/// * Type that implements the `Set` trait for the type from the previous item (like `SetBool` for
+///    `bool`, `SetUsize` for `usize`, etc.).
 macro_rules! setsockopt_impl {
     ($name:ident, $level:path, $flag:path, $ty:ty, $setter:ty) => {
         impl SetSockOpt for $name {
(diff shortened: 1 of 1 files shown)


mexus:
Is there something wrong with the request?

Susurrus:
I'd be more interested in exposing a proper `SO_ATTACH_FILTER` wrapper rather than exporting all of the macros instead. I'd think that's also be better for downstream as then we own that code rather than them. Was there a reason you took this second approach? If it's feasible to add this wrapper I'd like to add that and only expose this macros as a last resort.

mexus:
Well, the reason was not to force the users to make a pull request every time they need a socket option that is not implemented yet :) But it will be good for the crate, right? And eventually all the options, at least the common ones, will be implemented here.

But nevertheless, I've got your point so lets close this one and I'll just make a SO_ATTACH_FILTER implementation.

Susurrus:
Yes, I'd think it'd overall be better for the crate and our downstream users if we implemented everything in `nix` so that everyone could benefit.

Additionally, I would actually like to merge your documentation changes here, just not the part that exports the macros. Would you be willing to revise this PR in that fashion and we'll merge it?

mexus:
Sure, np.

mexus:
Although the examples will have to be marked with `no_run` :( But I guess they won't be changed much in the future.

src/sys/socket/sockopt.rs
```diff
@@ -25,6 +47,28 @@ macro_rules! setsockopt_impl {
...
+///
+/// This macro aims to help implementing `GetSockOpt` for different socket options that accept
+/// different kinds of data to be use with `getsockopt`.
+///
+/// Instead of using this macro directly consider using `sockopt_impl!`, especially if the option
+/// you are implementing represents a simple type.
+///
+/// # Arguments
+///
+/// 1. Name of the type you want to implement `GetSockOpt` for.
```
Susurrus: Please use bullet points here `*` instead of a numbered list.

src/sys/socket/sockopt.rs
```diff
@@ -25,6 +47,28 @@ macro_rules! setsockopt_impl {
...
+/// 1. Socket layer, or a `protocol level`: could be *raw sockets* (`lic::SOL_SOCKET`),  *ip
+///    protocol* (libc::IPPROTO_IP), *tcp protocol* (`libc::IPPROTO_TCP`),  and more. Please refere
```
Susurrus: "refere" -> "refer"

src/sys/socket/sockopt.rs
```diff
@@ -6,6 +6,28 @@ use libc::{self, c_int, uint8_t, c_void, socklen_t};
...
+///
+/// This macro aims to help implementing `SetSockOpt` for different socket options that accept
+/// different kinds of data to be used with `setsockopt`.
+///
+/// Instead of using this macro directly consider using `sockopt_impl!`, especially if the option
+/// you are implementing represents a simple type.
+///
+/// # Arguments
+///
+/// 1. `$name:ident`: name of the type you want to implement `SetSockOpt` for.
```
Susurrus: Change these to bullet points using `*` instead.

src/sys/socket/sockopt.rs
```diff
@@ -6,6 +6,28 @@ use libc::{self, c_int, uint8_t, c_void, socklen_t};
...
+/// 1. `$level:path` : socket layer, or a `protocol level`: could be *raw sockets*
+///    (`lic::SOL_SOCKET`), *ip protocol* (libc::IPPROTO_IP), *tcp protocol* (`libc::IPPROTO_TCP`),
+///    and more. Please refere to your system manual for more options. Will be passed as the second
```
Susurrus: "refere" -> "refer"

src/sys/socket/sockopt.rs
```diff
@@ -46,7 +90,31 @@ macro_rules! getsockopt_impl {
...
 
-// Helper to generate the sockopt accessors
+/// Helper to generate the sockopt accessors. See
+/// [`::sys::socket::GetSockOpt`](sys/socket/trait.GetSockOpt.html) and
+/// [`::sys::socket::SetSockOpt`](sys/socket/trait.SetSockOpt.html).
+///
+/// This macro aims to help implementing `GetSockOpt` and `SetSockOpt` for different socket options
+/// that accept different kinds of data to be use with `getsockopt` and `setsockopt` respectively.
+///
+/// Basically this macro wraps up [`getsockopt_impl!`](macro.getsockopt_impl.html) and
```
Susurrus: "wraps up " -> "wraps up the "

src/sys/socket/sockopt.rs
```diff
@@ -46,7 +90,31 @@ macro_rules! getsockopt_impl {
...
+/// [`setsockopt_impl!`](macro.setsockopt_impl.html) macros.
+///
+/// # Arguments
+///
+/// 1. `GetOnly`, `SetOnly` or `Both`: whether you want to implement only getter, only setter or
```
Susurrus: Change to bullet points using `*` instead

src/sys/socket/sockopt.rs
```diff
@@ -46,7 +90,31 @@ macro_rules! getsockopt_impl {
...
+///    both of them.
+/// 1. `$name:ident`: name of type `GetSockOpt`/`SetSockOpt` will be implemented for.
+/// 1. `$level:path` : socket layer, or a `protocol level`: could be *raw sockets*
+///    (`lic::SOL_SOCKET`), *ip protocol* (libc::IPPROTO_IP), *tcp protocol* (`libc::IPPROTO_TCP`),
+///    and more. Please refere to your system manual for more options. Will be passed as the second
```
Susurrus: "refere" -> "refer"

mexus:
Hope that fits :)

Susurrus:
Awesome, thanks!

bors r+

tomwhoiscontrary:
Do i understand correctly that the PR as merged documented the internal functions and types, but did not expose them? So there's still no way to define custom socket options?

mexus:
I'd say you've got right the part about the docs, but there is a way to at least use custom options: `setsockopt`/`getsockopt` accept objects that implement `SetSockOpt`/`GetSockOpt` which are public traits, so it shouldn't big a big deal. Yes, the `set` and `get` methods of the traits are hidden from the docs (for the reasons unknown to me), but still.

If by "custom" you mean something actually common (i.e. from the libc) but which hasn't been implemented yet inside the *nix* crate, it might be a good idea to either post an issue or make a PR with the desired option. If it's something really custom, then I guess implementing the traits once shouldn't be an issue..

Having said that I want to add that if I was a maintainer I'd definitely make the macro public with some comments like "please consider sending a PR if you're implementing something common!" :)

tomwhoiscontrary:
Aha, that is probably enough.

The specific situation we were thinking about was for some funky hardware where the vendor-supplied TCP/IP stack has a few extra socket options to configure it. I don't think it would make sense to put those into Nix.

We would like to see [SO_TIMESTAMPING](https://www.kernel.org/doc/Documentation/networking/timestamping.txt) in Nix, though. That would also need some machinery to decode the timestamps returned in the cmsg ancillary data, which is not terrifically complicated. We'll see if we can get a PR in for this ...

PR #848, opened 2018-01-27
waitpid() now has target pid as an enum. Resolves #840.

### LINKED ISSUE #840: Revise waitpid API
There are 4 different types of calls to `waitpid()`, all of which use a different range of a `c_int` to be expressed. Instead an enum type should be used that includes:

- wait for any child process with a specific process group ID
- wait for any child process
- wait for any child process whose process group ID is equal to that of the calling process
- wait for the child with a specific process ID

serejkus:
Please, note that it is an API-breaking change.  Resolves #840.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -1,4 +1,6 @@
-use libc::{self, c_int};
+use std::convert::{From, Into};
+
+use libc::{self, c_int, pid_t};
 use Result;
 use errno::Errno;
 use unistd::Pid;
@@ -45,6 +47,30 @@ libc_bitflags!(
         __WCLONE;
     }
 );
+
+/// Possible child targets of`waitpid()`.
+///
+/// The most common variants for waiting are waiting
+/// for a specific child via `WaitPid::Child(pid)` and
+/// waiting for any child via `WaitPid::AnyChild`.
+#[derive(Eq, PartialEq, Clone, Copy, Debug)]
+pub enum WaitTarget {
+    /// Waiting for a child with `Pid` process id.
+    Child(Pid),
+    /// Waiting for a child of the same process group as the calling process.
+    ThisGroupChild,
+    /// Waiting for any child.
+    AnyChild,
+    /// Waiting for a child of process group `Pid` (NB: do not negate this value,
+    /// just use process id of the group leader).
+    ChildOfGroup(Pid),
+}
+
+impl From<Pid> for WaitTarget {
+    fn from(pid: Pid) -> WaitTarget {
+        WaitTarget::Child(pid)
+    }
+}
 
 /// Possible return values from `wait()` or `waitpid()`.
 ///
@@ -207,7 +233,7 @@ impl WaitStatus {
     }
 }
 
-pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
+pub fn waitpid<P: Into<WaitTarget>>(who: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
     use self::WaitStatus::*;
 
     let mut status: i32 = 0;
@@ -217,9 +243,17 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
         None => 0,
     };
 
+    let who: WaitTarget = who.into();
+    let pid: pid_t = match who {
+        WaitTarget::Child(pid) => pid.into(),
+        WaitTarget::ThisGroupChild => 0,
+        WaitTarget::AnyChild => -1,
+        WaitTarget::ChildOfGroup(pid) => { let pid: pid_t = pid.into(); -pid },
+    };
+
     let res = unsafe {
         libc::waitpid(
-            pid.into().unwrap_or(Pid::from_raw(-1)).into(),
+            pid,
             &mut status as *mut c_int,
             option_bits,
         )
@@ -232,5 +266,5 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
 }
 
 pub fn wait() -> Result<WaitStatus> {
-    waitpid(None, None)
+    waitpid(WaitTarget::AnyChild, None)
 }


src/sys/wait.rs
```diff
@@ -46,6 +48,30 @@ libc_bitflags!(
...
+/// waiting for any child via `WaitPid::AnyChild`.
+#[derive(Eq, PartialEq, Clone, Copy, Debug)]
+pub enum WaitTarget {
+    /// Waiting for a child with `Pid` process id.
+    Child(Pid),
+    /// Waiting for a child of the same process group as the calling process.
+    ThisGroupChild,
+    /// Waiting for any child.
+    AnyChild,
+    /// Waiting for a child of process group `Pid` (NB: do not negate this value,
```
Susurrus: What does NB here mean?

src/sys/wait.rs
```diff
@@ -207,7 +233,7 @@ impl WaitStatus {
     }
 }
 
-pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
+pub fn waitpid<P: Into<WaitTarget>>(who: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
```
Susurrus: Can you provide docs for `waitpid` here? I think it's important to note in them that passing a straight `Pid` translates to `WaitTarget::Child(Pid)`

src/sys/wait.rs
```diff
@@ -232,5 +266,5 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
 }
 
 pub fn wait() -> Result<WaitStatus> {
```
Susurrus: Please add docs for this function as well.

src/sys/wait.rs
```diff
@@ -46,6 +48,30 @@ libc_bitflags!(
     }
 );
 
+/// Possible child targets of`waitpid()`.
```
Susurrus: Missing space here.

src/sys/wait.rs
```diff
@@ -46,6 +48,30 @@ libc_bitflags!(
...
+///
+/// The most common variants for waiting are waiting
```
Susurrus: Please wrap these comments a little longer so they match the wrapping elsewhere in the file (usually 79 or 99 characters)

src/sys/wait.rs
```diff
@@ -46,6 +48,30 @@ libc_bitflags!(
...
+    /// Waiting for a child of the same process group as the calling process.
+    ThisGroupChild,
+    /// Waiting for any child.
+    AnyChild,
+    /// Waiting for a child of process group `Pid` (NB: do not negate this value,
+    /// just use process id of the group leader).
+    ChildOfGroup(Pid),
+}
+
+impl From<Pid> for WaitTarget {
```
Susurrus: We probably want this to follow the mapping that `waitpid` uses internally, where only positive PID values are mapped to `WaitTarget::Child`, the other 3 mappings should also be here. This should make `waitpid` work safely with current code. Additionally it should make this a backwards-compatible change.

src/sys/wait.rs
```diff
@@ -46,6 +48,30 @@ libc_bitflags!(
     }
 );
 
+/// Possible child targets of`waitpid()`.
+///
+/// The most common variants for waiting are waiting
+/// for a specific child via `WaitPid::Child(pid)` and
+/// waiting for any child via `WaitPid::AnyChild`.
+#[derive(Eq, PartialEq, Clone, Copy, Debug)]
+pub enum WaitTarget {
```
Susurrus: My only concern with this enum is that constructing it allows for negative values in places where negative values shouldn't be allowed. This should be explicitly disallowed, but I'm not certain what the most Rustic way to do this is.

src/sys/wait.rs
Susurrus: You can reference #645 so it should be pretty easy.

serejkus:
Thanks for pointing these issues, I hope I'll fix them in a week.

 > The only big problem I see with this is that you can actually construct an incorrect WaitTarget that already contains negative PIDs, the values should be checked when it's constructed, but I'm not certain of the idiomatic way to do this.

This was my first thought that it should be protected somehow so that no user can call it with negative pids accidentally. The second thought was "how'd they get negative pids?".  `fork()` in `nix` returns `Result`, so it is not possible. `getpid()` and `getppid()` always succeed. `waitpid()` returns `Result` in `nix`. Probably, I'm missing something, but currently it is only possible to create negative pid only manually, but what's the reason behind this? On the other hand, making direct use of `Pid` returned from very ergonomic. Like, most uses are waiting for concrete pid or waiting for any child. Waiting for concrete pid is just a matter of
```rust
match fork() {
  ForkResult::Parent(child) => { waitpid(child, None); }
  ForkResult::Child => { ... }
}
```

instead of being more explicit
```rust
match fork() {
  ForkResult::Parent(child) => {
    let wait_target = WaitTarget::concrete_child(child)?;
    waitpid(wait_target, None);
  }
  ForkResult::Child => { ... }
}
```

What do you think about it?

And a note about idiomatic way: I think that it might be implemented like
```rust
pub enum WaitTarget {
  // same non-pub variants
}

impl WaitTarget {
  pub fn child_from_raw(pid: pid_t)-> Result<WaitTarget, BadWait> {
    // ...
  }
  pub fn child(pid: Pid) -> Result<WaitTarget, BadWait> {
    // here we have to get the innards of Pid struct somehow
  }
  // this one never fails, here might be Result, but it would introduce some inconsistency
  pub fn any_child() -> WaitTarget {
  }
  // etc
}
```

src/sys/wait.rs
serejkus: It's [Nota Bene](https://en.wikipedia.org/wiki/Nota_bene). I'll replace it with "please, note" since it adds confusion.

Susurrus:
I didn't quite follow what you were proposing in your last comment. I think if we just expand the `From` that you already have in your current implementation to also have asserts to check if there are invalid `Pid` values, we'll have the ergonomic API and safety that we want.

Susurrus:
@serejkus Are you planning on coming back to finish this PR?

pickfire:
Any news on this?

asomers-ax:
> Any news on this?

Apparently not.  If this issue matters to you, would you be willing to adopt the PR and finish it?

pickfire:
I believe this can be closed. https://github.com/nix-rust/nix/pull/1262

serejkus:
thank you @pickfire and sorry @Susurrus, I totally forgot about this PR. I'm closing it since it is updated by @pickfire.

Issue #850, opened 2018-01-29
add getsid


ggriffiniii:
add getsid wrapping the equivalent libc call.

Susurrus:
It is not necessary to create an issue and then a PR for it. If you plan to file a PR, go ahead and file one and skip the issue, it's just extra email noise for us maintainers.

I'll take a look at your PR this week.

Susurrus:
Closed by #851.

PR #851, opened 2018-01-29
Added `getsid` in `::nix::unistd`

ggriffiniii:
Resolves Issue #850


asomers:
Could we get a test here?

asomers:
If nothing else, a doc test demonstrating usage would be good.  You could also get that `getsid(None) == getsid(Some(getpid()))`

ggriffiniii:
Test added

PR #852, opened 2018-01-30
Add step function to ptrace

xd009642:
Added step function to ptrace, this follows the same form as the PTRACE_CONTINUE by advanced the tracee by a single step!

Found when I was updating to nix 0.10.0 that this function had been missed out. Minor addition as `SINGLESTEP` works the same as `CONTINUE`

### DIFF
--- a/src/sys/ptrace.rs
+++ b/src/sys/ptrace.rs
@@ -292,10 +292,17 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
 /// use nix::sys::ptrace::step;
 /// use nix::unistd::Pid;
 /// use nix::sys::signal::Signal; 
+/// use nix::sys::wait::*;
 /// fn main() {
-///     let dummy_pid = Pid::from_raw(0); 
-///
-///     let _ = step(dummy_pid, Some(Signal::SIGSTOP));
+///     // If a process changes state to the stopped state because of a SIGUSR1 
+///     // signal, this will step the process forward and forward the user 
+///     // signal to the stopped process
+///     match waitpid(Pid::from_raw(-1), None) {
+///         Ok(WaitStatus::Stopped(pid, Signal::SIGUSR1)) => {
+///             let _ = step(pid, Signal::SIGUSR1);
+///         }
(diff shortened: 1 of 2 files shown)


xd009642:
Once this has been approved I'll update the changelog with the PR number etc :smile:

Susurrus:
I'd love to get an example and a test in here. Specifically it'd be good to show the syntax for how the optional argument is called both with and without it.

xd009642:
Will do, am I right in recalling there's an issue with using waitpid and signals in tests?

src/sys/ptrace.rs
```diff
@@ -280,3 +280,30 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
     }
 }
 
+/// Move the stopped tracee process forward by a single step as with 
+/// `ptrace(PTRACE_SINGLESTEP, ...)`
+///
+/// Advances the execution of the process with PID `pid` by a single step optionally delivering a
+/// single specified by `sig`.
```
Susurrus: "signal"

xd009642: Fixed!

src/sys/ptrace.rs
```diff
@@ -280,3 +280,30 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
...
+/// # Example
+/// ```rust
+/// extern crate nix;
+/// use nix::sys::ptrace::step;
+/// use nix::unistd::Pid;
+/// use nix::sys::signal::Signal; 
+/// fn main() {
+///     let dummy_pid = Pid::from_raw(0); 
+///
+///     let _ = step(dummy_pid, Some(Signal::SIGSTOP));
```
Susurrus: Shouldn't you not need that `Some()` on this line?

Also, any way to make this a more realistic example? This looks more like a basic test than any realistic code that might help a user use this function.

xd009642:
Hi @Susurrus I've made the changes, just checking to see if this is ready to be merged yet?

Susurrus:
LGTM, thanks!

bors r+

PR #857, opened 2018-02-05
Add chmod, fchmod, fchmodat functions

antage:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Added `alarm`. ([#830](https://github.com/nix-rust/nix/pull/830))
+- Added `chmod`, `fchmod`, `fchmodat`.
```
asomers: Please add a link to the PR, as other entries do.

src/sys/stat.rs
```diff
@@ -122,3 +122,88 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
     Ok(dst)
 }
 
+/// The `chmod` function shall change `Mode::S_ISUID`, `Mode::S_ISGID`, `Mode::S_ISVTX`,
```
asomers: The doccomment block must begin with a brief description, < 80 characters, followed by a blank line.  And it's not necessary to copy the description from opengroup; nobody is going to check Nix's docs to understand what the underlying system call's compliance requirements are.  Just write what a Rust programmer would need to know.

src/sys/stat.rs
```diff
@@ -122,3 +122,88 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+/// the `Mode::S_ISUID` and `Mode::S_ISGID` bits in `mode` to be ignored.
+///
+/// Upon successful completion, `chmod` shall mark for update the last file
+/// status change timestamp of the file.
+///
+/// # References
+///
+/// See also [chmod(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html).
+pub fn chmod<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
+    let res = try!(path.with_nix_path(|cstr| unsafe {
```
asomers: You can replace `try!` with `?`.  See https://blog.rust-lang.org/2016/11/10/Rust-1.13.html for a description of how it works.

src/sys/stat.rs
```diff
@@ -122,3 +122,88 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+        libc::chmod(cstr.as_ptr(), mode.bits() as mode_t)
+    }));
+
+    Errno::result(res).map(drop)
```
asomers: I prefer `.map(|_| () )`.  I think it's more obvious what you're trying to accomplish.

antage:
Done.

src/sys/stat.rs
```diff
@@ -122,3 +122,48 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
     Ok(dst)
 }
 
+/// Change the file permission bits of the file named by the pathname.
+///
+/// # References
+///
+/// See also [chmod(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html).
```
asomers: If it's going to be in a References block, then you don't need "See also"

src/sys/stat.rs
```diff
@@ -122,3 +122,48 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
     Ok(dst)
 }
 
+/// Change the file permission bits of the file named by the pathname.
```
Susurrus: Change to "Change the file permission bits of a file specified by a path"

src/sys/stat.rs
```diff
@@ -122,3 +122,48 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+///
+/// [chmod(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html).
+pub fn chmod<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
+    let res =
+        path.with_nix_path(|cstr| unsafe { libc::chmod(cstr.as_ptr(), mode.bits() as mode_t) })?;
+
+    Errno::result(res).map(|_| ())
+}
+
+/// Change the file permission bits of the file specified by the file descriptor.
```
Susurrus: "Change the file permission bits of the file specified by the file descriptor." -> "Change the file permission bits of the file specified by a file descriptor"

src/sys/stat.rs
```diff
@@ -122,3 +122,48 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+/// # References
+///
+/// [fchmodat(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fchmodat.html).
+pub fn fchmodat<P: ?Sized + NixPath>(
+    dirfd: RawFd,
+    path: &P,
+    mode: Mode,
+    flag: AtFlags,
+) -> Result<()> {
+    let res = path.with_nix_path(|cstr| unsafe {
```
Susurrus: This function actually has 3 modes:
 * Path relative to the cwd of the calling process
 * Path relative to the specified directory
 * Absolute path

I'd therefore suggest replacing `dirfd` and `path` with the following:
```rust
pub enum PathMode {
  RelPath(P),
  RelPath(P, dirfd),
  AbsPath(P),
}```

Then internally we can do the whole `AT_FDCWD` dance and also `debug_assert!` that those paths are correctly relative or absolute.

Comment (asomers) [reply src/sys/stat.rs]: There's no point in providing both `chmod` and `fchmodat(PathMode::AbsPath)`; they're identical.  Also, I don't think we should force a programmatic distinction between absolute and relative paths, because there are likely to be situations where a Nix consumer doesn't know which type of path it's dealing with, without inspecting it.  How about we provide `chmod`, `fchmod`, and `fchmodat`, where the latter asserts that `fd` is nonnegative?

Comment (antage) [reply src/sys/stat.rs]: > asserts that fd is nonnegative

We can not because `AT_FDCWD` is negative integer usually.

Comment (Susurrus) [reply src/sys/stat.rs]: I'm wondering if we can't wrap up these 3 functions and 5 modes into a different object that provides a nicer API than these functions separately? There is definite overlap between these functions, and I don't like the terrible combinatorics of the inputs to `fchmodat`. Any ideas here?

Comment (antage) [reply src/sys/stat.rs]: I think we shouldn't merge `fchmod` and `chmod`/`fchmodat` because these functions are very different (path vs fd as main argument).
We can get rid of `chmod` as it duplicates `fchmodat` and leave `fchmodat` with the signature:
```
pub fn fchmodat<P: ?Sized + NixPath>(path: P, mode: Mode, dirfd: Option<RawFd>, nofollow_symlink: bool) -> Result<()>
```
Here `None` in `dirfd` substitutes `AT_FDCWD` constant.

Comment (asomers) [reply src/sys/stat.rs]: I like that proposal.  I'd like it even better if it had the same argument order as `libc::fchmodat`.

Comment (Susurrus) [reply src/sys/stat.rs]: I highly suggest removing that boolean option. It makes reading the code much harder because you don't know what `True` and `False` actually mean without jumping into the docs. This is explicitly mentioned in [the Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#c-custom-type).

Comment (antage) [reply src/sys/stat.rs]: Ok, we can replace the bool argument with:
```
pub enum FchmodatFlag {
    FollowSymlink,
    NoFollowSymlink,
}
```

Comment (antage) [reply src/sys/stat.rs]: @asomers, @Susurrus The following is a summary of proposed changes:
1. Get rid of `chmod` function.
2. Add `FchmodatFlag` enum instead of `AtFlags`.
3. Change a signature of `fchmodat` function to: `pub fn fchmodat<P: ?Sized + NixPath>(dirfd: Option<RawFd>, path: P, mode: Mode, flag: FchmodatFlag) -> Result<()>`.

Do you have any objections or/and proposals?

Comment (asomers) [reply src/sys/stat.rs]: Yes, that sounds good to me.

Comment (antage) [reply src/sys/stat.rs]: @Susurrus Any response?

Comment (asomers) [reply src/sys/stat.rs]: @antage @Susurrus is AFK.  Go ahead and make the changes we discussed, then I can approve the PR.

Comment (antage) [reply src/sys/stat.rs]: @asomers ok.

**Comment by antage** [comment]
@asomers Done the proposed changes.

File: src/sys/stat.rs [hunk 9]
```diff
@@ -122,3 +122,64 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+///
+/// The file to be changed is determined relative to the directory associated
+/// with the file descriptor `dirfd` or the current working directory
+/// if `dirfd` is `None`.
+///
+/// If `flag` is `FchmodatFlags::NoFollowSymlink` and `path` names a symbolic link,
+/// then the mode of the symbolic link is changed.
+///
+/// `fchmod(None, path, mode, FchmodatFlags::FollowSymlink)` is identical to
+/// a call `libc::chmod(path, mode)`. It's reason why `chmod` is unimplemented
```
asomers: s/It's reason/That's/

src/sys/stat.rs
```diff
@@ -122,3 +122,64 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+/// in `nix` crate.
```
asomers: s/in `nix`/in the `nix`/

test/test_stat.rs
```diff
@@ -102,3 +103,51 @@ fn test_stat_fstat_lstat() {
...
+    let tempdir = TempDir::new("nix-test_fchmodat").unwrap();
+    let filename = tempdir.path().join("foo.txt");
+    File::create(&filename).unwrap();
+
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    let mut mode1 = Mode::empty();
+    mode1.insert(Mode::S_IRUSR);
+    mode1.insert(Mode::S_IWUSR);
+    fchmodat(Some(dirfd), &filename, mode1, FchmodatFlags::FollowSymlink).unwrap();
```
asomers: Here you're supplying an absolute path for `filename`.  I think you should just supply the relative path, "foo.txt".

antage:
@asomers Fixed.

asomers:
Thanks, @antage.  It all looks good now.

bors r+

PR #860, opened 2018-02-19
Update the release procedure

### LINKED ISSUE #468: Let's get rid of the -pre suffix.
As @nox stated in #363, no one else does it, and it apparently might cause, problems I don't fully understand, yet. Since, I don't see an advantage in having the pre in the repository all the time, I don't mind.

In addition, I would like to fix the next version to currently 0.(prev + 1).0 and not 0.prev.1, if we ever want to release a hotfix, we can do so in a separate branch with a 0.current.1 version.

I'll prepare an RFC, when I have time.

Susurrus:
Be more explicit about the development version to specify after doing a release.

cc @nox

Closes #468


asomers:
I discovered that `cargo-release` can actually update the CHANGELOG for us.  It can't change the libc dependency, though.  See https://github.com/sunng87/cargo-release#maintaining-changelog .

asomers:
I'm going to merge this, then add `cargo-release` instructions in a separate PR.

bors r+

Issue #861, opened 2018-02-21
Non-blocking on macOS / iOS


kristate:
Thanks to everyone for your hard work.

`nix::sys::socket::socket` provides a fourth parameter that ends-up calling `fcntl()`and friends but seemingly does not support macos.

What is the best way to open a non-blocking socket on macos/ios using nix?

ref: https://github.com/nix-rust/nix/blob/v0.10.0/src/sys/socket/mod.rs#L684-L718

Susurrus:
First off, your welcome! Hopefully even with this missing/unknown functionality you've found `nix` useful.

Unfortunately `nix` isn't the best place to start looking for answers to questions like these. `nix` doesn't have the manpower to support these questions, instead outsourcing that work to existing documentation for these C APIs. Please search for other sources that provide the documentation you need. It will likely be in C, but once you know the C equivalent, translating that into Rust using `nix` is usually straightforward.

Additionally, if the functionality you need isn't available in `nix`, we welcome PRs for POSIX/libc APIs. We do not have the manpower to implement features either, so requests such as those will just sit unanswered.

kristate:
@Susurrus I understand that you are trying to be professional in your answer. I know there is a lot of work to be done with both `libc` and and `nix` and I am here to help. Please reopen this issue -- as it is an issue:

```rust
pub fn socket<T: Into<Option<SockProtocol>>>(domain: AddressFamily, ty: SockType, flags: SockFlag, protocol: T) -> Result<RawFd> {
```
has a giant cfg gate on it that doesn't support macOS even though macOS does support OFlag::O_NONBLOCK

In 0.9.0 it was okay to use `SockFlag:: SOCK_NONBLOCK` on macOS, but 0.10.0 changes this.

I will attempt to write a PR and look for your help. Thanks.

For anyone who hits this via google, etc. you will have to set it manually until this gets resolved:

```rust

use libc;
use nix;
use std::io;

/*
cvt from mio lib
https://github.com/carllerche/mio/blob/91f0832036303c7f83c286e049e31be29c43b519/src/sys/unix/mod.rs#L71
*/

trait IsMinusOne {
  fn is_minus_one(&self) -> bool;
}

impl IsMinusOne for i32 {
  fn is_minus_one(&self) -> bool { *self == -1 }
}
impl IsMinusOne for isize {
  fn is_minus_one(&self) -> bool { *self == -1 }
}

fn cvt<T: IsMinusOne>(t: T) -> ::io::Result<T> {
[...]

PR #863, opened 2018-02-22
Fix Broken SockFlag::SOCK_NONBLOCK and SockFlag::SOCK_CLOEXEC on MacOS/iOS

kristate:
`SockFlag::SOCK_NONBLOCK` and `SockFlag::SOCK_CLOEXEC` in 0.9.0 broke macOS/iOS support in 0.10.0 (f067e73bdbb589a0e3c991eab2c26de8bb00e924)

This commit solves the issue and also provides test cases to test for continued support.

ref #244 #861

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -86,32 +86,44 @@ pub enum SockProtocol {
     KextControl = libc::SYSPROTO_CONTROL,
 }
 
-libc_bitflags!{
-    /// Additional socket options
-    pub struct SockFlag: c_int {
-        /// Set non-blocking mode on the new socket
-        #[cfg(any(target_os = "android",
-                  target_os = "dragonfly",
-                  target_os = "freebsd",
-                  target_os = "linux",
-                  target_os = "netbsd",
-                  target_os = "openbsd"))]
-        SOCK_NONBLOCK;
-        /// Set close-on-exec on the new descriptor
-        #[cfg(any(target_os = "android",
-                  target_os = "dragonfly",
-                  target_os = "freebsd",
-                  target_os = "linux",
-                  target_os = "netbsd",
-                  target_os = "openbsd"))]
-        SOCK_CLOEXEC;
-        /// Return `EPIPE` instead of raising `SIGPIPE`
-        #[cfg(target_os = "netbsd")]
-        SOCK_NOSIGPIPE;
-        /// For domains `AF_INET(6)`, only allow `connect(2)`, `sendto(2)`, or `sendmsg(2)`
-        /// to the DNS port (typically 53)
-        #[cfg(target_os = "openbsd")]
-        SOCK_DNS;
+cfg_if! {
+    if #[cfg(any(target_os = "macos",
+                 target_os = "ios"))] {
+        bitflags! {
+            pub struct SockFlag: c_int {
+                const SOCK_NONBLOCK = libc::O_NONBLOCK;
+                const SOCK_CLOEXEC = libc::O_CLOEXEC;
+            }
+        }
+    } else {
+        libc_bitflags!{
+            /// Additional socket options
+            pub struct SockFlag: c_int {
+                /// Set non-blocking mode on the new socket
+                #[cfg(any(target_os = "android",
+                          target_os = "dragonfly",
+                          target_os = "freebsd",
+                          target_os = "linux",
+                          target_os = "netbsd",
+                          target_os = "openbsd"))]
+                SOCK_NONBLOCK;
+                /// Set close-on-exec on the new descriptor
+                #[cfg(any(target_os = "android",
+                          target_os = "dragonfly",
+                          target_os = "freebsd",
+                          target_os = "linux",
+                          target_os = "netbsd",
+                          target_os = "openbsd"))]
+                SOCK_CLOEXEC;
+                /// Return `EPIPE` instead of raising `SIGPIPE`
+                #[cfg(target_os = "netbsd")]
+                SOCK_NOSIGPIPE;
+                /// For domains `AF_INET(6)`, only allow `connect(2)`, `sendto(2)`, or `sendmsg(2)`
+                /// to the DNS port (typically 53)
+                #[cfg(target_os = "openbsd")]
+                SOCK_DNS;
+            }
+        }
     }
 }
 
@@ -709,7 +721,9 @@ pub fn socket<T: Into<Option<SockProtocol>>>(domain: AddressFamily, ty: SockType
     #[cfg(any(target_os = "android",
               target_os = "dragonfly",
               target_os = "freebsd",
+              target_os = "ios",
               target_os = "linux",
+              target_os = "macos",
               target_os = "netbsd",
               target_os = "openbsd"))]
     {
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -255,3 +255,61 @@ pub fn test_syscontrol() {
     // requires root privileges
     // connect(fd, &sockaddr).expect("connect failed");
 }
+
+/// Test non-blocking mode on new sockets via SockFlag::O_NONBLOCK
+#[cfg(any(target_os = "android",
+          target_os = "dragonfly",
+          target_os = "freebsd",
+          target_os = "ios",
+          target_os = "linux",
+          target_os = "macos",
+          target_os = "netbsd",
+          target_os = "openbsd"))]
+#[test]
+pub fn test_sockflag_nonblock() {
+    use libc;
+    use nix::fcntl::{fcntl};
+    use nix::fcntl::FcntlArg::{F_GETFL};
+    use nix::sys::socket::{socket, AddressFamily, SockType, SockFlag};
+
+    /* first, try without SockFlag::SOCK_NONBLOCK */
+    let sock = socket(AddressFamily::Unix, SockType::Stream, SockFlag::empty(), None)
+                 .expect("socket failed");
+
+    let fcntl_res = fcntl(sock, F_GETFL).expect("fcntl failed");
+
+    assert!(fcntl_res & libc::O_NONBLOCK == 0);
+
+    /* next, try with SockFlag::SOCK_NONBLOCK */
+    let sock = socket(AddressFamily::Unix, SockType::Stream, SockFlag::SOCK_NONBLOCK, None)
+                 .expect("socket failed");
+
+    let fcntl_res = fcntl(sock, F_GETFL).expect("fcntl failed");
+
+    assert!(fcntl_res & libc::O_NONBLOCK == libc::O_NONBLOCK);
+}
+
+/// Test close-on-exec on new sockets via SockFlag::SOCK_CLOEXEC
+#[test]
+pub fn test_sockflag_cloexec() {
+    use libc;
+    use nix::fcntl::{fcntl};
+    use nix::fcntl::FcntlArg::{F_GETFD};
+    use nix::sys::socket::{socket, AddressFamily, SockType, SockFlag};
+
+    /* first, test without SockFlag::SOCK_CLOEXEC */
+    let sock = socket(AddressFamily::Unix, SockType::Stream, SockFlag::empty(), None)
+                 .expect("socket failed");
+
+    let fcntl_res = fcntl(sock, F_GETFD).expect("fcntl failed");
+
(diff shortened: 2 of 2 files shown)


src/sys/socket/mod.rs
```diff
@@ -86,32 +86,44 @@ pub enum SockProtocol {
...
-        SOCK_CLOEXEC;
-        /// Return `EPIPE` instead of raising `SIGPIPE`
-        #[cfg(target_os = "netbsd")]
-        SOCK_NOSIGPIPE;
-        /// For domains `AF_INET(6)`, only allow `connect(2)`, `sendto(2)`, or `sendmsg(2)`
-        /// to the DNS port (typically 53)
-        #[cfg(target_os = "openbsd")]
-        SOCK_DNS;
+cfg_if! {
+    if #[cfg(any(target_os = "macos",
```
kristate: Note here: `nix::sys::socket::socket` fn API requires the linux specific `SockFlag::SOCK_NONBLOCK` `SockFlag::SOCK_CLOEXEC` and which is not included in macOS' libc -- the work around is to provide a bitflag that mimics such support.

Susurrus: Why can't this be fixed in `libc`? Do these constants not exist on OS X and iOS?

Susurrus:
I just searched all of `/usr/include` on my mac, and I don't see any `SOCK_NONBLOCK` and `SOCK_CLOEXEC` flags declared. So I'm not certain how async functionality is exposed on Mac, but this doesn't seem to be the right path forward here.

src/sys/socket/mod.rs
kristate: This is fundamentally an implementation concern.

It's because the `nix::sys::socket::socket` fn is not only assigning a socket, but also setting flags on the socket in the same function to match parity with Linux (which is an improvement). The normal API is defined in socket(2) as:
```c
int socket(int domain, int type, int protocol);
```
linux diverts from POSIX and BSD by supplying a `SOCK_NONBLOCK` flag as is defined in linux man socket(2) as:
```
SOCK_NONBLOCK
Set the O_NONBLOCK file status flag on the new open
file description.  Using this flag saves extra calls
to fcntl(2) to achieve the same result.

SOCK_CLOEXEC
Set the close-on-exec (FD_CLOEXEC) flag on the new
file descriptor.  See the description of the
O_CLOEXEC flag in open(2) for reasons why this may be
useful.
```

In summery:
* `SOCK_NONBLOCK` and `SOCK_CLOEXEC` are in linux since 2.6.27
* Some BSDs have conformed to this new API and such have a define in `libc`.
* macOS is the exception and sticks to directly calling `fcntl(2)`
* `nix` takes the high road and also only calls `fcntl(2)` with a new fourth parameter on the function call

Because `nix` provides a new API that does not conform to POSIX or BSD (or directly Linux), this patch creates virtual definitions for macOS and iOS that are backed by their equivalent `fcntl` flags.

I think this is the best course of action for two reasons:

0. Redefining the function at this stage would break compatibility
0. Defining virtual definitions for `SOCK_NONBLOCK` and `SOCK_CLOEXEC` are only used inside of `nix` and only for the socket fn.

kristate:
@Susurrus I replied to your comment above here: https://github.com/nix-rust/nix/pull/863#discussion_r170171843

> I just searched all of /usr/include on my mac, and I don't see any SOCK_NONBLOCK and SOCK_CLOEXEC flags declared. So I'm not certain how async functionality is exposed on Mac, but this doesn't seem to be the right path forward here.

Please read my above comment and then let me know is the right path forward. I put in the work to support macOS and iOS and am willing to maintain it.

Thanks.

src/sys/socket/mod.rs
kristate: For the sake of conversation, the alternative is:

Removing `pub struct SockFlag` and having the API user call `fcntl` directly.

Thus, these function defs need to be changed (breaking downstream code):
```rust
pub fn socket<T: Into<Option<SockProtocol>>>(domain: AddressFamily, ty: SockType, flags: SockFlag, protocol: T) -> Result<RawFd> {
```
```rust
/// Create a pair of connected sockets
///
/// [Further reading](http://pubs.opengroup.org/onlinepubs/9699919799/functions/socketpair.html)
pub fn socketpair<T: Into<Option<SockProtocol>>>(domain: AddressFamily, ty: SockType, protocol: T, flags: SockFlag) -> Result<(RawFd, RawFd)> {
```
```rust
/// Accept a connection on a socket
///
/// [Further reading](http://man7.org/linux/man-pages/man2/accept.2.html)
pub fn accept4(sockfd: RawFd, flags: SockFlag) -> Result<RawFd> {
```
```rust
fn accept4_polyfill(sockfd: RawFd, flags: SockFlag) -> Result<RawFd> {
```

Susurrus:
Okay, so this is all pretty interesting. The code as it stands at HEAD is incorrect and this code doesn't fix the problem properly as well I don't think.

What we have is that the `SOCK_NONBLOCK` and `SOCK_CLOEXEC` options are not properly past to `socket`, but instead are always emulated on all platforms, even those that support them properly! Basically on a few platforms there is no native support for this. Namely:
 * Linux pre-2.6.27 (released in 2008)
 * OpenBSD pre-5.7 (released in 2015)
 * DragonFlyBSD pre-4.4 (released in 2015)
 * NetBSD pre-6.0 (released in 2012)
 * FreeBSD pre-2013 (unknown what release, commit adding this is from early 2013)

But what we actually do is instead of allowing the user to use these values directly on support platforms, what `nix` actually do is emulate them by including calls to `fcntl` setting the `O_NONBLOCK` or `FD_CLOEXEC` flags as appropriate.

So I think this is part of a larger issue here of exposing these APIs. @asomers any input here?

I think what we should do is disable this emulation for all platforms and only emulate it on mac/ios. Additionally I think that on mac/ios we shouldn't reuse the `O_NONBLOCK` and `FD_CLOEXEC` values  and just use arbitrary ones, since they're discarded anyways. Otherwise it's confusing semantics.

asomers:
I think that Nix's purpose is to expose OS functionality in a Rusty way.  I don't think we should try to emulate missing OS functionality, especially since we can't do it correctly (the emulated socket flags are not set atomically).  So I'm in favor of removing the emulation altogether, even on OSX.  And BTW, we already have a wrapper for `fcntl`.  It's in src/fcntl.rs .

Susurrus:
@kristate I understand your rationale for wanting this feature as I did read your original reply, but I disagree that this functionality should exist. Regarding your two points:
 0. We do not guarantee backwards compatibility at this point because we aren't very set on many of our APIs. This is indicated by our pre-1.0 versioning. In fact the `Changed` section of our `CHANGELOG` are all APIs we've broken compatibility with as we are willing to break compatibility for correctness or better APIs at this stage of development.
 1. We are not trying to provide emulation functionality within this library, especially one like this where it's not obvious we're abstracting something away. By reusing those variable names we're hiding things from our users that can have a clear impact. I could easily see someone posting an issue or PR undoing this or changing the API and/or updating the docs to make it clear that we're hiding things here.

So I agree with @asomers above about removing the emulation entirely. It's probably best to close this PR and submit a new one to do that. @kristate if you'd like to do that, we'd appreciate it, otherwise I'll file one later. I want to resolve this issue for the 0.11.0 release.

Susurrus:
> I really think that the API was trying to be rusty from the beginning. I was just fixing a bug.

But the code is still wrong even after this fix. All BSDs and Linux support these flags natively, so they don't need the emulation, but it's turned on for all BSDs. I'm not certain if we should add the runtime detection for BSDs or we should actually assume that we're running versions at least as new as what I listed above. 2015 releases are still pretty new, but I'm not certain it's worth the effort when we can just document it as minimum requirements for the lib. I'd also be fine dropping the runtime checks for Linux given it was added 10 years ago. @asomers do you agree?

Susurrus:
As to address you specific question: "the readme for this project even states "The goal is to not provide a 100% unified interface, but to unify what can be while still providing platform specific APIs." -- is this not a unified interface for socket?"

This is one of those not-100% scenarios. We try to provide a unified interface, but we don't try to emulate or approximate functionality. This code change emulates an atomic API with a non-atomic implementation, which I consider bad form. We have always torn out this code and left the API more lacking in these circumstances.

kristate:
@Susurrus I think I understand your concern now.

So, C programmers are used to having the `socket(2)` syscall and those who are used to linux expect to have `SOCK_NONBLOCK` and `SOCK_CLOEXEC` available to them. If you are used to macOS, then you understand that `fcntl` must be called after setting up the socket.

The point I am making is that there needs to be a name and a value for the flags to our rust API. `SOCK_NONBLOCK` and `SOCK_CLOEXEC` was previously assigned and map well to the linux `socket(2)`, so that is an improvement and well received by C programmers.

I think that the question we have to first ask ourselves is: Does `pub fn socket` follow linux's lead and allow `SOCK_NONBLOCK` and `SOCK_CLOEXEC` to be called as flags (and take care of `fcntl` behind the scenes) OR do we simply wrap the `libc` crate and expect users downstream to again wrap code in cfg macros?

"Emulation" is a poor way to think about this problem. We are creating a Rusty API that helps calm the waters when working with unix. I chose to work with `nix` because it helps stabilize an API to work with unix systems. Likewise, if `nix` does not have any benefits over `libc`, it would seems that programmers would chose `libc` over `nix`.

So, I leave it to you: Does `pub fn socket` follow linux's lead and allow `SOCK_NONBLOCK` and `SOCK_CLOEXEC` to be called as flags (and take care of `fcntl` behind the scenes if required) OR do we simply wrap the `libc` crate and expect users downstream to again wrap code in cfg macros where applicable?

Thanks.

kristate:
@Susurrus 
> This code change emulates an atomic API with a non-atomic implementation, which I consider bad form. 

Where in my code does this change emulate an atomic API with a non-atomic implementation?

By the way, it does seem to check if the OS will `cloexec` atomically which was added by @carllerche  ref: https://github.com/nix-rust/nix/commit/c976be575f4fefcf03a70c5b87f898388150c8aa

kristate:
@asomers I agree with you in principle, but either way `fcntl` should be called directly after `socket` if no atomic behavior exists in the kernel, so does that not mean that we are providing the best practice for downstream users?

We have many years of network programming knowledge that we can pass-on downstream to users who might be less aware. Also, I am thinking about adding more documentation to match this discussion.

Thoughts?

Susurrus:
> should be called directly after socket if no atomic behavior exists in the kernel, so does that not mean that we are providing the best practice for downstream users?

Your implication is that we should be doing this and we've been explaining that we don't believe we should. There are numerous technical reasons why which we've enumerated previously. We all would like if we could have a nice API for this across all platforms, but that's not how things are unfortunately. So what we try to do is make the best of a bad situation.

As for better documentation, I'm all about that, so if you feel you can improve our documentation in this area, we'd welcome it!

> Then socket would return EPROTOTYPE if somebody tried to use those flags on OSX or RHEL5.

This wouldn't apply to `OSX`, where these flags don't exist. It looks like the `O_CLOEXEC` flag was added in OS X 10.7, which was released in 2011 and was unsupported in 2014.

So I think I agree with you @aomers, let's remove all emulation here. Since the constants aren't even available on macos/ios, then I think this should have minimal impact on downstream codebases.

kristate:
I guess I feel a little torn here. More or less because the Berkeley socket API sucks in so many ways and by safely normalizing as much as possible, we could build a better ecosystem in Rust.

Better yet, rust's standard library sets `CLOEXEC` for all fds on unix by default: rust-lang/rust#24034

More discussion here: rust-lang/rust#24237

SteveLauC:
> Was this ever addressed in some other PRs?

I guess no, and I am actually surprised that the `socket()` interface exposed by Nix does not conform to any implementation, so this is indeed an issue. Git blame shows me that that wrong interface was added 9 years ago, before this [commit](https://github.com/nix-rust/nix/commit/4059820cbaed9786596bfc8ea136617cac0dc8b0#diff-c0c25672b8b4509c9120357012d6918c1d0072815e2dc8e7b77c48d7f35c6615).

And I think @asomers's thought is right, we should expose the system interface without emulating anything, I will take a look at this and possibly get it fixed.

kskalski:
If the workaround is simply to call `fnctl` after creating a socket the current API seems reasonable, but maybe leaving a hint to the user in form of doc comment would be good (https://github.com/nix-rust/nix/pull/2619) - for a few weeks I was juggling working around blocking mode for mac in our codebase until I read through this issue in detail.

PR #865, opened 2018-02-23
Add SockProtocol::Htons(u16) for RAW sockets

### LINKED ISSUE #854: `SockProtocol` lacks protocols
Currently the only protocols supported in [`SockProtocol`](https://docs.rs/nix/0.10.0/nix/sys/socket/enum.SockProtocol.html) are `Tcp` and `Udp` (**for linux**), while there are many more.

For example, I'm working with raw sockets and I need to specify a special  `ETH_P_ALL` protocol (`man 7 packet`), and there is one important thing about it: its number has to be presented in network byte order when passed to the system `socket` call.

Currently the `SockProtocol` enum simply provides `i32` constants, so I can't put `htons` conversion there since there is no `const fn` call for that.

There is also no way to provide an arbitrary integer as a protocol number to the nix's [`socket`](https://docs.rs/nix/0.10.0/nix/sys/socket/fn.socket.html) call, hence I'm left with no options.

So I would say it would be great to have an option to pass the protocol number as a number (`i32` or whatever) to the [`socket`](https://docs.rs/nix/0.10.0/nix/sys/socket/fn.socket.html) call, but it might be there are other ways to address the issue.

kristate:
This PR closes #854.

* Maintains backwards compatibility with existing API.
* Tests htons conversion

Welcome to comments and suggestions. Thanks.

/cc @mexus @Susurrus

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -69,22 +69,38 @@ pub enum SockType {
 
 /// Constants used in [`socket`](fn.socket.html) and [`socketpair`](fn.socketpair.html)
 /// to specify the protocol to use.
-#[repr(i32)]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub enum SockProtocol {
-    /// TCP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
-    Tcp = libc::IPPROTO_TCP,
-    /// UDP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
-    Udp = libc::IPPROTO_UDP,
-    /// Allows applications and other KEXTs to be notified when certain kernel events occur
-    /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
-    #[cfg(any(target_os = "ios", target_os = "macos"))]
-    KextEvent = libc::SYSPROTO_EVENT,
+pub enum SockProtocol {   
+    /// htons conversion for RAW type
+    Htons(u16),
     /// Allows applications to configure and control a KEXT
     /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
     #[cfg(any(target_os = "ios", target_os = "macos"))]
-    KextControl = libc::SYSPROTO_CONTROL,
+    KextControl,
+    /// Allows applications and other KEXTs to be notified when certain kernel events occur
+    /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
+    #[cfg(any(target_os = "ios", target_os = "macos"))]
+    KextEvent,
+    /// TCP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
+    Tcp,
+    /// UDP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
+    Udp,
 }
+
+impl Into<i32> for SockProtocol {
+    fn into(self) -> i32 {
+        match self {
+            SockProtocol::Tcp => libc::IPPROTO_TCP,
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -69,23 +69,39 @@ pub enum SockType {
...
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub enum SockProtocol {
-    /// TCP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
-    Tcp = libc::IPPROTO_TCP,
-    /// UDP protocol ([ip(7)](http://man7.org/linux/man-pages/man7/ip.7.html))
-    Udp = libc::IPPROTO_UDP,
-    /// Allows applications and other KEXTs to be notified when certain kernel events occur
+pub enum SockProtocol {   
+    /// htons conversion for RAW type
+    Htons(u16),
```
mexus: Since we are going to use plain numbers instead of fixed enums, why not make an entry like `Custom` that accept just an integer instead of `Htons`?
I mean a procotol number (in general) is just an arbitrary number, which (again, in general) doesn't have to do anything with the byte order, at least that is my understanding.

What'd you say?

src/sys/socket/mod.rs
```diff
@@ -69,23 +69,39 @@ pub enum SockType {
...
+    /// Allows applications to configure and control a KEXT
     /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
     #[cfg(any(target_os = "ios", target_os = "macos"))]
-    KextEvent = libc::SYSPROTO_EVENT,
-    /// Allows applications to configure and control a KEXT
+    KextControl,
```
mexus: Just out of curiosity, why did you reorder the items? :)

kristate: Alphabetic ordering :)

src/sys/socket/mod.rs
kristate: Thanks for reviewing the code!
Htons gives a conversion from `u16` to `htons(u16)`,
but yes, perhaps it might be best to also have a Integer(i32).

Two things:
1. `man socket(2)` correctly states that `The protocol number to use is specific to the communication domain in which communication is to take place.` meaning that, as you said, a protocol is just an arbitrary number for the sock type.
2. I like the idea of looking at API documentation and getting a feel for what is possible. That's what makes the enum so great.

So, how about keeping the existing enums, but adding an Integer(i32) enum type?

mexus:
Well, as I've said on gitter, I would rather go with just that `Integer` enum. I mean these `Htons` and `Htonl` enumerators basically duplicate the `to_be` method. What's the reason to have them? Why there is no enum for `u64`? :) And what about `ntohs` conversions? :)

And I would definitely keep `KextEvent` and others as they are.

Just for the record, I'm not the one to decide :)

kristate:
@mexus 

> Well, as I've said on gitter, I would rather go with just that Integer enum. I mean these Htons and Htonl enumerators basically duplicate the to_be method. What's the reason to have them?

Two reasons:
1. You have to get the defines from somewhere. Using defines from the `nix` create keeps things cleaner.
2. Should you get the defines from somewhere, `Htons` guarantees that the type called into `socket(2)` will be okay. There is a chance that someone calls `.to_be()` on a u32 instead of a u16 which, of course, is the difference between `Htons` and `Htonl` -- this plays to rust's safety.

> Why there is no enum for u64?

There is not enum for u64 because the man `socket(2)` call is a c_int.

> And what about ntohs conversions?

The reason for `Htons` is to ensure that the sequence of bytes called to `socket(2)` are in the correct order when linux constructs a `struct sockaddr_ll` later on internally to match ethernet frame protocols in big endian (network). Essentially, `ntohs` is not required because we are ever only setting parameters from host to network, never network to host.

From man `packet(7)`:
```
The link-level header information is available
in a common format in a sockaddr_ll structure.
protocol is the IEEE 802.3 protocol number in
network byte order.
```

Thanks for the questions.

PR #866, opened 2018-02-25
getsockopt for Vec<u8>

kristate:
This PR implements `getsockopt` for variable bytes as `Vec<u8>`.

- [X] Implementation
- [ ] Documentation (would be great to get your help, @mexus)
- [ ] Tests (hard to test this one, see #867)

## Background
On macOS / iOS, to get the IFNAME of a UTUN, a `getsockopt(fd, SYSPROTO_CONTROL, UTUN_OPT_IFNAME, val, len);` call must be made passing not a struct, but a CString of `IFNAMSIZ` bytes, which returns how many bytes of the CString was written to (including null-terminating byte).

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -76,7 +76,7 @@ macro_rules! getsockopt_impl {
 
             fn get(&self, fd: RawFd) -> Result<$ty> {
                 unsafe {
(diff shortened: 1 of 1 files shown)


SteveLauC:
Close as this has been implemented

```
impl<O: AsMut<[u8]>> Get for  GetOsString<O>
```

PR #867, opened 2018-02-25
macOS: Add getsockopt for (libc::SYSPROTO_CONTROL, libc::UTUN_OPT_IFNAME)

kristate:
Adds `getsockopt` implementation for (`libc::SYSPROTO_CONTROL`, `libc::UTUN_OPT_IFNAME`).

- [X] Implementation (requires #866)
- [X] Tests (hard to test on CI because it requires root to initiate the UTUN)

## Usage
`getsockopt(fd, SYSPROTO_CONTROL, UTUN_OPT_IFNAME, name, len)` is used to get the interface name of the `connect`ed tunnel adapter. Usually returns `utun` followed by the interface number.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -76,7 +76,7 @@ macro_rules! getsockopt_impl {
 
             fn get(&self, fd: RawFd) -> Result<$ty> {
                 unsafe {
-                    let mut getter: $getter = Get::blank();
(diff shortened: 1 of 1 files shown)


asomers:
This is weird enough that I don't want to merge it without a test.  It's ok if the test requires root access.  We already have some like that, for example `test_setgroups`.

SteveLauC:
Close as superseded by #2325

PR #868, opened 2018-02-28
Don't derive Clone for sys::aio::Buffer

asomers:
It's deceptive, because BytesMut::clone doesn't do what you expect.  It
actually copies the object to new storage, which could cause problems if
the Buffer gets Clone'd while the aio operation is still active in the
kernel.

Also, fix some comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#837](https://github.com/nix-rust/nix/pull/837))
 
 ### Removed
+- Don't derive Clone for sys::aio::buffer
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#837](https://github.com/nix-rust/nix/pull/837))
 
 ### Fixed
+- Don't derive Clone for sys::aio::buffer
```
Susurrus: I'd put this under "Removed" since this actually removed something from the API.

asomers:
I've decided to make a more drastic change to `aio::Buffer`, that's going to supercede this one.  Abandon the PR.

PR #869, opened 2018-03-02
Change SigAction::flags to use from_bits_truncated

Detegr:
On Linux, if the signal trampoline code is in the C library, sigaction
sets the SA_RESTORER flag (0x04000000) in the sa_flags field of old
sigaction (see sigreturn(2)).

This is not intended for application use and is missing from SaFlags,
therefore from_bits fails and unwrapping panics the user program.

This fix just drops the bits that are not defined in SaFlags.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -389,7 +389,7 @@ impl SigAction {
     }
 
     pub fn flags(&self) -> SaFlags {
-        SaFlags::from_bits(self.sigaction.sa_flags).unwrap()
+        SaFlags::from_bits_truncate(self.sigaction.sa_flags)
     }
 
     pub fn mask(&self) -> SigSet {
--- a/test/sys/test_signal.rs
+++ b/test/sys/test_signal.rs
@@ -7,6 +7,20 @@ fn test_kill_none() {
 }
 
 #[test]
+fn test_old_sigaction_flags() {
+    extern "C" fn handler(_: ::libc::c_int) {}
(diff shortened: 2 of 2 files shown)


asomers:
Do you have an example testcase that demonstrates the failure?

Detegr:
Sure.
```rust
extern crate nix;
use nix::sys::signal;

extern "C" fn handler(_: nix::libc::c_int) {}
fn main() {
    let act = signal::SigAction::new(
        signal::SigHandler::Handler(handler),
        signal::SaFlags::empty(),
        signal::SigSet::empty(),
    );
    let oact = unsafe { signal::sigaction(signal::SIGINT, &act) }.unwrap();
    println!("{:?}", oact.flags()); // Works, the default handler flags do not have SO_RESTORER set
    let oact = unsafe { signal::sigaction(signal::SIGINT, &act) }.unwrap();
    println!("{:?}", oact.flags()); // Panics, as sa_flags contains SO_RESTORER
}
```

asomers:
Yeah, that fails for me too, on Linux but not FreeBSD.  Could you please add it to tests/sys/test_signal.rs?

Detegr:
Added.

Susurrus:
This needed a CHANGELOG entry indicating the fix as well. @Detegr Would you be willing to submit one?

Detegr:
Sure. Is it okay to have the commit in this PR or should I open a new one?

PR #871, opened 2018-03-05
WIP Replace AioCb::from_bytes with more generic from_boxed_slice

asomers:
Supporting the bytes crate was unnecessarily specific.  This change
replaces from_bytes and from_bytes_mut with from_boxed_slice and
from_boxed_mut_slice, which can work with anything that implements
Borrow<[u8]> and BorrowMut<[u8]>, respectively.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,16 +21,13 @@ void = "1.0.2"
 gcc = "0.3"
 
 [dev-dependencies]
-bytes = { version="0.5.0", default_features = false }
+# The examples use on a new feature of Bytes which should be available in 0.5.0
+# https://github.com/carllerche/bytes/pull/185
+bytes = { git = "https://github.com/asomers/bytes", rev = "b57934d" }
 lazy_static = "1"
 rand = "0.4"
 tempdir = "0.3"
 tempfile = "2"
-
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -17,20 +17,21 @@ bitflags = "1.0"
...
-[dependencies.bytes]
-version = "0.4.5"
-# Don't include the optional serde feature
-default-features = false
-
 [target.'cfg(target_os = "dragonfly")'.build-dependencies]
 gcc = "0.3"
 
 [dev-dependencies]
+bytes = { version="0.5.0", default_features = false }
```
asomers: The `bytes` crate is now only used for examples and tests.  I could've removed it entirely, but since we explicitly supported it in version 0.10.0, I think we ought to use it in the examples as a kind of migration guide.

asomers:
Looks like using Cargo's patch feature only works some of the time.  Builds are randomly passing or failing based on whether Cargo can download `bytes`.  I'll fix it in the morning.

asomers:
This is not quite right; I'm working on a fix.

asomers:
I'm combining this with PR #872

PR #872, opened 2018-03-05
Change sys::aio::lio_listio to sys::aio::LioCb::listio

### LINKED ISSUE #870: lio_listio needs to use unique storage locations for the list argument
When delivering notifications via `kevent`, the kernel treats the address of the `list` argument as a the `ident` of the event.  Within the kernel, `ident` is treated as a unique identifier.  So there can only be one event with a given `ident` value in the kernel at any one time.  However, Nix's `sys::aio::lio_listio` allows the user to supply a temporary `Vec` as the `list` argument.  If the user calls `lio_listio` several times in a row, each with the same value of `list` (but different contents), `kevent` may deliver too few notifications.

The most fool-proof solution will probably be to create a `struct LioCb` object that owns all of the `AioCb` objects, and also owns the `list` variable as a `Vec`.  That will achieve two things:
1) The `LioCb` object will be forced to outlive the kernel's knowledge of it, since it owns the individual `AioCb`s
2) The `list` argument is guaranteed to be unique for all concurrent `lio_listio` calls, because `Vec` is stack-allocated and the `LioCb` object must outlive the kernel's knowledge of the event.

https://www.freebsd.org/cgi/man.cgi?lio_listio

asomers:
The new LioCb structure allows us to control the exact arguments passed
to lio_listio, guaranteeing that each call gets a unique storage
location for the list argument.  This prevents clients from misusing
lio_listio in a way that causes events to get dropped from a kqueue

Fixes #870

### DIFF
--- a/src/sys/aio.rs
+++ b/src/sys/aio.rs
@@ -965,63 +965,6 @@ pub fn aio_suspend(list: &[&AioCb], timeout: Option<TimeSpec>) -> Result<()> {
     }).map(drop)
 }
 
-
-/// Submits multiple asynchronous I/O requests with a single system call.
-///
-/// They are not guaranteed to complete atomically, and the order in which the
-/// requests are carried out is not specified.  Reads, writes, and fsyncs may be
-/// freely mixed.
-///
-/// This function is useful for reducing the context-switch overhead of
-/// submitting many AIO operations.  It can also be used with
-/// `LioMode::LIO_WAIT` to block on the result of several independent
-/// operations.  Used that way, it is often useful in programs that otherwise
-/// make little use of AIO.
-///
-/// # Examples
-///
-/// Use `lio_listio` to submit an aio operation and wait for its completion.  In
-/// this case, there is no need to use `aio_suspend` to wait or `AioCb#error` to
-/// poll.
-///
-/// ```
-/// # extern crate tempfile;
-/// # extern crate nix;
-/// # use nix::sys::aio::*;
-/// # use nix::sys::signal::SigevNotify;
-/// # use std::os::unix::io::AsRawFd;
-/// # use tempfile::tempfile;
-/// # fn main() {
-/// const WBUF: &[u8] = b"abcdef123456";
-/// let mut f = tempfile().unwrap();
-/// let mut aiocb = AioCb::from_slice( f.as_raw_fd(),
-///     2,   //offset
-///     WBUF,
-///     0,   //priority
-///     SigevNotify::SigevNone,
-///     LioOpcode::LIO_WRITE);
-/// lio_listio(LioMode::LIO_WAIT,
-///            &[&mut aiocb],
-///            SigevNotify::SigevNone).unwrap();
-/// assert_eq!(aiocb.aio_return().unwrap() as usize, WBUF.len());
-/// # }
-/// ```
-///
-/// # References
-///
-/// [`lio_listio`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/lio_listio.html)
-#[cfg(not(any(target_os = "ios", target_os = "macos")))]
-pub fn lio_listio(mode: LioMode, list: &[&mut AioCb],
-                  sigev_notify: SigevNotify) -> Result<()> {
-    let sigev = SigEvent::new(sigev_notify);
-    let sigevp = &mut sigev.sigevent() as *mut libc::sigevent;
-    let plist = list as *const [&mut AioCb] as *const [*mut libc::aiocb];
-    let p = plist as *const *mut libc::aiocb;
-    Errno::result(unsafe {
-        libc::lio_listio(mode as i32, p, list.len() as i32, sigevp)
-    }).map(drop)
-}
-
 impl<'a> Debug for AioCb<'a> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         fmt.debug_struct("AioCb")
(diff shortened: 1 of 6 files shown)


asomers:
I found the copy/paste error that was causing the tests to fail on armv7 and i686-musl

asomers:
I discovered a problem with this PR: it makes it difficult to handle `EIO`.  I'll fix it before merging.

src/sys/aio.rs
```diff
@@ -165,6 +158,50 @@ impl<'a> AioCb<'a> {
         x
     }
 
+    /// Remove the inner boxed slice, if any, and return it.
+    ///
+    /// The returned value will be the argument that was passed to
+    /// `from_boxed_slice` when this `AioCb` was created.
+    ///
+    /// It is an error to call this method while the `AioCb` is still in
```
Susurrus: Would it be possible to actually make this a state machine so this error condition can be caught at compile time? So use a different type for in-progress `AioCb`s and not-in-progress ones?

Susurrus:
`from_bytes_mut` is mentioned on 269, and it doesn't exist anymore, so that line of documentation should be updated.

src/sys/aio.rs
```diff
@@ -1042,6 +1038,247 @@ impl<'a> Drop for AioCb<'a> {
...
+    }
+
+    /// Retrieve error status of an individual `AioCb` submitted as part of an
+    /// `LioCb`.
+    ///
+    /// This is just like [`AioCb::error`], except it takes into account
+    /// operations that were restarted by [`LioCb::listio_resubmit`]
+    ///
+    /// [`AioCb::error`] (struct.AioCb.html#method.error)
+    /// [`LioCb::listio_resubmit`] #method.listio_resubmit
```
Susurrus: This link is incorrect when rendered using rustdoc.

src/sys/aio.rs
```diff
@@ -1042,6 +1038,247 @@ impl<'a> Drop for AioCb<'a> {
...
+    /// # }
+    /// ```
+    ///
+    /// # References
+    ///
+    /// [`lio_listio`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/lio_listio.html)
+    ///
+    /// [`AioCb::aio_return`] (struct.AioCb.html#method.aio_return)
+    /// [`AioCb::error`] (struct.AioCb.html#method.error)
+    /// [`LioCb::aio_return`] #method.aio_return
```
Susurrus: Incorrect link format.

src/sys/aio.rs
```diff
@@ -1042,6 +1038,247 @@ impl<'a> Drop for AioCb<'a> {
...
+    /// submitting many AIO operations.  It can also be used with
+    /// `LioMode::LIO_WAIT` to block on the result of several independent
+    /// operations.  Used that way, it is often useful in programs that
+    /// otherwise make little use of AIO.
+    ///
+    /// # Examples
+    ///
+    /// Use `listio` to submit an aio operation and wait for its completion.  In
+    /// this case, there is no need to use `aio_suspend` to wait or
+    /// `AioCb#error` to poll.
```
Susurrus: Incorrect doc link here also.

src/sys/aio.rs
asomers: Hm.  It's conceivable.  I would have to change the signature of methods that change the in_progress state to consume the object and return a new one, like this:
```rust
pub fn read(self) -> result::Result<AioCbInProgress, (Error, AioCb)> {
let p: *mut libc::aiocb = &mut self.aiocb;
    Errno::result(unsafe {
        libc::aio_read(p)
    }).map(|_| {
        AioCbInProgress::from_aiocb(self)
    }).map_err(|e| {
        (self, e)
    })
}
```
I would also have to put a `Box` around `AioCb::aiocb`, because the `libc::aiocb` must have a stable address once it's been passed to the kernel.

But I can't guarantee that the compiler will always catch the problem, because if `lio_listio` returns `EIO`, then some operations may be in-progress but not all.  Which means that `LioCb::aiocbs` would have to become a `Vec` of an `Enum` type.  So you wouldn't be able to catch misuse of an `LioCb` at compile time.

I think I could make such an arrangement work with the downsteam mio-aio and tokio-file crates, but they would likewise have to use `Enum` types, so they wouldn't gain anything.  tokio-file already has a totally different way of ensuring that you don't abuse an in-progress aiocb (based on runtime checks).

Do you think it's worth it?

src/sys/aio.rs
asomers: There are two bugs fixed with the new API.  #870 could cause kqueue to deliver too few notifications if you issued multiple `lio_listio` calls with the `list` argument having the same stack location.  The new API makes that impossible, because `LioCb` owns the list.

The other bug dealt with `lio_listio` returning `EAGAIN` or `EIO` because not all operations were enqueued because of resource limitations.  I actually have written a test for this in the tokio-file crate.  However, reliably hitting that resource limit requires knowing more information about the system, which I got with the `sysrc` crate.  I didn't want to import that crate into `nix` as a dev dependency, so I didn't add the test here.

src/sys/aio.rs
```diff
@@ -1184,9 +1187,8 @@ impl<'a> LioCb<'a> {
     ///
     /// [`lio_listio`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/lio_listio.html)
     ///
-    /// [`AioCb::aio_return`] (struct.AioCb.html#method.aio_return)
-    /// [`AioCb::error`] (struct.AioCb.html#method.error)
-    /// [`LioCb::aio_return`] #method.aio_return
+    /// [`lio_listio`]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/lio_listio.html
```
Susurrus: Why doesn't this use the parens format for external URL like on line 1188? Is there a reason to prefer one syntax over the other and/or is this actually incorrect?

src/sys/aio.rs
Susurrus: Why do you think that dependency is too much for our testing? I'm always wary of fixing subtle bugs but not having a test to make sure they don't re-occur.

src/sys/aio.rs
Susurrus: I don't particularly think either way, the idea crossed my mind so I asked. I really don't understand this API very well and I don't plan to spend the time to do so. But I wanted to ask in case you hadn't thought about it and it would be a good idea. Current API LGTM.

Susurrus:
I think just that last comment, then squash everything and this is GTM.

src/sys/aio.rs
asomers: In this rare case, I actually got it right.  Line 1188 is an inline link.  Lines 1190-1191 are footnotes.

src/sys/aio.rs
asomers: Ok, I pushed a test for it.

Susurrus:
Looks pretty good I think, though I didn't dig too deep into AIO/LIO, but it seems reasonable. I'd resolve the one last issue I raised, then squash and I'm happy!

test/sys/test_lio_listio_resubmit.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+        assert_eq!(liocb.aio_return(j).unwrap(), BYTES_PER_OP as isize);
+    }
+}
+
+// Deliberately exceed system resource limits, causing lio_listio to return EIO.
+// This test must run in its own process since it deliberately uses all AIO
+// resources.  ATM it is only enabled on FreeBSD, because I don't know how to
+// check system AIO limits on other operating systems.
```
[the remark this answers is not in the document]
asomers: Probably, but these limits are OS-specific, and I don't know the rules on anything but FreeBSD.  The errors, however, are standardized.  So as long as `listio_resubmit` does the right thing when it gets the error on FreeBSD, then it should do the right thing everywhere.

asomers:
Which "one last issue" is that?

Susurrus:
To add a TODO or FIXME to add an implementation for other platforms. But it sounds like you don't think they're necessary.

asomers:
I really don't.  And I don't know how to do it, either.  I can't find any documentation about limits for Linux, DragonFly or NetBSD.

Susurrus:
Please squash them r+ me.

PR #873, opened 2018-03-08
add SO_MARK SetSockOpt for Linux

mcginty:
SO_MARK allows traffic to be filtered by a "tag" using fwmark (see: https://www.linux.org/docs/man8/tc-fw.html).

Tested on Linux as root - the test will skip over when not being run as root, as I noticed a few other tests do this in the suite already.


asomers:
Looks good, but could you please add an entry to the CHANGELOG too?

mcginty:
Rebased and added an entry to CHANGELOG.

Issue #756, opened 2017-09-01
Sending multiple ControlMessage in a single Unix Socket packet fails


roblabla:
This is a continuation of #464 and #473. For some reason, the attached test still fails.

IIRC, there's a problem in the way we encode multiple Control Messages which causes the kernel to only consider the first Control Message sent. The specifics of where the bug is, I have no idea.

<details>
<summary>The failing test</summary>

```rust
#[test]
fn test_scm_rights_multiple_cmsg() {
    use std::os::unix::net::UnixDatagram;
    use std::os::unix::io::{RawFd, AsRawFd};
    use std::thread;
    use nix::sys::socket::{CmsgSpace, ControlMessage, MsgFlags, sendmsg, recvmsg};
    use nix::sys::uio::IoVec;

    let (send, receive) = UnixDatagram::pair().unwrap();
    let thread = thread::spawn(move || {
        let mut buf = [0u8; 8];
        let iovec = [IoVec::from_mut_slice(&mut buf)];
        let mut space = CmsgSpace::<([RawFd; 1], CmsgSpace<[RawFd; 1]>)>::new();
        let (data, fds) = match recvmsg(receive.as_raw_fd(), &iovec, Some(&mut space), MsgFlags::empty()) {
            Ok(msg) => {
                let mut iter = msg.cmsgs();
                let fdone;
                let fdtwo;
                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
                    fdone = fds[0];
                } else {
                    panic!();
                }
                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
                    fdtwo = fds[0];
                } else {
                    panic!();
                }
                (iovec[0].as_slice(), [fdone, fdtwo])
            },
            Err(_) => {
                panic!();
            }
        };
        assert_eq!(data, [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8]);
        assert_eq!(fds.len(), 2);
    });

    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
    let iov = [IoVec::from_slice(&slice)];
    let arrone = [0];
    let arrtwo = [1]; // pass stdin and stdout
    let cmsg = [ControlMessage::ScmRights(&arrone), ControlMessage::ScmRights(&arrtwo)];
    sendmsg(send.as_raw_fd(), &iov, &cmsg, MsgFlags::empty(), None).unwrap();
[...]

PR #874, opened 2018-03-19
sendmsg: Fix padding for ControlMessages, and zero it rather than leave it uninitialised

alecmocatta:
I hit a valgrind complaint sending a single fd over a single ControlMessage. Nix was handing a length including padding to the `sendmsg` syscall, but the padding bytes were unaddressable (i.e. not just uninitialised, but after the end of the Vec, which is not good).

This fixes that. It possibly fixes/affects #756, as the old behaviour of not padding between ControlMessages despite the padded length being what's passed to the syscall seemed wrong. That however is yet to be tested.

Feel free to not accept until I/someone else gets a chance to more thoroughly test, I'm making it so that if someone else bumps into this problem they have something to go on in the meantime.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -302,6 +302,18 @@ unsafe fn copy_bytes<'a, 'b, T: ?Sized>(src: &T, dst: &'a mut &'b mut [u8]) {
     mem::swap(dst, &mut remainder);
 }
 
+/// Pad byte slice dst with `len` zeroes, and update the slice to point to
+/// the remainder.
+fn pad_bytes(len: usize, dst: &mut &mut [u8]) {
+    let mut tmpbuf = &mut [][..];
+    mem::swap(&mut tmpbuf, dst);
+    let (padding, mut remainder) = tmpbuf.split_at_mut(len);
+    for i in padding {
+        *i = 0;
+    }
+    mem::swap(dst, &mut remainder);
+}
+
 cfg_if! {
     // Darwin and DragonFly BSD always align struct cmsghdr to 32-bit only.
     if #[cfg(any(target_os = "dragonfly", target_os = "ios", target_os = "macos"))] {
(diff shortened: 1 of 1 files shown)


asomers:
Thanks for the bug report.  Could you also please add a test case that demonstrates the problem?  Perhaps a unit test for `encode_into`?

przygienda:
Sigh, the encoding seems to be broken beyond a single cmsg. When several cmsg are given only first is encoded but the size blows up to the point kernel starts to throw back excessive length. On top copy seems to be weird as well, it seems to copy @ the end. In short, it shows there are not unit tests for that stuff ...  Yes, unit-test for encode_into needed ...

przygienda:
Did more work & IMO there's a clear bug when you add stuff where cmsg.len() is not = cmsg.size(). Here's the necessary diff ... After I fixed that one multiple CMESG started to work so it seems the padding alignment code is actually working ...  

```
pub fn sendmsg<'a>(fd: RawFd, iov: &[IoVec<&'a [u8]>], cmsgs: &[ControlMessage<'a>], flags: MsgFlags, addr: Option<&'a SockAddr>) -> Result<usize> {
    let mut capacity = 0;
    for cmsg in cmsgs {
        capacity += cmsg.space();
    }
    // Note that the resulting vector claims to have length == capacity,
    // so it's presently uninitialized.
    let mut cmsg_buffer = unsafe {
        let mut vec = Vec::<u8>::with_capacity(capacity);
        vec.set_len(capacity);
        vec
    };
    {
        let mut ofs = 0;
        for cmsg in cmsgs {
            let mut ptr = &mut cmsg_buffer[ofs ..];
            unsafe { cmsg.encode_into(&mut ptr) };
            ofs += cmsg.space();
        }
    }
```

asomers:
Thanks for tracking this down!  Could you please submit your change in PR form, along with a test case?

PR #876, opened 2018-03-24
add mlockall and munlockall

### LINKED ISSUE #875: mlockall/munlockall
It would be great if nix had mlockall/munlockall in addition to mlock/munlock.

I'd be happy to try and add it. Or is there a specific reason why it was omitted?

afck:
Closes #875

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -215,18 +215,25 @@ libc_bitflags!{
     }
 }
 
+/// Locks all memory pages that contain part of the address range with `length` bytes starting at
+/// `addr`. Locked pages never move to the swap area.
 pub unsafe fn mlock(addr: *const c_void, length: size_t) -> Result<()> {
     Errno::result(libc::mlock(addr, length)).map(drop)
 }
 
+/// Unlocks all memory pages that contain part of the address range with `length` bytes starting at
+/// `addr`.
 pub unsafe fn munlock(addr: *const c_void, length: size_t) -> Result<()> {
     Errno::result(libc::munlock(addr, length)).map(drop)
 }
 
+/// Locks all memory pages mapped into this process' address space. Locked pages never move to the
+/// swap area.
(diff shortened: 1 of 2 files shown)


src/sys/mman.rs
```diff
@@ -213,6 +223,14 @@ pub unsafe fn munlock(addr: *const c_void, length: size_t) -> Result<()> {
     Errno::result(libc::munlock(addr, length)).map(drop)
 }
 
+pub unsafe fn mlockall(flags: MlockAllFlags) -> Result<()> {
```
asomers: Why are these functions `unsafe`?

afck: You're right, I suppose there's nothing actually unsafe here. Done.

afck:
I conditionalized the constants for now.

asomers:
Could you please check Linux/S390x?  I'd rather not needlessly disable obscure platforms simply out of neglect.

afck:
I'm not sure how to verify that; it looks like the constants are defined in `tools/include/uapi/asm-generic/mman.h`, which is included in `tools/arch/s390/include/uapi/asm/mman.h`, so I guess it is? I'll have a look at the libc crate.

asomers:
libc merged the PR, so you should be able to update this PR as well.  No need to update Nix's Cargo.toml.

afck:
Thank you, I removed the conditions.

asomers:
Could you please add some doc comments for the new functions, too?

afck:
I also documented `mlock`/`munlock`. Please check whether you're okay with my formulation.

asomers:
Ok, I think we're done here.

bors r+

asomers:
Hm, bors's behavior has changed.  It appears that I now need to click "approve changes" _before_ telling bors to merge.

bors r+

asomers:
Let's retry.

bors r+

PR #877, opened 2018-03-28
Replace the gcc crate by cc

Eijebong:


asomers:
This looks good, but it only affects Dragonfly and we don't have CI coverage for Dragonfly.  Have you tested it on DragonflyBSD?

Eijebong:
Nah, I removed those cfg to try that the syntax was ok and that the output didn't change between the two crates.

asomers:
I'm going to approve this change on the basis that it looks right, and it doesn't break any of the tested platforms, and it has no effect on the Tier 2 platforms.

bors r+

PR #879, opened 2018-04-01
Add getrlimit(2) and setrlimit(2)

### LINKED ISSUE #878: Add getrlimit(2)
I'm currently looking for a way to increase the maximum number of open file descriptors, it seems there's no crate yet besides the raw syscall in `libc`.

There was #364 which didn't get merged though due to some other setrlimit related problems.

kpcyrd:
This is a followup of #364 by @kamalmarhubi. The api has been changed to `(Option<rlim_t>, Option<rlim_t>)` to hide the rlim struct. This allows:

```rust
let (rlim_cur, rlim_max) = getrlimit()?;
```

Resolves #878.

### DIFF
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -18,11 +18,11 @@ libc_enum!{
         RLIMIT_STACK,
 
         // BSDs and Linux
-        #[cfg(all(unix, not(target_os = "solaris")))]
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "openbsd"))]
         RLIMIT_MEMLOCK,
-        #[cfg(all(unix, not(target_os = "solaris")))]
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "openbsd"))]
         RLIMIT_NPROC,
-        #[cfg(all(unix, not(target_os = "solaris")))]
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "openbsd"))]
         RLIMIT_RSS,
 
         // Android and Linux only
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -60,6 +60,18 @@ libc_enum!{
 ///
 /// * `resource`: The [`Resource`] that we want to get the limits of.
 ///
+/// # Examples
+///
+/// ```
+/// use nix::sys::resource::{getrlimit, Resource};
+///
+/// fn main() {
+///     let (soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+///     println!("current soft_limit: {:?}", soft_limit);
+///     println!("current hard_limit: {:?}", hard_limit);
+/// }
+/// ```
+///
 /// # References
 ///
 /// [getrlimit(2)](https://linux.die.net/man/2/getrlimit)
@@ -84,6 +96,18 @@ pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>
 /// * `soft_limit`: The value that the kenrel enforces for the corresponding resource.
 /// * `hard_limit`: The ceiling for the soft limit. Must be lower or equal to the current hard limit
 ///   for non-root users.
+///
+/// # Examples
+///
+/// ```no_run
+/// use nix::sys::resource::{setrlimit, Resource};
+///
+/// fn main() {
+///     let soft_limit = Some(1024);
+///     let hard_limit = None;
+///     setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit).unwrap();
+/// }
+/// ```
 ///
 /// # References
 ///
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -54,7 +54,7 @@ libc_enum!{
 
 /// Get the current processes resource limits
 ///
-/// A value of `None` corresponds to `RLIM_INFINITY`, which means there's no limit.
+/// A value of None indicates that there's no limit.
 ///
 /// # Parameters
 ///
@@ -64,11 +64,10 @@ libc_enum!{
 ///
 /// ```
 /// # use nix::sys::resource::{getrlimit, Resource};
-/// # fn main() {
+///
 /// let (soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
 /// println!("current soft_limit: {:?}", soft_limit);
 /// println!("current hard_limit: {:?}", hard_limit);
-/// # }
 /// ```
 ///
 /// # References
@@ -79,6 +78,7 @@ libc_enum!{
 pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)> {
     let mut rlim: rlimit = unsafe { mem::uninitialized() };
     let res = unsafe { libc::getrlimit(resource as c_int, &mut rlim as *mut _) };
+    // TODO: use Option::filter after it has been stabilized
     Errno::result(res).map(|_| {
         (if rlim.rlim_cur != RLIM_INFINITY { Some(rlim.rlim_cur) } else { None },
          if rlim.rlim_max != RLIM_INFINITY { Some(rlim.rlim_max) } else { None })
@@ -87,7 +87,7 @@ pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>
 
 /// Set the current processes resource limits
 ///
-/// A value of `None` corresponds to `RLIM_INFINITY`, which means there's no limit.
+/// A value of None indicates that there's no limit.
 ///
 /// # Parameters
 ///
@@ -100,11 +100,10 @@ pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>
 ///
 /// ```no_run
 /// # use nix::sys::resource::{setrlimit, Resource};
-/// # fn main() {
+///
 /// let soft_limit = Some(1024);
 /// let hard_limit = None;
 /// setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit).unwrap();
-/// # }
 /// ```
 ///
 /// # References
@@ -118,5 +117,5 @@ pub fn setrlimit(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Opt
     rlim.rlim_max = hard_limit.unwrap_or(RLIM_INFINITY);
 
     let res = unsafe { libc::setrlimit(resource as c_int, &rlim as *const _) };
-    Errno::result(res).map(drop)
+    Errno::result(res).map(|_| ())
 }
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -7,46 +7,67 @@ pub use libc::rlim_t;
 use {Errno, Result};
 
 libc_enum!{
+    /// A resource that limits apply to
     #[repr(i32)]
     pub enum Resource {
         // POSIX
+        /// This is the maximum size of the process's virtual memory (address space). The limit is specified in bytes, and is rounded down to the system page size.
         RLIMIT_AS,
+        /// This is the maximum size of a core file (see core(5)) in bytes that the process may dump.
         RLIMIT_CORE,
+        /// This is a limit, in seconds, on the amount of CPU time that the process can consume.
         RLIMIT_CPU,
+        /// This is the maximum size of the process's data segment (initialized data, uninitialized data, and heap). The limit is specified in bytes, and is rounded down to the system page size.
         RLIMIT_DATA,
+        /// This is the maximum size in bytes of files that the process may create. Attempts to extend a file beyond this limit result in delivery of a SIGXFSZ signal.
         RLIMIT_FSIZE,
+        /// This specifies a value one greater than the maximum file descriptor number that can be opened by this process.
         RLIMIT_NOFILE,
+        /// This is the maximum size of the process stack, in bytes. Upon reaching this limit, a SIGSEGV signal is generated.
         RLIMIT_STACK,
 
         // BSDs and Linux
+        /// This is the maximum number of bytes of memory that may be locked into RAM. This limit is in effect rounded down to the nearest multiple of the system page size.
         #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "openbsd"))]
         RLIMIT_MEMLOCK,
+        /// This is a limit on the number of extant process (or, more precisely on Linux, threads) for the real user ID of the calling process.
         #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "openbsd"))]
         RLIMIT_NPROC,
+        /// This is a limit (in bytes) on the process's resident set (the number of virtual pages resident in RAM).
         #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux", target_os = "openbsd"))]
         RLIMIT_RSS,
 
         // Android and Linux only
+        /// This is a limit on the combined number of flock(2) locks and fcntl(2) leases that this process may establish.
         #[cfg(any(target_os = "android", target_os = "linux"))]
         RLIMIT_LOCKS,
+        /// This is a limit on the number of bytes that can be allocated for POSIX message queues for the real user ID of the calling process.
         #[cfg(any(target_os = "android", target_os = "linux"))]
         RLIMIT_MSGQUEUE,
+        /// This specifies a ceiling to which the process's nice value can be raised using setpriority(2) or nice(2). The actual ceiling for the nice value is calculated as 20 - rlim_cur.
         #[cfg(any(target_os = "android", target_os = "linux"))]
         RLIMIT_NICE,
+        /// This specifies a ceiling on the real-time priority that may be set for this process using sched_setscheduler(2) and sched_setparam(2).
         #[cfg(any(target_os = "android", target_os = "linux"))]
         RLIMIT_RTPRIO,
+        /// This is a limit (in microseconds) on the amount of CPU time that a process scheduled under a real-time scheduling policy may consume without making a blocking system call.
         #[cfg(any(target_os = "android", target_os = "linux"))]
         RLIMIT_RTTIME,
+        /// This is a limit on the number of signals that may be queued for the real user ID of the calling process. Both standard and real-time signals are counted for the purpose of checking this limit.
         #[cfg(any(target_os = "android", target_os = "linux"))]
         RLIMIT_SIGPENDING,
 
         // Available on some BSD
+        /// The maximum number of kqueues this user id is allowed to create.
         #[cfg(target_os = "freebsd")]
         RLIMIT_KQUEUES,
+        /// The maximum number of pseudo-terminals this user id is allowed to create.
         #[cfg(target_os = "freebsd")]
         RLIMIT_NPTS,
+        /// The maximum size (in bytes) of socket buffer usage for this user.
         #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
         RLIMIT_SBSIZE,
+        /// The maximum size (in bytes) of the swap space that may be reserved or used by all of this user id's processes.
         #[cfg(target_os = "freebsd")]
         RLIMIT_SWAP,
     }
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -5,39 +5,40 @@ pub use libc::rlim_t;
 
 use {Errno, Result};
 
-#[derive(Clone, Copy, PartialEq, Eq, Debug)]
-#[repr(i32)]
-pub enum Resource {
-    // POSIX
-    RLIMIT_AS = libc::RLIMIT_AS,
-    RLIMIT_CORE = libc::RLIMIT_CORE,
-    RLIMIT_CPU = libc::RLIMIT_CPU,
-    RLIMIT_DATA = libc::RLIMIT_DATA,
-    RLIMIT_FSIZE = libc::RLIMIT_FSIZE,
-    RLIMIT_NOFILE = libc::RLIMIT_NOFILE,
-    RLIMIT_STACK = libc::RLIMIT_STACK,
+libc_enum!{
+    #[repr(i32)]
+    pub enum Resource {
+        // POSIX
+        RLIMIT_AS,
+        RLIMIT_CORE,
+        RLIMIT_CPU,
+        RLIMIT_DATA,
+        RLIMIT_FSIZE,
+        RLIMIT_NOFILE,
+        RLIMIT_STACK,
 
-    // BSDs and Linux
-    #[cfg(all(unix, not(target_os = "solaris")))]
-    RLIMIT_MEMLOCK = libc::RLIMIT_MEMLOCK,
-    #[cfg(all(unix, not(target_os = "solaris")))]
-    RLIMIT_NPROC = libc::RLIMIT_NPROC,
-    #[cfg(all(unix, not(target_os = "solaris")))]
-    RLIMIT_RSS = libc::RLIMIT_RSS,
+        // BSDs and Linux
+        #[cfg(all(unix, not(target_os = "solaris")))]
+        RLIMIT_MEMLOCK,
+        #[cfg(all(unix, not(target_os = "solaris")))]
+        RLIMIT_NPROC,
+        #[cfg(all(unix, not(target_os = "solaris")))]
+        RLIMIT_RSS,
 
-    // Android and Linux only
-    #[cfg(any(target_os = "android", target_os = "linux"))]
-    RLIMIT_LOCKS = libc::RLIMIT_LOCKS,
-    #[cfg(any(target_os = "android", target_os = "linux"))]
-    RLIMIT_MSGQUEUE = libc::RLIMIT_MSGQUEUE,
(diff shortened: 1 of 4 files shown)


src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
+use std::mem;
+
+use libc::{self, c_int};
+pub use libc::{rlimit, rlim_t, RLIM_INFINITY};
+
+#[cfg(any(target_os = "linux",
+          target_os = "openbsd",
+          target_os = "netbsd",
+          target_os = "bitrig"))]
```
Susurrus: Bitrig no longer exists, do go ahead and drop this. But please add `fuchsia` and `empscripten`, as they also support this API.

Additionally these targets should be alphabetized.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+pub use libc::{RLIM_SAVED_CUR, RLIM_SAVED_MAX};
+
+use {Errno, Result};
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(i32)]
+pub enum Resource {
```
Susurrus: Please use the `libc_enum!` macro instead.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[inline]
+fn rlim_to_option(rlim: rlim_t) -> Option<rlim_t> {
+    match rlim {
+        RLIM_INFINITY => None,
+        rlim => Some(rlim),
+    }
+}
+
+pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)> {
```
Susurrus: Needs a doccomment, should have an example as well.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    let mut rlim: rlimit = unsafe { mem::uninitialized() };
+    let res = unsafe { libc::getrlimit(resource as c_int, &mut rlim as *mut _) };
+    Errno::result(res)?;
+    Ok((rlim_to_option(rlim.rlim_cur), rlim_to_option(rlim.rlim_max)))
+}
+
+pub fn setrlimit(resource: Resource, limit: (Option<rlim_t>, Option<rlim_t>)) -> Result<()> {
```
Susurrus: I'm not a fan of this API because it's hard to know what `limit` means. I'd make this take a `soft_limit` and a `hard_limit` which are both `Option<rlim_t>`s so it's more explicit.

Additionally this should have doccomments and an example showing how this would be used (can be same as the one for `getrlimit`.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    let mut rlim: rlimit = unsafe { mem::uninitialized() };
+    rlim.rlim_cur = limit.0.unwrap_or(RLIM_INFINITY);
+    rlim.rlim_max = limit.1.unwrap_or(RLIM_INFINITY);
+
+    let res = unsafe { libc::setrlimit(resource as c_int, &rlim as *const _) };
+    Errno::result(res).map(drop)
```
Susurrus: Please do `Errno::result(res).map(|_| ())` as it's more clear about what it's actually doing.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    // POSIX
+    RLIMIT_CORE = libc::RLIMIT_CORE,
+    RLIMIT_CPU = libc::RLIMIT_CPU,
+    RLIMIT_DATA = libc::RLIMIT_DATA,
+    RLIMIT_FSIZE = libc::RLIMIT_FSIZE,
+    RLIMIT_NOFILE = libc::RLIMIT_NOFILE,
+    RLIMIT_STACK = libc::RLIMIT_STACK,
+    RLIMIT_AS = libc::RLIMIT_AS,
+    // BSDs and Linux
+    #[cfg(all(unix, not(target_os = "solaris")))]
```
Susurrus: Please don't use `not`, explicitly list targets here instead.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_RTPRIO = libc::RLIMIT_RTPRIO,
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_RTTIME = libc::RLIMIT_RTTIME,
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_SIGPENDING = libc::RLIMIT_SIGPENDING,
+}
+
+#[inline]
+fn rlim_to_option(rlim: rlim_t) -> Option<rlim_t> {
```
Susurrus: Is there not a better API to do this? I would think this would be a common enough idiom that there's a function somewhere in the Rust API to do this already.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    RLIMIT_AS = libc::RLIMIT_AS,
+    // BSDs and Linux
+    #[cfg(all(unix, not(target_os = "solaris")))]
+    RLIMIT_MEMLOCK = libc::RLIMIT_MEMLOCK,
+    #[cfg(all(unix, not(target_os = "solaris")))]
+    RLIMIT_NPROC = libc::RLIMIT_NPROC,
+    #[cfg(all(unix, not(target_os = "solaris")))]
+    RLIMIT_RSS = libc::RLIMIT_RSS,
+    // Linux-only
+    #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: These also need to be alphabetized.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    RLIMIT_STACK = libc::RLIMIT_STACK,
+    RLIMIT_AS = libc::RLIMIT_AS,
+    // BSDs and Linux
+    #[cfg(all(unix, not(target_os = "solaris")))]
+    RLIMIT_MEMLOCK = libc::RLIMIT_MEMLOCK,
+    #[cfg(all(unix, not(target_os = "solaris")))]
+    RLIMIT_NPROC = libc::RLIMIT_NPROC,
+    #[cfg(all(unix, not(target_os = "solaris")))]
+    RLIMIT_RSS = libc::RLIMIT_RSS,
+    // Linux-only
```
Susurrus: "Android and Linux only"

test/test_resource.rs
```diff
@@ -0,0 +1,22 @@
+use nix::sys::resource::{Resource, getrlimit, setrlimit};
+
+#[test]
+pub fn test_resource_limits() {
+    let mut limit = getrlimit(Resource::RLIMIT_STACK).unwrap();
+    assert!(limit.0 != limit.1);
```
asomers: This assertion need not hold universally.  On FreeBSD for example, the default hard limits and soft limits seem to be identical.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
+use std::mem;
+
+use libc::{self, c_int};
+pub use libc::{rlimit, rlim_t, RLIM_INFINITY};
```
asomers: I understand why you're reexporting `rlim_t`, but why reexport the others?

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[cfg(any(target_os = "linux",
+          target_os = "openbsd",
+          target_os = "netbsd",
+          target_os = "bitrig"))]
+pub use libc::{RLIM_SAVED_CUR, RLIM_SAVED_MAX};
```
asomers: Why reexport these symbols?  For that matter, why import them?  It seems like nothing is using them.

src/sys/resource.rs
```diff
@@ -0,0 +1,69 @@
...
+    RLIMIT_MSGQUEUE = libc::RLIMIT_MSGQUEUE,
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_NICE = libc::RLIMIT_NICE,
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_RTPRIO = libc::RLIMIT_RTPRIO,
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_RTTIME = libc::RLIMIT_RTTIME,
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    RLIMIT_SIGPENDING = libc::RLIMIT_SIGPENDING,
+}
```
asomers: There are more resources on non-Linux platforms, such as `RLIMIT_KQUEUES`, `RLIMIT_NPTS`, `RLIMIT_SBSIZE`, and `RLIMIT_SWAP`.

src/sys/resource.rs
kpcyrd: I've changed the api to
```rust
setrlimit(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Option<rlim_t>) -> Result<()>
```
I'd prefer to delay documentation until the api itself is approved.

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: I've changed it to a rust-style ternary expression. Option.filter would work, but is nightly only.

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

test/test_resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done, limited the reexport to `rlim_t`

src/sys/resource.rs
kpcyrd: Done, removed the import.

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done, removed the import.

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
```diff
@@ -1,68 +1,69 @@
...
+        RLIMIT_RTTIME,
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        RLIMIT_SIGPENDING,
 
-#[inline]
-fn rlim_to_option(rlim: rlim_t) -> Option<rlim_t> {
-    match rlim {
-        RLIM_INFINITY => None,
-        rlim => Some(rlim),
+        // Non-Linux
```
asomers: The comment makes it sound like these symbols are available everywhere except Linux.

src/sys/resource.rs
```diff
@@ -1,68 +1,69 @@
...
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_KQUEUES,
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_NPTS,
+        #[cfg(target_os = "freebsd")]
```
asomers: `RLIMIT_SBSIZE` is also available on Dragonfly.

test/test_resource.rs
```diff
@@ -2,19 +2,17 @@ use nix::sys::resource::{Resource, getrlimit, setrlimit};
 
 #[test]
 pub fn test_resource_limits() {
-    let mut limit = getrlimit(Resource::RLIMIT_STACK).unwrap();
-    assert!(limit.0 != limit.1);
+    let (mut soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_STACK).unwrap();
+    let orig_limit = (soft_limit, hard_limit);
```
asomers: Your changes to the test do fix the previous failure, but the coverage isn't very good, since the soft and hard limits are often identical.  How about setting `RLIMIT_NOFILE` to a hard-coded value of 1024 if the current soft limit is unlimited, or the current softlimit - 1 otherwise?

kpcyrd: Done

src/sys/resource.rs
kpcyrd: Changed to `// Available on some BSD`

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
```diff
@@ -39,18 +40,43 @@ libc_enum!{
...
+/// A value of `None` corresponds to `RLIM_INFINITY`, which means there's no limit.
+///
+/// # Parameters
+///
+/// * `resource`: The [`Resource`] that we want to get the limits of.
+///
+/// # Examples
+///
+/// ```
+/// use nix::sys::resource::{getrlimit, Resource};
```
asomers: In the doc test, you should put a `#` before the `use`, `fn main`, and final `}` lines so they will be hidden from the documentation.  Likewise, omit the blank line before `main`.

src/sys/resource.rs
```diff
@@ -60,6 +86,34 @@ pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>
...
 }
 
+/// Set the current processes resource limits
+///
+/// A value of `None` corresponds to `RLIM_INFINITY`, which means there's no limit.
+///
+/// # Parameters
+///
+/// * `resource`: The [`Resource`] that we want to set the limits of.
+/// * `soft_limit`: The value that the kenrel enforces for the corresponding resource.
```
asomers: s/kenrel/kernel

test/test_resource.rs
```diff
@@ -1,7 +1,22 @@
...
+
+    // make sure the soft limit and hard limit are not equal
+    let soft_limit = match soft_limit {
+        Some(nofile) => Some(nofile -1),
+        None => Some(1024),
+    };
+    setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit).unwrap();
+
+    let (new_soft_limit, new_hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+    assert!(new_soft_limit != new_hard_limit);
```
asomers: This should be `assert!(new_soft_limit, soft_limit)`

test/test_resource.rs
asomers: Do we really need two tests?  Does the stack test provide any extra coverage over the `RLIMIT_NOFILE` test?

test/test_resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use libc::{self, c_int, rlimit, RLIM_INFINITY};
+pub use libc::rlim_t;
+
+use {Errno, Result};
+
+libc_enum!{
+    #[repr(i32)]
+    pub enum Resource {
+        // POSIX
```
Susurrus: Please provide doc comments for every entry here to give users an understanding of what they mean.

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+use std::mem;
+
+use libc::{self, c_int, rlimit, RLIM_INFINITY};
+pub use libc::rlim_t;
+
+use {Errno, Result};
+
+libc_enum!{
+    #[repr(i32)]
+    pub enum Resource {
```
Susurrus: This enum needs a doccomment as well.

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+        #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+        RLIMIT_SBSIZE,
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_SWAP,
+    }
+}
+
+/// Get the current processes resource limits
+///
+/// A value of `None` corresponds to `RLIM_INFINITY`, which means there's no limit.
```
Susurrus: I'd just remove the `RLIM_INFINITY` reference. For the user that implementation detail is completely hidden: "A value of `None` indicates that there's no limit."

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// # Parameters
+///
+/// * `resource`: The [`Resource`] that we want to get the limits of.
+///
+/// # Examples
+///
+/// ```
+/// # use nix::sys::resource::{getrlimit, Resource};
+/// # fn main() {
```
Susurrus: You shouldn't need to declare `main()` here.

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+/// [setrlimit(2)](https://linux.die.net/man/2/setrlimit)
+///
+/// [`Resource`]: enum.Resource.html
+pub fn setrlimit(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Option<rlim_t>) -> Result<()> {
+    let mut rlim: rlimit = unsafe { mem::uninitialized() };
+    rlim.rlim_cur = soft_limit.unwrap_or(RLIM_INFINITY);
+    rlim.rlim_max = hard_limit.unwrap_or(RLIM_INFINITY);
+
+    let res = unsafe { libc::setrlimit(resource as c_int, &rlim as *const _) };
+    Errno::result(res).map(drop)
```
Susurrus: This should be `map(|_| ())`.

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+/// ```
+///
+/// # References
+///
+/// [setrlimit(2)](https://linux.die.net/man/2/setrlimit)
+///
+/// [`Resource`]: enum.Resource.html
+pub fn setrlimit(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Option<rlim_t>) -> Result<()> {
+    let mut rlim: rlimit = unsafe { mem::uninitialized() };
+    rlim.rlim_cur = soft_limit.unwrap_or(RLIM_INFINITY);
```
Susurrus: Should this be checked to be less than `rlim_max` or should that be deferred to the `setrlimit` call?

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+    let res = unsafe { libc::getrlimit(resource as c_int, &mut rlim as *mut _) };
+    Errno::result(res).map(|_| {
+        (if rlim.rlim_cur != RLIM_INFINITY { Some(rlim.rlim_cur) } else { None },
+         if rlim.rlim_max != RLIM_INFINITY { Some(rlim.rlim_max) } else { None })
+    })
+}
+
+/// Set the current processes resource limits
+///
+/// A value of `None` corresponds to `RLIM_INFINITY`, which means there's no limit.
```
Susurrus: Same as above, don't bother mentioning `RLIM_INFINITY`

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+/// * `resource`: The [`Resource`] that we want to set the limits of.
+/// * `soft_limit`: The value that the kernel enforces for the corresponding resource.
+/// * `hard_limit`: The ceiling for the soft limit. Must be lower or equal to the current hard limit
+///   for non-root users.
+///
+/// # Examples
+///
+/// ```no_run
+/// # use nix::sys::resource::{setrlimit, Resource};
+/// # fn main() {
```
Susurrus: I don't think you need the main here.

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// # References
+///
+/// [getrlimit(2)](https://linux.die.net/man/2/getrlimit)
+///
+/// [`Resource`]: enum.Resource.html
+pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)> {
+    let mut rlim: rlimit = unsafe { mem::uninitialized() };
+    let res = unsafe { libc::getrlimit(resource as c_int, &mut rlim as *mut _) };
+    Errno::result(res).map(|_| {
```
Susurrus: Ugly, but it works. Can you put a `TODO` in here to switch over to that other API you mentioned once it hits stable?

test/test_resource.rs
```diff
@@ -0,0 +1,35 @@
+use nix::sys::resource::{Resource, getrlimit, setrlimit};
+
+#[test]
+pub fn test_resource_limits_nofile() {
+    let (soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+
+    // make sure the soft limit and hard limit are not equal
+    let soft_limit = match soft_limit {
+        Some(nofile) => Some(nofile -1),
```
Susurrus: Need a space after the `-`

test/test_resource.rs
```diff
@@ -0,0 +1,35 @@
+use nix::sys::resource::{Resource, getrlimit, setrlimit};
+
+#[test]
+pub fn test_resource_limits_nofile() {
```
Susurrus: Can you summarize what this test is doing? Makes it easier to fix or modify them later.

src/sys/resource.rs
```diff
@@ -0,0 +1,122 @@
...
+///   for non-root users.
+///
+/// # Examples
+///
+/// ```no_run
+/// # use nix::sys::resource::{setrlimit, Resource};
+/// # fn main() {
+/// let soft_limit = Some(1024);
+/// let hard_limit = None;
+/// setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit).unwrap();
```
Susurrus: Is this a realistic use case auser might actually do? If not, we should try to find one more common to use as an example and provide a nice description of it before the codeblock.

test/test_resource.rs
```diff
@@ -0,0 +1,35 @@
...
+        None => Some(1024),
+    };
+    setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit).unwrap();
+
+    let (new_soft_limit, new_hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+    assert_eq!(new_soft_limit, soft_limit);
+}
+
+#[test]
+pub fn test_resource_limits_stack() {
```
Susurrus: Could you provide some comments or a general summarizing comment here? It makes it easier for us reviewers to understand and to modify/fix them later.

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

test/test_resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

src/sys/resource.rs
kpcyrd: Done

kpcyrd:
I'm having trouble coming up with a cross plattform compatible test that is also a realistic test case. Could you suggest one?

src/sys/resource.rs
```diff
@@ -0,0 +1,142 @@
...
+/// ```
+///
+/// # References
+///
+/// [setrlimit(2)](https://linux.die.net/man/2/setrlimit)
+///
+/// [`Resource`]: enum.Resource.html
+pub fn setrlimit(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Option<rlim_t>) -> Result<()> {
```
kamalmarhubi: I came in here to suggest changing this to
```rust
pub fn setrlimit<S=rlim_t, H=rlim_t>(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Option<rlim_t>) -> Result<()>
    where S: Into<Option<rlim_t>>, H: Into<Option<rlim_t>> { // ...
```
playground example

to take advantage for the `impl<T> From<T> for Option<T>` impl added in 1.12: https://github.com/rust-lang/rust/pull/34828.

*but* it turns out we can't use default type params in functions on stable. There's a WG looking into this, so maybe we can open it up later this year. (This would allow `setrlimit(RLIMIT_RSS, 1024 * 1024, None)` without the `Some` wrapping on the first arg.)

Susurrus:
A test doesn't need to run on all platforms, though it's nice if it does. It'd be good to get at least a test working on one of our Tier 1 platforms so it'll run regularly in CI tho.

src/sys/resource.rs
kamalmarhubi: I would prefer to leave this to the kernel.

Susurrus:
@kpcyrd friendly pin on this.

kpcyrd:
This issue is up for grabs if anybody is interested in it, but I might come back to it later. I'm currently using my own solution in the program I was working on when filing this PR.

rtzoeller:
Implemented by #1302.

PR #880, opened 2018-04-04
Update changelog for #869

Detegr:
Updated changelog to contain #869


Susurrus:
This was a bug fix though, correct? If so, this should be under the `Fixed` section. Probably should be stated as "Fixed panics when using `SigAction::flags()` in some situations on Linux". Does that seem like a succinct and correct description?

Detegr:
Sure. I'm sorry this took so long.

Susurrus:
Thanks, Detegr!

bors r+

PR #881, opened 2018-04-10
Implemented FromStr for Uid, Gid and Pid

Kixunil:
Example use case: parsing command line arguments and using it in `setuig`.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -13,6 +13,7 @@ use std::os::unix::io::RawFd;
 use std::path::{PathBuf};
 use void::Void;
 use sys::stat::Mode;
+use std::str::FromStr;
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 pub use self::pivot_root::*;
@@ -62,6 +63,14 @@ impl fmt::Display for Uid {
     }
 }
 
+impl FromStr for Uid {
+    type Err = <uid_t as FromStr>::Err;
+
+    fn from_str(s: &str) -> ::std::result::Result<Self, Self::Err> {
+        s.parse().map(Uid)
+    }
+}
+
 /// Constant for UID = 0
 pub const ROOT: Uid = Uid(0);
 
@@ -101,6 +110,14 @@ impl fmt::Display for Gid {
     }
 }
 
+impl FromStr for Gid {
+    type Err = <gid_t as FromStr>::Err;
+
+    fn from_str(s: &str) -> ::std::result::Result<Self, Self::Err> {
+        s.parse().map(Gid)
+    }
+}
+
 /// Process identifier
 ///
 /// Newtype pattern around `pid_t` (which is just alias). It prevents bugs caused by accidentally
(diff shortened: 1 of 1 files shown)


Susurrus:
Could you better explain what your use case is here? Specifically how's your argument specified for your app and what are you doing now to transform the string into the proper type for `setuig`?

Kixunil:
My real-world use case is to bind a web server on port 80 (root required) and then switch UID and GID to the one specified on the command line. I now do it by parsing it as `u32` and then converting, but parsing just `Uid` is nicer and more convenient.

The ability to parse `Uid` and others also makes sense because they can be displayed too and a reasonable mapping from `&str` to `Uid` exists.

Susurrus:
I think your point is sound, but I'd like to see our documentation fleshed out to support people finding and using this functionality. We have a lack of good documentation in `nix` and I'm requiring all PRs to add decent documentation to get added. So for this PR I'd like to see the following:

 * Module-level documentation for these items discussing both `FromStr` and `ToStr` for them. This should include specific examples, like your use case.
 * Tests for both `FromStr` and `ToStr` that include tests that should fail.

Kixunil:
I like the effort to make better documentation. I'm not sure if it's actually needed in this case, since `FromStr` is a standard trait from `std` and it's itself documented. I'll try to think of something reasonable and add it though.

What exactly do you mean by "tests that should fail"? Do you mean attempt to parse garbage should lead to error or something else?

Susurrus:
> I like the effort to make better documentation. I'm not sure if it's actually needed in this case, since FromStr is a standard trait from std and it's itself documented. I'll try to think of something reasonable and add it though.

Documentation has many facets. People think that if something is written, the job is done. But the goal here isn't to document `FromStr`, as you said, it should be already or self-documented. The goal here is to make this functionality *discoverable*. By discussing this use case in the module docs, it's easy for newcomers to the library, or even just this API, to find out this valuable functionality.

> What exactly do you mean by "tests that should fail"? Do you mean attempt to parse garbage should lead to error or something else?

For all functions there are valid and invalid inputs. This should be well-defined by the API. As part of good testing, the boundaries between valid and invalid inputs should be checked with the goal that a) everyone is clear on where that boundary lies and b) that boundary doesn't change during a refactoring. So if you only test *good* inputs, the space of good inputs could expand inappropriately, and no tests would fail. So it's critical that *bad* inputs are also checked that they remain bad.

kamalmarhubi:
I think as a general principle for nix types that are just newtype wrappers:
- we should impl most all traits that are defined on the wrapped type
- we can take a different approach to testing

Re the first, this seems like a good step. I would particularly like to see us also add `impl ::std::ops::Neg for Pid` to allow use in `kill` to send a signal to all processes in a process group.

Re the second: if we're just wrapping, we can add generative tests (quickcheck or similar) that assert that `Pid::from_str(s).map(|pid| pid.0) == pid_t::from_str(s)`. A tiny bit of care is needed to be sure that we hit both error and success cases in the generated tests.

Susurrus:
I'm still leaning towards not adding this implementation. These are things I think should be handled within the application, not the library, because it ties in with parsing user input. I don't think that's a business that `nix` should be in. Given that the workaround here is the one-liner of `uid.parse().map(Uid)`, it seems like very low overhead to have that integrated with the rest of the command-line argument parsing code that'd exist in your application. @asomers do you have any thoughts here before I close this?

asomers:
Yeah, I agree with @Susurrus .

PR #884, opened 2018-04-15
impl FromStr for sys::Signal

quodlibetor:
This implements both ALLCAPS and lowercase full name (including SIG) and short
name.

This matches what `kill` accepts in many shells, and it also allows the `Debug`
representation of `Signal`, which means it can be round-tripped, if desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#857](https://github.com/nix-rust/nix/pull/857))
 - Added `request_code_write_int!` on FreeBSD/DragonFlyBSD
   ([#833](https://github.com/nix-rust/nix/pull/833))
+- Added `FromStr` and `Display` impls for `nix::sys::Signal`
+  ([#884](https://github.com/nix-rust/nix/pull/884))
 
 ### Changed
 - Display and Debug for SysControlAddr now includes all fields.
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -5,6 +5,8 @@ use libc;
 use {Error, Result};
 use errno::Errno;
 use std::mem;
+use std::fmt;
+use std::str::FromStr;
 #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
 use std::os::unix::io::RawFd;
 use std::ptr;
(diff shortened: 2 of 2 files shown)


quodlibetor:
The failure looks spurious:

```text
error: failed to build archive: No space left on device
```

asomers:
I fixed the ENOSPC and restarted the builds.

quodlibetor:
Fair enough. Would you be more receptive to a change that only accepts a subset of these spellings? For example, just the exact matches (only SIGHUP, not SIGHUP and HUP and sighup and hup)? I'm happy to adjust. This is also just a way to reduce boilerplate and a newtype in one of my projects so I wouldn't be terribly upset if you just want to reject this.

Susurrus:
I don't like that there isn't a `ToStr` implementation to go with this. We should either have both or neither. I'm also of the opinion that there should only be one canonical representation of each signal, the full capitalized version.

@quodlibetor What's your use case here? You want to parse command-line arguments specifying signals or are these strings in a data stream?

quodlibetor:
@Susurrus I'm happy to implement `Display` and `AsRef<str>` for these. My use case is command-line flags ([here](https://github.com/quodlibetor/tabin-plugins/pull/14/files)).

This PR [verifies that the `Debug` impl does roundtrip](https://github.com/nix-rust/nix/pull/884/files#diff-d755fdcb8a25da8202c15b0e91e71a83R711), which isn't as neat as adding `to_string` but does work. I agree that round-tripping is important.

I would slightly prefer to keep both all-caps versions (e.g. `SIGHUP` and `HUP`), but that's just because I think of them as synonyms  and I would be happy to remove everything if coming in with a limited feature spec would be more likely to get accepted.

Susurrus:
What I would suggest is that we only support the `SIG[A-Z]*` format for signal identifiers. Additionally we implement `Display` for `Signal` and only test that that can be round-tripped. Additionally our round-tripped tests should do it for every variant and some ones that are clearly erroneous, rather than testing just a singular one.

quodlibetor:
I've done most of that. I implemented `AsRef<str>` as well, because it's possible to implement `Display` in terms of `AsRef` so the total amount of code is almost the same, but with all the benefits of not needing to allocate for the common-case in rust of just needing an `&str`.

quodlibetor:
I would still prefer to include the short-all-caps names, but if you're happy with this then so am I.

Mange:
I'd love to see this shipped. Is there anything I can help with?

Susurrus:
@Mange Take it for a spin in your own code, feel free to leave a review or feedback.

Susurrus:
This looks good to me at this point. @asomers any comments?

Before merging, it does need the following:
 * Probably a rebase
 * Squashed into a single commit with a good commit message
 * A CHANGELOG "Added" entry

PR #887, opened 2018-04-20
Implement Debug trait for PollFd (Fixes #885)

### LINKED ISSUE #885: No Debug trait for PollFd
I was trying to debug my program that uses poll(2) and discovered that it would be more convenient if nix supported Debug trait for PollFd.

The error message is

    `nix::poll::PollFd` cannot be formatted using `:?` because it doesn't implement `std::fmt::Debug`

Is there any reason we don't want this trait? It seems like to me, every struct should have the Debug trait to ease printf-debugging and logging.

kccqzy:
This is useful when using printf-style debugging to observe the variables and
of the program.

This is discussed in issue #885.


asomers:
How about a test?  And can you put "Fixes #885" in the commit message as well as the PR description?

PR #889, opened 2018-04-21
add TCP_KEEPINTVL and TCP_KEEPCNT

### LINKED ISSUE #888: TCP_KEEPIDLE and TCP_KEEPINTVL are missing from nix::sys::socket::SetSockOpt

aep:
closes #888


aep:
actually, osx doesnt have that. Would you be ok with target_os = "linux"  instead ?

asomers:
What do you mean by "inconsistent"?  If the parameter has a different size on different OSes or something, you should define it appropriately for each OS.

aep:
no, i just dont know bsd. Looks like some versions have it and some don't. It's difficult for me to judge if this ok. https://lists.freebsd.org/pipermail/freebsd-net/2013-May/035491.html

aep:
rustdoc doesnt pick them up. probably the reason why no other opts are documented. Should i remove it again?

asomers:
Just check the man pages:
https://www.freebsd.org/cgi/man.cgi?query=tcp&apropos=0&sektion=0&manpath=FreeBSD+11.1-RELEASE+and+Ports&arch=default&format=html
http://netbsd.gw.com/cgi-bin/man-cgi?tcp++NetBSD-current
https://man.openbsd.org/tcp
https://leaf.dragonflybsd.org/cgi/web-man?command=tcp&section=ANY

aep:
nah, libc needs the syscall values. not something i can do without background knowledge in which fork uses which libc. I opened a bug.

Susurrus:
@aep with rust-lang/libc#982 merged, are you ready to revisit this and get it cleaned up & merged?

povilasb:
I'd be happy to see this getting merged :)

PR #892, opened 2018-04-26
Fixed ptrace::Request cfg conditions

dalance:
The cfg condition of ptrace::Request seems to be different from libc.
For example, PTRACE_GETREGS is defined by libc to i686-unknown-linux-gnu target, but it is not defined in ptrace::Request.
I tried to change the cfg condition to the same as libc's condition.
I thinks this change covers the definitions under src/unix/notbsd directory in libc repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   `request_code_write`, and `request_code_readwrite` respectively. These have also now been exposed
   in the documentation.
   ([#833](https://github.com/nix-rust/nix/pull/833))
+- Enabled more `ptrace::Request` definitions for uncommon Linux platforms
+  ([#892](https://github.com/nix-rust/nix/pull/892))
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#833](https://github.com/nix-rust/nix/pull/833))
 - `ioctl_write_int!` now properly supports passing a `c_ulong` as the parameter on Linux non-musl targets
   ([#833](https://github.com/nix-rust/nix/pull/833))
+- Fixed `ptrace::Request` cfg conditions
```
asomers: How about "Enabled more `ptrace::Request` definitions for uncommon Linux platforms"?

dalance: I'll change to it.
Should it be moved to `Added` or `Changed` section?

dalance: I moved it to `Changed` setcion.

dalance:
I squashed and changed commit message.

PR #893, opened 2018-04-28
Remove IFF_NOTRAILERS for openbsd

kpcyrd:
nix currently doesn't compile on openbsd because `IFF_NOTRAILERS` isn't found. This PR fixes the openbsd build again (tested on 6.3). It seems it was changed in this commit:

https://github.com/openbsd/src/commit/beb8b0dd5985e55a615b52e593da6e75bab33f3f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#833](https://github.com/nix-rust/nix/pull/833))
 - Enabled more `ptrace::Request` definitions for uncommon Linux platforms
   ([#892](https://github.com/nix-rust/nix/pull/892))
+- Remove `IFF_NOTRAILERS` on OpenBSD, as it has been removed in OpenBSD 6.3
+  ([#893](https://github.com/nix-rust/nix/pull/893))
 
 ### Fixed
 - Properly exposed 460800 and 921600 baud rates on NetBSD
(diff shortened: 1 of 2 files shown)


asomers:
Did you mean to enable `IFF_STATICARP` too?  If so, can you do that as a separate commit?

asomers:
Ahh, well in that case you can leave it in one commit, but you should still mention it.  Also, it looks like this change is going to break compiling Nix for older versions of OpenBSD.  Can you mention that in the CHANGELOG?

asomers:
So removing `IFF_NOTRAILERS` fixes Nix for Openbsd >= 6.3, and adding `IFF_STATICARP` breaks Nix for OpenBSD < 6.3.  Would it be better not to add `IFF_STATICARP` to ensure that Nix will continue working on older versions of OpenBSD?  That seems to me like the less disruptive option.  Unfortunately Rust has no equivalent of `#ifdef`, so I don't see a way to do this automatically.

asomers:
I suppose I should ask what OpenBSD's support policy is.  When will 6.2 go EOL?

kpcyrd:
I've updated the changelog and the commit message.

asomers:
bors retry

kpcyrd:
The crate is now successfully building on OpenBSD 6.3: https://openbsd.rs/#/builders/8

Any chance we could get a release soon? :)

PR #276, opened 2016-02-19
Move SigSet, TimeVal to libc, add TimeSpec and pselect syscall.

abbradar:
Part and extension of #270 and should be merged after #272. Maybe we want to split this across multiple PRs too (although this time these things are dependent on each other).

PR #894, opened 2018-04-28
Add pselect syscall

antifuchs:
I saw that #276 was closed, and now I need `pselect`, so here it is! I copied the function body from @abbradar's work, updated the type signatures, added two tests and added a doc comment.

Hope this works!


src/sys/select.rs
```diff
@@ -131,6 +132,78 @@ where
...
+/// When this function returns, the original signal mask is restored.
+///
+/// Unlike [`select`](#fn.select), `pselect` does not mutate the `timeout` value.
+///
+/// # Parameters
+///
+/// * `nfds`: The highest file descriptor set in any of the passed `FdSet`s, plus 1. If `None`, this
+///   is calculated automatically by calling [`FdSet::highest`] on all descriptor sets and adding 1
+///   to the maximum of that.
+/// * `readfds`: File descriptors to check for being ready to read.
```
asomers: How about `s/being ready to read/read readiness` here and on the next two lines?

src/sys/select.rs
```diff
@@ -227,6 +300,29 @@ mod tests {
         assert!(!fd_set.contains(r2));
     }
 
+    #[test]
+    #[cfg_attr(any(target_arch = "powerpc", target_arch = "mips"), ignore)]
```
asomers: If you're going to ignore the test on those two arches, you should explain why in a comment.

src/sys/select.rs
```diff
@@ -227,6 +300,29 @@ mod tests {
...
+        fd_set.insert(r2);
+
+        let timeout = TimeSpec::seconds(10);
+        let sigmask = SigSet::empty();
+        assert_eq!(1, pselect(None,
+                              &mut fd_set,
+                              None,
+                              None,
+                              &timeout,
+                              &sigmask).unwrap());
```
asomers: Altering the signal mask has side effects which may interfere with other concurrent tests.  Nix has a way of dealing with that: you should this test into test/sys/test_select.rs and then add this line to the top:
```rust
let _mtx = ::SIGNAL_MTX.lock().expect("Mutex got poisoned by another test");
```
Since the other test does not modify the sigmask, you don't need the mutex.  It's up to you whether or not to move it.

src/sys/select.rs
antifuchs: Excellent point. Fixed.

src/sys/select.rs
antifuchs: Ah! I forgot to drop that line when I cargo-culted this test. Thanks for spotting it!

src/sys/select.rs
antifuchs: Thanks - I'll move both.

test/sys/test_select.rs
```diff
@@ -0,0 +1,55 @@
...
+        1,
+        pselect(None, &mut fd_set, None, None, &timeout, &sigmask).unwrap()
+    );
+    assert!(fd_set.contains(r1));
+    assert!(!fd_set.contains(r2));
+}
+
+#[test]
+pub fn test_pselect_nfds2() {
+    let _mtx = ::SIGNAL_MTX
```
asomers: You put `_mtx` in the wrong test; only `test_pselect` modifies the sigmask.

antifuchs: 🤦🏻‍♂️  oh no

antifuchs:
Updated once again, hahaha, I should probably go to bed (:

asomers:
thanks for your contribution!

bors r+

PR #896, opened 2018-05-01
Switch to Into<Option<Pid>> for Pid arguments in unistd

antifuchs:
This PR adjusts the unistd functions `setpgid`, `getpgid` and `getsid` to take `Into<Option<Pid>>` arguments, which (as of Rust 1.12) allows users to pass plain `Pid`s without wrapping them in `Some`.

Additionally, this allows users to pass `None` as both the `pid` and the `pgrp` argument in `setpgrp`, defaulting them to 0, as specified by POSIX.

I hope this is useful! If you like this, please let me know if I should go on an `Into<Option>` spree on the rest of the code (:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#833](https://github.com/nix-rust/nix/pull/833))
 
 ### Changed
+- `setpgid`, `getpgid` and `getsid` now take `Into<Option<Pid>>`
+  arguments. It is now possible to specify these optional Pid
+  arguments as plain `Pid`s or as `None`.
+  ([#896](https://github.com/nix-rust/nix/pull/896))
 - Display and Debug for SysControlAddr now includes all fields.
   ([#837](https://github.com/nix-rust/nix/pull/837))
 - `ioctl!` has been replaced with a family of `ioctl_*!` macros.
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -240,20 +240,30 @@ pub fn getppid() -> Pid {
 /// Set a process group ID (see
 /// [setpgid(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgid.html)).
 ///
-/// Set the process group id (PGID) of a particular process.  If a pid of zero
-/// is specified, then the pid of the calling process is used.  Process groups
-/// may be used to group together a set of processes in order for the OS to
-/// apply some operations across the group.
+/// Set the process group id (PGID) of a particular process.  If None
+/// or a pid of zero is specified, then the pid of the calling process
+/// is used.  Process groups may be used to group together a set of
+/// processes in order for the OS to apply some operations across the
+/// group.
 ///
 /// `setsid()` may be used to create a new process group.
 #[inline]
-pub fn setpgid(pid: Pid, pgid: Pid) -> Result<()> {
+pub fn setpgid<P: Into<Option<Pid>>>(pid: P, pgid: P) -> Result<()> {
+    let pid = pid.into().unwrap_or(Pid(0));
+    let pgid = pgid.into().unwrap_or(Pid(0));
     let res = unsafe { libc::setpgid(pid.into(), pgid.into()) };
     Errno::result(res).map(drop)
 }
+
+/// Get the process group ID for a process (see
+/// [getpgid(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/getpgid.html)).
+///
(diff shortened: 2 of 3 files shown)


asomers:
LGTM, but it needs a CHANGELOG entry.

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#833](https://github.com/nix-rust/nix/pull/833))
 
 ### Changed
+- `setpgid`, `getpgid` and `getsid` now take `Into<Option<Pid>>`
+  arguments. It is now possible to specify these optional Pid
+  arguments as plain `Pid`s or as `None`.
```
Susurrus: Please link to the PR as the other CHANGELOG entries do.

src/unistd.rs
```diff
@@ -240,20 +240,23 @@ pub fn getppid() -> Pid {
 /// Set a process group ID (see
 /// [setpgid(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgid.html)).
 ///
-/// Set the process group id (PGID) of a particular process.  If a pid of zero
-/// is specified, then the pid of the calling process is used.  Process groups
-/// may be used to group together a set of processes in order for the OS to
-/// apply some operations across the group.
+/// Set the process group id (PGID) of a particular process.  If None
+/// or a pid of zero is specified, then the pid of the calling process
```
Susurrus: Do we really want to still support a Pid of 0 here? I'd argue that the semantics would be improved by forcing people o use `None` or a valid `Pid` rather than having two possible values for `None`.

src/unistd.rs
```diff
@@ -240,20 +240,23 @@ pub fn getppid() -> Pid {
...
+pub fn setpgid<P: Into<Option<Pid>>>(pid: P, pgid: P) -> Result<()> {
+    let pid = pid.into().unwrap_or(Pid(0));
+    let pgid = pgid.into().unwrap_or(Pid(0));
     let res = unsafe { libc::setpgid(pid.into(), pgid.into()) };
     Errno::result(res).map(drop)
 }
 #[inline]
-pub fn getpgid(pid: Option<Pid>) -> Result<Pid> {
-    let res = unsafe { libc::getpgid(pid.unwrap_or(Pid(0)).into()) };
+pub fn getpgid<P: Into<Option<Pid>>>(pid: P) -> Result<Pid> {
```
Susurrus: Can you add doccomments for this function since you're modifying it?

src/unistd.rs
```diff
@@ -268,14 +271,13 @@ pub fn setsid() -> Result<Pid> {
 /// [getsid(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/getsid.html).
 ///
 /// Obtain the process group ID of the process that is the session leader of the process specified
-/// by pid. If pid is zero, it specifies the calling process.
+/// by pid. If passed None or pid is or zero, it specifies the calling process.
```
Susurrus: Same as the above, do we really want to allow `0` as a valid input. I'd argue that we shouldn't.

test/test_unistd.rs
```diff
@@ -112,8 +112,10 @@ fn test_getpid() {
 fn test_getsid() {
     let none_sid: ::libc::pid_t = getsid(None).unwrap().into();
     let pid_sid: ::libc::pid_t = getsid(Some(getpid())).unwrap().into();
+    let plain_sid: ::libc::pid_t = getsid(getpid()).unwrap().into();
```
Susurrus: None of these test cases cover `0`. I'm a fan of removing that functionality, but wanted to point it out for completeness.

src/unistd.rs
antifuchs: I have spent a while thinking about this now, and I don't see a better option than the one here:

* To disallow `0`, we'd need to be able to have an `.into` that can fail. The only way I know to do that would be to use `TryInto`, which isn't stable yet.
* If we disallowed `0` pid arguments, we'd become incompatible with other code that treats `0` as the "self" pid argument (e.g. C code).
* It's a super incompatible change with rust client code, as stuff that passed in `0` pids before no longer can (the program for which I am making this change would be one of them).

So I'm not sure those tradeoffs are worth introducing the additional complexity (and ergonomic setbacks) of a pid-but-not-zero type. Does that make sense?

antifuchs:
I squashed, added the changelog link, documented `getpgid` and added the test case; as outlined above, I'm not sure a 0-disallowing interface is feasible/super desirable at the moment, but if you can think of a way to do it nicely, I'm happy to change it!

PR #901, opened 2018-05-16
Implement sendfile on FreeBSD and Darwin

morrowa:
This PR exposes the `sendfile` system call on libc's supported BSD-likes:
* FreeBSD
* Darwin (macOS/iOS)

DragonFly could be supported in the future, but I was unable to build rustc to test.

Note that NetBSD has no equivalent system call.

### DIFF
--- a/src/sys/sendfile.rs
+++ b/src/sys/sendfile.rs
@@ -6,8 +6,195 @@ use libc::{self, off_t};
 use Result;
 use errno::Errno;
 
-pub fn sendfile(out_fd: RawFd, in_fd: RawFd, offset: Option<&mut off_t>, count: usize) -> Result<usize> {
-    let offset = offset.map(|offset| offset as *mut _).unwrap_or(ptr::null_mut());
+/// Copy up to `count` bytes to `out_fd` from `in_fd` starting at `offset`.
+///
+/// Returns a `Result` with the number of bytes written.
+///
+/// If `offset` is `None`, `sendfile` will begin reading at the current offset of `in_fd`and will
+/// update the offset of `in_fd`. If `offset` is `Some`, `sendfile` will begin at the specified
+/// offset and will not update the offset of `in_fd`. Instead, it will mutate `offset` to point to
+/// the byte after the last byte copied.
+///
+/// `in_fd` must support `mmap`-like operations and therefore cannot be a socket.
+///
+/// For more information, see [the sendfile(2) man page.](http://man7.org/linux/man-pages/man2/sendfile.2.html)
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub fn sendfile(
+    out_fd: RawFd,
+    in_fd: RawFd,
+    offset: Option<&mut off_t>,
+    count: usize,
+) -> Result<usize> {
+    let offset = offset
+        .map(|offset| offset as *mut _)
+        .unwrap_or(ptr::null_mut());
     let ret = unsafe { libc::sendfile(out_fd, in_fd, offset, count) };
     Errno::result(ret).map(|r| r as usize)
+}
+
+cfg_if! {
+    if #[cfg(any(target_os = "freebsd",
+                 target_os = "ios",
+                 target_os = "macos"))] {
+        use sys::uio::IoVec;
+
+        #[allow(missing_debug_implementations)]
+        struct SendfileHeaderTrailer<'a>(
+            libc::sf_hdtr,
+            Option<Vec<IoVec<&'a [u8]>>>,
+            Option<Vec<IoVec<&'a [u8]>>>,
+        );
+
+        impl<'a> SendfileHeaderTrailer<'a> {
+            fn new(
+                headers: Option<&'a [&'a [u8]]>,
+                trailers: Option<&'a [&'a [u8]]>
+            ) -> SendfileHeaderTrailer<'a> {
+                let header_iovecs: Option<Vec<IoVec<&[u8]>>> =
+                    headers.map(|s| s.iter().map(|b| IoVec::from_slice(b)).collect());
+                let trailer_iovecs: Option<Vec<IoVec<&[u8]>>> =
+                    trailers.map(|s| s.iter().map(|b| IoVec::from_slice(b)).collect());
+                SendfileHeaderTrailer(
+                    libc::sf_hdtr {
+                        headers: {
+                            header_iovecs
+                                .as_ref()
+                                .map_or(ptr::null(), |v| v.as_ptr()) as *mut libc::iovec
+                        },
+                        hdr_cnt: header_iovecs.as_ref().map(|v| v.len()).unwrap_or(0) as i32,
+                        trailers: {
+                            trailer_iovecs
+                                .as_ref()
+                                .map_or(ptr::null(), |v| v.as_ptr()) as *mut libc::iovec
+                        },
+                        trl_cnt: trailer_iovecs.as_ref().map(|v| v.len()).unwrap_or(0) as i32
+                    },
+                    header_iovecs,
+                    trailer_iovecs,
+                )
+            }
+        }
+    }
+}
+
+cfg_if! {
+    if #[cfg(target_os = "freebsd")] {
+        use libc::c_int;
+
+        libc_bitflags!{
+            /// Configuration options for [`sendfile`.](fn.sendfile.html)
+            pub struct SfFlags: c_int {
+                /// Causes `sendfile` to return EBUSY instead of blocking when attempting to read a
+                /// busy page.
+                SF_NODISKIO;
+                /// Causes `sendfile` to sleep until the network stack releases its reference to the
+                /// VM pages read. When `sendfile` returns, the data is not guaranteed to have been
+                /// sent, but it is safe to modify the file.
+                SF_SYNC;
+                /// Causes `sendfile` to cache exactly the number of pages specified in the
+                /// `readahead` parameter, disabling caching heuristics.
+                SF_USER_READAHEAD;
+                /// Causes `sendfile` not to cache the data read.
+                SF_NOCACHE;
+            }
+        }
+
+        /// Read up to `count` bytes from `in_fd` starting at `offset` and write to `out_sock`.
+        ///
+        /// Returns a `Result` and a count of bytes written. Bytes written may be non-zero even if
+        /// an error occurs.
+        ///
+        /// `in_fd` must describe a regular file or shared memory object. `out_sock` must describe a
+        /// stream socket.
+        ///
+        /// If `offset` falls past the end of the file, the function returns success and zero bytes
+        /// written.
+        ///
+        /// If `count` is `None` or 0, bytes will be read from `in_fd` until reaching the end of
+        /// file (EOF).
+        ///
+        /// `headers` and `trailers` specify optional slices of byte slices to be sent before and
+        /// after the data read from `in_fd`, respectively. The length of headers and trailers sent
+        /// is included in the returned count of bytes written. The values of `offset` and `count`
+        /// do not apply to headers or trailers.
+        ///
+        /// `readahead` specifies the minimum number of pages to cache in memory ahead of the page
+        /// currently being sent.
+        ///
+        /// For more information, see
+        /// [the sendfile(2) man page.](https://www.freebsd.org/cgi/man.cgi?query=sendfile&sektion=2)
+        pub fn sendfile(
+            in_fd: RawFd,
+            out_sock: RawFd,
+            offset: off_t,
+            count: Option<usize>,
+            headers: Option<&[&[u8]]>,
+            trailers: Option<&[&[u8]]>,
+            flags: SfFlags,
+            readahead: u16
+        ) -> (Result<()>, off_t) {
+            // Readahead goes in upper 16 bits
+            // Flags goes in lower 16 bits
+            // see `man 2 sendfile`
+            let flags: u32 = ((readahead as u32) << 16) | (flags.bits() as u32);
+            let mut bytes_sent: off_t = 0;
+            let hdtr = headers.or(trailers).map(|_| SendfileHeaderTrailer::new(headers, trailers));
+            let hdtr_ptr = hdtr.as_ref().map_or(ptr::null(), |s| &s.0 as *const libc::sf_hdtr);
+            let return_code = unsafe {
+                libc::sendfile(in_fd,
+                               out_sock,
+                               offset,
+                               count.unwrap_or(0),
+                               hdtr_ptr as *mut libc::sf_hdtr,
+                               &mut bytes_sent as *mut off_t,
+                               flags as c_int)
+            };
+            (Errno::result(return_code).and(Ok(())), bytes_sent)
+        }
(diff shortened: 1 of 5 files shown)


src/sys/sendfile.rs
```diff
@@ -2,12 +2,83 @@ use std::os::unix::io::RawFd;
...
 }
+
+#[cfg(any(target_os = "dragonfly",
+          target_os = "freebsd",
+          target_os = "ios",
+          target_os = "macos"))]
+#[allow(missing_debug_implementations)]
+#[allow(missing_copy_implementations)]
+#[repr(C)]
+pub struct SfHdTr(libc::sf_hdtr);
```
asomers: You should add a `PhantomData` field to keep `sf_hdtr`'s pointers from dangling.  See `AioCb::from_mut_slice` in src/sys/aio.rs for an example.

src/sys/sendfile.rs
```diff
@@ -2,12 +2,83 @@ use std::os::unix::io::RawFd;
...
+        })
+    }
+}
+
+#[cfg(target_os = "freebsd")]
+libc_bitflags!{
+    pub struct SfFlags: c_int {
+        SF_NODISKIO;
+        SF_MNOWAIT;
+        // SF_NOCACHE; // TODO: not yet exported by libc
```
asomers: Have you opened a PR with libc for `SF_NOCACHE` support?  If not, you should.  It's very easy to get PRs accepted to libc.  When you do, you should also include `SF_USER_READAHEAD`, which is new in FreeBSD 12.

src/sys/sendfile.rs
```diff
@@ -2,12 +2,83 @@ use std::os::unix::io::RawFd;
...
+            trl_cnt: trailers.len() as i32
+        })
+    }
+}
+
+#[cfg(target_os = "freebsd")]
+libc_bitflags!{
+    pub struct SfFlags: c_int {
+        SF_NODISKIO;
+        SF_MNOWAIT;
```
asomers: `SF_MNOWAIT` is obsolete in FreeBSD 11.0 and above.  The last release that makes use of it, 10.4, goes EoL on Oct 31 of this year.  So I'd rather not expose this symbol in Nix unless there is specific demand.

src/sys/sendfile.rs
```diff
@@ -2,12 +2,83 @@ use std::os::unix::io::RawFd;
...
+        // SF_NOCACHE; // TODO: not yet exported by libc
+        SF_SYNC;
+    }
+}
+
+#[cfg(target_os = "freebsd")]
+pub fn sendfile(in_fd: RawFd,
```
asomers: Could you please add some usage documentation?

test/test_sendfile.rs
```diff
@@ -28,3 +35,42 @@ fn test_sendfile() {
...
+    let headers: Vec<IoVec<&[u8]>> = header_strings.iter().map(|s| IoVec::from_slice(s.as_bytes())).collect();
+    let trailers: Vec<IoVec<&[u8]>> = trailer_strings.iter().map(|s| IoVec::from_slice(s.as_bytes())).collect();
+
+    let hdtr = SfHdTr::new(headers.as_slice(), trailers.as_slice());
+
+    // Prepare socket pair
+    let (mut rd, wr) = UnixStream::pair().unwrap();
+
+    // Call the test method
+    let (res, bytes_written) = sendfile(tmp.as_raw_fd(), wr.as_raw_fd(), body_offset as off_t, 0, Some(&hdtr), SfFlags::empty(), 0);
```
asomers: Wrap to 80 columns, please.

src/sys/sendfile.rs
```diff
@@ -2,12 +2,83 @@ use std::os::unix::io::RawFd;
...
+                out_sock: RawFd,
+                offset: off_t,
+                count: usize,
```
asomers: It would be more Rusty for `count` to be an `Option<usize>` instead of giving a special meaning to `0`.

morrowa:
Thanks for the very quick feedback! I'll update this PR soon.

src/sys/sendfile.rs
Susurrus: I'd love to see a better name for this struct compared to the cryptic shorthand libc uses. `SendfileHeader` I think would work

morrowa: How about `SendfileHeaderTrailer`?

test/test_sendfile.rs
morrowa: Should I just run `rustfmt` on the files I've touched?

src/sys/sendfile.rs
morrowa: Done! Does it look correct?

src/sys/sendfile.rs
morrowa: Working on a PR against libc now. The tests fail when I add the FreeBSD 12 constant and run them on my FreeBSD 11.1 machine. Once I figure that out, I'll submit it.

morrowa: https://github.com/rust-lang/libc/pull/988

src/sys/sendfile.rs
```diff
@@ -6,8 +6,98 @@ use libc::{self, off_t};
...
+
+cfg_if! {
+    if #[cfg(target_os = "freebsd")] {
+        use libc::c_int;
+
+        libc_bitflags!{
+            pub struct SfFlags: c_int {
+                SF_NODISKIO;
+                SF_SYNC;
+                // SF_USER_READAHEAD; // TODO: waiting on libc release
```
asomers: Nix uses a git dependency on libc (except when we release), so you can go ahead and enable these now.

src/sys/sendfile.rs
asomers: Yes

asomers:
I would still like to see some doc comments and a CHANGELOG entry.  I would consider Dragonfly support to be optional, since we don't test that platform anyway.

morrowa:
I haven't been able to build rustc on DragonFly after several tries, so I will not implement that in this PR. I'll add the doc comments, add the change log entry, clean up my TODOs, rebase, and ping you when it's ready. Thanks again!

morrowa:
@asomers and @Susurrus this is ready for re-review! I believe it's complete.

src/sys/sendfile.rs
```diff
@@ -6,8 +6,171 @@ use libc::{self, off_t};
...
+    }
+}
+
+cfg_if! {
+    if #[cfg(target_os = "freebsd")] {
+        use libc::c_int;
+
+        libc_bitflags!{
+            pub struct SfFlags: c_int {
+                SF_NODISKIO;
```
asomers: Could you please add some documentation for the flags?

src/sys/sendfile.rs
```diff
@@ -6,8 +6,171 @@ use libc::{self, off_t};
...
+/// Returns a `Result` with the number of bytes written.
+///
+/// If `offset` is `None`, `sendfile` will begin reading at the current offset of `in_fd`and will
+/// update the offset of `in_fd`. If `offset` is `Some`, `sendfile` will begin at the specified
+/// offset and will not update the offset of `in_fd`. Instead, it will mutate `offset` to point to
+/// the byte after the last byte copied.
+///
+/// `in_fd` must support `mmap`-like operations and therefore cannot be a socket.
+///
+/// For more information, see `man 2 sendfile`
```
asomers: This should be a link, like other man pages references throughout our docs.

src/sys/sendfile.rs
```diff
@@ -6,8 +6,171 @@ use libc::{self, off_t};
...
+        ///
+        /// If `count` is `None` or 0, bytes will be read from `in_fd` until reaching the end of
+        /// file (EOF).
+        ///
+        /// `hdtr` specifies an optional list of headers and trailers to be sent before and after
+        /// the data read from `in_fd`, respectively. The length of headers and trailers sent is
+        /// included in the returned count of bytes written. The values of `offset` and `count` do
+        /// not apply to headers or trailers.
+        ///
+        /// `readahead` specifies how many pages may be cached in memory ahead of the page currently
```
asomers: s/how many pages may be cached/the minimum number of pages to cache/

test/test_sendfile.rs
```diff
@@ -28,3 +37,109 @@ fn test_sendfile() {
...
+    let body = "Xabcdef123456";
+    let body_offset = 1;
+    let trailer_strings = vec!["\n", "Served by Make Believe\n"];
+
+    // Write the body to a file
+    let mut tmp = tempfile().unwrap();
+    tmp.write_all(body.as_bytes()).unwrap();
+
+    // Prepare headers and trailers for sendfile
+    let headers: Vec<IoVec<&[u8]>> = header_strings
```
asomers: Would it make sense for the API to accept a slice of slices instead of a slice of `IoVec`s?  That would simplify the user's code.  In fact, if you did this, then `SendfileHeaderTrailer` wouldn't even need to be public.

morrowa: How about a slice of byte slices? I feel like it makes the API even less obtuse.

morrowa:
@asomers ready for another look!

src/sys/sendfile.rs
```diff
@@ -36,30 +36,42 @@ cfg_if! {
...
                  target_os = "macos"))] {
-        use std::marker::PhantomData;
         use sys::uio::IoVec;
 
         #[allow(missing_debug_implementations)]
-        #[repr(C)]
-        pub struct SendfileHeaderTrailer<'a>(libc::sf_hdtr, PhantomData<&'a [IoVec<&'a [u8]>]>);
+        struct SendfileHeaderTrailer<'a>(
+            libc::sf_hdtr,
+            Option<Vec<IoVec<&'a [u8]>>>,
```
asomers: Why is it necessary to store the `Option<Vec<` fields in `SendfileHeaderTrailer`?

morrowa: Because it's creating the `IoVec` structs from byte slices, it's now the logical owner of those structs.

asomers: It still doesn't own the data, however.  `PhantomData` should be sufficient to keep the caller from dropping the byte slices.  Or is there something I'm missing?

morrowa: It doesn't own the byte slices, but it does own all of the `IoVec` instances, which do occupy memory.

asomers: Ahh, that makes sense.  In that case, this PR LGTM.  All it needs is a squash.

src/sys/sendfile.rs
```diff
@@ -6,8 +6,196 @@ use libc::{self, off_t};
...
+                    },
+                    header_iovecs,
+                    trailer_iovecs,
+                )
+            }
+        }
+    }
+}
+
+cfg_if! {
```
Susurrus: This doesn't need a new one I don't think and could be bundled into the above macro, yes?

src/sys/sendfile.rs
```diff
@@ -6,8 +6,196 @@ use libc::{self, off_t};
...
+///
+/// If `offset` is `None`, `sendfile` will begin reading at the current offset of `in_fd`and will
+/// update the offset of `in_fd`. If `offset` is `Some`, `sendfile` will begin at the specified
+/// offset and will not update the offset of `in_fd`. Instead, it will mutate `offset` to point to
+/// the byte after the last byte copied.
+///
+/// `in_fd` must support `mmap`-like operations and therefore cannot be a socket.
+///
+/// For more information, see [the sendfile(2) man page.](http://man7.org/linux/man-pages/man2/sendfile.2.html)
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize the OSes listed here.

test/test_sendfile.rs
```diff
@@ -1,15 +1,23 @@
...
+    if #[cfg(any(target_os = "linux", target_os = "android"))] {
+        use nix::unistd::{close, pipe, read};
+    }
+    else if #[cfg(any(target_os = "freebsd", target_os = "ios", target_os = "macos"))] {
+        use std::net::Shutdown;
+        use std::os::unix::net::UnixStream;
+    }
+}
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize the target OS list here.

test/test_sendfile.rs
```diff
@@ -1,15 +1,23 @@
...
-use tempfile::tempfile;
-
 use libc::off_t;
+use nix::sys::sendfile::*;
+use tempfile::tempfile;
 
-use nix::unistd::{close, pipe, read};
-use nix::sys::sendfile::sendfile;
+cfg_if! {
+    if #[cfg(any(target_os = "linux", target_os = "android"))] {
```
Susurrus: Please alphabetize the target OS list here.

test/test_sendfile.rs
```diff
@@ -1,15 +1,23 @@
...
+        use nix::unistd::{close, pipe, read};
+    }
+    else if #[cfg(any(target_os = "freebsd", target_os = "ios", target_os = "macos"))] {
```
Susurrus: Please put `else` expressions on the same line as the closing brace as suggested by [the style guidelines](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/expressions.md#control-flow-expressions).

src/sys/sendfile.rs
```diff
@@ -6,8 +6,196 @@ use libc::{self, off_t};
...
+                               offset,
+                               count.unwrap_or(0),
+                               hdtr_ptr as *mut libc::sf_hdtr,
+                               &mut bytes_sent as *mut off_t,
+                               flags as c_int)
+            };
+            (Errno::result(return_code).and(Ok(())), bytes_sent)
+        }
+    }
+    else if #[cfg(any(target_os = "ios", target_os = "macos"))] {
```
Susurrus: Please put `else` expressions on the same line as the closing brace as suggested by [the style guidelines](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/expressions.md#control-flow-expressions).

src/sys/sendfile.rs
morrowa: The header/trailer struct is the same across FreeBSD and Darwin, but the flags and function signatures are different, so they can't share all code. I could nest the ifs, but this felt like it caused less rightward drift.

morrowa: Yeah, if I try to put multiple independent if/else chains inside a single macro block, it won't compile. :(

morrowa:
@Susurrus style comments addressed! Ready for another look.

asomers:
The test failure is weird.  It looks like Nix's code is wrong, and AFAICT this should've been failing ever since 53789456df265e685475f8eeeea956d8ec92ded9, but it never failed until now.  There have been no recent changes in either libc or japaric/cross that look related.  I don't know why it just began to fail.

morrowa:
[A recent PR merged into libc](https://github.com/rust-lang/libc/pull/994) changed the `libc::bind` function signature for 64-bit Android. That's what's causing the build failure. I'm quite confident it's unrelated to these changes.

asomers:
You're right; I just couldn't find that with git blame for some reason.

Susurrus:
Please see #906 for a fix.

PR #902, opened 2018-05-17
Add Support for SCM_TIMESTAMPING

gz:
It's currently impossible to use hardware timestamps for network packets on Linux since the SCM_TIMESTAMPING type is missing in the ControlMessage enum.

The following commits add the given type. Note that this is a newer control message and has a different format from the already existing SCM_TIMESTAMP (i.e., can report more timestamps in a message and allows for nanosecond precision). Let me know if there is anything missing that you would like to add for this. 

I'm currently using these patches successfully with a network benchmark tool that can be found here https://github.com/gz/netlat.


CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#892](https://github.com/nix-rust/nix/pull/892))
 - Remove `IFF_NOTRAILERS` on OpenBSD, as it has been removed in OpenBSD 6.3
   ([#893](https://github.com/nix-rust/nix/pull/893))
+- ControlMessage enum now supports SCM_TIMESTAMPING messages on Linux
+  ([#XXX](https://github.com/nix-rust/nix/pull/XXX))
```
asomers: Don't forget to fix the XXX

gz:
Yes I can add a test case. There also seems to be no support for the SCM_TIMESTAMPING in the libc version used for ARM so I will remove them from the supported architectures of this flag.

gz:
@asomers I added a test case for it let me know if you have anything else that needs attention.

src/sys/socket/sockopt.rs
```diff
@@ -255,6 +255,11 @@ sockopt_impl!(Both, BindAny, libc::SOL_SOCKET, libc::SO_BINDANY, bool);
 sockopt_impl!(Both, BindAny, libc::IPPROTO_IP, libc::IP_BINDANY, bool);
 #[cfg(target_os = "linux")]
 sockopt_impl!(Both, Mark, libc::SOL_SOCKET, libc::SO_MARK, u32);
+#[cfg(all(target_os = "linux", target_arch = "x86_64", not(target_env = "musl")))]
+const SO_TIMESTAMPING: libc::c_int = 37;
```
asomers: No magic numbers please.  Use a symbol from the libc crate instead.

test/sys/test_socket.rs
```diff
@@ -255,3 +255,62 @@ pub fn test_syscontrol() {
...
+// Test creating and using named system control sockets
+#[cfg(all(target_os = "linux", target_arch = "x86_64", not(target_env = "musl")))]
+#[test]
+pub fn test_scm_timestamping() {
+    use nix::sys::uio::IoVec;
+    use nix::unistd::close;
+    use nix::sys::socket::{socket, bind, connect, sendmsg, recvmsg, setsockopt,
+                           AddressFamily, SockType, SockFlag, SockAddr,
+                           CmsgSpace, MsgFlags, ControlMessage};
+    use nix::sys::socket::sockopt::{Timestamping};
```
asomers: No need for curly braces here.

test/sys/test_socket.rs
```diff
@@ -255,3 +255,62 @@ pub fn test_syscontrol() {
...
+    use nix::sys::time::TimeVal;
+    use std::net;
+
+    let actual: net::SocketAddr = FromStr::from_str("127.0.0.1:3000").unwrap();
```
asomers: port 3000 is already used by a different test.  Can you choose a different port?

test/sys/test_socket.rs
```diff
@@ -255,3 +255,62 @@ pub fn test_syscontrol() {
...
+    let addr = SockAddr::new_inet(InetAddr::from_std(&actual));
+
+    let fd1 = socket(AddressFamily::Inet, SockType::Datagram, SockFlag::empty(), None).expect("socket failed");
+    connect(fd1, &addr).expect("connect failed");
+
+    let fd2 = socket(AddressFamily::Inet, SockType::Datagram, SockFlag::empty(), None).expect("socket failed");
+    bind(fd2, &addr).expect("binding failed");
+
+    // Definitions from linux/net_tstamp.h
+    const SOF_TIMESTAMPING_SOFTWARE: i32 = 1 << 4;
```
asomers: These symbols should also be in libc.

src/sys/socket/mod.rs
```diff
@@ -414,6 +417,11 @@ impl<'a> Iterator for CmsgIterator<'a> {
                 Some(ControlMessage::ScmTimestamp(
                     &*(cmsg_data.as_ptr() as *const _)))
             },
+            #[cfg(all(target_os = "linux", target_arch = "x86_64", not(target_env = "musl")))]
+            (libc::SOL_SOCKET, libc::SCM_TIMESTAMPING) => unsafe {
```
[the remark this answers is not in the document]
gz: AFAIK the problem lies with the rust libc version which defines the symbol only in a x86_64 specific file:
https://docs.rs/libc/0.2.42/src/libc/unix/notbsd/linux/other/b64/x86_64.rs.html#375

The timestamping itself is not restricted to x86-64 only, works fine on other architectures.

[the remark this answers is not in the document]
asomers: Then you should submit a PR to libc to get it fixed.  libc is very responsive to this kind of thing.

test/sys/test_socket.rs
asomers: No, I think it's fine.  libc is not just for POSIX stuff; there's plenty of OS-specific things in there.

asomers:
The tests are failing because they can't check out your branch.  There's something screwy about it.  Plus, you need to rebase anyway due to the conflict in CHANGELOG.md.  Please rebase.

gz:
@asomers sorry about that I was busy the last few days. I still need to get the necessary additions and changes in libc before we can merge this. But due to deadlines it wont happen before end of Jul. I will ping you here when it's done.

asomers:
I don't get it.  What does libvma have to do with anything?  And why did you disable that assertion?

gz:
Sorry about the confusion, I accidentally pushed this change you're confused about to this repo, it has nothing to do with this. I reverted and opened bug requests in libc for the remaining issues with this pull request. Once the constants are added to libc I will resolve the conflicts and we can hopefully merge this.

PR #907, opened 2018-06-01
Remove emulation of FD_CLOEXEC/O_NONBLOCK

Susurrus:
Rather than using the native implementation of these constants
on supported platforms, the native implementation was instead
emulated. This was also hidden from the user even though this
could result in data races and the functionality being broken.

Native functionality is, however, not support on macos/ios.
Rather than enable this emulation solely for this platform, it
should be removed as this is a dangerous abstraction.

This is a replacement for #863. There is much previous discussion there which I recommend you read to familiarize yourself with this decision.

I'm looking to push this through rather quickly as it's the last thing blocking our next 0.11.0 release.

cc @aomser @kristate

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -3,7 +3,6 @@
 //! [Further reading](http://man7.org/linux/man-pages/man7/socket.7.html)
 use {Error, Result};
 use errno::Errno;
-use features;
 use libc::{self, c_void, c_int, socklen_t, size_t};
 use std::{fmt, mem, ptr, slice};
 use std::os::unix::io::RawFd;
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -692,87 +691,44 @@ pub fn recvmsg<'a, T>(fd: RawFd, iov: &[IoVec<&mut [u8]>], cmsg_buffer: Option<&
...
-    if feat_atomic {
-        ty |= flags.bits();
-    }
+    // SockFlags are usually embedded into `ty`, but we don't do that in `nix` because it's a
+    // little easier to understand by separating it out. So we have to merge these bitfields
+    // here.
+    let mut ty = ty as c_int;
+    ty |= flags.bits();
 
     // TODO: Check the kernel version
```
asomers: This comment should go.

src/sys/socket/mod.rs
```diff
@@ -810,45 +766,9 @@ pub fn accept(sockfd: RawFd) -> Result<RawFd> {
 ///
 /// [Further reading](http://man7.org/linux/man-pages/man2/accept.2.html)
 pub fn accept4(sockfd: RawFd, flags: SockFlag) -> Result<RawFd> {
```
asomers: Looks like OSX lacks accept4, and NetBSD only has it on the unstable development version.  So you should gate this function.

Susurrus:
@asomers should be good here I think.

kristate:
@Susurrus You seem to be using [my test cases](https://github.com/nix-rust/nix/pull/863/files#diff-d76bcdb0ac1e11a568ac60c293d81be1) from #863 in this commit [[ref](https://github.com/nix-rust/nix/pull/907/files#diff-d76bcdb0ac1e11a568ac60c293d81be1)], but I am curious why you did not add me as a committer to this commit/patch?

Susurrus:
Sorry, you're right! I meant to just run them through the tests before merging them, but I should have talked to you before doing it at all. I'm really sorry about that!

How would you like to handle this? I'd be happy to remove the test cases and you could file an additional PR with just those test cases so we can get it logged in git history that they're your work.

PR #908, opened 2018-06-02
Prep for 0.11 release

Susurrus:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name        = "nix"
 description = "Rust friendly bindings to *nix APIs"
-version     = "0.11.0-pre"
+version     = "0.11"
```
asomers: Why no trailing ".0" ?

Susurrus: Good point, fixed.

Susurrus:
Interesting that Cargo won't successfully parse a manifest file with only an X.Y version number.

Susurrus:
We're going to block on this until I sort out the tests I stole in #907.

PR #909, opened 2018-06-02
Remove tests that weren't mine

Susurrus:
cc @kristate


Susurrus:
This should pass, so let's see if we can push this through.

bors r+

Susurrus:
Just a connection timeout hiccup. Let's try this again.

bors r+

PR #911, opened 2018-06-02
Start the next dev cycle

Susurrus:
bors r+


asomers:
Are you not going to bump the rev to 0.12.0-pre ?

Susurrus:
It wasn't in the release procedure doc so I didn't even think about it. Is there any reason to actually do this?

PR #900, opened 2018-05-11
deps: update tempfile to 3

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

PR #916, opened 2018-06-28
new dir module

### LINKED ISSUE #915: support fdopendir + readdir operations
I'm writing some static file serving code for a webserver. It opens paths using `openat`, determines what type they are (plain file, directory, ...), then tries to serve them. So I want to pass my existing file descriptor to `fdopendir` rather than pass a path to `std::fs::read_dir`.

I have a prototype implementation which I'll open a pull request for shortly. This is my first contribution to nix. Apologies if I didn't match style well.

One particular point: I wasn't sure if it'd be better to add a new module or put this into an existing one such as `fcntl`. I went with the former for the moment.

scottlamb:
Fixes #915 

This is a lower-level interface than `std::fs::ReadDir`. Notable differences:

   * can be opened from a file descriptor (as returned by `openat`, perhaps
     before knowing if the path represents a file or directory). Uses
     `fdopendir` for this, available on all Unix platforms as of
     rust-lang/libc#1018.

   * implements `AsRawFd`, so it can be passed to `fstat`, `openat`, etc.

   * can be iterated through multiple times without closing and reopening the
     file descriptor. Each iteration rewinds when finished.

   * returns entries for `.` (current directory) and `..` (parent directory).

   * returns entries' names as a `CStr` (no allocation or conversion beyond
     whatever libc does).

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -1,0 +1,210 @@
+use {Error, NixPath, Result};
+use errno::Errno;
+use fcntl::{self, OFlag};
+use libc;
+use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
+use std::{ffi, fmt, ptr};
+use sys;
+
+#[cfg(target_os = "linux")]
+use libc::{dirent64 as dirent, readdir64_r as readdir_r};
+
+#[cfg(not(target_os = "linux"))]
+use libc::{dirent, readdir_r};
+
+/// An open directory.
+///
+/// This is a lower-level interface than `std::fs::ReadDir`. Notable differences:
+///    * can be opened from a file descriptor (as returned by `openat`, perhaps before knowing
+///      if the path represents a file or directory).
+///    * implements `AsRawFd`, so it can be passed to `fstat`, `openat`, etc.
+///      The file descriptor continues to be owned by the `Dir`, so callers must not keep a `RawFd`
+///      after the `Dir` is dropped.
+///    * can be iterated through multiple times without closing and reopening the file
+///      descriptor. Each iteration rewinds when finished.
+///    * returns entries for `.` (current directory) and `..` (parent directory).
+///    * returns entries' names as a `CStr` (no allocation or conversion beyond whatever libc
+///      does).
+pub struct Dir(
+    // This could be ptr::NonNull once nix requires Rust 1.25.
+    *mut libc::DIR
+);
+
+impl Dir {
+    /// Opens the given path as with `fcntl::open`.
+    pub fn open<P: ?Sized + NixPath>(path: &P, oflag: OFlag,
+                                     mode: sys::stat::Mode) -> Result<Self> {
+        let fd = fcntl::open(path, oflag, mode)?;
+        Dir::from_fd(fd)
+    }
+
+    /// Opens the given path as with `fcntl::openat`.
+    pub fn openat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P, oflag: OFlag,
+                                       mode: sys::stat::Mode) -> Result<Self> {
+        let fd = fcntl::openat(dirfd, path, oflag, mode)?;
+        Dir::from_fd(fd)
+    }
+
+    /// Converts from a descriptor-based object, closing the descriptor on success or failure.
+    #[inline]
+    pub fn from<F: IntoRawFd>(fd: F) -> Result<Self> {
+        Dir::from_fd(fd.into_raw_fd())
+    }
+
+    /// Converts from a file descriptor, closing it on success or failure.
+    pub fn from_fd(fd: RawFd) -> Result<Self> {
+        let d = unsafe { libc::fdopendir(fd) };
+        if d.is_null() {
+            let e = Error::last();
+            unsafe { libc::close(fd) };
+            return Err(e);
+        };
+        Ok(Dir(d))
+    }
+
+    /// Returns an iterator of `Result<Entry>` which rewinds when finished.
+    pub fn iter(&mut self) -> Iter {
+        Iter(self)
+    }
+}
+
+// `Dir` is not `Sync`. With the current implementation, it could be, but according to
+// https://www.gnu.org/software/libc/manual/html_node/Reading_002fClosing-Directory.html,
+// future versions of POSIX are likely to obsolete `readdir_r` and specify that it's unsafe to
+// call `readdir` simultaneously from multiple threads.
+//
+// `Dir` is safe to pass from one thread to another, as it's not reference-counted.
+unsafe impl Send for Dir {}
+
+impl AsRawFd for Dir {
+    fn as_raw_fd(&self) -> RawFd {
+        unsafe { libc::dirfd(self.0) }
+    }
+}
+
+impl fmt::Debug for Dir {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Dir")
+            .field("fd", &self.as_raw_fd())
+            .finish()
+    }
+}
+
+impl Drop for Dir {
+    fn drop(&mut self) {
+        unsafe { libc::closedir(self.0) };
+    }
+}
+
+#[derive(Debug)]
+pub struct Iter<'d>(&'d mut Dir);
+
+impl<'d> Iterator for Iter<'d> {
+    type Item = Result<Entry>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe {
+            // Note: POSIX specifies that portable applications should dynamically allocate a
+            // buffer with room for a `d_name` field of size `pathconf(..., _PC_NAME_MAX)` plus 1
+            // for the NUL byte. It doesn't look like the std library does this; it just uses
+            // fixed-sized buffers (and libc's dirent seems to be sized so this is appropriate).
+            // Probably fine here too then.
+            let mut ent: Entry = Entry(::std::mem::uninitialized());
+            let mut result = ptr::null_mut();
+            if let Err(e) = Errno::result(readdir_r((self.0).0, &mut ent.0, &mut result)) {
+                return Some(Err(e));
+            }
+            if result == ptr::null_mut() {
+                return None;
+            }
+            assert_eq!(result, &mut ent.0 as *mut dirent);
+            return Some(Ok(ent));
+        }
+    }
+}
+
+impl<'d> Drop for Iter<'d> {
+    fn drop(&mut self) {
+        unsafe { libc::rewinddir((self.0).0) }
+    }
+}
+
+/// A directory entry, similar to `std::fs::DirEntry`.
+///
+/// Note that unlike the std version, this may represent the `.` or `..` entries.
+#[derive(Copy, Clone)]
+pub struct Entry(dirent);
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Type {
+    Fifo,
+    CharacterDevice,
+    Directory,
+    BlockDevice,
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Entry {
+    /// Returns the inode number (`d_ino`) of the underlying `dirent`.
+    #[cfg(any(target_os = "android",
+              target_os = "emscripten",
+              target_os = "fuchsia",
+              target_os = "haiku",
+              target_os = "ios",
+              target_os = "l4re",
+              target_os = "linux",
+              target_os = "macos",
+              target_os = "solaris"))]
+    pub fn ino(&self) -> u64 {
+        self.0.d_ino as u64
+    }
+
+    /// Returns the inode number (`d_fileno`) of the underlying `dirent`.
+    #[cfg(not(any(target_os = "android",
+                  target_os = "emscripten",
+                  target_os = "fuchsia",
+                  target_os = "haiku",
+                  target_os = "ios",
+                  target_os = "l4re",
+                  target_os = "linux",
+                  target_os = "macos",
+                  target_os = "solaris")))]
+    pub fn ino(&self) -> u64 {
+        self.0.d_fileno as u64
+    }
+
+    /// Returns the bare file name of this directory entry without any other leading path component.
+    pub fn file_name(&self) -> &ffi::CStr {
+        unsafe { ::std::ffi::CStr::from_ptr(self.0.d_name.as_ptr()) }
+    }
+
+    /// Returns the type of this directory entry, if known.
+    ///
+    /// See platform `readdir(3)` or `dirent(5)` manpage for when the file type is known;
+    /// notably, some Linux filesystems don't implement this. The caller should use `stat` or
+    /// `fstat` if this returns `None`.
+    pub fn file_type(&self) -> Option<Type> {
+        match self.0.d_type {
+            libc::DT_FIFO => Some(Type::Fifo),
+            libc::DT_CHR => Some(Type::CharacterDevice),
+            libc::DT_DIR => Some(Type::Directory),
+            libc::DT_BLK => Some(Type::BlockDevice),
+            libc::DT_REG => Some(Type::File),
+            libc::DT_LNK => Some(Type::Symlink),
+            libc::DT_SOCK => Some(Type::Socket),
+            /* libc::DT_UNKNOWN | */ _ => None,
+        }
+    }
+}
+
+impl fmt::Debug for Entry {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Entry")
+            .field("ino", &self.ino())
+            .field("file_name", &self.file_name())
+            .field("file_type", &self.file_type())
+            .finish()
+    }
+}
--- a/test/test_dir.rs
+++ b/test/test_dir.rs
@@ -1,0 +1,46 @@
+extern crate nix;
+extern crate tempfile;
+
+use nix::dir::{Dir, Type};
+use nix::fcntl::OFlag;
+use nix::sys::stat::Mode;
+use std::fs::File;
+use self::tempfile::tempdir;
+
+#[test]
+fn read() {
+    let tmp = tempdir().unwrap();
+    File::create(&tmp.path().join("foo")).unwrap();
+    ::std::os::unix::fs::symlink("foo", tmp.path().join("bar")).unwrap();
+    let mut dir = Dir::open(tmp.path(), OFlag::O_DIRECTORY | OFlag::O_RDONLY | OFlag::O_CLOEXEC,
+                            Mode::empty()).unwrap();
+    let mut entries: Vec<_> = dir.iter().map(|e| e.unwrap()).collect();
+    entries.sort_by(|a, b| a.file_name().cmp(b.file_name()));
+    let entry_names: Vec<_> = entries
+        .iter()
+        .map(|e| e.file_name().to_str().unwrap().to_owned())
+        .collect();
+    assert_eq!(&entry_names[..], &[".", "..", "bar", "foo"]);
+
+    // Check file types. The system is allowed to return DT_UNKNOWN (aka None here) but if it does
+    // return a type, ensure it's correct.
+    assert!(&[Some(Type::Directory), None].contains(&entries[0].file_type())); // .: dir
+    assert!(&[Some(Type::Directory), None].contains(&entries[1].file_type())); // ..: dir
+    assert!(&[Some(Type::Symlink), None].contains(&entries[2].file_type())); // bar: symlink
+    assert!(&[Some(Type::File), None].contains(&entries[3].file_type())); // foo: regular file
+}
+
+#[test]
+fn rewind() {
+    let tmp = tempdir().unwrap();
+    let mut dir = Dir::open(tmp.path(), OFlag::O_DIRECTORY | OFlag::O_RDONLY | OFlag::O_CLOEXEC,
+                            Mode::empty()).unwrap();
+    let entries1: Vec<_> = dir.iter().map(|e| e.unwrap().file_name().to_owned()).collect();
+    let entries2: Vec<_> = dir.iter().map(|e| e.unwrap().file_name().to_owned()).collect();
+    assert_eq!(entries1, entries2);
+}
+
+#[test]
+fn ebadf() {
+    assert_eq!(Dir::from_fd(-1).unwrap_err(), nix::Error::Sys(nix::errno::Errno::EBADF));
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#921](https://github.com/nix-rust/nix/pull/921))
 - Added support for `SCM_CREDENTIALS`, allowing to send process credentials over Unix sockets.
   ([#923](https://github.com/nix-rust/nix/pull/923))
+- Added a `dir` module for reading directories (wraps `fdopendir`, `readdir`, and `rewinddir`).
+  ([#916](https://github.com/nix-rust/nix/pull/916))
 
 ### Changed
 - Increased required Rust version to 1.22.1/
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,7 @@ pub extern crate libc;
 #[macro_use] mod macros;
 
 // Public crates
+pub mod dir;
 pub mod errno;
 #[deny(missing_docs)]
 pub mod features;
--- a/test/test.rs
+++ b/test/test.rs
@@ -10,6 +10,7 @@ extern crate rand;
 extern crate tempfile;
 
 mod sys;
+mod test_dir;
 mod test_fcntl;
 #[cfg(any(target_os = "dragonfly",
           target_os = "freebsd",


src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+    Directory,
+    BlockDevice,
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Entry {
+    pub fn inode(&self) -> u64 {
+        self.0.d_ino
```
asomers: This field is called `d_fileno` on FreeBSD.  I don't know about the other BSDs; you should check.

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+                let e = Error::last();
+                unsafe { libc::close(fd) };
+                Err(e)
+            }
+            Some(d) => Ok(Dir(d)),
+        }
+    }
+}
+
+// `Dir` is not `Sync`, but it's not reference-counted either, so it can be safely passed from one
```
asomers: What makes it non-`Sync` ?

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+
+#[cfg(not(target_os = "linux"))]
+use libc::{dirent, readdir_r};
+
+/// An open directory.
+///! This is a lower-level interface than `std::fs::ReadDir`. Notable differences:
+///!    * can be opened from a file descriptor (as returned by `openat`, perhaps before knowing
+///!      if the path represents a file or directory).
+///!    * implements `AsRawFd`, so it can be passed to `fstat`, `openat`, etc.
+///!    * can be iterated through multiple times without closing and reopening the file
```
asomers: How can it be iterated through multiple times if it only implemented `IntoIterator` and not `Iterator`?

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub fn from<F: IntoRawFd>(fd: F) -> Result<Self> {
+        Dir::from_fd(fd.into_raw_fd())
+    }
+
+    /// Converts from a file descriptor, closing it on success or failure.
+    pub fn from_fd(fd: RawFd) -> Result<Self> {
+        match ptr::NonNull::new(unsafe { libc::fdopendir(fd) }) {
+            None => {
+                let e = Error::last();
+                unsafe { libc::close(fd) };
```
asomers: `Dir` takes ownership of the file descriptor, and closes it on `Drop`.  Unfortunately, `RawFd` is defined as an alias, so there's no way to enforce that the caller doesn't hold onto a copy.  But could you at least warn the caller in the documentation?  Bad stuff will happen if he accesses fd outside of the `Dir` interface.

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+impl Drop for Dir {
+    fn drop(&mut self) {
+        unsafe { libc::closedir(self.0.as_ptr()) };
+    }
+}
+
+#[allow(missing_debug_implementations)]
```
asomers: Generally, an ugly derived `Debug` implementation is better than none at all, because none at all prevents `derive` from working on larger types composed with this one.  Is there any reason not to `derive` it?

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+            }
+            assert_eq!(result, &mut ent.0 as *mut dirent);
+            return Some(Ok(ent));
+        }
+    }
+}
+
+impl<'d> Drop for Iter<'d> {
+    fn drop(&mut self) {
+        unsafe { libc::rewinddir((self.0).0.as_ptr()) }
```
asomers: Are the parens around `self.0` really necessary?

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+/// A directory entry, similar to `std::fs::DirEntry`.
+/// Note that unlike the std version, this may represent the `.` or `..` entries.
+#[derive(Copy, Clone)]
```
asomers: How about `Debug`?

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Entry {
+    pub fn inode(&self) -> u64 {
+        self.0.d_ino
+    }
+    pub fn name(&self) -> &ffi::CStr {
```
asomers: spaces between methods, please.  Also, please document all public methods.

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+        unsafe { ::std::ffi::CStr::from_ptr(self.0.d_name.as_ptr()) }
+    }
+
+    /// Returns the type of this directory entry, if known.
+    /// See platform `readdir(3)` or `dirent(5)` manpage for when the file type is known;
+    /// notably, some Linux filesystems don't implement this. The caller should use `stat` or
+    /// `fstat` if this returns `None`.
+    pub fn type_(&self) -> Option<Type> {
```
asomers: Since `Entry` is basically a replacement for `std::fs::DirEntry`, why not make the API as close as possible?  This method should be named `file_type`, and should return an object supporting methods like those of `std::fs::FileType`.

src/dir.rs
asomers: Should be named `file_name` to match `std::fs::DirEntry`

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+    CharacterDevice,
+    Directory,
+    BlockDevice,
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Entry {
+    pub fn inode(&self) -> u64 {
```
asomers: Should be named `ino` to match `std::fs::DirEntry`

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+            libc::DT_LNK => Some(Type::Symlink),
+            libc::DT_SOCK => Some(Type::Socket),
+            /* libc::DT_UNKNOWN | */ _ => None,
+        }
+    }
+}
+
+impl PartialEq<Entry> for Entry {
+    fn eq(&self, other: &Entry) -> bool {
+        self.inode() == other.inode() && self.name() == other.name()
```
asomers: This method could give incorrect results for files opened from different file systems.  Probably best just to compare `name` and the `Dir` object's file descriptor.

src/dir.rs
scottlamb: `&mut Directory` implements `IntoIterator`, so you can get an iterator more than once. The test does this.

I could have `Dir` implement `Iterator` instead but you'd have to remember to rewind it.

src/dir.rs
scottlamb: Hmm, I was under the impression it was only thread-safe when used with different `DIR` objects. It turns out I was wrong with `readdir_r`, but according to [this glibc documentation](https://www.gnu.org/software/libc/manual/html_node/Reading_002fClosing-Directory.html), that call might be obsoleted, and `readdir` will be specified to work as I'd imagined.

I'm inclined to update the comment and leave it non-`Sync`. Opinions?

src/dir.rs
scottlamb: I manually implemented `Debug`, so that `name` would be shown as a `CStr` rather than a `[::c_char; 256]` or whatever.

src/dir.rs
scottlamb: Fixed: now matches the std implementation's cfg stuff.

src/dir.rs
scottlamb: `AsRawFd::as_raw_fd` has such a warning:

> This method does not pass ownership of the raw file descriptor to the caller. The descriptor is only guaranteed to be valid while the original object has not yet been destroyed.

but I'll add one to `Dir`'s docs as well where I mention the `AsRawFd` implementation.

src/dir.rs
scottlamb: I suppose not. Done.

src/dir.rs
scottlamb: Yeah, I put them here at the suggestion of the compiler:

```
error: unexpected token: `0.0`
  --> src/dir.rs:97:58
   |
97 |             if let Err(e) = Errno::result(readdir_r(self.0.0.as_ptr(), &mut ent.0, &mut result)) {
   |                                                     -----^^^
   |                                                     |    |
   |                                                     |    unexpected token
   |                                                     help: try parenthesizing the first index: `(self.0).0`
```

src/dir.rs
scottlamb: All done.

src/dir.rs
scottlamb: Done.

src/dir.rs
scottlamb: Hmm, honestly, comparing two `Entry`s is probably not that useful at all. I just have it here for my test. What would you think of removing it and do something else there?

src/dir.rs
scottlamb: Hmm. I'd like to preserve the ability to see exactly what the OS returned, for a couple reasons:

   * general educational value
   * you might want to use this to filter out uninteresting types without calling stat. If you decide to look more closely, you might want to do the stat in a different way (openat + fstat), and you might want to see the other fields without a second stat call.

My general impression was that nix is supposed to pretty closely match what the Unix systems offer, where the std library is more high-level. Does that sound right?

I made a similar decision to preserve the potentially-useful `.` and `..` entries (you might want to call `ino()` on them, for example) where std filters it out.

src/dir.rs
```diff
@@ -0,0 +1,167 @@
...
+///! This is a lower-level interface than `std::fs::ReadDir`. Notable differences:
+///!    * can be opened from a file descriptor (as returned by `openat`, perhaps before knowing
+///!      if the path represents a file or directory).
+///!    * implements `AsRawFd`, so it can be passed to `fstat`, `openat`, etc.
+///!    * can be iterated through multiple times without closing and reopening the file
+///!      descriptor. Each iteration rewinds when finished.
+///!    * returns entries for `.` (current directory) and `..` (parent directory).
+///!    * returns entries' names as a `CStr` (no allocation or conversion beyond whatever libc
+///!      does).
+pub struct Dir(ptr::NonNull<libc::DIR>);
```
scottlamb: This didn't go over well with the continuous build:

```
error[E0412]: cannot find type `NonNull` in module `ptr`
  --> src/dir.rs:23:21
   |
23 | pub struct Dir(ptr::NonNull<libc::DIR>);
   |                     ^^^^^^^ not found in `ptr`
```

so I switched to `*mut libc::DIR` instead.

scottlamb:
Pushed a new commit that should address most of these comments.

Thanks for the quick review! I appreciate it even though I wasn't able to respond for a while (got sick, yuck).

src/dir.rs
asomers: Removing it would be reasonable.  Is there a reason that `Entry` needs to implement `PartialEq`?

src/dir.rs
asomers: That's because `NonNull` only got added in Rust 1.25.0.  Nix's minimum requirement is 1.20.0.

src/dir.rs
asomers: Oh, I see.  You implemented `IntoIterator` for `&Dir`, not `Dir`.  I think it would be more idiomatic instead for `Dir` to have a `iter()` method, that returns an object which implemented `Iterator`.

src/dir.rs
```diff
@@ -0,0 +1,198 @@
...
+use std::{ffi, fmt, ptr};
+
+#[cfg(target_os = "linux")]
+use libc::{dirent64 as dirent, readdir64_r as readdir_r};
+
+#[cfg(not(target_os = "linux"))]
+use libc::{dirent, readdir_r};
+
+/// An open directory.
+///! This is a lower-level interface than `std::fs::ReadDir`. Notable differences:
```
asomers: Does this render correctly in Rust Doc?  Normally `///!` annotates the enclosing object, which in this case would be the file.  To describe `Dir` instead, you should have all `///`s, with a blank line after the first.

src/dir.rs
```diff
@@ -0,0 +1,198 @@
...
+    Directory,
+    BlockDevice,
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Entry {
+    /// Returns the inode number (`d_ino`) of the underlying `dirent`.
+    #[cfg(any(target_os = "macos",
```
asomers: alphabetical ordering, please.

src/dir.rs
```diff
@@ -0,0 +1,198 @@
...
+              target_os = "l4re",
+              target_os = "fuchsia"))]
+    pub fn ino(&self) -> u64 {
+        self.0.d_ino as u64
+    }
+
+    /// Returns the inode number (`d_fileno`) of the underlying `dirent`.
+    #[cfg(any(target_os = "freebsd",
+              target_os = "openbsd",
+              target_os = "bitrig",
```
asomers: bitrig is dead.  You can remove it.  Also, the others should be alphabetically ordered.  Does Nix support any OSes that can't implement `ino`?  If not, you can simplify this with `cfg_if`, like this:
```rust
pub fn ino(&self) -> u64 {
  cfg_if! {
    if #[cfg(any(target_os = ...))] {
       self.0.d_ino as u64
    } else {
      self.0.d_filenno as u64
    }
  }
}

File: src/dir.rs (hunk 9 above)
Comment (asomers) [reply src/dir.rs]: That's reasonable.  But I would still rather call this method `file_type` or `filetype`.  The trailing underscore is weird and unprecedented.

File: test/test_dir.rs [hunk 16]
```diff
@@ -0,0 +1,49 @@
+extern crate nix;
+extern crate tempdir;
+
+use nix::dir::{Dir, Type};
+use std::fs::File;
+use self::tempdir::TempDir;
+
+fn open(path: &::std::path::Path) -> nix::Result<Dir> {
```
asomers: Would it be worthwhile to make this a method of `Dir`, so that you can open by either `RawFd` or a path?

test/test_dir.rs
```diff
@@ -0,0 +1,49 @@
...
+    ::std::os::unix::fs::symlink("foo", tmp.path().join("bar")).unwrap();
+    let mut dir = open(tmp.path()).unwrap();
+    let mut entries: Vec<_> = dir.into_iter().map(|e| e.unwrap()).collect();
+    entries.sort_by(|a, b| a.file_name().cmp(b.file_name()));
+    let entry_names: Vec<_> = entries
+        .iter()
+        .map(|e| e.file_name().to_str().unwrap().to_owned())
+        .collect();
+    assert_eq!(&entry_names[..], &[".", "..", "bar", "foo"]);
+    assert!(&[Some(Type::Directory), None].contains(&entries[0].type_())); // .: dir
```
asomers: It would be worthwhile adding a comment explaining why the check for None is necessary.

src/dir.rs
scottlamb: Hmm, that's neat, but I'm having trouble getting it to work. Tried a couple variations. Any ideas?

```rust
    pub fn ino(&self) -> u64 {
        cfg_if! {
            if #[cfg(any(target_os = "android",
                         target_os = "emscripten",
                         target_os = "fuchsia",
                         target_os = "haiku",
                         target_os = "ios",
                         target_os = "l4re",
                         target_os = "linux",
                         target_os = "macos",
                         target_os = "solaris"))] {
                self.0.d_ino as u64
            } else {
                self.0.d_fileno as u64
            }
        }
    }
```

produces

```
error: expected one of `!` or `::`, found `.`
   --> src/dir.rs:158:21
    |
158 |                 self.0.d_ino as u64
    |                     ^ expected one of `!` or `::` here
```

and

```rust
    cfg_if! {
        if #[cfg(any(target_os = "android",
                     target_os = "emscripten",
                     target_os = "fuchsia",
                     target_os = "haiku",
                     target_os = "ios",
                     target_os = "l4re",
                     target_os = "linux",
                     target_os = "macos",
                     target_os = "solaris"))] {
            pub fn ino(&self) -> u64 { self.0.d_ino as u64 }
        } else {
            pub fn ino(&self) -> u64 { self.0.d_fileno as u64 }
        }
    }
```

produces a bunch of errors, starting with:

```
error: expected one of `::` or `:`, found `)`
   --> src/dir.rs:175:29
    |
175 |             pub fn ino(&self) -> u64 { self.0.d_ino as u64 }
    |                             ^ expected one of `::` or `:` here
```

asomers: Looks like `cfg_if` doesn't work with expressions, only items.  An item is something like a module or a function.
https://docs.rs/cfg-if/0.1.4/cfg_if/macro.cfg_if.html
https://doc.rust-lang.org/reference/items.html

scottlamb: Hmm, function didn't work either (second try above). I got it to work by having a separate `impl` for this vs the other stuff, but then the two impls are separate in the documentation, which is annoying.

I think I'm giving up on `cfg_if!` here, but I can do `#[cfg(any(target_os = "bar", target_os="foo"))]` vs `#[cfg(not(any(target_os = "bar", target_os="foo")))]` so it's obviously comprehensive.

test/test_dir.rs
scottlamb: Added.

test/test_dir.rs
scottlamb: Done - both `fcntl::open` and `fcntl::openat` variants. Unlike my quick test version, I'm using a `NixPath`, and I'm expecting the caller to provide `OFlags`, so they can do `O_NOFOLLOW` or `O_EXLOCK` or whatever as desired.

src/dir.rs
scottlamb: Done.

src/dir.rs
scottlamb: It did, but fixed anyway. Also adding some extra whitespace to the docstrings so the one-line extract would be less overwhelming.

scottlamb:
> One more thing: it's not safe to have two Iter objects alive at the same time, because they'll affect each other's directory pointers. It may be better to implement Iterator directly on Dir, and add an explicit rewind method.

The lifetime should prevent that problem. Each `Iter` has a `&'d mut Dir`, and there can only be one of those at once.

src/dir.rs
```diff
@@ -0,0 +1,208 @@
...
+#[cfg(not(target_os = "linux"))]
+use libc::{dirent, readdir_r};
+
+/// An open directory.
+///
+/// This is a lower-level interface than `std::fs::ReadDir`. Notable differences:
+///    * can be opened from a file descriptor (as returned by `openat`, perhaps before knowing
+///      if the path represents a file or directory).
+///    * implements `AsRawFd`, so it can be passed to `fstat`, `openat`, etc.
+///      The descriptor continues to be owned by the `Dir`, so callers must not keep a `RawFd`
```
asomers: s/descriptor/file/

src/dir.rs
```diff
@@ -0,0 +1,208 @@
...
+
+impl Dir {
+    /// Opens the given path as with `fcntl::open`.
+    pub fn open<P: ?Sized + NixPath>(path: &P, oflag: OFlag) -> Result<Self> {
+        let fd = fcntl::open(path, oflag, sys::stat::Mode::empty())?;
+        Dir::from_fd(fd)
+    }
+
+    /// Opens the given path as with `fcntl::openat`.
+    pub fn openat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P, oflag: OFlag) -> Result<Self> {
```
asomers: hardcoding `Mode::empty` prevents this API from being used to create new directories.  Any reason not to expose the `mode` argument of `fcntl::openat` and `fcntl::open`?

scottlamb: Done. I just never imagined using this to create directories would actually work, but it does on Linux. (On macOS, it returns EINVAL.)

src/dir.rs
scottlamb: file descriptor? it's really not the file.

src/dir.rs
asomers: Any idea why?  Can `openat` not create directories on OSX?

src/dir.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub fn file_type(&self) -> Option<Type> {
+        match self.0.d_type {
+            libc::DT_FIFO => Some(Type::Fifo),
+            libc::DT_CHR => Some(Type::CharacterDevice),
+            libc::DT_DIR => Some(Type::Directory),
+            libc::DT_BLK => Some(Type::BlockDevice),
+            libc::DT_REG => Some(Type::File),
+            libc::DT_LNK => Some(Type::Symlink),
+            libc::DT_SOCK => Some(Type::Socket),
+            /* libc::DT_UNKNOWN | */ _ => None,
```
Susurrus: Wouldn't this be better to actually use `DT_UNKNOWN` and then have another catchall that maps to `unreachable!()`?

test/test_dir.rs
```diff
@@ -0,0 +1,46 @@
...
+    // return a type, ensure it's correct.
+    assert!(&[Some(Type::Directory), None].contains(&entries[0].file_type())); // .: dir
+    assert!(&[Some(Type::Directory), None].contains(&entries[1].file_type())); // ..: dir
+    assert!(&[Some(Type::Symlink), None].contains(&entries[2].file_type())); // bar: symlink
+    assert!(&[Some(Type::File), None].contains(&entries[3].file_type())); // foo: regular file
+}
+
+#[test]
+fn rewind() {
+    let tmp = TempDir::new("nix-dir-rewind").unwrap();
```
Susurrus: Can't we have anonymous folders? I'd prefer we use that style rather than naming everything.

test/test_dir.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use nix::dir::{Dir, Type};
+use nix::fcntl::OFlag;
+use nix::sys::stat::Mode;
+use std::fs::File;
+use self::tempdir::TempDir;
+
+#[test]
+fn read() {
+    let tmp = TempDir::new("nix-dir-read").unwrap();
```
Susurrus: Can't we have anonymous folders? I'd prefer we use that style rather than naming everything.

test/test_dir.rs
```diff
@@ -0,0 +1,46 @@
...
+                            Mode::empty()).unwrap();
+    let mut entries: Vec<_> = dir.iter().map(|e| e.unwrap()).collect();
+    entries.sort_by(|a, b| a.file_name().cmp(b.file_name()));
+    let entry_names: Vec<_> = entries
+        .iter()
+        .map(|e| e.file_name().to_str().unwrap().to_owned())
+        .collect();
+    assert_eq!(&entry_names[..], &[".", "..", "bar", "foo"]);
+
+    // Check file types. The system is allowed to return DT_UNKNOWN (aka None here) but if it does
```
Susurrus: But why would it return `DT_UNKNOWN`? Does it do this in practice on any systems?

src/dir.rs
```diff
@@ -0,0 +1,210 @@
...
+    Directory,
+    BlockDevice,
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Entry {
+    /// Returns the inode number (`d_ino`) of the underlying `dirent`.
+    #[cfg(any(target_os = "android",
```
Susurrus: Can we use `cfg-if` here instead of these huge if/else blocks?

src/dir.rs
```diff
@@ -0,0 +1,210 @@
+use {Error, NixPath, Result};
+use errno::Errno;
+use fcntl::{self, OFlag};
+use libc;
+use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
+use std::{ffi, fmt, ptr};
+use sys;
+
+#[cfg(target_os = "linux")]
```
Susurrus: Can we use the `cfg_if!` macro here instead? It's a little more verbose but I find it easier to read an if-else rather than trying to determine that through consecutive `cfg` statements.

src/dir.rs
scottlamb: Hmm, I think this argument applies equally well to anywhere else that uses `AsRawFd`, and Google doesn't find any results for `AsRef<RawFd>`. I think I'd prefer to match the rest of the ecosystem.

test/test_dir.rs
scottlamb: The `tempdir` crate doesn't seem to support that. `tempfile` version 3 [does](https://docs.rs/tempfile/3.0.3/tempfile/fn.tempdir.html) (and that obsoletes `tempdir`), but nix still seems to be on version 2. This call to `TempDir::new` matches the other ones in the `tests` dir.

test/test_dir.rs
scottlamb: likewise

test/test_dir.rs
scottlamb: Yes, it does. From the Linux `readdir(3)` manpage:

```
              Currently,  only  some  filesystems  (among them: Btrfs, ext2, ext3, and ext4) have
              full support for returning the file type in d_type.  All applications must properly
              handle a return of DT_UNKNOWN.
```

src/dir.rs
scottlamb: and `RawFd` is `Copy`, so `AsRef<RawFd>` is probably just as easy to mess up anyway.

src/dir.rs
scottlamb: If the list of `DT_` types is incomplete (which seems possible—I haven't read all the platforms' manpages, and a platform could add one anyway), what should this code do? With `unreachable!`, it'd panic. I'd prefer it return `None`. Calling code is supposed to handle `DT_UNKNOWN` gracefully, and it makes sense to me to handle an unexpected enum type in the same way.

src/dir.rs
scottlamb: I tried this on July 5th, and the only way I could get it to work was by having two `impl` blocks, which makes the `rustdoc` output look weird. `cfg_if` can only be attached at the "item" level; apparently neither functions nor expressions are items.

src/dir.rs
scottlamb: likewise

src/dir.rs
asomers: I forgot about `Copy`.  In that case you're right that `AsRef<RawRd>` wouldn't have much advantage.

scottlamb:
Anything else I need to do to finish this?

Susurrus:
LGTM. @asomers any last changes you think are necessary?

asomers:
No changes except a squash.

scottlamb:
Squashed. Thank you!

asomers:
Looks like we need to merge PR #900 first.

Susurrus:
#900 was just merged

bors r+

asomers:
Looks like you need to rebase and replace the tempdir crate with tempfile.

scottlamb:
Rebased, replaced tempfile with tempdir.

Susurrus:
Can you add an Added entry into the CHANGELOG? Then I think this LGTM.

scottlamb:
Done

Susurrus:
Awesome! Thanks @scottlamb!

bors r+

PR #918, opened 2018-07-01
Fix passing multiple file descriptors / control messages via sendmsg

### LINKED ISSUE #464: Passing multiple FD over an Unix Socket in a single Control Message doesn't work

### LINKED ISSUE #756: Sending multiple ControlMessage in a single Unix Socket packet fails

jonas-schievink:
Fixes #464
Closes #874 because it's incorporated here
Closes #756 because it adds the test from that issue (with fixes)

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -287,18 +287,37 @@ impl fmt::Debug for Ipv6MembershipRequest {
     }
 }
 
-/// Copy the in-memory representation of src into the byte slice dst,
-/// updating the slice to point to the remainder of dst only. Unsafe
-/// because it exposes all bytes in src, which may be UB if some of them
-/// are uninitialized (including padding).
-unsafe fn copy_bytes<'a, 'b, T: ?Sized>(src: &T, dst: &'a mut &'b mut [u8]) {
+/// Copy the in-memory representation of `src` into the byte slice `dst`.
+///
+/// Returns the remainder of `dst`.
+///
+/// Panics when `dst` is too small for `src` (more precisely, panics if
+/// `mem::size_of_val(src) >= dst.len()`).
+///
+/// Unsafe because it transmutes `src` to raw bytes, which is only safe for some
+/// types `T`. Refer to the [Rustonomicon] for details.
+///
+/// [Rustonomicon]: https://doc.rust-lang.org/nomicon/transmutes.html
+unsafe fn copy_bytes<'a, T: ?Sized>(src: &T, dst: &'a mut [u8]) -> &'a mut [u8] {
     let srclen = mem::size_of_val(src);
-    let mut tmpdst = &mut [][..];
-    mem::swap(&mut tmpdst, dst);
-    let (target, mut remainder) = tmpdst.split_at_mut(srclen);
-    // Safe because the mutable borrow of dst guarantees that src does not alias it.
-    ptr::copy_nonoverlapping(src as *const T as *const u8, target.as_mut_ptr(), srclen);
-    mem::swap(dst, &mut remainder);
+    ptr::copy_nonoverlapping(
+        src as *const T as *const u8,
+        dst[..srclen].as_mut_ptr(),
+        srclen
+    );
+
+    &mut dst[srclen..]
+}
+
+/// Fills `dst` with `len` zero bytes and returns the remainder of the slice.
+///
+/// Panics when `len >= dst.len()`.
+fn pad_bytes(len: usize, dst: &mut [u8]) -> &mut [u8] {
+    for pad in &mut dst[..len] {
+        *pad = 0;
+    }
+
+    &mut dst[len..]
 }
 
 cfg_if! {
@@ -434,6 +453,11 @@ pub enum ControlMessage<'a> {
     ///
     /// See the description in the "Ancillary messages" section of the
     /// [unix(7) man page](http://man7.org/linux/man-pages/man7/unix.7.html).
+    ///
+    /// Using multiple `ScmRights` messages for a single `sendmsg` call isn't recommended since it
+    /// causes platform-dependent behaviour: It might swallow all but the first `ScmRights` message
+    /// or fail with `EINVAL`. Instead, you can put all fds to be passed into a single `ScmRights`
+    /// message.
     ScmRights(&'a [RawFd]),
     /// A message of type `SCM_TIMESTAMP`, containing the time the
     /// packet was received by the kernel.
@@ -545,7 +569,7 @@ impl<'a> ControlMessage<'a> {
 
     // Unsafe: start and end of buffer must be cmsg_align'd. Updates
     // the provided slice; panics if the buffer is too small.
-    unsafe fn encode_into<'b>(&self, buf: &mut &'b mut [u8]) {
+    unsafe fn encode_into(&self, buf: &mut [u8]) {
         match *self {
             ControlMessage::ScmRights(fds) => {
                 let cmsg = cmsghdr {
@@ -554,17 +578,16 @@ impl<'a> ControlMessage<'a> {
                     cmsg_type: libc::SCM_RIGHTS,
                     ..mem::uninitialized()
                 };
-                copy_bytes(&cmsg, buf);
+                let buf = copy_bytes(&cmsg, buf);
 
                 let padlen = cmsg_align(mem::size_of_val(&cmsg)) -
                     mem::size_of_val(&cmsg);
+                let buf = pad_bytes(padlen, buf);
 
-                let mut tmpbuf = &mut [][..];
-                mem::swap(&mut tmpbuf, buf);
-                let (_padding, mut remainder) = tmpbuf.split_at_mut(padlen);
-                mem::swap(buf, &mut remainder);
+                let buf = copy_bytes(fds, buf);
 
-                copy_bytes(fds, buf);
+                let padlen = self.space() - self.len();
+                pad_bytes(padlen, buf);
             },
             ControlMessage::ScmTimestamp(t) => {
                 let cmsg = cmsghdr {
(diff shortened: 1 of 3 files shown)


jonas-schievink:
Current CI failures:

* Only 1 fd was successfully received, the other got lost
  * FreeBSD x86_64
  * powerpc64-unknown-linux-gnu
  * powerpc64le-unknown-linux-gnu
  * mips64-unknown-linux-gnuabi64
  * mips64el-unknown-linux-gnuabi64
  * aarch64-unknown-linux-gnu
* `EINVAL`
  * i686-apple-darwin
  * x86_64-apple-darwin

jonas-schievink:
**Every single** CI job that failed passes another test that *also* passes 2 fds, but only using a *single* control message. Seems like you're not supposed to use more than 1 `SCM_RIGHTS` message.

I've deleted the broken test and added a note.

test/test_pass_fds.rs
```diff
@@ -0,0 +1,37 @@
...
+            Ok(msg) => {
+                let mut iter = msg.cmsgs();
+                if let Some(ControlMessage::ScmRights(fds)) = iter.next() {
+                    (iovec[0].as_slice(), fds.to_vec())
+                } else {
+                    panic!();
+                }
+            },
+            Err(_) => {
+                panic!();
```
asomers: These two panic messages are indistinguishable since you don't provide descriptions for them.  And if the error case is a simple `panic`, you may as well just use `unwrap` instead of `if let`.

test/test_pass_fds.rs
```diff
@@ -0,0 +1,37 @@
...
+use nix::sys::uio::IoVec;
+
+
+fn main() {
+    let (send, receive) = UnixDatagram::pair().unwrap();
+    let thread = std::thread::spawn(move || {
+        let mut buf = [0u8; 8];
+        let iovec = [IoVec::from_mut_slice(&mut buf)];
+        let mut space = CmsgSpace::<[RawFd; 2]>::new();
+        let (paneid, fds) = match recvmsg(receive.as_raw_fd(), &iovec, Some(&mut space), MsgFlags::empty()) {
```
asomers: What is a "paneid" ?

test/test_pass_fds.rs
```diff
@@ -0,0 +1,37 @@
...
+            Err(_) => {
+                panic!();
+            }
+        };
+        println!("paneid = {:?}, fds = {:?}", paneid, fds);
+    });
+
+    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
+    let iov = [IoVec::from_slice(&slice)];
+    let arr = [0, 1]; // Pass stdin and stdout
```
asomers: This should be `libc::STDIN_FILENO` and `libc::STDOUT_FILENO`.

test/sys/test_socket.rs
```diff
@@ -167,6 +167,41 @@ pub fn test_scm_rights() {
...
+            _ => panic!(),
+        }
+        assert!(cmsgs.next().is_none(), "unexpected control msg");
+
+        assert_eq!(iovec[0].as_slice(), [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8]);
+    });
+
+    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
+    let iov = [IoVec::from_slice(&slice)];
+    let fds = [0, 1];    // pass stdin and stdout
```
asomers: should be libc::STDIN_FILENO and libc::STDOUT_FILENO.

Cargo.toml
```diff
@@ -50,3 +50,8 @@ harness = false
 [[test]]
 name = "test-ptymaster-drop"
 path = "test/test_ptymaster_drop.rs"
+
+[[test]]
```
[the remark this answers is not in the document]
jonas-schievink: Turns out the test I added in a recent commit essentially does the same thing (pass 2 fds using 1 cmsg), so I'll remove this one.

Adding examples to nix is a very good idea, but I guess it should be done as a more coordinated approach rather than in this PR.

jonas-schievink:
> Does Nix also have a bug with sending two dissimilar control messages in a single sendmsg call, and if so does this change fix it?

This *should* fix that, but I haven't added a test since there's no convenient `ControlMessage` that can be used to test it. I might add support for `SCM_CREDENTIALS` later and use that along with `SCM_RIGHTS` to test this.

asomers:
Why not use `ScmRights` with `ScmTimestamp`?

jonas-schievink:
Attempting to send a `SCM_TIMESTAMP` results in `EINVAL` on my system (x64 Linux). Even when it's just a single `SCM_TIMESTAMP` control message, and even from C++, so it's not a bug in nix, you're just not supposed to send those by yourself (the kernel does).

jonas-schievink:
@asomers Anything left to do here?

src/sys/socket/mod.rs
```diff
@@ -287,18 +287,30 @@ impl fmt::Debug for Ipv6MembershipRequest {
...
-/// because it exposes all bytes in src, which may be UB if some of them
-/// are uninitialized (including padding).
-unsafe fn copy_bytes<'a, 'b, T: ?Sized>(src: &T, dst: &'a mut &'b mut [u8]) {
+/// Copy the in-memory representation of `src` into the byte slice `dst`.
+///
+/// Returns the remainder of `dst`.
+///
+/// Panics when `dst` is too small for `src` (panics if `mem::size_of_val(src) >= dst.len()`).
+///
+/// Unsafe because it exposes all bytes in `src`, which may be UB if some of them are uninitialized
```
asomers: It's not actually unsafe to read an uninitialized slice, because slices are always supposed to be initialized.  The unsafe part is creating the uninitialized slice in the first place.  However, it _is_ unsafe to call `copy_nonoverlapping` without verifying that the slices don't overlap, without verifying that `T` implements `Copy`, and without `drop`ing the contents of `dst`.  Please update the comment.

src/sys/socket/mod.rs
```diff
@@ -287,18 +287,30 @@ impl fmt::Debug for Ipv6MembershipRequest {
...
-    mem::swap(dst, &mut remainder);
+    ptr::copy_nonoverlapping(src as *const T as *const u8, dst[..srclen].as_mut_ptr(), srclen);
+
+    &mut dst[srclen..]
+}
+
+/// Fills `dst` with `len` zero bytes and returns the remainder of the slice.
+///
+/// Panics when `len >= dst.len()`.
+fn pad_bytes(len: usize, dst: &mut [u8]) -> &mut [u8] {
```
asomers: Much nicer than the previous version.  I like.

test/sys/test_socket.rs
```diff
@@ -167,6 +167,42 @@ pub fn test_scm_rights() {
...
+    let thread = thread::spawn(move || {
+        let mut buf = [0u8; 8];
+        let iovec = [IoVec::from_mut_slice(&mut buf)];
+        let mut space = CmsgSpace::<[RawFd; 2]>::new();
+        let msg = recvmsg(receive.as_raw_fd(), &iovec, Some(&mut space), MsgFlags::empty()).unwrap();
+        assert!(!msg.flags.intersects(MsgFlags::MSG_TRUNC | MsgFlags::MSG_CTRUNC));
+
+        let mut cmsgs = msg.cmsgs();
+        match cmsgs.next() {
+            Some(ControlMessage::ScmRights(fds)) => assert_eq!(fds.len(), 2, "unexpected fd count (expected 2 fds)"),
```
asomers: 80 chars per line, please.

src/sys/socket/mod.rs
jonas-schievink: The problem is that this is reading the bytes of an arbitrary `T`, including its padding. If reading the bytes from a `T` is safe, then so is this function. It doesn't create any uninitialized memory or slices.

AFAIK, LLVM considers reads of padding bytes `undef`, which would be unsafe, but I don't know if rustc makes it safe in some way (eg. by providing its own padding members that read as junk instead of `undef`).

asomers: So the unsafe part is that it's effectively transmuting the `T`.  Transmutation is unsafe regardless of padding issues.

jonas-schievink: I don't think transmutation to just bytes is unsafe except when padding is involved (as there's no way to read data that isn't valid as a `u8`), but I've clarified what's going on here

jonas-schievink:
```
     Running /target/i686-unknown-linux-gnu/release/deps/test_aio_drop-88766c7ced1811bd
running 1 test
test test_drop ... error: An unknown error occurred
```

That doesn't look good - but also spurious

asomers:
bors retry

PR #920, opened 2018-07-04
Closure transfer fix for `sched::clone`

### LINKED ISSUE #919: `std::sched::clone` doesn't pass ownership of closure to new thread.
The current implementation of of `std::sched::clone` doesn't pass ownership of the the closure called by `callback` into the newly created thread, only a reference. This means that when the call to `clone` ends, the callback is actually dropped, releasing all associated memory. This most likely means that the closure that actually gets run in the child thread is a reference to invalid memory when it is run.

Recommended fix is to unbox the `CloneCb` into a raw pointer and box it back up again in `callback`. This would mean that the closure would be dropped at the end of the callback in the child thread, rather than at the end of the call to `clone`.

xurtis:
Fixes #919

Annotate the correct lifetime (`'static`) for the closure transferred to
the child process of the clone.

Properly box the closure in the parent then rebox it in the child to
ensure that the closure is dropped at the end of the invocation in the
child rather than at the end of the call to `clone` in the parent.

For some reason, unwrapping the boxed function itself caused all kinds
of mayhem, so it instead ends up wrapped in an additional box from which
it is then unwrapped.

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -36,7 +36,7 @@ libc_bitflags!{
     }
 }
 
-pub type CloneCb<'a> = Box<FnMut() -> isize + 'a>;
+pub type CloneCb = Box<FnMut() -> isize + Send + 'static>;
 
 #[repr(C)]
 #[derive(Clone, Copy)]
@@ -85,24 +85,24 @@ pub fn sched_setaffinity(pid: Pid, cpuset: &CpuSet) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-pub fn clone(mut cb: CloneCb,
+pub fn clone(cb: CloneCb,
              stack: &mut [u8],
              flags: CloneFlags,
              signal: Option<c_int>)
              -> Result<Pid> {
     extern "C" fn callback(data: *mut CloneCb) -> c_int {
-        let cb: &mut CloneCb = unsafe { &mut *data };
-        (*cb)() as c_int
+        let mut cb: CloneCb = unsafe { *Box::from_raw(data) };
+        cb() as c_int
     }
 
     let res = unsafe {
         let combined = flags.bits() | signal.unwrap_or(0);
         let ptr = stack.as_mut_ptr().offset(stack.len() as isize);
         let ptr_aligned = ptr.offset((ptr as usize % 16) as isize * -1);
-        libc::clone(mem::transmute(callback as extern "C" fn(*mut Box<::std::ops::FnMut() -> isize>) -> i32),
+        libc::clone(mem::transmute(callback as extern "C" fn(*mut CloneCb) -> i32),
                    ptr_aligned as *mut c_void,
                    combined,
-                   &mut cb as *mut _ as *mut c_void)
+                   Box::into_raw(Box::new(cb)) as *mut c_void)
     };
 
     Errno::result(res).map(Pid::from_raw)


xurtis:
After looking at how `std::thread` achieves the same result, I changed the patch to mimic that instead of doing strange dances with boxes. The (much simpler) solution turns out to be just forgetting the boxed function rather than dropping it.

xurtis:
It also appears that even at its lowest level, in order to implement the FnOnce semantics, `libstd::thread` uses a `Box<Box<FnBox()>>` which it then unwraps, sends, then boxes again before calling. Still not sure on the why though.

I think it would make more sense to require the callback to be a FnOnce, but I can't see a way of doing that in stable currently.

asomers:
I'm skeptical of the `mem::forget`.  That function deliberately leaks memory, and is normally used when you've made a low-level (unsafe) copy of a reference somewhere.  But normally at least one copy gets destroyed, and I don't see that happening.  Is it really necessary to `forget` the callback, when it's already `static`?

xurtis:
In this case the `'static` refers to the fact that the function that the box wraps is not bound to the lifetime of any other object (i.e. all items that it refers to have been moved into it). It _does not_ mean that the boxed closure is static.

The existing implementation of clones as you have ti will only work when the child thread does not use the `CLONE_VM` flag, meaning it gets its own entire copy of the VM. The reason it doesn't work is because the boxed closure will be dropped by the end of the call to clone, as it is moved into that function and that function consumes it. There is only one copy to this box and the child thread gets a reference to it, but if that box is dropped be for the child thread exits, then the child thread could rely entirely on invalid memory and, in most cases, won't even get to the point of executing the closure.

Either, the box needs to be copied (which it can't) or have its ownership transferred to the child process, or forgotten and never dropped at all. The only to transfer ownership to the child process involves wrapping a the boxed closure in an additional box (I'm still not entirely clear on the why there).

xurtis:
Another alternative fix that could be used here is raising that concern to the caller and using a more direct analogous type to what the syscall is actually expecting and have the wrapper take a `fn(T) -> isize` (a function pointer rather than a boxed closure) and a `Box<T>` (where `T: Send + 'static`) instead. This would leave more abstract implementations of executing closures to caller instead, whcih may be preferable in this case.

asomers:
I agree.  But I'm not comfortable with the double-box solution without knowing why it's necessary.  Did you check the history of the file in `libstd` that uses that technique?  That might be instructive.

xurtis:
So, on your recommendation, I'm going back through the years of commits to https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/thread.rs and it goes back quite some way through pre-stable rust (back before `~` notation for boxes was a thing).

I asked around on `#rust` and got some help answering the question. The short version is, don't cast between pointers of static and DSTs. DST pointers are actually 'fat' pointers which are a tuple of a pointer to data and a pointer to a vtable. When those get cast to a pointer to a void, the vtable data is dropped. To resolve this, an additional box should be used, so that you have a pointer to a static type (which contains a DST pointer) that can be safely casted to pointer to a static type, like `*mut c_void`.

The existing code masked a fatal compile error that would have caught this using a `mem::transmute`.

src/sched.rs
```diff
@@ -119,3 +119,34 @@ pub fn setns(fd: RawFd, nstype: CloneFlags) -> Result<()> {
 
     Errno::result(res).map(drop)
 }
+
+#[cfg(not)]
```
asomers: cfg what?

src/sched.rs
```diff
@@ -119,3 +119,34 @@ pub fn setns(fd: RawFd, nstype: CloneFlags) -> Result<()> {
...
+        Box::new(move || {
+            assert_eq!(numbers.iter().sum::<i32>(), 5050);
+            0
+        })
+    }
+
+    #[test]
+    fn simple_clone() {
+        // Stack *must* outlive the child.
+        let mut stack = Vec::new();
```
asomers: May as well combine these lines into `let mut stack = vec![0u8; 4096]`

src/sched.rs
xurtis: I wanted to see if the tests on travis would pass if the test was disabled. It seems that with this test in place, both this test and the `sys::wait::tests::test_waitpid` test would fail in various situations, but not in a way that I could replicate locally, even when building for targets that were failing in travis.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [0.11.0] 2018-06-01
 
 ### Added
+- Changed mechanism used to send closure to child thread in `sched::clone`.
```
asomers: This description doesn't say why you "changed mechanism".  Did you fix a bug?  Add a feature?  Change just for the heck of it?  Also, it's in the wrong section.  "Added" is for new features.  "Fixed" is for bug fixes.  "Changed" is for backwards-incompatible changes, including bug fixes.

asomers:
Linux tests that aren't x86_64 or i386 get executed in QEMU, which has some limitations.  You should research if there are any known problems using clone(2) in QEMU.  If so, you'll have to disable those tests except for i386 and x86_64.

xurtis:
I don't think this is an issue with QEMU though. On arm (which requires stack alignment), the musl implementation (which fixes the alignment rather than returning an error) works, but the glibc implementations (which passes the value unaltered to the syscall) fails.

oblique:
Any news on this?

asomers:
I think we can merge it as soon as we figure out why it fails in QEMU.  If anybody has actual arm, mips, or ppc hardware running Linux that would make a good point of comparison.

jD91mZM2:
Anyone still working on this?

PR #921, opened 2018-07-04
Get `SO_PEERCRED` working on all Linux targets

jonas-schievink:
These were disabled for ARM way back in 0db6ed1a28b4fb4d408cd9b7c4dba0a79bccf1f7 and 09c00ed7d9d92db2a79baa3ed212e7e239edceb9. Try to enable them for all arches and Android as well, since the removal wasn't really explained and I see no reason why this shouldn't work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Added a `sysinfo` wrapper.
   ([#922](https://github.com/nix-rust/nix/pull/922))
+- Support the `SO_PEERCRED` socket option and the `UnixCredentials` type on all Linux and Android targets.
+  ([#921](https://github.com/nix-rust/nix/pull/921))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


jonas-schievink:
[CI passes](https://travis-ci.org/nix-rust/nix/builds/400012144), so I've squashed and added a changelog entry.

asomers:
@berkowski @Susurrus can you comment about why these socket options were disabled?  Is there any problem with reenabling them?

berkowski:
They were disabled at the time to get nix passing arm-based tests.  There's been a lot of changes in the upstream libs since then so maybe a few missing symbols were added?  

I don't have the knowledge to say definitively whether that flag is valid on all platforms.  I'd trust it is if libc exports it at this point.

asomers:
@berkowski do you recall if the arm-based tests were failing deterministically, or intermittently?

berkowski:
@asomers, I don't remember specifics, sorry.  Most of my effort at that time was just getting nix (and libc) to compile under the new architectures exposed by trust.  I'd guess that I was chasing down compile-time issues rather than run-time test failings.

PR #923, opened 2018-07-06
Fix control message *decoding* and add support for `ScmCredentials`

jonas-schievink:
While https://github.com/nix-rust/nix/pull/918 fixed the *encoding*, the *decoding* done by the `CmsgIterator` still remained broken when multiple messages are received. However, since nix didn't support any control message that could reliably be sent twice in one `sendmsg` call, this couldn't be tested properly.

This PR addresses this by adding `SCM_CREDENTIALS` support and testing all of this by passing both an `SCM_CREDENTIALS` and a `SCM_RIGHTS` message at the same time.

I've also verified that the resulting encoding is the same as for roughly equivalent C code.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -205,6 +205,18 @@ cfg_if! {
         }
         impl Eq for UnixCredentials {}
 
+        impl From<libc::ucred> for UnixCredentials {
+            fn from(cred: libc::ucred) -> Self {
+                UnixCredentials(cred)
+            }
+        }
+
+        impl Into<libc::ucred> for UnixCredentials {
+            fn into(self) -> libc::ucred {
+                self.0
+            }
+        }
+
         impl fmt::Debug for UnixCredentials {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 f.debug_struct("UnixCredentials")
@@ -359,7 +371,7 @@ impl<T> CmsgSpace<T> {
     }
 }
 
-#[allow(missing_debug_implementations)]
+#[derive(Debug)]
 pub struct RecvMsg<'a> {
     // The number of bytes received.
     pub bytes: usize,
@@ -374,15 +386,14 @@ impl<'a> RecvMsg<'a> {
     pub fn cmsgs(&self) -> CmsgIterator {
         CmsgIterator {
             buf: self.cmsg_buffer,
-            next: 0
         }
     }
 }
 
-#[allow(missing_debug_implementations)]
+#[derive(Debug)]
 pub struct CmsgIterator<'a> {
+    /// Control message buffer to decode from. Must adhere to cmsg alignment.
     buf: &'a [u8],
-    next: usize,
 }
 
 impl<'a> Iterator for CmsgIterator<'a> {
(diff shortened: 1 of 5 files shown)


src/sys/socket/mod.rs
```diff
@@ -392,53 +403,25 @@ impl<'a> Iterator for CmsgIterator<'a> {
...
+            // bytes we need, so we're at the end when the buffer is empty.
             return None;
         }
-        let cmsg: &'a cmsghdr = unsafe { &*(self.buf.as_ptr() as *const cmsghdr) };
 
-        // This check is only in the glibc implementation of CMSG_NXTHDR
-        // (although it claims the kernel header checks this), but such
-        // a structure is clearly invalid, either way.
+        // Safe if: `self.buf` is `cmsghdr`-aligned.
+        let cmsg: &'a cmsghdr = unsafe { &*(self.buf[..mem::size_of::<cmsghdr>()].as_ptr() as *const cmsghdr) };
```
asomers: How do you guarantee that self.buf is correctly aligned?

src/sys/socket/mod.rs
```diff
@@ -392,53 +403,25 @@ impl<'a> Iterator for CmsgIterator<'a> {
...
-                    &*(cmsg_data.as_ptr() as *const _)))
-            },
-            (_, _) => unsafe {
-                Some(ControlMessage::Unknown(UnknownCmsg(
-                    cmsg,
-                    slice::from_raw_parts(
-                        cmsg_data.as_ptr() as *const _,
-                        len))))
-            }
+        let cmsg_data = &self.buf[cmsg_align(mem::size_of::<cmsghdr>())..cmsg_len];
```
asomers: Up above you requird self.buf to be aligned.  Here, you seem to be handling the case where it isn't.  Or am I misunderstanding?

src/sys/socket/mod.rs
```diff
@@ -567,57 +569,95 @@ impl<'a> ControlMessage<'a> {
         }
     }
 
+    /// Returns the value to put into the `cmsg_type` field of the header.
+    fn type_(&self) -> libc::c_int {
```
asomers: I don't like trailing underscores.  How about naming this method `cmsg_type` instead?

src/sys/socket/mod.rs
```diff
@@ -567,57 +569,95 @@ impl<'a> ControlMessage<'a> {
...
+                cmsg_len: self.len() as _,
+                cmsg_level: libc::SOL_SOCKET,
+                cmsg_type: self.type_(),
+                ..mem::zeroed() // zero out platform-dependent padding fields
+            };
+            let buf = copy_bytes(&cmsg, buf);
+
+            match *self {
+                ControlMessage::ScmRights(fds) => {
+                    let padlen = cmsg_align(mem::size_of_val(&cmsg)) -
```
asomers: It looks like `padlen` and buf are identical for all cases.  Can you move their definitions outside of the `match` block?

src/sys/socket/mod.rs
jonas-schievink: The caller does. In this case its done by the `CmsgSpace`, which ensures alignment using a properly aligned member.

src/sys/socket/mod.rs
jonas-schievink: The `cmsg_align` just rounds up the size of `cmsghdr` to the right alignment. This is needed to include the right padding to reach the payload.

src/sys/socket/mod.rs
jonas-schievink: Well, it's not the caller but whoever creates the iterator. Here, the buffer comes from `recvmsg` and is a `CmsgSpace` which guarantees alignment.

jonas-schievink:
> Any idea why Travis didn't build this PR?

I added `[ci skip]` to the last commit as it was just updating the changelog. This causes Travis to ignore the commit which saves CI time.

jonas-schievink:
Uh, now the *old* test fails? That's odd.

jonas-schievink:
I think the issues are in qemu. I haven't managed to get any socket tests failing on my Raspberry Pi 1 (armv6) or 2 (armv7). Additionally, I wrote a test program in C (https://gist.github.com/jonas-schievink/cb6e6584a055539d2113f22d91068e2d) that works fine. The control message encoding for nix and the test program is identical, and it's the kernel (or qemu) that strips out the second message, not our decoder doing weird stuff.

I'm going to build a smaller reproducer and report the bug to qemu, then `#[ignore]` the tests here.

jonas-schievink:
@asomers Alright, after a long journey this might be finally done (except the commit message for 
386cd04 is wrong - but I'll fix that when squashing later). Care to take another look?

test/sys/test_socket.rs
```diff
@@ -301,14 +301,36 @@ fn test_scm_credentials() {
 /// Ensure that we can send `SCM_CREDENTIALS` and `SCM_RIGHTS` with a single
 /// `sendmsg` call.
 #[cfg(any(target_os = "android", target_os = "linux"))]
+// qemu's handling of multiple cmsgs is bugged, ignore tests on non-x86-64
+#[cfg_attr(not(target_arch = "x86_64"), ignore)]
```
asomers: What about i386?  That doesn't use qemu on FreeBSD, and I don't think it does on Linux either.

test/sys/test_socket.rs
```diff
@@ -301,14 +301,36 @@ fn test_scm_credentials() {
 /// Ensure that we can send `SCM_CREDENTIALS` and `SCM_RIGHTS` with a single
 /// `sendmsg` call.
 #[cfg(any(target_os = "android", target_os = "linux"))]
+// qemu's handling of multiple cmsgs is bugged, ignore tests on non-x86-64
```
asomers: Can you add an upstream bug link?  Your C test case should make it easy to report.

src/sys/socket/mod.rs
asomers: Wrap to 80 columns, please.

asomers:
@jonas-schievink please squash this; then I'll merge it.

Susurrus:
@asomers This all look good? You asked for it to be squashed, which is has been. We could split this up into two commits with the other one added the `as_raw` implementations for Pid and friends, but it's a minor nit, so I'm fine not worrying about it.

asomers:
bors retry

PR #927, opened 2018-07-11
Enable the select tests on powerpc and mips

jonas-schievink:
They pass at least on my machine with QEMU, so the underlying bugs might
have been fixed.

### DIFF
--- a/src/sys/select.rs
+++ b/src/sys/select.rs
@@ -273,14 +273,7 @@ mod tests {
         assert_eq!(set.highest(), Some(7));
     }
 
-    // powerpc-unknown-linux-gnu currently fails on the first `assert_eq` because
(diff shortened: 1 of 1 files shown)


asomers:
When you originally disabled those tests, did they fail reliably or intermittently?

jonas-schievink:
Oh, I didn't disable them, I just copied them from another file in https://github.com/nix-rust/nix/commit/885a1f751497c265a532a42a0c98829e32ae7f82. @Susurrus disabled them in https://github.com/nix-rust/nix/pull/590. I've run them a few times and they haven't failed yet, so I doubt it's an intermittent failure.

asomers:
Ok, that should be fine then.

bors r+

PR #928, opened 2018-07-11
Wrap libc::statfs

### LINKED ISSUE #926: Can't use sys::statfs::statfs without unsafe
Current implementation of `statfs` requires mutable reference of result variable, but I can't pass it to `statfs` without initialization. For example this code
 ```rust
let mut statf: libc::statfs;
sys::statfs::statfs(Path::new(&String::from("/")), &mut statf).unwrap();
```
 gives me this error: 
```rust
error[E0381]: use of possibly uninitialized variable: `statf`
sys::statfs::statfs(Path::new(&String::from("/")), &mut statf).unwrap();
                                                        ^^^^^ use of possibly uninitialized `statf`
```
So I need to somehow initialize the variable:
 ```rust
let mut statf: libc::statfs = unsafe { mem::uninitialized() };
sys::statfs::statfs(Path::new(&String::from("/")), &mut statf).unwrap();
```
Maybe it will be better to return result with `libc::statfs` struct instead of empty result?

### PR DESCRIPTION
**Pull request by alesharik**
Fix for [#926](https://github.com/nix-rust/nix/issues/926)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1077](https://github.com/nix-rust/nix/pull/1077))
 - Minimum supported Rust version is now 1.25.0 
   ([#1035](https://github.com/nix-rust/nix/pull/1035))
+- Now functions `statfs()` and `fstatfs()` return result with `Statfs` wrapper
+  ([#928](https://github.com/nix-rust/nix/pull/928))
 
 ### Fixed
 ### Removed
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -71,7 +71,14 @@ pub mod socket;
 
 pub mod stat;
 
-#[cfg(any(target_os = "android", target_os = "linux", target_os = "macos"))]
+#[cfg(any(target_os = "android",
+          target_os = "dragonfly",
+          target_os = "freebsd",
+          target_os = "ios",
+          target_os = "linux",
+          target_os = "macos",
+          target_os = "openbsd"
+))]
 pub mod statfs;
 
 pub mod statvfs;
--- a/src/sys/statfs.rs
+++ b/src/sys/statfs.rs
@@ -1,20 +1,548 @@
-use {Result, NixPath};
-use errno::Errno;
+use std::fmt::{self, Debug};
+use std::mem;
 use std::os::unix::io::AsRawFd;
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+use std::ffi::CStr;
+
 use libc;
 
-pub fn statfs<P: ?Sized + NixPath>(path: &P, stat: &mut libc::statfs) -> Result<()> {
-    unsafe {
-        Errno::clear();
-        let res = path.with_nix_path(|path| libc::statfs(path.as_ptr(), stat))?;
+use {NixPath, Result};
+use errno::Errno;
 
-        Errno::result(res).map(drop)
+#[cfg(target_os = "android")]
+pub type fsid_t = libc::__fsid_t;
+#[cfg(not(target_os = "android"))]
+pub type fsid_t = libc::fsid_t;
+
+#[derive(Clone, Copy)]
+pub struct Statfs(libc::statfs);
+
+#[cfg(target_os = "freebsd")]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(u32);
+#[cfg(target_os = "android")]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(libc::c_ulong);
+#[cfg(all(target_os = "linux", target_arch = "s390x"))]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(u32);
+#[cfg(all(target_os = "linux", target_env = "musl"))]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(libc::c_ulong);
+#[cfg(all(target_os = "linux", not(any(target_arch = "s390x", target_env = "musl"))))]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(libc::c_long);
+
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const ADFS_SUPER_MAGIC: FsType = FsType(libc::ADFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const AFFS_SUPER_MAGIC: FsType = FsType(libc::AFFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const CODA_SUPER_MAGIC: FsType = FsType(libc::CODA_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const CRAMFS_MAGIC: FsType = FsType(libc::CRAMFS_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const EFS_SUPER_MAGIC: FsType = FsType(libc::EFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const EXT2_SUPER_MAGIC: FsType = FsType(libc::EXT2_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const EXT3_SUPER_MAGIC: FsType = FsType(libc::EXT3_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const EXT4_SUPER_MAGIC: FsType = FsType(libc::EXT4_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const HPFS_SUPER_MAGIC: FsType = FsType(libc::HPFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const HUGETLBFS_MAGIC: FsType = FsType(libc::HUGETLBFS_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const ISOFS_SUPER_MAGIC: FsType = FsType(libc::ISOFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const JFFS2_SUPER_MAGIC: FsType = FsType(libc::JFFS2_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const MINIX_SUPER_MAGIC: FsType = FsType(libc::MINIX_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const MINIX_SUPER_MAGIC2: FsType = FsType(libc::MINIX_SUPER_MAGIC2);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const MINIX2_SUPER_MAGIC: FsType = FsType(libc::MINIX2_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const MINIX2_SUPER_MAGIC2: FsType = FsType(libc::MINIX2_SUPER_MAGIC2);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const MSDOS_SUPER_MAGIC: FsType = FsType(libc::MSDOS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const NCP_SUPER_MAGIC: FsType = FsType(libc::NCP_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const NFS_SUPER_MAGIC: FsType = FsType(libc::NFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const OPENPROM_SUPER_MAGIC: FsType = FsType(libc::OPENPROM_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const PROC_SUPER_MAGIC: FsType = FsType(libc::PROC_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const QNX4_SUPER_MAGIC: FsType = FsType(libc::QNX4_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
+pub const REISERFS_SUPER_MAGIC: FsType = FsType(libc::REISERFS_SUPER_MAGIC);
+#[cfg(all(target_os = "linux", not(target_env = "musl"), not(target_arch = "s390x")))]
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: src/sys/statfs.rs [hunk 1]
```diff
@@ -1,22 +1,239 @@
...
+
+#[cfg(test)]
+mod test {
+    use sys::statfs::*;
+    use std::mem::uninitialized;
+    use std::fs::File;
+
+    #[test]
+    fn statfs_new_call() {
+        let statfs1 = statfs_new("/proc".as_bytes()).unwrap();
```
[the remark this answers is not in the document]
alesharik: Is there any better example file? `/tmp`?

asomers:
Yeah, `/tmp` should be safe.  But you don't know what filesystem it will be using.

src/sys/statfs.rs
```diff
@@ -1,22 +1,118 @@
...
+pub struct Statfs(libc::statfs);
+
+impl Statfs {
+    /// Magic code defining system type
+    pub fn filesystem_type(&self) -> c_ulong {
+        self.0.f_type as c_ulong
+    }
+
+    /// Optimal transfer block size
+    #[cfg(any(target_os = "freebsd", target_os="macos", target_os="ios"))]
```
asomers: NetBSD does it this way too.  I haven't checked Dragonfly or Openbsd.

src/sys/statfs.rs
```diff
@@ -1,22 +1,118 @@
...
+        self.0.f_ffree as u64
+    }
+
+    /// Filesystem ID
+    pub fn filesystem_id(&self) -> fsid_t {
+        self.0.f_fsid
+    }
+
+    /// Maximum length of filenames
+    #[cfg(all(not(any(target_os="ios", target_os="macos")), target_os = "freebsd"))]
```
asomers: This `#[cfg()]` directive is redundant.  Is it really what you intended?

src/sys/statfs.rs
```diff
@@ -1,22 +1,120 @@
...
+use libc::fsid_t;
+
+#[derive(Clone, Copy)]
+#[allow(missing_debug_implementations)]
+pub struct Statfs(libc::statfs);
+
+impl Statfs {
+    /// Magic code defining system type
+    pub fn filesystem_type(&self) -> c_ulong {
+        self.0.f_type as c_ulong
```
asomers: This API is really nasty.  Depending on the platform, `f_type` can be defined as either a long, ulong, int, uint32_t, or uint64_t.  On Linux, you'll most likely want to compare it to the magic constants, which are defined as either long or int.  On BSD, the value means nothing by itself, and there are no constants to compare it to.  So I think the best API that Nix can present would be to return a Newtype that derives `Eq`.  On Linux, it can define constants for the various magic constants.  On BSD, it should define none, and be totally opaque.

src/sys/statfs.rs
```diff
@@ -1,22 +1,120 @@
...
+    /// Optimal transfer block size
+    #[cfg(any(target_os = "freebsd", target_os="macos", target_os="ios"))]
+    pub fn optimal_transfer_size(&self) -> c_ulong {
+        self.0.f_iosize as c_ulong
+    }
+
+    /// Optimal transfer block size
+    #[cfg(not(any(target_os = "freebsd", target_os="macos", target_os="ios")))]
+    pub fn optimal_transfer_size(&self) -> c_ulong {
+        self.0.f_bsize as c_ulong
```
asomers: This API is almost as bad as `f_type`.  It is variously defined as `uint64_t`, `c_uint`, `c_ulong`, or `c_long`.  In cases like this, Nix usually exposes the underlying feature with as little modification as possible, except for what's required to make it Rust-friendly.  In this case, I think Nix should use the native type of each platform instead of casting everything to the lowest common denominator.

src/sys/statfs.rs
asomers: Yeah, but use the constants from libc instead of hard-coding them.  And the inner type should be whatever libc exports instead of u64.

src/sys/statfs.rs
```diff
@@ -1,46 +1,148 @@
...
+
 impl Statfs {
     /// Magic code defining system type
-    pub fn filesystem_type(&self) -> c_ulong {
-        self.0.f_type as c_ulong
+    pub fn filesystem_type(&self) -> FsType {
+        FsType(self.0.f_type)
+    }
+
+
```
asomers: No need to double space.

src/sys/mod.rs
```diff
@@ -65,7 +65,7 @@ pub mod socket;
 
 pub mod stat;
 
-#[cfg(any(target_os = "android", target_os = "linux", target_os = "macos"))]
+#[cfg(not(any(target_os = "freebsd", target_os = "netbsd")))]
```
asomers: Why are you excluding freebsd and netbsd?

asomers: Well, I'll be darned.  Can you please submit a PR to libc to fix that?

src/sys/mod.rs
```diff
@@ -65,7 +65,7 @@ pub mod socket;
 
 pub mod stat;
 
-#[cfg(not(any(target_os = "freebsd", target_os = "netbsd")))]
+#[cfg(not(target_os = "netbsd"))]
```
asomers: What's wrong with netbsd?

src/sys/statfs.rs
```diff
@@ -88,31 +88,41 @@ impl Statfs {
...
-
     /// Optimal transfer block size
     #[cfg(any(target_os = "ios", target_os = "macos"))]
     pub fn optimal_transfer_size(&self) -> libc::int32_t {
         self.0.f_iosize
     }
 
     /// Optimal transfer block size
-    #[cfg(target_os = "android")]
+    #[cfg(target_arch = "s390x")]
```
asomers: You should probably also have a target_os = "linux" check in the guard.

src/sys/statfs.rs
```diff
@@ -88,31 +88,41 @@ impl Statfs {
...
     /// Optimal transfer block size
-    #[cfg(not(any(target_os = "linux", target_os = "android", target_os = "ios", target_os = "macos")))]
-    pub fn optimal_transfer_size(&self) -> libc::c_ulong {
+    #[cfg(target_os = "freebsd")]
+    pub fn optimal_transfer_size(&self) -> libc::uint64_t {
         self.0.f_iosize
     }
 
+    /// Optimal transfer block size
+    #[cfg(not(any(target_os = "linux", target_os = "android", target_os = "ios", target_os = "macos", target_os = "freebsd")))]
```
asomers: This guard would be clearer expressed in the positive sense than in the negative.

src/sys/statfs.rs
```diff
@@ -149,7 +149,8 @@ impl Statfs {
     }
 
     /// Size of a block
-    #[cfg(not(any(target_os = "macos", target_os = "freebsd", target_os = "ios", target_os = "android", target_arch = "s390x", all(target_os = "linux", target_env = "musl"))))]
+
+    #[cfg(all(target_os = "linux", not(target_arch = "s390x"), not(target_env = "musl")))]
```
asomers: 80 columns per line, please

src/sys/statfs.rs
```diff
@@ -1,22 +1,244 @@
...
+
+#[cfg(all(target_os = "linux", not(any(target_env = "musl", target_arch = "s390x"))))]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(libc::c_long);
+
+#[cfg(any(target_os = "android", all(target_os = "linux", target_env = "musl")))]
+#[derive(Eq, Copy, Clone, PartialEq, Debug)]
+pub struct FsType(libc::c_ulong);
+
+#[cfg(not(any(target_os = "linux", target_os = "ios", target_os = "macos", target_os = "android", target_os="freebsd")))]
```
asomers: This `#[cfg()]` condition is getting very confusing.  Could you please convert it into the positive form?  Also wrap to 80 columns.

src/sys/statfs.rs
```diff
@@ -1,22 +1,244 @@
...
+    }
+
+    /// Optimal transfer block size
+    #[cfg(target_os = "freebsd")]
+    pub fn optimal_transfer_size(&self) -> libc::uint64_t {
+        self.0.f_iosize
+    }
+
+    /// Optimal transfer block size
+    #[cfg(not(any(target_os = "linux", target_os = "android", target_os = "ios", target_os = "macos", target_os = "freebsd")))]
```
asomers: This `#[cfg()]` would also be better written in the positive sense, and wrapped to 80 columns.

src/sys/statfs.rs
```diff
@@ -1,22 +1,300 @@
...
-        Errno::clear();
-        Errno::result(libc::fstatfs(fd.as_raw_fd(), stat)).map(drop)
+        let mut stat: Statfs = mem::uninitialized();
+        Errno::result(libc::fstatfs(fd.as_raw_fd(), &mut stat.0)).map(|_| stat)
+    }
+}
+
+#[cfg(test)]
+#[cfg(not(any(
+    target_os = "macos",
```
asomers: Why exclude the tests on OSX?

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#921](https://github.com/nix-rust/nix/pull/921))
 - Added support for `SCM_CREDENTIALS`, allowing to send process credentials over Unix sockets.
   ([#923](https://github.com/nix-rust/nix/pull/923))
+- Added a `statfs` wrapper.
```
asomers: You didn't add a `statfs` wrapper because it already existed.  You should remove this changelog entry, but leave the entry in the Changed section.

src/sys/mod.rs
asomers: Probably better to use a positive list here, not a negative list.  That's Nix's usual style, and this way you'll avoid breaking the build on platforms that don't have CI, like Dragonfly and emscripten.

src/sys/statfs.rs
```diff
@@ -1,22 +1,432 @@
...
-            path.with_nix_path(|path| libc::statfs(path.as_ptr(), stat))
-        );
+#[cfg(target_os = "android")]
+pub type fsid_t = libc::__fsid_t;
+#[cfg(not(target_os = "android"))]
+pub type fsid_t = libc::fsid_t;
+
+#[derive(Clone, Copy)]
+#[allow(missing_debug_implementations)]
+pub struct Statfs(libc::statfs);
```
asomers: It would be very handy to have Debug for this structure.  You could implement it manually.  See SigEvent in sys/signal.rs for an example.  I wouldn't hold up the PR just for that, though.

src/sys/statfs.rs
```diff
@@ -1,22 +1,432 @@
...
+        self.0.f_namelen
+    }
+
+    /// Total data blocks in filesystem
+    #[cfg(any(
+        target_os = "ios",
+        target_os = "macos",
+        target_os = "android",
+        target_os = "freebsd",
+        target_pointer_width = "64",
```
asomers: Here and elsewhere, this condition is confusing and incorrect.  Better written as `all(target_os = "linux", target_pointer_width = "64")`.  OpenBSD should be added to the list here too.

src/sys/statfs.rs
```diff
@@ -1,22 +1,432 @@
...
+
+#[cfg(test)]
+#[cfg(not(target_os = "ios"))]
+mod test {
+    use std::fs::File;
+    use sys::statfs::*;
+
+    #[test]
+    fn statfs_call() {
+        statfs("/tmp".as_bytes()).unwrap();
```
asomers: These tests don't provide much coverage.  Could you check the output somehow?  One way would be to compare the output of `statfs` with `statvfs`.

asomers:
Since it's been so long since the original submission, I'm going to ask that you rebase onto master before I review it.  Be sure to rebase, don't just merge.  Merging makes the changes harder to review.  And I probably won't have time to review before Sunday night.  If I still haven't reviewed by Monday, feel free to ping me.

alesharik:
Hi, @asomers . Any updates on reviewing my changes?

src/sys/statfs.rs
```diff
@@ -1,20 +1,665 @@
...
+        FsType(self.0.f_type)
+    }
+
+    /// Magic code defining system type
+    #[cfg(any(
+        target_os = "openbsd",
+        target_os = "ios",
+        target_os = "macos"
+    ))]
+    pub fn filesystem_type(&self) -> &str {
```
asomers: The two `filesystem_type` methods are not equivalent.  They aren't mutually exclusive, either.  FreeBSD for example has both fields.  You should probably expose both methods wherever they are supported.

src/sys/statfs.rs
```diff
@@ -1,20 +1,665 @@
...
+        let c_str = unsafe { CStr::from_ptr(self.0.f_fstypename.as_ptr()) };
+        c_str.to_str().unwrap()
+    }
+
+    /// Optimal transfer block size
+    #[cfg(any(
```
asomers: These long chains of `#[cfg()]`ed methods might be more legible if you write them with `cfg_if!`.

src/sys/statfs.rs
```diff
@@ -1,20 +1,665 @@
...
+        self.0.f_iosize
+    }
+
+    /// Size of a block
+    #[cfg(any(
+        target_os = "ios",
+        target_os = "macos",
+        target_os = "openbsd"
+    ))]
+    pub fn block_size(&self) -> libc::uint32_t {
```
asomers: Both the FreeBSD and Linux man pages describe this field as the fragment size.  Should you change the name accordingly?

src/sys/statfs.rs
```diff
@@ -1,20 +1,665 @@
...
+    pub fn files_free(&self) -> libc::uint32_t {
+        self.0.f_ffree
+    }
+
+    /// Free file nodes in filesystem
+    #[cfg(any(
+        target_os = "freebsd",
+        target_os = "openbsd"
+    ))]
+    pub fn files_free(&self) -> i64 {
```
asomers: Why use `i64` here but `libc::int64_t` elsewhere?

src/sys/statfs.rs
```diff
@@ -1,20 +1,665 @@
...
+    }
+
+    fn assert_fs_equals(fs: Statfs, vfs: Statvfs) {
+        assert_eq!(fs.files() as u64, vfs.files() as u64);
+        assert_eq!(fs.blocks() as u64, vfs.blocks() as u64);
+        assert_eq!(fs.block_size() as u64, vfs.fragment_size() as u64);
+    }
+
+    #[test]
+    #[ignore]
```
asomers: Why ignored?  You should always explain why.

src/sys/statfs.rs
```diff
@@ -1,20 +1,665 @@
...
+    fn fstatfs_call() {
+        check_fstatfs("/tmp");
+        check_fstatfs("/dev");
+        check_fstatfs("/run");
+        check_fstatfs("/");
+    }
+
+    fn check_fstatfs(path: &str) {
+        let vfs = statvfs(path.as_bytes());
+        if vfs.is_err() {
```
asomers: You shouldn't ignore all errors from `statvfs` and friends.  Better to check that the path exists and then assert on error.

src/sys/statfs.rs
alesharik: Seems like [cfg_if doesn't support functions inside impl](https://github.com/alexcrichton/cfg-if/issues/10). Should I use my own macro to replace `#[cfg()]`?

asomers: I guess you should probably just use plain old `#[cfg]` in that case.

alesharik:
Hi, @asomers . [File `src/sys/signal.rs` is broken an android](https://travis-ci.org/nix-rust/nix/jobs/537350981) therefore Travis fails the build. Can you do something about it?

asomers:
It looks like this is still a WIP; let me know when you're ready for a review.  Also, you'll need to rebase.  Be sure to fix the CHANGELOG entry when you rebase.

alesharik:
@asomers I'm ready for review

Cargo.toml
```diff
@@ -16,7 +16,7 @@ exclude     = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc" }
+libc = { version = "0.2.57", features = ["extra_traits"] }
```
asomers: Don't switch libc from a git dependency to a version dependency.  We only do that at release time.  Also, we'll probably merge #1035 first, so you won't have to change this file.

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ```
 
 ### Changed
+- Now functions `statfs()` and `fstatfs()` return result with `Statfs` wrapper
```
asomers: The change log entry should go in the [Unreleased] section.

src/sys/statfs.rs
```diff
@@ -1,20 +1,549 @@
...
+        if !Path::new(path).exists() {
+            return;
+        }
+        let vfs = statvfs(path.as_bytes()).unwrap();
+        let file = File::open(path).unwrap();
+        let fs = fstatfs(&file).unwrap();
+        assert_fs_equals(fs, vfs);
+    }
+
+    fn check_statfs(path: &str) {
```
asomers: What's the difference between check_fstatfs and check_statfs ?

alesharik:
@asomers Ready for review

alesharik:
Ok, no problem. I'll fix conflicts when #1035 is merged.

asomers:
Ok, 1035 is merged.  You should be able to rebase now.  You should no longer need `impl Debug for StatFs`.

asomers:
Please squash your commits.

alesharik:
@asomers any updates on merging this PR?

PR #929, opened 2018-07-14
Enable most AIO tests on musl targets and clean them up

jonas-schievink:
I've run them a bunch of times on my machine and haven't witnessed any spurious failures so far.

The exception are the signal event tests, which I've kept ignored. They deadlock on some target combinations - apparently the signal handler never runs. This is not a QEMU bug since it also happens on `i686-unknown-linux-musl`.

### DIFF
--- a/test/sys/test_aio.rs
+++ b/test/sys/test_aio.rs
@@ -6,7 +6,6 @@ use nix::sys::aio::*;
 use nix::sys::signal::{SaFlags, SigAction, sigaction, SigevNotify, SigHandler, Signal, SigSet};
 use nix::sys::time::{TimeSpec, TimeValLike};
 use std::io::{Write, Read, Seek, SeekFrom};
-use std::ops::Deref;
 use std::os::unix::io::AsRawFd;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::{thread, time};
@@ -168,8 +167,8 @@ fn test_aio_suspend() {
         }
     }
 
-    assert!(wcb.aio_return().unwrap() as usize == WBUF.len());
-    assert!(rcb.aio_return().unwrap() as usize == rlen);
+    assert_eq!(wcb.aio_return().unwrap() as usize, WBUF.len());
+    assert_eq!(rcb.aio_return().unwrap() as usize, rlen);
 }
 
 // Test a simple aio operation with no completion notification.  We must poll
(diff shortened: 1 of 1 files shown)


jonas-schievink:
`x86_64-unknown-linux-musl` fails with:

```
test sys::test_aio::test_liocb_listio_read_immutable ... fatal runtime error: failed to initiate panic, error 5
error: An unknown error occurred
```

That's... interesting. I can't reproduce this locally with the same rustc version (1.20). Neither with `cross test --target x86_64-unknown-linux-musl -- test_aio` nor with `cargo` using the same command line. Panics also work without any problems on my system with musl.

I was thinking that this might be related to system library versions, but musl doesn't really have problems with that, right (I'm on Arch Linux, so I'm running very recent versions of everything)? At least I don't have my system's musl installed, and both I and Travis are using the regular `rustup`-provided Rust releases.

I guess I'll disable the test for now.

test/sys/test_aio.rs
```diff
@@ -433,20 +418,19 @@ lazy_static! {
 }
 
 extern fn sigfunc(_: c_int) {
-    SIGNALED.store(true, Ordering::Relaxed);
+    SIGNALED.store(true, Ordering::SeqCst);
```
asomers: `Relaxed` is fine here, because `SIGNALED` isn't being used to control access to any other variables.  It's only being accessed by itself, so stricter atomic instructions aren't needed.  This is the kind of application that, in C, wouldn't even use special instructions, just a `sigatomic_t` variable.

test/sys/test_aio.rs
```diff
@@ -562,26 +544,25 @@ fn test_liocb_listio_nowait() {
...
+    assert_eq!(rbuf2, EXPECT);
 }
 
 // Test LioCb::listio with LIO_NOWAIT and a SigEvent to indicate when all
 // AioCb's are complete.
 // FIXME: This test is ignored on mips/mips64 because of failures in qemu in CI.
 #[test]
 #[cfg(not(any(target_os = "ios", target_os = "macos")))]
-#[cfg_attr(any(target_arch = "mips", target_arch = "mips64", target_env = "musl"), ignore)]
+#[cfg_attr(any(all(target_arch = "x86", target_env = "musl"), target_arch = "mips", target_arch = "mips64"), ignore)]
```
asomers: If this test fails on x86 with musl, why do you think it will work better on other architectures with musl?

jonas-schievink:
Sorry, don't really have the time to finish this PR (mostly because doing it properly would involve debugging musl). If anyone wants to finish this, feel free, although I don't really see why you'd want to given that there's better alternatives to aio.

PR #930, opened 2018-07-20
Add wrapper for linux kernel module loading

bachp:
- init_module and finit_module to load kernel modules
- delete_module to unload kernel modules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#923](https://github.com/nix-rust/nix/pull/923))
 - Added a `dir` module for reading directories (wraps `fdopendir`, `readdir`, and `rewinddir`).
   ([#916](https://github.com/nix-rust/nix/pull/916))
+- Added `kmod` module that allows loading and unloading kernel modules on Linux.
+  ([#930](https://github.com/nix-rust/nix/pull/930))
 
 ### Changed
 - Increased required Rust version to 1.22.1/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ cc = "1"
 [dev-dependencies]
 bytes = "0.4.8"
 lazy_static = "1"
-rand = "0.4"
+rand = "0.5"
 tempfile = "3"
 
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
--- a/src/kmod.rs
+++ b/src/kmod.rs
@@ -1,0 +1,123 @@
+//! Load and unload kernel modules.
+//!
+//! For more details see
+
+use libc;
+use std::ffi::CStr;
+use std::os::unix::io::AsRawFd;
+
+use errno::Errno;
+use Result;
+
+/// Loads a kernel module from a buffer.
+///
+/// It loads an ELF image into kernel space,
+/// performs any necessary symbol relocations,
+/// initializes module parameters to values provided by the caller,
+/// and then runs the module's init function.
+///
+/// This function requires `CAP_SYS_MODULE` privilege.
+///
+/// The `module_image` argument points to a buffer containing the binary image
+/// to be loaded. The buffer should contain a valid ELF image
+/// built for the running kernel.
+///
+/// The `param_values` argument is a string containing space-delimited specifications
+/// of the values for module parameters.
+/// Each of the parameter specifications has the form:
+///
+/// `name[=value[,value...]]`
+///
+/// # Example
+///
+/// ```no_run
+/// use std::fs::File;
+/// use std::io::Read;
+/// use std::ffi::CString;
+/// use nix::kmod::init_module;
+///
+/// let mut f = File::open("mykernel.ko").unwrap();
+/// let mut contents: Vec<u8> = Vec::new();
+/// f.read_to_end(&mut contents).unwrap();
+/// init_module(&mut contents, &CString::new("who=Rust when=Now,12").unwrap()).unwrap();
+/// ```
+///
+/// See [`man init_module(2)`](http://man7.org/linux/man-pages/man2/init_module.2.html) for more information.
+pub fn init_module(module_image: &[u8], param_values: &CStr) -> Result<()> {
+    let res = unsafe {
+        libc::syscall(
+            libc::SYS_init_module,
+            module_image.as_ptr(),
+            module_image.len(),
+            param_values.as_ptr(),
+        )
+    };
+
+    Errno::result(res).map(drop)
+}
+
+libc_bitflags!(
+    /// Flags used by the `finit_module` function.
+    pub struct ModuleInitFlags: libc::c_uint {
+        /// Ignore symbol version hashes.
+        MODULE_INIT_IGNORE_MODVERSIONS;
+        /// Ignore kernel version magic.
+        MODULE_INIT_IGNORE_VERMAGIC;
+    }
+);
+
+/// Loads a kernel module from a given file descriptor.
+///
+/// # Example
+///
+/// ```no_run
+/// use std::fs::File;
+/// use std::ffi::CString;
+/// use nix::kmod::{finit_module, ModuleInitFlags};
+///
+/// let f = File::open("mymod.ko").unwrap();
+/// finit_module(&f, &CString::new("").unwrap(), ModuleInitFlags::empty()).unwrap();
+/// ```
+///
+/// See [`man init_module(2)`](http://man7.org/linux/man-pages/man2/init_module.2.html) for more information.
+pub fn finit_module<T: AsRawFd>(fd: &T, param_values: &CStr, flags: ModuleInitFlags) -> Result<()> {
+    let res = unsafe {
+        libc::syscall(
+            libc::SYS_finit_module,
+            fd.as_raw_fd(),
+            param_values.as_ptr(),
+            flags.bits(),
+        )
+    };
+
+    Errno::result(res).map(drop)
+}
+
+libc_bitflags!(
+    /// Flags used by `delete_module`.
+    ///
+    /// See [`man delete_module(2)`](http://man7.org/linux/man-pages/man2/delete_module.2.html)
+    /// for a detailed description how these flags work.
+    pub struct DeleteModuleFlags: libc::c_int {
+        O_NONBLOCK;
+        O_TRUNC;
+    }
+);
+
+/// Unloads the kernel module with the given name.
+///
+/// # Example
+///
+/// ```no_run
+/// use std::ffi::CString;
+/// use nix::kmod::{delete_module, DeleteModuleFlags};
+///
+/// delete_module(&CString::new("mymod").unwrap(), DeleteModuleFlags::O_NONBLOCK).unwrap();
+/// ```
+///
+/// See [`man delete_module(2)`](http://man7.org/linux/man-pages/man2/delete_module.2.html) for more information.
+pub fn delete_module(name: &CStr, flags: DeleteModuleFlags) -> Result<()> {
+    let res = unsafe { libc::syscall(libc::SYS_delete_module, name.as_ptr(), flags.bits()) };
+
+    Errno::result(res).map(drop)
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,7 +43,11 @@ pub mod fcntl;
           target_os = "netbsd",
           target_os = "openbsd"))]
 pub mod ifaddrs;
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(any(target_os = "android",
+          target_os = "linux"))]
+pub mod kmod;
+#[cfg(any(target_os = "android",
+          target_os = "linux"))]
 pub mod mount;
 #[cfg(any(target_os = "dragonfly",
           target_os = "freebsd",
@@ -57,7 +61,8 @@ pub mod net;
 pub mod poll;
 #[deny(missing_docs)]
 pub mod pty;
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(any(target_os = "android",
+          target_os = "linux"))]
 pub mod sched;
 pub mod sys;
 // This can be implemented for other platforms as soon as libc
--- a/test/test.rs
+++ b/test/test.rs
@@ -9,6 +9,21 @@ extern crate libc;
 extern crate rand;
 extern crate tempfile;
 
+macro_rules! skip_if_not_root {
+    ($name:expr) => {
+        use nix::unistd::Uid;
+        use std;
+        use std::io::Write;
+
+        if !Uid::current().is_root() {
+            let stderr = std::io::stderr();
+            let mut handle = stderr.lock();
+            writeln!(handle, "{} requires root privileges. Skipping test.", $name).unwrap();
+            return;
+        }
+    };
+}
+
 mod sys;
 mod test_dir;
 mod test_fcntl;
--- a/test/test.rs
+++ b/test/test.rs
@@ -27,6 +27,9 @@ macro_rules! skip_if_not_root {
 mod sys;
 mod test_dir;
 mod test_fcntl;
+#[cfg(any(target_os = "android",
+          target_os = "linux"))]
+mod test_kmod;
 #[cfg(any(target_os = "dragonfly",
           target_os = "freebsd",
           target_os = "fushsia",
--- a/test/test_kmod/hello_mod/Makefile
+++ b/test/test_kmod/hello_mod/Makefile
@@ -1,0 +1,7 @@
+obj-m += hello.o
+
+all:
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) modules
+
+clean:
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) clean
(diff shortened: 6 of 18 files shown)


src/kmod.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Example usage:
+/// 
+/// ```
+/// let mut f = File::open("mykernel.ko").unwrap();
+/// let mut contents: Vec<u8> = Vec::new();
+/// f.read_to_end(&mut contents).unwrap();
+/// init_module(&mut contents, &CString::new("").unwrap()).unwrap();
+/// ```
+/// 
+pub fn init_module(module_image: &[u8], param_values: &CStr) -> Result<i64> {
```
bachp: Is there a more flexible way then passing the binary as `&[u8]`?

src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub mod fcntl;
           target_os = "openbsd"))]
 pub mod ifaddrs;
 #[cfg(any(target_os = "linux", target_os = "android"))]
+pub mod kmod;
```
bachp: Not sure if this is the right place to put this nor if this is the proper name. Inputs are welcome.

bachp:
@asomers A simple hello world module as describe [here](https://www.tldp.org/LDP/lkmpg/2.6/html/x121.html) could be used. But as the kernel ABI is not stable it needs to be compiled for the currently running version of Linux. I'm trying to get it compiled during the test using the `cc` crate, but with no luck so far.

bachp:
@asomers What is the best way to run tests as root? Should I run all of the build as root or is there a better way?

asomers:
Do it just like the `test_setgroups` test.  The test should quit early if it's not running as root.  When you want to actually run that test by hand, run all of the tests as root.

bachp:
@asomers I added some tests

test/test_kmod/hello_mod/hello-1.c
```diff
@@ -0,0 +1,22 @@
...
+	 */
+	return 0;
+}
+
+void cleanup_module(void)
+{
+	printk(KERN_INFO "Goodbye world 1.\n");
+}
+
+MODULE_LICENSE("GPL");
```
asomers: Why GPL?  The rest of Nix uses MIT.

test/test_kmod/mod.rs
```diff
@@ -0,0 +1,93 @@
+use std::process::Command;
+
+fn compile_kernel_module() -> String {
+    #[allow(unused_variables)]
```
asomers: The `#[allow(unused_variables)]` syntax is holdover from an older version of rustc.  Nowadays you can just do `let _, = ...` and it won't complain.

asomers:
On my Debian system the test_finit_and_delete_module test fails with this error:
```
make -C /lib/modules/4.9.0-7-amd64/build M=/home/somers/src/rust/nix/test/test_kmod/hello_mod modules
make[1]: *** /lib/modules/4.9.0-7-amd64/build: No such file or directory.  Stop.
Makefile:4: recipe for target 'all' failed
make: *** [all] Error 2
FAILED

failures:

---- test_kmod::test_finit_and_delete_module stdout ----
        thread 'test_kmod::test_finit_and_delete_module' panicked at 'assertion failed: status.success()', test/test_kmod/mod.rs:12:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

bachp:
@asomers You most probably don't have the kernel sources installed on your system. Building a kernel module requires the kernel source. I'm not sure what packages provides these on Debian.

test/test_kmod/hello_mod/hello-1.c
bachp: The module is taken from tldp.org where it is licensed GPL.

test/test_kmod/mod.rs
asomers: Sorry, my last comment had a typo.  You need to use `let _m = ...`, not `let _ = ...`.  There is a difference between `_m` and `_`.

asomers:
I installed the linux-source package, and now I have sources in /usr/src, but I still get the same error.  The fact that the error message mentions a "build" directory suggests that maybe there's a missing step?

bachp:
I tried on a debian system and the required package is `linux-headers-*`. 

You can install it via:
```
apt install linux-headers-$(uname -r)
```

test/test_kmod/mod.rs
bachp: Good to know. I didn't know they behave differently. I update the code.

test/test_kmod/hello_mod/hello-1.c
Susurrus: This also isn't specific enough, as the GPL has versions. I assume that means 2.0. Let's get a new one written that we can license MIT. Also, can we name it `hello.c`. That `-1` is really killing me.

test/test_kmod/hello_mod/Makefile
```diff
@@ -0,0 +1,7 @@
+obj-m += hello-1.o
```
Susurrus: Also make this `hello.o`

test/test.rs
```diff
@@ -10,8 +10,25 @@ extern crate rand;
...
+            let mut handle = stderr.lock();
+            writeln!(handle, "{} requires root privileges. Skipping test.", $name).unwrap();
+            return;
+        }
+    };
+}
+
 mod sys;
 mod test_fcntl;
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize the OSes

src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub mod fcntl;
...
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize here.

src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub mod fcntl;
           target_os = "openbsd"))]
 pub mod ifaddrs;
 #[cfg(any(target_os = "linux", target_os = "android"))]
```
Susurrus: Please alphabetize this.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::os::unix::io::AsRawFd;
+
+use errno::Errno;
+use Result;
+
+/// Loads an ELF image into kernel space, performs any necessary symbol relocations,
+/// initializes module parameters to values provided by the caller, and then runs the module's init function.
+/// This function requires privilege.
+///
+/// The module_image argument points to a buffer containing the binary image to be loaded.
```
Susurrus: Backticks around `module_image`

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+/// The module image should be a valid ELF image, built for the running kernel.
```
Susurrus: Remove the comma.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// The param_values argument is a string containing space-delimited specifications of the values for module parameters.
```
Susurrus: Backticks around variable names.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
+//! Load and unload kernel modules.
+
+use libc;
+use std::ffi::CStr;
+use std::os::unix::io::AsRawFd;
+
+use errno::Errno;
+use Result;
+
+/// Loads an ELF image into kernel space, performs any necessary symbol relocations,
```
Susurrus: Please use a short (60 char or less) title sentence, followed by a blank comment line, then the rest of the text. See how other functions are commented for examples.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+/// initializes module parameters to values provided by the caller, and then runs the module's init function.
+/// This function requires privilege.
```
Susurrus: What privilege? Probably root privileges is correct here, yes? Also, can you separate this into a separate paragraph by putting a blank line above it or integrate it into the existing one?

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// The module_image argument points to a buffer containing the binary image to be loaded.
+/// The module image should be a valid ELF image, built for the running kernel.
+///
+/// The param_values argument is a string containing space-delimited specifications of the values for module parameters.
+/// Each of the parameter specifications has the form:
```
Susurrus: Could use provide an example that actually passes parameters? This format specifier is technically correct but can be hard to understand. Providing an example would be helpful.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// `name[=value[,value...]]`
+///
+/// Example usage:
```
Susurrus: Please follow the format of `# Example` rather than this text. Other comments follow a similar pattern.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// ```no_run
+/// use std::fs::File;
+/// use std::ffi::CString;
+/// use nix::kmod::{finit_module, ModuleInitFlags};
+///
+/// let f = File::open("mymod.ko").unwrap();
+/// finit_module(&f, &CString::new("").unwrap(), ModuleInitFlags::empty()).unwrap();
+/// ```
+///
```
Susurrus: Remove this blank line.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+        /// Ignore symbol version hashes.
+        MODULE_INIT_IGNORE_MODVERSIONS;
+        /// Ignore kernel version magic.
+        MODULE_INIT_IGNORE_VERMAGIC;
+    }
+);
+
+/// Loads a kernel module from a given file descriptor.
+///
+/// Example usage:
```
Susurrus: Same as the above, use `# Example` instead.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+        )
+    };
+
+    Errno::result(res).map(drop)
+}
+
+libc_bitflags!(
+    /// Flags used by `delete_module`.
+    pub struct OFlags: libc::c_int {
+        O_NONBLOCK;
```
Susurrus: Please add doccomments for these items.

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+            flags.bits(),
+        )
+    };
+
+    Errno::result(res).map(drop)
+}
+
+libc_bitflags!(
+    /// Flags used by `delete_module`.
+    pub struct OFlags: libc::c_int {
```
Susurrus: Can we rename this to `OutputFlags` or something less short-hand?

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+        O_NONBLOCK;
+        O_TRUNC;
+    }
+);
+
+/// Unloads the kernel module with the given name.
+///
+/// Example usage:
```
Susurrus: `# Example`

src/kmod.rs
```diff
@@ -0,0 +1,108 @@
...
+
+libc_bitflags!(
+    /// Flags used by `delete_module`.
+    pub struct OFlags: libc::c_int {
+        O_NONBLOCK;
+        O_TRUNC;
+    }
+);
+
+/// Unloads the kernel module with the given name.
```
Susurrus: Does "Unloads the given module from the currently-running kernel" read better here?

test/test_kmod/hello_mod/hello-1.c
bachp: A pure MIT licensed module would taint the kernel if loaded. The option I see would be: [`Dual MIT/GPL`](https://elixir.bootlin.com/linux/latest/source/include/linux/module.h#L180). I willl create a new module and license it accordingly.

src/kmod.rs
bachp: This is taken from the original documentation (https://linux.die.net/man/2/delete_module) I think explicitly state the given name is more accurate.

bachp:
@Susurrus Kernel module rewritten, it is now dual licensed MIT or GPL-2.0+ and includes parameters. These are also used in tests.

@asomers I didn't find a way how to build external kernel modules with a separate build directory.

asomers:
Did you try simply cd'ing to ${TMPDIR:-/tmp} before running make?

bachp:
@asomers I don't think switching the workdir before doing make would work as the workdir needs to be in the same directory. One option would be to copy everything to a TMP location, do the build and then delete it again after the test. But it seems hacky.

asomers:
@bachp could you try that then?  No test should be creating files in the source directory.  I can't even run these tests, because my source directory is mounted over NFS with rootsquash on.

bachp:
@asomers I bumped the minimum required rust version to 1.24.1 as this is the version packaged in Debian. So it should allow them to still package rust tools depending on nix.

I think the freebsd jobs should be updated to this version too.

src/kmod.rs
bachp: Done

asomers:
Why 1.24.1?  If Nix only depends on 1.22.0 features, then the minimum rustc version should be 1.22.0.  That shouldn't be a problem for building Debian packages.  Nor should Nix depend in any way on Debian.

bachp:
@asomers Rust 1.22 produces some compile errors that are not present in 1.20 and are no longer present in 1.24.1. See: https://travis-ci.org/nix-rust/nix/builds/413801114 for details on the error.

asomers:
That looks like a bug in Nix.  `ptr` may be a mutable reference.  But since it's passed by value to `encode_into`, it itself does not need to be declared mutable.  Try removing the `mut` and rebuilding on 1.22.0.

bachp:
But why does this warning only pop up with 1.22 and not with 1.24.1 or 1.28?

bachp:
I updated the MR to build with 1.22.1 and changed the offending code. But travis seems to have failed for some other reason now. Seems unrelated to the change.

asomers:
The Travis failures were spurious.  Restarting the jobs fixed them.  I'm upgrading buildbot now.

bachp:
Anything else that needs to be fixed?

CHANGELOG.md
```diff
@@ -12,8 +12,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#921](https://github.com/nix-rust/nix/pull/921))
 - Added support for `SCM_CREDENTIALS`, allowing to send process credentials over Unix sockets.
   ([#923](https://github.com/nix-rust/nix/pull/923))
+- Added `kmod` module that allows loading and unloading kernel modules on Linux.
+  ([#930](https://github.com/nix-rust/nix/pull/930))
 
 ### Changed
+- Bump minimum tested Rust version to 1.22.1.
```
Susurrus: "tested" -> "required"

Susurrus:
There are a bunch of changes unrelated to adding kmod, like the tempfile update and the various style changes. I'd prefer to merge #900 and then rebase and squash this merge into 2 commits:
 1) updating the Rust requirements to 1.22.1
 2) Adding `kmod`

We'll give #900 a couple of days.

bachp:
@Susurrus I think the 1.22.1 update should go into #900 as it is related to `tempfile`.

CHANGELOG.md
bachp: This commit/change will be removed once #900 is merged

bachp:
Rebased after #900 got merged

Cargo.toml
```diff
@@ -21,13 +21,13 @@ void = "1.0.2"
...
 [dev-dependencies]
-bytes = "0.4.8"
+bytes = "0.4"
 lazy_static = "1"
 rand = "0.5"
 tempfile = "3"
 
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
-sysctl = "0.1"
+sysctl = "0.2"
```
asomers: Why did you update sysctl and bytes?  What was wrong with the old versions?

bachp: This PR is open so long I don't remember anymore.

On the other side what is wrong with the new versions?

asomers: It's unnecessarily restrictive.  It might conflict with a crate that depends on nix and also has a restrictive requirement like `sysctl = "0.1.0"` or `sysctl < "0.2"`.

bachp: Isn't the same true the other way around? If I use `bytes = 0.2` I would still get a copy of `0.1`  because of nix?

asomers: Nope.  The dependency resolution is a little unintuitive.  Specifying "0.1" really means anything in the half-open range [0.1 - 1.0).  So version 0.2.0 would match.  If you want exactly 0.1.0, then you need to say "=0.1.0".  If you want anything in the range [0.1 - 2.0), you would specify "0.1.0".

bachp: Are you sure this is true for `0.x` too? According to the [documentation](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements):

> While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

But the example mentions:

> `^0.2 := >= 0.2.0 < 0.3.0`

Which seems kind of contradicting?

bachp: I removed the update commit. Expect for the `rand` crate.
It's not needed for this.

bachp:
I rebased again to fix the changelog conflict.

bachp:
@Susurrus PR reworked into 4 logical commits.

bachp:
Anything still missing?

Issue #931, opened 2018-07-28
-1 parameters to setresuid


sevagh:
Hello. The nix docs link to the Linux manpage for `setresuid`: http://man7.org/linux/man-pages/man2/setresuid.2.html

In Linux you can pass -1 as a parameter of setresuid. How can I do this in nix since the parameters are of type `u32`?

```
error[E0600]: cannot apply unary operator `-` to type `u32`
 --> src/main.rs:9:72
  |
9 |     match setresuid(Uid::from_raw(-1), Uid::from_raw(1), Uid::from_raw(-1)) {
  |                                                                        ^^ cannot apply unary operator `-`
```

asomers:
Uh, you can't.  Best would be if Nix's `setresuid` took an `Option<Uid>` instead.  Would you submit a PR that makes that change?

sevagh:
Sounds good, I'll work on that.

sevagh:
Is there a reason `getresuid` isn't implemented? I would like to take a shot at that as well if not.

asomers:
The only reason is because nobody's submitted it yet.

Susurrus:
Reopening this because it is still a bug with our implementation.

Susurrus:
Probably the solution here is to make it take `Option` for each argument and continue to not allow `-1`.

PR #932, opened 2018-07-29
*resuid improvements

sevagh:
This change allows one to pass an Option<Uid>, where None denotes no change. This should be the equivalent of passing a value of -1 in C setresuid.

To test that `setresuid(-1, ...)` is equivalent to `setresuid(UINT_MAX, ...)` (i.e. `setresuid(4294967295, ...)`), I did the following:

I tested this on Fedora 28, on an x86_64 system (Thinkpad T450) with `4.17.5-200` kernel. First, I changed `UID_MAX` in `/etc/login.defs` to `UINT_MAX` I wanted to confirm that the actual `UINT_MAX` value is reserved:

```
sevagh:nix $ grep '^UID_MAX' /etc/login.defs
UID_MAX                  4294967295
sevagh:nix $ useradd fakeuser -u 4294967295
useradd: invalid user ID '4294967295'
sevagh:nix $ useradd fakeuser -u 4294967294
useradd: cannot open /etc/passwd
useradd: failed to unlock /etc/passwd
```

Also wrote up a little C file to make sure:

```
#define _GNU_SOURCE
#include <unistd.h>
#include <stdio.h>
#include <limits.h>

void main() {
    printf("Hello world\n");

    uid_t x = -1;
    uid_t y = UINT_MAX;
    if (x == y) {
        printf("(uid_t x = -1) == (uid_t UINT_MAX)\n");
    } else {
        printf("(uid_t x = -1) != (uid_t UINT_MAX)\n");
    }
    printf("printf uid_t -1 with %%u: %u\n", x);
    printf("printf UINT_MAX with %%u: %u\n", UINT_MAX);
    static uid_t euid, ruid, suid;
    setresuid(-1, -1, -1);
    getresuid(&ruid, &euid, &suid);
    printf("Result of setresuid(-1, -1, -1):\n\t%u %u %u\n", euid, ruid, suid);
    setresuid(UINT_MAX, UINT_MAX, UINT_MAX);
    getresuid(&ruid, &euid, &suid);
    printf("Result of setresuid(UINT_MAX, UINT_MAX, UINT_MAX):\n\t%u %u %u\n", euid, ruid, suid);
}
```
Output:
```
(uid_t x = -1) == (uid_t UINT_MAX)
printf uid_t -1 with %u: 4294967295
printf UINT_MAX with %u: 4294967295
Result of setresuid(-1, -1, -1):
        1000 1000 1000
Result of setresuid(UINT_MAX, UINT_MAX, UINT_MAX):
        1000 1000 1000
```

I'm out of my element w.r.t. platform-specific behavior and whether this cause problems on different architecture, so I'd appreciate help/feedback in that regard.

Solves https://github.com/nix-rust/nix/issues/931

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
-- Change `setresuid` and `setresgid` functions to accept Option<Uid> instead of Uid
+- Add `getresuid` and `getresgid` functions
   ([#932](https://github.com/nix-rust/nix/pull/932))
 - Added a `sysinfo` wrapper.
   ([#922](https://github.com/nix-rust/nix/pull/922))
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#921](https://github.com/nix-rust/nix/pull/921))
 
 ### Changed
+- Change `setresuid` and `setresgid` functions to accept Option<Uid> instead of Uid
+  ([#932](https://github.com/nix-rust/nix/pull/932))
 
 ### Fixed
 - Made `preadv` take immutable slice of IoVec.


src/unistd.rs
```diff
@@ -2112,10 +2112,12 @@ mod setres {
...
     /// * returns: Ok or libc error code.
     ///
+    /// Use None to denote an unchanged value (equivalent to passing -1 in C's setresuid)
+    ///
     /// Err is returned if the user doesn't have permission to set this UID.
     #[inline]
-    pub fn setresuid(ruid: Uid, euid: Uid, suid: Uid) -> Result<()> {
-        let res = unsafe { libc::setresuid(ruid.into(), euid.into(), suid.into()) };
+    pub fn setresuid(ruid: Option<Uid>, euid: Option<Uid>, suid: Option<Uid>) -> Result<()> {
+        let res = unsafe { libc::setresuid(ruid.unwrap_or(Uid::from_raw(<u32>::max_value())).into(), euid.unwrap_or(Uid::from_raw(<u32>::max_value())).into(), suid.unwrap_or(Uid::from_raw(<u32>::max_value())).into()) };
```
asomers: 80 columns per line, please.

sevagh:
Is this better?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Change `setresuid` and `setresgid` functions to accept Option<Uid> instead of Uid
```
asomers: Because it's a breaking change, this should go into the Changed section, not Added.

src/unistd.rs
```diff
@@ -2112,10 +2112,17 @@ mod setres {
     /// * `suid`: saved user id
     /// * returns: Ok or libc error code.
     ///
+    /// Pass None to denote no change
```
asomers: Better if you can format the comment blocks in the more standard way, with Parameters and Returns sections, like `fpathconf`` does.

sevagh:
>Better, but not there yet. Also, have you figured out how to write a test?

Does it sound like a bad idea to combine my `getresuid/getresgid` PR here and have a combined test like:

```
orig_values = getresuid()
setresuid(None, None, None)
new_values = getresuid()
assert_eq(orig_values, new_values)
```
To me this seems like a safe way to test both functions without making assumptions on the running user's `uid`/etc.

sevagh:
I generalized the PR to "*resuid improvements" and closed the other one. I will work on the docs.

Susurrus:
I'm wondering if we can't do a better API than this, but I'm not certain how this API is used in practice. Is it actually common to actually work with all 3 uid/gid value? Or is it more common to change just 1 or 2 of them? I'm thinking that breaking up these functions to separate getters and setters might be more ergonomic.

The main reason I'm thinking about this is because I don't like the signature of `getres*id()`. C returns things through pointers, but in Rust it's preferably to return things by value. But if you do that, you have a 3-tuple. Which I'm not a big fan of either. It almost seems like were should have some methods on a `Process` struct where you can get and set these values as they apply to the calling process. I'm not certain it makes sense to go that far, but I'm sure other APIs could work within that framework.

sevagh:
I definitely agree that this is painful to type:

```rust
let mut rgid2 = Gid::from_raw(0);
let mut egid2 = Gid::from_raw(0);
let mut sgid2 = Gid::from_raw(0);

let _r3 = getresgid(&mut rgid2, &mut egid2, &mut sgid2);
```
I don't know if the nix-rust philosophy is to replicate C function signatures, or that if things deserve to be changed then they should be changed - I'll leave it up to the maintainers to decide.

Susurrus:
It will need to be changed to returning a 3-tuple at a minimum. I'm just wondering if going a little bit further with the abstraction would be worth it. @asomers any opinion here?

sevagh:
The documentation strings for `setresuid` and `setresgid` have been incorrect since 2016  (https://github.com/nix-rust/nix/commit/6c85f437c1df07a27e9e60c25c1322e9effc3a04); my corrections being scrutinized is interesting. I'd rather just drop my doc commits in that case.

asomers:
I think a 3-tuple would be ok, or it would be ok to return a `struct RESuid` with `ruid`, `euid`, and `suid` getters.  I don't think we should return a `Process` that has more interfaces than just `setresuid` and `setresgid`.  I like to keep Nix as close as possible to libc, while still being Rusty.

sevagh:
I'll work on that.

sevagh:
I think in practise I am preferring the 3-tuple. Introducing a `struct RESuid` somehow compels me to include that into `setres*id` which ends up being ugly with the `Option<Uid>`.

sevagh:
On second thought, since you two want to talk past me and to each other, go ahead and implement it yourselves.

PR #933, opened 2018-07-29
Add getresuid and getresgid functions

sevagh:
Also clean up inconsistencies in setresuid/setresgid doc comments.


sevagh:
I wonder if `assert_eq!(ruid, 1000)` is a bad idea - I don't really know the ruid/suid/euid of the user running the tests.

PR #936, opened 2018-08-17
Add the ptrace GET/SETREGS request methods

jabedude:


Susurrus:
Are these on all platforms? It looks like there are compilation failures on some platforms because various constants or structs are missing from `libc`. Could you submit a PR to `libc` adding them for the relevant platforms?

jabedude:
It looks like the PTRACE_GETREGS/SETREGS requests are not on all platforms. From `man 2 ptrace`
`PTRACE_GETREGS and PTRACE_GETFPREGS are not present on all architectures.`
Ditto for PTRACE_SETREGS.

Would it be best to use a `#[cfg]` here?

Susurrus:
Yes, you'll need to use `#[cfg]` to restrict this to certain platforms. Note, however, that you'll want to look at the platforms sources to determine which ones support it, not `libc`. You may still need to submit a PR to `libc` adding missing functionality.

xd009642:
So I don't know if this will be any help but some of the missing ptrace commands seem to just be missing from apples woefully incomplete implementation. (You can see their header [here](https://opensource.apple.com/source/xnu/xnu-792.13.8/bsd/sys/ptrace.h)). 

I've started work on apple support for #947 starting in libc before I start work in nix. But it looks like there may be some overlap in the updates for both PRs so I don't know if you want to do it or I could do the cfgs as part of my work and pull in my changes when they're merged. (Depending on when I get to it of course)

PR #940, opened 2018-09-10
Document ioctl_param_type

rukai:
Currently the docs claim ioctl_param_type to always be libc::c_int which is not the case on linux.

Also removes an unused import that was triggering a warning.

### DIFF
--- a/src/sys/ioctl/mod.rs
+++ b/src/sys/ioctl/mod.rs
@@ -221,7 +221,6 @@
 //!
 //! # fn main() {}
 //! ```
-//!
 #[cfg(any(target_os = "android", target_os = "linux"))]
 #[macro_use]
 mod linux;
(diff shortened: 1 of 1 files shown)


rukai:
Hi, this is a pretty simple fix, can someone please review it?

asomers:
@Susurrus you're most familiar with the ioctl stuff.  Want to give this a lookover?

Susurrus:
I'm wondering, do we want to have our documentation like this or do we just want to use the proper value on the given platform? I'm wondering how often this will be used for cross-platform stuff and whether every platform should know about all the possibilities or just the current platform needs to know their data type. Any opinion here?

rukai:
A case where that would be an issue is when you view the docs on a service like docs.rs. Then the type would just be whatever the docs were compiled on, presumably linux.

rukai:
Uh, you can squash when you merge the PR right?

asomers:
No we can't, because we use bors instead of the merge button.

rukai:
Oh, my bad, it's squashed now. Thanks for taking the time to review it.

asomers:
Looks like the build failure was probably Travis's fault.

bors retry

PR #942, opened 2018-09-17
Add wrapper for chmod

jmmv:
As the subject says, add a trivial wrapper for the chmod system call.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -557,6 +557,18 @@ pub fn getcwd() -> Result<PathBuf> {
     }
 }
 
+/// Change the mode of the file at `path` to have the permissions specified by
+/// `perm` (see
+/// [chmod(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html)).
+#[inline]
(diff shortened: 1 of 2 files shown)


asomers:
Welcome to Rust, Julio!  One question about this PR: can you give an example of when this function would be necessary as opposed to using [`std::os::unix::PermissionsExt::set_mode`](https://doc.rust-lang.org/std/os/unix/fs/trait.PermissionsExt.html#tymethod.set_mode) ?

jmmv:
Hey! I know your name :)

Uh... well, never? Hadn't seen that yet. So I guess this is useless :-P

jmmv:
Hold on... the `set_mode` you quote modifies a `Permissions` in-memory object... but how do you later apply those permissions to the actual file?

asomers:
@jmmv see [`std::fs::set_permissions`](https://doc.rust-lang.org/std/fs/fn.set_permissions.html)

jmmv:
Oh nice; thanks a lot. See https://github.com/jmmv/sandboxfs/commit/2dc36620fb2e90a84eade4deeb14bdc0c80e2898 for context.

PR #944, opened 2018-09-24
Add wrappers for futimens(2) and utimesat(2)

jmmv:
Hello again!

Let's see how this looks like. I don't think there is any other way of doing this via the standard library, though I see there is a `filetime` crate in the Rust nursery to add this missing functionality. Not sure what your policy for adding features into `nix` is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#916](https://github.com/nix-rust/nix/pull/916))
 - Added `kmod` module that allows loading and unloading kernel modules on Linux.
   ([#930](https://github.com/nix-rust/nix/pull/930))
+- Added `futimens` and `utimesat` wrappers.
+  ([#944](https://github.com/nix-rust/nix/pull/944))
 
 ### Changed
 - Increased required Rust version to 1.22.1/
--- a/src/sys/stat.rs
+++ b/src/sys/stat.rs
@@ -6,7 +6,9 @@ use errno::Errno;
 use fcntl::AtFlags;
 use libc::{self, mode_t};
 use std::mem;
+use std::os::raw;
 use std::os::unix::io::RawFd;
+use sys::time::TimeSpec;
 
 libc_bitflags!(
     pub struct SFlag: mode_t {
@@ -133,6 +135,15 @@ pub fn fchmod(fd: RawFd, mode: Mode) -> Result<()> {
     Errno::result(res).map(|_| ())
 }
 
+/// Computes the raw fd consumed by a function of the form `*at`.
+#[inline]
+fn actual_atfd(fd: Option<RawFd>) -> raw::c_int {
+    match fd {
+        None => libc::AT_FDCWD,
+        Some(fd) => fd,
+    }
+}
+
 /// Flags for `fchmodat` function.
 #[derive(Clone, Copy, Debug)]
 pub enum FchmodatFlags {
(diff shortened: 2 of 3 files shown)


asomers:
utimes(2) is obsolete.  futimens and utimensat are the newer replacements.  So you should be using one of those instead.  Or are you working on an OS that doesn't have them?

jmmv:
Why is `utimes` obsolete? I don't see it documented as such in the manual page (of macOS).

But I can change this to implement the other two calls if you prefer; I'm not bound to using `utimes` at the moment. Is there anything else major to change?

jmmv:
Grr, except that `libc` lacks both `utimensat` and `futimens` support for macOS. I guess we should fix that first...

jmmv:
Sent PR for libc in https://github.com/rust-lang/libc/pull/1084.

asomers:
Sigh.  It looks like lib needs a PR for FreeBSD, too.

src/sys/stat.rs
```diff
@@ -183,3 +189,64 @@ pub fn fchmodat<P: ?Sized + NixPath>(
...
+///
+/// The file to be changed is determined relative to the directory associated
+/// with the file descriptor `dirfd` or the current working directory
+/// if `dirfd` is `None`.
+///
+/// If `flag` is `UtimensatFlags::NoFollowSymlink` and `path` names a symbolic link,
+/// then the mode of the symbolic link is changed.
+///
+/// `utimensat(None, path, times, UtimensatFlags::FollowSymlink)` is identical to
+/// a call `libc::utimes(path, times)`. That's why `utimes` is unimplemented
```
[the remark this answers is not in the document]
jmmv: Done. But note that this was a copy/paste from `fchmodat` above, which suffers from the same problem.

asomers:
Thanks.  But one more thing: please add an entry in CHANGELOG.

jmmv:
CHANGELOG entry added and branch rebased on master to avoid conflicts.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#916](https://github.com/nix-rust/nix/pull/916))
 - Added `kmod` module that allows loading and unloading kernel modules on Linux.
   ([#930](https://github.com/nix-rust/nix/pull/930))
+- Added `futimens` and `utimesat` wrappers.
```
asomers: Please link to the PR in the changelog entry.

jmmv: Not sure why I missed that. Done.

jmmv:
Ugh, I wasn't expecting the original message in the PR review thread would appear in the commit message :-( I guess I know for next time.

jmmv:
And more: I just found out the hard way that `futimens` and `utimesat` are not available on macOS 10.12 or below. Not that I care much about supporting that old version, but I have to use this version on Travis as it's the only one that allows loading the FUSE kext.

What's the policy for `nix` regarding OS supported versions?

asomers:
Nix doesn't attempt to guarantee consistent features across different OSes and versions, since its objective is simply to expose OS functionality.  If a Nix function returns ENOSYS on older OSes, that's fine.

jmmv:
But then, would it be OK to *also* implement `utimes`? This deprecated function is going to exist pretty much everywhere, and for my specific use case, it's sufficient. Otherwise, in those systems where these functions return `ENOSYS`, we need to fall back to raw libc calls...

asomers:
Yeah, I think that's ok.

PR #948, opened 2018-10-06
socket: add AF_UNSPEC to AddressFamily

levex:
Hi!

I work on a [crate](https://github.com/levex/network-bridge-rs) where it would make sense to use `AF_UNSPEC` in place of `AF_INET`, since that's the API most bridge libraries use. Unfortunately, it seems that `nix` is missing a representation for `AF_UNSPEC`. This PR adds `Unspec` to `AddressFamily`, so we can now represent `AF_UNSPEC`.


asomers:
Sure, this looks fine.  Just add an entry to CHANGELOG and I'll merge it.

levex:
@asomers great! I've added an entry to the CHANGELOG.

PR #949, opened 2018-10-07
ptrace support for BSDs

### LINKED ISSUE #947: ptrace mac support
It's came up as part of https://github.com/xd009642/tarpaulin/issues/152 but ptrace is partially supported in macos ([source](http://uninformed.org/index.cgi?v=4&a=3&p=14)). I'll look into it myself this weekend but I think it would be useful to go through the ptrace api and cfg in the parts supported on Macs.

Any thoughts, opinions, objections?

xd009642:
This PR adds support to the ptrace API for BSDs to close #947. It also adds a read and write method for reading and writing to a traced processes memory. The ptrace API created for linux offers this via a deprecated function so I added this so they can be feature equivalent without replicating a deprecated part of the API.

Due to the differences in ptrace on BSD and linux I've made a ptrace module to keep things readable.

Still to do - revert travis config to remove my feature branch and update the changelog.

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -38,7 +38,13 @@ pub mod mman;
 
 pub mod pthread;
 
-#[cfg(any(target_os = "android", target_os = "linux"))]
+#[cfg(any(target_os = "android",
+          target_os = "dragonfly",
+          target_os = "freebsd",
+          target_os = "linux",
+          target_os = "macos",
+          target_os = "netbsd",
+          target_os = "openbsd"))]
 pub mod ptrace;
 
 #[cfg(target_os = "linux")]
--- a/src/sys/ptrace/bsd.rs
+++ b/src/sys/ptrace/bsd.rs
@@ -1,0 +1,170 @@
+use errno::Errno;
+use libc::{self, c_int};
+use std::ptr;
+use sys::signal::Signal;
+use unistd::Pid;
+use Result;
+
+pub type RequestType = c_int;
+
+cfg_if! {
+    if #[cfg(any(target_os = "dragonfly", 
+                 target_os = "freebsd", 
+                 target_os = "macos",
+                 target_os = "openbsd"))] {
+        #[doc(hidden)]
+        pub type AddressType = *mut ::libc::c_char;
+    } else {
+        #[doc(hidden)]
+        pub type AddressType = *mut ::libc::c_void;
+    }
+}
+
+libc_enum! {
+    #[repr(i32)]
+    /// Ptrace Request enum defining the action to be taken.
+    pub enum Request {
+        PT_TRACE_ME,
+        PT_READ_I,
+        PT_READ_D,
+        #[cfg(target_os = "macos")]
+        PT_READ_U,
+        PT_WRITE_I,
+        PT_WRITE_D,
+        #[cfg(target_os = "macos")]
+        PT_WRITE_U,
+        PT_CONTINUE,
+        PT_KILL,
+        #[cfg(any(any(target_os = "dragonfly",
+                  target_os = "freebsd",
+                  target_os = "macos"),
+                  all(target_os = "openbsd", target_arch = "x86_64"),
+                  all(target_os = "netbsd", any(target_arch = "x86_64",
+                                                target_arch = "powerpc"))))]
+        PT_STEP,
+        PT_ATTACH,
+        PT_DETACH,
+        #[cfg(target_os = "macos")]
+        PT_SIGEXC,
+        #[cfg(target_os = "macos")]
+        PT_THUPDATE,
+        #[cfg(target_os = "macos")]
+        PT_ATTACHEXC
+    }
+}
+
+unsafe fn ptrace_other(
+    request: Request,
+    pid: Pid,
+    addr: AddressType,
+    data: c_int,
+) -> Result<c_int> {
+    Errno::result(libc::ptrace(
+        request as RequestType,
+        libc::pid_t::from(pid),
+        addr,
+        data,
+    )).map(|_| 0)
+}
+
+/// Sets the process as traceable, as with `ptrace(PT_TRACEME, ...)`
+///
+/// Indicates that this process is to be traced by its parent.
+/// This is the only ptrace request to be issued by the tracee.
+pub fn traceme() -> Result<()> {
+    unsafe { ptrace_other(Request::PT_TRACE_ME, Pid::from_raw(0), ptr::null_mut(), 0).map(|_| ()) }
+}
+
+/// Attach to a running process, as with `ptrace(PT_ATTACH, ...)`
+///
+/// Attaches to the process specified in pid, making it a tracee of the calling process.
+pub fn attach(pid: Pid) -> Result<()> {
+    unsafe { ptrace_other(Request::PT_ATTACH, pid, ptr::null_mut(), 0).map(|_| ()) }
+}
+
+/// Detaches the current running process, as with `ptrace(PT_DETACH, ...)`
+///
+/// Detaches from the process specified in pid allowing it to run freely
+pub fn detach(pid: Pid) -> Result<()> {
+    unsafe { ptrace_other(Request::PT_DETACH, pid, ptr::null_mut(), 0).map(|_| ()) }
+}
+
+/// Restart the stopped tracee process, as with `ptrace(PTRACE_CONT, ...)`
+///
+/// Continues the execution of the process with PID `pid`, optionally
+/// delivering a signal specified by `sig`.
+pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
+    let data = match sig.into() {
+        Some(s) => s as c_int,
+        None => 0,
+    };
+    unsafe {
+        // Ignore the useless return value
+        ptrace_other(Request::PT_CONTINUE, pid, 1 as AddressType, data).map(|_| ())
+    }
+}
+
+/// Issues a kill request as with `ptrace(PT_KILL, ...)`
+///
+/// This request is equivalent to `ptrace(PT_CONTINUE, ..., SIGKILL);` 
+pub fn kill(pid: Pid) -> Result<()> {
+    unsafe {
+        ptrace_other(Request::PT_KILL, pid, 0 as AddressType, 0).map(|_| ())
+    }
+}
+
+/// Move the stopped tracee process forward by a single step as with
+/// `ptrace(PT_STEP, ...)`
+///
+/// Advances the execution of the process with PID `pid` by a single step optionally delivering a
+/// signal specified by `sig`.
+///
+/// # Example
+/// ```rust
+/// extern crate nix;
+/// use nix::sys::ptrace::step;
+/// use nix::unistd::Pid;
+/// use nix::sys::signal::Signal;
+/// use nix::sys::wait::*;
+/// fn main() {
+///     // If a process changes state to the stopped state because of a SIGUSR1
+///     // signal, this will step the process forward and forward the user
+///     // signal to the stopped process
+///     match waitpid(Pid::from_raw(-1), None) {
+///         Ok(WaitStatus::Stopped(pid, Signal::SIGUSR1)) => {
+///             let _ = step(pid, Signal::SIGUSR1);
+///         }
+///         _ => {},
+///     }
+/// }
+/// ```
+#[cfg(
+    any(
+        any(target_os = "dragonfly", target_os = "freebsd", target_os = "macos"),
+        all(target_os = "openbsd", target_arch = "x86_64"),
+        all(target_os = "netbsd",
+            any(target_arch = "x86_64", target_arch = "powerpc")
+        )
+    )
+)]
+pub fn step<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
+    let data = match sig.into() {
+        Some(s) => s as c_int,
+        None => 0,
+    };
+    unsafe { ptrace_other(Request::PT_STEP, pid, ptr::null_mut(), data).map(|_| ()) }
+}
+
+/// Reads a word from a processes memory at the given address
+pub fn read(pid: Pid, addr: AddressType) -> Result<c_int> {
+    unsafe {
+        // Traditionally there was a difference between reading data or
+        // instruction memory but not in modern systems.
+        ptrace_other(Request::PT_READ_D, pid, addr, 0)
+    }
+}
+
+/// Writes a word into the processes memory at the given address
+pub fn write(pid: Pid, addr: AddressType, data: c_int) -> Result<()> {
+    unsafe { ptrace_other(Request::PT_WRITE_D, pid, addr, data).map(|_| ()) }
+}
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -173,8 +173,10 @@ libc_bitflags! {
 pub unsafe fn ptrace(request: Request, pid: Pid, addr: *mut c_void, data: *mut c_void) -> Result<c_long> {
     use self::Request::*;
     match request {
-        PTRACE_PEEKTEXT | PTRACE_PEEKDATA | PTRACE_PEEKUSER => ptrace_peek(request, pid, addr, data),
-        PTRACE_GETSIGINFO | PTRACE_GETEVENTMSG | PTRACE_SETSIGINFO | PTRACE_SETOPTIONS => Err(Error::UnsupportedOperation),
+        PTRACE_PEEKTEXT | PTRACE_PEEKDATA | PTRACE_PEEKUSER | PTRACE_GETSIGINFO | 
+            PTRACE_GETEVENTMSG | PTRACE_SETSIGINFO | PTRACE_SETOPTIONS | 
+            PTRACE_POKETEXT | PTRACE_POKEDATA | PTRACE_POKEUSER | 
+            PTRACE_KILL => Err(Error::UnsupportedOperation),
         _ => ptrace_other(request, pid, addr, data)
     }
 }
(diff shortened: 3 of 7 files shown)


.travis.yml
```diff
@@ -141,6 +141,7 @@ branches:
     # bors-ng branches; see https://bors-ng.github.io/getting-started/
     - trying
     - staging
+    - issue-947
```
asomers: This line shouldn't be necessary.  Travis automatically builds PRs.  This line is only necessary if you want nix-rust/nix to have a branch named "issue-947".

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,171 @@
...
+///
+/// Continues the execution of the process with PID `pid`, optionally
+/// delivering a signal specified by `sig`.
+pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
+    let data = match sig.into() {
+        Some(s) => s as c_int,
+        None => 0,
+    };
+    unsafe {
+        ptrace_other(Request::PT_CONTINUE, pid, ptr::null_mut(), data).map(|_| ()) // ignore the useless return value
```
asomers: Please line wrap to 80 cols.

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,171 @@
...
+///     // signal to the stopped process
+///     match waitpid(Pid::from_raw(-1), None) {
+///         Ok(WaitStatus::Stopped(pid, Signal::SIGUSR1)) => {
+///             let _ = step(pid, Signal::SIGUSR1);
+///         }
+///         _ => {},
+///     }
+/// }
+/// ```
+#[cfg(not(any(target_os = "netbsd", target_os = "openbsd")))]
```
asomers: Why not enabled for these OSes?  They both define `PT_STEP`.

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,171 @@
...
+        None => 0,
+    };
+    unsafe {
+        ptrace_other(Request::PT_STEP, pid, ptr::null_mut(), data).map(|_| ())
+    }
+}
+
+/// Step isn't supported on netbsd or openbsd
+#[cfg(any(target_os = "netbsd", target_os = "openbsd"))]
+pub fn Step<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
```
asomers: You shouldn't define this function.  It's better to make programs fail at compile time than at runtime.  Also, the capitalization is wrong.

src/sys/ptrace/linux.rs
```diff
@@ -20,7 +20,7 @@ cfg_if! {
 }
 
 libc_enum!{
-    #[cfg_attr(not(any(target_env = "musl", target_os = "android")), repr(u32))]
+    #[cfg_attr(not(any(target_env = "musl", target_os = "android", target_os = "macos")), repr(u32))]
```
asomers: This change shouldn't be necessary, since the file is not included on OSX.

src/sys/ptrace/bsd.rs
xd009642: Do they really? I was going off https://github.com/openbsd/src/blob/master/sys/sys/ptrace.h and https://github.com/IIJ-NetBSD/netbsd-src/blob/master/sys/sys/ptrace.h#L38

xd009642:
Well that didn't quite go as expected. I'm unsure what setup the buildbot instances are (guessing freebsd though) and the apple ptrace calls just seem to be returning `ENOTSUP` As that's returned if you use things like `PT_DENY_ATTACH` in apple I'm thinking maybe it's blocking ptrace calls for security reasons. Though that is just a guess.

src/sys/ptrace/bsd.rs
asomers: They must be in a different header then.  I'm looking at https://man.openbsd.org/ptrace and http://netbsd.gw.com/cgi-bin/man-cgi?ptrace++NetBSD-current

test/sys/mod.rs
```diff
@@ -26,6 +26,5 @@ mod test_uio;
 #[cfg(target_os = "linux")]
 mod test_epoll;
 mod test_pthread;
-#[cfg(any(target_os = "android",
-          target_os = "linux"))]
+#[cfg(not(target_os = "ios"))]
```
asomers: This `#[cfg()]` statement is too broad.  Nix runs on many platforms.  You should opt-in, not opt-out.

src/sys/ptrace/bsd.rs
krytarowski: NetBSD supports PT_STEP in CPU specific headers, not all CPUs support it.

xd009642:
Okay well I've just opened a libc PR that adds PT_STEP for the relevant archs. So once that's merged I'll update accordingly.

Any idea about the `ENOTSUP` for the apple ptrace tests? Might be a qemu based issue? I noticed that in the test comments it says it returns `ENOSYS` when it can't ptrace, might it do `ENOTSUP` on apple?

krytarowski:
"libc PR that adds PT_STEP" where?

xd009642:
In libc https://github.com/rust-lang/libc/pull/1091#issuecomment-427952816 none of the ptrace requests were available in BSDs so I did a PR, did another one to add the arch specific requests today :smile:

asomers:
> Okay well I've just opened a libc PR that adds PT_STEP for the relevant archs. So once that's merged I'll update accordingly.
> 
> Any idea about the `ENOTSUP` for the apple ptrace tests? Might be a qemu based issue? I noticed that in the test comments it says it returns `ENOSYS` when it can't ptrace, might it do `ENOTSUP` on apple?

I don't know.  Have you tried it on real Apple hardware?  I don't have any.

xd009642:
I don't have any either, I'm trying to add OSX support for one of my projects for a user that asked for it. I'll ask them if they can run it

xd009642:
Well it doesn't seem to work on an actual apple pc as well. I think it might need additional fiddling via functions that aren't in libc but are in https://github.com/fitzgen/mach/ 

In the interest of moving this along I could remove apple support from bsd, fix the buildbot errors then work on apple support as an additional PR? Or try and do it all in this one.

How would you feel about adding a dependency on https://github.com/fitzgen/mach/ for apple users?

xd009642:
I might be onto something, I've got `test_ptrace_cont` passing on apple! So I'm going to assume there's now no need for `mach` as a crate

xd009642:
Okay I might be all good now! 

One potential issue with the tests I noticed there was a spurious test_wait failure, this might be caused by test_wait and test_ptrace_cont being ran at the same time by rustc-test if waitpid will respond to any signal from the same `PGID` on the system in question. I've had similar issues in my own project when trying to test things that use wait or ptrace.

Either way I'll check back tomorrow and see what travis/buildbot says

asomers:
Tsk Tsk.  You should always compile and test before pushing to Travis.  Travis shouldn't be finding your unclosed delimiters for you.

xd009642:
I know, it's this apple stuff has made me sloppy. I realised doing stuff in the bsd mode that cargo check and build didn't flag up syntax errors at all.

If I make a habit of OSX based PRs I'll try to figure out how to cross compile for OSX

xd009642:
So I know I can't just change `wait` to `waitpid` in `test_wait`, but I just wanted to prove my theory on the last failure was corrected. With the design of `test_wait` you're vulnerable to stealing `test_ptrace_cont` signals if `cargo test` is ran with multiple threads.

As a result I've made `test-threads=1` I know this will slow down CI runs, so there's also an alternative to make `test_wait` ignored then run it on it's own in a separate call to `cargo test` than the others.

test/sys/test_ptrace.rs
```diff
@@ -10,25 +12,29 @@ fn test_ptrace() {
     // Just make sure ptrace can be called at all, for now.
     // FIXME: qemu-user doesn't implement ptrace on all arches, so permit ENOSYS
     let err = ptrace::attach(getpid()).unwrap_err();
-    assert!(err == Error::Sys(Errno::EPERM) || err == Error::Sys(Errno::ENOSYS));
+    assert!(err == Error::Sys(Errno::EPERM) || err == Error::Sys(Errno::EINVAL) ||
```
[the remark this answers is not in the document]
xd009642: https://www.freebsd.org/cgi/man.cgi?query=ptrace the BSDs seem to use EINVAL for attempting to attach to yourself. I don't know if all BSDs do that or if some mimic linux and do EPERM so I just had it as that for now. I can investigate and make it more specific though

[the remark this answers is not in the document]
krytarowski: NetBSD returns EINVAL

xd009642:
@asomers `FORK_MTX` appears to have had no effect which seems super odd to me :confused:

krytarowski:
If you find anything wrong with NetBSD and ptrace(2), please let me know!

krytarowski:
BTW. you might find it useful http://netbsd.org/~kamil/ptrace-netbsd/presentation.html a little bit dated, but quite good starter.

asomers:
The compile failure on NetBSD is purely a problem in your PR.  You have an unused `use` statement on line 7.  Delete that and it should compile.  The runtime failure on OSX is more interesting.  It looks like the `test_ptrace_const` test is leaving behind a zombie process on OSX, even though it doesn't on Linux or FreeBSD.  It would be interesting to run the test on OSX with this patch, and see if there is any zombie in the output.  You'll have to run it with `--nocapture`, of course.

```
diff --git a/test/sys/test_ptrace.rs b/test/sys/test_ptrace.rs
index 68d7922a..73966b30 100644
--- a/test/sys/test_ptrace.rs
+++ b/test/sys/test_ptrace.rs
@@ -96,6 +96,9 @@ fn test_ptrace_cont() {
                 }
                 _ => panic!("The process should have been killed"),
             }
+            use std::process;
+            let output = process::Command::new("ps").arg("-axwwd").output().unwrap();
+            println!("{}", String::from_utf8_lossy(&output.stdout));
         },
     }
 }
```

xd009642:
Great, I figured as much.

I'll try things that come to mind as potential fixes, (like currently ptrace::detach) and also in parallel try to find some documentation or blogs that shed light on the differences between apple and the rest of the world

xd009642:
I've seemingly cracked it, this part of the linux man page for ptrace helped me realise what may be happening:

```
PTRACE_KILL
      Send  the  tracee a SIGKILL to terminate it.  (addr and data are
      ignored.)

      This operation is deprecated; do not use it!   Instead,  send  a
      SIGKILL  directly  using kill(2) or tgkill(2).  The problem with
      PTRACE_KILL is that it requires the  tracee  to  be  in  signal-
      delivery-stop,  otherwise  it  may  not work (i.e., may complete
      successfully but won't kill the tracee).  By contrast, sending a
      SIGKILL directly has no such limitation.
```

And then in the apple man page for ptrace

```
PT_KILL       The traced process terminates, as if PT_CONTINUE had been
              used with SIGKILL given as the signal to be delivered.
```

I know they're different OSes but if PT_KILL in apple is subject to the same potential pitfalls of linux that would explain the issues we've observed. Therefore after waitpid catches the kill we sent, just in case it's  a false positive kill I send a direct kill to the tracee. 

I've also added a comment to document this behaviour in the test to explain the rationale for future maintainers.

asomers:
> I've seemingly cracked it, this part of the linux man page for ptrace helped me realise what may be happening:
> 
> ```
> PTRACE_KILL
>       Send  the  tracee a SIGKILL to terminate it.  (addr and data are
>       ignored.)
> 
>       This operation is deprecated; do not use it!   Instead,  send  a
>       SIGKILL  directly  using kill(2) or tgkill(2).  The problem with
>       PTRACE_KILL is that it requires the  tracee  to  be  in  signal-
>       delivery-stop,  otherwise  it  may  not work (i.e., may complete
>       successfully but won't kill the tracee).  By contrast, sending a
>       SIGKILL directly has no such limitation.
> ```
> 
> And then in the apple man page for ptrace
> 
> ```
> PT_KILL       The traced process terminates, as if PT_CONTINUE had been
>               used with SIGKILL given as the signal to be delivered.
> ```
> 
> I know they're different OSes but if PT_KILL in apple is subject to the same potential pitfalls of linux that would explain the issues we've observed. Therefore after waitpid catches the kill we sent, just in case it's a false positive kill I send a direct kill to the tracee.
> 
> I've also added a comment to document this behaviour in the test to explain the rationale for future maintainers.

The Linux manpage surely doesn't apply.  ptrace(2) is very different in Linux than in the BSDs.  OSX takes after the BSDs, but adds its own differences to ptrace(2) due to Mach.  I'm not saying that your solution is wrong, but I don't accept the rationale.

krytarowski:
No idea about Darwin, but while PT_KILL is broken on Linux, it's perfectly fine on BSDs (at least NetBSD).

xd009642:
It might be tenuous but it is the only source that could explain this strange apple behaviour. I'm now going to see if `PT_KILL` works on apple  different to `PT_CONTINUE` + `SIGKILL`.

I've also gone over all the signal and tracing stuff in http://newosxbook.com/MOXiI.pdf and can find no justification. Apart from the fact ptrace is crippled on apple for security/obfuscation reasons, there are undocumented parts and it's generally a pain as a result.

xd009642:
Thinking of the kill, does it actually effect what the test is attempting to show. `test_ptrace_cont` wants to show that `PT_CONTINUE` works and also that you can use it to forward signals onto the tracee. This behaviour is still shown with the additional kill call in as we check the kill signal was received when sent via ptrace.

xd009642:
@asomers any thoughts/issues RE my last comment?

xd009642:
Yeah that's exactly the point I was making :smile: also rebased and squashed!

xd009642:
So I've just had to rebase this onto master again because of another changelog change, @Susurrus any comments before merging?

src/sys/mod.rs
```diff
@@ -38,7 +38,13 @@ pub mod mman;
 
 pub mod pthread;
 
-#[cfg(any(target_os = "android", target_os = "linux"))]
+#[cfg(any(target_os = "dragonfly",
```
Susurrus: These should be in alphabetical order.

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,167 @@
...
+use libc::{self, c_int};
+use std::ptr;
+use sys::signal::Signal;
+use unistd::Pid;
+use Result;
+
+pub type RequestType = c_int;
+
+cfg_if! {
+    if #[cfg(any(target_os = "macos",
```
Susurrus: Alphabetical order here as well.

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,167 @@
...
+        PT_READ_D,
+        #[cfg(target_os = "macos")]
+        PT_READ_U,
+        PT_WRITE_I,
+        PT_WRITE_D,
+        #[cfg(target_os = "macos")]
+        PT_WRITE_U,
+        PT_CONTINUE,
+        PT_KILL,
+        #[cfg(any(any(target_os = "macos",
```
Susurrus: Alphabetical order

test/sys/mod.rs
```diff
@@ -26,6 +26,11 @@ mod test_uio;
 #[cfg(target_os = "linux")]
 mod test_epoll;
 mod test_pthread;
-#[cfg(any(target_os = "android",
-          target_os = "linux"))]
+#[cfg(any(target_os = "dragonfly",
```
Susurrus: Alphabetical order here too.

test/sys/test_ptrace.rs
```diff
@@ -50,10 +57,12 @@ fn test_ptrace_setsiginfo() {
 fn test_ptrace_cont() {
     use nix::sys::ptrace;
     use nix::sys::signal::{raise, Signal};
-    use nix::sys::wait::{waitpid, WaitStatus};
+    use nix::sys::wait::{waitpid, WaitStatus, WaitPidFlag};
```
Susurrus: Alphabetical order

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,167 @@
...
+        None => 0,
+    };
+    unsafe {
+        // Ignore the useless return value
+        ptrace_other(Request::PT_CONTINUE, pid, 1 as AddressType, data).map(|_| ())
+    }
+}
+
+/// Issues a kill request as with ptrace(PT_KILL, ..-)
+/// This request is equivalent to ptrace(PT_CONTINUE, ..., SIGKILL);
```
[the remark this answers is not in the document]
xd009642: Fix it as in a blank line above each function? I'm not sure which bit looks weird in the generated docs. Is it the function summary? https://docs.rs/nix/0.11.0/nix/sys/ptrace/index.html

src/sys/ptrace/linux.rs
```diff
@@ -320,9 +322,9 @@ pub fn cont<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
     }
 }
 
-
-/// Issues a kill request as with ptrace(PTRACE_KILL, ..-)
-/// This request is equivalent to ptrace(PTRACE_CONT, ..., SIGKILL); 
+///
```
Susurrus: Sorry I wasn't clear, can you put this blank comment line between the two existing comment lines on 326 and 327? A blank line separates a brief summary line from a further longer description. Otherwise this shows as one really long line on the main ptrace doc page.

src/sys/ptrace/bsd.rs
```diff
@@ -0,0 +1,170 @@
...
+use libc::{self, c_int};
+use std::ptr;
+use sys::signal::Signal;
+use unistd::Pid;
+use Result;
+
+pub type RequestType = c_int;
+
+cfg_if! {
+    if #[cfg(any(target_os = "dragonfly",
```
Susurrus: Could you alphabetize this?

xd009642:
@Susurrus I think we're all good to go now :+1:

src/sys/mod.rs
xd009642: Fixed, squashed and pushed. Sorry about that!

PR #952, opened 2018-10-12
Add wrapper for acct(2)

jabedude:
This PR aims to add a nix wrapper for acct(2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added support for `ptrace` on BSD operating systems ([#949](https://github.com/nix-rust/nix/pull/949))
 - Added `ptrace` functions for reads and writes to tracee memory and ptrace kill 
   ([#949](https://github.com/nix-rust/nix/pull/949))
+- Added a `acct` wrapper module for enabling and disabling process accounting
+  ([#952](https://github.com/nix-rust/nix/pull/952))
 
 ### Changed
 - Increased required Rust version to 1.22.1/
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1508,6 +1508,31 @@ pub fn sleep(seconds: c_uint) -> c_uint {
     unsafe { libc::sleep(seconds) }
 }
 
+pub mod acct {
+    use libc;
+    use {Result, NixPath};
+    use errno::Errno;
+    use std::ptr;
+
+    /// Enable process accounting
+    ///
+    /// See also [acct(2)](https://linux.die.net/man/2/acct)
+    pub fn enable<P: ?Sized + NixPath>(filename: &P) -> Result<()> {
+        let res = try!(filename.with_nix_path(|cstr| {
+            unsafe { libc::acct(cstr.as_ptr()) }
+        }));
+
+        Errno::result(res).map(drop)
+    }
+
+    /// Disable process accounting
+    pub fn disable() -> Result<()> {
+        let res = unsafe { libc::acct(ptr::null()) };
+
+        Errno::result(res).map(drop)
+    }
+}
+
 /// Creates a regular file which persists even after process termination
 ///
 /// * `template`: a path whose 6 rightmost characters must be X, e.g. `/tmp/tmpfile_XXXXXX`
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -4,12 +4,13 @@ use nix::unistd::ForkResult::*;
 use nix::sys::signal::{SaFlags, SigAction, SigHandler, SigSet, Signal, sigaction};
 use nix::sys::wait::*;
 use nix::sys::stat::{self, Mode, SFlag};
-use std::{env, iter};
+use std::{env, iter, thread, time};
 use std::ffi::CString;
-use std::fs::{self, File};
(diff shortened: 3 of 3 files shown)


src/unistd.rs
```diff
@@ -1439,6 +1439,18 @@ pub fn sleep(seconds: c_uint) -> c_uint {
     unsafe { libc::sleep(seconds) }
 }
 
+/// Switch process accounting on or off
+///
+/// See also [acct(2)](https://linux.die.net/man/2/acct)
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn acct(filename: &CString) -> Result<Void> {
```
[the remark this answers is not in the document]
jabedude: Wonderful ideas. I will definitely work on those changes.

jabedude:
I am currently planning on using `lastcomm`, but I might make a separate crate for parsing pacct files.

jabedude:
I added acct(2) to the BSD tree of libc [See here.](https://github.com/rust-lang/libc/pull/1093). CI should pass now

src/unistd.rs
```diff
@@ -1439,6 +1439,32 @@ pub fn sleep(seconds: c_uint) -> c_uint {
     unsafe { libc::sleep(seconds) }
 }
 
+#[cfg(target_family = "unix")]
```
asomers: The `target_family` should be unnecessary.  Nix only runs on unix operating systems.

test/test_unistd.rs
```diff
@@ -335,6 +336,21 @@ fn test_lseek64() {
     close(tmpfd).unwrap();
 }
 
+#[cfg(target_family = "unix")]
+#[test]
+fn test_acct() {
+    skip_if_not_root!("test_acct");
```
asomers: The test is failing in CI because CI runs in a jail, and jails aren't allowed to control the system-wide process accounting.  You can check whether you're jailed by checking the `security.jail.jailed` sysctl.  However, since the test is not FreeBSD-specific, I'd accept simply disabling it on FreeBSD.  I've done the same thing in mio.

test/test_unistd.rs
```diff
@@ -335,6 +336,21 @@ fn test_lseek64() {
...
+    let file = NamedTempFile::new().unwrap();
+    let path = file.path().to_str().unwrap();
+
+    acct::enable(path).unwrap();
+    Command::new("echo").arg("Hello world");
+    acct::disable().unwrap();
+
+    let md = metadata(path).unwrap();
+    assert!(md.len() > 0);
```
asomers: This is a race.  acct is an asynchronous interface.  There's no guarantee that the accounting file is updated by the time that `echo` returns.

jabedude: Would a sleep be appropriate here?

asomers: No, you need a polling loop.

test/test_unistd.rs
asomers: Please add a comment indicating why this test is skipped on FreeBSD.  It isn't obvious.

test/test_unistd.rs
jabedude: Would [inotify](https://github.com/inotify-rs/inotify) be okay to use?

asomers: Just for a simple test case?  No, that's much too heavyweight.

jabedude: I used a simple loop, hopefully that works...

test/test_unistd.rs
```diff
@@ -390,8 +391,13 @@ fn test_acct() {
     Command::new("echo").arg("Hello world");
     acct::disable().unwrap();
 
-    let md = metadata(path).unwrap();
-    assert!(md.len() > 0);
+    let len;
+    loop {
+        len = metadata(path).unwrap().len();
+        if len > 0 { break; }
```
asomers: You should add a short sleep so we don't spin the CPU.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#956](https://github.com/nix-rust/nix/pull/956))
 - Added a `fchownat` wrapper.
   ([#955](https://github.com/nix-rust/nix/pull/955))
+- Added a `acct` wrapper module for enabling and disabling process accounting ([#952](https://github.com/nix-rust/nix/pull/952))
```
asomers: Wrap to 80 columns, please.

test/test_unistd.rs
```diff
@@ -378,6 +379,27 @@ fn test_lseek64() {
...
+    Command::new("echo").arg("Hello world");
+    acct::disable().unwrap();
+
+    let len;
+    loop {
+        len = metadata(path).unwrap().len();
+        if len > 0 { break; }
+    }
+
+    assert!(len > 0);
```
asomers: The loop makes this assertion guaranteed.  You may as well remove the assertion.

jabedude: Ah I guess I misunderstood the way #[test] blocks of code work. I thought that test cases required an assertion to pass `cargo test`

asomers: Nope.  They're really quite simple.  Anything that doesn't panic is considered a success.

jabedude:
Done!

PR #955, opened 2018-10-17
Add a fchownat(2) wrapper

jmmv:


jmmv:
Per the TODO, do you think `chown` should be removed?

The way I'd validate that is: add a trivial test for `fchownat` to ensure we reference this function somewhere and then ask CI to validate all platforms. If that succeeds, I can later try to see if this is sufficient for older macOS versions, which still exist in Travis, to prevent the issue we saw with `utimes`.

I can do this before the merge; just let me know if there are other considerations.

src/fcntl.rs
```diff
@@ -182,6 +183,15 @@ pub fn readlinkat<'a, P: ?Sized + NixPath>(dirfd: RawFd, path: &P, buffer: &'a m
     wrap_readlink_result(buffer, res)
 }
 
+/// Computes the raw fd consumed by a function of the form `*at`.
+#[inline]
+pub fn at_rawfd(fd: Option<RawFd>) -> raw::c_int {
```
asomers: Does this need to be public?  Will Nix consumers ever need to use it directly?

src/unistd.rs
```diff
@@ -557,27 +557,89 @@ pub fn getcwd() -> Result<PathBuf> {
     }
 }
 
+/// Computes the raw UID and GID values to pass to a `*chown` call.
+///
+/// According to the POSIX specification, -1 is used to indicate that owner and group
```
asomers: This part doesn't belong in the doc comment.  You can make it a regular comment instead.

src/unistd.rs
```diff
@@ -557,27 +557,89 @@ pub fn getcwd() -> Result<PathBuf> {
...
+/// are not to be changed.  Since uid_t and gid_t are unsigned types, we have to wrap
+/// around to get -1.
+#[inline]
```
asomers: `#[inline]` doesn't do anything for non-public functions, so you can remove it.

src/unistd.rs
```diff
@@ -557,27 +557,89 @@ pub fn getcwd() -> Result<PathBuf> {
...
+/// then the mode of the symbolic link is changed.
+///
+/// `fchownat(None, path, mode, FchownatFlags::NoFollowSymlink)` is identical to
+/// a call `libc::lchown(path, mode)`.  That's why `lchmod` is unimplemented in
+/// the `nix` crate.
+///
+/// # References
+///
+/// [fchownat(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fchownat.html).
+#[inline]
```
asomers: I favor not using `#[inline]` at all.  Application authors can already use LTO if they really care.

src/fcntl.rs
asomers: If you need to use it from within multiple of Nix's modules, then maybe you should be using `pub(crate)` or `pub(super)`.

jmmv: Oh, I wasn't aware of such a thing. Done.

jmmv:
I don't mind keeping `chown`, but then would the same rationale apply for `lchown` and we should add it? Why doesn't the same apply for other traditional functions like `utimes`?

asomers:
> I don't mind keeping `chown`, but then would the same rationale apply for `lchown` and we should add it? Why doesn't the same apply for other traditional functions like `utimes`?
> 
> In fact, I think that if the OS exposes such functions, they should be exposed in `nix` as well. `nix` shouldn't be in the business of calling "traditional but ugly" functions obsolete... because all the new-ish counterparts are not always available :(

The same rationale would apply to `lchown` only if we already provide it.  I don't want to withdraw `chown` because consumers may already be using it.  `lchown`, OTOH, is unnecessary unless somebody ports Nix to a weird platform that doesn't have `fchownat`.

jmmv:
I've added tests to ensure these functions exists in all CI platforms, and also confirmed in my project that `fchownat` exists for the older macOS that I need. So no need to add `lchown` from my point of view, which is what I wanted to confirm.

src/fcntl.rs
```diff
@@ -182,6 +183,15 @@ pub fn readlinkat<'a, P: ?Sized + NixPath>(dirfd: RawFd, path: &P, buffer: &'a m
     wrap_readlink_result(buffer, res)
 }
 
+/// Computes the raw fd consumed by a function of the form `*at`.
+#[inline]
```
asomers: Again, no need for `#[inline]` for private functions.

test/test_unistd.rs
```diff
@@ -301,6 +301,45 @@ fn test_getcwd() {
...
+
+#[test]
+fn test_fchownat() {
+    // Testing for anything other than our own UID/GID is hard.
+    let uid = Some(getuid());
+    let gid = Some(getgid());
+
+    let tempdir = tempfile::tempdir().unwrap();
+    let path = tempdir.path().join("file");
+    drop(File::create(&path).unwrap());
```
asomers: As in the `ftruncate` review, I prefer using lexical scope to `drop`.

jmmv:
Updated the commit message and removed the original contents of the PR message, which I think would be added to the history as well.

Also rebased on top of the truncate merge to avoid `CHANGELOG` conflicts.

PR #956, opened 2018-10-17
Add a truncate(2) wrapper

jmmv:
This also adds a test for truncate (obviously) but, while doing so, also
adds a test for the already-existing ftruncate.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -989,6 +989,20 @@ fn pipe2_setflags(fd1: RawFd, fd2: RawFd, flags: OFlag) -> Result<()> {
 /// Truncate a file to a specified length
 ///
 /// See also
+/// [truncate(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/truncate.html)
+pub fn truncate<P: ?Sized + NixPath>(path: &P, len: off_t) -> Result<()> {
+    let res = try!(path.with_nix_path(|cstr| {
+        unsafe {
+            libc::truncate(cstr.as_ptr(), len)
+        }
+    }));
+
+    Errno::result(res).map(drop)
+}
+
+/// Truncate a file to a specified length
+///
+/// See also
 /// [ftruncate(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/ftruncate.html)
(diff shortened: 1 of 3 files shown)


src/unistd.rs
```diff
@@ -986,6 +986,20 @@ fn pipe2_setflags(fd1: RawFd, fd2: RawFd, flags: OFlag) -> Result<()> {
...
+/// See also
+/// [truncate(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/truncate.html)
+pub fn truncate<P: ?Sized + NixPath>(path: &P, len: off_t) -> Result<()> {
+    let res = try!(path.with_nix_path(|cstr| {
+        unsafe {
+            libc::truncate(cstr.as_ptr(), len)
+        }
+    }));
+
+    Errno::result(res).map(drop)
```
asomers: I prefer `map(|| ())` for stylistic reasons.  `drop`, in this context, is basically being called for its side effect of returning nothing.

test/test_unistd.rs
```diff
@@ -390,6 +390,40 @@ fn test_pipe2() {
...
 
+#[test]
+fn test_truncate() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let path = tempdir.path().join("file");
+
+    let mut tmp = File::create(&path).unwrap();
+    const CONTENTS: &[u8] = b"12345678";
+    tmp.write_all(CONTENTS).unwrap();
+    drop(tmp);
```
asomers: Do you really need to `drop(tmp)`? If so, then I think it's better to simply put the preceding three lines within a `{}` block so `tmp` will lose lexical scope and drop automatically.

src/unistd.rs
jmmv: I'm not exactly sure what you'd like to see here, but given that this is a copy/paste from the many, many other functions that follow this style, I think keeping this for consistency is good.

test/test_unistd.rs
jmmv: Not really, but given that we are accessing the file through a different "handle" (filename vs. `File` object), it seems nicer to ensure that the object is closed. Used a scope now.

PR #957, opened 2018-10-19
Added direct support for `fchown` and `dup3`

### LINKED ISSUE #939: `unistd::dup3` never uses the `dup3` syscall (even if it might be available)

koutheir:
Added direct support for `fchown` and `dup3`.

Fixes #939.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -373,13 +373,13 @@ pub fn dup2(oldfd: RawFd, newfd: RawFd) -> Result<RawFd> {
 ///
 /// This function behaves similar to `dup2()` but allows for flags to be
 /// specified.
-#[cfg(any(target_os = "fuchsia",
-          target_os = "emscripten",
+#[cfg(any(target_os = "emscripten",
           target_os = "freebsd",
+          target_os = "fuchsia",
           target_os = "linux",
-          target_os = "solaris",
           target_os = "netbsd",
-          target_os = "openbsd"))]
+          target_os = "openbsd",
+          target_os = "solaris"))]
 pub fn dup3(oldfd: RawFd, newfd: RawFd, flags: OFlag) -> Result<RawFd> {
     let res = unsafe { libc::dup3(oldfd, newfd, flags.bits()) };
 


koutheir:
In order to be able to use `dup3()` on FreeBSD, I also added the pull request [#1100](https://github.com/rust-lang/libc/pull/1100) to the `libc` crate.

src/unistd.rs
```diff
@@ -374,26 +373,11 @@ pub fn dup2(oldfd: RawFd, newfd: RawFd) -> Result<RawFd> {
...
-
-    let fd = try!(dup2(oldfd, newfd));
-
-    if flags.contains(OFlag::O_CLOEXEC) {
-        if let Err(e) = fcntl(fd, F_SETFD(FdFlag::FD_CLOEXEC)) {
-            let _ = close(fd);
-            return Err(e);
-        }
-    }
+pub fn dup3(oldfd: RawFd, newfd: RawFd, flags: OFlag) -> Result<RawFd> {
```
asomers: OSX doesn't support `dup3`.  You'll have to disable this function on that, and possibly other, platforms.

src/unistd.rs
```diff
@@ -578,6 +562,55 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
...
+/// Change the ownership of the file at `path` to be owned by the specified
+/// `owner` (user) and `group` (see
+/// [lchown(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/lchown.html)).
+/// If `path` names a symbolic link, then ownership of the symbolic link file itself is changed.
+///
+/// The owner/group for the provided path name will not be modified if `None` is
+/// provided for that argument.  Ownership change will be attempted for the path
+/// only if `Some` owner/group is provided.
+#[inline]
+pub fn lchown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gid>) -> Result<()> {
```
asomers: `lchown` is redundant with `fchownat`, which is being added as part of PR #955 .  Unless there's a reason why `fchownat` won't work for you, you shouldn't bother with `lchown`.

src/unistd.rs
```diff
@@ -578,6 +562,55 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
...
+/// specified `owner` (user) and `group` (see
+/// [fchown(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fchown.html)).
+///
+/// The owner/group for the provided path name will not be modified if `None` is
+/// provided for that argument.  Ownership change will be attempted for the path
+/// only if `Some` owner/group is provided.
+#[inline]
+pub fn fchown(fd: RawFd, owner: Option<Uid>, group: Option<Gid>) -> Result<()> {
+    let res = unsafe { libc::fchown(fd,
+                             owner.map(Into::into).unwrap_or((0 as uid_t).wrapping_sub(1)),
```
asomers: You should probably wait until #955 lands, then rebase this PR and use the private `chown_raw_ids` method here.

src/unistd.rs
```diff
@@ -578,6 +562,55 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
...
+}
+
+/// Change the ownership of the file specified by the file descriptor `fd` to be owned by the
+/// specified `owner` (user) and `group` (see
+/// [fchown(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fchown.html)).
+///
+/// The owner/group for the provided path name will not be modified if `None` is
+/// provided for that argument.  Ownership change will be attempted for the path
+/// only if `Some` owner/group is provided.
+#[inline]
```
asomers: I wouldn't use `#[inline]`.  The compiler can figure out for itself whether that's called for if the user builds his applications with LTO.

src/unistd.rs
```diff
@@ -578,6 +562,55 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
...
+                             owner.map(Into::into).unwrap_or((0 as uid_t).wrapping_sub(1)),
+                             group.map(Into::into).unwrap_or((0 as gid_t).wrapping_sub(1))) };
+
+    Errno::result(res).map(drop)
+}
+
+/// Change the mode of the file at `path` (see
+/// [chmod(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html)).
+#[inline]
+pub fn chmod<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
```
asomers: This is also redundant with `fchmodat`.  I don't think we should add it to Nix unless you know of a reason why `fchmodat` won't work.

src/unistd.rs
koutheir: Done.

src/unistd.rs
koutheir: Agreed.

src/unistd.rs
koutheir: OK.

asomers:
Don't forget the CHANGELOG.  Make it clear that dup3 has been removed from OSX.

koutheir:
> Don't forget the CHANGELOG. Make it clear that dup3 has been removed from OSX.

The previous implementation of `dup3` was buggy because it was vulnerable to a race condition. I would argue that it should have never been implemented for OSX (or at least, not via `dup2` and `fcntl`). Furthermore, one of the principles in this project is to avoid providing APIs that are not actually implemented on the platform.

src/unistd.rs
koutheir: I would do so if I still have spare time by the time #955 lands.

koutheir: Done.

Susurrus:
> The previous implementation of `dup3` was buggy because it was vulnerable to a race condition. I would argue that it should have never been implemented for OSX (or at least, not via `dup2` and `fcntl`). Furthermore, one of the principles in this project is to avoid providing APIs that are not actually implemented on the platform.

The CHANGELOG should also be viewed as everything someone needs to read to allow them to upgrade. So if a user was using it on OS X, and we broke their code during an upgrade, there should be a part under the `Removed` section of the CHANGELOG stating a) what we removed and b) why we did so. Could you add a line there for this?

koutheir:
> Could you add a line there for this?

For `dup3`, I wrote two change log entries:
- Under `Changed`, to mention the direct system call, instead of emulation, and
- Under `Removed`, to inform OSX users of the removal.

src/unistd.rs
koutheir: OK

Susurrus:
There are Travis failures that need to be addressed. Besides that this looks good, though it should be squashed into 2 commits:
 * Add `fchown`
 * Modify `dup3`

After those are addressed, this is ready for merging.

PR #958, opened 2018-10-22
Added AddressType type to ptrace::linux + peek/poke user fix

xd009642:
This was added to the BSD ptrace API and probably should have been added to the linux API to make it easier to write code for both platforms without the user having to reexport the types to their own crate.

I went to use the latest nix myself and found that having this added would improve usability for nix users.

Could potentially add a type for what they return (BSDs: `c_int`, linux: `c_long`). Data input might be trickier as linux is `*mut c_void` and BSD just takes data as `c_int`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#955](https://github.com/nix-rust/nix/pull/955))
 - Added support for `ptrace` on BSD operating systems ([#949](https://github.com/nix-rust/nix/pull/949))
 - Added `ptrace` functions for reads and writes to tracee memory and ptrace kill 
-  ([#949](https://github.com/nix-rust/nix/pull/949))
+  ([#949](https://github.com/nix-rust/nix/pull/949)) ([#958](https://github.com/nix-rust/nix/pull/958))
 - Added a `acct` wrapper module for enabling and disabling process accounting
   ([#952](https://github.com/nix-rust/nix/pull/952))
 
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -174,10 +174,9 @@ libc_bitflags! {
 pub unsafe fn ptrace(request: Request, pid: Pid, addr: AddressType, data: *mut c_void) -> Result<c_long> {
     use self::Request::*;
     match request {
-        PTRACE_PEEKTEXT | PTRACE_PEEKDATA | PTRACE_PEEKUSER | PTRACE_GETSIGINFO | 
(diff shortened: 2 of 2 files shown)


xd009642:
@asomers do you have time to have a look at this?

asomers:
So the `PTRACE_PEEKUSER` mistake crept in in PR #949 ?  Then you should fix it.  It isn't really a regression from the users' point of view, because there hasn't been a release since then.  I think this PR is fine, but you do need to update the existing CHANGELOG entry for 949 with the additional PR link.

xd009642:
Yeah there's no tests for it so I didn't realise I'd made a mistake. Unfortunately it's pretty hard to test as it relies on knowledge of the user struct and what's happening in the test executable - unless you can think of a better way?

asomers:
No.  I'm not even clear on what `PTRACE_PEEK_USER` does.  Is it redundant with `PTRACE_GETREGS`?

xd009642:
I think it might be, `PTRACE_PEEK_USER` reads a single word from a given offset into the user struct, from looking at the man page it seems `PTRACE_GETREGS` reads the entire user struct not just one word.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#949](https://github.com/nix-rust/nix/pull/949))
 - Added a `acct` wrapper module for enabling and disabling process accounting
   ([#952](https://github.com/nix-rust/nix/pull/952))
+  ([#949](https://github.com/nix-rust/nix/pull/949)) ([#958](https://github.com/nix-rust/nix/pull/958))
```
asomers: Looks like you got a merge error when you rebased.

CHANGELOG.md
```diff
@@ -27,9 +27,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#955](https://github.com/nix-rust/nix/pull/955))
 - Added support for `ptrace` on BSD operating systems ([#949](https://github.com/nix-rust/nix/pull/949))
 - Added `ptrace` functions for reads and writes to tracee memory and ptrace kill 
-  ([#949](https://github.com/nix-rust/nix/pull/949))
+  ([#949](https://github.com/nix-rust/nix/pull/949)) ([#958](https://github.com/nix-rust/nix/pull/958))
 - Added a `acct` wrapper module for enabling and disabling process accounting
   ([#952](https://github.com/nix-rust/nix/pull/952))
+  ([#949](https://github.com/nix-rust/nix/pull/949))
```
asomers: CHANGELOG still isn't right.

xd009642:
Fixed (I hope)

PR #960, opened 2018-11-01
Added direct support for fchown() and dup3()

### LINKED ISSUE #939: `unistd::dup3` never uses the `dup3` syscall (even if it might be available)

koutheir:
- Added direct support for `dup3`.
- Removed emulation of `dup3`.
- Added support for `fchown`.

Fixes #939.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#949](https://github.com/nix-rust/nix/pull/949)) ([#958](https://github.com/nix-rust/nix/pull/958))
 - Added a `acct` wrapper module for enabling and disabling process accounting
   ([#952](https://github.com/nix-rust/nix/pull/952))
+- Added a `fchown` wrapper.
 
 ### Changed
 - Increased required Rust version to 1.22.1/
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -30,10 +30,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#949](https://github.com/nix-rust/nix/pull/949)) ([#958](https://github.com/nix-rust/nix/pull/958))
 - Added a `acct` wrapper module for enabling and disabling process accounting
   ([#952](https://github.com/nix-rust/nix/pull/952))
+- Added a `fchown` wrapper.
```
asomers: Needs a link to the PR here and below.

src/unistd.rs
```diff
@@ -633,6 +623,20 @@ pub fn fchownat<P: ?Sized + NixPath>(
...
+/// [fchown(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fchown.html)).
+///
+/// The owner/group for the provided path name will not be modified if `None` is
+/// provided for that argument.  Ownership change will be attempted for the path
+/// only if `Some` owner/group is provided.
+pub fn fchown(fd: RawFd, owner: Option<Uid>, group: Option<Gid>) -> Result<()> {
+    let (uid, gid) = chown_raw_ids(owner, group);
+    let res = unsafe { libc::fchown(fd, uid, gid) };
+
+    Errno::result(res).map(drop)
```
asomers: As a matter of style, I prefer `.map(|| ())`.  `drop`, in this case, is really just being used for its "side effect" of returning nothing.

PR #971, opened 2018-11-09
Implement copy_file_range()

### LINKED ISSUE #941: Implement copy_file_range() system call for Linux
[copy_file_range()](http://man7.org/linux/man-pages/man2/copy_file_range.2.html) has yet to be implemented.

ArniDagur:
This closes #941 (see issue for context).

I've decided that it is better for API forwards-compatibility to implement an empty CopyFileRangeFlags enum. I'm not sure how to do this in a good way, however, so please see the commit diff (https://github.com/nix-rust/nix/commit/48c9e6b431a08f8ce03252829f0afe0769e7bee6) and advise me how to do this in a good way, if you'd like. I've also allowed edits from maintainers, so you can make a commit yourself.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -6,6 +6,7 @@ use std::os::raw;
 use std::os::unix::io::RawFd;
 use std::ffi::OsStr;
 use std::os::unix::ffi::OsStrExt;
+use std::ptr;
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 use sys::uio::IoVec;  // For vmsplice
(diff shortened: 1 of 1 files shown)


src/fcntl.rs
```diff
@@ -305,6 +306,29 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
 
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+bitflags! {
+    pub struct CopyFileRangeFlags: c_int {
+        const NONE = 0 as c_int;
+    }
+}
+
+/// The flags argument is currently unused, and is provided to allow for future
```
asomers: How about documenting the rest of the function too, not just flags?

ArniDagur:
I've been struggling with the testing, so I'm wondering if you could lend a hand.

This code (using explicit file names `/tmp/test.txt` and `/tmp/test2.txt` for debugging purposes) works:

```rust
        let mut reader = File::open("/tmp/test.txt").unwrap();
        let mut writer = OpenOptions::new().read(true).write(true).truncate(true)
                            .open("/tmp/test2.txt").unwrap();

        let res = copy_file_range(reader.as_raw_fd(), None,
                                  writer.as_raw_fd(), None,
                                  2, CopyFileRangeFlags::empty())
                                  .unwrap();
        assert_eq!(2, res);
```

As seen here:
```
[arni@arni-pc][~/OpenSource/nix]% cat /tmp/test.txt
hello world
[arni@arni-pc][~/OpenSource/nix]% cat /tmp/test2.txt
he%      
```

But when trying to read the contents of `/tmp/test2.txt` after the copy, no bytes seem to be returned. I tried various methods.

asomers:
I'm not very familiar with `copy_file_range`.  Does it update the target's seek pointer?  If so, that would explain your problem.  Try seeking back to zero before reading test2.txt.  Also, you'll need to rebase due to some changes to our CI setup.

ArniDagur:
Superseded by #1069.

PR #972, opened 2018-11-16
Add support of TCP_CONGESTION for setsockopt

Fensteer:
Implementation proposal for support of TCP_CONGESTION param for `setsockopt`and `getsockopt` with the CString type.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -5,6 +5,14 @@ use sys::time::TimeVal;
 use libc::{self, c_int, uint8_t, c_void, socklen_t};
 use std::mem;
 use std::os::unix::io::RawFd;
+use std::ffi::{OsStr, OsString};
+#[cfg(target_family = "unix")]
+use std::os::unix::ffi::OsStrExt;
+
+// Constants
+// TCP_CA_NAME_MAX isn't defined in user space include files
+#[cfg(any(target_os = "freebsd", target_os = "linux"))] 
+const TCP_CA_NAME_MAX: usize = 16;
 
 /// Helper for implementing `SetSockOpt` for a given socket option. See
 /// [`::sys::socket::SetSockOpt`](sys/socket/trait.SetSockOpt.html).
@@ -152,6 +160,10 @@ macro_rules! sockopt_impl {
         sockopt_impl!(Both, $name, $level, $flag, usize, GetUsize, SetUsize);
     };
 
+    (Both, $name:ident, $level:path, $flag:path, OsString<$array:ty>) => {
+        sockopt_impl!(Both, $name, $level, $flag, OsString, GetOsString<$array>, SetOsString);
+    };
+
     /*
      * Matchers with generic getter types must be placed at the end, so
      * they'll only match _after_ specialized matchers fail
@@ -257,6 +269,8 @@ sockopt_impl!(Both, BindAny, libc::IPPROTO_IP, libc::IP_BINDANY, bool);
 sockopt_impl!(Both, Mark, libc::SOL_SOCKET, libc::SO_MARK, u32);
 #[cfg(any(target_os = "android", target_os = "linux"))]
 sockopt_impl!(Both, PassCred, libc::SOL_SOCKET, libc::SO_PASSCRED, bool);
+#[cfg(any(target_os = "freebsd", target_os = "linux"))] 
+sockopt_impl!(Both, TcpCongestion, libc::IPPROTO_TCP, libc::TCP_CONGESTION, OsString<[u8; TCP_CA_NAME_MAX]>);
 
 /*
  *
@@ -477,6 +491,53 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
         mem::size_of::<c_int>() as socklen_t
     }
 }
+
+/// Getter for a `OsString` value.
+struct GetOsString<T: AsMut<[u8]>> {
+    len: socklen_t,
+    val: T,
+}
+
+unsafe impl<T: AsMut<[u8]>> Get<OsString> for GetOsString<T> {
+    unsafe fn blank() -> Self {
+        GetOsString {
+            len: mem::size_of::<T>() as socklen_t,
+            val: mem::zeroed(),
+        }
+    }
+
+    fn ffi_ptr(&mut self) -> *mut c_void {
+        &mut self.val as *mut T as *mut c_void
+    }
+
+    fn ffi_len(&mut self) -> *mut socklen_t {
+        &mut self.len
+    }
+
+    unsafe fn unwrap(mut self) -> OsString {
+        OsStr::from_bytes(self.val.as_mut()).to_owned()
+    }
+}
+
+/// Setter for a `OsString` value.
+struct SetOsString<'a> {
+    val: &'a OsStr,
+}
+
+unsafe impl<'a> Set<'a, OsString> for SetOsString<'a> {
+    fn new(val: &'a OsString) -> SetOsString {
+        SetOsString { val: val.as_os_str() }
+    }
+
+    fn ffi_ptr(&self) -> *const c_void {
+        self.val.as_bytes().as_ptr() as *const c_void
+    }
+
+    fn ffi_len(&self) -> socklen_t {
+        self.val.len() as socklen_t
+    }
+}
+
 
 #[cfg(test)]
 mod test {
--- a/test/sys/test_sockopt.rs
+++ b/test/sys/test_sockopt.rs
@@ -13,3 +13,28 @@ fn test_so_buf() {
     let actual = getsockopt(fd, sockopt::RcvBuf).unwrap();
     assert!(actual >= bufsize);
 }
+
+// The CI doesn't supported getsockopt and setsockopt on emulated processors.
+// It's beleived that a QEMU issue, the tests run ok on a fully emulated system.
+// Current CI just run the binary with QEMU but the Kernel remains the same as the host.
+// So the syscall doesn't work properly unless the kernel is also emulated.
+#[test]
+#[cfg(all(
+    any(target_arch = "x86", target_arch = "x86_64"),
+    any(target_os = "freebsd", target_os = "linux")
+))]
+fn test_tcp_congestion() {
+    use std::ffi::OsString;
+
+    let fd = socket(AddressFamily::Inet, SockType::Stream, SockFlag::empty(), None).unwrap();
+
+    let val = getsockopt(fd, sockopt::TcpCongestion).unwrap();
+    setsockopt(fd, sockopt::TcpCongestion, &val).unwrap();
+
+    setsockopt(fd, sockopt::TcpCongestion, &OsString::from("tcp_congestion_does_not_exist")).unwrap_err();
+
+    assert_eq!(
+        getsockopt(fd, sockopt::TcpCongestion).unwrap(),
+        val
+    );
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+- Added support of CString type in `setsockopt`.
+  ([#972](https://github.com/nix-rust/nix/pull/972))
+- Added option `TCP_CONGESTION` in `setsockopt`.
+  ([#972](https://github.com/nix-rust/nix/pull/972))
+
 ### Changed
 ### Fixed
 ### Removed


src/sys/socket/sockopt.rs
```diff
@@ -152,6 +153,10 @@ macro_rules! sockopt_impl {
         sockopt_impl!(Both, $name, $level, $flag, usize, GetUsize, SetUsize);
     };
 
+    (Both, $name:ident, $level:path, $flag:path, usize) => {
+        sockopt_impl!(Both, $name, $level, $flag, CString, GetCString, SetCString);
```
asomers: I think `OsString` would be more appropriate than `CString`.

src/sys/socket/sockopt.rs
```diff
@@ -478,6 +485,67 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
     }
 }
 
+/// Size of empty CString buffer
+/// There is no default size in kernel documentation
+const SIZE_CSTRING_MAX : usize = 64;
```
asomers: I believe the symbol you're looking for is `TCP_CA_NAME_MAX`.  You should add it to libc.

src/sys/socket/sockopt.rs
```diff
@@ -478,6 +485,67 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
...
+    fn ffi_ptr(&mut self) -> *mut c_void {
+        self.val.as_mut_ptr() as *mut c_void
+    }
+
+    fn ffi_len(&mut self) -> *mut socklen_t {
+        &mut self.len
+    }
+
+    unsafe fn unwrap(self) -> CString {
+        // Crop length of val with the first null byte (because CString::new add null byte)
```
asomers: s/add/adds a/

src/sys/socket/sockopt.rs
```diff
@@ -478,6 +485,67 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
...
+        let len = {
+            let mut i = 0;
+            loop {
+                if i == self.len || self.val[i as usize] == 0 {
```
asomers: You can replace the entire body of this method with `CStr::from_ptr(self.ffi_ptr()).to_owned()`.  Or, if you replace `CString` with `OsString`, you can use `OsStr::from_bytes(self.val).to_owned()`.

src/sys/socket/sockopt.rs
Fensteer: I'm not sure of the difference between `OsString` and `CString`, and which is the best in this situation.

src/sys/socket/sockopt.rs
Fensteer: `TCP_CA_NAME_MAX` is only for TCP_CONGESTION parameter. I wanted to make a generic string type for setsockopt / getsockopt.

Fensteer:
I use this code to test the feature but I don't know how to integrate into the CI (I don't know if `reno` is in `tcp_available_congestion_control`)

```rust
extern crate nix;

use nix::sys::socket::sockopt;
use nix::sys::socket::{getsockopt, setsockopt, socket, AddressFamily, SockFlag, SockType};
use std::ffi::CString;

fn main() {
    let socket = socket(
        AddressFamily::Inet,
        SockType::Stream,
        SockFlag::empty(),
        None,
    ).unwrap();

    let algo = CString::new("reno").expect("Cannot create CString");

    setsockopt(socket, sockopt::TcpCongestion, &algo).expect("setsockopt");

    assert_eq!(algo, CString::new("reno").unwrap());

    assert_eq!(
        getsockopt(socket, sockopt::TcpCongestion).unwrap(),
        CString::new("reno").unwrap()
    );
}
```

Fensteer:
I don't understand why the buildbot is failing because `libc::TCP_CONGESTION` exists in latest version of `libc` : https://docs.rs/libc/0.2.44/libc/constant.TCP_CONGESTION.html

asomers:
Looks like libc has only defined `TCP_CONGESTION` for Linux and uclibc.  You'll have to submit a PR to libc to add the definition for FreeBSD.  You should also check if it's defined on any other OS.

src/sys/socket/sockopt.rs
asomers: `CString` should be used when you're interfacing with C libraries.  `OsString` should be used when you're interfacing with the operating system.  On Unix, they're usually the same thing.  But since `setsockopt` is provided by the OS, we should use `OsString` instead.

src/sys/socket/sockopt.rs
asomers: Oh, I see now.  Do you think the `GetCString` and `SetCString` structs should be generic on the string's size?

Fensteer:
Done : https://github.com/rust-lang/libc/pull/1151

src/sys/socket/sockopt.rs
Fensteer: To make size generic, we have to add an `Option<usize>` parameter to the `blank` method like: https://github.com/nix-rust/nix/pull/866/files#diff-16839e6bfe96cb0a5f52aef846eefe8eR296
Or do you have another idea?

asomers: Can't you just do something like this?
```rust
struct GetString<T> {
    len: usize,
    val: T
}
impl<T> GetString<T> {
    unsafe fn blank() -> Self {
        GetStruct {
            len: mem::size_of::<T>() as socklen_t,
            val: mem::zeroed(),
         }
    }
}
```
Then instantiate it like `GetStruct<[u8; 16]>`?

Fensteer: Done

src/sys/socket/sockopt.rs
Fensteer: Ok thank you.
However I cannot use `OsString` because I can't get ptr on it.

Fensteer: Yes, but self.val is parametrized so I can't call `as_bytes()`.
```
mismatched types
expected &[u8], found type parameter
note: expected type `&[u8]` found type `T`
```

asomers: Oh, I thought you were talking about the `as_ptr` call in `SetCString`.  Where does `OsString` cause a problem?

Fensteer: Oh no sorry I mis explained. It's in `GetOsString`, I can't call `OsStr::from_bytes` because generic T has not conditions to be bytes. 
```
/// Getter for a `OsString` value.
struct GetOsString<T> {
    len: socklen_t,
    val: T,
}

unsafe impl<T> Get<OsString> for GetOsString<T> {
   ...
    unsafe fn unwrap(mut self) -> OsString {
        OsStr::from_bytes(self.val).to_owned()
    }
```

asomers: You probably need to add a where guard to T: `T: AsRef<[u8]>`.

Fensteer: Thank you. It's done but with OsString, `getsockopt(socket, sockopt::TcpCongestion)` returns an OsString with "reno\0\0\0\0\0\0\0\0\0\0\0\0" (with CString it's cropped to "reno").

src/sys/socket/sockopt.rs
```diff
@@ -513,23 +514,23 @@ unsafe impl<T> Get<CString> for GetCString<T> {
         &mut self.len
     }
 
-    unsafe fn unwrap(mut self) -> CString {
-        CStr::from_ptr(self.ffi_ptr() as *const c_char).to_owned()
+    unsafe fn unwrap(self) -> OsString {
```
asomers: Does this method still need to be `unsafe`?

src/sys/socket/sockopt.rs
```diff
@@ -513,23 +514,23 @@ unsafe impl<T> Get<CString> for GetCString<T> {
...
-    val: &'a CString,
+/// Setter for a `OsString` value.
+struct SetOsString<'a> {
+    val: &'a OsString,
 }
 
-unsafe impl<'a> Set<'a, CString> for SetCString<'a> {
-    fn new(val: &'a CString) -> SetCString {
-        SetCString { val }
+unsafe impl<'a> Set<'a, OsString> for SetOsString<'a> {
```
asomers: I think this entire `impl` block doesn't need to be `unsafe`.

src/sys/socket/sockopt.rs
```diff
@@ -478,6 +492,53 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
     }
 }
 
+/// Getter for a `OsString` value.
+struct GetOsString<T: AsRef<[u8]>> {
+    len: socklen_t,
+    val: T,
+}
+
+unsafe impl<T: AsRef<[u8]>> Get<OsString> for GetOsString<T> {
```
asomers: You can remove the `unsafe` keyword on the `impl` block, because you use it on each individual method.

src/sys/socket/sockopt.rs
```diff
@@ -152,6 +160,10 @@ macro_rules! sockopt_impl {
         sockopt_impl!(Both, $name, $level, $flag, usize, GetUsize, SetUsize);
     };
 
+    (Both, $name:ident, $level:path, $flag:path, OsString<$ty:ty>) => {
```
asomers: The variable name `ty` is confusing here, because that name is also used by the `getsockopt_impl` macro.  Could you please rename it?

src/sys/socket/sockopt.rs
```diff
@@ -478,6 +492,53 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
     }
 }
 
+/// Getter for a `OsString` value.
+struct GetOsString<T: AsRef<[u8]>> {
```
asomers: This should be `AsMut`, not `AsRef`, since `libc::getsockopt` is going to modify it.

src/sys/socket/sockopt.rs
```diff
@@ -478,6 +492,53 @@ unsafe impl<'a> Set<'a, usize> for SetUsize {
...
+    }
+
+    unsafe fn unwrap(self) -> OsString {
+        OsStr::from_bytes(self.val.as_ref()).to_owned()
+    }
+}
+
+/// Setter for a `OsString` value.
+struct SetOsString<'a> {
+    val: &'a OsString,
```
asomers: You should basically never use a reference to an `OsString`.  Instead, use its borrowed form: `&OsStr`.

src/sys/socket/sockopt.rs
Fensteer: You're right, but this is defined by the trait:
``` rust
unsafe trait Get<T> {
     ...
     unsafe fn unwrap(self) -> T;
}
```

src/sys/socket/sockopt.rs
Fensteer: You're right, but this is defined by the trait
```rust
unsafe trait Set<'a, T>
```

src/sys/socket/sockopt.rs
Fensteer: The same

src/sys/socket/sockopt.rs
Fensteer: Right. Done.

src/sys/socket/sockopt.rs
Fensteer: Done

src/sys/socket/sockopt.rs
Fensteer: Done

src/sys/socket/sockopt.rs
asomers: Hm, that's a mistake.  A Trait should only be marked as `unsafe` if the `impl` itself can cause unsafe behavior, not just because it has unsafe methods.  For example, `Send`, and `Sync` are unsafe methods.  It's fine if you want to leave that in for this PR, but we should remove it in another PR.

asomers:
Ok, this PR is starting to look pretty good.  I think there are only two issues left:
1) Could you add a test?  On FreeBSD, I believe getting the option should work fine.  Setting it is a little harder to test, because there's only one valid option unless you load a kernel module.  But you should be able to set it back to the same value as the default.  I'm not sure what will happen on Linux; you'll have to try to find out.
2) You'll need to rebase and fix the changelog, since Nix just released 0.12.0.

Fensteer:
CI fails but actually the code is OK. I tried CI on a full emulated MIPS or PowerPC system and the test run well. I think that the CI fails because only the binary is run on a emulated processor but the kernel is still the same than the host.

asomers:
> CI fails but actually the code is OK. I tried CI on a full emulated MIPS or PowerPC system and the test run well. I think that the CI fails because only the binary is run on a emulated processor but the kernel is still the same than the host.

QEMU does emulate the full system.  It's not just a binary translator.  However, it's got a lot of bugs.  What did you use for a "full emulated MIPS or PowerPC system"?

test/sys/test_sockopt.rs
```diff
@@ -17,28 +14,24 @@ fn test_so_buf() {
     assert!(actual >= bufsize);
 }
 
+// The CI doesn't supported getsockopt and setsockopt on emulated processors
```
asomers: You should mention in your comment that it's believed to be a QEMU issue.

Fensteer: > > CI fails but actually the code is OK. I tried CI on a full emulated MIPS or PowerPC system and the test run well. I think that the CI fails because only the binary is run on a emulated processor but the kernel is still the same than the host.
> 
> QEMU does emulate the full system. It's not just a binary translator. However, it's got a lot of bugs. What did you use for a "full emulated MIPS or PowerPC system"?

I installed a full MIPS debian in a virtual machine with `qemu-system-mips` and the run the test inside it. The CI just cross compile the binary and run it with `qemu-mips` so the syscalls are not made by a native kernel.

Fensteer:
Rebase done and I squashed some commits.
CHANGELOG fixed.

asomers:
Could you please squash all commits together?

asomers:
Could you please squash all commits together?

asomers:
Could you please squash all commits together?

asomers:
Could you please squash all commits together?

Fensteer:
Done

Fensteer:
Thanks for the reviews and advices !

PR #973, opened 2018-11-16
CHANGELOG: move entry from #884 from 0.11.0 to [Unreleased]

mpasternacki:
PR #884 added changelog entry in section for 0.11.0, which has been already released, instead of in the [Unreleased] section (which wasn't present yet). Moved the entry where it belongs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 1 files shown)


mpasternacki:
Original entry seems to be in section [Unreleased] for 0.11.0, as the commit is dated before the 0.11.0 release. PR has been merged after the release, though, but the entry merged where it was added, at the end of 0.11.0's entries.

asomers:
Ahh, that makes sense.  My faith in bors is restored.  I guess we need a policy of rebasing all PRs after a release.

PR #984, opened 2018-11-28
Prefer `map(drop)` to `map(|_| ())`

asomers:
I previously advocated for the latter syntax on stylistic grounds.  But
it generates less efficient code, because it creates a new lambda
function for each usage.  The optimizer does not combine them.  This
change saves about 6KB of code.


asomers:
Bors r+

PR #986, opened 2018-12-01
Add more platforms to CI

Susurrus:
* aarch64-unknown-linux-musl
 * sparc64-unknown-linux-gnu
 * x86_64-unknown-linux-gnux32

These don't pass yet, but should be supportable by `nix`, so let's add them now as platforms that are allowed to fail so we can move towards fully supporting them.


Susurrus:
Not certain why these weren't working by just adding them. Tried hacking them in, but need to come back and relearn our CI setup to actually get these fixed.

asomers:
Don't forget to add new platforms to the README, if we're going to support them.

Susurrus:
Coming back to this now, so rebased and updated the CHANGELOG and README, but will likely wait until #1035 is merged to really dig into why these platforms are failing.

SteveLauC:
Close this PR for the same reason as #2255

PR #987, opened 2018-12-01
Fix builds on sparc64-unknown-linux-gnu

Susurrus:


glaubitz:
Any chance a new version can be released with those changes included?

rust-nix in Debian still fails to build on sparc64 for version 0.12.0: https://buildd.debian.org/status/fetch.php?pkg=rust-nix&arch=sparc64&ver=0.12.0-1&stamp=1547420163&raw=0

PR #989, opened 2018-12-03
DragonflyBSD: Remove unused Errno's

levex:
EUNUSED* were removed from <sys/errno.h> in DragonflyBSD, so there is no
need for them to be in nix either. This also fixes the build on
DragonflyBSD.

r? @asomers 
cc/ @mneumann 
Signed-off-by: Levente Kurusa <lkurusa@acm.org>

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -1,6 +1,6 @@
 #[cfg(not(target_os = "dragonfly"))]
 use libc;
-use libc::c_int;
+use libc::{c_int, c_void};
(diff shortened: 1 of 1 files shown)


levex:
@asomers ping can we merge this? it fixes the build on DragonflyBSD.

asomers:
Is everything good on the libc front?

levex:
Yup, see https://github.com/rust-lang/libc/pull/1160

PR #891, opened 2018-04-25
support PKTINFO cmsgs for Linux/macOS/iOS

mcginty:
The libc changes have been merged (https://github.com/rust-lang/libc/pull/980) now, so looks like it's finally time to submit a PR for this :).

PR #990, opened 2018-12-04
Add cmsg PKTINFO for IPv4 and IPv6.

pusateri:
Replaces #891 and attempts to address all previous concerns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Added PKTINFO(V4) & V6PKTINFO cmsg support - Android/FreeBSD/iOS/Linux/MacOS.
+  ([#990](https://github.com/nix-rust/nix/pull/990))
 ### Added
 - Added support of CString type in `setsockopt`.
   ([#972](https://github.com/nix-rust/nix/pull/972))
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -201,10 +201,10 @@ fn test_scm_rights_single_cmsg_multiple_fds() {
         }
         assert!(cmsgs.next().is_none(), "unexpected control msg");
 
-        assert_eq!(iovec[0].as_slice(), [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8]);
+        assert_eq!(iovec[0].as_slice(), [1u8, 2, 3, 4, 5, 6, 7, 8]);
     });
 
-    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
+    let slice = [1u8, 2, 3, 4, 5, 6, 7, 8];
     let iov = [IoVec::from_slice(&slice)];
     let fds = [libc::STDIN_FILENO, libc::STDOUT_FILENO];    // pass stdin and stdout
     let cmsg = [ControlMessage::ScmRights(&fds)];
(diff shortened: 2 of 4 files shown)


pusateri:
FreeBSD has IP_RECVIF and IP_RECVDSTADDR/IP_SENDSRCADDR to accomplish the same thing as IP_PKTINFO. I haven't added that yet. I will once this is committed.

pusateri:
OpenBSD has IPV6_PKTINFO but cargo test fails before these changes so I did not yet include OpenBSD cfg support.

pusateri:
Not sure if ```test_recv_ipv6pktinfo()``` should pass if there is no IPv6 address on the loopback interface. ```::1``` is missing on the amd64_fbsd11 ```buildbot``` automated test. The other alternative is to mark the test as ignore and let is only be run intentionally.

pusateri:
mips-unknown-linux-gnu looks like an endian problem. Not sure how deep. Index probably should have been 1 and it was 1000000 (hex).

asomers:
> I tested on stable-armv7-unknown-linux-gnueabihf, Debian GNU/Linux 8.11 (jessie) with stable 1.30.0 and it passed. Not sure what's different with the test apparatus other than it using 1.24.1.

Were you actually running Debian on an arm?  The test apparatus uses qemu to emulate the embedded architectures, and qemu's got some bugs.

asomers:
Looks like you're getting `EPFNOSUPPORT` on many of the QEMU platforms.  You can either disable the test for those platforms, or else accept `EPFNOSUPPORT` (and print a warning) in the test.

pusateri:
Cool. All that's left to fix is the big endian problem.

pusateri:
After some trouble, I got Ubuntu 16.04 LTS installed on a PowerMac Dual G5 and ran the pktinfo tests natively and both IPv4 and IPv6 tests passed. Maybe the endian bug is in the qemu emulator? I think the code is ok and don't have any more changes to make.

pusateri:
This leaves us in a bad situation for testing though. I tested with version 1.31.0. I will test with 1.24.1 and verify since that's what the emulator is running.

asomers:
There definitely are bugs in QEMU.  That's why many tests are disabled for the QEMU platforms.  It's not necessary for this PR that you fix QEMU; just make sure the tests pass on real platforms and all checkmarks are green.

pusateri:
The powerpc tests passed on 1.24.1 as well. Currently, other tests are disabled for qemu bugs:
```
// qemu's handling of multiple cmsgs is bugged, ignore tests on non-x86
// see https://bugs.launchpad.net/qemu/+bug/1781280
#[cfg_attr(not(any(target_arch = "x86_64", target_arch = "x86")), ignore)]
```
Should I turn my tests off for powerpc and mips?

asomers:
That's what I would do.

asomers:
I'm afraid that you'll need to rebase.  You somehow merged several other PRs into this one.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- Added PKTINFO cmsg support on Android/iOS/Linux/MacOS.
+- Added V6PKTINFO cmsg support on Android/FreeBSD/iOS/Linux/MacOS.
+- ([#990](https://github.com/nix-rust/nix/pull/990))
```
asomers: The PR doesn't get its own bullet.  These three lines should either all share a bullet, or the "Added" lines should each get their own and each have the PR link.

test/sys/test_socket.rs
```diff
@@ -451,3 +451,242 @@ pub fn test_syscontrol() {
...
+        }
+        None
+    }
+
+    let _m = ::LOOP_MTX
+        .lock()
+        .expect("Mutex got poisoned by another test");
+
+    let lo_ifaddr = loopback_v4addr();
+    let (lo_name, lo) = match lo_ifaddr {
```
asomers: Could you simplify the test by hardcoding this to `("lo0", "127.0.0.1")`?

test/test.rs
```diff
@@ -77,4 +77,6 @@ lazy_static! {
     pub static ref PTSNAME_MTX: Mutex<()> = Mutex::new(());
     /// Any test that alters signal handling must grab this mutex.
     pub static ref SIGNAL_MTX: Mutex<()> = Mutex::new(());
+    /// Any test sending packets over the loopback interface
+    pub static ref LOOP_MTX: Mutex<()> = Mutex::new(());
```
asomers: Is this really necessary?  Why not just bind each test to a different port?

test/sys/test_socket.rs
```diff
@@ -451,3 +451,242 @@ pub fn test_syscontrol() {
...
+            _ => (),
+        }
+        assert!(cmsgs.next().is_none(), "unexpected additional control msg");
+        assert_eq!(
+            iovec[0].as_slice(),
+            [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8]
+        );
+    });
+
+    let slice = [1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8];
```
asomers: Pro tip: You can leave out the "u8" on the second and subsequent elements.

pusateri:
From the nix gitter from me:
Is it ok to send packets through the loopback interface with sendmsg()/recvmsg() in a test (test/sys/test_socket.rs)?

@asomers replied:
@pusateri it's ok as long as you can guarantee that they won't interfere with anything else
Use the appropriate mutex from test/test.rs or create a new one to protect access to the loopback interface.
And don't assume that its address will be 127.0.0.1.

So that's what I did. Also Linux uses "lo" and BSDs and macOS use "lo0". Rather than hard coding it, it's safer to look for the LOOPBACK flag.

So now which way do you prefer?

test/sys/test_socket.rs
pusateri: Do you want me to fix the versions above in the other existing tests or just this new one?

asomers:
> From the nix gitter from me:
> Is it ok to send packets through the loopback interface with sendmsg()/recvmsg() in a test (test/sys/test_socket.rs)?
> 
> @asomers replied:
> @pusateri it's ok as long as you can guarantee that they won't interfere with anything else
> Use the appropriate mutex from test/test.rs or create a new one to protect access to the loopback interface.
> And don't assume that its address will be 127.0.0.1.
> 
> So that's what I did. Also Linux uses "lo" and BSDs and macOS use "lo0". Rather than hard coding it, it's safer to look for the LOOPBACK flag.
> 
> So now which way do you prefer?

Well, that answers the question about hardcoding the interface name and address.  I guess it's not an issue for most tests, because most tests only need the address, not the name.  As for the mutex, I think it's better to use separate port numbers.  Using distinct ports is a way to guarantee that the tests won't interfere with each other.

pusateri:
It already has the kernel assign a random unused port number so all I should need to do is remove the mutex, fix the CHANGELOG, and remove the unnecessary u8's?

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 - Added PKTINFO cmsg support on Android/iOS/Linux/MacOS.
 - Added V6PKTINFO cmsg support on Android/FreeBSD/iOS/Linux/MacOS.
-- ([#990](https://github.com/nix-rust/nix/pull/990))
+  ([#990](https://github.com/nix-rust/nix/pull/990))
```
asomers: This still isn't correct because the link is only associated with the V6PKTINFO bullet.  You need to either duplicate the link for both bullets, or collapse everything into a single bullet.

asomers:
Ok, everything is good!  Please squash and then I'll merge.

PR #991, opened 2018-12-05
Add `mman::mprotect`

acfoltzer:
It doesn't look like there's an existing test suite for `mmap` and friends, so this is regrettably untested.

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -258,6 +258,35 @@ pub unsafe fn madvise(addr: *mut c_void, length: size_t, advise: MmapAdvise) -> 
     Errno::result(libc::madvise(addr, length, advise as i32)).map(drop)
 }
 
+/// Set protection of memory mapping.
+///
+/// See [`mprotect(3)`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mprotect.html) for
+/// details.
+///
+/// # Safety
+///
+/// Calls to `mprotect` are inherently unsafe, as changes to memory protections can lead to
+/// SIGSEGVs.
+///
+/// ```
+/// # use nix::libc::size_t;
+/// # use nix::sys::mman::{mmap, mprotect, MapFlags, ProtFlags};
+/// # use std::ptr;
+/// const ONE_K: size_t = 1024;
+/// let mut slice: &mut [u8] = unsafe {
+///     let mem = mmap(ptr::null_mut(), ONE_K, ProtFlags::PROT_NONE,
+///                    MapFlags::MAP_ANON | MapFlags::MAP_PRIVATE, -1, 0).unwrap();
+///     mprotect(mem, ONE_K, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE).unwrap();
+///     std::slice::from_raw_parts_mut(mem as *mut u8, ONE_K)
+/// };
+/// assert_eq!(slice[0], 0x00);
+/// slice[0] = 0xFF;
+/// assert_eq!(slice[0], 0xFF);
+/// ```
+pub unsafe fn mprotect(addr: *mut c_void, length: size_t, prot: ProtFlags) -> Result<()> {
+    Errno::result(libc::mprotect(addr, length, prot.bits())).map(drop)
+}
+
 pub unsafe fn msync(addr: *mut c_void, length: size_t, flags: MsFlags) -> Result<()> {
     Errno::result(libc::msync(addr, length, flags.bits())).map(drop)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1010](https://github.com/nix-rust/nix/pull/1010))
 - Added `nix::sys::signal::signal`.
   ([#817](https://github.com/nix-rust/nix/pull/817))
+- Added an `mprotect` wrapper.
+  ([#991](https://github.com/nix-rust/nix/pull/991))
 
 ### Changed
 ### Fixed


asomers:
Would you be willing to write a testsuite for `mmap`?

acfoltzer:
I don't think I'm the right person to do that; I don't know the APIs very deeply.

acfoltzer:
Is there anything else I need to do so that this can be merged? It keeps picking up conflicts in the changelog...

asomers:
I would feel much better about the PR if there were some tests.  Even a simple doc test would be good.  Right now it's not obvious how to use it.  Also, `unsafe` functions' documentation blocks should say why they're `unsafe`, and what the user must do to use them safely.

acfoltzer:
Got it; I was following the pattern of the rest of the definitions in that module, which don't have those things; I'll get a doctest merged shortly.

acfoltzer:
Well, it looks like my test is only gonna pass on Linux, due to differences in how `mmap` works there. This is part of why I was hesitant to add any tests for `mprotect` while `mmap` itself doesn't have any; it's a big, complicated API and I am not equipped to portably test it.

src/sys/mman.rs
```diff
@@ -258,6 +258,27 @@ pub unsafe fn madvise(addr: *mut c_void, length: size_t, advise: MmapAdvise) ->
     Errno::result(libc::madvise(addr, length, advise as i32)).map(drop)
 }
 
+/// Calls to `mprotect` are inherently unsafe, as changes to memory protections can lead to
```
asomers: The doc comment should start with a one-line description of the function.  This can optionallly be followed by a multiline description.  Safety information should be in a `#Safety` section.  There should also be a link to an online man page.  For standardized functions like this one, you should use the open group's man page: http://pubs.opengroup.org/onlinepubs/9699919799/functions/mprotect.html .

src/sys/mman.rs
```diff
@@ -258,6 +258,27 @@ pub unsafe fn madvise(addr: *mut c_void, length: size_t, advise: MmapAdvise) ->
...
+/// SIGSEGVs.
+///
+/// ```
+/// use nix::sys::mman::{mmap, mprotect, MapFlags, ProtFlags};
```
asomers: For doctests, the use lines should usually be hidden by a leading `#`.

src/sys/mman.rs
```diff
@@ -258,6 +258,27 @@ pub unsafe fn madvise(addr: *mut c_void, length: size_t, advise: MmapAdvise) ->
...
+/// use std::ptr;
+///
+/// let mut slice: &mut [u8] = unsafe {
+///     let mem = mmap(ptr::null_mut(), 1024, ProtFlags::PROT_NONE,
+///                    MapFlags::MAP_ANONYMOUS | MapFlags::MAP_PRIVATE, -1, 0).unwrap();
```
asomers: `MAP_ANONYMOUS` is a nonstandard flag, so it doesn't make a great example.

src/sys/mman.rs
```diff
@@ -258,6 +258,27 @@ pub unsafe fn madvise(addr: *mut c_void, length: size_t, advise: MmapAdvise) ->
...
 
+/// Calls to `mprotect` are inherently unsafe, as changes to memory protections can lead to
+/// SIGSEGVs.
+///
+/// ```
+/// use nix::sys::mman::{mmap, mprotect, MapFlags, ProtFlags};
+/// use std::ptr;
+///
+/// let mut slice: &mut [u8] = unsafe {
+///     let mem = mmap(ptr::null_mut(), 1024, ProtFlags::PROT_NONE,
```
asomers: The `1024` should be defined as a constant, rather than repeated 3 times.

src/sys/mman.rs
acfoltzer: I've only ever used `mmap` with `MAP_ANONYMOUS`, but from what I can tell by the POSIX manpage, it would need a file descriptor if that flag isn't used. This has to either be backed by a real file or a typed memory object, the API for which `nix` doesn't support. This is starting to feel _way_ too complicated for a 3 line wrapper function, and would be more appropriate for a larger `mman` test suite.

acfoltzer: It looks like `MAP_ANON` at least passes on all the CI platforms, even though it's also not in POSIX. Good enough?

src/sys/mman.rs
```diff
@@ -258,6 +258,33 @@ pub unsafe fn madvise(addr: *mut c_void, length: size_t, advise: MmapAdvise) ->
     Errno::result(libc::madvise(addr, length, advise as i32)).map(drop)
 }
 
+/// Set protection of memory mapping.
+///
+/// See [`mprotect(3)`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mprotect.html) for
+/// details.
+///
+/// # Safety Calls to `mprotect` are inherently unsafe, as changes to memory protections can lead to
```
asomers: You need a newline after "Safety".  Also, is there any advice you can offer on how the user can use this function safely?

acfoltzer: Newline fixed; Emacs got overly enthusiastic about reflowing that paragraph.

As for advice, anything I write would be just rewriting the linked manpage. It's such a broad API with so many potential implications that I think it would serve folks best to just use that as a resource.

asomers:
Can you please squash your two commits into one?  We can't do that ourselves because of bors.

acfoltzer:
Squashed and rebased on the current master; I think this CI failure is unrelated to this PR though

acfoltzer:
@asomers is this ready to merge now?

asomers:
Uh, could you remove that odd blank line in the CHANGELOG?

acfoltzer:
Got it; of course the online editor first tried to make a merge conflict out of it :(

PR #993, opened 2018-12-05
Don't reference packed structs.

### LINKED ISSUE #992: borrow of packed field is unsafe
[rust lang issue #46043](https://github.com/rust-lang/rust/issues/46043) describes the problem. The error shows up on NetBSD tests for sys/socket/mod.rs:262 and sys/socket/addr.rs:500
```
--> src/sys/socket/addr.rs:500:9
    |
500 |         self.0.s_addr.hash(s)
    |         ^^^^^^^^^^^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
```

pusateri:
Fixes #992.
Don't merge this yet. I have more testing to do. I just am pushing it up for others.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -497,7 +497,8 @@ impl Eq for Ipv4Addr {
 
 impl hash::Hash for Ipv4Addr {
     fn hash<H: hash::Hasher>(&self, s: &mut H) {
-        self.0.s_addr.hash(s)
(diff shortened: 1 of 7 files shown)


asomers:
Good catch.

pusateri:
Ok, I think this is ready to merge. Also, if you want to add NetBSD to CI, that would be cool.

PR #995, opened 2018-12-08
Replace try! with ?

asomers:
try! is not available in Rust 2018.  It would be premature to convert the entire project to Rust 2018, since that would bump the minimum compiler to 1.31.0.  But his change will help us when we do convert it eventually.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -139,17 +139,17 @@ libc_bitflags!(
 );
 
 pub fn open<P: ?Sized + NixPath>(path: &P, oflag: OFlag, mode: Mode) -> Result<RawFd> {
-    let fd = try!(path.with_nix_path(|cstr| {
+    let fd = path.with_nix_path(|cstr| {
         unsafe { libc::open(cstr.as_ptr(), oflag.bits(), mode.bits() as c_uint) }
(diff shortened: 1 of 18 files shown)


Susurrus:
LGTM. However, what's the point of doing this now without a test checking that new PRs don't reintroduce these? I'd think it'd be easier if we moved to rust2018 in a concerted effort once we're willing to bump our Rust requirement to 1.31.

asomers:
New PRs could absolutely break it.  But I had to do the work of building with Rust 2018 in order to make sure that Nix wouldn't cause any problems for dependent 2018 crates.  For example, another crate of mine had a public `try` method, so it couldn't easily be used by a 2018 crate.  I didn't find any problems, but since I had to make all of those changes anyway I figured I would push them.

Susurrus:
I thought that there were no implications on other crates when setting the Rust edition? I guess that's not the case?

asomers:
Technically there aren't.  But if the _consuming_ crate wants to use Rust 2018, then it would have an awkward time calling a Nix method called `try` or `async`.  It would have to call them like `foo.r#try()`, or something like that.  I wanted to make sure we would have no such pitfalls.

PR #997, opened 2018-12-11
Implement symlinkat

oblique:


asomers:
Does this `symlink` have any advantages over [std::os::unix::fs::symlink](https://doc.rust-lang.org/std/os/unix/fs/fn.symlink.html)?  If not, then I don't think we should provide a redundant binding.

oblique:
No. However it is convenient when someone wants to write a function that returns `nix::Error`.

asomers:
I don't think that's important enough to justify the bloat.  Please remove `symlink`.  Keep `symlinkat` though; that's useful.

oblique:
Ready

PR #1005, opened 2019-01-06
Make ScmTimestamp's doc test more robust

asomers:
The old test made assumptions about the responsiveness of the test
environment.  The new test does not, and it's simpler too.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -480,52 +480,41 @@ pub enum ControlMessage<'a> {
     /// use std::time::*;
     ///
     /// // Set up
-    /// let message1 = "Ohayō!".as_bytes();
-    /// let message2 = "Jā ne".as_bytes();
-    /// let in_socket = socket(AddressFamily::Inet, SockType::Datagram, SockFlag::empty(), None).unwrap();
+    /// let message = "Ohayō!".as_bytes();
+    /// let in_socket = socket(
+    ///     AddressFamily::Inet,
+    ///     SockType::Datagram,
+    ///     SockFlag::empty(),
+    ///     None).unwrap();
     /// setsockopt(in_socket, sockopt::ReceiveTimestamp, &true).unwrap();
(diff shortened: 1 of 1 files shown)


asomers:
cc @pusateri @Wolvereness

Wolvereness:
You've missed one of the fundamental assumptions for whether or not this API is functional in the target environment. The point of this particular method of time-stamping is that the OS is assigning it, and it's independent of when the caller finally reads the packets.

If the test environment isn't "responsive", then the API probably doesn't work, as it's fundamentally supposed to be a method of time-stamping in that particular environment.

The previous test sends two packets, 250ms apart, and assumes the maximum system discrepancy is less than 5ms. If this treshold is problematic, then I suggest simply increasing it, or if a target system is giving responses with anywhere near the maximum 250ms discrepancy, consider the API to not be functional on that target.

Wolvereness:
I don't think *tests* should ever expect to handle non-monotonic time. The test is also not expecting the sleep to be 250ms; it's actually timing the delay itself, and then seeing if the delay between packets is within 5ms of what was measured. So, if the system time changed, it should be accounted for unless it was between the packet send and the timestamp get.

Are you getting the output for the tests? How off is it?

asomers:
Well, the old test was measuring _a_ delay.  But it still didn't account for the time taken up by the `sendmsg` calls themselves.  They aren't instantaneous.  Nor does the received timestamp necessarily get recorded during the `sendmsg` call.  That's an implementation detail.  It could be recorded when the data is copied from the sending socket's buffer to the receiving socket's buffer, and that could happen at any point before `recvmsg` returns.  The discrepancy, in one failing case, was 13 ms.

Wolvereness:
I suggest just changing it from 5ms to 50ms then, and one extra check for the initial sleep to be at least 200ms.

asomers:
Increasing the time limit wouldn't remove the test's sensitivity to the environment's responsiveness.  It would just loosen it a bit.  The new test should work even on the most unresponsive systems, even if the process gets paused for an hour.  It only requires that the clock be monotonic.

Issue #903, opened 2018-05-18
Missing bindings for `fchown`, `lchown`, and the `chmod` family


SolraBizna:
There are occasional situations where `fchown` is preferable to `chown`. The usefulness of `chmod` and its siblings is more self-evident.

Implementing `fchown` and `lchown` at least should be pretty easy, since nearly all the logic would be the same as for `chown`.

Edit: When documenting `fchown`/`fchmod`, it ought to be mentioned that they can't be used to portably change the permissions on a UNIX domain socket. (A lot of people make the mistake of assuming they can, including me ten minutes ago.)

asomers:
Do you want to submit a PR for this?  It sounds useful to me.

jbaublitz:
I see this is still open. I do in fact need this functionality and am happy to put together a PR if this hasn't been done yet. It looks like the PR referenced at the bottom of the issue here was never merged.

SteveLauC:
* fchown is added
* lchown is not because you can use fchownat
* chmod is added
* fchmodat is added and thus we don't need lchmod

Close this issue as it is complete

PR #1007, opened 2019-01-07
Issue #903: Adds fchown and chmod wrappers.

edwinbdr:
Does not add lchown, because it is a subpart of fchownat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#972](https://github.com/nix-rust/nix/pull/972))
 - Added `symlinkat` wrapper.
   ([#997](https://github.com/nix-rust/nix/pull/997))
+- Added `ptrace::{getregs, setregs}`.
+  ([#1010](https://github.com/nix-rust/nix/pull/1010))
 
 ### Changed
 ### Fixed
@@ -46,7 +48,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added a `fchownat` wrapper.
   ([#955](https://github.com/nix-rust/nix/pull/955))
 - Added support for `ptrace` on BSD operating systems ([#949](https://github.com/nix-rust/nix/pull/949))
-- Added `ptrace` functions for reads and writes to tracee memory and ptrace kill 
+- Added `ptrace` functions for reads and writes to tracee memory and ptrace kill
   ([#949](https://github.com/nix-rust/nix/pull/949)) ([#958](https://github.com/nix-rust/nix/pull/958))
 - Added a `acct` wrapper module for enabling and disabling process accounting
   ([#952](https://github.com/nix-rust/nix/pull/952))
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -9,6 +9,12 @@ use sys::signal::Signal;
 
 pub type AddressType = *mut ::libc::c_void;
 
+#[cfg(all(target_os = "linux",
+          any(target_arch = "x86_64",
+              target_arch = "x86"),
+          target_env = "gnu"))]
+use libc::user_regs_struct;
+
 cfg_if! {
     if #[cfg(any(all(target_os = "linux", target_arch = "s390x"),
                  all(target_os = "linux", target_env = "gnu")))] {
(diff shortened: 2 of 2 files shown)


edwinbdr:
Fix for https://github.com/nix-rust/nix/issues/903

src/sys/stat.rs
```diff
@@ -123,8 +123,32 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
     Ok(dst)
 }
 
+/// The chmod() function shall change S_ISUID, S_ISGID, S_ISVTX,
+/// and the file permission bits of the file named by the pathname
+/// pointed to by the path argument to the corresponding bits in the
+/// mode argument.
+///
+/// The fchmod() call cannot be used to portably change the
```
asomers: s/fchmod/chmod

src/sys/stat.rs
```diff
@@ -123,8 +123,32 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
     Ok(dst)
 }
 
+/// The chmod() function shall change S_ISUID, S_ISGID, S_ISVTX,
```
asomers: Please put identifiers in backticks

src/sys/stat.rs
```diff
@@ -123,8 +123,32 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
...
+/// pointed to by the path argument to the corresponding bits in the
+/// mode argument.
+///
+/// The fchmod() call cannot be used to portably change the
+/// permission bits of a file descriptor of a UNIX domain socket.
+///
+/// # References
+///
+/// [chmod(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html).
+pub fn chmod<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
```
asomers: Does directly binding `chmod(2)` have any advantages over `std::fs::set_permissions`?  It looks like that does the same thing.

src/unistd.rs
```diff
@@ -612,6 +612,24 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
...
+/// `owner` (user) and `group` (see
+/// [fchown(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chown.html)).
+///
+/// The owner/group for the provided file descriptor will not be modified if `None` is
+/// provided for that argument.  Ownership change will be attempted for the path
+/// only if `Some` owner/group is provided.
+///
+/// The fchown() call cannot be used to portably change the ownership
+/// of a UNIX domain socket.
+#[inline]
```
asomers: No need to add `#[inline]`

src/unistd.rs
```diff
@@ -612,6 +612,24 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
...
 
+/// Change the ownership of the file descriptor `fd` to be owned by the specified
+/// `owner` (user) and `group` (see
+/// [fchown(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chown.html)).
+///
+/// The owner/group for the provided file descriptor will not be modified if `None` is
+/// provided for that argument.  Ownership change will be attempted for the path
+/// only if `Some` owner/group is provided.
+///
+/// The fchown() call cannot be used to portably change the ownership
```
asomers: What exactly is the problem with unix domain sockets?  Do you have a reference?

src/sys/stat.rs
edwinbdr: done

src/sys/stat.rs
edwinbdr: done

src/sys/stat.rs
edwinbdr: difference is that this uses the Mode flags, instead of the octal bits, which could be more descriptive. Still, it seems to offer similar functionality. Skip this method, or include it for completeness?

src/unistd.rs
edwinbdr: I don't. It's in the original request for this feature. I'll look it up...

edwinbdr: https://lkml.org/lkml/2004/11/1/84

src/sys/stat.rs
asomers: Better not to bloat Nix if the method has no advantage.  Just ensure that Nix's `Mode` implements `Into<u32>` so it can be used with `std::fs::set_permissions`.

src/unistd.rs
asomers: On the BSDs, OSX, and Illumos that returns `EINVAL`.  Only on Linux does it silently fail.  In any case I think this detail is too obscure to warrant mentioning in Nix's documentation.

src/sys/stat.rs
edwinbdr: Removed

src/unistd.rs
edwinbdr: Removed

src/unistd.rs
edwinbdr: Comment is removed

PR #1008, opened 2019-01-08
Adds implementation of copy_file_range()

edwinbdr:
Also adds CHANGELOG.md and tests

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -246,6 +246,9 @@ pub enum FcntlArg<'a> {
 }
 pub use self::FcntlArg::*;
 
+#[cfg(any(target_os = "android", target_os = "linux"))]
+use std::ptr;
+
 // TODO: Figure out how to handle value fcntl returns
 pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
     let res = unsafe {
@@ -305,6 +308,36 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Copy a range of data from one file to another
+///
+/// The `copy_file_range` system call performs an in-kernel copy between
+/// file descriptor `fd_in` and `fd_out` without the additional cost of transferring data
+/// from the kernel to user space and then back into the kernel.  It
+/// copies up to `len` bytes of data from file descriptor `fd_in` to file
+/// descriptor `fd_out`, overwriting any data that exists within the
+/// requested range of the target file.
+///
+/// If the `off_in` and/or `off_out` arguments are used, the values
+/// will be mutated to reflect the new position within the file after
+/// copying. If they are not used, the relevant filedescriptors will be seeked
+/// to the new position.
+///
+/// On completion the number of bytes actually copied will be returned.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub fn copy_file_range(
+    fd_in: RawFd, off_in: Option<&mut i64>,
+    fd_out: RawFd, off_out: Option<&mut i64>,
+    len: usize) -> Result<usize> {
+
+    let off_in = off_in.map(|offset| offset as *mut _).unwrap_or(ptr::null_mut());
+    let off_out = off_out.map(|offset| offset as *mut _).unwrap_or(ptr::null_mut());
+
+    let ret = unsafe {
+        libc::syscall(libc::SYS_copy_file_range, fd_in, off_in, fd_out, off_out, len, 0)
+    };
+    Errno::result(ret).map(|r| r as usize)
+}
+
 #[cfg(any(target_os = "android", target_os = "linux"))]
 libc_bitflags! {
     /// Additional flags to `splice` and friends.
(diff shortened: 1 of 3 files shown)


edwinbdr:
This fixes https://github.com/nix-rust/nix/issues/941

asomers:
This doesn't even compile.  Did you forget to commit something?

src/fcntl.rs
```diff
@@ -305,6 +308,43 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+/// descriptor `fd_out`, overwriting any data that exists within the
+/// requested range of the target file.
+///
+/// The flags argument is currently unused, and is provided to allow for future
+/// extensions and currently must be set to `CopyFileRangeFlags::empty()`.
+///
+/// On completion the number of bytes actually copied will be returned.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub fn copy_file_range(
+    fd_in: RawFd, off_in: Option<&mut libc::loff_t>,
```
asomers: Why are the offsets provided as mutable references?  If this operation changes them, could you describe how in the doc comment?

Also, it's ugly to expose raw libc types in our public API.  In this case, I think it would be preferable for the arguments to be `&mut i64`.

src/fcntl.rs
```diff
@@ -305,6 +308,43 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+    fd_out: RawFd, off_out: Option<&mut libc::loff_t>,
+    len: usize,
+    _flags: CopyFileRangeFlags) -> Result<usize> {
+
+    let off_in = off_in.map(|offset| offset as *mut _).unwrap_or(ptr::null_mut());
+    let off_out = off_out.map(|offset| offset as *mut _).unwrap_or(ptr::null_mut());
+
+    let ret = unsafe {
+        libc::syscall(libc::SYS_copy_file_range, fd_in, off_in, fd_out, off_out, len, 0)
```
asomers: Shouldn't the last argument be `flags` instead of 0?

test/test_fcntl.rs
```diff
@@ -52,12 +52,42 @@ mod linux_android {
...
+
+        tmp1.write_all(CONTENTS).unwrap();
+        tmp1.flush().unwrap();
+
+        let mut from_offset: libc::loff_t = 3;
+        copy_file_range(tmp1.as_raw_fd(), Some(&mut from_offset),
+                                    tmp2.as_raw_fd(), None,
+                                    3, CopyFileRangeFlags::NONE).unwrap();
+
+        tmp2.sync_all().unwrap();
```
asomers: I the `sync_all` necessary?  I would think that the test should work fine without it.

src/fcntl.rs
```diff
@@ -305,6 +308,43 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+/// The flags argument is currently unused, and is provided to allow for future
+/// extensions and currently must be set to `CopyFileRangeFlags::empty()`.
+///
+/// On completion the number of bytes actually copied will be returned.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub fn copy_file_range(
+    fd_in: RawFd, off_in: Option<&mut libc::loff_t>,
+    fd_out: RawFd, off_out: Option<&mut libc::loff_t>,
+    len: usize,
+    _flags: CopyFileRangeFlags) -> Result<usize> {
```
asomers: I must admit I'm not enthusiastic about the `flags` argument.  It currently provides no benefit to the user, and there's no guarantee that Linux will ever use it.  But it does require a lot of typing.  Maybe we should omit it.  Or we could at least reduce the typing by changing the type to `Into<CopyFileRangeFlags>` so the user can simply supply `0`.  What do you think?

src/fcntl.rs
edwinbdr: Done

src/fcntl.rs
edwinbdr: Done (flags removed)

test/test_fcntl.rs
edwinbdr: Apparently not. I was trying to be overly cautious.

src/fcntl.rs
edwinbdr: Removed the flags.

src/fcntl.rs
asomers: I agree.  Returning a struct would be more Rusty than using mutable arguments.

edwinbdr: Hmm, it would be a bit of an overkill and possibly also a memory management issue to return new structs. Purpose of this method is to quickly copy large parts of files incrementally, so it will most likely be called in an iterative fashion with feedback. Agree to leave it the way it is now, or do we want some extra helper methods?

asomers: I suppose it would be ok to leave as is.  Returning a structure with offsets could be up to 256 bits.

asomers:
The tests are failing because Travis's Linux kernel is too old.  We could switch Linux testing from Travis to Cirrus, but I don't want to hold up this PR just for that.  Why don't you just add an `#[ignore]` attribute for that test?  Also, you'll need to rebase your commits to remove that merge commit from your history.

test/test_fcntl.rs
```diff
@@ -49,15 +49,49 @@ fn test_readlink() {
...
+    /// 3 bytes at offset 3 (`bar`) to another empty file at offset 0. The
+    /// resulting file is read and should contain the contents `bar`.
+    /// The from_offset should be updated by the call to reflect
+    /// the 3 bytes read (6).
+    ///
+    /// Fix me: test is disabled for linux based builds, because Jenkins
+    /// Linux version is too old for `copy_file_range`. Should work
+    /// on netbsd build.
+    #[test]
+    #[cfg_attr(target_env = "linux", ignore)]
```
asomers: Linux is a target_os, not a target_env.  Also, you should leave a message with the ignore attribute.  And it's not going to work on a "netbsd build", because NetBSD doesn't have this syscall at all.  Finally, we use Travis, not Jenkins.

ArniDagur: Is there anything else that is holding this PR back?

asomers: Yes.  The OP needs to address my last comments, fix the CI failures, and rebase.

ArniDagur:
This PR can be closed, right?

asomers:
Yeah, since it was replaced by #1069 this PR can be closed.

PR #1010, opened 2019-01-09
Added `ptrace::{getregs, setregs}`

wpovell:
I needed `ptrace`'s `getregs` and `setregs` for a project I was working on, so I added them! First time contributor, so let me know if I should change / fix anything. Things work fine in my project which uses both functions, let me know if further testing is needed.


wpovell:
Not entirely sure why CI is failing; it seems like `user_regs_struct` can't be found on some platforms, even when wrapping the use with the same `cfg` used by `PTRACE_GETREGS`

asomers:
Should be easy to debug.  Just checkout the libc repository and have a look.  It's also possible that the libc repository has errors.  If you find one, just send libc a PR.

PR #1016, opened 2019-01-20
Added inotify bindings.

vdagonneau:
Hi !

I needed inotify bindings and noticed that nix did not have any so here is a PR to add it.

There was another PR from 2015 to add support for inotify that was never merged. I took some of the feedback and applied it here.

### DIFF
--- a/src/sys/inotify.rs
+++ b/src/sys/inotify.rs
@@ -1,0 +1,231 @@
+//! Monitoring API for filesystem events.
+//!
+//! Inotify is a Linux-only API to monitor filesystems events.
+//!
+//! For more documentation, please read [inotify(7)](http://man7.org/linux/man-pages/man7/inotify.7.html).
+//! 
+//! # Examples
+//!
+//! Monitor all events happening in directory "test":
+//! ```no_run
+//! # use nix::sys::inotify::{AddWatchFlags,InitFlags,Inotify};
+//! #
+//! // We create a new inotify instance.
+//! let instance = Inotify::init(InitFlags::empty()).unwrap();
+//!
+//! // We add a new watch on directory "test" for all events.
+//! let wd = instance.add_watch("test", AddWatchFlags::IN_ALL_EVENTS).unwrap();
+//!
+//! loop {
+//!     // We read from our inotify instance for events.
+//!     let events = instance.read_events().unwrap();
+//!     println!("Events: {:?}", events);
+//! }
+//! ```
+
+use libc;
+use libc::{
+    c_char,
+    c_int,
+    uint32_t
+};
+use std::ffi::{OsString,OsStr,CStr};
+use std::os::unix::ffi::OsStrExt;
+use std::mem::size_of;
+use std::os::unix::io::{RawFd,AsRawFd,FromRawFd};
+use unistd::read;
+use Result;
+use NixPath;
+use errno::Errno;
+
+libc_bitflags! {
+    /// Configuration options for [`inotify_add_watch`](fn.inotify_add_watch.html).
+    pub struct AddWatchFlags: uint32_t {
+        IN_ACCESS;
+        IN_MODIFY;
+        IN_ATTRIB;
+        IN_CLOSE_WRITE;
+        IN_CLOSE_NOWRITE;
+        IN_OPEN;
+        IN_MOVED_FROM;
+        IN_MOVED_TO;
+        IN_CREATE;
+        IN_DELETE;
+        IN_DELETE_SELF;
+        IN_MOVE_SELF;
+
+        IN_UNMOUNT;
+        IN_Q_OVERFLOW;
+        IN_IGNORED;
+
+        IN_CLOSE;
+        IN_MOVE;
+
+        IN_ONLYDIR;
+        IN_DONT_FOLLOW;
+
+        IN_ISDIR;
+        IN_ONESHOT;
+        IN_ALL_EVENTS;
+    }
+}
+
+libc_bitflags! {
+    /// Configuration options for [`inotify_init1`](fn.inotify_init1.html).
+    pub struct InitFlags: c_int {
+        IN_CLOEXEC;
+        IN_NONBLOCK;
+    }
+}
+
+/// An inotify instance. This is also a file descriptor, you can feed it to
+/// other interfaces consuming file descriptors, epoll for example.
+#[derive(Debug, Clone, Copy)]
+pub struct Inotify {
+    fd: RawFd
+}
+
+/// This object is returned when you create a new watch on an inotify instance.
+/// It is then returned as part of an event once triggered. It allows you to
+/// know which watch triggered which event. 
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct WatchDescriptor {
+    wd: i32
+}
+
+/// A single inotify event.
+///
+/// For more documentation see, [inotify(7)](http://man7.org/linux/man-pages/man7/inotify.7.html).
+#[derive(Debug)]
+pub struct InotifyEvent {
+    /// Watch descriptor. This field corresponds to the watch descriptor you
+    /// were issued when calling add_watch. It allows you to know which watch
+    /// this event comes from. 
+    pub wd: WatchDescriptor,
+    /// Event mask. This field is a bitfield describing the exact event that
+    /// occured.
+    pub mask: AddWatchFlags,
+    /// This cookie is a number that allows you to connect related events. For
+    /// now only IN_MOVED_FROM and IN_MOVED_TO can be connected. 
+    pub cookie: u32,
+    /// Filename. This field exists only if the event was triggered for a file
+    /// inside the watched directory.
+    pub name: Option<OsString>
+}
+
+impl Inotify {
+    /// Initialize a new inotify instance.
+    ///
+    /// Returns a Result containing an inotify instance.
+    ///
+    /// For more information see, [inotify_init(2)](http://man7.org/linux/man-pages/man2/inotify_init.2.html).
+    pub fn init(flags: InitFlags) -> Result<Inotify> {
+        let res = Errno::result(unsafe {
+            libc::inotify_init1(flags.bits())
+        });
+
+        res.map(|fd| Inotify { fd })
+    }
+
+    /// Adds a new watch on the target file or directory. 
+    ///
+    /// Returns a watch descriptor. This is not a File Descriptor! 
+    ///
+    /// For more information see, [inotify_add_watch(2)](http://man7.org/linux/man-pages/man2/inotify_add_watch.2.html).
+    pub fn add_watch<P: ?Sized + NixPath>(&self,
+                                          path: &P,
+                                          mask: AddWatchFlags) 
+                                            -> Result<WatchDescriptor>
+    {
+        let res = path.with_nix_path(|cstr| {
+            unsafe {
+                libc::inotify_add_watch(self.fd, cstr.as_ptr(), mask.bits())
+            }
+        })?;
+
+        Errno::result(res).map(|wd| WatchDescriptor { wd })
+    }
+
+    /// Removes an existing watch using the watch descriptor returned by
+    /// inotify_add_watch.
+    ///
+    /// Returns an EINVAL error if the watch descriptor is invalid.
+    ///
+    /// For more information see, [inotify_rm_watch(2)](http://man7.org/linux/man-pages/man2/inotify_rm_watch.2.html).
+    #[cfg(target_os = "linux")]
+    pub fn rm_watch(&self, wd: WatchDescriptor) -> Result<()> {
+        let res = unsafe { libc::inotify_rm_watch(self.fd, wd.wd) };
+
+        Errno::result(res).map(drop)
+    }
+
+    #[cfg(target_os = "android")]
+    pub fn rm_watch(&self, wd: WatchDescriptor) -> Result<()> {
+        let res = unsafe { libc::inotify_rm_watch(self.fd, wd.wd as u32) };
+
+        Errno::result(res).map(drop)
+    }
+
+    /// Reads a collection of events from the inotify file descriptor. This call
+    /// can either be blocking or non blocking depending on whether IN_NONBLOCK
+    /// was set at initialization. 
+    /// 
+    /// Returns as many events as available. If the call was non blocking and no
+    /// events could be read then the EAGAIN error is returned.
+    pub fn read_events(&self) -> Result<Vec<InotifyEvent>> {
+        let header_size = size_of::<libc::inotify_event>();
+        let mut buffer = [0u8; 4096];
+        let mut events = Vec::new();
+        let mut offset = 0;
+
+        let nread = read(self.fd, &mut buffer)?;
+
+        while (nread - offset) >= header_size {
+            let event = unsafe {
+                &*(
+                    buffer
+                        .as_ptr()
+                        .offset(offset as isize) as *const libc::inotify_event
+                )
+            };
+
+            let name = match event.len {
+                0 => None,
+                _ => {
+                    let ptr = unsafe { 
+                        buffer
+                            .as_ptr()
+                            .offset(offset as isize + header_size as isize)
+                            as *const c_char
+                    };
+                    let cstr = unsafe { CStr::from_ptr(ptr) };
+                
+                    Some(OsStr::from_bytes(cstr.to_bytes()).to_owned())
+                }
+            };
+
+            events.push(InotifyEvent {
+                wd: WatchDescriptor { wd: event.wd },
+                mask: AddWatchFlags::from_bits_truncate(event.mask),
+                cookie: event.cookie,
+                name
+            });
+
+            offset += header_size + event.len as usize;
+        }
+
+        Ok(events)
+    }
+}
+
+impl AsRawFd for Inotify {
+    fn as_raw_fd(&self) -> RawFd {
+        self.fd
+    }
+}
+
+impl FromRawFd for Inotify {
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        Inotify { fd }
+    }
+}
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -88,3 +88,6 @@ pub mod uio;
 pub mod utsname;
 
 pub mod wait;
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub mod inotify;
--- a/test/sys/test_inotify.rs
+++ b/test/sys/test_inotify.rs
@@ -1,0 +1,65 @@
+use nix::sys::inotify::{AddWatchFlags,InitFlags,Inotify};
+use nix::Error;
+use nix::errno::Errno;
+use tempfile;
+use std::ffi::OsString;
+use std::fs::{rename, File};
+
+#[test]
+pub fn test_inotify() {
+    let instance = Inotify::init(InitFlags::IN_NONBLOCK)
+        .unwrap();
+    let tempdir = tempfile::tempdir().unwrap();
+
+    instance.add_watch(tempdir.path(), AddWatchFlags::IN_ALL_EVENTS).unwrap();
+
+    let events = instance.read_events();
+    assert_eq!(events.unwrap_err(), Error::Sys(Errno::EAGAIN));
+
+    File::create(tempdir.path().join("test")).unwrap();
+
+    let events = instance.read_events().unwrap();
+    assert_eq!(events[0].name, Some(OsString::from("test")));
+}
+
+#[test]
+pub fn test_inotify_multi_events() {
+    let instance = Inotify::init(InitFlags::IN_NONBLOCK)
+        .unwrap();
+    let tempdir = tempfile::tempdir().unwrap();
+
+    instance.add_watch(tempdir.path(), AddWatchFlags::IN_ALL_EVENTS).unwrap();
+
+    let events = instance.read_events();
+    assert_eq!(events.unwrap_err(), Error::Sys(Errno::EAGAIN));
+
+    File::create(tempdir.path().join("test")).unwrap();
+    rename(tempdir.path().join("test"), tempdir.path().join("test2")).unwrap();
+
+    // Now there should be 5 events in queue:
+    //   - IN_CREATE on test
+    //   - IN_OPEN on test
+    //   - IN_CLOSE_WRITE on test
+    //   - IN_MOVED_FROM on test with a cookie
+    //   - IN_MOVED_TO on test2 with the same cookie
+
+    let events = instance.read_events().unwrap();
+    assert_eq!(events.len(), 5);
+
+    assert_eq!(events[0].mask, AddWatchFlags::IN_CREATE);
+    assert_eq!(events[0].name, Some(OsString::from("test")));
+
+    assert_eq!(events[1].mask, AddWatchFlags::IN_OPEN);
+    assert_eq!(events[1].name, Some(OsString::from("test")));
+
+    assert_eq!(events[2].mask, AddWatchFlags::IN_CLOSE_WRITE);
+    assert_eq!(events[2].name, Some(OsString::from("test")));
+
+    assert_eq!(events[3].mask, AddWatchFlags::IN_MOVED_FROM);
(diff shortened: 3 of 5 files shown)


src/sys/inotify.rs
```diff
@@ -0,0 +1,120 @@
+use libc::{
+    c_int,
+    uint32_t
+};
+use std::os::unix::io::RawFd;
+use {Error, Result};
+use NixPath;
+use errno::Errno;
+
+pub mod ffi {
```
asomers: Nix no longer uses FFI directly.  C symbols should be placed in the libc crate instead.  Then Nix consumes libc.

src/sys/inotify.rs
```diff
@@ -0,0 +1,120 @@
...
+
+bitflags! {
+    pub struct InotifyInitFlags: c_int {
+        const IN_CLOEXEC    = 0o02000000;
+        const IN_NONBLOCK   = 0o00004000;
+    }
+}
+
+#[inline]
+pub fn inotify_init() -> Result<RawFd> {
```
asomers: It looks like `inotify_init1` superseded `inotify_init`.  If so, then Nix should bother with `inotify_init`.  Just use the newer interface.

src/sys/inotify.rs
```diff
@@ -0,0 +1,120 @@
...
+    let res = unsafe { ffi::inotify_init() };
+
+    if res < 0 {
+        return Err(Error::Sys(Errno::last()));
+    }
+
+    Ok(res)
+}
+
+#[inline]
```
asomers: Don't use `#[inline]`.  The compiler can figure out on its own when it makes sense to inline stuff.

src/sys/inotify.rs
```diff
@@ -0,0 +1,120 @@
...
+pub fn inotify_init1(flags: InotifyInitFlags) -> Result<RawFd> {
+    let res = unsafe { ffi::inotify_init1(flags.bits()) };
+
+    if res < 0 {
+        return Err(Error::Sys(Errno::last()));
```
asomers: You can do `Errno::result(res)` instead.  That will automatically take care of errno.

src/sys/inotify.rs
vdagonneau: I decided to include it after seeing that Nix exposes both `epoll_create` and `epoll_create1`. Isn't it the same situation ? If I'm not mistaken `epoll_create` was superseded by `epoll_create1` as well.

vdagonneau:
I updated the code taking your remarks into account, I also added some documentation and testing.

src/sys/inotify.rs
asomers: If `inotify_create1` requires a very recent kernel, then that would be a reason to keep `inotify_create` around.  Otherwise, there's probably not much point.

src/sys/inotify.rs
asomers: Yeah, Nix is an old create, and coding standards have evolved over time.  We haven't gone back and updated all of the old stuff.

test/sys/test_inotify.rs
```diff
@@ -0,0 +1,54 @@
...
+pub fn test_inotify() {
+	let ifd = inotify_init1(InotifyInitFlags::IN_NONBLOCK).unwrap();
+	let tempdir = tempfile::tempdir().unwrap();
+
+	let wd = inotify_add_watch(ifd, tempdir.path(), EventFlags::IN_ALL_EVENTS);
+	assert!(wd.is_ok());
+
+	let mut buffer = [0u8; 4096];
+
+	let mut _nread = read(ifd, &mut buffer);
```
asomers: Why the leading underscore on `_nread`?  That usually indicates a variable that will not be used.

asomers:
Submitting a PR to libc is easy.  Just checkout https://github.com/rust-lang/libc/ .  Copy the format for existing code.  The hardest part is sometimes figuring out which file to modify.  If inotify is the same on all Linux architectures, then your changes should go into src/unix/notbsd/linux/mod.rs.

src/sys/inotify.rs
```diff
@@ -137,6 +117,43 @@ bitflags! {
...
+
+    read(fd, &mut buffer)?;
+
+    let event_size = std::mem::size_of::<ffi::inotify_event>();
+    assert!(buffer.len() >= event_size);
+
+    let event = buffer.as_ptr() as *const ffi::inotify_event;
+    let event = unsafe { &*event };
+
+    let bytes_left_in_buffer = buffer.len() - event_size;
```
asomers: Instead of doing this trick, you could just make the FFI `inotify_event` struct's last field be an unsized slice.  That's what the Linux headers do.  Like this:
```rust
struct inotify_event<T: ?Sized> {
    pub stuff: Stuff,
    pub name: T
}
let header_size = mem::size_of::<inotify_event<[u8; 0]>>();
let name_size = buffer.len() - header_size;
let event = buffer.as_ptr() as *const inotify_event<[u8; name_size]>()
```

src/sys/inotify.rs
```diff
@@ -137,6 +117,43 @@ bitflags! {
...
+    assert!(bytes_left_in_buffer >= event.len as usize);
+
+    let bytes_consumed = event_size + event.len as usize;
+    let name = &buffer[event_size..bytes_consumed];
+    let name = name
+        .splitn(2, |b| b == &0u8)
```
asomers: Don't implement your own ASCII string parsing code.  Just use [`OsStrExt::from_bytes`](https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html#tymethod.from_bytes)

src/sys/inotify.rs
```diff
@@ -137,6 +117,43 @@ bitflags! {
     }
 }
 
+#[derive(Debug)]
+pub struct InotifyEvent {
+    pub wd: i32,
```
asomers: What do these fields mean?  If they're going to be public, then they should be documented, especially the enigmatic `wd` and `cookie`.

src/sys/inotify.rs
vdagonneau: I'm not sure I understand your example correctly, `name_size` is not constant therefore `[u8; name_size]` is invalid, no ?

I do understand the general meaning though.

asomers: It would be invalid to construct a `[u8; name_size]` and store it on the stack.  However, a pointer to an array of unknown size *does* have a constant size; it's a pointer.  So the expression is valid.

vdagonneau: Ok I think I am missing something because this doesn't compile. 

I'm sorry if my level in Rust is not great; I'm coming from C.

I made a small rust playground with the smallest code necessary to reproduce:

[](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=45520d14233c767d2fcb0b22b0e730b0)

asomers: Well, I guess I was wrong.  In that case, you'll probably have to use slice::from_raw_parts.

vdagonneau: So, I refactored some of that code but I still end up with a problem. The name_size is actually aligned which means that my string ends up with a lot of trailing NULLs in some cases. The original code for trimming it was ugly : 
```
let name = name.splitn(2, |b| b == &0u8).next().unwrap();
```
However I can't find a much better way. `OsStr::from_bytes` doesn't work here.

I found this thread on StackOverflow, but it is arguably uglier: [https://stackoverflow.com/questions/42066381/how-to-get-a-str-from-a-nul-terminated-byte-slice-if-the-nul-terminator-isnt](https://stackoverflow.com/questions/42066381/how-to-get-a-str-from-a-nul-terminated-byte-slice-if-the-nul-terminator-isnt)

asomers: What's wrong with `OsStrExt::from_bytes`?

vdagonneau: `OsStrExt::from_bytes` does not take of NULL bytes. It happily accepts to have NULL bytes in the middle of the resulting `OsStr`. In my case, I end up with trailing NULLs.

asomers: Oh, I forgot that it requires you to take care of that yourself.  Well, there's always `CStr::from_ptr`.

vdagonneau: Ok fixed it with CStr::from_ptr.

src/sys/inotify.rs
```diff
@@ -5,17 +5,17 @@
 //!
 //! Inotify is a Linux-only API to monitor filesystems events.
 //!
-//! For more documentation, please read [the inotify(7) man page.](http://man7.org/linux/man-pages/man7/inotify.7.html)
+//! For more documentation, please read [inotify(7).](http://man7.org/linux/man-pages/man7/inotify.7.html)
```
asomers: Sorry to nitpick, but the trailing '.' should not be part of the link.

src/sys/inotify.rs
```diff
@@ -5,17 +5,17 @@
...
 //! 
 //! Example
 //! =======
 //!
 //! * Monitor all events happening in directory "test":
 //! ```rust,no_run
-//! use nix::sys::inotify::{EventFlags,InotifyInitFlags,InotifyInstance}
+//! use nix::sys::inotify::{EventFlags,InotifyInitFlags,Inotify};
```
asomers: In doc tests, you can prefix the use lines with `#`.  That way they won't show up in the formatted docs.

src/sys/inotify.rs
```diff
@@ -5,17 +5,17 @@
...
 //! Inotify is a Linux-only API to monitor filesystems events.
 //!
-//! For more documentation, please read [the inotify(7) man page.](http://man7.org/linux/man-pages/man7/inotify.7.html)
+//! For more documentation, please read [inotify(7).](http://man7.org/linux/man-pages/man7/inotify.7.html)
 //! 
 //! Example
 //! =======
 //!
 //! * Monitor all events happening in directory "test":
 //! ```rust,no_run
```
asomers: Pro tip: rustdoc is smart enough to figure out that the syntax is Rust; you don't need to tell it.  It's ok if you want to leave that in there, but you don't need to.

src/sys/inotify.rs
```diff
@@ -5,17 +5,17 @@
...
-//! use nix::sys::inotify::{EventFlags,InotifyInitFlags,InotifyInstance}
+//! use nix::sys::inotify::{EventFlags,InotifyInitFlags,Inotify};
 //!
 //! // We create a new inotify instance.
-//! let instance = InotifyInstance::init(InotifyInitFlags::empty())
+//! let instance = Inotify::init(InotifyInitFlags::empty())
 //!     .expect("inotify_init failed");
```
asomers: For the doc tests, you can replace the `.expect` with a simple `.unwrap`.  It gets the point across.

src/sys/inotify.rs
```diff
@@ -27,7 +27,7 @@
 //!     let event = instance.read_event()
 //!         .expect("Failed reading inotify events");
 //!     
-//!     println!("wd: {}, mask: {:?}, cookie: {}, name: {:?}",
+//!     println!("wd: {:?}, mask: {:?}, cookie: {}, name: {:?}",
```
asomers: Since you derive `Debug` for `InotifyEvent`, you can simply do `println!("{:?}", event);` instead of printing each field individually.

src/sys/inotify.rs
```diff
@@ -125,18 +125,26 @@ bitflags! {
...
 #[derive(Debug, Clone, Copy)]
-pub struct InotifyInstance {
+pub struct Inotify {
     fd: RawFd
 }
 
+/// This object is returned when you create a new watch on an inotify instance.
+/// It is then returned as part of an event once triggered. It allows you to
+/// know which watch triggered which event. 
+#[derive(Debug, Clone, Copy)]
```
asomers: You should derive `Hash`, `PartialEq`, `Eq`, `PartialOrd`, and `Ord` for `WatchDescriptor`.  That way they can be stored in all the standard collections.

src/sys/inotify.rs
```diff
@@ -145,50 +153,47 @@ pub struct InotifyEvent {
     pub cookie: u32,
     /// Filename. This field exists only if the event was triggered for a file
     /// inside the watched directory.
-    pub name: OsString
+    pub name: Option<CString>
```
asomers: I'm a little hesitant about putting `CString` in a public API.  `CString` is harder to work with than other types, like `String`, `OsString`, or even `PathBuf`.

src/sys/inotify.rs
```diff
@@ -213,33 +218,36 @@ impl InotifyInstance {
...
+            0 => None,
+            _ => {
+                let ptr = unsafe { buffer.as_ptr().offset(header_size as isize) as *const c_char };
+                let cstr = unsafe { CStr::from_ptr(ptr) };
+            
+                Some(cstr.to_owned())
+            }
+        };
 
         // TODO: Get rid of the trailing NULLs due to alignement
```
asomers: Is this comment obsolete now?

src/sys/inotify.rs
```diff
@@ -213,33 +218,36 @@ impl InotifyInstance {
...
 
-        let bytes_left_in_buffer = buffer.len() - header_size;
-        assert!(bytes_left_in_buffer >= event.len as usize);
-
-        let bytes_consumed = header_size + event.len as usize;
-        let name = &buffer[header_size..bytes_consumed];
+        let name = match event.len {
+            0 => None,
+            _ => {
+                let ptr = unsafe { buffer.as_ptr().offset(header_size as isize) as *const c_char };
```
asomers: According to the man page, `read` can return more than one event at a time.  If that happens, your code is going to ignore events past the first.

src/sys/inotify.rs
```diff
@@ -213,33 +218,36 @@ impl InotifyInstance {
...
         // TODO: Get rid of the trailing NULLs due to alignement
         // https://stackoverflow.com/questions/42066381/how-to-get-a-str-from-a-nul-terminated-byte-slice-if-the-nul-terminator-isnt
 
         Ok(InotifyEvent {
-            wd: event.wd,
+            wd: WatchDescriptor { wd: event.wd },
             mask: EventFlags::from_bits_truncate(event.mask),
             cookie: event.cookie,
-            name: OsStr::from_bytes(&name).to_os_string()
+            name
```
asomers: Trailing space

src/sys/inotify.rs
vdagonneau: I agree with that. However no other type fits particularly well. I could maybe return a `String` using `CStr::to_string_lossy` but we would lose data. There is no easy path from `CString` to `OsString` and I'm not sure `PathBuf` is the right structure to use.

src/sys/inotify.rs
```diff
@@ -191,52 +185,61 @@ impl Inotify {
...
-    /// For more information see, [inotify_rm_watch(2).](http://man7.org/linux/man-pages/man2/inotify_rm_watch.2.html)
+    /// For more information see, [inotify_rm_watch(2)](http://man7.org/linux/man-pages/man2/inotify_rm_watch.2.html).
     pub fn rm_watch(&self, wd: WatchDescriptor) -> Result<()> {
         let res = unsafe { ffi::inotify_rm_watch(self.fd, wd.wd) };
 
         Errno::result(res).map(drop)
     }
 
-    /// Reads a new event from the inotify file descriptor. This call can either
+    /// Reads ne events from the inotify file descriptor. This call can either
```
asomers: Reads what now?

src/sys/inotify.rs
```diff
@@ -191,52 +185,61 @@ impl Inotify {
...
-        let event = unsafe {
-            &*(
-                buffer.as_ptr() as *const ffi::inotify_event
-            )
-        };
+            let name = match event.len {
+                0 => None,
+                _ => {
+                    let ptr = unsafe { 
+                        buffer
```
asomers: This pointer arithmetic is wrong for events past the first because you don't update buffer.

src/sys/inotify.rs
```diff
@@ -191,52 +185,61 @@ impl Inotify {
...
+            events.push(InotifyEvent {
+                wd: WatchDescriptor { wd: event.wd },
+                mask: EventFlags::from_bits_truncate(event.mask),
+                cookie: event.cookie,
+                name
+            });
 
-        // TODO: Get rid of the trailing NULLs due to alignement
-        // https://stackoverflow.com/questions/42066381/how-to-get-a-str-from-a-nul-terminated-byte-slice-if-the-nul-terminator-isnt
+            offset = offset + header_size + event.len as usize;
```
asomers: It would be more idiomatic to use the `+=` operator here.

src/sys/inotify.rs
vdagonneau: Replacing `header_size as isize` by `offset as isize + header_size as isize` should do it right ? Got my arithmetic mixed up. I did not even notice in my tests.

vdagonneau:
I removed all the ffi code from the pull request and submitted a pull request in the libc.

asomers:
Could you please link this PR to the libc PR?

vdagonneau:
Hi, I'm sorry for the delay, I was on vacation with very limited access to internet. The libc PR is at: https://github.com/rust-lang/libc/pull/1230.

vdagonneau:
The code I needed has been merged into the libc.

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
+//! Monitoring API for filesystem events.
+//!
+//! Documentation
```
asomers: "Documentation" is redundant.  You can remove it.

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
...
+//! =============
+//!
+//! Inotify is a Linux-only API to monitor filesystems events.
+//!
+//! For more documentation, please read [inotify(7)](http://man7.org/linux/man-pages/man7/inotify.7.html).
+//! 
+//! Example
```
asomers: For consistency, use `# Examples` instead of plain `Example` with an underline.

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
...
+//! =======
+//!
+//! * Monitor all events happening in directory "test":
```
asomers: You don't need a bullet here.

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
...
+use std::mem::size_of;
+use std::os::unix::io::{RawFd,AsRawFd,FromRawFd};
+use unistd::read;
+use Result;
+use NixPath;
+use errno::Errno;
+
+libc_bitflags! {
+    /// Configuration options for [`inotify_add_watch`](fn.inotify_add_watch.html).
+    pub struct EventFlags: uint32_t {
```
asomers: According to our guidelines in CONVENTIONS.md, this struct should be named `InFlags`.  However, the guidelines say that the `InotifyInitFlags` should be named the same way.  How about renaming them `AddWatchFlags` and `InitFlags`?  That way it's clear which functions they should be used with, and they don't repeat the "inotify" part of the module name.

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
...
+        IN_ACCESS;
+		IN_MODIFY;
```
asomers: Why two levels of indentation?  Is there a tabs/spaces problem here?

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
...
+/// This object is returned when you create a new watch on an inotify instance.
+/// It is then returned as part of an event once triggered. It allows you to
+/// know which watch triggered which event. 
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct WatchDescriptor {
+    wd: i32
+}
+
+/// A single inotify event.
+/// For more documentation see, [inotify(7)](http://man7.org/linux/man-pages/man7/inotify.7.html).
```
asomers: You should probably put a blank line between these two so the link will be hidden in the collapsed html documentation.

src/sys/inotify.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        Errno::result(res).map(|wd| WatchDescriptor { wd })
+    }
+
+    /// Removes an existing watch using the watch descriptor returned by
+    /// inotify_add_watch.
+    ///
+    /// Returns an EINVAL error if the watch descriptor is invalid.
+    ///
+    /// For more information see, [inotify_rm_watch(2)](http://man7.org/linux/man-pages/man2/inotify_rm_watch.2.html).
```
asomers: 80 columns per line, please.  But don't split the link, of course.

src/sys/mod.rs
```diff
@@ -88,3 +88,6 @@ pub mod uio;
 pub mod utsname;
 
 pub mod wait;
+
+#[cfg(target_os = "linux")]
```
asomers: Any reason not to include Android, too?

test/sys/test_inotify.rs
```diff
@@ -0,0 +1,66 @@
...
+use std::ffi::OsString;
+use std::fs::{rename, File};
+
+#[test]
+pub fn test_inotify() {
+	let instance = Inotify::init(InotifyInitFlags::IN_NONBLOCK)
+		.unwrap();
+	let tempdir = tempfile::tempdir().unwrap();
+
+	instance.add_watch(tempdir.path(), EventFlags::IN_ALL_EVENTS).unwrap();
```
asomers: No hard tabs.  Use four spaces for each level of indentation.

src/sys/mod.rs
vdagonneau: No, I was concerned that android might not have some of the flags introduced in 4.17 (IN_MASK_CREATE and IN_MASK_ADD) but the libc doesn't either.

src/sys/inotify.rs
vdagonneau: I like the name EventFlags as the flags are used not only in `add_watch` but also as a result in the struct `InotifyEvent`.

asomers:
The build is failing on Android because of undefined symbols.  You'll have to figure out whether Android actually has those symbols or not.  If so, then you should add them to libc.  If not, then you can remove this module on Android.

vdagonneau:
I'm having problems integrating the changes for android into libc; It is likely not as simple as I initially thought to support inotify for android.

Maybe we can start by merging the support for Linux only?

asomers:
What's the problem?  Does android use a different API?

vdagonneau:
The Android API is mostly the same with one little exception: inotify_rm_watch takes a uint32_t as its second argument on Android and an int on Linux.

Android: [https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/SYSCALLS.TXT](https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/SYSCALLS.TXT)

Linux: [https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/unix/sysv/linux/sys/inotify.h;hb=HEAD](https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/unix/sysv/linux/sys/inotify.h;hb=HEAD)

asomers:
That sounds pretty simple.  What's the problem?

vdagonneau:
Nothing really, it just took me a while to track down that slight difference in API.

asomers:
So the only difference between Linux and Android is signededness?  Lame.  Android is full of unforced errors like that.  Anyway, this PR looks good to me!  All it needs now is a squash.

vdagonneau:
Yes just the signededness, do you know where this kind of errors comes from ? It seems very weird to me.

asomers:
Honestly, I don't know why Android has so many errors like that.  There doesn't seem to be much excuse for it.

vdagonneau:
I squashed my commits, seems like everything is ready for merging!

asomers:
Oh I forgot: you need a CHANGELOG entry too.

asomers:
The CHANGELOG looks good, but I'm afraid your commits have become unsquashed.  Could you resquash them?

asomers:
Unfortunately that latested merge screwed up the PR history.  I'm afraid you'll have to rebase onto master.

vdagonneau:
Ah, I'm a little lost in my git history now. It might take me some time to figure everything out.

vdagonneau:
Ok, I think the git history looks good now.

asomers:
bors retry

vdagonneau:
Seems like something went wrong with bors. Any way I can help ?

PR #1017, opened 2019-01-23
Add sendmmsg() / recvmmsg()

dholroyd:
This is my initial go at adding `sendmmsg()` / `recvmmsg()`.

I am planning on updating this pull request with at least support for control messages, but would welcome any other feedback in the meantime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add several errno constants from OpenBSD 6.2
   ([#1036](https://github.com/nix-rust/nix/pull/1036))
  - Added support for sendmmsg() / recvmmsg() on Linux
+  ([#1017](https://github.com/nix-rust/nix/pull/1017))
 
 ### Changed
 - `PollFd` event flags renamed to `PollFlags` ([#1024](https://github.com/nix-rust/nix/pull/1024/))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#969](https://github.com/nix-rust/nix/pull/969))
 - Add several errno constants from OpenBSD 6.2
   ([#1036](https://github.com/nix-rust/nix/pull/1036))
+ - Added support for sendmmsg() / recvmmsg() on Linux
 
 ### Changed
 - `PollFd` event flags renamed to `PollFlags` ([#1024](https://github.com/nix-rust/nix/pull/1024/))
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -890,6 +890,8 @@ impl SockAddr {
             SockAddr::Unix(UnixAddr(ref mut addr, len)) => (mem::transmute(addr), (len + offset_of!(libc::sockaddr_un, sun_path)) as libc::socklen_t),
             #[cfg(any(target_os = "android", target_os = "linux"))]
             SockAddr::Netlink(NetlinkAddr(ref mut sa)) => (mem::transmute(sa), mem::size_of::<libc::sockaddr_nl>() as libc::socklen_t),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            SockAddr::Alg(AlgAddr(ref mut sa)) => (mem::transmute(sa), mem::size_of::<libc::sockaddr_alg>() as libc::socklen_t),
             #[cfg(any(target_os = "ios", target_os = "macos"))]
             SockAddr::SysControl(SysControlAddr(ref mut sa)) => (mem::transmute(sa), mem::size_of::<libc::sockaddr_ctl>() as libc::socklen_t),
             #[cfg(any(target_os = "android", target_os = "linux"))]
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -882,6 +882,27 @@ impl SockAddr {
             SockAddr::Link(LinkAddr(ref ether_addr)) => (mem::transmute(ether_addr), mem::size_of::<libc::sockaddr_dl>() as libc::socklen_t),
         }
     }
+
+    pub unsafe fn as_ffi_pair_mut(&mut self) -> (&mut libc::sockaddr, libc::socklen_t) {
+        match *self {
+            SockAddr::Inet(InetAddr::V4(ref mut addr)) => (mem::transmute(addr), mem::size_of::<libc::sockaddr_in>() as libc::socklen_t),
+            SockAddr::Inet(InetAddr::V6(ref mut addr)) => (mem::transmute(addr), mem::size_of::<libc::sockaddr_in6>() as libc::socklen_t),
+            SockAddr::Unix(UnixAddr(ref mut addr, len)) => (mem::transmute(addr), (len + offset_of!(libc::sockaddr_un, sun_path)) as libc::socklen_t),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            SockAddr::Netlink(NetlinkAddr(ref mut sa)) => (mem::transmute(sa), mem::size_of::<libc::sockaddr_nl>() as libc::socklen_t),
+            #[cfg(any(target_os = "ios", target_os = "macos"))]
+            SockAddr::SysControl(SysControlAddr(ref mut sa)) => (mem::transmute(sa), mem::size_of::<libc::sockaddr_ctl>() as libc::socklen_t),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            SockAddr::Link(LinkAddr(ref mut ether_addr)) => (mem::transmute(ether_addr), mem::size_of::<libc::sockaddr_ll>() as libc::socklen_t),
+            #[cfg(any(target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "ios",
+            target_os = "macos",
+            target_os = "netbsd",
+            target_os = "openbsd"))]
+            SockAddr::Link(LinkAddr(ref mut ether_addr)) => (mem::transmute(ether_addr), mem::size_of::<libc::sockaddr_dl>() as libc::socklen_t),
+        }
+    }
 }
 
 impl PartialEq for SockAddr {
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1416,7 +1416,7 @@ pub fn recvmmsg(fd: RawFd, msgvec: &mut[MMsgHdr], flags: MsgFlags, mut timeout: 
         libc::recvmmsg(
             fd,
             msgvec.as_mut_ptr() as *mut libc::mmsghdr,
-            msgvec.len() as u32,
+            msgvec.len() as _,
             flags.bits(),
             tptr as *mut libc::timespec,
         )
@@ -1433,7 +1433,7 @@ pub fn sendmmsg(fd: RawFd, msgvec: &mut[MMsgHdr]) -> Result<usize> {
         libc::sendmmsg(
             fd,
             msgvec.as_mut_ptr() as *mut libc::mmsghdr,
-            msgvec.len() as u32,
+            msgvec.len() as _,
             0,
         )
     };
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1355,12 +1355,12 @@ impl<'a> MMsgHdr<'a> {
         MMsgHdr (
             libc::mmsghdr {
                 msg_hdr: msghdr {
-                    msg_control: std::ptr::null_mut(),
+                    msg_control: ptr::null_mut(),
                     msg_controllen: 0,
                     msg_flags: flags.bits(),
                     msg_iov: iov.as_ptr() as *mut libc::iovec,
                     msg_iovlen: vlen,
-                    msg_name: std::ptr::null_mut(),
+                    msg_name: ptr::null_mut(),
                     msg_namelen: 0,
                 },
                 msg_len: 0,
@@ -1384,7 +1384,7 @@ pub fn recvmmsg(fd: RawFd, msgvec: &mut[MMsgHdr], flags: MsgFlags, timeout: Opti
     });
     let tptr = match t {
         Some(ref mut time) => time,
-        None => std::ptr::null_mut(),
+        None => ptr::null_mut(),
     };
     let ret = unsafe {
         libc::recvmmsg(
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1402,7 +1402,7 @@ impl<'a> SendMMsgHdr<'a> {
         };
         let mut hdr: libc::mmsghdr = unsafe { mem::uninitialized() };
         hdr.msg_hdr.msg_control = cmsg_ptr;
-        hdr.msg_hdr.msg_controllen = capacity;
+        hdr.msg_hdr.msg_controllen = capacity as _;
         hdr.msg_hdr.msg_flags = flags.bits();
         hdr.msg_hdr.msg_iov = iov.as_ptr() as *mut libc::iovec;
         hdr.msg_hdr.msg_iovlen = iov.len() as _;
@@ -1474,7 +1474,7 @@ impl<'a> RecvMMsgHdr<'a> {
         };
         let mut hdr: libc::mmsghdr = unsafe { mem::uninitialized() };
         hdr.msg_hdr.msg_control = msg_control as *mut _;
-        hdr.msg_hdr.msg_controllen = msg_controllen;
+        hdr.msg_hdr.msg_controllen = msg_controllen as _;
         hdr.msg_hdr.msg_flags = flags.bits();
         hdr.msg_hdr.msg_iov = iov.as_ptr() as *mut libc::iovec;
         hdr.msg_hdr.msg_iovlen = iov.len() as _;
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+ - Added support for sendmmsg() / recvmmsg() on Linux
```
asomers: CHANGELOG entries should have PR links.

src/sys/socket/mod.rs
```diff
@@ -1202,3 +1206,70 @@ pub fn shutdown(df: RawFd, how: Shutdown) -> Result<()> {
...
+        MMsgHdr(hdr, PhantomData)
+    }
+
+    pub fn msg_len(&self) -> usize {
+        self.0.msg_len as usize
+    }
+}
+
+#[cfg(any(
+target_os = "linux",
```
asomers: sendmmsg and recvmmsg are not unique to Linux.  You should enable them on all platforms that support them.

src/sys/socket/mod.rs
```diff
@@ -1202,3 +1206,70 @@ pub fn shutdown(df: RawFd, how: Shutdown) -> Result<()> {
...
+))]
+impl<'a> MMsgHdr<'a> {
+    pub fn new(iov: &mut[IoVec<&'a mut [u8]>], flags: MsgFlags) -> MMsgHdr<'a> {
+        // TODO:
+        //  - support 'control'
+        //  - support 'name'
+        let mut hdr: libc::mmsghdr = unsafe { mem::uninitialized() };
```
asomers: It looks like you're initializing every one of hdr's fields.  So there's no need to use `mem::uninitialized`.  Just initialize the whole thing at once.

src/sys/socket/mod.rs
```diff
@@ -1202,3 +1206,70 @@ pub fn shutdown(df: RawFd, how: Shutdown) -> Result<()> {
...
+
+#[cfg(any(
+target_os = "linux",
+))]
+pub fn sendmmsg(fd: RawFd, msgvec: &mut[MMsgHdr]) -> Result<usize> {
+    let ret = unsafe {
+        libc::sendmmsg(
+            fd,
+            msgvec.as_mut_ptr() as *mut libc::mmsghdr,
+            msgvec.len() as u32,
```
asomers: Beware: this field is usize on FreeBSD.

src/sys/socket/mod.rs
```diff
@@ -1202,3 +1206,70 @@ pub fn shutdown(df: RawFd, how: Shutdown) -> Result<()> {
...
+
+    pub fn msg_len(&self) -> usize {
+        self.0.msg_len as usize
+    }
+}
+
+#[cfg(any(
+target_os = "linux",
+))]
+pub fn recvmmsg(fd: RawFd, msgvec: &mut[MMsgHdr], flags: MsgFlags, mut timeout: Option<TimeSpec>) -> Result<usize> {
```
asomers: 80 columns per line, please.

test/sys/test_socket.rs
```diff
@@ -683,3 +683,66 @@ pub fn test_recv_ipv6pktinfo() {
...
+target_os = "linux",
+))]
+#[test]
+pub fn test_mmsg() {
+    use std::thread;
+    use nix::sys::uio::IoVec;
+    use nix::sys::socket::{socket, bind, connect, recvmmsg, sendmmsg, MMsgHdr, MsgFlags, AddressFamily, SockType, SockAddr, SockFlag, InetAddr};
+    use std::time;
+
+    let sender = thread::spawn(move || {
```
asomers: You probably don't need to spawn a thread here.  You should try removing it, and the test will probably still work.  The order of operations should be:
1) Create and bind the receiving socket
2) Created, bind, and connect the sending socket
3) sendmmsg
4) recvmmsg

test/sys/test_socket.rs
```diff
@@ -683,3 +683,66 @@ pub fn test_recv_ipv6pktinfo() {
...
+        let so = socket(
+            AddressFamily::Inet,
+            SockType::Datagram,
+            SockFlag::empty(),
+            None,
+        ).expect("send socket failed");
+        let sockaddr = SockAddr::new_inet(InetAddr::from_std(&SocketAddr::from_str("127.0.0.1:0").unwrap()));
+        bind(so, &sockaddr).unwrap();
```
asomers: Why do you bother binding the sending socket?  Isn't that automatic?

test/sys/test_socket.rs
```diff
@@ -683,3 +683,66 @@ pub fn test_recv_ipv6pktinfo() {
...
+        thread::park();
+        sendmmsg(so, &mut msgs[..]).unwrap();
+    });
+    let so = socket(
+        AddressFamily::Inet,
+        SockType::Datagram,
+        SockFlag::empty(),
+        None,
+    ).expect("send socket failed");
+    let sockaddr = SockAddr::new_inet(InetAddr::from_std(&SocketAddr::from_str("127.0.0.1:3456").unwrap()));
```
asomers: The address and port number is a magic number.  You shouldn't duplicate it.  Just define it as a const.

test/sys/test_socket.rs
```diff
@@ -683,3 +683,66 @@ pub fn test_recv_ipv6pktinfo() {
...
+    let mut c = [0u8; 1500];
+    let mut iov_a = [ IoVec::from_mut_slice(&mut a[..]) ];
+    let mut iov_b = [ IoVec::from_mut_slice(&mut b[..]) ];
+    let mut iov_c = [ IoVec::from_mut_slice(&mut c[..]) ];
+    let mut msgs = [
+        MMsgHdr::new(&mut iov_a[..], MsgFlags::empty()),
+        MMsgHdr::new(&mut iov_b[..], MsgFlags::empty()),
+        MMsgHdr::new(&mut iov_c[..], MsgFlags::empty()),
+    ];
+    let count = recvmmsg(so, &mut msgs[..], MsgFlags::MSG_DONTWAIT, None).unwrap();
```
asomers: Why `MSG_DONTWAIT`?

src/sys/socket/mod.rs
dholroyd: I've added freebsd and netbsd.  Let me know if I missed something.

src/sys/socket/mod.rs
dholroyd: I had originally gone that way, but the x86_64-unknown-linux-musl build failed due to missing `_pad1` / `_pad2` initialisation -- https://travis-ci.org/nix-rust/nix/jobs/483633104

src/sys/socket/mod.rs
dholroyd: Actually netbsd and freebsd builds failed e.g. https://travis-ci.org/nix-rust/nix/jobs/484619844 , since the libc crate doesn't define these.

So removed those again.

asomers:
Rather than disabling this feature on FreeBSD and NetBSD, you should submit a PR to libc to add the definitions there.  When you do, check other OSes too.

src/sys/socket/mod.rs
```diff
@@ -9,6 +9,20 @@ use std::{fmt, mem, ptr, slice};
 use std::os::unix::io::RawFd;
 use sys::time::TimeVal;
 use sys::uio::IoVec;
+#[cfg(any(
+target_os = "linux",
```
asomers: Indent for consistency's sake.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#969](https://github.com/nix-rust/nix/pull/969))
 - Add several errno constants from OpenBSD 6.2
   ([#1036](https://github.com/nix-rust/nix/pull/1036))
+ - Added support for sendmmsg() / recvmmsg() on Linux
```
asomers: This description is blisfully out-of-date.  You can remove the "on Linux" part.

src/sys/socket/addr.rs
```diff
@@ -882,6 +882,29 @@ impl SockAddr {
             SockAddr::Link(LinkAddr(ref ether_addr)) => (mem::transmute(ether_addr), mem::size_of::<libc::sockaddr_dl>() as libc::socklen_t),
         }
     }
+
+    pub unsafe fn as_ffi_pair_mut(&mut self) -> (&mut libc::sockaddr, libc::socklen_t) {
```
asomers: This is duplicative.  Can you think of any way to share code between this method and `as_ffi_pair`?  Also, you should wrap to 80 cols.

src/sys/socket/mod.rs
```diff
@@ -1334,3 +1348,233 @@ pub fn shutdown(df: RawFd, how: Shutdown) -> Result<()> {
...
+}
+
+/// Transmit multiple messages on a socket using a single system call.
+#[cfg(any(
+    target_os = "linux",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "android",
+))]
+pub fn sendmmsg(fd: RawFd, msgvec: &mut[SendMMsgHdr]) -> Result<usize> {
```
asomers: This is a very complicated function to use.  Could you provide an example?  Also, why no `flags` argument?

dholroyd:
Just want to say that I plan to address all the comments you've kindly provided Alan; all look reasonable.

I'm currently trying to integrate the API in the form from my branch into a higher-level API within actual application code, in hopes that exercise might knock off some rough edges here.

src/sys/socket/mod.rs
```diff
@@ -1334,3 +1348,233 @@ pub fn shutdown(df: RawFd, how: Shutdown) -> Result<()> {
...
+    let tptr = match timeout {
+        Some(ref mut time) => time.as_ref() as *const libc::timespec,
+        None => ptr::null_mut(),
+    };
+    let ret = unsafe {
+        libc::recvmmsg(
+            fd,
+            msgvec.as_mut_ptr() as *mut libc::mmsghdr,
+            msgvec.len() as _,
+            flags.bits(),
```
gnzlbg: Note: on `musl` the `flag` argument is an `unsigned int`, while on `gnu`, it is a signed int.

kawakami-o3:
What are you struggling with, @dholroyd ? I want to help you.

dholroyd:
> What are you struggling with, @dholroyd ?

There are the usual issues of time and priorities, but further than that, I think the interfaces I made don't work very well:
 - An app that would benefit from these optimisations (vs non-batched sendmsg/recvmsg) will also want to avoid allocations / copies per packet
 - RecvMMsgHdr and SendMMsgHdr both borrow the underlying buffers, so while these instances exist the underlying buffers can't be modified
 - modifying underlying buffers in between sendmmsg/recvmmsg calls is something that applications will naturally need to do (e.g. so as to be able to supply different payloads on each call to sendmmsg)
 - I think calling code has to drop the sets of RecvMMsgHdr and SendMMsgHdr instances in between each call in order to regain mutable access to underlying buffers, and will have to recreate new instances for subsequent calls, which is a drain on efficiency

In order to try and deal with these problems I tried to come up with an interface that ends up looking significantly different than sendmmsg/recvmmsg while fixing these problems and being safe.  I didn't finish that exercise either, but here's a sketch of the WIP (quoted out of context, so it will not work as-is).  My idea was to provide a datastruture to own the underlying `libc::mmsghdr` instances and allow them to be safely reused -- hiding complexity that the previous implementation forces on the caller.  I'm not sure if such a 'thick' layer on top of sendmmsg/recvmmsg is in keeping with nix philosophy though?

```rust
struct RecvPk<P: Packet> {
    packet: P,
    iovec: [ libc::iovec; 1],
    _pin: marker::PhantomPinned,
}
impl<P: Packet> RecvPk<P> {
    fn new(mut packet: P) -> RecvPk<P> {
        let pay = packet.payload_mut();
        let iov = libc::iovec {
            iov_base: pay.as_ptr() as *mut libc::c_void,
            iov_len: pay.len() as libc::size_t,
        };
        let mut result = RecvPk {
            packet,
            iovec: [ iov ],
            _pin: marker::PhantomPinned
        };
        result
    }

    fn into_packet(self, len: usize) -> P {
        let mut pk = self.packet;
        pk.truncate(len);
        pk
    }
}

pub struct RecvQueue<BP: BufferPool> {
    buffer_pool: BP,
    // TODO: use a type with a similar interface that guarantees stable addresses for the contained
    //       items (VecDeque could relocate items if capacity is exceeded, which we do take care to
    //       avoid).
    recv_packets: VecDeque<RecvPk<BP::P>>,
    // storage can't be a VecDeque, since we will need to pass a contiguous slice
    // to recevmmsg()
    storage: SliceDeque<libc::mmsghdr>,
    hist: Vec<u64>,
    count: usize,
}
impl<BP: BufferPool> RecvQueue<BP> {
    pub fn new(buffer_pool: BP) -> RecvQueue<BP> {
        let mut queue = RecvQueue {
            buffer_pool,
            recv_packets: VecDeque::with_capacity(BATCH_SIZE),
            storage: SliceDeque::with_capacity(BATCH_SIZE),
            hist: vec![0; BATCH_SIZE],
            count: 0,
        };
        queue.replenish();
        queue
    }

    fn replenish(&mut self) {
        for _ in self.storage.len()..BATCH_SIZE {
            self.add(self.buffer_pool.allocate().unwrap(/*FIXME*/));
        }
        // if the list which contains the iovec elements does not have the expected capacity, this
        // likely means it has been reallocated, and the pointers from the members of the 'storage'
        // Vec will now point at random memory; try to detect this and fail fast!
        assert_eq!(self.recv_packets.len(), BATCH_SIZE);
        assert_eq!(self.storage.len(), BATCH_SIZE);
    }
    fn add(&mut self, packet: BP::P) {
        let pk = RecvPk::new(packet);
        self.recv_packets.push_back(pk);
        let pk = self.recv_packets.get_mut(self.recv_packets.len()-1).unwrap();
        unsafe {
            self.storage.push_back(mem::uninitialized());
        }
        let mut hdr = self.storage.last_mut().unwrap();
        let flags = 0;
        hdr.msg_hdr.msg_control = ptr::null_mut();
        hdr.msg_hdr.msg_controllen = 0;
        hdr.msg_hdr.msg_flags = flags;
        hdr.msg_hdr.msg_iov = pk.iovec.as_mut_ptr() as *mut libc::iovec;
        hdr.msg_hdr.msg_iovlen = pk.iovec.len() as _;
        hdr.msg_hdr.msg_name = ptr::null_mut();
        hdr.msg_hdr.msg_namelen = 0;
        hdr.msg_len = 0;

    }

    pub fn recvbatch<'a>(&'a mut self, sock: &mio::net::UdpSocket) -> Result<(Batch, Option<impl Iterator<Item = BP::P> + 'a>), io::Error> {
        self.replenish();
        let msgvec = self.storage[..].as_mut_ptr();
        let vlen = self.storage.len() as u32;
        let flags = 0;
        let timeout = ptr::null_mut();
        let res = unsafe { libc::recvmmsg(sock.as_raw_fd(), msgvec, vlen, flags, timeout) };
        let size = match nix::errno::Errno::result(res) {
            Ok(s) => s as usize,
            Err(nix::Error::Sys(errno)) if errno == nix::errno::Errno::EAGAIN => {
                return Ok((Batch::Exhausted, None));
            }
            Err(nix::Error::Sys(e)) => return Err(io::Error::from(e)),
            Err(e) => return Err(io::Error::new(io::ErrorKind::Other, e.description())),
        };
        self.hist[size - 1] += 1;
        self.count += 1;
        if self.count & 0b111111111111 == 0 {
            println!("stats: {:?}", self.hist);
            self.hist.iter_mut().for_each(|v| *v = 0 )
        }
        let pks = self.recv_packets.drain(0..size);
        // we want SliceDeque::drain(), but that's unstable so we can't use it.  This hack with
        // repeat_with() lacks the guarantees that drain() would give about actually dropping
        // the items no matter if the caller actually consumes the iterator; it's therefore not
        // really a good enough solution 'for real', but lets us benchmark
        /*let store = &mut self.storage;
        let hdrs = iter::repeat_with(move || { store.remove(0) } ).take(size);
        //let hdrs = self.storage.drain(0..size);
        let result = hdrs.zip(pks).map(|(hdr, pk)| {
            pk.into_packet(hdr.msg_len as usize)
        });*/
        let result = Drain { iter: pks, store: &mut self.storage };

        if size < BATCH_SIZE {
            // either there were no more packets to receive (so it's legitimate to
            // avoid calling recvmmsg() again any avoid getting EAGAIN) or an error
            // occurred after receiving at least one packet.  Since we can't
            // differentiate these two cases, and we would like to avoid the extra
            // syscall that will immediately get EAGAIN, we break out of the recv loop
            // here, and rely on the socket having been registered with MIO using
            // PollOpt::level() to let any remaining buffered packets be read on
            // another iteration of the overall event loop,
            Ok((Batch::Exhausted, Some(result)))
        } else {
            Ok((Batch::Continues, Some(result)))
        }
    }
}
struct Drain<'a, P: Packet> {
    iter: std::collections::vec_deque::Drain<'a, RecvPk<P>>,
    store: &'a mut SliceDeque<libc::mmsghdr>,
}
impl<'a, P: Packet> Iterator for Drain<'a, P> {
    type Item = P;

    fn next(&mut self) -> Option<Self::Item> {
        let pk = self.iter.next()?;
        let hdr = &self.store[0];
        let result = pk.into_packet(hdr.msg_len as usize);
        self.store.remove(0);
        Some(result)
    }
}

#[derive(PartialEq, Debug)]
pub enum Batch {
    Continues,
    Exhausted,
}
```

kawakami-o3:
Thank you for explaining the current state. Honestly, I'm strugling to implement a wrapper for recvmmsg(). Your idea is very insightful, although it needs dicision making. I will also try to make a proof of concept.

dholroyd:
Superseded by #1208

PR #1018, opened 2019-01-25
Add sigaltstack and related types

acfoltzer:
`sigaltstack` is part of the `sigaction` family of APIs, used to control which stack signal handlers run on.


src/sys/signal.rs
```diff
@@ -831,6 +831,95 @@ mod sigevent {
...
+#[derive(Copy, Clone)]
+#[allow(missing_debug_implementations)]
+pub struct SigStack {
+    stack: libc::stack_t,
+}
+
+impl SigStack {
+    /// Define an alternate signal stack for use with `sigaction`.
+    ///
+    /// See [`sigaltstack(3posix)`](http://man7.org/linux/man-pages/man3/sigaltstack.3p.html) for
```
asomers: For standardized functions, link to the open group's man pages instead of Linux's.  In this case: http://pubs.opengroup.org/onlinepubs/9699919799/ .

src/sys/signal.rs
```diff
@@ -831,6 +831,95 @@ mod sigevent {
...
+///         ptr::null_mut(),
+///         SIGSTKSZ,
+///         ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+///         MapFlags::MAP_ANON | MapFlags::MAP_PRIVATE,
+///         -1,
+///         0,
+///     )
+///     .unwrap()
+/// };
+/// let sigstack = SigStack::new(stack_mem, SigStackFlags::empty(), SIGSTKSZ);
```
asomers: For the docs, `SIGSTKSZ` looks mysterious and undefined.  Better to refer to it as `libc::SIGSTKSZ`.

src/sys/signal.rs
```diff
@@ -831,6 +831,95 @@ mod sigevent {
...
+pub struct SigStack {
+    stack: libc::stack_t,
+}
+
+impl SigStack {
+    /// Define an alternate signal stack for use with `sigaction`.
+    ///
+    /// See [`sigaltstack(3posix)`](http://man7.org/linux/man-pages/man3/sigaltstack.3p.html) for
+    /// details on the requirements for an alternate signal stack.
+    pub fn new(sp: *mut libc::c_void, flags: SigStackFlags, size: libc::size_t) -> SigStack {
```
asomers: I really don't like this API.  It isn't Rusty.  For one thing, it exposes raw pointers in its public API.  For another, it forces the user to specify the size twice: here and wherever he allocated the memory.  Can you try to come up with a nicer API?  Taking the memory region as a mutable slice would be one option.  Handling allocation too would be even better, though we need to be careful not to proscribe valid use cases.

acfoltzer: I added an alternate interface that takes a mutable byte slice. Having `SigStack` itself own the allocation would increase complexity and decrease safety, particularly around how `sigaction`'s would be handled. Better for the user to know they have to explicitly manage unsafe lifetimes.

acfoltzer-fastly:
Hmm, looks like an error during dependency builds for the FreeBSD check

acfoltzer:
Hmm, looks like an error during dependency builds for the FreeBSD check

src/sys/signal.rs
```diff
@@ -831,6 +831,127 @@ mod sigevent {
...
+    /// };
+    /// let sigstack = SigStack::from_raw(stack_mem, SigStackFlags::empty(), libc::SIGSTKSZ);
+    /// unsafe {
+    ///     let old_sigstack = sigaltstack(&sigstack).unwrap();
+    ///     sigaltstack(&old_sigstack).unwrap();
+    ///     // unmap memory only after the stack is unregistered
+    ///     munmap(stack_mem, libc::SIGSTKSZ);
+    /// }
+    /// ```
+    pub fn from_raw(sp: *mut libc::c_void, flags: SigStackFlags, size: libc::size_t) -> SigStack {
```
asomers: I think that the only way to create the SigStack should be from a slice.  It's a cleaner interface than using the raw pointer, and it's easy for the user to construct his own slice if he must.

src/sys/signal.rs
```diff
@@ -831,6 +831,127 @@ mod sigevent {
...
+    /// let sigstack = SigStack::from_slice(stack_mem.as_mut_slice(), SigStackFlags::empty());
+    /// unsafe {
+    ///     let old_sigstack = sigaltstack(&sigstack).unwrap();
+    ///     sigaltstack(&old_sigstack).unwrap();
+    /// }
+    /// // keep `stack_mem` alive until after the stack is unregistered
+    /// drop(stack_mem);
+    /// ```
+    pub fn from_slice(stack: &mut [u8], flags: SigStackFlags) -> SigStack {
+        let mut st = unsafe { mem::uninitialized::<libc::stack_t>() };
```
asomers: Why use mem::uninitialized when you're going to initialize every field?

src/sys/signal.rs
```diff
@@ -831,6 +831,127 @@ mod sigevent {
...
+        st.ss_sp = sp;
+        st.ss_flags = flags.bits();
+        st.ss_size = size;
+        SigStack { stack: st }
+    }
+
+    /// Define an alternate signal stack for use with `sigaction` using a mutable byte slice.
+    ///
+    /// The `stack` slice must live at least as long as any signal handlers are registered to use
+    /// this stack. Since the lifetime of `stack` is not reflected in the type of `SigStack`, this
```
asomers: You can ensure that the lifetime of `stack` contains the lifetime of the slice by adding a `std::marker::PhantomData` field.  In fact, do you think you should also add a `Drop` implementation that deregisters the signal stack?  That would ensure that the byte slice never gets dropped while it's still being used.

acfoltzer: Unfortunately the lifetime story is more complicated than that; calling `sigaltstack` essentially writes the stack pointer to a thread-local variable out in FFI land. We could add a lifetime parameter to `SigStack`, but we can't thread that lifetime through FFI land, so that parameter would be bogus in the return type of `sigaltstack`.

We also can't unregister in the `Drop` impl, because if other code has come along and also called `sigaltstack`, we might end up clobbering someone else's stack.

src/sys/signal.rs
acfoltzer: Constructing a slice as an intermediate step introduces an unnecessary use of `unsafe`, so I'd much rather keep `from_raw`.

AlexApps99:
@acfoltzer do you have plans to finish this PR?

acfoltzer:
@AlexApps99 apologies for the substantial delay in my reply: no, not at this time. I am closing this particular PR but will keep the branch around; please feel free to adopt it if this is still relevant to you.

PR #1019, opened 2019-01-30
WIP use doc_cfg for cross-platform documentation

asomers:
This commit demonstrates the use of the unstable doc_cfg feature to make
cross-platform documentation.  Properly annotated items will be visible
in the documentation regardless of which platform it was built on.  But
there are still a few warts:

* The feature is unstable, so it must be gated by a "nightly-docs"
  feature flag.
* Rust-2015-style macros must be annotated individually, even if they're
  behind an annotated module.
* Struct members, even private ones, must be #[cfg()] gated even if
  their struct is already gated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,13 @@ exclude     = [
   "/bors.toml"
 ]
 
+[package.metadata.docs.rs]
+features = ["nightly-docs"]
+
+[features]
+# For building documentation only; no functional change to the library.
+nightly-docs = []
+
 [dependencies]
 libc = { git = "https://github.com/rust-lang/libc" }
 bitflags = "1.0"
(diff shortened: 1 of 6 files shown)


asomers:
cc @nix-rust/nix-maintainers for review.  What do you think?  I think this would be tremendously useful to Nix.  My only doubt comes from the need to laboriously regate all private struct members.  We might be able to fix that in rustdoc.

Susurrus:
Found the docs for this feature [here](https://doc.rust-lang.org/unstable-book/language-features/doc-cfg.html). Sounds like a great idea. Might make it much more clear for our audience when they see something they expect to work for a platform, but it doesn't, what's going on. I wonder if/when `docs.rs` will support it.

asomers:
Actually, docs.rs already supports it, because docs.rs uses the nightly compiler.  See for example https://docs.rs/futures-locks/0.3.3/futures_locks/struct.Mutex.html#method.with .

asomers:
The need for this has largely been eliminated by #1337 .

PR #1020, opened 2019-01-30
Major cmsg cleanup

### LINKED ISSUE #994: CmsgSpace::new unsafely creates uninitialized, arbitrary types
Even though the uninitialized field is private, it still gets dropped. For example `nix::sys::socket::CmsgSpace::<String>::new();` easily causes a segfault.

The field should probably be `ManuallyDrop<T>`, or there should be a `T: Copy` bound.

### LINKED ISSUE #999: Misaligned reference constructed by ControlMessage::ScmTimestamp
On macOS (and maybe other systems?) cmsg data is 4-byte aligned. In 64-bit macOS environments, `libc::timeval` contains 8-byte values, and is presumably 8-byte aligned. `ControlMessage` directly converts the pointer to the cmsg data into a typed reference, which is therefore apparently unaligned and incurs UB. It seems like the only safe thing to do here would be to use `ptr::read_unaligned`.

### LINKED ISSUE #1013: Audit all cmsg code
control messages used by `sendmsg` and `recvmsg` are one of the hardest Unix APIs.  Some of Nix's cmsg-related tests currently fail on OpenBSD (see #1001), and others rely on a hard-coded os-dependent value that is not defined anywhere in libc (`align_of_cmsg_data`).  Both of these things hint at underlying bugs in our cmsg implementation.  We should audit the entire thing for correctness, and possibly rewrite it to be more C-like.  libc already exports versions of the `CMSG_DATA` family of macros.  Nix should use those rather than reinvent their logic.

asomers:
This PR fixes multiple bugs in the cmsg code.
Fixes #994 
Fixes #999 
Fixes #1013

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -333,13 +333,13 @@ pub trait CmsgBuffer {
 /// # use nix::sys::time::TimeVal;
 /// # use std::os::unix::io::RawFd;
 /// # fn main() {
-/// // Create a buffer for a `ControlMessage::ScmTimestamp` message
+/// // Create a buffer for a `ControlMessageOwned::ScmTimestamp` message
 /// let _ = cmsg_space!(TimeVal);
-/// // Create a buffer big enough for a `ControlMessage::ScmRights` message
+/// // Create a buffer big enough for a `ControlMessageOwned::ScmRights` message
 /// // with two file descriptors
 /// let _ = cmsg_space!([RawFd; 2]);
-/// // Create a buffer big enough for a `ControlMessage::ScmRights` message
-/// // and a `ControlMessage::ScmTimestamp` message
+/// // Create a buffer big enough for a `ControlMessageOwned::ScmRights` message
+/// // and a `ControlMessageOwned::ScmTimestamp` message
 /// let _ = cmsg_space!(RawFd, TimeVal);
 /// # }
 /// ```
@@ -440,15 +440,15 @@ pub struct CmsgIterator<'a> {
 }
 
 impl<'a> Iterator for CmsgIterator<'a> {
-    type Item = ControlMessage<'a>;
+    type Item = ControlMessageOwned;
 
-    fn next(&mut self) -> Option<ControlMessage<'a>> {
+    fn next(&mut self) -> Option<ControlMessageOwned> {
         match self.cmsghdr {
             None => None,   // No more messages
             Some(hdr) => {
                 // Get the data.
                 // Safe if cmsghdr points to valid data returned by recvmsg(2)
-                let cm = unsafe { Some(ControlMessage::decode_from(hdr))};
+                let cm = unsafe { Some(ControlMessageOwned::decode_from(hdr))};
                 // Advance the internal pointer.  Safe if mhdr and cmsghdr point
                 // to valid data returned by recvmsg(2)
                 self.cmsghdr = unsafe {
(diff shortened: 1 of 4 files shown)


asomers:
cc @pusateri @ralith @goffrie

src/sys/socket/mod.rs
```diff
@@ -369,99 +389,95 @@ pub struct CmsgSpace<T> {
...
-    /// This is similar to the socket option `SO_PEERCRED`, but requires a
-    /// process to explicitly send its credentials. A process running as root is
-    /// allowed to specify any credentials, while credentials sent by other
-    /// processes are verified by the kernel.
-    ///
-    /// For further information, please refer to the
-    /// [`unix(7)`](http://man7.org/linux/man-pages/man7/unix.7.html) man page.
-    // FIXME: When `#[repr(transparent)]` is stable, use it on `UnixCredentials`
-    // and put that in here instead of a raw ucred.
+pub enum ControlMessageOwned {
```
Ralith: Per [my previous comment](https://github.com/nix-rust/nix/issues/999#issuecomment-454601121), I think this is the wrong way to go, as it incurs avoidable heap allocation and copying.

src/sys/socket/mod.rs
```diff
@@ -345,6 +318,53 @@ cfg_if! {
...
+/// let _ = cmsg_space!(TimeVal);
+/// // Create a buffer big enough for a `ControlMessageOwned::ScmRights` message
+/// // with two file descriptors
+/// let _ = cmsg_space!([RawFd; 2]);
+/// // Create a buffer big enough for a `ControlMessageOwned::ScmRights` message
+/// // and a `ControlMessageOwned::ScmTimestamp` message
+/// let _ = cmsg_space!(RawFd, TimeVal);
+/// ```
+// Unfortunately, CMSG_SPACE isn't a const_fn, or else we could return a
+// stack-allocated array.
```
Ralith: Could this be corrected? This seems like exactly the special case that const fns can already handle gracefully, and is the intended use of the C macros.

src/sys/socket/mod.rs
```diff
@@ -345,6 +318,53 @@ cfg_if! {
...
+/// # use std::os::unix::io::RawFd;
+/// // Create a buffer for a `ControlMessageOwned::ScmTimestamp` message
+/// let _ = cmsg_space!(TimeVal);
+/// // Create a buffer big enough for a `ControlMessageOwned::ScmRights` message
+/// // with two file descriptors
+/// let _ = cmsg_space!([RawFd; 2]);
+/// // Create a buffer big enough for a `ControlMessageOwned::ScmRights` message
+/// // and a `ControlMessageOwned::ScmTimestamp` message
+/// let _ = cmsg_space!(RawFd, TimeVal);
+/// ```
```
Ralith: I love this interface!

src/sys/socket/mod.rs
```diff
@@ -369,99 +389,95 @@ pub struct CmsgSpace<T> {
...
+        &mut self[..]
+    }
+}
+
+#[allow(missing_debug_implementations)] // msghdr isn't Debug
 pub struct RecvMsg<'a> {
-    // The number of bytes received.
-    pub bytes: usize,
-    cmsg_buffer: &'a [u8],
+    cmsghdr: *const cmsghdr,
```
Ralith: Why not a real reference?

Ralith:
Thanks for your work on the libc macros, by the way!

src/sys/socket/mod.rs
asomers: The libc people are resistant to that kind of change. https://github.com/rust-lang/libc/issues/1213

src/sys/socket/mod.rs
asomers: I chose to go this path because it's closer to backwards-compatible.  I tried the 100% backwards-compatible method first, but the code got very ugly very quick, and it involved even more copying.

Ralith: It seems to be that you've settled on the worst of both worlds: both a breaking change and inefficient.

src/sys/socket/mod.rs
Ralith: References can convert to pointers implicitly, and the `PhantomData` would be rendered unnecessary.

asomers: As a matter of fact, this turned out to be pretty easy.  I'll do it.

src/sys/socket/mod.rs
asomers: Can you suggest something more efficient that causes no more pain for existing users?  This solution, while not exactly backwards compatible, only requires a simple search and replace.

Ralith: I don't think the specific level of breakage makes sense as a line in the sand that cannot be crossed by a single step. That said, you could compromise by only using the flyweight pattern where heap allocation is otherwise necessary, i.e. in `ScmRights`, though this would make "Owned" no longer an accurate descriptor.

asomers:
The QEMU failures are due to a bug in QEMU itself.  For once I'm sure of it.  The emulated recvmsg is writing beyond the end of the supplied msg_control buffer.  I think it's probably due to the linked qemu bug, which is already fixed.  I'll disable the test.  Once japaric/cross updates its image to use qemu 2.12.0, then we can reenable it.
https://bugs.launchpad.net/qemu/+bug/1701808
https://github.com/japaric/rust-cross/issues/49

src/sys/socket/mod.rs
asomers: But what's to say that `ScmRights`'s payload is aligned on all platforms that Nix will ever support?  I don't think that's guaranteed, especially if a user ever passes an unaligned msg_control field to recvmsg.

asomers:
Sigh.  Travis got confused somehow and thinks I'm missing a merge commit.  I guess I may as well rebase.

src/sys/socket/mod.rs
Ralith: That's why I said "flyweight pattern" and not "a slice", yes. Refer to the comment I linked for an example implementation.

asomers:
I'm going to go ahead and merge this.  @Ralith if you're still concerned, then please submit a separate PR to reduce data copies.

Ralith:
>  @Ralith if you're still concerned, then please submit a separate PR to reduce data copies

I don't actually use Nix at present, so I'm just going to stick with my current implementation that doesn't have these issues.

Issue #317, opened 2016-03-16
Fix flag type names to match FOO_* => FooFlags convention


kamalmarhubi:
The convention was added in #287, and there are a few places that need changing. I'm going through all the bitflags as part of #315, and noticing them. Here's a (partial) list:

- [ ] `poll::EventFlags` => `PollFlags` (1)
- [ ] `sys::epoll::EpollEventKind` => `EpollFlags` (1)
- [ ] `sys::event::EventFlag` => `EvFlags`
- [ ] `sys::event::FilterFlag` => `NoteFlags`
- [ ] `sys::eventfd::EventFdFlag` => `EFdFlags`
- [ ] `sys::memfd::MemFdCreateFlag` => `MFdFlags`
- [ ] `sys::quota::QuotaValidFlags` => `QIFFlags` (unsure about capitalization)
- [ ] `sys::stat::{SFlag, Mode}` => `SFlags`? (2)
- [ ] `sys::statvfs::FsFlags` => use `sys::mount::MsFlags`? (3)
- [ ] `sys::termios::{InputFlags, OutputFlags, ControlFlags, LocalFlags}` => ? (4)

(1) these might need amending the convention. The flags are named `POLLIN`, `POLLOUT`, `EPOLLIN`, `EPOLLOUT`, etc: there is no underscore

(2) these are used in different parameters; needs to looked at a bit more carefully

(3) man page for `statvfs(2)` says
> The field f_flag is a bit mask (of mount flags, see mount(8)).

(4) These flags have no common prefix.

amanjeev:
I guess this can be closed as https://github.com/nix-rust/nix/pull/1024 is merged?

asomers:
Some of the other types haven't been changed yet, though.

amanjeev:
@asomers oh let me see which ones. Sorry I missed those. If you can point me to some other, I'd be glad to change those as well. It will give me some way to peruse through the source.

palash25:
Hi :wave: I made a PR for `sys::event` can someone please review it.

PR #1024, opened 2019-02-11
Fix for #317 - Some event flags renaming

amanjeev:
I tried to grep and replace for #317. Please let me know if I need to fix something else as well but in my search I saw that most were fixed earlier.


asomers:
Looks good!  But could you please add a CHANGELOG entry and also describe what you did in your commit message?

amanjeev:
@asomers will do!

PR #1032, opened 2019-02-22
Add getpagesize for linux

condy0919:
:)

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -2249,6 +2249,14 @@ pub fn sysconf(var: SysconfVar) -> Result<Option<c_long>> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


asomers:
Is this function really necessary?  It's in C because it predates `sysconf`.  But since Nix already has `sysconf`, I'm not sure there's any value to adding it.  It doesn't save the user any code.

condy0919:
It can facilitate getting PAGESIZE without `unwrap` a `sysconf` call.
I think it helpful.

asomers:
IMHO saving an `unwrap` is an insufficient reason to add a new function, especially one that's no longer even in POSIX.

condy0919:
ok, I will close this pr

PR #1033, opened 2019-03-01
Fix the tests with libc 0.2.49 and Rust 1.33.0

asomers:
Several symbols are now marked as deprecated on OSX.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1020](https://github.com/nix-rust/nix/pull/1020))
 
 ### Removed
+- `Daemon`, `NOTE_REAP`, and `NOTE_EXIT_REPARENTED` are now deprecated on OSX
+  and iOS.
+  ([#1033](https://github.com/nix-rust/nix/pull/1033))
 
 ## [0.13.0] - 2019-01-15
 ### Added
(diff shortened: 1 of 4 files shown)


asomers:
This should fix PR #1031

asomers:
Can one of the @nix-rust/nix-maintainers with an OSX machine please verify whether Nix 0.13.0 can even build with libc 0.2.49?

asomers:
Ok, I'm stumped.  Why didn't the `#[allow(deprecated)]` fix the tests on OSX?

Susurrus:
I have a machine and can dig into this tonight if that'd be helpful.

asomers:
That would indeed be helpful.  But don't bother with my first question; 0.13.0 should still work fine, because we only `#[deny(warnings)]` during `#[cfg(test)]`.

asomers:
Ok, I'm an idiot.  I never actually committed the important part.

asomers:
The Linux failure looks like a regression in Cargo 1.32.0.  I'll fix it.

asomers:
It looks like Cargo <= 1.32.0 would create a new thread for every test, even if `--test-threads=1`.  And in-between tests it would destroy the test thread and recreate it. But Cargo 1.33.0 skips creating a new thread when `--test-threads=1`.  And even if `--test-threads=1`, it reuses threads for multiple tests.  This has exposed a failure to teardown some per-thread data in one of our unit tests.

Weirdly, this is a case where `--test-threads=1` is _more_ likely to make the test fail.

PR #1034, opened 2019-03-04
Add killpg

### LINKED ISSUE #644: Missing killpg
Needed by Ion to send signals to all members of a process group.

DanSnow:
It's seem that #644 is inactively about 1 year. But I really want this API that can land in nix.

Actually I not really understand how to check the availability of an API for other platform except Linux. But I saw that this API in `libc` is wrapping inside a `#[cfg(unix)]`. So I don't need to add any `#[cfg]` on this API, am I right?

Resolved #644

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -680,6 +680,22 @@ pub fn kill<T: Into<Option<Signal>>>(pid: ::unistd::Pid, signal: T) -> Result<()
     Errno::result(res).map(drop)
 }
 
+/// Send a signal to a process group [(see
+/// killpg(3))](http://pubs.opengroup.org/onlinepubs/9699919799/functions/killpg.html).
+///
+/// If `pgrp` less then or equal 1, the behavior is platform-specific.
+/// If `signal` is `None`, `killpg` will only preform error checking and won't
+/// send any signal.
+pub fn killpg<T: Into<Option<Signal>>>(pgrp: ::unistd::Pid, signal: T) -> Result<()> {
+    let res = unsafe { libc::killpg(pgrp.into(),
+                                  match signal.into() {
+                                      Some(s) => s as libc::c_int,
+                                      None => 0,
+                                  }) };
+
+    Errno::result(res).map(drop)
+}
+
 pub fn raise(signal: Signal) -> Result<()> {
     let res = unsafe { libc::raise(signal as libc::c_int) };
 
--- a/test/sys/test_signal.rs
+++ b/test/sys/test_signal.rs
@@ -10,6 +10,12 @@ fn test_kill_none() {
 }
 
 #[test]
+fn test_killpg_none() {
+    killpg(getpgrp(), None)
+        .expect("Should be able to send signal to my process group.");
+}
+
+#[test]
 fn test_old_sigaction_flags() {
     extern "C" fn handler(_: ::libc::c_int) {}
     let act = SigAction::new(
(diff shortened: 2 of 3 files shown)


src/sys/signal.rs
```diff
@@ -680,6 +680,16 @@ pub fn kill<T: Into<Option<Signal>>>(pid: ::unistd::Pid, signal: T) -> Result<()
     Errno::result(res).map(drop)
 }
 
+pub fn killpg<T: Into<Option<Signal>>>(pid: ::unistd::Pid, signal: T) -> Result<()> {
+    let res = unsafe { libc::killpg(pid.into(),
+                                  match signal.into() {
+                                      Some(s) => s as libc::c_int,
+                                      None => 0,
```
asomers: Why would you ever want to send 0?  Is that even valid?

DanSnow: It's valid. It just perform error checking and won't send any signal.

According to the [POSIX spec of `killpg`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/killpg.html):

> If pgrp is greater than 1, killpg(pgrp, sig) shall be equivalent to kill(-pgrp, sig).

And in the [POSIX spec of `kill`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/kill.html):

> The signal to be sent is specified by sig and is either one from the list given in <signal.h> or 0. If sig is 0 (the null signal), error checking is performed but no signal is actually sent. The null signal can be used to check the validity of pid.

Also in the [manpage of `killpg`](http://man7.org/linux/man-pages/man3/killpg.3.html):

> On Linux, killpg() is implemented as a library function that makes the call kill(-pgrp, sig).

DanSnow:
Thanks for reviewing. Actually I just copy the implement of `nix::sys::signal::kill` and it doesn't have any document either. But I think that I can also copy the test of `kill`.

DanSnow:
@asomers I have add test & document. Can you review again please?

src/sys/signal.rs
```diff
@@ -680,6 +680,8 @@ pub fn kill<T: Into<Option<Signal>>>(pid: ::unistd::Pid, signal: T) -> Result<()
     Errno::result(res).map(drop)
 }
 
+/// Send a signal to a process group [(see
+/// killpg(3))](http://pubs.opengroup.org/onlinepubs/9699919799/functions/killpg.html).
```
asomers: The docs should say what the arguments mean.  It's not obvious why `signal` should be an `Option`.  Also, I see from the man page that `killpg` does something special when `pgrp` is 0.  Perhaps that should be an option too?

test/sys/test_signal.rs
```diff
@@ -9,6 +9,11 @@ fn test_kill_none() {
     kill(getpid(), None).expect("Should be able to send signal to myself.");
 }
 
+#[test]
+fn test_killpg_none() {
+    killpg(getpgrp(), None).expect("Should be able to send signal to my process group.");
```
asomers: 80 columns, please.

DanSnow:
@asomers I have update the document. Can you check again?

src/sys/signal.rs
```diff
@@ -680,6 +680,22 @@ pub fn kill<T: Into<Option<Signal>>>(pid: ::unistd::Pid, signal: T) -> Result<()
...
+///
+/// If `pgrp` less then or equal 1, the behavior is undefined.
```
asomers: In compiler-speak, "Undefined behavior" means something specific and very bad.  It would be better to say "platform-specific".

DanSnow: @asomers OK. I have updated it.

asomers:
Ok, looks good.  It just needs a squash now.

DanSnow:
Thanks again. I just squashed all commits.

asomers:
Ahh, it looks like the PR I merged yesterday now conflicts with your CHANGELOG entry.  I'm afraid you'll have to rebase.

DanSnow:
OK. I think I fixed it.

PR #1035, opened 2019-03-10
Implement extra traits for all types

Susurrus:
Now that I've gotten all the extra traits implemented for `libc`, they can be easily derived for `nix`'s types. Note that this requires a bump to the minimum supported Rust version, but it's still at least 2 versions behind, so it fits in with policy.

One thing I did notice is that we have an inconsistent approach to our newtypes, where some use a struct and some a tuple struct. We should be consistent here, and should probably use a tuple struct since the name of the single field is irrelevant. This style is already suggested in our `CONVENTIONS.md` doc, so this should be uncontroversial. I'll file a PR after this is merged adding that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Added `MSG_WAITALL` to `MsgFlags` in `sys::socket`.
   ([#1079](https://github.com/nix-rust/nix/pull/1079))
+- Implemented `Clone`, `Copy`, `Debug`, `Eq`, `Hash`, and `PartialEq` for most
+  types that support them. ([#1035](https://github.com/nix-rust/nix/pull/1035))
 
 ### Changed
 - Support for `ifaddrs` now present when building for Android.
   ([#1077](https://github.com/nix-rust/nix/pull/1077))
+- Minimum supported Rust version is now 1.25.0 
+  ([#1035](https://github.com/nix-rust/nix/pull/1035))
 
 ### Fixed
 ### Removed
@@ -35,8 +39,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   native equivalents like `u32.`
   ([#1072](https://github.com/nix-rust/nix/pull/1072/commits))
 
-- Minimum supported Rust version is now 1.25.0 
-  ([#1035](https://github.com/nix-rust/nix/pull/1035))
 ### Fixed
 - Fix the build on Android and Linux/mips with recent versions of libc.
   ([#1072](https://github.com/nix-rust/nix/pull/1072/commits))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   native equivalents like `u32.`
   ([#1072](https://github.com/nix-rust/nix/pull/1072/commits))
 
+- Minimum supported Rust version is now 1.25.0 
+  ([#1035](https://github.com/nix-rust/nix/pull/1035))
 ### Fixed
 - Fix the build on Android and Linux/mips with recent versions of libc.
   ([#1072](https://github.com/nix-rust/nix/pull/1072/commits))
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -3,7 +3,7 @@ use errno::Errno;
 use fcntl::{self, OFlag};
 use libc;
 use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
-use std::{ffi, fmt, ptr};
+use std::{ffi, ptr};
 use sys;
 
 #[cfg(target_os = "linux")]
@@ -25,6 +25,7 @@ use libc::{dirent, readdir_r};
 ///    * returns entries for `.` (current directory) and `..` (parent directory).
 ///    * returns entries' names as a `CStr` (no allocation or conversion beyond whatever libc
 ///      does).
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Dir(
     // This could be ptr::NonNull once nix requires Rust 1.25.
     *mut libc::DIR
@@ -82,21 +83,13 @@ impl AsRawFd for Dir {
     }
 }
 
-impl fmt::Debug for Dir {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_struct("Dir")
-            .field("fd", &self.as_raw_fd())
-            .finish()
-    }
-}
-
 impl Drop for Dir {
     fn drop(&mut self) {
         unsafe { libc::closedir(self.0) };
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Eq, Hash, PartialEq)]
 pub struct Iter<'d>(&'d mut Dir);
 
 impl<'d> Iterator for Iter<'d> {
@@ -132,10 +125,10 @@ impl<'d> Drop for Iter<'d> {
 /// A directory entry, similar to `std::fs::DirEntry`.
 ///
 /// Note that unlike the std version, this may represent the `.` or `..` entries.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Entry(dirent);
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum Type {
     Fifo,
     CharacterDevice,
@@ -196,15 +189,5 @@ impl Entry {
             libc::DT_SOCK => Some(Type::Socket),
             /* libc::DT_UNKNOWN | */ _ => None,
         }
-    }
-}
-
-impl fmt::Debug for Entry {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_struct("Entry")
-            .field("ino", &self.ino())
-            .field("file_name", &self.file_name())
-            .field("file_type", &self.file_type())
-            .finish()
     }
 }
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -27,8 +27,7 @@ use libc::{dirent, readdir_r};
 ///      does).
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Dir(
-    // This could be ptr::NonNull once nix requires Rust 1.25.
-    *mut libc::DIR
+    ptr::NonNull<libc::DIR>
 );
 
 impl Dir {
@@ -60,7 +59,8 @@ impl Dir {
             unsafe { libc::close(fd) };
             return Err(e);
         };
-        Ok(Dir(d))
+        // Always guaranteed to be non-null by the previous check
+        Ok(Dir(ptr::NonNull::new(d).unwrap()))
     }
 
     /// Returns an iterator of `Result<Entry>` which rewinds when finished.
@@ -79,13 +79,13 @@ unsafe impl Send for Dir {}
 
 impl AsRawFd for Dir {
     fn as_raw_fd(&self) -> RawFd {
-        unsafe { libc::dirfd(self.0) }
+        unsafe { libc::dirfd(self.0.as_ptr()) }
     }
 }
 
 impl Drop for Dir {
     fn drop(&mut self) {
-        unsafe { libc::closedir(self.0) };
+        unsafe { libc::closedir(self.0.as_ptr()) };
     }
 }
 
@@ -104,7 +104,7 @@ impl<'d> Iterator for Iter<'d> {
             // Probably fine here too then.
             let mut ent: Entry = Entry(::std::mem::uninitialized());
             let mut result = ptr::null_mut();
-            if let Err(e) = Errno::result(readdir_r((self.0).0, &mut ent.0, &mut result)) {
+            if let Err(e) = Errno::result(readdir_r((self.0).0.as_ptr(), &mut ent.0, &mut result)) {
                 return Some(Err(e));
             }
             if result == ptr::null_mut() {
@@ -118,7 +118,7 @@ impl<'d> Iterator for Iter<'d> {
 
 impl<'d> Drop for Iter<'d> {
     fn drop(&mut self) {
-        unsafe { libc::rewinddir((self.0).0) }
+        unsafe { libc::rewinddir((self.0).0.as_ptr()) }
     }
 }
 
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -29,20 +29,9 @@ libc_bitflags!{
 }
 
 #[repr(C)]
-#[derive(Clone, Copy)]
-#[allow(missing_debug_implementations)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct MqAttr {
     mq_attr: libc::mq_attr,
-}
-
-impl PartialEq<MqAttr> for MqAttr {
-    fn eq(&self, other: &MqAttr) -> bool {
-        let self_attr = self.mq_attr;
-        let other_attr = other.mq_attr;
-        self_attr.mq_flags == other_attr.mq_flags && self_attr.mq_maxmsg == other_attr.mq_maxmsg &&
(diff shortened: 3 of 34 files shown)


src/sys/socket/addr.rs
```diff
@@ -687,27 +595,6 @@ impl UnixAddr {
...
-        self.sun_path() == other.sun_path()
-    }
-}
-
-impl Eq for UnixAddr {
-}
-
-impl hash::Hash for UnixAddr {
-    fn hash<H: hash::Hasher>(&self, s: &mut H) {
-        ( self.0.sun_family, self.sun_path() ).hash(s)
```
asomers: Same thing here.  We shouldn't be hashing the bytes after NUL.

src/sys/socket/addr.rs
```diff
@@ -958,30 +779,10 @@ pub mod netlink {
...
     use libc::{sa_family_t, sockaddr_nl};
     use std::{fmt, mem};
-    use std::hash::{Hash, Hasher};
 
-    #[derive(Copy, Clone)]
+    #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
     pub struct NetlinkAddr(pub sockaddr_nl);
 
-    // , PartialEq, Eq, Debug, Hash
-    impl PartialEq for NetlinkAddr {
```
asomers: The derived PartialEq and Hash will use the sockaddr_nl.nl_pad field, which is probably not what we want.

src/sys/socket/addr.rs
```diff
@@ -1035,28 +832,10 @@ pub mod sys_control {
...
 
-    #[derive(Copy, Clone)]
     #[repr(C)]
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
     pub struct SysControlAddr(pub libc::sockaddr_ctl);
 
-    impl PartialEq for SysControlAddr {
-        fn eq(&self, other: &Self) -> bool {
-            let (inner, other) = (self.0, other.0);
-            (inner.sc_id, inner.sc_unit) ==
```
asomers: This is weird.  Why did the old implementation ignore the sc_len, sc_family, and sc_sysaddr fields?

src/mqueue.rs
```diff
@@ -29,22 +29,11 @@ libc_bitflags!{
...
 
 #[repr(C)]
-#[derive(Clone, Copy)]
-#[allow(missing_debug_implementations)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct MqAttr {
     mq_attr: libc::mq_attr,
 }
 
-impl PartialEq<MqAttr> for MqAttr {
```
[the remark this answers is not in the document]
Susurrus: I'm not certain we do, however I think we should tackle that in `libc` itself. We are seeing test failures due to this difference.

euclio:
Would it be possible to split out just the `Debug` portion of this PR if the rest of the traits are blocked on upstream libc?

Susurrus:
Now that rust-lang/libc#1371 was merged, it's time to come back to this. That should have resolved most of the issues. Gonna test this through CI then do another pass through @asomers' comments.

src/sys/socket/addr.rs
Susurrus: Dunno, it's possible this code is from when that struct was still defined in `nix` and only had these two fields as they were all that was required. Regardless, I think the `libc` implementation is now the correct one.

src/sys/socket/addr.rs
```diff
@@ -687,27 +595,6 @@ impl UnixAddr {
     }
 }
 
-impl PartialEq for UnixAddr {
-    fn eq(&self, other: &UnixAddr) -> bool {
-        self.sun_path() == other.sun_path()
```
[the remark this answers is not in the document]
Susurrus: So, I was reading up on `sockaddr_un` following [this link](http://man7.org/linux/man-pages/man7/unix.7.html). I don't know if we need to worry about the 2nd condition, where there is no path, and it's trivial to compare these. However, I'm not certain if this makes sense to implement in `libc`, or if this implementation should really live in `nix`. I don't know how much `libc`  wants to have standards logic in their code.

However, looking at our code, it doesn't appear that we properly extract the `sun_path` for abstract sockets via `UnixAddr::sun_path()`.

We have some tests involving the `sun_path` in `test/sys/test_socket.rs`, however they don't test the hash or equality functions of this struct. I'm thinking before continuing a discussion, I should add some tests for that. Do you agree? I'm looking at it however and I don't immediately know how these tests should be written. So if you have any ideas, @asomers, I'd love to hear them. Otherwise I'll take a stab at these.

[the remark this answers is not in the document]
asomers: Yeah, we should definitely be testing equality for abstract sockets.  I would suggest just manually creating some so we can get all of the edge cases.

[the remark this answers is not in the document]
asomers: Maybe we should suggest that libc not implement PartialEq or Hash for sockaddr_un, because it's so weird.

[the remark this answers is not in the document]
Susurrus: > Yeah, we should definitely be testing equality for abstract sockets. I would suggest just manually creating some so we can get all of the edge cases.

Turns out we have some tests that check the paths for regular and abstract sockets. I was thinking of extending them, but I'm not certain the right way to do it. I was going to create a `u8` array of 0xFF the same size of the struct, transmute it to a `UnixAddr` and then manually set the `sockaddr_un` and `usize` fields. This will have a bunch of custom logic to set those fields the way I think they should be set.

What I would prefer to do is to initialize a `UnixAddr`, then set the stack to contain `0xFF` for a large chunk of memory, and initialize an identical `UnixAddr`, and then compare them for equality and hashing. That should test it very effectively and remove all of the complexity of stuffing data in correctly by hand. Do you know of a way to do that with Rust?

> Maybe we should suggest that libc not implement PartialEq or Hash for sockaddr_un, because it's so weird.

I'm not going to bother with this. There's no promise made there that equality isn't just binary equality, but Nix definitely implies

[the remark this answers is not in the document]
asomers: > > Yeah, we should definitely be testing equality for abstract sockets. I would suggest just manually creating some so we can get all of the edge cases.
> 
> Turns out we have some tests that check the paths for regular and abstract sockets. I was thinking of extending them, but I'm not certain the right way to do it. I was going to create a `u8` array of 0xFF the same size of the struct, transmute it to a `UnixAddr` and then manually set the `sockaddr_un` and `usize` fields. This will have a bunch of custom logic to set those fields the way I think they should be set.
> 
> What I would prefer to do is to initialize a `UnixAddr`, then set the stack to contain `0xFF` for a large chunk of memory, and initialize an identical `UnixAddr`, and then compare them for equality and hashing. That should test it very effectively and remove all of the complexity of stuffing data in correctly by hand. Do you know of a way to do that with Rust?

That sounds cumbersome.  Instead of modifying those tests, why not write new tests directly in src/sys/socket/addr.rs?  That way you'll be able to access the underlying libc::sockaddr_un directly.

Susurrus:
Sorry I wasn't clear. I was planning to write new tests. The problem is that I dont know how to write them. I can handcraft sockaddr_un structs to test equality, but that seems like I'm coding the same logic in the test and I'm testing for and I dont think it'll provide much value. Or I can modify the stack so that all values are nonzero and use UnixAddr directly to do the comparison. I think the latter would be a better implementation, but I don't know how to do it.

On Wed, Jun 5, 2019, at 8:44 AM, Alan Somers wrote:
> *@asomers* commented on this pull request.

> 
> In src/sys/socket/addr.rs <https://github.com/nix-rust/nix/pull/1035#discussion_r290808764>:

> > @@ -687,27 +595,6 @@ impl UnixAddr {
     }
 }
 
-impl PartialEq for UnixAddr {
-    fn eq(&self, other: &UnixAddr) -> bool {
-        self.sun_path() == other.sun_path()

asomers:
> Sorry I wasn't clear. I was planning to write new tests. The problem is that I dont know how to write them. I can handcraft sockaddr_un structs to test equality, but that seems like I'm coding the same logic in the test and I'm testing for and I dont think it'll provide much value. Or I can modify the stack so that all values are nonzero and use UnixAddr directly to do the comparison. I think the latter would be a better implementation, but I don't know how to do it.

I think the most straightforward way would be to handcraft a `libc::sockaddr_un`, cast to a `libc::sockaddr_storage`, convert to a `UnixAddr` via `sockaddr_storage_to_addr`, then compare `UnixAddr`s.  That way you'll have byte-level control of the data using only the regular API.

Susurrus:
Alright, I implemented something that I think is a reasonable test. Tests passed on my local computer, so if CI passed, I think this is good to go.

Susurrus:
Looks like `PTRACE_ATTACH` was changed in a commit at some point I guess and now it can't be found for `mips64` targets. However, I'm not finding anywhere where that changed in `libc` or `nix`. So that's it for me tonight, I'll need to come back and dig in to it again later.

asomers:
> Looks like `PTRACE_ATTACH` was changed in a commit at some point I guess and now it can't be found for `mips64` targets. However, I'm not finding anywhere where that changed in `libc` or `nix`. So that's it for me tonight, I'll need to come back and dig in to it again later.

Yeah, that one stumped me too.  I installed the exact mips64 target for the 1.24.1 toolchain and checked that I'm building against the exact same revision of libc as Travis used, but for me it works.

Susurrus:
Figured out I wasn't synced to HEAD and 4bd419ebec2d691f55331e91e799d5fa96180044 was the problem. Filed rust-lang/libc#1393 to resolve.

Susurrus:
Alright, looks like we should be all good to go here now that my LINC PR was merged!

test/sys/test_socket.rs
```diff
@@ -66,6 +68,37 @@ pub fn test_path_to_sock_addr() {
...
+    assert_eq!(calculate_hash(&addr1), calculate_hash(&addr2));
+}
+
+#[cfg(target_os = "linux")]
+#[test]
+pub fn test_addr_equality_abstract() {
+    let name = String::from("nix\0abstract\0test");
+    let addr1 = UnixAddr::new_abstract(name.as_bytes()).unwrap();
+    let mut addr2 = addr1.clone();
+    addr2.0.sun_path[18] = 127;
```
asomers: Why should these two addrs be equal?  I thought that in an abstract socket, all bytes were relevant?

src/sys/signal.rs
```diff
@@ -788,28 +784,6 @@ mod sigevent {
         }
     }
 
-    impl Debug for SigEvent {
```
asomers: I think `SigEvent`'s manual `Debug` implementation is better than the derived one, because it won't print unused fields.

src/sys/aio.rs
```diff
@@ -1018,22 +1019,6 @@ pub fn aio_suspend(list: &[&AioCb], timeout: Option<TimeSpec>) -> Result<()> {
     }).map(drop)
 }
 
-impl<'a> Debug for AioCb<'a> {
```
asomers: I think `AioCb`'s manual `Debug` implementation is better than the derived one, because it won't print unused fields.  It also doesn't print the `Buffer`, which is really just to track ownership.

src/dir.rs
```diff
@@ -25,6 +25,7 @@ use libc::{dirent, readdir_r};
 ///    * returns entries for `.` (current directory) and `..` (parent directory).
 ///    * returns entries' names as a `CStr` (no allocation or conversion beyond whatever libc
 ///      does).
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Dir(
     // This could be ptr::NonNull once nix requires Rust 1.25.
```
asomers: May as well change this since you're raising the MSRV.

CHANGELOG.md
```diff
@@ -23,12 +23,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
   ioctl_read_bad!(tcgets, libc::TCGETS, libc::termios);
   ```
+- Implemented `Clone`, `Copy`, `Debug`, `Eq`, `Hash`, and `PartialEq` for most
```
asomers: Your CHANGELOG entry got dislocated.

src/sys/signal.rs
Susurrus: I'm thinking this should be moved into `libc`, so I'll need to submit another PR to do that.

src/sys/socket/addr.rs
```diff
@@ -243,7 +243,7 @@ impl AddressFamily {
     }
 }
 
-#[derive(Copy)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum InetAddr {
     V4(libc::sockaddr_in),
```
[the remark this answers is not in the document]
Susurrus: Broke off into #1082

src/sys/signal.rs
Susurrus: Opened rust-lang/libc#1394 to move this there

test/sys/test_socket.rs
Susurrus: It actually looks like our test code was wrong here, as it truncated abstract address names to the end of whatever string was passed in, when really it should be the entire array (minus the first byte). So I've also modified that behavior here. Need to add a CHANGELOG entry if you agree this new implementation is the correct one.

src/sys/signal.rs
Susurrus: That PR's been merged so resolving.

Susurrus:
Alright, I think we're all good with this one. Libc changes have all been merged and tests are all padsing

src/dir.rs
```diff
@@ -59,7 +59,7 @@ impl Dir {
             unsafe { libc::close(fd) };
             return Err(e);
         };
-        Ok(Dir(d))
+        Ok(Dir(unsafe { ptr::NonNull::new_unchecked(d) }))
```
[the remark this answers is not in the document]
Susurrus: This is guaranteed to be safe given that we check for `NULL` right above. Do you think it's more clear this way or the way you suggest?

test/sys/test_socket.rs
```diff
@@ -66,26 +68,65 @@ pub fn test_path_to_sock_addr() {
...
+}
+
 // Test getting/setting abstract addresses (without unix socket creation)
 #[cfg(target_os = "linux")]
 #[test]
 pub fn test_abstract_uds_addr() {
     let empty = String::new();
     let addr = UnixAddr::new_abstract(empty.as_bytes()).unwrap();
-    assert_eq!(addr.as_abstract(), Some(empty.as_bytes()));
+    let sun_path = [0u8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```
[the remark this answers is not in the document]
Susurrus: The default for `rustfmt` is 100 as the maximum line length and 80 as the ideal line length. There's no clear RFC accepted for this convention, however, so it could change in the future. We don't have a spec specified in our `CONVENTIONS` doc either. Where possible, I'd suggest we stick to `rustfmt`s defaults (however we could create a `rustfmt.toml` if we want to alter them, but I'd like to do this *and* wrap these in this PR rather than just wrapping this as a one-off.

src/dir.rs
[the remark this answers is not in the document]
asomers: I think it would look better (and raise fewer eyeballs) done the way that doesn't involve `unsafe`.

test/sys/test_socket.rs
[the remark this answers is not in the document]
asomers: Sure, there's an argument to be made for sticking to 100 columns.  But this line has 343!  Can we fix it now?  In fact, rather than wrap the columns, it would make sense to use the `let sun_path = [0u8; X]` syntax to define an array of identical members.

Susurrus:
Hopefully this is the final necessary batch of changes!

PR #1036, opened 2019-03-15
Add several errno values for OpenBSD

sourgrasses:
It looks like OpenBSD added several new errno values in 6.2, and they've already been added to `libc`, so this should allow any crates using those to build on OpenBSD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1034](https://github.com/nix-rust/nix/pull/1034))
 - Added ENOTSUP errno support for Linux and Android.
   ([#969](https://github.com/nix-rust/nix/pull/969))
+- Add several errno constants from OpenBSD 6.2
+  ([#1036](https://github.com/nix-rust/nix/pull/1036))
 
 ### Changed
 - `PollFd` event flags renamed to `PollFlags` ([#1024](https://github.com/nix-rust/nix/pull/1024/))
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -360,7 +360,8 @@ fn desc(errno: Errno) -> &'static str {
         #[cfg(any(target_os = "linux", target_os = "android"))]
         EDQUOT          => "Quota exceeded",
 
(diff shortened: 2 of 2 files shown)


asomers:
Yikes, those are some long lines!  While you're here, would you mind wrapping them to 80 columns?

sourgrasses:
Ha, yeah, I was wondering about that and decided just to mimic the line above. Should I just `rustfmt` that whole file with a max width of 80? Those long lines come out like this:
```rust
#[cfg(any(target_os = "macos", target_os = "freebsd",
          target_os = "dragonfly", target_os = "ios",
          target_os = "openbsd", target_os = "netbsd"))]
EPROTO => "Protocol error",
```
It wraps some of the other cfg lines also in more or less the same style.

asomers:
I'm not a fan of everything that rustfmt does.  I wouldn't use it on parts of the file that you aren't changing.

sourgrasses:
Sounds good. Should I go ahead and wrap those other really long lines I was mimicking?

rennis250:
Is it mainly formatting that is holding this back? It would be helpful to have this available, because newer versions of some packages I depend on already check for EPROTO and won't compile on OpenBSD, forcing me to stay with older versions.

Thanks,
Rob

rennis250:
Sorry if I came off demanding. I was merely just inquiring. Thanks! :-)

--Rob

rennis250:
Same! :)

https://github.com/Smithay/wayland-rs/issues/259

asomers:
Thanks for handling the formatting.  Could you please do two more things?
1) Add a CHANGELOG entry that says something like "Added new errno constants from OpenBSD 6.2"
2) Squash the commits.

sourgrasses:
👍

asomers:
Looks like Travis is super-slow today :'( .  I'll try again late at night.

asomers:
Sorry, but merging master isn't the right way to resolve the merge conflict.  You need to rebase instead.

sourgrasses:
Yeahhh, whoops lol.

rennis250:
Thanks!

PR #1041, opened 2019-04-02
allow importing macros without helpers in the 2018 edition

euclio:
~This PR adds the `$crate::` prefix to all uses of helper macros, which will allow users of the 2018 edition to import the macros without needing to also import their helpers.~

See https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-with-crate-prefix for more info.

~While this is a cleaner solution than using `#[macro_export(local_inner_macros)]`, it also requires bumping the minimum supported compiler version to 1.30. If this is acceptable, I'll add a commit to update the Travis configuration and docs.~

This PR now uses `local_inner_macros` to preserve compatibility with the current minimum Rust version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,18 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Fixed multiple bugs when using `sendmsg` and `recvmsg` with ancillary control messages
   ([#1020](https://github.com/nix-rust/nix/pull/1020))
+- Macros exported by `nix` may now be imported via `use` on the Rust 2018
+  edition without importing helper macros.
+  ([#1041](https://github.com/nix-rust/nix/pull/1041))
+
+  For example, in Rust 2018, the `ioctl_read_bad!` macro can now be imported
+  without importing the `convert_ioctl_res!` macro.
+
+  ```rust
+  use nix::ioctl_read_bad;
+
+  ioctl_read_bad!(tcgets, libc::TCGETS, libc::termios);
+  ```
 
 ### Removed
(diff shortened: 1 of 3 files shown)


asomers:
Nix has traditionally been pretty conservative about raising the minimum compiler version, and this is a big bump.  It could make Nix unusable by some distros' Rust packages.  Worse, it might break platforms not supported by rustup that don't yet have compiler 1.30.0.  Do you have any data about how widespread this compiler is?

Also, how did you tell github that this PR is a draft?  I haven't seen that feature before.

euclio:
I don't have any data for the usage of 1.30.0, unfortunately. If we're not comfortable bumping to 1.30.0, I can see how feasible using `local_inner_macros` is.

When you create a PR, you can click a dropdown next to the PR submission button to mark a PR as "Draft" :)

euclio:
Actually, here is some data on 1.30.0 usage: https://repology.org/project/rust/versions

euclio:
@asomers I'm now using `local_inner_macros`, which should preserve compatibility with the current minimum Rust version.

asomers:
Thanks for making that change.  And thanks for the Rust versions link; I didn't know about that.  But this needs a CHANGELOG entry.  Can you please add one and describe how the visibility of macros has changed?

euclio:
Added an entry to the changelog.

asomers:
If the helper macros are no longer importable, then they should be mentioned in the "Removed" section of the changelog.

euclio:
They're still importable, it's just that you don't need to explicitly import them to use macros that use them internally.

euclio:
Friendly ping @asomers

PR #1042, opened 2019-04-09
pty: Add forkpty

keur:


keur:
Note: `forkpty` re-added to linux in rust-lang/libc#1313

keur:
Err... failed CI for OSX but it passed all the tests?

asomers:
It hung on OSX.  Probably a bug in `test_forkpty`.  It also appears to be hanging on FreeBSD.  I'll terminate it.

keur:
@asomers yeah... i think i relied on linux kernel for how it handles a `read` after `write`. will fix in a bit.

keur:
@asomers Got CI passing. Take a look

test/test_pty.rs
```diff
@@ -201,3 +201,33 @@ fn test_openpty_with_termios() {
...
+    let _m1 = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+    let string = "naninani\n";
+    let echoed_string = "naninani\r\n";
+    let pty = forkpty(None, None).unwrap();
+    match pty.fork_result {
+        Child => {
+            write(0, string.as_bytes()).unwrap();
+            pause();  // we need the child to stay alive until the parent calls read
+        },
```
asomers: In a Rust test, you cannot return from a forked child.  That can cause deadlocks.  Instead, you must `_exit(0)`.  The `SIGTERM` will _probably_ suffice, but I would `_exit(0)` just in case.

src/pty.rs
```diff
@@ -266,3 +280,76 @@ pub fn openpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>
...
+
+/// Create a new pseudoterminal, returning the master file descriptor and forked pid.
+/// in `ForkptyResult`
+/// (see [`forkpty`](http://man7.org/linux/man-pages/man3/forkpty.3.html)).
+///
+/// If `winsize` is not `None`, the window size of the slave will be set to
+/// the values in `winsize`. If `termios` is not `None`, the pseudoterminal's
+/// terminal settings of the slave will be set to the values in `termios`.
+#[inline]
+pub fn forkpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>>>(winsize: T, termios: U) -> Result<ForkptyResult> {
```
asomers: 80 cols, please.

src/pty.rs
```diff
@@ -266,3 +280,76 @@ pub fn openpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>
...
 }
+
+/// Create a new pseudoterminal, returning the master file descriptor and forked pid.
+/// in `ForkptyResult`
+/// (see [`forkpty`](http://man7.org/linux/man-pages/man3/forkpty.3.html)).
+///
+/// If `winsize` is not `None`, the window size of the slave will be set to
+/// the values in `winsize`. If `termios` is not `None`, the pseudoterminal's
+/// terminal settings of the slave will be set to the values in `termios`.
+#[inline]
```
asomers: Why inline such a big function?  In fact, it's rarely a good idea to use `#[inline]` at all.  Better to let the compiler decide.

src/pty.rs
```diff
@@ -266,3 +280,76 @@ pub fn openpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>
...
+pub fn forkpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>>>(winsize: T, termios: U) -> Result<ForkptyResult> {
+    use std::ptr;
+    use unistd::Pid;
+    use unistd::ForkResult::*;
+
+    let mut master: libc::c_int = unsafe { mem::uninitialized() };
+    let res = {
+        match (termios.into(), winsize.into()) {
```
asomers: I think you could considerably shorten this block by using something like `termios.into().unwrap_or(ptr::null_mut())`.

src/pty.rs
```diff
@@ -26,6 +27,19 @@ pub struct OpenptyResult {
     pub slave: RawFd,
 }
 
+/// Representation of a master with a forked pty
+///
+/// This is returned by `forkpty`. Note that this type does *not* implement `Drop`, so the user
+/// must manually close the file descriptors.
+#[derive(Clone, Copy)]
+#[allow(missing_debug_implementations)]
```
asomers: Why not derive `Debug`?

test/test_pty.rs
```diff
@@ -201,3 +201,33 @@ fn test_openpty_with_termios() {
...
+    let _m0 = ::PTSNAME_MTX.lock().expect("Mutex got poisoned by another test");
+    // forkpty spawns a child process
+    let _m1 = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+    let string = "naninani\n";
+    let echoed_string = "naninani\r\n";
+    let pty = forkpty(None, None).unwrap();
+    match pty.fork_result {
+        Child => {
+            write(0, string.as_bytes()).unwrap();
```
asomers: Yikes!  Can you eliminate that magic number?

keur:
@asomers Made the suggested changes.

keur:
@asomers Thanks for the comments. Since the initial code was based largely off of `openpty` a lot of the review comments could be applied to that function; I am happy to make another PR to refactor that.

keur:
@asomers anything left to fix for this?

src/pty.rs
```diff
@@ -266,3 +279,49 @@ pub fn openpty<'a, 'b, T: Into<Option<&'a Winsize>>, U: Into<Option<&'b Termios>
...
+    };
+
+    let fork_result = Errno::result(res).map(|res| match res {
+        0 => Child,
+        res => Parent { child: Pid::from_raw(res) },
+    })?;
+
+    Ok(ForkptyResult {
+        master: master,
+        fork_result: fork_result,
```
asomers: If `libc::forkpty` fails, then `master` will be uninitialized here.  For that matter, the return value will be wrong, too.  You should ensure that it returns `Err` if `libc::forkpty` fails.

keur: `Errno::result(res)` should convert to an `Err`, and `?` on line 320 should propagate the error.

asomers: Ahh, I didn't see that little `?`.

keur:
@asomers done

PR #1045, opened 2019-04-19
Add a access(2) wrapper

snsmac:


src/unistd.rs
```diff
@@ -2334,3 +2334,29 @@ mod setres {
...
+        /// Test for write permission.
+        W_OK;
+        /// Test for execute (search) permission.
+        X_OK;
+    }
+}
+
+/// Checks the file named by `path` for accessibility according to the flags given by `amode`
+/// See [access(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/access.html)
+#[inline]
```
asomers: No need for `#[inline]`.  It's better to allow the compiler to decide.

test/test_unistd.rs
```diff
@@ -574,3 +574,22 @@ fn test_symlinkat() {
...
+               nix::errno::Errno::ENOENT);
+}
+
+#[test]
+fn test_access_file_exists() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let path  = tempdir.path().join("does_exist.txt");
+    let _file = File::create(path.clone()).unwrap();
+    assert_eq!(
+        access(&path, AccessFlags::F_OK | AccessFlags::R_OK | AccessFlags::W_OK).unwrap(),
```
asomers: Instead of unwrapping, just should just `assert!(access(...).is_ok())`

src/unistd.rs
snsmac: Ok, i did this because many other functions in unistd.rs also use inline

asomers: Yeah, it's an anachronism that we haven't fully purged.

snsmac:
@asomers I made the changes you requested

PR #1048, opened 2019-04-27
Restored the bytes field on RecvMsg

vdagonneau:
This pull request restores the `bytes` field on the `RecvMsg` structure in order to be able to know the amount of bytes read during a call to `recvmsg`.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -417,6 +417,7 @@ impl CmsgBuffer for Vec<u8> {
 
 #[allow(missing_debug_implementations)] // msghdr isn't Debug
 pub struct RecvMsg<'a> {
+    pub bytes: usize,
     cmsghdr: Option<&'a cmsghdr>,
     pub address: Option<SockAddr>,
     pub flags: MsgFlags,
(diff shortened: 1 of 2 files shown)


src/sys/socket/mod.rs
```diff
@@ -986,6 +987,7 @@ pub fn recvmsg<'a>(fd: RawFd, iov: &[IoVec<&mut [u8]>],
             sockaddr_storage_to_addr(&address, mhdr.msg_namelen as usize).ok()
         };
         RecvMsg {
+            bytes: ret as usize,
```
asomers: While this will technically work, it's bad style because it won't work for other data types.  Instead of using `ret`, you should use the argument provided to `map`, like this:
```
Errno::result(ret).map(|r| {
    ...
    bytes: r as usize
}
```

asomers:
Ok, that looks good.  But please squash it.  BTW, ordinarly nix requires CHANGELOG entries for most PRs.  But this is an exception, because it fixes a bug that was introduced since the last release.  So no CHANGELOG entry is required.

vdagonneau:
I squashed it.

PR #1052, opened 2019-05-16
Implement unlinkat

sunfishcode:
This adds the unlinkat function, which is part of POSIX:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/unlinkat.html
and widely implmented on Unix-family platforms.


sunfishcode:
Oops, I accidentally hit the wrong button!

PR #1058, opened 2019-05-21
Implement unlinkat

jlb6740:
This adds the unlinkat function, which is part of POSIX:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/unlinkat.html
and widely implmented on Unix-family platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1089](https://github.com/nix-rust/nix/pull/1089))
 - Added `AF_VSOCK` to `AddressFamily`.
   ([#1091](https://github.com/nix-rust/nix/pull/1091))
+- Add `unlinkat`
+  ([#1058](https://github.com/nix-rust/nix/pull/1058))
 
 ### Changed
 - Support for `ifaddrs` now present when building for Android.
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1144,6 +1144,42 @@ pub fn unlink<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Flags for `unlinkat` function.
+#[derive(Clone, Copy, Debug)]
+pub enum UnlinkatFlags {
+    RemoveDir,
+    NoRemoveDir,
+}
+
+/// Remove a directory entry
+///
+/// In the case of a relative path, the directory entry to be removed is determined relative to
+/// the directory associated with the file descriptor `dirfd` or the current working directory
+/// if `dirfd` is `None`. In the case of an absolute `path` `dirfd` is ignored. If `flag` is
+/// `UnlinkatFlags::RemoveDir` then removal of the directory entry specified by `dirfd` and `path`
+/// is performed.
+///
+/// # References
+/// See also [unlinkat(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/unlinkat.html)
+pub fn unlinkat<P: ?Sized + NixPath>(
+    dirfd: Option<RawFd>,
+    path: &P,
+    flag: UnlinkatFlags,
+) -> Result<()> {
+    let atflag =
+        match flag {
+            UnlinkatFlags::RemoveDir => AtFlags::AT_REMOVEDIR,
+            UnlinkatFlags::NoRemoveDir => AtFlags::empty(),
+        };
+    let res = path.with_nix_path(|cstr| {
+        unsafe {
+            libc::unlinkat(at_rawfd(dirfd), cstr.as_ptr(), atflag.bits() as libc::c_int)
+        }
+    })?;
+    Errno::result(res).map(drop)
+}
+
+
 #[inline]
 pub fn chroot<P: ?Sized + NixPath>(path: &P) -> Result<()> {
     let res = path.with_nix_path(|cstr| {
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -1,13 +1,14 @@
-use nix::fcntl::{fcntl, FcntlArg, FdFlag, open, OFlag, readlink};
+use nix::fcntl::{self, fcntl, FcntlArg, FdFlag, open, OFlag, readlink};
 use nix::unistd::*;
 use nix::unistd::ForkResult::*;
 use nix::sys::signal::{SaFlags, SigAction, SigHandler, SigSet, Signal, sigaction};
 use nix::sys::wait::*;
 use nix::sys::stat::{self, Mode, SFlag};
 use nix::errno::Errno;
+use nix::Error;
 use std::{env, iter};
 use std::ffi::CString;
-use std::fs::{self, File};
+use std::fs::{self, DirBuilder, File};
 use std::io::Write;
 use std::os::unix::prelude::*;
 use tempfile::{self, tempfile};
@@ -598,6 +599,58 @@ fn test_symlinkat() {
         target
     );
 }
+
+
+#[test]
+fn test_unlinkat_dir_noremovedir() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let dirname = "foo_dir";
+    let dirpath = tempdir.path().join(dirname);
+
+    // Create dir
+    DirBuilder::new().recursive(true).create(&dirpath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink dir at relative path without proper flag
+    let err_result = unlinkat(Some(dirfd), dirname, UnlinkatFlags::NoRemoveDir).unwrap_err();
+    assert!(err_result == Error::Sys(Errno::EISDIR) || err_result == Error::Sys(Errno::EPERM));
+ }
+
+#[test]
+fn test_unlinkat_dir_removedir() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let dirname = "foo_dir";
+    let dirpath = tempdir.path().join(dirname);
+
+    // Create dir
+    DirBuilder::new().recursive(true).create(&dirpath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink dir at relative path with proper flag
+    unlinkat(Some(dirfd), dirname, UnlinkatFlags::RemoveDir).unwrap();
+    assert!(!dirpath.exists());
+ }
+
+#[test]
+fn test_unlinkat_file() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let filename = "foo.txt";
+    let filepath = tempdir.path().join(filename);
+
+    // Create file
+    File::create(&filepath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink file at relative path
+    unlinkat(Some(dirfd), filename, UnlinkatFlags::NoRemoveDir).unwrap();
+    assert!(!filepath.exists());
+ }
 
 #[test]
 fn test_access_not_existing() {
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -23,6 +23,7 @@ pub use self::posix_fadvise::*;
 
 libc_bitflags!{
     pub struct AtFlags: c_int {
+        AT_REMOVEDIR;
         AT_SYMLINK_NOFOLLOW;
         #[cfg(any(target_os = "android", target_os = "linux"))]
         AT_NO_AUTOMOUNT;


src/unistd.rs
```diff
@@ -1144,6 +1144,40 @@ pub fn unlink<P: ?Sized + NixPath>(path: &P) -> Result<()> {
...
+#[derive(Clone, Copy, Debug)]
+pub enum UnlinkatFlags {
+    RemoveDir,
+    NoRemoveDir,
+}
+
+/// Remove a directory entry
+///
+/// In the case of a relative path, the directory entry to be removed is determined relative to
+/// the directory associated with the file descriptor dirfd. If flag is UnlinkatFlags::RemoveDir
```
asomers: Symbols like `dirfd`, `flag`, and `UnlinkatFlags::RemoveDir` should be in backticks.  The doc comment should also explain what it means for `dirfd` to be `None`.

test/test_unistd.rs
```diff
@@ -576,6 +577,39 @@ fn test_symlinkat() {
...
+
+    // Create dir
+    DirBuilder::new().recursive(true).create(&dirpath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink dir at relative path without proper flag
+    panic::set_hook(Box::new(|_info| {}));
+    let result = std::panic::catch_unwind(|| unlinkat(Some(dirfd), dirname, UnlinkatFlags::NoRemoveDir).unwrap());
```
asomers: This is way too complicated.  The right way to test code that ought to panic is with the `#[should_panic()]` attribute.  See sys/test_aio.rs for an example.

test/test_unistd.rs
```diff
@@ -576,6 +577,39 @@ fn test_symlinkat() {
...
+    assert!(result.is_err());
+    assert!(dirpath.exists());
+
+    // Attempt unlink dir at relative path with proper flag
+    unlinkat(Some(dirfd), dirname, UnlinkatFlags::RemoveDir).unwrap();
+    assert!(!dirpath.exists());
+
+    // Attempt unlink file at relative path
+    unlinkat(Some(dirfd), filename, UnlinkatFlags::NoRemoveDir).unwrap();
```
asomers: A good unit test should do only the minimum.  You should break this test up into 3: the panic test, the successful remove file test, and the successful remove dir test.

test/test_unistd.rs
jlb6740: Hi asomers .. thanks a lot for the feedback. I am definitely new to rust so I appreciate the suggestions. Being fairly new though I did spend time to research the best way to do this and this was the suggestion I came across. Specifically this was suggested over should_panic for more fine grain panic control. For testing this code I wanted the test code to create a folder (and file) and then try to delete the folder first without the proper flag and then try again with. If I use [should_panic()] its my understanding no unexpected panics in the code function would be properly caught hence this suggestion: https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/controlling-panics-with-std-panic.html. Any comments on this concern of mine? Looks like below you may be suggesting three different commits which I guess would mitigate my concern of using [should_panic()] where I don't want that coarse grain coverage of the attribute?

asomers: Yes, if you split the test up, then some tests would have `should_panic` and others wouldn't.  Why do you prefer `std::panic` over the simpler `should_panic`?

jlb6740: Hi .. I simply did not want to use an attribute that I realized could hide other unexpected exceptions that should not happen. Using std:panic since was the only suggested way I saw to handle this, plus it was just cutting and pasting two lines of code. Its all unfamiliar syntax for me so it didn't stand out to me as anymore complicated than some of the other lines. Certainly using should_panic should be easier to follow. If I break up this test case as you suggest I am creating more folders/files and using more lines of code but I can then use should_panic. I didn't think to do this initially since I guess I was trying to capture the testing of unlinkat in a single function instead of multiple.

asomers: Actually, `should_panic` is more specific than you realize.  You can provide an expected string, and the test will fail if it panics with any other message.  See test/sys/test_aio.rs for an example.

jlb6740:
@asomers I am a little uncertain as to the github way to amend a pull request. Instead of amending and doing a force push I instead followed some fairly old instructions here https://stackoverflow.com/questions/9790448/how-to-update-a-pull-request-from-forked-repo and pushed the patch. I am able to see this second commit. I assume this is all part of the same original pull request but please advice if there is more that I should do to have this latest patch reviewed. Thanks.

asomers:
The easiest way is to amend and/or rebase your changes locally, and then do a force push, like this:
```
git checkout -b mybranch
# Edit some files
git commit
git push -u git@github.com:myusername/myproject.git
# Edit some more
git amend
git push -f
```

jlb6740:
@Asomers .. Ok. I was afraid that would erase comment history asking for the update but I wasn't sure and also I wasn't sure if it mattered to the project. The update https://github.com/nix-rust/nix/pull/1058/commits/ce88eddff21e343860949c899c1286584128217c is here but I don't see a way to ask for a new pull request so I assume you can see and are notified of it? Let me know if you need me to somehow resubmit. Thanks.

test/test_unistd.rs
```diff
@@ -576,6 +577,60 @@ fn test_symlinkat() {
     );
 }
 
+
+#[test]
+#[should_panic]
```
asomers: You should the `expected("whatever")` string to `should_panic` to ensure that it's only catching the desired panic, and no others.

test/test_unistd.rs
```diff
@@ -576,6 +577,60 @@ fn test_symlinkat() {
...
+
+    // Create dir
+    DirBuilder::new().recursive(true).create(&dirpath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink dir at relative path without proper flag
+    unlinkat(Some(dirfd), dirname, UnlinkatFlags::NoRemoveDir).unwrap();
+    assert!(dirpath.exists());
```
asomers: This line is dead code, assuming the line above panics.

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1045](https://github.com/nix-rust/nix/pull/1045))
 - Add `forkpty`
   ([#1042](https://github.com/nix-rust/nix/pull/1042))
+- Add `unlinkat`
+  ([#TBD](https://github.com/nix-rust/nix/pull/TBD))
```
asomers: Don't forget to replace the TBD.

test/test_unistd.rs
jlb6740: Hi ... Thanks for the tip. Update #3 includes this change.

test/test_unistd.rs
jlb6740: Thanks. You are right. Not sure why I left this in but it is now removed.

CHANGELOG.md
asomers: You missed a TBD.

test/test_unistd.rs
```diff
@@ -576,6 +577,59 @@ fn test_symlinkat() {
...
+    let dirpath = tempdir.path().join(dirname);
+
+    // Create dir
+    DirBuilder::new().recursive(true).create(&dirpath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink dir at relative path without proper flag
+    unlinkat(Some(dirfd), dirname, UnlinkatFlags::NoRemoveDir).unwrap();
```
asomers: You don't need `should_panic` for this.  Just replace the `unwrap` with an `unwrap_err` and check the result.

jlb6740:
@asomers I tried to be more specific by matching on the error message, but there is a discrepancy in the error returned by i686-appledarwin and x86_64-apple-darwin "Sys(EPERM)", and every other target being tested Sys(EISDIR) https://travis-ci.org/nix-rust/nix/builds/541975264?utm_source=github_status&utm_medium=notification. I will resubmit and have to be a less specific about the type of error expected. Not sure if this indicates a bug on those apple targets or not?

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1042](https://github.com/nix-rust/nix/pull/1042))
 - Add `sched_yield`
   ([#1050](https://github.com/nix-rust/nix/pull/1050))
+- Add `unlinkat`
+  ([#TBD](https://github.com/nix-rust/nix/pull/1058))
```
asomers: Don't forget to replace the other TBD.

jlb6740: I see. Expecting either is the better option but it's not clear to me the expect way to account for this when assigning a string to expected for the match. Any pointer?

jlb6740:
I see. Expecting either is the better option but it's not clear to me the expect way to account for this when assigning a string to expected for the match. Any pointer?

> This is not a case of Apple being wrong, this is a case of POSIX being stupid.  EPERM is the error that POSIX says should be returned in this case.  But it looks like both FreeBSD And Linux choose to return EISDIR instead, which makes more sense.  I suggest that your test should accept either error.
> https://pubs.opengroup.org/onlinepubs/9699919799/functions/unlink.html

asomers:
It's easy if you do as I suggested earlier and use `unwrap_err` instead of `should_panic`.

jlb6740:
> This is way too complicated.  The right way to test code that ought to panic is with the #[should_panic()] attribute.  See sys/test_aio.rs for an example.
>You should the expected("whatever") string to should_panic to ensure that it's only catching the desired panic, and no others.

I used should_panic and expected based on your suggests. I am just revisiting this patch today and didn't even notice that latest comment ... I went right to understanding the failure that precedes it. I'll change to using un_wrap error, but I'd like some learnings here. Is this only being suggested now because we realize we want to match against multiple potential error messages or should I have avoided should_panic from the beginning? In general when do you suggest using should_panic? I noticed a few other tests in this folder that use the syntax, but not many have elected to.

asomers:
It's true.  When you first posted this PR, I didn't see the forest for the trees.  The `std::panic::catch_unwind` was such a big red flag that I failed to realize that the test didn't even need to panic at all.  The takeaway is:

1) `should_panic` is usually the correct way to test code that ought to panic.  `std::panic::catch_unwind` should rarely be used unless you're doing something like implementing `should_panic`.

2) The best way to test code that returns an error is to simply look at the error it returns.  Asserting that it returned ok (which is what `unwrap` does) is roundabout and overkill.  In tests/sys/test_aio.rs I used `should_panic` because the `panic`s I was trying to exercise lay in the crate's source, in src/sys/aio.rs.  But that's not the case here.  Your panic lays in the test.  It's the `unwrap` statement that's panicing.

jlb6740:
@asomers Ok, thanks. So to recap from the two bullets .. the first is saying generally prefer using should_panic, while the second bullet is saying but when it it's possible to just compare the error directly then to do that? It still seems a little ambigious considering I thought the point of "expected" was to compare the error message though in this specific case it is not as helpful since we'd like to accept multiple error values. If those apple targets had returned EISDIR the same as FreeBSD and Linux (or they all returned EPERM) I am still unsure what would be the preferred way to check the error? In any case I made an update here: https://gist.github.com/jlb6740/9848944cbab4a1ab5165a3f5ee2b3e1f. If it looks ok, I'll push this and update the CHANGELOG as well.

Thanks.

asomers:
The reason why this test shouldn't use `should_panic` is because the code that you're trying to test doesn't panic.  It simply returns an error.  So the best way to test that is to just check the error code.  Don't make the test more complicated than it needs to be.

jlb6740:
> The reason why this test shouldn't use should_panic is because the code that you're trying to test doesn't panic.  It simply returns an error.  So the best way to test that is to just check the error code.  Don't make the test more complicated than it needs to be.

Ok ... Thanks. I think I see your point. I'll go ahead and resubmit. Any more suggestion just let me know. Thanks for your patients.

jlb6740:
@asomers Github is showing me a "Changes requested" message but I think the previous requests have been addressed. Clicking "resolve conversations" doesn't seem to get rid of the message .. not sure what github wants from me to make the message go away. If there is anything else you see amiss just let me know. My latest update seemed ok, but I did need a rebase in order to pass all tests.

jlb6740:
Hi @asomers, 

Was there anything left here you'd like updated? I think commit 4 should have resolved the remaining requests?

CHANGELOG.md
```diff
@@ -76,6 +76,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1042](https://github.com/nix-rust/nix/pull/1042))
 - Add `sched_yield`
   ([#1050](https://github.com/nix-rust/nix/pull/1050))
+- Add `unlinkat`
```
asomers: Your changelog entry got dislocated during a rebase.

test/test_unistd.rs
```diff
@@ -599,6 +600,63 @@ fn test_symlinkat() {
...
+    // Create dir
+    DirBuilder::new().recursive(true).create(&dirpath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink dir at relative path without proper flag
+    let err_result = unlinkat(Some(dirfd), dirname, UnlinkatFlags::NoRemoveDir).unwrap_err();
+    match err_result {
+        Error::Sys(Errno::EISDIR) => println!("{}", err_result),
```
asomers: passing tests shouldn't print anything.  Just do `assert!(err_result == Error::Sys(Errno::EISDIR) || err_result == Error::Sys(Errno::EPERM))`

CHANGELOG.md
jlb6740: Hi ... what do you mean by dislocated? For the latest patch this is the change I intended for changelog.md:

-  ([#TBD](https://github.com/nix-rust/nix/pull/1058))
+  ([#1058](https://github.com/nix-rust/nix/pull/1058))

Is this what you can see? Is there another push needed for this file?

test/test_unistd.rs
jlb6740: I see .. good catch/comment. I've corrected this in the 5th commit.

jlb6740:
@asomers I've made changes to test/test_unistd.rs .. but maybe didn't address CHANGELOG.md comment? Let me know. Thanks.

asomers:
The changelog entry is in the section for release 0.14. It should be in the unreleased section.

jlb6740:
> The changelog entry is in the section for release 0.14. It should be in the unreleased section.

Ahh .. I see. I've moved it to the "Added" part of the "Unreleased" section. Thanks.

CHANGELOG.md
```diff
@@ -77,6 +79,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add `sched_yield`
   ([#1050](https://github.com/nix-rust/nix/pull/1050))
 
+
```
asomers: Eliminate this newline that you added.

test/test_unistd.rs
```diff
@@ -599,6 +600,59 @@ fn test_symlinkat() {
...
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt unlink file at relative path
+    unlinkat(Some(dirfd), filename, UnlinkatFlags::NoRemoveDir).unwrap();
+    assert!(!filepath.exists());
+ }
+
+
```
asomers: Too many newlines here.

jlb6740: Hi @asomers .. Done. Last commit should address this.

asomers:
There are conflicts now that you must resolve.

jlb6740:
I've rebased and pushed ... Looks like the latest push is passing all tests.

asomers:
You still haven't squashed your commits, though.

jlb6740:
> You still haven't squashed your commits, though.

Ok ... I am familiar with the best way to push and merge patches to gerrit but obviously I am still new to github. I was thinking my commits would be squashed automatically when you merged here. Is the work flow to squash on my local branch and then push here (i.e. git rebase -i HEAD~7), or does github provide some UI for this. I am seeing this: https://help.github.com/en/articles/about-pull-request-merges#squash-and-merge-your-pull-request-commits but I can't find this button or tell if it is for me or the person doing the merging?

asomers:
Many projects squash commits when they merge the PR.  However, Nix uses the bors merge bot, so that isn't possible for us.  Instead, you must squash yourself.

jlb6740:
> Many projects squash commits when they merge the PR.  However, Nix uses the bors merge bot, so that isn't possible for us.  Instead, you must squash yourself.

You respond quick .. thanks. Ok. So do a git rebase -i and then do a force push. I'll give that a try.
Thanks!

jlb6740:
@asmoers Hi .. Not sure how bors works or if anything else is needed on my end for the pull request. Let me know. I'll be starting another api, mimicking what's been done here.

PR #1071, opened 2019-05-28
Implement lchown call

antialize:
I am missing the lchown call, that allows to chown symlinks instead of the files they point to.


asomers:
Is there anything you can do with `lchown` that you can't do with `fchownat`?  If not, then I don't want to merge this PR.  No point in adding redundant functionality.

antialize:
No there is not. I see this is also documented in the fchownat call. I wonder if there is any way to make this show up in the documentation for developers that are more familiar with lchown. The current search https://docs.rs/nix/0.14.0/nix/unistd/fn.chown.html?search=lchown, points to chown. I have updated the pull request to document also here that fchown at should be used instead of lchown.

asomers:
Good question.  Is there a way to make RustDoc document a function that does not exist?

njam:
Maybe add a comment to the documentation of `chown` like: 
> For changing the ownership of a symlink (without following it), please use `fchownat` instead

PR #1072, opened 2019-05-31
Fix libc deprecations

dario23:
The `libc` crate now uses rust types in some places where previously types like `libc::uint8_t` had been defined. This PR adjusts nix's code to reflect these simplifications.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude     = [
 ]
 
 [dependencies]
-libc = "0.2.55"
+libc = "0.2.57"
 bitflags = "1.0"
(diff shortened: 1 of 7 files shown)


dario23:
there seem to be more deprecations than showed up on my machine, travis has some failures.. i'll push an update soon-ish.

dario23:
there seem to be more deprecations than showed up on my machine, travis has some failures.. i'll push an update soon-ish.

asomers:
Any update?  We're still very interested in this PR.

dario23:
@asomers thanks for completing; wasn't available the last two days :-)

PR #1076, opened 2019-06-03
Miscellaneous test fixes

asomers:
Miscellaneous fixes needed to get the tests working more reliably when --test-threads > 1 , or when run inside of Linux containers.


asomers:
The build script for the rand crate has begun to segfault since I upgraded the image to 12.0.  I can't reproduce that failure locally.  I'll probably have to create my own VM on GCP to reproduce it :( .

PR #1083, opened 2019-06-09
Allow signal injection in ptrace::syscall and ptrace::detach

### LINKED ISSUE #1049: Some ptrace functions don't allow signal injection
From [`ptrace(2)`](https://linux.die.net/man/2/ptrace):
> After signal-delivery-stop is observed by the tracer, the tracer should restart the tracee with the call `ptrace(PTRACE_restart, pid, 0, sig)` where `PTRACE_restart` is one of the restarting ptrace requests. If `sig` is 0, then a signal is not delivered. Otherwise, the signal `sig` is delivered. This operation is called signal injection

As far as I understand, the restarting requests correspond to `cont`, `detach`, `step`, and `syscall` in the `sys::ptrace` module. `cont` and `step` allow signal injection through an `Option<Signal>` argument. `detach` and `syscall` don't have such an argument so signal injection is not possible. I think they should all have this optional signal argument.

frangio:
Fixes #1049

Should I add tests for this functionality?

By the way, I noticed that the BSD module doesn't have `ptrace::syscall`. I couldn't find a reason behind this in #949. Should we add it?

### DIFF
--- a/src/sys/ptrace/bsd.rs
+++ b/src/sys/ptrace/bsd.rs
@@ -77,16 +77,23 @@ pub fn traceme() -> Result<()> {
 
 /// Attach to a running process, as with `ptrace(PT_ATTACH, ...)`
 ///
-/// Attaches to the process specified in pid, making it a tracee of the calling process.
+/// Attaches to the process specified by `pid`, making it a tracee of the calling process.
 pub fn attach(pid: Pid) -> Result<()> {
     unsafe { ptrace_other(Request::PT_ATTACH, pid, ptr::null_mut(), 0).map(drop) }
 }
 
 /// Detaches the current running process, as with `ptrace(PT_DETACH, ...)`
 ///
-/// Detaches from the process specified in pid allowing it to run freely
-pub fn detach(pid: Pid) -> Result<()> {
-    unsafe { ptrace_other(Request::PT_DETACH, pid, ptr::null_mut(), 0).map(drop) }
+/// Detaches from the process specified by `pid` allowing it to run freely, optionally delivering a
+/// signal specified by `sig`.
+pub fn detach<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
+    let data = match sig.into() {
+        Some(s) => s as c_int,
+        None => 0,
+    };
+    unsafe {
+        ptrace_other(Request::PT_DETACH, pid, ptr::null_mut(), data).map(drop)
+    }
 }
 
 /// Restart the stopped tracee process, as with `ptrace(PTRACE_CONT, ...)`
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -289,21 +289,26 @@ pub fn traceme() -> Result<()> {
 
 /// Ask for next syscall, as with `ptrace(PTRACE_SYSCALL, ...)`
 ///
-/// Arranges for the tracee to be stopped at the next entry to or exit from a system call.
-pub fn syscall(pid: Pid) -> Result<()> {
+/// Arranges for the tracee to be stopped at the next entry to or exit from a system call,
+/// optionally delivering a signal specified by `sig`.
+pub fn syscall<T: Into<Option<Signal>>>(pid: Pid, sig: T) -> Result<()> {
+    let data = match sig.into() {
+        Some(s) => s as i32 as *mut c_void,
+        None => ptr::null_mut(),
+    };
     unsafe {
         ptrace_other(
             Request::PTRACE_SYSCALL,
             pid,
             ptr::null_mut(),
-            ptr::null_mut(),
+            data,
         ).map(drop) // ignore the useless return value
     }
 }
 
 /// Attach to a running process, as with `ptrace(PTRACE_ATTACH, ...)`
 ///
-/// Attaches to the process specified in pid, making it a tracee of the calling process.
+/// Attaches to the process specified by `pid`, making it a tracee of the calling process.
 pub fn attach(pid: Pid) -> Result<()> {
     unsafe {
         ptrace_other(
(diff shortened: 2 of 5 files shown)


asomers:
Tests: yes please.  Especially for ptrace, which is very difficult to use correctly.  I can't figure out a good reason why BSD lacks `ptrace::syscall` either.  Adding that would be appreciated.  If you do it, please do it in a separate commit.  Also, what about `PTRACE_CONT`?  Can't it inject signals too?

frangio:
Yes, `cont` and `step` can also inject signals, but this was already present in the nix API. Not sure why that was done inconsistently.

frangio:
I've added tests now, but restricted them to Linux/Android only. I'm not familiar with the ptrace API in BSD so I wouldn't know how to test it correctly.

frangio:
I removed my implementation of `ptrace::syscall` for BSD. The build had failed and when I looked into it I found that not all BSDs have this part of the ptrace API, and among those who have it it looks like it works differently. I'll open an issue so that someone more knowledgeable in BSD can work on it.

src/sys/ptrace/bsd.rs
```diff
@@ -84,9 +84,16 @@ pub fn attach(pid: Pid) -> Result<()> {
 
 /// Detaches the current running process, as with `ptrace(PT_DETACH, ...)`
 ///
-/// Detaches from the process specified in pid allowing it to run freely
-pub fn detach(pid: Pid) -> Result<()> {
-    unsafe { ptrace_other(Request::PT_DETACH, pid, ptr::null_mut(), 0).map(drop) }
+/// Detaches from the process specified in pid allowing it to run freely, optionally delivering a
```
asomers: "pid" should be in backticks.  And "in pid" should be "by pid".

src/sys/ptrace/linux.rs
```diff
@@ -318,14 +323,19 @@ pub fn attach(pid: Pid) -> Result<()> {
 
 /// Detaches the current running process, as with `ptrace(PTRACE_DETACH, ...)`
 ///
-/// Detaches from the process specified in pid allowing it to run freely
-pub fn detach(pid: Pid) -> Result<()> {
+/// Detaches from the process specified in pid allowing it to run freely, optionally delivering a
```
asomers: "in pid" should be "by `pid`".

test/sys/test_ptrace.rs
```diff
@@ -105,3 +105,78 @@ fn test_ptrace_cont() {
...
+    use nix::sys::signal::Signal;
+    use nix::sys::wait::{waitpid, WaitStatus};
+    use nix::unistd::fork;
+    use nix::unistd::getpid;
+    use nix::unistd::ForkResult::*;
+
+    let _m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+    // FIXME: qemu-user doesn't implement ptrace on all architectures
+    // and retunrs ENOSYS in this case.
```
asomers: This is a little bit redundant.  On the one hand, you only compile the test on platforms that don't use qemu-user in CI.  OTOH, you try to detect qemu-user here.

test/sys/test_ptrace.rs
```diff
@@ -105,3 +105,78 @@ fn test_ptrace_cont() {
...
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::PtraceSyscall(child)));
+            assert_eq!(get_syscall_id(), ::libc::SYS_kill);
+
+            // receive signal
+            ptrace::syscall(child, None).unwrap();
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::Stopped(child, Signal::SIGTERM)));
+
+            // inject signal
+            ptrace::syscall(child, Signal::SIGTERM).unwrap();
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::Signaled(child, Signal::SIGTERM, false)));
```
asomers: So the child sends `SIGTERM` to itself, _and_ you inject it here?  Why two places?

frangio: That's just how `ptrace` works. It intercepts signal delivery and allows the tracer to decide whether the child will actually receive the signal or not. The mechanism to do that is signal injection. (It's a bad name, and a bad mechanism tbh.) If you don't inject the signal that was intercepted, it won't be delivered at all.

test/sys/test_ptrace.rs
frangio: Good catch, thank you. I had copied this over from another test.

frangio:
This is very weird, the tests are failing in the CI but not in my machine. I'll keep looking into it.

frangio:
CI is green now!

asomers:
Due to an upstream problem with the tempfile crate, you'll have to rebase.

frangio:
@asomers Done!

MikailBag:
I would like to use signal injection with PTRACE_SYSCALL. 
While it is not hard to manually call `libc::ptrace`, I'd like to see this PR merged.

frangio:
Is there anything pending on my end?

asomers:
Sorry that this dropped off of my radar.  I should have some time this week to look at it.

test/sys/test_ptrace.rs
```diff
@@ -105,3 +105,62 @@ fn test_ptrace_cont() {
...
+    let _m = ::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+    match fork().expect("Error: Fork Failed") {
+        Child => {
+            ptrace::traceme().unwrap();
+            // first sigstop until parent is ready to continue
+            let pid = getpid();
+            kill(pid, Signal::SIGSTOP).unwrap();
+            kill(pid, Signal::SIGTERM).unwrap();
+        },
```
asomers: You need to call `_exit(0)` here.  Otherwise, if the process isn't killed, it will do very bad things while cleaning up.

test/sys/test_ptrace.rs
```diff
@@ -105,3 +105,62 @@ fn test_ptrace_cont() {
...
+
+        Parent { child } => {
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::Stopped(child, Signal::SIGSTOP)));
+
+            // set this option to recognize syscall-stops
+            ptrace::setoptions(child, ptrace::Options::PTRACE_O_TRACESYSGOOD).unwrap();
+
+            #[cfg(target_arch = "x86_64")]
```
asomers: How about using `target_pointer_width` instead of `target_arch`?

frangio: Good idea.

frangio:
Done! Please let me know if the changelog entry looks ok.

frangio:
Should be done!

asomers:
I'm sorry, but you'll have to rebase again.  And be sure to rebase; don't merge master into your branch.

frangio:
@asomers Done.

PR #1084, opened 2019-06-10
sys/stat: implement mkdirat

kevinwern:
See: https://github.com/CraneStation/wasi-common/issues/16
https://linux.die.net/man/2/mkdirat

My first contribution to this repo. Tests were probably overkill...

Also, out of curiosity, is there any reason why `mkdir` is located in `unistd`? The documentation I read mentioned the function definition also being located in `sys/stat.h`.

### DIFF
--- a/test/test_stat.rs
+++ b/test/test_stat.rs
@@ -1,13 +1,15 @@
 use std::fs::{self, File};
-use std::os::unix::fs::symlink;
+use std::os::unix::fs::{symlink, PermissionsExt};
 use std::os::unix::prelude::AsRawFd;
 use std::time::{Duration, UNIX_EPOCH};
+use std::path::Path;
 
 #[cfg(not(any(target_os = "netbsd")))]
-use libc::{S_IFMT, S_IFLNK};
+use libc::{S_IFMT, S_IFLNK, mode_t};
 
-use nix::fcntl;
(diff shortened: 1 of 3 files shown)


test/test_stat.rs
```diff
@@ -260,3 +261,36 @@ fn test_utimensat() {
...
+#[test]
+#[should_panic="EACCES"]
+fn test_mkdirat_fail() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let filename = "example_subdir";
+    let mut perms = fs::metadata(tempdir.path()).unwrap().permissions();
+    perms.set_readonly(true);
+    fs::set_permissions(tempdir.path(), perms).unwrap();
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+    let result = mkdirat(dirfd, filename, Mode::S_IRWXU).unwrap();
```
asomers: Instead of using `should_panic`, you can just replace `unwrap` with `unwrap_err`.

kevinwern:
Switched the error condition to ENOTDIR because I suspected the FreeBSD build was failing because of something related to `set_permissions`... not sure if that's the case, but seems to be passing now.

kevinwern:
Rebased again...just leaving a comment because I forget if GH sends notifications for force pushes.

PR #1091, opened 2019-06-25
socket: add support for AF_VSOCK

stefano-garzarella:
This patch adds the support of AF_VSOCK in the socket module.
VSOCK is present since Linux 3.9.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -24,6 +24,8 @@ use ::sys::socket::addr::sys_control::SysControlAddr;
           target_os = "netbsd",
           target_os = "openbsd"))]
 pub use self::datalink::LinkAddr;
+#[cfg(target_os = "linux")]
+pub use self::vsock::VsockAddr;
 
 /// These constants specify the protocol family to be used
 /// in [`socket`](fn.socket.html) and [`socketpair`](fn.socketpair.html)
@@ -241,6 +243,8 @@ impl AddressFamily {
                       target_os = "netbsd",
                       target_os = "openbsd"))]
             libc::AF_LINK => Some(AddressFamily::Link),
+            #[cfg(target_os = "linux")]
+            libc::AF_VSOCK => Some(AddressFamily::Vsock),
             _ => None
         }
     }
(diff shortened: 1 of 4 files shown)


stefano-garzarella:
I added an entry in the CHANGELOG, more description in the commit message and forced push.

stefano-garzarella:
> The patch looks good so far. But could you please add some tests for the functionality?

Sure, I'll add some tests for VSOCK.

stefano-garzarella:
@asomers I add a test. Unfortunately, the current VSOCK implementation in Linux doesn't support loopback devices, so, for now, I add a simple test that tries to bind forbidden address and tries to connect, receiving an error.

For the 'AddressFamily' I think it is enough, what do you think?

asomers:
All the test needs to do is ensure that Nix is binding correctly; we don't need a full regression test for vsock.  So if `bind` returns a distinctive error code then that's enough.

stefano-garzarella:
> All the test needs to do is ensure that Nix is binding correctly; we don't need a full regression test for vsock. So if `bind` returns a distinctive error code then that's enough.

@asomers Okay, so what I tried in the second patch that I pushed on this PR is exactly what you said, because a bind on reserved address should return EADDRNOTAVAIL.

src/sys/socket/addr.rs
```diff
@@ -1124,6 +1144,66 @@ mod datalink {
...
+pub mod vsock {
+    use ::sys::socket::addr::AddressFamily;
+    use libc::{sa_family_t, sockaddr_vm};
+    use std::{fmt, mem};
+    use std::hash::{Hash, Hasher};
+
+    #[derive(Copy, Clone)]
+    pub struct VsockAddr(pub sockaddr_vm);
+
+    // , PartialEq, Eq, Debug, Hash
```
asomers: What exactly does this comment mean, and why the leading comma?

stefano-garzarella: Oh, sorry about that! I think was a line copied by struct AlgAddr that I used as a template for VsockAddr. I'll remove this line.

stefano-garzarella:
@asomers I removed the useless comment line that you pointed out and added few comments on VsockAddr to explain cid and port fields.

asomers:
Please rebase to fix the merge conflict.

stefano-garzarella:
> Please rebase to fix the merge conflict.

@asomers done!

Thanks,
Stefano

Issue #1096, opened 2019-07-07
Replace all calls to mem::uninitialized with MaybeUninit


asomers:
Rust will deprecate `mem::uninitialized` beginning with Rust 1.38.0.  The recommended replacement is `MaybeUninit`.  Unfortunately, `MaybeUninit` was only introduced in Rust 1.36.0.  So we'll have to raise the MSRV.  We should therefore wait to fix this issue until shortly before 1.38.0 is released, which will be sometime around September-29.
https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html

PR #1100, opened 2019-07-24
implement clock_gettime, clock_getres, clock_settime

kevinwern:
See: https://pubs.opengroup.org/onlinepubs/007908799/xsh/clock_settime.html
Again related to: https://github.com/CraneStation/wasi-common/issues/16

Conditional inclusion of clock IDs based on looking through libc.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -1,0 +1,184 @@
+use sys::time::TimeSpec;
+use std::mem::MaybeUninit;
+use {Result, Errno};
+use libc;
+
+libc_enum! {
+    #[cfg_attr(any(
+                target_env = "uclibc",
+                target_os = "fuchsia",
+                target_os = "redox",
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd",
+                target_os = "haiku",
+                all(not(target_env = "newlib"),
+                    any(target_os = "linux",
+                        target_os = "android",
+                        target_os = "emscripten",
+                        target_os = "solaris",
+                        target_os = "illumos")),
+               ), repr(i32))]
+    #[cfg_attr(any(
+            all(target_env = "newlib", target_arch = "arm"),
+            target_os = "macos",
+            target_os = "ios",
+            ), repr(u32))]
+    #[cfg_attr(any(
+                all(target_env = "newlib", target_arch = "aarch64"),
+                all(not(any(target_env = "newlib", target_env = "uclibc")), target_os = "hermit"),
+                target_os = "dragonfly",
+              ), repr(u64))]
+    pub enum ClockId {
+        #[cfg(any(target_os = "fuchsia",
+                  all(not(any(target_env = "uclibc",
+                              target_env = "newlib")),
+                      any(target_os = "linux",
+                          target_os = "android",
+                          target_os = "emscripten"),
+                      )
+                  ))]
+        CLOCK_BOOTTIME,
+        #[cfg(any(target_os = "fuchsia",
+                  all(not(any(target_env = "uclibc",
+                              target_env = "newlib")),
+                      any(target_os = "linux",
+                          target_os = "android",
+                          target_os = "emscripten"))))]
+        CLOCK_BOOTTIME_ALARM,
+        CLOCK_MONOTONIC,
+        #[cfg(any(target_os = "fuchsia",
+                  all(not(any(target_env = "uclibc",
+                              target_env = "newlib")),
+                      any(target_os = "linux",
+                          target_os = "android",
+                          target_os = "emscripten"))))]
+        CLOCK_MONOTONIC_COARSE,
+        #[cfg(any(target_os = "freebsd",
+                  target_os = "dragonfly"))]
+        CLOCK_MONOTONIC_FAST,
+        #[cfg(any(target_os = "freebsd",
+                  target_os = "dragonfly"))]
+        CLOCK_MONOTONIC_PRECISE,
+        #[cfg(any(target_os = "fuchsia",
+                  all(not(any(target_env = "uclibc",
+                              target_env = "newlib")),
+                      any(target_os = "linux",
+                          target_os = "android",
+                          target_os = "emscripten"))))]
+        CLOCK_MONOTONIC_RAW,
+        #[cfg(any(target_os = "fuchsia",
+                  target_env = "uclibc",
+                  target_os = "macos",
+                  target_os = "ios",
+                  target_os = "freebsd",
+                  target_os = "dragonfly",
+                  all(not(target_env = "newlib"),
+                      any(target_os = "linux",
+                          target_os = "android",
+                          target_os = "emscripten"))))]
+        CLOCK_PROCESS_CPUTIME_ID,
+        #[cfg(any(target_os = "freebsd",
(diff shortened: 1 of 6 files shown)


src/time.rs
```diff
@@ -0,0 +1,118 @@
...
+                  target_os = "dragonfly"))]
+        CLOCK_UPTIME_PRECISE,
+        #[cfg(any(target_os = "freebsd",
+                  target_os = "dragonfly"))]
+        CLOCK_VIRTUAL,
+    }
+}
+
+pub fn clock_getres(clk_id: ClockId) -> Result<TimeSpec> {
+    let mut c_time = libc::timespec {tv_sec: 0, tv_nsec: 0};
```
asomers: Please use `MaybeUninit` instead of zero-initializing.  That will cause the test to fail for now, but they'll pass as soon as #1096 is fixed.

test/test_time.rs
```diff
@@ -0,0 +1,32 @@
...
+    let res1 = clock_gettime(ClockId::CLOCK_REALTIME).unwrap();
+    let res2 = clock_gettime(ClockId::CLOCK_REALTIME).unwrap();
+    assert!(res1 < res2);
+}
+
+#[test]
+pub fn test_clock_settime() {
+    require_capability!(CAP_SYS_TIME);
+    let ts = TimeSpec::from(timespec{tv_sec: 10000000, tv_nsec: 100});
+    let res = clock_settime(ClockId::CLOCK_REALTIME, ts).unwrap();
```
asomers: Lol, this will set the system time back to April 26, 1970!  That's not a safe test.  There's probably little you can do to test `clock_settime`.  Perhaps it would be ok to set it to the current time.

test/test_time.rs
```diff
@@ -0,0 +1,32 @@
...
+#[test]
+pub fn test_clock_gettime() {
+    let res1 = clock_gettime(ClockId::CLOCK_REALTIME).unwrap();
+    let res2 = clock_gettime(ClockId::CLOCK_REALTIME).unwrap();
+    assert!(res1 < res2);
+}
+
+#[test]
+pub fn test_clock_settime() {
+    require_capability!(CAP_SYS_TIME);
```
asomers: `require_capability` is a Linux-only macro.

kevinwern:
Was making some of the changes but getting on a flight now...will follow up with the rest

src/time.rs
```diff
@@ -3,7 +3,8 @@ use {Result, Errno};
 use libc;
 
 libc_enum! {
-    #[repr(i32)]
+    #[cfg_attr(target_os = "darwin", repr(i32))]
+    #[cfg_attr(not(target_os = "darwin"), repr(u32))]
```
asomers: This isn't correct.  Check the libc crate to see what the actual size is on various platforms:
```
> rg type.clockid_t
linux_like/mod.rs
4:pub type clockid_t = ::c_int;

newlib/mod.rs
4:pub type clockid_t = ::c_ulong;

uclibc/mod.rs
6:pub type clockid_t = ::c_int;

solarish/mod.rs
5:pub type clockid_t = ::c_int;

haiku/mod.rs
9:pub type clockid_t = i32;

redox/mod.rs
9:pub type clockid_t = ::c_int;

hermit/mod.rs
39:pub type clockid_t = c_ulong;

bsd/netbsdlike/mod.rs
10:pub type clockid_t = ::c_int;

bsd/apple/mod.rs
16:pub type clockid_t = ::c_uint;

bsd/freebsdlike/dragonfly/mod.rs
8:pub type clockid_t = ::c_ulong;

bsd/freebsdlike/freebsd/mod.rs
6:pub type clockid_t = ::c_int;
```

src/time.rs
asomers: `MaybeUninit` is the correct thing to do here, not zero-initialization.

src/time.rs
```diff
@@ -0,0 +1,149 @@
...
+                target_os = "openbsd",
+                target_os = "haiku",
+                target_os = "linux",
+                target_os = "android",
+                target_os = "emscripten",
+                target_os = "solaris",
+                target_os = "illumos",
+                all(target_env = "newlib", target_arch = "arm"),
+               ), repr(i32))]
+    #[cfg_attr(target_os = "macos", repr(u32))]
```
asomers: Don't forget ios

src/time.rs
```diff
@@ -0,0 +1,149 @@
...
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd",
+                target_os = "haiku",
+                target_os = "linux",
+                target_os = "android",
+                target_os = "emscripten",
+                target_os = "solaris",
+                target_os = "illumos",
+                all(target_env = "newlib", target_arch = "arm"),
```
asomers: Why the special case for `target_arch` on newlib?  I don't see that in libc.

src/time.rs
```diff
@@ -0,0 +1,149 @@
...
+libc_enum! {
+    #[cfg_attr(any(
+                target_os = "fuchsia",
+                target_os = "redox",
+                target_env = "uclibc",
+                target_os = "freebsd",
+                target_os = "netbsd",
+                target_os = "openbsd",
+                target_os = "haiku",
+                target_os = "linux",
```
asomers: This should be `all(target_os = "linux", not(target_env = "newlib"))` and likewise for android and emscripten.

kevinwern: Ah, that's very tricky. Added that in the case of everything that appears after newlib/uclibc. In some cases it had to be `all(target_os = "<x>", not(target_env = "newlib", target_env = "uclibc"))`.

src/time.rs
kevinwern: For `newlib`, the supported `target_archs` are only `arm `and `aarch64`. While the type `clockid_t` is defined here:
https://github.com/rust-lang/libc/blob/49c1d138e3b4719e21ee7cbce7dd903543cda80d/src/unix/newlib/mod.rs#L4

`c_ulong` is defined for each arch in their respective modules: 
https://github.com/rust-lang/libc/blob/49c1d138e3b4719e21ee7cbce7dd903543cda80d/src/unix/newlib/arm/mod.rs#L5
https://github.com/rust-lang/libc/blob/49c1d138e3b4719e21ee7cbce7dd903543cda80d/src/unix/newlib/aarch64/mod.rs#L5

kevinwern: Just realized that I had put the "arm" case under i32 rather than u32. Hopefully makes more sense now.

asomers:
You should be able to fix the test failures by rebasing now.

asomers:
Ok, I think this is good, but you need to squash before we can merge.

kevinwern:
Squashed a while ago and just rebased again for the changelog. Hopefully good to merge now.

asomers:
I've seen this before.  I think it's a bug in glibc.  It's certainly nothing to do with your PR.

bors retry

asomers:
Please rebase to fix the musl failure.

xonatius:
I'm interested in some of these functions too for #1275. @kevinwern do you mind if I pick up this PR to implement them?

cemeyer:
Can we close this one after #1281?

asomers:
Yep

PR #1101, opened 2019-07-25
Implment linkat

jlb6740:
This adds the linkat function which is part of POSIX:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/linkat.html
and widely implmented on Unix-Family platforms.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1169,6 +1169,58 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
     }
 }
 
+/// Flags for `linkat` function.
+#[derive(Clone, Copy, Debug)]
+pub enum LinkatFlags {
+    SymlinkFollow,
+    NoSymlinkFollow,
+}
+
+/// Link one file to another file
+///
+/// Creates a new link (directory entry) at `newpath` for the existing file at `oldpath`. In the
+/// case of a relative `oldpath`, the path is interpreted relative to the directory associated
+/// with file descriptor `olddirfd` instead of the current working directory and similiarly for
+/// `newpath` and file descriptor `newdirfd`. In case `flag` is LinkatFlags::SymlinkFollow and
+/// `oldpath` names a symoblic link, a new link for the target of the symbolic link is created.
+/// If either `olddirfd` or `newdirfd` is `None`, `AT_FDCWD` is used respectively where `oldpath`
+/// and/or `newpath` is then interpreted relative to the current working directory of the calling
+/// process. If either `oldpath` or `newpath` is absolute, then `dirfd` is ignored.
+///
+/// # References
+/// See also [linkat(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/linkat.html)
+pub fn linkat<P: ?Sized + NixPath>(
+    olddirfd: Option<RawFd>,
+    oldpath: &P,
+    newdirfd: Option<RawFd>,
+    newpath: &P,
+    flag: LinkatFlags,
+) -> Result<()> {
+
+    let atflag =
+        match flag {
+            LinkatFlags::SymlinkFollow => AtFlags::AT_SYMLINK_FOLLOW,
+            LinkatFlags::NoSymlinkFollow => AtFlags::empty(),
+        };
+
+    let res =
+        oldpath.with_nix_path(|oldcstr| {
+            newpath.with_nix_path(|newcstr| {
+            unsafe {
+                libc::linkat(
+                    at_rawfd(olddirfd),
+                    oldcstr.as_ptr(),
+                    at_rawfd(newdirfd),
+                    newcstr.as_ptr(),
+                    atflag.bits() as libc::c_int
+                    )
+                }
+            })
+        })??;
+    Errno::result(res).map(drop)
+}
+
+
 /// Remove a directory entry
 ///
 /// See also [unlink(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/unlink.html)
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -660,6 +660,137 @@ fn test_symlinkat() {
     );
 }
 
+#[test]
+fn test_linkat_file() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let oldfilename = "foo.txt";
+    let oldfilepath = tempdir.path().join(oldfilename);
+
+    let newfilename = "bar.txt";
+    let newfilepath = tempdir.path().join(newfilename);
+
+    // Create file
+    File::create(&oldfilepath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt hard link file at relative path
+    linkat(Some(dirfd), oldfilename, Some(dirfd), newfilename, LinkatFlags::SymlinkFollow).unwrap();
+    assert!(newfilepath.exists());
+}
+
+#[test]
+fn test_linkat_olddirfd_none() {
+    let tempdir_oldfile = tempfile::tempdir().unwrap();
+    let oldfilename = "foo.txt";
+    let oldfilepath = tempdir_oldfile.path().join(oldfilename);
+
+    let tempdir_newfile = tempfile::tempdir().unwrap();
+    let newfilename = "bar.txt";
+    let newfilepath = tempdir_newfile.path().join(newfilename);
+
+    // Create file
+    File::create(&oldfilepath).unwrap();
+
+    // Get file descriptor for base directory of new file
+    let dirfd = fcntl::open(tempdir_newfile.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt hard link file using curent working directory as relative path for old file path
+    chdir(tempdir_oldfile.path()).unwrap();
+    linkat(None, oldfilename, Some(dirfd), newfilename, LinkatFlags::SymlinkFollow).unwrap();
+    assert!(newfilepath.exists());
+}
+
+#[test]
+fn test_linkat_newdirfd_none() {
+    let tempdir_oldfile = tempfile::tempdir().unwrap();
+    let oldfilename = "foo.txt";
+    let oldfilepath = tempdir_oldfile.path().join(oldfilename);
+
+    let tempdir_newfile = tempfile::tempdir().unwrap();
+    let newfilename = "bar.txt";
+    let newfilepath = tempdir_newfile.path().join(newfilename);
+
+    // Create file
+    File::create(&oldfilepath).unwrap();
+
+    // Get file descriptor for base directory of old file
+    let dirfd = fcntl::open(tempdir_oldfile.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt hard link file using current working directory as relative path for new file path
+    chdir(tempdir_newfile.path()).unwrap();
+    linkat(Some(dirfd), oldfilename, None, newfilename, LinkatFlags::SymlinkFollow).unwrap();
+    assert!(newfilepath.exists());
+}
+
+#[test]
+#[cfg(not(any(target_os = "ios", target_os = "macos")))]
+fn test_linkat_no_follow_symlink() {
+    let tempdir = tempfile::tempdir().unwrap();
+    let oldfilename = "foo.txt";
+    let oldfilepath = tempdir.path().join(oldfilename);
+
+    let symoldfilename = "symfoo.txt";
+    let symoldfilepath = tempdir.path().join(symoldfilename);
+
+    let newfilename = "nofollowsymbar.txt";
+    let newfilepath = tempdir.path().join(newfilename);
+
+    // Create file
+    File::create(&oldfilepath).unwrap();
+
+    // Create symlink to file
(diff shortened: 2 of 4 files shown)


test/test_unistd.rs
```diff
@@ -601,6 +601,92 @@ fn test_symlinkat() {
...
+    symlinkat(&oldfilepath, None, &symoldfilepath).unwrap();
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt link target of symlink of file at relative path
+    linkat(Some(dirfd), symoldfilename, Some(dirfd), newfilename, LinkatFlags::SymlinkFollow).unwrap();
+
+    let err_result =  readlink(&newfilepath, &mut buf).unwrap_err();
+    assert!(err_result == Error::Sys(Errno::EINVAL));
```
jlb6740: Could not find a convenient way to tell I've created a hard link and not another soft link so I simply try to follow the hardlink and when that fails check the error.

asomers: It's easy to tell.  Just `stat` both old file and new file.

jlb6740: Thanks ... yesterday I googled "rust stat file" and using std::fs::metadata came up. When I compared the metadata for the original file, the symlink, and the hardlink the metadata all looked the same. I then found something for libc::stat but it was describing a struct. I didn't notice there was also a function call of the same name available. Following https://github.com/rust-lang/libc/issues/403 I've been able to see the filetype and number of hardlinks for a file through stat and will post this update.

jlb6740:
@asomer Another example were mac is doing something different than all other targets. You have any thoughts on what flag it is expecting?

image

I am testing not following the symlink. http://pubs.opengroup.org/onlinepubs/9699919799/functions/linkat.html

Says to use 

AT_SYMLINK_FOLLOW
If path1 names a symbolic link, a new link for the target of the symbolic link is created.

I found this first thread from a few years back that says it is not supported:
https://discussions.apple.com/thread/5759427

And I couldn't find it here:
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/

But the other tests for linkat pass on the target and I see this reference here: 
https://www.unix.com/man-page/mojave/2/linkat/

> If linkat() is passed the special value AT_FDCWD in the fd1 or fd2 parameter, the current working directory is used for the respective name
>      argument.	If both fd1 and fd2 have value AT_FDCWD, the behavior is identical to a call to link().  Unless flag contains the
>      AT_SYMLINK_FOLLOW flag, if name1 names a symbolic link, a new link is created for the symbolic link name1 and not its target. On OS X, not
>      assigning AT_SYMLINK_FOLLOW to flag may result in some filesystems returning an error.

where it is said not assigning anything results in an error. Any comments on what I should be adding as a FLAG? The last sentence makes it sound like this i inconsistently supported on mac and that some systems will only accept this call if AT_SYMLINK_FOLLOW is passed to it?

jlb6740:
@asomer In this scenario can we just a check that allows ENOTSOP error or should I just get rid of this test all together and comment that NOSYMLINKFOLLOW is not supported for all targets?

asomers:
Since the OSX man page documents that `AT_SYMLINK_FOLLOW` is required, you should just disable that test on OSX.

jlb6740:
> Since the OSX man page documents that `AT_SYMLINK_FOLLOW` is required, you should just disable that test on OSX.

Ok, make sense. What is the process for doing that .. I guess manually triggering this patch to be retested while omitting specific targets?

jlb6740:
> > Since the OSX man page documents that `AT_SYMLINK_FOLLOW` is required, you should just disable that test on OSX.
> 
> Ok, make sense. What is the process for doing that .. I guess manually triggering this patch to be retested while omitting specific targets?

Ahh  .. I see a travis.yml. I'll try to modify that!

jlb6740:
Actually .. will 

> > > Since the OSX man page documents that `AT_SYMLINK_FOLLOW` is required, you should just disable that test on OSX.
> > 
> > 
> > Ok, make sense. What is the process for doing that .. I guess manually triggering this patch to be retested while omitting specific targets?
> 
> Ahh .. I see a travis.yml. I'll try to modify that!

I see modifying the travis.xml would disable all tests for the target. I see I should just add an attribute based guard as is done on another test.

> #[cfg(not(any(target_os = "ios", target_os = "macos")))]

jlb6740:
@asomers Ok, cool. Tests passed and/or were excluded as expected. It think this is ready but needs review.

src/unistd.rs
```diff
@@ -1132,6 +1132,57 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
     }
 }
 
+
```
asomers: Too many blank lines here.

src/unistd.rs
```diff
@@ -1132,6 +1132,57 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
...
+    NoSymlinkFollow,
+}
+
+/// Link one file to another file
+///
+/// Creates a new link (directory entry) at `newpath` for the existing file at `oldpath`. In the 
+/// case of a relative `oldpath`, the path is interpreted relative to the directory associated 
+/// with file descriptor `olddirfd` instead of the current working directory and similiarly for 
+/// `newpath` and file descriptor `newdirfd`. In case `flag` is LinkatFlags::SymlinkFollow and 
+/// `oldpath` names a symoblic link, a new link for the target of the symbolic link is created.
```
asomers: The docs should mention what it means for `olddirfd` or `newdirfd` to be `None`.

src/unistd.rs
jlb6740: Hi ... didn't notice these comments. Will remove.

src/unistd.rs
jlb6740: Hi ... didn't notice these comments. Will update the doc.

jlb6740: Ok, Updated ... commented on the error conditions that are thrown when olddirfd or newdirfd is  set to None.

jlb6740:
Thanks ... Let me know if there is anything else.

src/unistd.rs
asomers: You still haven't removed them.  Please don't mark conversations as resolved unless you've actually resolved them.

src/unistd.rs
```diff
@@ -1132,6 +1132,58 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
...
+/// The function shall fail if either olddirfd or newdirfd is None, returning ENOENT and EEXIST
```
asomers: WTF?  No it shall not fail.  That's not what will happen.

jlb6740: > WTF? No it shall not fail. That's not what will happen.

What should happen? The spec I am referring to is here: http://pubs.opengroup.org/onlinepubs/9699919799/functions/linkat.html 

Failing is language the spec uses in the ERRORS section to describe errors returned. I.E.

> ERRORS
> These functions shall fail if:
> [EEXIST]
> [ENOENT]

When I tested by inserting "None" at the calls ENOENT and EEXIST were returned. 

Please comment on what you think should happen. If failing with ENOENT and EEXIST is not correct behavior then an update to the patch is needed.

Johnnie

jlb6740:
>You still haven't removed them. Please don't mark conversations as resolved unless you've actually resolved them.

The change most have been unknowingly reverted before I pushed. I believe the newline is gone now but I no longer see the conversation to close. Does that happen in the github UI when an issue such as a newline is resolved as expected?

src/unistd.rs
asomers: What _should_ happen when `olddirfd` is `None` is that you call `libc::linkat` with the `AT_FDCWD` argument, which the operating system interprets to mean that `oldpath` is relative to the current directory.  Similarly with `newdirfd`.

jlb6740: @asomers Ok .. I see, thanks. I saw this mentioned and read the spec to mean this special value could be passed in directly by the user. However in case of None, instead of an returning an error, we could define behavior such that this value is used in this case as well. I'll make this update and use newdirfd as a reference. Thanks.

asomers: No, Johnnie.  The code that you wrote *already uses `AT_FDCWD`*.  How did you not notice that?

jlb6740: @asomers Thanks for the suggestion. The comment has been updated. Perhaps the patch is good now, but let me know if you'd like more changes. Thanks.

jlb6740: @asomers Thanks for the suggestion. The comment has been updated. Perhaps the patch is good now, but let me know if you'd like more changes. Thanks.

src/unistd.rs
```diff
@@ -1149,6 +1149,58 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
...
+}
+
+/// Link one file to another file
+///
+/// Creates a new link (directory entry) at `newpath` for the existing file at `oldpath`. In the
+/// case of a relative `oldpath`, the path is interpreted relative to the directory associated
+/// with file descriptor `olddirfd` instead of the current working directory and similiarly for
+/// `newpath` and file descriptor `newdirfd`. In case `flag` is LinkatFlags::SymlinkFollow and
+/// `oldpath` names a symoblic link, a new link for the target of the symbolic link is created.
+/// If either olddirfd or newdirfd is None, AT_FDCWD is used respectively where pathname is then
```
asomers: Need backticks on this line.  Also, "pathname" should be "oldpath" and/or "newpath".

test/test_unistd.rs
```diff
@@ -609,6 +609,94 @@ fn test_symlinkat() {
...
+
+    // Get file descriptor for base directory
+    let dirfd = fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty()).unwrap();
+
+    // Attempt hard link file at relative path
+    linkat(Some(dirfd), oldfilename, Some(dirfd), newfilename, LinkatFlags::SymlinkFollow).unwrap();
+    assert!(newfilepath.exists());
+}
+
+#[cfg(target_os = "freebsd")]
```
asomers: Why is this test restricted to FreeBSD?

jlb6740:
@asomers Thanks for the suggestions. I think they've been handled. In particular I added tests for the cases of setting olddirfd and newdirfd to None. I think the test cover the input as we'd like, but if you have something else in mind I can definitely make a change no problem. I also updated the CHANGELOG file. Let me know anything else you'd like to see. Thanks.

src/unistd.rs
```diff
@@ -1169,6 +1169,58 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
...
+}
+
+/// Link one file to another file
+///
+/// Creates a new link (directory entry) at `newpath` for the existing file at `oldpath`. In the
+/// case of a relative `oldpath`, the path is interpreted relative to the directory associated
+/// with file descriptor `olddirfd` instead of the current working directory and similiarly for
+/// `newpath` and file descriptor `newdirfd`. In case `flag` is LinkatFlags::SymlinkFollow and
+/// `oldpath` names a symoblic link, a new link for the target of the symbolic link is created.
+/// If either `olddirfd` or `newdirfd` is None, AT_FDCWD is used respectively where `oldpath`
```
asomers: Missing backticks for "None" and "AT_FDCWD"

src/unistd.rs
```diff
@@ -1169,6 +1169,58 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
...
+/// and/or `newpath` is then interpreted relative to the current working directory of the calling
+/// process. If either `oldpath` or `newpath` is absolute, then dirfd is ignored.
```
asomers: Missing backticks for `dirfd`

jlb6740: Ok, was only adding for function parameter names. Updated now. Thanks.

asomers:
Where did all of the whitespace changes come from?  They seem unrelated.  If you're going to modify whitespace that's unrelated to your PR, it needs to be in a separate commit (not necessarily a separate PR).  Also, though it's too late now, in the future please don't force-push until after review is complete and I ask you to squash.

jlb6740:
> Also, though it's too late now, in the future please don't force-push until after review is complete and I ask you to squash.

Ok, will do. 

At some point I've changed IDE settings to remove trailing white spaces to adhere to standards for a different project. I removed trailing whitespaces that existed for an unrelated function as you pointed out. In the latest commit I added the whitespace back. I am thinking that is the change you wanted but if it is not then please let me know. Thanks.

asomers:
Ok, it looks good.  Now you can squash and force-push.

jlb6740:
I assume this will get merged correctly? When doing a rebase I choose a large than needed  look back to make sure I could see all the patches to squash "git rebase -i HEAD~3" and squashed the appropriate patches. But when I forced pushed looks like it captures previous commits not involved in the squash. The files changed tab looks as intended, but if further update is needed let me know.

jlb6740:
> Ok, it looks good. Now you can squash and force-push.

Ok, thanks a lot. Hopefully this is good to go.

src/unistd.rs
jlb6740: Hi @asmoers .. I am not sure the etiquette concerning who marks "Resolve conversation" for change requests such as this. Do you have more comment for this request?

test/test_unistd.rs
jlb6740: Hi @asmoers .. I am not sure the etiquette concerning who marks "Resolve conversation" for change requests such as this. Do you have more comment for this request?

PR #1103, opened 2019-08-09
Add support for UTIME_NOW and UTIME_OMIT to TimeSpec

lpetre:
From the [documentation](https://linux.die.net/man/3/futimens):

> If the tv_nsec field of one of the timespec structures has the special value UTIME_NOW, then the corresponding file timestamp is set to the current time. If the tv_nsec field of one of the timespec structures has the special value UTIME_OMIT, then the corresponding file timestamp is left unchanged. In both of these cases, the value of the corresponding tv_sec field is ignored. 

Given the current API, there is no way to set tv_nsec to UTIME_NOW, since TimeSpec::nanoseconds mangles the value.

### DIFF
--- a/src/sys/time.rs
+++ b/src/sys/time.rs
@@ -143,6 +143,25 @@ impl TimeValLike for TimeSpec {
     }
 }
 
+#[cfg(any(
+    target_os = "linux",
+    target_os = "android",
+    target_os = "emscripten",
+    target_os = "fuchsia",
+    target_env = "uclibc"
+))]
+impl TimeSpec {
+    /// Makes a new `TimeSpec` with the special UTIME_NOW constant
+    pub fn utime_now() -> TimeSpec {
+        TimeSpec(timespec { tv_sec: 0, tv_nsec: libc::UTIME_NOW })
+    }
+
+    /// Makes a new `TimeSpec` with the special UTIME_OMIT constant
+    pub fn utime_omit() -> TimeSpec {
+        TimeSpec(timespec { tv_sec: 0, tv_nsec: libc::UTIME_OMIT })
+    }
+}
+
 impl TimeSpec {
     fn nanos_mod_sec(&self) -> c_long {
         if self.tv_sec() < 0 && self.tv_nsec() > 0 {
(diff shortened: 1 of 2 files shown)


lpetre:
I don't think this works on OSX

src/sys/time.rs
```diff
@@ -144,6 +144,20 @@ impl TimeValLike for TimeSpec {
 }
 
 impl TimeSpec {
+    /// Makes a new `TimeSpec` with the special UTIME_NOW constant
+    #[inline]
```
asomers: No need for `#[inline]`.  The compiler can figure that out for itself.

src/sys/time.rs
```diff
@@ -144,6 +144,20 @@ impl TimeValLike for TimeSpec {
...
+    #[cfg(target_os = "linux")]
```
asomers: Plenty of OSes besides Linux have `UTIME_NOW` and `UTIME_OMIT`.  Can you please enable all of them here?  You may need to make a PR to libc first.

lpetre: Yes, I'm happy to do that. I just don't know how to get that exhaustive list of OSes. Any suggestions?

asomers: There's no easy way.  All of the open source OSes have online manpages; that's what I use.  The hard one is OSX.  I haven't found a good, authoritative source for OSX man pages.

asomers: I think you're still leaving out some OS support.  FreeBSD, for one, defines these symbols.  I haven't checked others.  Can you please check their online man pages, and open a PR to libc to add support where it's missing?

lpetre: I've got a PR open here: https://github.com/rust-lang/libc/pull/1474

src/sys/time.rs
```diff
@@ -143,6 +143,35 @@ impl TimeValLike for TimeSpec {
     }
 }
 
+#[cfg(any(
+    target_env = "uclibc",
+    target_env = "wasi",
+    target_os = "android",
+    target_os = "emscripten",
```
asomers: You can add Dragonfly.  Besides that, what targets are missing from this list?  Does it not cover everything?

lpetre: Opened another PR: https://github.com/rust-lang/libc/pull/1487

SteveLauC:
Hi @lpetre, I just noticed that I created a duplicate PR of this one, since you come first, I would like to merge this and close mine, would you like to rebase your branch and finish this?

SteveLauC:
Alright, I just realized that you have deleted your Nix fork, then I will close this PR and do it in 1897, thanks for your interest in contributing to Nix!

PR #1105, opened 2019-08-11
posix_fallocate support

dingxiangfei2009:
This PR add [`posix_fallocate`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_fallocate.html), which is available on
- Linux
- FreeBSD
- Android
- Emscripten
- Fuchsia
- WASI

Here is a question: for some reason, `posix_fallocate` returns `EBADF` instead of `EPIPE` if a FIFO file descriptor is passed in on Linux 4.19.64. In the test `EBADF` is used for now, but I would like to know if such behaviour is expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased] - ReleaseDate
 ### Added
 
+- Added `posix_fallocate`.
+  ([#1105](https://github.com/nix-rust/nix/pull/1105))
+
 - Implemented `Default` for `FdSet`
   ([#1107](https://github.com/nix-rust/nix/pull/1107))
 
@@ -779,5 +782,3 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#335](https://github.com/nix-rust/nix/pull/335))
 
 ## [0.5.0] 2016-03-01
-- Added `posix_fallocate`.
-  ([#1105](https://github.com/nix-rust/nix/pull/1105))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -779,3 +779,5 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#335](https://github.com/nix-rust/nix/pull/335))
 
 ## [0.5.0] 2016-03-01
+- Added `posix_fallocate`.
+  ([#1105](https://github.com/nix-rust/nix/pull/1105))
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -505,3 +505,24 @@ mod posix_fadvise {
         Errno::result(res)
     }
 }
+
+#[cfg(any(
+    target_os = "linux",
+    target_os = "android",
+    target_os = "emscripten",
+    target_os = "fuchsia",
+    any(target_os = "wasi", target_env = "wasi"),
+    target_os = "freebsd"
+))]
+pub fn posix_fallocate(
+    fd: RawFd,
+    offset: libc::off_t,
+    len: libc::off_t
+) -> Result<()> {
+    let res = unsafe { libc::posix_fallocate(fd, offset, len) };
+    match Errno::result(res) {
+        Err(err) => Err(err),
+        Ok(0) => Ok(()),
+        Ok(errno) => Err(crate::Error::Sys(Errno::from_i32(errno))),
+    }
+}
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -256,7 +256,7 @@ mod test_posix_fallocate {
             Ok(_) => {
                 let mut data = [1u8; LEN];
                 assert_eq!(tmp.read(&mut data).expect("read failure"), LEN);
-                assert_eq!(&data as &[u8], &[0u8; LEN] as &[u8]);
+                assert_eq!(&data[..], &[0u8; LEN][..]);
             }
             Err(nix::Error::Sys(Errno::EINVAL)) => {
                 // POSIX requires posix_fallocate to return EINVAL both for
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -231,3 +231,68 @@ mod test_posix_fadvise {
         assert_eq!(errno, Errno::ESPIPE as i32);
     }
 }
+
+#[cfg(any(target_os = "linux",
+          target_os = "android",
+          target_os = "emscripten",
+          target_os = "fuchsia",
+          any(target_os = "wasi", target_env = "wasi"),
+          target_os = "freebsd"))]
+mod test_posix_fallocate {
+
+    use tempfile::NamedTempFile;
+    use std::{io::Read, os::unix::io::{RawFd, AsRawFd}};
+    use nix::errno::Errno;
+    use nix::fcntl::*;
+    use nix::unistd::pipe;
+
+    #[test]
+    fn success() {
+        const LEN: usize = 100;
+        let mut tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let res = posix_fallocate(fd, 0, LEN as libc::off_t);
+        match res {
+            Ok(_) => {
+                let mut data = [1u8; LEN];
+                assert_eq!(tmp.read(&mut data).expect("read failure"), LEN);
+                assert_eq!(&data as &[u8], &[0u8; LEN] as &[u8]);
+            }
+            Err(nix::Error::Sys(Errno::EINVAL)) => {
+                // `posix_fallocate` returned EINVAL.
+                // According to POSIX.1-2008, its implementation shall
+                // return `EINVAL` if `len` is 0, `offset` is negative or
+                // the filesystem does not support this operation.
+                // According to POSIX.1-2001, its implementation shall
+                // return `EINVAL` if `len` is negative, `offset` is
+                // negative or the filesystem does not support this
+                // operation; and may return `EINVAL` if `len` is 0.
+                // However, this test is using `offset=0` and non-zero
+                // `len`.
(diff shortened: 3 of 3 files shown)


asomers:
It's not nix's job to fix kernel bugs; we have to live with them.  You should adjust the test so that it accepts all error codes that the kernel might return in that situation.

dingxiangfei2009:
Sorry, forgot to cross reference here: CraneStation/wasi-common#16

test/test_fcntl.rs
```diff
@@ -232,3 +232,46 @@ mod test_posix_fadvise {
...
+    use nix::errno::Errno;
+    use nix::fcntl::*;
+    use nix::unistd::pipe;
+
+    #[test]
+    fn test_success() {
+        const LEN: usize = 100;
+        let mut tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let res = posix_fallocate(fd, 0, LEN as libc::off_t).unwrap();
```
asomers: This test will fail if /tmp resides on a file system that doesn't support `posix_fallocate`, like ZFS.  The test must be adjusted to handle that error.  Unfortunately, POSIX stipulates that the errno in that case shall be `EINVAL`, which can also indicate several other problems.

test/test_fcntl.rs
```diff
@@ -232,3 +232,46 @@ mod test_posix_fadvise {
...
+        assert_eq!(&data as &[u8], &[0u8; LEN] as &[u8]);
+        assert_eq!(res, 0);
+    }
+
+    #[test]
+    fn test_errno() {
+        let (rd, _wr) = pipe().unwrap();
+        let errno = posix_fallocate(rd as RawFd, 0, 100).unwrap();
+        match Errno::from_i32(errno) {
+            Errno::EBADF
```
asomers: Can't you narrow this list down?  I don't think there are very many different error codes that could be returned in this case.

test/test_fcntl.rs
dingxiangfei2009: Thanks! I wasn't aware of this case.

test/test_fcntl.rs
dingxiangfei2009: Sure!

test/test_fcntl.rs
dingxiangfei2009: I have added in some explanation in the test, in case the test fails possibly due to filesystem not supporting this operation.

src/fcntl.rs
```diff
@@ -504,3 +504,16 @@ mod posix_fadvise {
         Errno::result(res)
     }
 }
+
+#[cfg(any(
```
asomers: You can add dragonfly, too.  You can also add NetBSD, if you make a PR to libc first.

test/test_fcntl.rs
```diff
@@ -232,3 +232,56 @@ mod test_posix_fadvise {
...
+        let res = posix_fallocate(fd, 0, LEN as libc::off_t).unwrap();
+        if res == 0 {
+            let mut data = [1u8; LEN];
+            assert_eq!(tmp.read(&mut data).expect("read failure"), LEN);
+            assert_eq!(&data as &[u8], &[0u8; LEN] as &[u8]);
+        } else {
+            match Errno::from_i32(res) {
+                Errno::EINVAL => {
+                    eprintln!(r#"
+`posix_fallocate` returned EINVAL.
```
asomers: No need to print this info out; a comment is sufficient.  And please keep it to 80 characters.

test/test_fcntl.rs
```diff
@@ -232,3 +232,56 @@ mod test_posix_fadvise {
...
+mod test_posix_fallocate {
+
+    use tempfile::NamedTempFile;
+    use std::{io::Read, os::unix::io::{RawFd, AsRawFd}};
+    use nix::errno::Errno;
+    use nix::fcntl::*;
+    use nix::unistd::pipe;
+
+    #[test]
+    fn test_success() {
```
asomers: Putting "test" in the function name is redundant with the "test" in the module name.  You can shorten it to "success".

src/fcntl.rs
```diff
@@ -504,3 +504,16 @@ mod posix_fadvise {
...
+    target_os = "linux",
+    target_os = "android",
+    target_os = "emscripten",
+    target_os = "fuchsia",
+    any(target_os = "wasi", target_env = "wasi"),
+    target_env = "freebsd"
+))]
+pub fn posix_fallocate(fd: RawFd, offset: libc::off_t, len: libc::off_t) -> Result<libc::c_int> {
```
asomers: A better return type would be `Result<()>`, because the only non-error value ever returned is 0.

test/test_fcntl.rs
```diff
@@ -232,3 +232,56 @@ mod test_posix_fadvise {
...
+                }
+                errno => panic!("unexpected errno={}", errno),
+            }
+        }
+    }
+
+    #[test]
+    fn test_errno() {
+        let (rd, _wr) = pipe().unwrap();
+        let errno = posix_fallocate(rd as RawFd, 0, 100).unwrap();
```
asomers: Shouldn't this be `unwrap_err`?

test/test_fcntl.rs
```diff
@@ -232,3 +232,56 @@ mod test_posix_fadvise {
         assert_eq!(errno, Errno::ESPIPE as i32);
     }
 }
+
+#[cfg(any(target_os = "linux",
+          target_os = "android",
+          target_os = "emscripten",
+          target_os = "fuchsia",
+          any(target_os = "wasi", target_env = "wasi"),
+          target_env = "freebsd"))]
```
asomers: s/target_env/target_os/

src/fcntl.rs
```diff
@@ -504,3 +504,16 @@ mod posix_fadvise {
...
     }
 }
+
+#[cfg(any(
+    target_os = "linux",
+    target_os = "android",
+    target_os = "emscripten",
+    target_os = "fuchsia",
+    any(target_os = "wasi", target_env = "wasi"),
+    target_env = "freebsd"
```
asomers: s/target_env/target_os/

src/fcntl.rs
dingxiangfei2009: The spec says that `posix_fallocate` returns the error number and it does not set `errno`.
`fallocate` returns `-1` on error and set `errno`. The two functions are quite different.

test/test_fcntl.rs
dingxiangfei2009: Okay. It is in comment now.

src/fcntl.rs
dingxiangfei2009: I have converted to return type to `Result<()>`. However, I have concerns on whether we should follow POSIX and return Errno directly?

test/test_fcntl.rs
dingxiangfei2009: Thank you!

dingxiangfei2009:
Oh, sorry for squashing again. I will not do this in later commits.
Also, I am going to add an entry to the `CHANGELOG`, once my question about the function signature of `posix_fallocate` is resolved.

test/test_fcntl.rs
```diff
@@ -232,3 +232,68 @@ mod test_posix_fadvise {
...
+        let fd = tmp.as_raw_fd();
+        let res = posix_fallocate(fd, 0, LEN as libc::off_t);
+        match res {
+            Ok(_) => {
+                let mut data = [1u8; LEN];
+                assert_eq!(tmp.read(&mut data).expect("read failure"), LEN);
+                assert_eq!(&data as &[u8], &[0u8; LEN] as &[u8]);
+            }
+            Err(nix::Error::Sys(Errno::EINVAL)) => {
+                // `posix_fallocate` returned EINVAL.
```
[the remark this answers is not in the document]
dingxiangfei2009: Thank you for your suggestion! :smile:

test/test_fcntl.rs
```diff
@@ -232,3 +232,68 @@ mod test_posix_fadvise {
...
+                // According to POSIX.1-2008, its implementation shall
+                // return `EINVAL` if `len` is 0, `offset` is negative or
+                // the filesystem does not support this operation.
+                // According to POSIX.1-2001, its implementation shall
```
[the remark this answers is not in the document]
dingxiangfei2009: :scissors: ed

test/test_fcntl.rs
```diff
@@ -231,3 +231,60 @@ mod test_posix_fadvise {
...
+    fn success() {
+        const LEN: usize = 100;
+        let mut tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let res = posix_fallocate(fd, 0, LEN as libc::off_t);
+        match res {
+            Ok(_) => {
+                let mut data = [1u8; LEN];
+                assert_eq!(tmp.read(&mut data).expect("read failure"), LEN);
+                assert_eq!(&data as &[u8], &[0u8; LEN] as &[u8]);
```
asomers: Instead of casting, you should be able to do `&data[..], &[0u8; LEN][..]`

CHANGELOG.md
```diff
@@ -773,3 +773,5 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#335](https://github.com/nix-rust/nix/pull/335))
 
 ## [0.5.0] 2016-03-01
+- Added `posix_fallocate`.
```
asomers: This CHANGELOG entry is seriously dislocated.  It should be up in the "Unreleased" section.

dingxiangfei2009: Oh, sorry

dingxiangfei2009:
It is surprising to see tests failing under `x86_64-unknown-linux-gnu`.

asomers:
Sorry, but you need to rebase again to fix the test failures.

PR #1106, opened 2019-08-11
Fix TODO: exclude fdatasync from Apple systems

jabedude:
Small PR to fix a TODO in unistd.rs.

Signed-off-by: Josh Abraham <sinisterpatrician@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,6 +95,11 @@ matrix:
       script:
         - cargo update -Zminimal-versions
         - cargo check
+    - language: generic
+      name: redox
+      script:
+        - docker pull redoxos/redoxer
+        - docker run  -v $(pwd):$(pwd) -w $(pwd) redoxos/redoxer sh -c 'rm -rf ~/.redoxer && redoxer test'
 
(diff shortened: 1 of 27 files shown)


asomers:
Did you check whether libc defines fdatasync for dragonfly BSD?

asomers:
And openbsd too. Neither of those is checked in ci.

jabedude:
Sure thing, looks like it is not defined for fdatasync and is defined for openbsd. I'll go ahead and exclude it.

asomers:
Could you please add a mention to the CHANGELOG?

jabedude:
Sure thing, done!

asomers:
The CHANGELOG entry got inserted into the wrong section; probably a merge error on your side.  Please move it up to the `[Unreleased]` section.  And I would put it under `Fixed` rather than `Changed`.  The latter suggests that some action might be required by a user when upgrading.  Do that and squash the commits, and I'll merge it.

asomers:
Please squash the commits, too.

PR #1109, opened 2019-08-22
fix readlink/readlinkat to return too long only when it is long

sendilkumarn:
Currently readlink returns `ENAMETOOLONG` when the bufferSize is equal to the result size.  

Consider the below case

```c++
int main(int argc, const char * argv[]) {
    
    std::string filename = "/tmp/test-dir/target";
    
    size_t bufferSize = 9;
    char* buffer = new char[bufferSize];
    size_t rc = readlink (filename.c_str(), buffer, bufferSize); 
   // Since `readlink's` output depends on the `bufferSize`. If the`bufferSize` is 9 or greater than 
   // 9 then it will return 9 or else it will return `bufferSize` as the value of `rc`. 

    return 0;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,18 +10,17 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1109](https://github.com/nix-rust/nix/pull/1109))
 
   ```rust
-  use nix::fcntl::{readlink, readlinkat};
-
+  # use nix::fcntl::{readlink, readlinkat};
   // the buffer argument of `readlink` and `readlinkat` has been removed,
   // and the return value is now an owned type (`OsString`).
   // Existing code can be updated by removing the buffer argument
   // and removing any clone or similar operation on the output
 
   // old code `readlink(&path, &mut buf)` can be replaced with the following
-  readlink(&path); // this returns OsString
+  let _: OsString = readlink(&path);
   
   // old code `readlinkat(dirfd, &path, &mut buf)` can be replaced with the following
-  readlinkat(dirfd, &path); // this returns OsString
+  let _: OsString = readlinkat(dirfd, &path);
   ```
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ```rust
   use nix::fcntl::{readlink, readlinkat};
 
-  readlink(&path);
-  readlinkat(dirfd, &path);
+  // the buffer argument of `readlink` and `readlinkat` has been removed,
+  // and the return value is now an owned type (`OsString`).
+  // Existing code can be updated by removing the buffer argument
+  // and removing any clone or similar operation on the output
+
+  // old code `readlink(&path, &mut buf)` can be replaced with the following
+  readlink(&path); // this returns OsString
+  
+  // old code `readlinkat(dirfd, &path, &mut buf)` can be replaced with the following
+  readlinkat(dirfd, &path); // this returns OsString
   ```
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased] - ReleaseDate
 ### Added
 ### Changed
-- Changed `readlink` and `readlinkat` to return `osString`
+- Changed `readlink` and `readlinkat` to return `OsString`
   ([#1109](https://github.com/nix-rust/nix/pull/1109))
+
+  ```rust
+  use nix::fcntl::readlink;
+
+  readlink!(&path);
+  ```
+
 ### Fixed
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ```rust
   use nix::fcntl::{readlink, readlinkat};
 
-  readlink!(&path);
-  readlinkat!(dirfd, &path);
+  readlink(&path);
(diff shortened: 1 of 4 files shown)


asomers:
I'm afraid I don't understand the problem.  Could you please be more clear?  Give a complete example demonstrating the wrong behavior, and what you think the correct behavior should be.

sendilkumarn:
> I'm afraid I don't understand the problem. 

Oops sorry! Let me try to clarify.

Consider the following folder structure

```
/tmp 
  |____ target
  |____ test-dir
       |____ target (symlinked folder)
```

With nix when we try to `readLink` this scenario.
```Rust
let src = "/tmp/target; 
let dest = "tmp/target/test-dir";
```

It works correctly when we pass in a buffer that is bigger than the src length

```Rust
let mut buf = vec![0; src.len() + 1]; 
println!("{}", str::from_utf8(&buf).unwrap()); // prints "/tmp/target"
```

But when we pass in a buffer that is smaller than the `src` length, it returns `ENAMETOOLONG` exception.

```Rust
let mut buf = vec![0; src.len()];  // returns `ENAMETOOLONG` 
```
Instead it should truncate the contents (to a length of buffer size), in case the buffer is too small to hold all of the contents. - from [here](https://linux.die.net/man/2/readlink)

asomers:
Truncating the result is what readlink(2) does.  But Nix is supposed to be safe and easy to use; we don't want to truncate any results.  Perhaps `ENAMETOOLONG` isn't the best error to return, but a truncated result would be even worse.

sendilkumarn:
> but a truncated result would be even worse.

That makes sense. I agree but this means we should always send in a buffer that is greater than the required size or if not it fails. 

We should at least allow the buffer to have the same size as that of the required result. What do you think? 

> Perhaps ENAMETOOLONG isn't the best error to return

Can we return EINVAL instead of ENAMETOOLONG?

asomers:
> > but a truncated result would be even worse.
> 
> That makes sense. I agree but this means we should always send in a buffer that is greater than the required size or if not it fails.
> 
> We should at least allow the buffer to have the same size as that of the required result. What do you think?
> 
> > Perhaps ENAMETOOLONG isn't the best error to return
> 
> Can we return EINVAL instead of ENAMETOOLONG?

No, that would have the same problem.  The problem with both of those error codes is that they overload errors used by the OS.  Better to use our own error type.  Better yet would be to never return an error.  How about changing that function to allocate its storage internally and return an `OsString`, a little like this?

```rust
fn readlink<P: ...>(path: P) -> Result<OsString> {
    let v = Vec::with_capacity(libc::PATH_MAX);
    let r = libc::readlink(..., v.as_mut_ptr() as *mut c_char, v.capacity());
    if let Ok(len) = Errno::result(r) {
        unsafe { v.set_len(len) };
        OsString::from_vec(v)
    } else {
        /* error handling */
    }
}

**Comment by sendilkumarn** [comment]
@asomers That is a good idea. I updated the code to reflect your suggested changes.

File: CHANGELOG.md [hunk 1]
```diff
@@ -6,8 +6,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased] - ReleaseDate
 ### Added
 ### Changed
-- Changed `readlink` and `readlinkat` to return `osString`
+- Changed `readlink` and `readlinkat` to return `OsString`
   ([#1109](https://github.com/nix-rust/nix/pull/1109))
+
```
asomers: This example isn't going to mean much to a reader.  I'm thinking something in English, like "the `buffer` argument of `readlink` and `readlinkat` has been removed, and the return value is now an owned type.  Existing code can be updated by removing the buffer argument and removing any `clone` or similar operation on the output."

src/fcntl.rs
```diff
@@ -188,19 +188,21 @@ fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
 }
 
 pub fn readlink<'a, P: ?Sized + NixPath>(path: &P) -> Result<OsString> {
-    let mut v = vec![0u8; libc::PATH_MAX as usize];
+    let len = libc::PATH_MAX as usize;
+    let mut v = Vec::with_capacity(len);
     let res = path.with_nix_path(|cstr| {
-        unsafe { libc::readlink(cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.len() as size_t) }
+        unsafe { libc::readlink(cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, len as size_t) }
```
asomers: You seem to be unaware of the `capacity` method.  You should write it like this:
```rust
let mut v = Vec::with_capacity(libc::PATH_MAX as usize);
let res = path.with_nix_path(|cstr| {
    unsafe { libc::readlink(cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.capacity() as size_t) }
```
```

File: CHANGELOG.md [hunk 3]
```diff
@@ -12,8 +12,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ```rust
   use nix::fcntl::{readlink, readlinkat};
```
asomers: This line isn't important to readers.  You can suppress it with a leading `#`, and delete the following blank line.

CHANGELOG.md
```diff
@@ -12,8 +12,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
 
-  readlink(&path);
-  readlinkat(dirfd, &path);
+  // the buffer argument of `readlink` and `readlinkat` has been removed,
+  // and the return value is now an owned type (`OsString`).
+  // Existing code can be updated by removing the buffer argument
+  // and removing any clone or similar operation on the output
+
+  // old code `readlink(&path, &mut buf)` can be replaced with the following
+  readlink(&path); // this returns OsString
```
asomers: The code would be more self-documenting like this:
```rust
let _: OsString = readlink(&path);
```

PR #1114, opened 2019-08-31
Replace most instances of mem::uninitialized with mem::MaybeUninit

asomers:
Only two instances remain:

* For the deprecated sys::socket::CmsgSpace::new.  We should probably
  just remove that method.

* For sys::termios::Termios::default_uninit.  This will require some
  more thought.


asomers:
@Susurrus this is starting to hold up other PRs, so I'm going to merge it tonight.

PR #1117, opened 2019-09-01
Fixes #1116 - Feature flag and target_env check to selectively include GNU function wrappers

### LINKED ISSUE #1116: Add a feature + cfg attribute to exclude GNU extension functions
Hi, 
I was wondering if it would be possible to add a feature flag and the appropriate cfg calls to separate POSIX functions from the GNU specific extensions (usable when _GNU_SOURCE is defined). 

My use case is as follows. I am using some packages (lucet) that have a dependency on the nix packages. While I'm using these packages in a Linux system, we are linking against libraries without the GNU extension functions... Thus the presence of calls to `process_vm_writev`, `process_vm_readv` and `setns` is causing problems. Additionally, it seems that the nix wrappers for these functions cannot be eliminated by dead code removal, and so persist in the final library, resulting in missing symbol errors.

Currently, it seems that these functions are protected only with `#[cfg(any(target_os = "linux"))]`. Thus the easiest workaround that comes to mind is to pretend that target_os is not linux, but this feels hacky at best... I was wondering if it would be possible to add a new feature flag which we can set at compile time, that would exclude these GNU specific functions...

(also, I apologize if I got some details wrong... I'm still quite new to rust :) )

Please let me know... If required, I am also happy to contribute a patch for the same...

shravanrn:
All tests pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ caps = "0.3.1"
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
 sysctl = "0.1"
 
+[features]
+exclude_gnu = []
+
 [[test]]
 name = "test"
 path = "test/test.rs"
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -10,6 +10,7 @@ mod sched_linux_like {
     use libc::{self, c_int, c_void};
     use std::mem;
     use std::option::Option;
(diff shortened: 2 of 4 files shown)


asomers:
As discussed in the issue, we won't be merging this solution.  It's too narrow, and not required by most users.

PR #1118, opened 2019-09-02
Document unimplemented flag for fchmodat

0xpr03:


0xpr03:
I can understand that, but it's really irritating for flag to throw errors just so you can find out later that they just never implemented this.

PR #1120, opened 2019-09-04
Fix length of abstract socket address

### LINKED ISSUE #1119: UnixAddr::new_abstract() is using the wrong length
https://github.com/nix-rust/nix/blob/a4a465d25567f163f9552b977eb4d17435251d41/src/sys/socket/addr.rs#L549-L565

Here, the `ret.sun_path.len()` should be `path.len() + 1`

yshui:
NULL bytes have no special significance in an abstrace address, and the
length of the address is solely decided by the length member. If the
length is set to sun_path.len(), all the NULL bytes will be considered
part of the address.

Tests are updated accordingly.

Closes #1119

Signed-off-by: Yuxuan Shui <yshuiv7@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   ([#1107](https://github.com/nix-rust/nix/pull/1107))
 
 ### Fixed
+- Fix length of abstract socket addresses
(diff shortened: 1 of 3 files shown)


yshui:
@asomers done

asomers:
CHANGELOG entries should all have a link to the PR.

yshui:
@asomers fixed

PR #1121, opened 2019-09-06
Fix #1093

### LINKED ISSUE #1093: exec family functions take CString instead of &CStr
For example, the signature of `execve` is:

    pub fn execve(path: &CString, args: &[CString], env: &[CString]) -> Result<Void>

Surely there's no point in taking a `&CString` as opposed to a `&CStr`, right? And couldn't the slices just contain `&CStr`s since it's not possible to move out of a slice anyway?

tathanhdinh:
Hello,
This PR fixes the problem discovered in #1093.
Thanks for any comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   `self` by value.
   ([#1107](https://github.com/nix-rust/nix/pull/1107))
 
+- Type `&CString` for parameters of `exec(v|ve|vp|vpe|veat)` are changed to `&CStr`.
+  ([#1121](https://github.com/nix-rust/nix/pull/1121))
+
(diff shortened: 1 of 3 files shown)


tathanhdinh:
Thank you for the feedback. CHANGELOG modified.

asomers:
Looks good now.  As soon as PR #1123 is done, you'll be able to rebase this one and the test failures will be fixed.

tathanhdinh:
Thank you. I'm wating for PR #1123 now.

asomers:
You should be able to rebase now.

tathanhdinh:
Thank you. The PR is rebased.

otavio:
@tathanhdinh the commit log could receive some love, don't you think?

I know the issue is detailed but it'd be good if the commit log could be extended with it.

PR #1124, opened 2019-09-14
Where relevant, replace equality checks in assert! with assert_eq!

AdminXVII:
`assert_eq!` gives more debug info when the test fails by default than `assert!`. This should help make debugging easier.

### DIFF
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -329,23 +329,17 @@ pub fn ev_set(ev: &mut KEvent,
 fn test_struct_kevent() {
     let udata : intptr_t = 12345;
 
-    let expected = libc::kevent{ident: 0xdead_beef,
-                                filter: libc::EVFILT_READ,
-                                flags: libc::EV_ONESHOT | libc::EV_ADD,
-                                fflags: libc::NOTE_CHILD | libc::NOTE_EXIT,
-                                data: 0x1337,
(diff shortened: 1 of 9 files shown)


asomers:
Thanks!  This looks great.  BTW, I hope to have the test failures fixed soon in PR #1123 .

AdminXVII:
Cool!

asomers:
I don't understand.  What problem did that last commit fix?

AdminXVII:
On mac, the `kevent` struct use a packed alignment (according to the previously failing tests). `assert_eq` took a reference to the field, which is unsafe because the value can be unaligned. By adding brackets, the value is copied, so it's correctly aligned.

asomers:
I don't think that's true, at least not in general.  Braces can certainly capture references.  For this specific case, how about removing the assignment of `expected` altogether?  It's not necessary.  Instead, you can put literals directly into the `assert_eq!` statements.

AdminXVII:
Braces can capture references because they are copy. Since what's inside the brace is a copy value, not a reference, this is not a problem, even if the macro expands to something that takes the reference of it. That said I fixed it with your suggestion.

asomers:
Please squash the commits and I'll merge.  Oh, and you'll probably have to rebase now that PR #1123 is done.

AdminXVII:
@asomers Done

Issue #1029, opened 2019-02-13
Missing Flag For nix::unistd::fchownat


gatoWololo:
Hello!

Currently `fchownat<P: ?Sized + NixPath>(dirfd: Option<RawFd>, path: &P, owner: Option<Uid>, group: Option<Gid>,  flag: FchownatFlags) -> Result<()>` takes an enum as it's `flag` argument nix::unistd::FchownatFlags. Where the variants are either `FollowSymlink` or `NoFollowSymlink`. According to `man 2 fchownat` since Linux 2.6.39 `fchownat()` may also takes `AT_EMPTY_PATH` as part of the flags bitmask.

This is currently not possible with the current implementation provided by Nix, the enum will probably have to be changed to a bitflag.

asomers:
This sounds like an easy thing to fix.  Would you like to write a PR to add the new flag?

gatoWololo:
Sure! I can do that.

otavio:
I was looking at the flags and the `AT_EMPTY_PATH` is Linux specific. Couldn't same thing to be done using `fchown()`?

dubiousjim:
Bump. Would be great to get the functionality of `fchown` exposed, either as a standalone, or via the `AT_EMPTY_PATH` flag for `fchownat`. If the latter is Linux-specific, then the former seems preferable.

asomers:
Yes, I think `fchown` would be a good addition.  And an easy one.  Would you care to submit a PR< @dubiousjim ?

PR #1128, opened 2019-09-23
Add EmptyPath for FchownatFlags

tathanhdinh:
Hello,

This PR is probably a fix for #1029. Thanks for any response.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -631,6 +631,8 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
 pub enum FchownatFlags {
     FollowSymlink,
     NoFollowSymlink,
+    #[cfg(target_os = "linux")]
+    EmptyPath,    
 }
 
 /// Change the ownership of the file at `path` to be owned by the specified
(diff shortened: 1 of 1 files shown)


otavio:
Couldn't same thing to be done using `fchown()`?

tathanhdinh:
@otavio thank you for the feedback but IMHO `fchown` doesn't need flags.

otavio:
@tathanhdinh you got me wrong. I meant if it is worth adding this or using the `fchown` as an alternative.

tathanhdinh:
Ah, sorry.
I don't know whether `fchown` can always be used. Does it always follow symbolic link?

otavio:
The `fchown` operates on the file description and in this case, seems equivalent to `fchownat` with `EmptyPath` flag.

src/unistd.rs
```diff
@@ -631,6 +631,8 @@ pub fn chown<P: ?Sized + NixPath>(path: &P, owner: Option<Uid>, group: Option<Gi
 pub enum FchownatFlags {
     FollowSymlink,
     NoFollowSymlink,
+    #[cfg(target_os = "linux")]
+    EmptyPath,
```
SteveLauC, line 635: If we add this flag, then this type should no longer be an enum but a bitflag cause you can do `AT_EMPTY_PATH` and ` AT_SYMLINK_NOFOLLOW` at the same time

PR #1131, opened 2019-09-29
implement mknodat

zmlcc:
This adds the `mknodat` function, which is part of POSIX [https://pubs.opengroup.org/onlinepubs/9699919799/functions/mknod.html](https://pubs.opengroup.org/onlinepubs/9699919799/functions/mknod.html)

`mknodat` seems not implemented in OSX, so I add conditional compilation attributes to exclude `ios` and `macos`

I don't find any test code about `mknod`, So tests of `mknodat` are not added either

### DIFF
--- a/src/sys/stat.rs
+++ b/src/sys/stat.rs
@@ -52,6 +52,27 @@ pub fn mknod<P: ?Sized + NixPath>(path: &P, kind: SFlag, perm: Mode, dev: dev_t)
     Errno::result(res).map(drop)
 }
 
+// mknodat not implemented in OSX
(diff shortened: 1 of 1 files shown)


asomers:
Yes, `mknod` may not have any test code, and that's a shame.  You should still add tests for `mknodat`, and while you're at it it would be great to add them for `mknod` as well.

zmlcc:
`mknod` test failed in many arch/os.   Does  `mknod` was not widely implemented in different systems ?

asomers:
The tests are probably failing due to permissions errors.  Creating a device node requires root.

PR #1136, opened 2019-10-15
Properly initialize msghdr when using musl

yshui:
Because of the use of MaybeUninit::uninit, the padding fields in msghdr,
which only present on musl builds, are not initialized. Causing garbage
data to be sent to the kernel. This change ensures the paddings are
always zeroed.


posborne:
Thanks @yshui change LGTM.  Can you amend the commit with a changelog addition per https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests

yshui:
@posborne Done!

PR #1137, opened 2019-10-15
Rustup minimal

asomers:
Fix CI on FreeBSD after rustup 1.20.0.


asomers:
The builds failed with errors like
```
Unable to find image 'rustembedded/cross:x86_64-unknown-linux-gnu-0.1.16' locally
docker: Error response from daemon: received unexpected HTTP status: 503 Service Unavailable.
```
That sounds like it might be a transient error.  Let's try again.

bors retry

asomers:
That command doesn't seem to be working.

bors r+

PR #1138, opened 2019-10-16
Add Signal::as_str() to get representation as static string

MikailBag:
# Motivation
Currently string representation of signal can be obtained with AsRef<str> impl.
But it has downside: returned string's lifetime is bound to lifetime of signal, so &str must be converted to String. This allocation is avoidable, because as_ref() only returns static strings. To fix this problem, my PR adds Signal::as_str() method, which returns static string.

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -112,9 +112,14 @@ impl FromStr for Signal {
     }
 }
 
-impl AsRef<str> for Signal {
-    fn as_ref(&self) -> &str {
-        match *self {
+impl Signal {
+    /// Returns name of signal.
+    ///
+    /// This function is equivalent to `<Signal as AsRef<str>>::as_ref()`,
+    /// with difference that returned string is `'static`
+    /// and not bound to `self`'s lifetime.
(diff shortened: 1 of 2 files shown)


src/sys/signal.rs
```diff
@@ -112,8 +112,11 @@ impl FromStr for Signal {
     }
 }
 
-impl AsRef<str> for Signal {
-    fn as_ref(&self) -> &str {
+impl Signal {
+    /// Returns name of signal
+    /// It is equivalent to <Signal as AsRef<str>>::as_ref(), with difference that returned string is
```
asomers: Needs backticks.

MikailBag:
Do docs look OK now?

asomers:
Yeah, looks good.  Just wrap the comment to 80 columns and I'll merge.

MikailBag:
Done

MikailBag:
Note: I changed signature of as_str to take self by-value, because Signal is Copy

MikailBag:
Squashed.

PR #1139, opened 2019-10-19
Add `Users` and `Group` related functions

otavio:
This was a collaborative work between Johannes Schilling
<dario@deaktualisierung.org>, Fredrick Brennan <copypaste@kittens.ph>
and myself.

This PR is a split-off from #864 so we merge ready parts first. Next, we work on the iterators.

Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

PR #1140, opened 2019-10-20
Expose getpwuid

FauxFaux:
Implement `getpwuid_r`, which loads the `passwd` entry for a given `Uid`, allowing you to find the user's name, shell, homedir, etc.

I did not know what to do about the string ownership model. It returns us a `struct` full of `*const c_char` pointers, which supposedly point to our buffer. I wrote new code to check these, to some extent, then trust `CStr` to do the rest. I did this because I couldn't find any other code in `nix` that did this (which must exist, someone point it out so I can slap my forehead), and to defer the cost of the checks to the time of use, i.e. so if you only want the `name`, you don't have to pay to check the `shell`.

The code should be safe without the extra checks, if the system follows the rules, and I have no reason to believe otherwise; perhaps just deleting the extra checks is reasonable?

Also not handled: dynamically increasing the buffer size if the function returns that the buffer is too small. `GETPW_R_SIZE_MAX` is `1024` on Ubuntu 19.10 / 5.3 / amd64.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -7,6 +7,7 @@ use fcntl::FcntlArg::F_SETFD;
 use libc::{self, c_char, c_void, c_int, c_long, c_uint, size_t, pid_t, off_t,
            uid_t, gid_t, mode_t, PATH_MAX};
 use std::{fmt, mem, ptr};
+use std::convert::TryFrom;
 use std::ffi::{CStr, OsString, OsStr};
 use std::os::unix::ffi::{OsStringExt, OsStrExt};
 use std::os::unix::io::RawFd;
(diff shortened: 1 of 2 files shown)


src/unistd.rs
```diff
@@ -1346,6 +1347,88 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    }
+
+    Ok(unsafe { CStr::from_ptr(ptr) })
+}
+
+/// Retrieve the specified user's `passwd` file entry.
+///
+/// See also [getpwuid(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getpwuid.html)
+pub fn getpwuid(uid: Uid) -> Result<Option<Passwd>> {
+    let mut inner: libc::passwd = unsafe { mem::zeroed() };
```
asomers: This should use `mem::MaybeUninit` instead of `mem::zeroed`.

src/unistd.rs
```diff
@@ -1346,6 +1347,88 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+    let mut result = ptr::null_mut();
```
asomers: `result`, too, should use `MaybeUninit`.

test/test_unistd.rs
```diff
@@ -473,6 +473,24 @@ cfg_if!{
     }
 }
 
+#[test]
+fn test_passwd() -> nix::Result<()> {
+    let passwd = getpwuid(getuid())?
```
asomers: Test functions should all return `()`.

asomers:
You should probably look at #1139.  It overlaps with this one.

otavio:
@FauxFaux please look at #1139 and see if you could provide any feedback on this.

asomers:
I'm going to close this PR, because I like #1139's approach better.  It solves the same problem but in a different way.

PR #1148, opened 2019-10-30
Implement sched::sched_getaffinity()

thib-ack:
Hello,

I found the function `sched::sched_setaffinity()` but I also needed to get the process affinity and I did not find the function `sched::sched_getaffinity()`
So I added the function which maps [sched_getaffinity(2)](https://linux.die.net/man/2/sched_getaffinity) which "get a process's CPU affinity mask" to the sched.rs file.
I hope the code match your guidelines (returned `Result<CpuSet>` instead of pointer parameter)
If that's not the case, tell me, I will fix asap.

I hope this will help !

Thanks,
Thibaut

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -46,6 +46,11 @@ mod sched_linux_like {
 
     pub type CloneCb<'a> = Box<dyn FnMut() -> isize + 'a>;
 
+    /// CpuSet represent a bit-mask of CPUs.
+    /// CpuSets are used by sched_setaffinity and
+    /// sched_getaffinity for example.
+    ///
+    /// This is a wrapper around `libc::cpu_set_t`.
     #[repr(C)]
     #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
     pub struct CpuSet {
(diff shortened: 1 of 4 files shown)


src/sched.rs
```diff
@@ -96,6 +96,22 @@ mod sched_linux_like {
         Errno::result(res).map(drop)
     }
 
+    pub fn sched_getaffinity(pid: Pid) -> Result<CpuSet> {
+        let mut cpuset = CpuSet::new();
+        let res = unsafe {
+            libc::sched_getaffinity(
+                pid.into(),
+                mem::size_of::<CpuSet>() as libc::size_t,
+                & mut cpuset.cpu_set,
```
asomers: Style nit: there should be no space between `&` and `mut`.

src/sched.rs
```diff
@@ -96,6 +96,22 @@ mod sched_linux_like {
...
+            )
+        };
+
+        match res {
```
asomers: I think you can do this more simply by `Errno::result(res).and(Ok(cpuset))`

thib-ack:
- Issues fixed
- test added (in new dedicated file test_sched.rs)
- changelog added

Thanks

test/test_sched.rs
```diff
@@ -0,0 +1,36 @@
...
+fn test_sched_affinity() {
+    use nix::sched::{sched_getaffinity, sched_setaffinity, CpuSet};
+    use nix::unistd::Pid;
+    use std::mem;
+
+    // If pid is zero, then the mask of the calling process is returned.
+    let initial_affinity = sched_getaffinity(Pid::from_raw(0)).unwrap();
+    let mut at_least_one_cpu = false;
+    let mut last_valid_cpu = 0;
+    for field in 0..(8 * mem::size_of::<libc::cpu_set_t>()) {
```
asomers: What does the magic number 8 mean?  You should remove it if possible, or define it as a const if not.

test/test_sched.rs
```diff
@@ -0,0 +1,36 @@
+
+#[cfg(any(target_os = "android",
```
asomers: You should move the `#[cfg()]` into test/test.rs, for consistency with other files.

test/test_sched.rs
thib-ack: This value comes from the `CpuSet::set()` and `CpuSet::is_set()` methods.
https://github.com/nix-rust/nix/blob/master/src/sched.rs#L63
```rust
        pub fn is_set(&self, field: usize) -> Result<bool> {
            if field >= 8 * mem::size_of::<libc::cpu_set_t>() {
                Err(Error::Sys(Errno::EINVAL))
            } else {
                Ok(unsafe { libc::CPU_ISSET(field, &self.cpu_set) })
            }
        }
```

That's because there is no easy way to know the size of a CpuSet and in the test I want to check each bits of the CpuSet.

I can do differently:
- add a method `CpuSet::size()` which will return this value.
- call `CpuSet::is_set()` until I get a `Errno::EINVAL`
What do you prefer ?

thib-ack:
Hi,

I'll be AFK for the next 2 weeks unfortunately. I'll continue to work on this when I come back.
For example, I'm thinking the `sched::sched_getaffinity()` should take a `&CpuSet` as parameter instead of building one internally. That way, users could allocate the CpuSet once and use it many times in `sched_getaffinity()`.
You agree ?

Thanks,

asomers:
> Hi,
> 
> I'll be AFK for the next 2 weeks unfortunately. I'll continue to work on this when I come back.
> For example, I'm thinking the `sched::sched_getaffinity()` should take a `&CpuSet` as parameter instead of building one internally. That way, users could allocate the CpuSet once and use it many times in `sched_getaffinity()`.
> You agree ?
> 
> Thanks,

No, I disgree.  Return arguments are very C-ish.  It's much Rustier to use return values.  I doubt there will be any performance difference, because rustc is able to do what C++ people call "return by move".

test/test_sched.rs
asomers: Better to add a method to `CpuSet`.  But I would call it `count`, because that's what the C macros call it.

asomers:
You'll have to rebase to fix the build error on nightly.

src/sched.rs
```diff
@@ -46,44 +46,84 @@ mod sched_linux_like {
...
             } else {
                 Ok(unsafe { libc::CPU_SET(field, &mut self.cpu_set) })
             }
         }
 
+        /// Remove a CPU from CpuSet.
+        /// `field` is the CPU id to remove
         pub fn unset(&mut self, field: usize) -> Result<()> {
-            if field >= 8 * mem::size_of::<libc::cpu_set_t>() {
+            CpuSet::count();
```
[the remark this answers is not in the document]
thib-ack: oops, copy-paste mistake

src/sched.rs
```diff
@@ -46,44 +46,82 @@ mod sched_linux_like {
...
+    /// If pid is zero, then the calling thread is updated.
+    ///
+    /// The `cpuset` argument specifies the set of CPUs on which the thread
+    /// will be eligible to run.
+    ///
+    /// # Example
+    ///
+    /// Binding the current thread to CPU 0 can be done as follows:
+    ///
+    /// ```rust
```
asomers: I would probably make this `rust,no_run`.  Otherwise, you'll be changing the affinity of the doc test process.

PR #1150, opened 2019-11-01
sys: termios: Define VMIN and VTIME as VEOF and VEOL on sparc64

### LINKED ISSUE #1149: VMIN and VTIME should be defined on sparc64
Currently, the code assumes that ```VMIN``` and ```VTIME``` are not defined on sparc64.

However, the glibc has in https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/sparc/bits/termios-c_cc.h;h=bfe91f652dc9bd1bc885d228a7952cd156eaa9d0;hb=HEAD

```
/* User apps assume vmin/vtime is shared with eof/eol */
#define VMIN     VEOF
#define VTIME    VEOL
```

I'll check whether this needs to be fixed in the libc crate for sparc64. In any case, the conditional for sparc64 in src/sys/termios.rs to not define VMIN and VTIME should be removed as this makes rust-rustyline fail to build on sparc64: https://buildd.debian.org/status/fetch.php?pkg=rust-rustyline&arch=sparc64&ver=5.0.2-1&stamp=1570657917&raw=0

glaubitz:
On sparc64, glibc defines VMIN as VEOF and VTIME as VEOL for
termios, so we need to inherit these alias definitions for
nix-rust as well.

Fixes #1149

### DIFF
--- a/src/sys/termios.rs
+++ b/src/sys/termios.rs
@@ -583,6 +583,11 @@ libc_enum! {
     }
 }
 
+#[cfg(not(all(target_os = "linux", target_arch = "sparc64")))]
+pub const VMIN: SpecialCharacterIndices = SpecialCharacterIndices::VEOF;
+#[cfg(not(all(target_os = "linux", target_arch = "sparc64")))]
+pub const VTIME: SpecialCharacterIndices = SpecialCharacterIndices::VEOL;
+
 pub use libc::NCCS;
 #[cfg(any(target_os = "dragonfly",
           target_os = "freebsd",


glaubitz:
Yikes, you're right. I shouldn't be doing such things in the middle of the night.

glaubitz:
Okay, just opened another PR #1151 to fix this. Sorry for that.

glaubitz:
Also, I think the definitions need to be moved into ```libc_enum``` but I haven't verified that yet.

PR #1151, opened 2019-11-01
sys: termios: Fix inverted logic for [cfg()] conditional for sparc64

glaubitz:
The fix for #1149 has the logic for the [cfg()] conditional inverted
so that the aliases for VMIN and VTIME are defined on targets that
are not linux-sparc64.

### DIFF
--- a/src/sys/termios.rs
+++ b/src/sys/termios.rs
@@ -584,9 +584,10 @@ libc_enum! {
 }
 
 #[cfg(all(target_os = "linux", target_arch = "sparc64"))]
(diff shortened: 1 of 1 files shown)


glaubitz:
Before merging this, please let me have another test run. I need to verify again that the fix works properly.

glaubitz:
Okay, this works now and rustyline builds fine on sparc64.

asomers:
You need to rebase to fix the build failure on nightly.

glaubitz:
Thanks. I have done that now. Also, simplified the code a bit more.

PR #1157, opened 2019-12-01
Fix some race conditions in the integration tests

asomers:


asomers:
I found some other tests that need mutexes, too.

bors r-

PR #1161, opened 2019-12-02
Add support for IP_RECVERR

mcpherrinm:
This pull request adds support for setting the IP_RECVERR sockopt, and then decoding the resulting messages from recvmsg.  My intent is to use this to implement a traceroute program (And so need ICMP TTL exceeded messages), though IP_RECVERR is useful in general when writing UDP network programs.  It is, unfortunately, a linux-specific API.  

There's no tests here, but I've been trying it out with this bit of code:
https://github.com/mcpherrinm/traceroute/commit/8a92b6d66d7a44a60b261ff382f38f307b435ef5

SockExtendedErr is the struct type returned as a Cmsg from recvmsg.   It (like the C linux/errqueue.h which defines it) uses u8 and u32.  I think we'll want something more user-friendly here, so I'd say this PR is still a work-in-progress.

I am new to the `nix` codebase (and am not very experienced with Rust), so would appreciate early feedback.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -558,7 +558,7 @@ impl ControlMessageOwned {
                 ControlMessageOwned::Ipv4RecvErr(err, addr)
             },
             #[cfg(any(target_os = "android", target_os = "linux"))]
-            (libc::IPPROTO_IPV6, libc::IP_RECVERR) => {
+            (libc::IPPROTO_IPV6, libc::IPV6_RECVERR) => {
                 let ee = p as *const libc::sock_extended_err;
                 let err = ptr::read_unaligned(ee);
                 let addr = ptr::read_unaligned(libc::SO_EE_OFFENDER(ee) as *const sockaddr_in6);
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -261,21 +261,6 @@ impl Ipv6MembershipRequest {
     }
 }
 
-/// SockExtendedErr is the struct returned when IP_RECVERR is used to
-/// receive asyncronous IP errors.
-#[cfg(any(target_os = "android", target_os = "linux"))]
-#[repr(C)]
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub struct SockExtendedErr {
-    errno: u32,
-    origin: u8,
-    typ: u8,
-    code: u8,
-    pad: u8,
-    info: u32,
-    data: u32,
-}
-
 /// Create a buffer large enough for storing some control messages as returned
 /// by [`recvmsg`](fn.recvmsg.html).
 ///
@@ -479,9 +464,9 @@ pub enum ControlMessageOwned {
     ))]
     Ipv4RecvDstAddr(libc::in_addr),
     #[cfg(any(target_os = "android", target_os = "linux"))]
-    Ipv4RecvErr(SockExtendedErr, sockaddr_in),
+    Ipv4RecvErr(libc::sock_extended_err, sockaddr_in),
     #[cfg(any(target_os = "android", target_os = "linux"))]
-    Ipv6RecvErr(SockExtendedErr, sockaddr_in6),
+    Ipv6RecvErr(libc::sock_extended_err, sockaddr_in6),
     /// Catch-all variant for unimplemented cmsg types.
     #[doc(hidden)]
     Unknown(UnknownCmsg),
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -261,6 +261,21 @@ impl Ipv6MembershipRequest {
     }
 }
 
+/// SockExtendedErr is the struct returned when IP_RECVERR is used to
+/// receive asyncronous IP errors.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct SockExtendedErr {
```
asomers: Is this struct defined anywhere in the C headers?  If so, you should add its definition to the libc crate, and reference that here.  Nix shouldn't be in the business of repeating the headers.

mcpherrinm:
Yes, won't force-push anymore.  Just wanted to quickly fix a few cfg errors that caused the freebsd CI to fail.

src/sys/socket/mod.rs
mcpherrinm: It's defined in https://elixir.bootlin.com/linux/v5.4.1/source/include/uapi/linux/errqueue.h

I read the libc crate's RFC https://github.com/rust-lang/rfcs/blob/master/text/1291-promote-libc.md and it wasn't clear to me if this struct belonged there or not.

I am happy to open a PR there first and define it there.

asomers: Yes, please submit a PR to libc.

mcpherrinm:
I have opened a PR against libc @ https://github.com/rust-lang/libc/pull/1613

I will revise this pull request to use the struct definition from there once reviewed and merged.

Thank you for your time so far.

asomers:
Looks like this PR is still waiting on https://github.com/rust-lang/libc/pull/1616

src/sys/socket/mod.rs
```diff
@@ -546,6 +550,20 @@ impl ControlMessageOwned {
...
             },
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            (libc::IPPROTO_IP, libc::IP_RECVERR) => {
+                let ee = p as *const libc::sock_extended_err;
+                let err = ptr::read_unaligned(ee);
+                let addr = ptr::read_unaligned(libc::SO_EE_OFFENDER(ee) as *const sockaddr_in);
+                ControlMessageOwned::Ipv4RecvErr(err, addr)
+            },
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            (libc::IPPROTO_IPV6, libc::IP_RECVERR) => {
```
WGH-: It should be `IPV6_RECVERR` (tested that on Linux).

src/sys/socket/mod.rs
```diff
@@ -546,6 +550,20 @@ impl ControlMessageOwned {
...
+                let ee = p as *const libc::sock_extended_err;
+                let err = ptr::read_unaligned(ee);
+                let addr = ptr::read_unaligned(libc::SO_EE_OFFENDER(ee) as *const sockaddr_in6);
```
WGH-: `man 2 recvmsg`

> For local errors, no address is passed (this can be checked with the cmsg_len member  of  the  cmsghdr). 

Given that, I think `addr` should become an `Option`, and `len` should be handled appropriately.

WGH-:
Anyway, I tested the code with my toy TCP SYN scanner, and it works fine (except that `IPV6_RECVERR` thing that I fixed myself).

WGH-:
libc-0.2.67 has been released recently, which includes the missing bits, so it's no longer necessary to use a patch.

```diff
iff --git a/Cargo.toml b/Cargo.toml
index 29f0fca..8fb953e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,14 +16,11 @@ exclude     = [
 ]
 
 [dependencies]
-libc = { version = "0.2.60", features = [ "extra_traits" ] }
+libc = { version = "0.2.67", features = [ "extra_traits" ] }
 bitflags = "1.0"
 cfg-if = "0.1.2"
 void = "1.0.2"
 
-[patch.crates-io]
-libc = { git = "https://github.com/mcpherrinm/libc" }
-
 [target.'cfg(target_os = "dragonfly")'.build-dependencies]
 cc = "1"
```

mcpherrinm:
Thanks @WGH-.  I'll update this PR soon.

asomers:
Looks like you need to raise  the minimum version of libc

mcpherrinm:
Thanks. Sorry this is taking so long, but I have little free time for side project programming right now.

cemeyer:
We can probably close this one now.  Thanks!

asomers:
Yep.

PR #1163, opened 2019-12-15
Add setfsuid and setfsgid implementation for filesystem checks

gliderkite:
I noticed that the filesystem checks API `setfsuid` and `setfsgid` where missing (while available in `libc`). This PR adds the implementation for both of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (#[1185](https://github.com/nix-rust/nix/pull/1185))
 - `FsType` inner value made public.
   (#[1187](https://github.com/nix-rust/nix/pull/1187))
+- Added `unistd::setfsuid` and `unistd::setfsgid` to set the user or group
+  identity for filesystem checks per-thread.
+  (#[1163](https://github.com/nix-rust/nix/pull/1163))
 ### Changed
 ### Fixed
 ### Removed
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1398,6 +1398,28 @@ pub fn setgid(gid: Gid) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Set the user identity used for filesystem checks per-thread.
+/// On both success and failure, this call returns the previous filesystem user
+/// ID of the caller.
+///
+/// See also [setfsuid(2)](http://man7.org/linux/man-pages/man2/setfsuid.2.html)
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn setfsuid(uid: Uid) -> Uid {
+    let prev_fsuid = unsafe { libc::setfsuid(uid.into()) };
+    Uid::from_raw(prev_fsuid as uid_t)
+}
+
+/// Set the group identity used for filesystem checks per-thread.
+/// On both success and failure, this call returns the previous filesystem group
+/// ID of the caller.
+///
+/// See also [setfsgid(2)](http://man7.org/linux/man-pages/man2/setfsgid.2.html)
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn setfsgid(gid: Gid) -> Gid {
+    let prev_fsgid = unsafe { libc::setfsgid(gid.into()) };
+    Gid::from_raw(prev_fsgid as gid_t)
+}
+
 /// Get the list of supplementary group IDs of the calling process.
 ///
 /// [Further reading](http://pubs.opengroup.org/onlinepubs/009695399/functions/getgroups.html)
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -866,3 +866,41 @@ fn test_access_file_exists() {
     let _file = File::create(path.clone()).unwrap();
     assert!(access(&path, AccessFlags::R_OK | AccessFlags::W_OK).is_ok());
 }
+
+/// Tests setting the filesystem UID with `setfsuid`.
+#[cfg(any(target_os = "linux", target_os = "android"))]
+#[test]
+fn test_setfsuid() {
+    use std::os::unix::fs::PermissionsExt;
+    use std::{fs, thread};
+    require_capability!(CAP_SETUID);
+
+    // get the UID of the "nobody" user
+    let nobody = User::from_name("nobody").unwrap().unwrap();
+
+    // create a temporary file with permissions '-rw-r-----'
+    let file = tempfile::NamedTempFile::new().unwrap();
+    let temp_path = file.into_temp_path();
+    let temp_path_2 = (&temp_path).to_path_buf();
+    let mut permissions = fs::metadata(&temp_path).unwrap().permissions();
+    permissions.set_mode(640);
+
+    // spawn a new thread where to test setfsuid
+    thread::spawn(move || {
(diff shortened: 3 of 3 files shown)


asomers:
According to the man page, those syscalls are obsolete.  Do you have an actual use case for them?  If not, then we should reject this PR in the interest of resisting bloat.

gliderkite:
As far as I understand the note you are referring to:

> At the time when this system call was introduced, one process could send a signal to another process with the same effective user ID. This meant that if a privileged process changed its effective user ID for the purpose of file permission checking, then it could become vulnerable to receiving signals sent by another (unprivileged) process with the same user ID. The filesystem user ID attribute was thus added to allow a process to change its user ID for the purposes of file permission checking without at the same time becoming vulnerable to receiving unwanted signals. Since Linux 2.0, signal permission handling is different (see kill(2)), with the result that a process can change its effective user ID without being vulnerable to receiving signals from unwanted processes. Thus, setfsuid() is nowadays unneeded and should be avoided in new applications (likewise for setfsgid(2)).

states that `setfsuid` is not required anymore if the purpose was to protect the process from receiving signals from another process having the same effective UID, while it doesn't say that `setfsuid` should be avoided in general because obsolete.

I have a specific use case for this, that is I need to be able to check filesystem permissions per thread, and using `seteuid` would require to lock every thread with write access, not allowing parallel checks ([source](https://stackoverflow.com/a/1225487/995246)).

src/unistd.rs
```diff
@@ -1398,6 +1398,30 @@ pub fn setgid(gid: Gid) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Set the user identity used for filesystem checks.
+/// On both success and failure, this call returns the previous filesystem user
+/// ID of the caller.
+///
+/// See also [setfsuid(2)](http://man7.org/linux/man-pages/man2/setfsuid.2.html)
+#[cfg(target_os = "linux")]
```
asomers: I think we can enable this on Android, too, right?

src/unistd.rs
```diff
@@ -1398,6 +1398,30 @@ pub fn setgid(gid: Gid) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
+/// Set the user identity used for filesystem checks.
```
asomers: Perhaps add a "for the current thread" here?  I found it non-obvious that this is a per-thread variable.

src/unistd.rs
```diff
@@ -1398,6 +1398,30 @@ pub fn setgid(gid: Gid) -> Result<()> {
...
+/// On both success and failure, this call returns the previous filesystem user
+/// ID of the caller.
+///
+/// See also [setfsuid(2)](http://man7.org/linux/man-pages/man2/setfsuid.2.html)
+#[cfg(target_os = "linux")]
+#[inline]
```
asomers: No need for `#[inline]`

gliderkite:
> Ok, that sounds like a reasonable use case. Could you please add a test and a CHANGELOG entry?

I added a new entry in the changelog, but regarding the tests I'm not sure what tests would you like to see. I couldn't find any test for any equivalent function such as `seteuid` or `setegid` for example.

src/unistd.rs
gliderkite: I removed the attribute (fyi I added it just because it is used for all the other equivalent functions in the same module).

src/unistd.rs
gliderkite: Done.

src/unistd.rs
gliderkite: Added Android as a new target.

asomers:
> I added a new entry in the changelog, but regarding the tests I'm not sure what tests would you like to see. I couldn't find any test for any equivalent function such as `seteuid` or `setegid` for example.

Anything that minimally exercises those syscalls.  For example, creating a thread, setting its fs uid to nobody, and verifying that it gets EACCES when trying to read a 640 file.

gliderkite:
I added the test, but it fails on CI. At the moment I can only test it locally using `--target x86_64-unknown-linux-gnu`, which fails on CI, but it succeeds on my machine. I'll try to debug it by adding `--nocapture` to the CI tests when I have some time.

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
-- Add `CLK_TCK` to `SysconfVar`
+- Added `CLK_TCK` to `SysconfVar`
   (#[1177](https://github.com/nix-rust/nix/pull/1177))
+
+- Added `unistd::setfsuid` and `unistd::setfsgid` to set the user or group
```
asomers: Don't forget a PR link.

src/unistd.rs
```diff
@@ -2729,3 +2751,42 @@ impl Group {
...
+mod tests {
+    use super::*;
+    use std::{fs, thread, io};
+
+    /// Tests setting the filesystem UID with `setfsuid`.
+    #[test]
+    fn test_setfsuid() {
+        use std::os::unix::fs::PermissionsExt;
+
+        // get the UID of the "nobody" user
```
asomers: Per the man page, `setfsuid` requires the `CAP_SETUID` capability.  You should check for that here, like so:
```rust
require_capability!(CAP_SETUID);
```

src/unistd.rs
```diff
@@ -2729,3 +2751,42 @@ impl Group {
...
+        let file = tempfile::NamedTempFile::new().unwrap();
+        let temp_path = file.into_temp_path();
+        let temp_path_2 = (&temp_path).to_path_buf();
+        let mut permissions = fs::metadata(&temp_path).unwrap().permissions();
+        permissions.set_mode(640);
+
+        // spawn a new thread where to test setfsuid
+        thread::spawn(move || {
+            // set filesystem UID
+            let _ = setfsuid(nobody.uid);
```
asomers: To determine success, you should follow the procedure described in the man page.  Call `setfsuid(-1)` and check whether the fsuid was actually changed.  I suspect that when running on Travis, it won't be.

http://man7.org/linux/man-pages/man2/setfsuid.2.html

src/unistd.rs
gliderkite: Added it.

src/unistd.rs
gliderkite: I added an assertion for the fsuid change.

CHANGELOG.md
gliderkite: Done.

gliderkite:
It looks like it failed on nightly the second run (https://travis-ci.org/nix-rust/nix/jobs/656330684?utm_medium=notification&utm_source=github_status), where the first run passed and the only change only included the changelog. Could you please have a look @asomers?

asomers:
It looks like a new Rust nightly has added some lints.  I'll fix this in a separate PR, then you can rebase.

asomers:
Actually, it looks like the failures were due to a bug in the compiler.  I just restarted the build, it used a newer compiler version than the previous build (nightly-2020-02-28 vs nightly-2020-02-27), and it passed this time.

asomers:
Please squash your commits.  Then I'll merge.

asomers:
That failure happened deep within libstd.  Perhaps a QEMU bug?

bors retry

Issue #1164, opened 2019-12-22
There are no macros in sys::stat to determine file type from st_mode


Chronophagus:
Can't find any analogues to these macros: 
S_ISREG()
S_ISDIR()
S_ISCHR()
S_ISBLK()
S_ISFIFO()
S_ISLNK()
S_ISSOCK()

rrnewton:
This is not particularly clear in the documentation, but `S_IFMT` is the thing to bitwise AND with to get only the bits that specify the file type.  Those macros that are missing from the Rust wrapper are defined like:

```
    #define 	S_ISBLK(m)   (((m) & S_IFMT) == S_IFBLK)
```

PR #1165, opened 2019-12-22
Add S_ISREG, S_ISDIR,... macro analogues to stat

Chronophagus:
So I've just added them #1164


asomers:
Please rebase to fix the musl failure.

asomers:
CI is failing because you need to raise the minimum version of the bitflags crate.  It looks like you need version 1.2.0 to get the `from_bits_unchecked` function.

src/sys/stat.rs
```diff
@@ -123,6 +123,41 @@ pub fn fstatat<P: ?Sized + NixPath>(dirfd: RawFd, pathname: &P, f: AtFlags) -> R
     Ok(unsafe{dst.assume_init()})
 }
 
+pub fn s_isreg(st_mode: mode_t) -> bool {
+    let st_mode = unsafe { SFlag::from_bits_unchecked(st_mode) };
```
asomers: You can eliminate `unsafe` by doing:
```rust
(st_mode & SFlag::S_IFFMT.bits()) == SFlag::S_IFREG.bits())
```

Chronophagus:
> I'm not sure I like the interface. It's not very Rusty. It seems like a better API would be for the `stat` function to return a newtype, and for these functions to be methods on that newtype. Do you agree?

I completely agree! Originally I've implemented `FileType` enum with `is_dir()`, `is_sock()...` methods. I wanted to add `fn flie_type() -> FileType` method to the `FileStat` structure, but then I saw `FileStat` is simply the alias for libc::stat struct. I didn't want to add a function like `fn file_type(stat: &FileStat) -> FileType` because all free functions in stat.rs are part of stat C API. So I made the simplest patch as possible.

Chronophagus:
Pushed another API design there: https://github.com/nix-rust/nix/pull/1174 . But I really don't like this free `file_type` funciton.

asomers:
I think we can solve this problem more simply just by changing `FileStat` from a type alias to a newtype.  It won't be completely backwards-compatible, but Nix doesn't guarantee backwards compatibility between minor versions.  And it will be both more ergonomic,  more Rusty and more cross-platform.

Chronophagus:
Ok. If backwards-compatibility is not a concern I will make a new type.

asomers:
Yes, please do.

jabedude:
@Chronophagus any update on this?

PR #864, opened 2018-02-23
Add `getpwnam` and related functions

ctrlcctrlv:
This PR closes #862.

Implemented functions:
* As `User::query`:
  * getpwnam
  * getpwuid
* As `Group::query`:
  * getgrgid
  * getgrnam
* As `Users` iterator:
  * getpwent
  * setpwent
  * endpwent
* As `Groups` iterator:
  * getgrent
  * setgrent
  * endgrent

~I wanted to implement `getgrent`, `setgrent` and `endgrent`, but they are missing from `libc`. (See https://github.com/rust-lang/libc/issues/923 )~

PR #1173, opened 2019-12-24
Add `shadow.h` functions

ArniDagur:


asomers:
It looks like you need to set appropriate `#[cfg()]` guards around your new code.  Also, any particular reason you want to merge the shadow crate into Nix?

ArniDagur:
> Also, any particular reason you want to merge the shadow crate into Nix?

1. I think `nix` would be incomplete without the `shadow(3)` suite of functions. It's nice having only one crate that fulfils all of my "safe(r) libc"  needs.
2. It doesn't look like the `shadow` crate is maintained; it hasn't been touched in 3 years. The `nix` crate is well maintained (e.g. https://github.com/nix-rust/nix/pull/1114).

ArniDagur:
@asomers Some of the builds were failing on 32-bit systems because I assumed that `long int` was always 64 bits, which caused the implicit cast to fail. Are you happy with using `i64` as the integer type? Another possible option could be `usize`, but are we sure that the size of `usize` is always equal to or larger than `libc::c_long`?

asomers:
What's wrong with declaring the fields as `c_long`?

ArniDagur:
I thought we didn't want to use C types, but now that I think about it, this project does it all the time. I will change it to use `c_long`.

ArniDagur:
On another note, I tried GitHub's inbuilt "resolve merge conflict" functionality, and it seems to have merged master into this branch. Should I revert that change manually also?

asomers:
Yeah, you should manually rebase rather than use the "resolve conflicts" button.

ArniDagur:
I've rebased and squashed.

src/shadow.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// It is expressed as the number of days since Jan 1, 1970. The value 0
+    /// has a special meaning, which is that the user should change their
+    /// password the next time they will log in the system
+    pub last_change: libc::c_long,
+    /// Minimum password age
+    ///
+    /// The minimum password age is the number of days the user will have to
+    /// wait before she will be allowed to change their password again.
+    /// An empty field and value 0 mean that there are no minimum password age.
+    pub min: libc::c_long,
```
asomers: Instead of mysterious flag values, how about making this field a `NonZeroIsize`?

src/shadow.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// user cannot change their password.
+    pub max: libc::c_long,
+    /// Password warning period
+    ///
+    /// The number of days before a password is going to expire (see the
+    /// maximum password age above) during which the user should be warned.
+    ///
+    /// An empty field and value 0 mean that there are no password warning
+    /// period.
+    pub warn: libc::c_long,
```
asomers: An Option type would also be useful here.

src/shadow.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// since Jan 1, 1970.
+    ///
+    /// Note that an account expiration differs from a password expiration. In
+    /// case of an account expiration, the user shall not be allowed to login.
+    /// In case of a password expiration, the user is not allowed to login using
+    /// their password.
+    ///
+    /// The value 0 should not be used as it is interpreted as either an account
+    /// with no expiration, or as an expiration on Jan 1, 1970.
+    pub expire: libc::c_long,
```
asomers: Sounds like this could be a `NonZeroIsize`.

src/shadow.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// other means).
+    ///
+    /// This field may be empty, in which case no passwords are required to
+    /// authenticate as the specified login name. However, some applications
+    /// which read the /etc/shadow file may decide not to permit any access at
+    /// all if the password field is empty.
+    ///
+    /// A password field which starts with a exclamation mark means that the
+    /// password is locked.
+    pub password: CString,
```
asomers: Why use `String` for name, but `CString` for password?

src/shadow.rs
```diff
@@ -0,0 +1,174 @@
...
+/// println!("There are {} shadow entries", shadows.len());
+/// ```
+///
+/// # Safety
+/// Care should be taken when this iterator is used in different threads
+/// without synchronization. This is because the underlying functions used
+/// ([`getspent()`][1], and [`endspent()`][1]) are not thread safe.
+///
+/// [1]: http://man7.org/linux/man-pages/man3/shadow.3.html
+#[derive(Debug, Default)]
```
asomers: Shouldn't you call `setspent` during construction?

src/shadow.rs
```diff
@@ -0,0 +1,174 @@
...
+///
+/// # Examples
+/// ```
+/// # use nix::shadow::ShadowIterator;
+/// # use nix::shadow::Shadow;
+/// let shadows = ShadowIterator::default().collect::<Vec<Shadow>>();
+/// println!("There are {} shadow entries", shadows.len());
+/// ```
+///
+/// # Safety
```
asomers: What about using the reentrant versions `getspent_r` etc to solve the thread-safety problem?

src/shadow.rs
ArniDagur: Just to clarify, do you mean Option<NonZeroIsize>, where zero corresponds to `None`?

src/shadow.rs
ArniDagur: Because of what I read from @ctrlcctrlv here: https://github.com/nix-rust/nix/pull/864/files
> Because it's guaranteed that the fields I made `String` will conform to [`NAME_REGEX`](https://serverfault.com/a/73101/407341). `gecos` ("real name") has no such guarantee.

src/shadow.rs
ArniDagur: I'll do that.

ArniDagur: I'm currently blocked by https://github.com/rust-lang/libc/pull/1653

PR #1176, opened 2020-01-09
Add RISC-V support on GNU/Linux

msizanoen1:


asomers:
How does one build this?  Does rustup include a risc-v toolchain now?

msizanoen1:
No but libc and libstd are ported.
You can get a RISC-V cross toolchain and use xargo or cargo's `-Zbuild-std`.

msizanoen1:
This is the PR for building rust libstd on CI: https://github.com/rust-lang/rust/pull/68037
After it is merged you can install it from rustup.

asomers:
What exactly does this PR accomplish?  Is there more to it than those handfuls of constants?

msizanoen1:
That's everything needed to make it compile on RISC-V.

PR #1177, opened 2020-01-11
Add CLK_TCK to SysconfVar

cjbassi:
http://man7.org/linux/man-pages/man3/sysconf.3.html says that the
corresponding variable is obsolete, but I think that just means the
constant defined in limits.h is obsolete. Checking the value using
sysconf is still valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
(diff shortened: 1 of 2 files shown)


asomers:
I think you're correct.  Could you please add a CHANGELOG entry?

cjbassi:
Done.

asomers:
Good.  Please squash your commits, and I'll merge them.  I can't simply use Github's squash button, because we merge with bors.

cjbassi:
Just squashed.

Issue #678, opened 2017-07-15
Replace `RawFd` with `IntoRawFd` or `AsRawFd` as appropriate


Susurrus:
This is blocking on the resolution of rust-lang/rust#43254. Right now you can't have both `RawFd` and `T: AsRawFd` supported as inputs to the same function, so implementing this doesn't actually improve ergonomics. We may be able to create a newtype wrapper in `nix`, but I want to see if there might be a solution from the Rust devs here before we do our own newtype-wrapper based implementation.

Susurrus:
I think `unistd::close` is the only function that should use `IntoRawFd`, the others can just use `AsRawFd`, but this should be checked.

roblabla:
I am very much in favor of this. Wouldn't newtype-ing `RawFd` be a good idea anyway ? The fact that it's an int should be, IMO, hidden from the user's view, as a lot of operations on int don't yield interesting results on fds (what does RawFd + int do ?).

We could have our own newtype'd RawFd, and implement AsRawFd/IntoRawFd/FromRawFd ourselves. Is there a good reason not to do this (other than it being a breaking change) ?

lucab:
> a lot of operations on int don't yield interesting results on fds (what does RawFd + int do?)

Integer arithmetic on fd is quite common with IPC and socket passing. If you want a specific concrete usecase, see https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html.

In general, I'm in favor of having a more sound `RawFd` replacement (see https://github.com/nix-rust/nix/issues/594) but I still need to do integer arithmetic around it.

roblabla:
Well, you could still use `AsRawFd` to get the raw integer, and FromRawFd (which is unsafe) to create the new Fd.

Kixunil:
@lucab such operation is basically `unsafe`.

kahing:
What about operations that return RawFd? For example, dup2() returns RawFd, how would one pass that to another function if everything takes AsRawFd?

Kixunil:
@kahing The interface should be changed to return `Fd`, just as I did it in case of `Pid`, `Uid`, and `Gid`.

asomers:
I agree that we should use a Newtype.  But we should name it `FileDesc`, because that's what libstd uses.  The libstd type isn't public now.  But if it ever becomes public in the future, it would be easy for us to delete our own implementation and for our users to change `nix::FileDesc` into `FileDesc`.

Susurrus:
[...]

PR #1184, opened 2020-02-08
Safe fd wrappers

### LINKED ISSUE #594: Taking RawFd by value is not sound

Chronophagus:
Here I'm trying to resolve #594 #678. Wish me luck!

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -9,7 +9,7 @@ use libc::{self, c_char, c_void, c_int, c_long, c_uint, size_t, pid_t, off_t,
 use std::{fmt, mem, ptr};
 use std::ffi::{CString, CStr, OsString, OsStr};
 use std::os::unix::ffi::{OsStringExt, OsStrExt};
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{FromRawFd, IntoRawFd, AsRawFd, RawFd};
 use std::path::PathBuf;
 use void::Void;
 use sys::stat::Mode;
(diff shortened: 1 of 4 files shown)


src/unistd.rs
```diff
@@ -21,6 +21,143 @@ pub use self::pivot_root::*;
...
+
+    pub fn stderr() -> Self {
+        let stderr = unsafe { StdHandle::from_raw_fd(libc::STDERR_FILENO) };
+        stderr
+    }
+}
+
+impl FromRawFd for StdHandle {
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        assert!(fd < 3);
```
asomers: -1 to magic numbers.  Instead of hard-coding 3, you should compare to the `libc::STDERR_FILENO` etc

PR #1185, opened 2020-02-11
Provide clearenv()

### LINKED ISSUE #481: Provide clearenv?
Can clearenv() be supported, with a workaround of environ = NULL where it's not supported?

jgallagher:
Maybe closes #481.

"Maybe" for a couple reasons that are discussed on that issue:

* The issue mentions adding the other environment-variable-related functions (`setenv`, `getenv`, etc.), but I omitted those because they're already covered by `std::env` AFAICT, plus they're wildly unthreadsafe. It looks like `std::env` avoids the thread unsafety via a library-level lock and a warning to be careful when using ffi functions that might modify the environment concurrently.
* I opted to return `Error::UnsupportedOperation` in the (impossible?) case of `libc::clearenv()` returning non-zero instead of having the function return a `Result<(), ()>`. The latter seems like it would be onerous (since every other possibly-failing function in nix returns a `nix::Result`), but if you'd prefer a different error return I'm not going to put up any argument.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ exclude     = [
 [dependencies]
 libc = { version = "0.2.60", features = [ "extra_traits" ] }
 bitflags = "1.1"
-cfg-if = "0.1.2"
+cfg-if = "0.1.10"
 void = "1.0.2"
 
 [target.'cfg(target_os = "dragonfly")'.build-dependencies]
@@ -43,6 +43,10 @@ path = "test/test.rs"
 [[test]]
 name = "test-aio-drop"
 path = "test/sys/test_aio_drop.rs"
+
+[[test]]
+name = "test-clearenv"
+path = "test/test_clearenv.rs"
 
 [[test]]
 name = "test-lio-listio-resubmit"
--- a/src/env.rs
+++ b/src/env.rs
@@ -1,0 +1,52 @@
+use {Error, Result};
+
+/// Clear the environment of all name-value pairs.
+///
+/// On platforms where libc provides `clearenv()`, it will be used. libc's
+/// `clearenv()` is documented to return an error code but not set errno; if the
+/// return value indicates a failure, this function will return
+/// `Error::UnsupportedOperation`.
+///
+/// On platforms where libc does not provide `clearenv()`, a fallback
+/// implementation will be used that iterates over all environment variables and
+/// removes them one-by-one.
+///
+/// # Safety
+///
+/// This function is not threadsafe and can cause undefined behavior in
+/// combination with `std::env` or other program components that access the
+/// environment. See, for example, the discussion on `std::env::remove_var`; this
+/// function is a case of an "inherently unsafe non-threadsafe API" dealing with
+/// the environment.
+///
+///  The caller must ensure no other threads access the process environment while
+///  this function executes and that no raw pointers to an element of libc's
+///  `environ` is currently held. The latter is not an issue if the only other
+///  environment access in the program is via `std::env`, but the requirement on
+///  thread safety must still be upheld.
+pub unsafe fn clearenv() -> Result<()> {
+    let ret;
+    cfg_if! {
+        if #[cfg(any(target_os = "fuchsia",
+                     target_os = "wasi",
+                     target_env = "wasi",
+                     target_env = "uclibc",
+                     target_os = "linux",
+                     target_os = "android",
+                     target_os = "emscripten"))] {
+            ret = libc::clearenv();
+        } else {
+            use std::env;
+            for (name, _) in env::vars_os() {
+                env::remove_var(name);
+            }
+            ret = 0;
+        }
+    }
+
+    if ret == 0 {
+        Ok(())
+    } else {
+        Err(Error::UnsupportedOperation)
+    }
+}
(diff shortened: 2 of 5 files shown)


src/env.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+///  The caller must ensure no other threads access the process environment while
+///  this function executes and that no raw pointers to an element of libc's
+///  `environ` is currently held. The latter is not an issue if the only other
+///  environment access in the program is via `std::env`, but the requirement on
+///  thread safety must still be upheld.
+pub unsafe fn clearenv() -> Result<()> {
+    let ret;
+    cfg_if! {
+        if #[cfg(target_os = "fuchsia")] {
```
asomers: You should combine these three `#[cfg]` expressions.

src/env.rs
```diff
@@ -0,0 +1,57 @@
...
+            ret = libc::clearenv();
+        } else if #[cfg(any(target_os = "wasi", target_env = "wasi"))] {
+            ret = libc::clearenv();
+        } else if #[cfg(all(unix, any(target_env = "uclibc",
```
asomers: Checking for `unix` is redundant since those targets are always unix.

src/env.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    if ret == 0 {
+        Ok(())
+    } else {
+        Err(Error::UnsupportedOperation)
+    }
+}
+
+fn manual_clearenv() {
+    for (name, _) in env::vars() {
```
asomers: This entails converting from OsString to String and back again.  Faster if you use `env::vars_os` instead.

src/env.rs
```diff
@@ -0,0 +1,57 @@
...
+/// function is a case of an "inherently unsafe non-threadsafe API" dealing with
+/// the environment.
+///
+///  The caller must ensure no other threads access the process environment while
+///  this function executes and that no raw pointers to an element of libc's
+///  `environ` is currently held. The latter is not an issue if the only other
+///  environment access in the program is via `std::env`, but the requirement on
+///  thread safety must still be upheld.
+pub unsafe fn clearenv() -> Result<()> {
+    let ret;
```
asomers: As a matter of style, I suggest:
```rust
let ret = cfg_if! {
    if something {
        libc::clearenv()
    } else {
        0
    }
};
```

src/env.rs
jgallagher: Ah, hah, I was scraping through `libc`'s breakdown of where it included `clearenv` and lost the forest for the trees. All of these can be combined into one `#[cfg(any(...)]`, right?

src/env.rs
jgallagher: 👍

src/env.rs
jgallagher: I tried that first, but it looks like that's not allowed with `cfg_if!`. It generates a lot of compilation errors; the most relevant looks like:

```
error: macro expansion ignores token `$crate` and any following
  --> <::cfg_if::cfg_if macros>:23:37
   |
23 |       { @ __identity $ ($ tokens) * } $ crate :: cfg_if !
   |                                       ^^^^^^^
   |
  ::: src/env.rs:28:15
   |
28 |       let ret = cfg_if! {
   |  _______________-
29 | |         if #[cfg(any(target_os = "fuchsia",
30 | |                      target_os = "wasi",
31 | |                      target_env = "wasi",
...  |
43 | |         }
44 | |     };
   | |     -
   | |     |
   | |_____caused by the macro expansion here
   |       help: you might be missing a semicolon here: `;`
   |
   = note: the usage of `cfg_if!` is likely invalid in expression context
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

asomers:
The build failed on nightly.  Could you please investigate, @jgallagher ?

jgallagher:
@asomers Squashed down to one commit. I'm not sure what to make of the nightly failure; I can't reproduce it locally, and the error message itself is pretty bizarre.

Do you know off hand if there's a description of the travis environment somewhere? I could try spinning up a VM that more closely matches it.

asomers:
That test was run on the latest nightly.  It may well be a bug in Rust.  Could you try updating your nightly toolchain and reproducing?

jgallagher:
I had already tried updating nightly without any luck, but I was just able to reproduce it. I hadn't noticed the `cargo update -Zminimal-versions` in the travis log, but if I do that first, I get the same error. Bumping `cfg-if` in `Cargo.toml` to `0.1.10` fixes the error (the error is still there on `0.1.9`).

Do you want me to squash that bump into this PR?

asomers:
Yes please!  Good sleuthing.

jgallagher:
Looks like that did the trick. 👍

Cargo.toml
```diff
@@ -18,7 +18,7 @@ exclude     = [
 [dependencies]
 libc = { version = "0.2.60", features = [ "extra_traits" ] }
 bitflags = "1.1"
-cfg-if = "0.1.2"
+cfg-if = "0.1.10"
```
asomers: Did you actually verify that 0.1.10 is the minimum version that works?

jgallagher: I think I mentioned before that the error is still there on 0.1.9. I didn't try every version in between 0.1.2 and 0.1.9; it seems unlikely any of those work work if the same error is in both 0.1.2 and 0.1.9, but I guess it's not impossible. Is there an impact to jumping to 0.1.10?

asomers: Well, given that 0.1.10 comes immediately after 0.1.9, I'd say that confirms it.  Too bad that the cfg-if crate doesn't maintain a CHANGELOG.

PR #1187, opened 2020-02-12
Make FsType's raw type public

MikailBag:
# Motivation
Currently, `FsType` API is limited. In fact, only operation is supports is comparison with one of well-known IDs.
But:
* This list is incomplete. For example, it misses cgroupfs and cgroup2fs.
* This approach doesn't work with custom FSs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   where it's available, and clears the environment of all variables
   via `std::env::vars` and `std::env::remove_var` on others.
   (#[1185](https://github.com/nix-rust/nix/pull/1185))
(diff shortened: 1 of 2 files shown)


asomers:
That's certainly a real problem.  I think this solution is fair.  But could you please add a CHANGELOG entry?  Also, we would appreciate PRs that add any missing values.

MikailBag:
Thanks for review!
I added CHANGELOG entry.
I can't submit PR for `CGROUP_SUPER_MAGIC` and `CGROUP2_SUPER_MAGIC` until [libc PR](https://github.com/rust-lang/libc/pull/1660) is merged.

asomers:
Sorry, but you'll need to rebase.

MikailBag:
Done

PR #1189, opened 2020-02-29
Derive `Ord`, `PartialOrd` for `Pid`

### LINKED ISSUE #1188: Derive `Ord`, `PartialOrd` for `Pid`
This would enable, for example, keying `BTree{Map,Set}` on a `Pid` without having to invoke `{as,from}_raw()`, or adding a wrapper around `Pid` itself.

It looks like these traits are currently missing on all platforms.

I'm running:
- **Platform:** x86-64, Linux 5.4
- **rustc:** 1.41.0

ranweiler:
Closes #1188.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `unistd::setfsuid` and `unistd::setfsgid` to set the user or group
   identity for filesystem checks per-thread.
   (#[1163](https://github.com/nix-rust/nix/pull/1163))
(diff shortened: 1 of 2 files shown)


posborne:
Bah, bors r+

ranweiler:
@posborne, would you like me to add a changelog entry with this PR, or do maintainers prefer to do that separately post-merge?

posborne:
@ranweiler Yes, good call.  We do prefer them to be a part of the PR.  Looks like things didn't go through before so go ahead and add now.

ranweiler:
@posborne added!

posborne:
Thanks @ranweiler, looks good.  Can you squash down to a single commit and then I'll kick off the merge.

ranweiler:
@posborne done.

Issue #1192, opened 2020-03-09
Does nix support mount for MacOS?


pwang7:
I'm trying nix to make some syscalls, instead of using libc.
But it seems nix has no mount for MacOS, am I right?

asomers:
Yep.  Though it probably wouldn't be hard to extend it to work on OSX, too.

pwang7:
@asomers could you please advise me how to extend mount to work on OSX?

asomers:
Uh, not really.  I don't have a Mac and I haven't used one since 10.3.  You'll have to figure out how mount works on that platform.

pwang7:
Let me explain what I found first.

It seems `mount(2)` differs a lot w.r.t. Linux and OSX.
Linux: `int mount(const char *source, const char *target, const char *filesystemtype, unsigned long mountflags, const void *data);` 5 input arguments;
OSX: `int mount(const char *type, const char *dir, int flags, void *data);` 4 input arguments.

The trick part is the last input argument of `mount(2)`, `void *data`, which has different meaning w.r.t. Linux and OSX. According to Linux man page, the `data` argument is a c-string of comma-separated options understood by this filesystem. Whereas, under OSX, the `data` argument is a pointer to a structure that contains the type specific arguments to mount, and the format for these argument structures is different w.r.t. each filesystem.

Another thing is the mount flags are different w.r.t. Linux and OSX.

I used `libc::mount()` under OSX to successfully mount a Filesystem in User Space (FUSE), which requires a struct as following:
```
struct fuse_mount_args {
    char     mntpath[MAXPATHLEN];        // path to the mount point
    char     fsname[MAXPATHLEN];         // file system description string
    char     fstypename[MFSTYPENAMELEN]; // file system type name
    char     volname[MAXPATHLEN];        // volume name
    uint64_t altflags;                   // see mount-time flags below
    uint32_t blocksize;                  // fictitious block size of our "storage"
    uint32_t daemon_timeout;             // timeout in seconds for upcalls to daemon
    uint32_t fsid;                       // optional custom value for part of fsid[0]
    uint32_t fssubtype;                  // file system sub type id
    uint32_t iosize;                     // maximum size for reading or writing
    uint32_t random;                     // random "secret" from device
    uint32_t rdev;                       // dev_t for the /dev/osxfuse{n} in question
};
```
Whereas it seems the struct to mount Mac HFS under OSX is something else, which I haven't tried yet:
```
struct hfs_mount_args {
[...]

PR #1193, opened 2020-03-14
add macOS mount flags

pwang7:
#1192 add macOS mount flags according to
https://github.com/apple/darwin-xnu/blob/master/bsd/sys/mount.h#L288


src/mount.rs
```diff
@@ -48,6 +48,59 @@ libc_bitflags!(
 
 libc_bitflags!(
     pub struct MntFlags: c_int {
+        /// Read only filesystem
+        MNT_RDONLY;
```
asomers: You need to `#[cfg(...)]` gate each of these according to which OSes have it.

pwang7: Added macOS target.
Should I add more OS targets?

asomers: Yes, it would be great to add other targets.  I think most of these flags are also supported by one or more of the BSDs, for example.  Also, please don't force push until I ask you to squash your commits.  It makes incremental review harder.

pwang7: OK, thanks for your reminder. I added more BSD-like targets.

pwang7: @asomers any comments?

asomers:
You can remove bitrig.  It's dead, and even libc doesn't support it any more.  Also, you should add a note in CHANGELOG.

src/mount.rs
asomers: Just remove bitrig and add a CHANGELOG entry.

PR #1195, opened 2020-03-17
Implement open file descriptor locks in fcntl

andrenth:
Hello

This PR updates libc to 0.2.68, which adds the `F_OFD_*` fcntl commands, and uses them in `nix::fcntl::fcntl`.

### DIFF
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -211,7 +211,14 @@ mod linux_android {
     }
 
     #[test]
-    #[cfg(not(target_arch = "mips"))]
+    #[cfg(not(any(target_arch = "aarch64",
+                  target_arch = "arm",
+                  target_arch = "armv7",
+                  target_arch = "i686",
+                  target_arch = "mips",
+                  target_arch = "mips64",
+                  target_arch = "mips64el",
+                  target_arch = "powerpc64")))]
     fn test_ofd_locks() {
         let tmp = NamedTempFile::new().unwrap();
 
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -207,7 +207,7 @@ mod linux_android {
     #[cfg(not(any(target_arch = "aarch64",
                   target_arch = "arm",
                   target_arch = "armv7",
-                  target_arch = "i686",
+                  target_arch = "x86",
                   target_arch = "mips",
                   target_arch = "mips64",
                   target_arch = "mips64el",
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -205,7 +205,7 @@ mod linux_android {
         let tmp = NamedTempFile::new().unwrap();
 
         let fd = tmp.as_raw_fd();
-        let inode = fstat(fd).unwrap().st_ino;
+        let inode = fstat(fd).unwrap().st_ino as usize;
 
         let mut flock = libc::flock {
             l_type: libc::F_WRLCK as libc::c_short,
@@ -230,7 +230,7 @@ mod linux_android {
         assert_eq!(None, lock_info(inode));
     }
 
-    fn lock_info(inode: u64) -> Option<(String, String)> {
+    fn lock_info(inode: usize) -> Option<(String, String)> {
         let file = File::open("/proc/locks").unwrap();
         let buf = BufReader::new(file);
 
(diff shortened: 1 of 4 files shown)


andrenth:
I've added the CHANGELOG entry. I'll try to come up with a test. Does the CI run on Linux? I see only FreeBSD mentioned, and `O_OFD_*` is Linux-specific.

asomers:
CI runs on quite a few platforms.  Cirrus does FreeBSD, and Travis does Linux, OSX, NetBSD (build only), and Android (build only)

andrenth:
So, I've added `#[cfg(not(target_arch = "mips"))]` to the test because I can't seem to instantiate a `libc::flock` on mips, due to the `pad` field being private.

The failing tests in some architectures are caused by fcntl returning `EINVAL`, which means the `F_OFD_SETLKW` command is not recognized. This seems to be because of old kernel/libc versions in the test environments; for example, the tests do pass on a Debian Buster 32-bit VM.

Do you think it would be fine for now to simply not run this test for aarch64, arm, armv7, i686, mips64, mips64el and powerpc64?

andrenth:
Travis now only fails for i686. I don't know why that test is running despite the `#[cfg(not(any(..., target_arch = "i686", ...)))]`.

test/test_fcntl.rs
```diff
@@ -197,22 +209,87 @@ mod linux_android {
         let mut buf = [0u8; 200];
         assert_eq!(100, read(fd, &mut buf).unwrap());
     }
+
+    #[test]
+    #[cfg(not(any(target_arch = "aarch64",
+                  target_arch = "arm",
+                  target_arch = "armv7",
+                  target_arch = "i686",
```
asomers: Try "x86" instead of "i686".  Also, please explain in a comment why the test is disabled  on certain architectures.

test/test_fcntl.rs
```diff
@@ -1,26 +1,27 @@
...
     tmp.write_all(CONTENTS).unwrap();
 
-    let dirfd = open(tmp.path().parent().unwrap(),
-                     OFlag::empty(),
-                     Mode::empty()).unwrap();
-    let fd = openat(dirfd,
-                    tmp.path().file_name().unwrap(),
-                    OFlag::O_RDONLY,
-                    Mode::empty()).unwrap();
+    let dirfd = open(tmp.path().parent().unwrap(), OFlag::empty(), Mode::empty()).unwrap();
```
asomers: It looks like you made some formatting changes unrelated to this PR.  Could you please back those out?

test/test_fcntl.rs
andrenth: I've added a comment and switched to x86.

test/test_fcntl.rs
andrenth: Sorry, forgot to disable rustfmt while editing. It's fixed.

test/test_fcntl.rs
```diff
@@ -197,6 +199,72 @@ mod linux_android {
         let mut buf = [0u8; 200];
         assert_eq!(100, read(fd, &mut buf).unwrap());
     }
+
+    // This test is disabled for the target architectures below
+    // due to OFD locks not being available in the kernel/libc
+    // versions used in the CI environment.
```
asomers, line 205: It's probably related to the fact that CI for these platforms runs under QEMU.

test/test_fcntl.rs
```diff
@@ -197,6 +199,72 @@ mod linux_android {
...
+        assert_eq!(
+            Some(("OFDLCK".to_string(), "WRITE".to_string())),
+            lock_info(inode)
+        );
+
+        flock.l_type = libc::F_UNLCK as libc::c_short;
+        fcntl(fd, FcntlArg::F_OFD_SETLKW(&flock)).expect("write unlock failed");
+        assert_eq!(None, lock_info(inode));
+
+        flock.l_type = libc::F_RDLCK as libc::c_short;
```
asomers, line 239: It looks like this test is doing two separate things: setting and clearing a write lock, and setting and clearing a read lock.  Please split it up into two separate tests.  That will make it easier to debug.

test/test_fcntl.rs
andrenth: I've amended the comment.

test/test_fcntl.rs
andrenth: Done.

Issue #1153, opened 2019-11-14
Iterator over all available waitpid events


KillingSpark:
I am using the wait module to check after the child processes after receiving a SIGCHLD signal. Naturally I want to collect all available events at once as SIGCHLD might be sent only once for multiple child process events.

I am using an iterator for that which might be useful for other users. It would be nice to see this integrated into the modules API

```
// this would be a cool addition to the API
pub fn event_iter() -> std::iter::FromFn<fn() -> Option<Result<(i32, i8), nix::Error>> {
    let event_iter = std::iter::from_fn(get_next_exited_child)
}

// I use this function to make my iterator
fn get_next_exited_child() -> Option<Result<(i32, i8), nix::Error>> {
    match nix::sys::wait::waitpid(-1, Some(nix::sys::wait::WNOHANG)) {
        Ok(exit_status) => match exit_status {
            nix::sys::wait::WaitStatus::Exited(pid, code) => Some(Ok((pid, code))),
            nix::sys::wait::WaitStatus::Signaled(pid, signal, dumped_core) => {
                // signals get handed to the parent if the child got killed by it but didnt handle the
                // signal itself
                if signal == libc::SIGTERM {
                    if dumped_core {
                        Some(Ok((pid, signal as i8)))
                    } else {
                        Some(Ok((pid, signal as i8)))
                    }
                } else {
                    None
                }
            }
            nix::sys::wait::WaitStatus::StillAlive => {
                trace!("No more state changes to poll");
                None
            }
            _ => {
                trace!("Child signaled with code: {:?}", exit_status);
                None
            }
        },
        Err(e) => {
            if let nix::Error::Sys(nix::errno::ECHILD) = e {
            } else {
                trace!("Error while waiting: {}", e.description().to_owned());
            }
            Some(Err(e))
        }
    }
}
```

asomers:
Care to turn it into a PR?

KillingSpark:
Sure. I just wanted to see if you have any interest in this conceptually

PR #1196, opened 2020-03-19
Add an iterator over all events pollable with waitpid(-1, WNOHANG)

KillingSpark:
PR for the issue #1153. It basically provides an iterator over all currently available child events. This makes handling of the SIGCHLD signal easier, since it may be sent for multiple events and not just one.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -1,8 +1,8 @@
-use libc::{self, c_int};
-use Result;
 use errno::Errno;
+use libc::{self, c_int};
 use std::convert::TryFrom;
 use unistd::Pid;
+use Result;
 
 use sys::signal::Signal;
 
@@ -105,8 +105,7 @@ impl WaitStatus {
     pub fn pid(&self) -> Option<Pid> {
         use self::WaitStatus::*;
         match *self {
-            Exited(p, _)  | Signaled(p, _, _) |
-                Stopped(p, _) | Continued(p) => Some(p),
+            Exited(p, _) | Signaled(p, _, _) | Stopped(p, _) | Continued(p) => Some(p),
             StillAlive => None,
             #[cfg(any(target_os = "android", target_os = "linux"))]
             PtraceEvent(p, _, _) | PtraceSyscall(p) => Some(p),
@@ -211,6 +210,7 @@ impl WaitStatus {
     }
 }
 
+/// Wait for an event of a child process, either for a specific PID, or all children by passing None as pid 
 pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Result<WaitStatus> {
     use self::WaitStatus::*;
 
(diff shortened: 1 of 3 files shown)


src/sys/wait.rs
```diff
@@ -235,6 +235,57 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
     }
 }
 
+
+/// The same as calling waitpid(None, None)
```
asomers: Thanks for adding docs where they were missing.  Do you think you could turn this one into an internal link?  It would look something like [`waitpid(None, None)`](fn.waitpid.html)`.

src/sys/wait.rs
```diff
@@ -235,6 +235,57 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
...
 pub fn wait() -> Result<WaitStatus> {
     waitpid(None, None)
 }
+
+
+/// Returns an iterator over all currently available events of child processes.
+/// 
+/// This iterator is best used with a signal handler on the signal SIGCHLD
```
asomers: Stuff like `SIGCHLD` should be in backticks.

src/sys/wait.rs
```diff
@@ -235,6 +235,57 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
...
+                // No more events to be waited on
+                None
+            } else {
+                // Some other error, which we cannot handle here
+                //
+                // This might be unecessary, since the other errors are either
+                // 1. EINTR which only occurs if WNOHANG is not used, but it is used here
+                // 2. EINVAL which should be impossible since only valid WaitPidFlags can be created
+                //
+                // Not having this error handling would make the API nicer, but it could introduce breakage on some
```
asomers: It's a good idea to have the error handling.  Plenty of syscalls occasionally return undocumented errors.

src/sys/wait.rs
```diff
@@ -235,6 +235,57 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
...
 }
+
+
+/// Returns an iterator over all currently available events of child processes.
+/// 
+/// This iterator is best used with a signal handler on the signal SIGCHLD
+/// (Note that this example uses the signal_hook crate. This is just for clarity, it would of course work with any other mechanism to
+/// receive signals)
+/// ```ignore
+/// for signal in signals.forever() {
```
asomers: The example isn't very illustrative since it relies on an external crate.  But nix includes signal handling functionality.  Could you rework the example to eliminate the external crate?

KillingSpark: This is a bit tough, since handling signals is a bit tricky if there is actual work to be done. i think this is too elaborated to have as an example. I removed the external crate but I added some more text to point this out to users.

Does this count as the doc test you mentioned above? :)

asomers: Well, no, because it doesn't do anything.  It doesn't create any child processes, and never gets SIGCHLD.  What I would do if I were you would be to set a signal handler, then fork the child processes, then poll until the signal is received.  That's the simplest way to handle signals.  For an example, see `test_write_sigev_signal` in tests/sys/test_aio.rs .

KillingSpark: I went a bit of another route for the test since the signal is not actually needed and makes the test more complex than I think it needs to be. I hope this is acceptable regardless. If you think that signal handling should be part of this I will add it of course.

src/sys/wait.rs
```diff
@@ -235,6 +235,59 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
     }
 }
 
+/// The same as calling [waitpid](fn.waitpid.html)(None, None)
 pub fn wait() -> Result<WaitStatus> {
     waitpid(None, None)
 }
+
+/// Returns an iterator over all currently available events of child processes, as would be available via waitpid with the
```
asomers, line 243: Wrap to 80 cols, please

src/sys/wait.rs
```diff
@@ -240,19 +245,27 @@ pub fn wait() -> Result<WaitStatus> {
...
+/// This iterator is best used with a signal handler on the signal `SIGCHLD`
+/// to handle all pending events. The iterator ends either
+/// if all remaining children are still running normally or no more children
+/// are running.
+///
+/// Note that you should not call this directly in a signal handler, since
+/// those have some restrictions on what may
+/// happen in them. Use a conditional variable to notify some other routine
+/// or use a crate like [signal_hook](https://crates.io/crates/signal-hook)
 /// ```
```
asomers, line 261: I think you need to put this in an `# EXAMPLES` section so rustdoc will render it appropriately.  You'll also want to add `no_run` so it won't hang during `cargo test`

test/sys/test_wait.rs
```diff
@@ -56,6 +56,61 @@ fn test_waitstatus_pid() {
...
+    
+    // this runs until all children have been signaled. It might have to call the iterator a few times, because the events might
+    // be delivered in chunks
+    while exited_children.len() < children.len() {
+        // this prevents the test from running for ever in case of a bug 
+        assert!(
+            start_time.elapsed() < std::time::Duration::from_secs(2), 
+            "It takes the children way too long to exit, something is probably broken"
+        );
+        exited_children.extend(child_event_iter());
```
asomers, line 94: If this is a polling loop, shouldn't it sleep somewhere?  Also, the comment on line 86 is >> 80 cols.

src/sys/wait.rs
KillingSpark: It should not hang, because it exits immediately if there are no events available but yeah it should not run, just compile. I'll add this.

test/sys/test_wait.rs
KillingSpark: It is not really necessary to sleep here, the children exit immediately and are collected yb the loop. Sleeping would in most cases increase the time this tests needs.

At worst (when a bug occurs) this will occupy one cpu core for 2 seconds.

PR #1198, opened 2020-03-19
Enable getregs/setregs on x86_64 musl

coord-e:
Hi, I needed `ptrace::getregs` in my project. I noticed that they're only enabled on glibc while they can be compiled with musl. This PR enables getregs/setregs on x86_64 musl libc environment.


asomers:
This looks good; now please squash your commits.  The CI failure I think is caused by a bug in glibc.

coord-e:
Squashed and pushed. Thanks

asomers:
Sorry, but you'll have to rebase.  Some other PR just made a conflict.

coord-e:
That's OK, I've rebased

PR #1201, opened 2020-03-28
Fix fallocate signature

### LINKED ISSUE #1200: fallocate shouldn't return c_int
`fallocate`, like `posix_fallocate`, doesn't have a meaningful return value.

From the man page

> On success, fallocate() returns zero.  On error, -1 is returned and errno is set to indicate the error.

DianaNites:
Fixes #1200


DianaNites:
> follow existing convention in the CHANGELOG.

Oh sorry, didn't notice. That convention should probably be documented in CONVENTIONS.md?

PR #1205, opened 2020-04-06
unistd: avoid infinite loop caused by reserve_double_buffer_size

stevendanna:
Functions such as Group::from_anything use reserve_double_buffer_size
in a loop, expecting it to return ERANGE if the passed limit is
reached.

However, the returned vector is passed as pointer to a libc function
that writes data into memory and doesn't update the length of the
Vec. Because of this, the previous code would never return ERANGE and
the calling loops would never exit if they hit a case where the
required buffer was larger than the maximum buffer.

This fixes the problem by checking the capacity rather than the
length.

Signed-off-by: Steven Danna <steve@chef.io>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (#[1198](https://github.com/nix-rust/nix/pull/1198))
 
 ### Fixed
+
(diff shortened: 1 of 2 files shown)


asomers:
Oh!  Thanks for spotting this.  Could you please add a CHANGELOG entry?

PR #1206, opened 2020-04-07
Fix unaligned casting of cmsg data to af_alg_iv

glebpom:
Casting a pointer to `cmsg_data` to `af_alg_iv` is incorrect since it's not properly aligned. As of the [`cmsg` man page](http://man7.org/linux/man-pages/man3/cmsg.3.html) "Applications should not cast it to a pointer type matching the payload, but should instead use memcpy(3) to copy data to or from a suitably declared object."


asomers:
Thanks for the fix.  Might I ask how you found the problem?  Also, could you please add a CHANGELOG entry?

glebpom:
Sure, that was pretty easy with clippy. BTW, few other things are treated as errors by clippy as well. It would make sense to add clippy check as a CI step.

asomers:
bors retry

PR #1207, opened 2020-04-08
Add select::FdSet::fds() method

zombiezen:
To be more consistent with most Rust APIs and enable cloning of the iterator, I made `FdSet::contains` operate on an immutable borrow instead of a mutable one by copying the set. If this is not desirable, I can roll that back from this PR and focus purely on the `fds()` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Changed `fallocate` return type from `c_int` to `()` (#[1201](https://github.com/nix-rust/nix/pull/1201))
 - Enabled `sys::ptrace::setregs` and `sys::ptrace::getregs` on x86_64-unknown-linux-musl target
   (#[1198](https://github.com/nix-rust/nix/pull/1198))
-- `select::FdSet::contains` and `select::FdSet::highest` now operate on `&FdSet`
-  instead of `&mut FdSet`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   identity for filesystem checks per-thread.
   (#[1163](https://github.com/nix-rust/nix/pull/1163))
 - Derived `Ord`, `PartialOrd` for `unistd::Pid` (#[1189](https://github.com/nix-rust/nix/pull/1189))
+- Added `select::FdSet::fds` method to iterate over file descriptors in a set.
 
 ### Changed
 - Changed `fallocate` return type from `c_int` to `()` (#[1201](https://github.com/nix-rust/nix/pull/1201))
 - Enabled `sys::ptrace::setregs` and `sys::ptrace::getregs` on x86_64-unknown-linux-musl target
   (#[1198](https://github.com/nix-rust/nix/pull/1198))
+- `select::FdSet::contains` and `select::FdSet::highest` now operate on `&FdSet`
+  instead of `&mut FdSet`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (#[1163](https://github.com/nix-rust/nix/pull/1163))
 - Derived `Ord`, `PartialOrd` for `unistd::Pid` (#[1189](https://github.com/nix-rust/nix/pull/1189))
 - Added `select::FdSet::fds` method to iterate over file descriptors in a set.
+  (#[#1207](https://github.com/nix-rust/nix/pull/1207))
 
 ### Changed
 - Changed `fallocate` return type from `c_int` to `()` (#[1201](https://github.com/nix-rust/nix/pull/1201))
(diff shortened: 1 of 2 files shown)


src/sys/select.rs
```diff
@@ -30,8 +32,9 @@ impl FdSet {
         unsafe { libc::FD_CLR(fd, &mut self.0) };
     }
 
-    pub fn contains(&mut self, fd: RawFd) -> bool {
-        unsafe { libc::FD_ISSET(fd, &mut self.0) }
+    pub fn contains(&self, fd: RawFd) -> bool {
+        let mut copy = self.0;
```
asomers, line 36: `libc::FD_ISSET` should be taking a `*const` pointer rather than `*mut`.  I just opened a PR to fix that.
https://github.com/rust-lang/libc/pull/1725

src/sys/select.rs
```diff
@@ -58,15 +61,32 @@ impl FdSet {
...
-        None
+    /// Returns an iterator over the file descriptors in the set.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # extern crate nix;
+    /// # use nix::sys::select::FdSet;
+    /// # use std::os::unix::io::RawFd;
+    /// # fn main() {
```
asomers, line 76: Usually you don't need the `fn main` line.  Do you really need it here?

zombiezen: You're right. Removed.

src/sys/select.rs
zombiezen: As an aside, if we're worried about the cost of this copy, then we might want to consider not deriving `Copy` for `FdSet`.

asomers: You know, that's an excellent point.  On my system, `FdSet` is 128 bytes big.  Users probably shouldn't copy such a time unless they really need to. _However_ removing `Copy` doesn't necessarily prevent the value from ever being `memcpy`'d, and the consensus is that it's best to impl `Copy` regardless of size.  See https://www.reddit.com/r/rust/comments/2xxjda/when_should_my_type_be_copy/ and https://internals.rust-lang.org/t/pre-rfc-copy-like-trait-for-large-types/11852/14 for some discussion.

src/sys/select.rs
```diff
@@ -58,15 +61,30 @@ impl FdSet {
...
+    /// # use nix::sys::select::FdSet;
+    /// # use std::os::unix::io::RawFd;
+    /// let mut set = FdSet::new();
+    /// set.insert(4);
+    /// set.insert(9);
+    /// let fds: Vec<RawFd> = set.fds().collect();
+    /// assert_eq!(fds, vec![4, 9]);
+    /// ```
+    #[inline]
+    pub fn fds(&self) -> Fds {
```
asomers, line 83: Typically a C programmer will know what the maximum file descriptor he registered was, and when inspecting the `fd_set` he won't need to iterate all the way to `FD_SETSIZE`, which can be up to 64k on some platforms.  But as written your iterator will always iterate that high.  That's not good for performance.  What if you add an `max: Option<RawFd>` argument to`fds`?  If supplied, the iterator won't check any higher than that.  Such an argument could also be useful to `highest`.

src/sys/select.rs
asomers: I think that for now we should proceed without waiting for https://github.com/rust-lang/libc/pull/1725 .  But copying the entire structure on each loop of the iterator is woefully inefficient.  I think you should retain the `&mut` reference instead.

src/sys/select.rs
zombiezen: Added to `fds()`. I didn't add to `highest()` in the interest of not introducing unnecessary breakage. If someone wants to get that performance gain, they can use `set.fds(Some(my_limit)).next_back()`, which isn't terribly hard to write and is pretty clear.

src/sys/select.rs
zombiezen: Done. This sadly means that the iterator cannot implement `Clone`, but I don't see a good way around that.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   identity for filesystem checks per-thread.
   (#[1163](https://github.com/nix-rust/nix/pull/1163))
 - Derived `Ord`, `PartialOrd` for `unistd::Pid` (#[1189](https://github.com/nix-rust/nix/pull/1189))
+- Added `select::FdSet::fds` method to iterate over file descriptors in a set.
```
asomers, line 18: Don't forget to put a link to the PR in the CHANGELOG

src/sys/select.rs
```diff
@@ -59,14 +61,31 @@ impl FdSet {
...
-            let i = i as RawFd;
-            if unsafe { libc::FD_ISSET(i, self as *mut _ as *mut libc::fd_set) } {
-                return Some(i)
-            }
-        }
+        self.fds(None).next_back()
+    }
 
-        None
+    /// Returns an iterator over the file descriptors in the set. For
```
asomers, line 67: Put a full blank line after the first full sentence.  That will make the docs on docs.rs formatted nicely.

CHANGELOG.md
zombiezen: Done.

src/sys/select.rs
zombiezen: Done.

asomers:
The CI failure was an intermittent network hiccup.

PR #1209, opened 2020-04-08
Support UDP GSO and GRO on linux

glebpom:
This PR implements support for UDP GSO and GRO on Linux. It provides the way to send/receive UDP payloads bigger than interface MTU. The goal is to improve UDP performance. 

GSO was introduced in Linux 4.18, GRO in 5.3

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -463,6 +463,18 @@ pub enum ControlMessageOwned {
         target_os = "openbsd",
     ))]
     Ipv4RecvDstAddr(libc::in_addr),
+
+    /// UDP Generic Receive Offload (GRO) allows receiving multiple UDP
+    /// packets from a single sender.
+    /// Fixed-size payloads are following one by one in a receive buffer.
+    /// This Control Message indicates the size of all smaller packets,
+    /// except, maybe, the last one.
+    ///
+    /// `UdpGroSegment` socket option should be enabled on a socket
+    /// to allow receiving GRO packets.
+    #[cfg(target_os = "linux")]
+    UdpGroSegments(u16),
+
     /// Catch-all variant for unimplemented cmsg types.
     #[doc(hidden)]
     Unknown(UnknownCmsg),
@@ -546,6 +558,11 @@ impl ControlMessageOwned {
                 let dl = ptr::read_unaligned(p as *const libc::in_addr);
                 ControlMessageOwned::Ipv4RecvDstAddr(dl)
             },
+            #[cfg(target_os = "linux")]
+            (libc::SOL_UDP, libc::UDP_GRO) => {
+                let gso_size: u16 = ptr::read_unaligned(p as *const _);
+                ControlMessageOwned::UdpGroSegments(gso_size)
+            },
             (_, _) => {
                 let sl = slice::from_raw_parts(p, len);
                 let ucmsg = UnknownCmsg(*header, Vec::<u8>::from(&sl[..]));
@@ -617,6 +634,16 @@ pub enum ControlMessage<'a> {
     ))]
     AlgSetAeadAssoclen(&'a u32),
 
+    /// UDP GSO makes it possible for applications to generate network packets
+    /// for a virtual MTU much greater than the real one.
+    /// The length of the send data no longer matches the expected length on
+    /// the wire.
+    /// The size of the datagram payload as it should appear on the wire may be
+    /// passed through this control message.
+    /// Send buffer should consist of multiple fixed-size wire payloads
+    /// following one by one, and the last, possibly smaller one.
+    #[cfg(target_os = "linux")]
+    UdpGsoSegments(&'a u16),
 }
 
 // An opaque structure used to prevent cmsghdr from being a public type
@@ -687,6 +714,10 @@ impl<'a> ControlMessage<'a> {
             ControlMessage::AlgSetAeadAssoclen(len) => {
                 len as *const _ as *const u8
             },
+            #[cfg(target_os = "linux")]
+            ControlMessage::UdpGsoSegments(gso_size) => {
+                gso_size as *const _ as *const u8
+            },
         };
         unsafe {
             ptr::copy_nonoverlapping(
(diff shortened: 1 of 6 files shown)


asomers:
Looks like your tests are failing because this feature isn't implemented in Travis.  You'll have to figure out how to test for the feature's availability at runtime.  We have a few other tests that do that already.

Also, I wonder if it would be worth converting the Linux CI tests to use Cirrus-ci.com, if it can get us more coverage.

glebpom:
@asomers I've added the macro to skip tests if kernel version doesn't match. What do you think?

asomers:
Is there not a more direct way to check for kernel support?  Support could be removed from the kernel configuration, for example, in which case this wouldn't work.

glebpom:
Unfortunately, I didn't find any direct way to check if UDP offload is supported.

Checking the error code on actual call (like `ENOSYS`) is not the best idea, since it can be a result of the bug, with using wrong constants or something like that.  

I think that considering the kernel version is generally a good idea because we typically know at what version the particular call appeared. 

Regarding the customization of the kernel, I think it's a pretty common problem for almost any part of nix tests. The good solution would be to check if the particular configuration flag is enabled in kernel config (see [here](https://superuser.com/questions/287371/obtain-kernel-config-from-currently-running-linux-system)).

src/sys/socket/mod.rs
```diff
@@ -463,6 +463,10 @@ pub enum ControlMessageOwned {
         target_os = "openbsd",
     ))]
     Ipv4RecvDstAddr(libc::in_addr),
+
+    #[cfg(target_os = "linux")]
+    UdpGroSegments(u16),
```
asomers, line 468: Document this enum variant, please

src/sys/socket/mod.rs
```diff
@@ -617,6 +626,9 @@ pub enum ControlMessage<'a> {
     ))]
     AlgSetAeadAssoclen(&'a u32),
 
+    ///Gso UDP
```
asomers, line 629: Please document this enum variant.

test/sys/test_socket.rs
```diff
@@ -169,7 +169,7 @@ mod recvfrom {
 
     const MSG: &'static [u8] = b"Hello, World!";
 
-    fn sendrecv<F>(rsock: RawFd, ssock: RawFd, f: F) -> Option<SockAddr>
+    fn sendrecv<F>(rsock: RawFd, ssock: RawFd, max_recv_once: Option<usize>, f: F) -> Option<SockAddr>
```
asomers, line 172: The `max_recv_once` argument seems specific to the GSO tests.  Could you try to make `sendrecv` remain more generic?  Perhaps you could make the new argument an optional `Fn` that would be called after each `recvfrom`.

test/sys/test_socket.rs
```diff
@@ -220,12 +223,84 @@ mod recvfrom {
...
+            let std_sa = SocketAddr::from_str("127.0.0.1:6791").unwrap();
+            let inet_addr = InetAddr::from_std(&std_sa);
+            let sock_addr = SockAddr::new_inet(inet_addr);
+            let rsock = socket(AddressFamily::Inet,
+                               SockType::Datagram,
+                               SockFlag::empty(),
+                               None
+            ).unwrap();
+
+            setsockopt(rsock, UdpGsoSegment, &2).expect("setsockopt UDP_SEGMENT failed");
```
asomers, line 252: This `2` is a magic number.  Please assign it to a named variable, and reference that variable here and on line 262.

test/sys/test_socket.rs
```diff
@@ -220,12 +223,84 @@ mod recvfrom {
...
+                let iov = [IoVec::from_slice(m)];
+                let cmsg = ControlMessage::UdpGsoSegments(&2);
+                sendmsg(s, &iov, &[cmsg], flags, Some(&sock_addr))
+            });
+            // UDP sockets should set the from address
+            assert_eq!(AddressFamily::Inet, from.unwrap().family());
+        }
+
+        #[test]
+        pub fn gro() {
```
asomers, line 272: How does this test work?  I don't understand what effect the UDP_GRO socket option has, or what role the ControlMessage::UdpGsoSegments has.

test/test.rs
```diff
@@ -100,6 +101,48 @@ cfg_if! {
     }
 }
 
+cfg_if! {
+    if #[cfg(any(target_os = "android", target_os = "linux"))] {
+        macro_rules! require_kernel_version {
```
asomers, line 106: This macro would be a lot simpler if it used `nix::src::sys::utsname` rather than read from procfs.  That would make it portable, too.

glebpom:
More details can be found here http://vger.kernel.org/lpc_net2018_talks/willemdebruijn-lpc2018-udpgso-paper-DRAFT-1.pdf

test/sys/test_socket.rs
glebpom: I've lost the comment with the explanation. The problem is that it's not clear how to guarantee that incoming packets are GROed. It may vary between kernel versions, and it may just be impossible on a physical network interface. I've seen some GRO tests working on veth interfaces. I will need some time to research. This test is just setting the socket option and check that it doesn't fail with an error.

asomers: If it's truly unobservable then there's no point in writing a test that doesn't do anything.  Just make sure that you can set the socket options without error.  There are a few functions like that in Nix.

test/sys/test_socket.rs
```diff
@@ -249,7 +249,9 @@ mod recvfrom {
                                None
             ).unwrap();
 
-            setsockopt(rsock, UdpGsoSegment, &2).expect("setsockopt UDP_SEGMENT failed");
```
asomers: Could you perhaps add a comment explaining the principle behind this test?

test/sys/test_socket.rs
```diff
@@ -291,11 +295,11 @@ mod recvfrom {
                 None,
             ).expect("send socket failed");
 
-            let from = sendrecv(rsock, ssock, None, move |s, m, flags| {
```
asomers: The `gro` test, too, could benefit from a comment

PR #1210, opened 2020-04-09
impl io::{Read,Write} for PtyMaster

cuviper:
`PtyMaster` acts like an owned file descriptor, even closing on `Drop`.
Implementing `io::Read` and `io::Write` lets it be used directly in
standard I/O operations.

### DIFF
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -6,6 +6,7 @@ pub use libc::pid_t as SessionId;
 pub use libc::winsize as Winsize;
 
 use std::ffi::CStr;
+use std::io;
 use std::mem;
 use std::os::unix::prelude::*;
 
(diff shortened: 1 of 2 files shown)


asomers:
This change makes sense to me.  However, please split that test method up.  It's too big, and tests too much.  I'm also going to request a review from @Susurrus because he's more knowledgeable about pseudoterminals than I am.

cuviper:
>  However, please split that test method up. It's too big, and tests too much.

I kept it together because the existing test code is all necessary setup to get a usable pair for my read/write tests. Even if I extracted that setup into some common function, the only difference from the current test would be the two asserts for `fd > 0`.

asomers:
> > However, please split that test method up. It's too big, and tests too much.
> 
> I kept it together because the existing test code is all necessary setup to get a usable pair for my read/write tests. Even if I extracted that setup into some common function, the only difference from the current test would be the two asserts for `fd > 0`.

My objection is that you're testing `io::Read` and `io::Write` in the same test, even though they're logically separate functionality.  Those should be in separate test methods, sharing a common setup function.

cuviper:
> Those should be in separate test methods, sharing a common setup function.

OK, I've now split that up.

cuviper:
I've rebased and squashed to one commit.

I think CI isn't triggering with the current GitHub outages, but I'll check back in a while and re-push if that still hasn't gone anywhere...

cuviper:
OK, CI is green now.

asomers:
bors retry

asomers:
bors retry

PR #1213, opened 2020-04-12
Formatting files using cargo fmt

srikwit:


asomers:
I'm going to have to reject this PR.  I'm not saying that there aren't inconsistencies in our style; there are.  But it's not worth breaking every outstanding PR just to fix the style.  If you do a major update to one file however, feel free to update that file's style (in its own commit).

srikwit:
Hi @asomers Thanks for your feedback. I will keep that in mind before my next PR.

PR #1215, opened 2020-04-12
Remove sys::socket::addr::from_libc_sockaddr from the public API

asomers:
This function never should've been public, since it's basically
impossible to use directly.  It's only public due to an oversight from
PR #667 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Removed
 
+- Removed `sys::socket::addr::from_libc_sockaddr` from the public API.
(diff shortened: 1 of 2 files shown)


asomers:
CC @mwanner .  This is basically cleanup from PR #667 .

coolreader18:
This would actually be really useful to me; it's not impossible to use directly if you're doing FFI and get a `*const sockaddr`.

asomers:
> This would actually be really useful to me; it's not impossible to use directly if you're doing FFI and get a `*const sockaddr`.

What is your use case?  How are you getting the `*const sockaddr`?

coolreader18:
Well I suppose I actually have a `socket2::SockAddr`, so at the low level it's from something like `getsockname` or `getpeername` or `recvfrom`

PR #1216, opened 2020-04-14
Add UnixCredentials support on FreeBSD/DragonFly

myfreeweb:
This allows working with `SCM_CREDS` messages, which are like `SCM_CREDENTIALS` on Linux, but slightly different (always overwritten by the kernel, contain a bit more info — euid and groups).

With this PR, it is possible to write portable code that would use the appropriate message for the platform, but one remaining quirk is that `PassCred` thing still has to be present and `cfg`'d to Linux.

Adding the `SCM_CREDS` constant to libc: https://github.com/rust-lang/libc/pull/1740

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -189,12 +189,22 @@ cfg_if! {
     if #[cfg(any(target_os = "android", target_os = "linux"))] {
         /// Unix credentials of the sending process.
         ///
-        /// This struct is used with the `SO_PEERCRED` ancillary message for UNIX sockets.
+        /// This struct is used with the `SO_PEERCRED` ancillary message
+        /// and the `SCM_CREDENTIALS` control message for UNIX sockets.
         #[repr(transparent)]
         #[derive(Clone, Copy, Debug, Eq, PartialEq)]
         pub struct UnixCredentials(libc::ucred);
 
         impl UnixCredentials {
+            /// Creates a new instance with the credentials of the current process
+            pub fn new() -> Self {
+                UnixCredentials(libc::ucred {
+                    pid: crate::unistd::getpid().as_raw(),
+                    uid: crate::unistd::getuid().as_raw(),
+                    gid: crate::unistd::getgid().as_raw(),
+                })
+            }
+
             /// Returns the process identifier
             pub fn pid(&self) -> libc::pid_t {
                 self.0.pid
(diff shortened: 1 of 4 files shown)


src/sys/socket/mod.rs
```diff
@@ -498,6 +561,11 @@ impl ControlMessageOwned {
                 let cred: libc::ucred = ptr::read_unaligned(p as *const _);
                 ControlMessageOwned::ScmCredentials(cred.into())
             }
+            #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+            (libc::SOL_SOCKET, /* libc::SCM_CREDS */ 0x03) => {
```
asomers, line 565: Adding a note to this line so we don't forget to fix it before committing.

src/sys/socket/mod.rs
```diff
@@ -574,17 +642,25 @@ pub enum ControlMessage<'a> {
...
-    /// processes are verified by the kernel.
+    /// process to explicitly send its credentials.
+    ///
+    /// On Linux, a process running as root is allowed to specify any credentials,
+    /// while credentials sent by other processes are verified by the kernel.
+    ///
+    /// On FreeBSD/DragonFly, credentials are always overwritten by the kernel.
     ///
     /// For further information, please refer to the
     /// [`unix(7)`](http://man7.org/linux/man-pages/man7/unix.7.html) man page.
```
asomers, line 657: You should add the FreeBSD man page reference, too.

src/sys/socket/mod.rs
```diff
@@ -729,6 +819,8 @@ impl<'a> ControlMessage<'a> {
             ControlMessage::ScmRights(_) => libc::SCM_RIGHTS,
             #[cfg(any(target_os = "android", target_os = "linux"))]
             ControlMessage::ScmCredentials(_) => libc::SCM_CREDENTIALS,
+            #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+            ControlMessage::ScmCredentials(_) => /* libc::SCM_CREDS */ 0x03, // https://github.com/rust-lang/libc/pull/1740
```
asomers, line 823: Adding a comment here to ensure we fix this before merging

src/sys/socket/mod.rs
```diff
@@ -222,8 +232,56 @@ cfg_if! {
...
+                })
+            }
+
+            /// Returns the process identifier
+            pub fn pid(&self) -> libc::pid_t {
+                self.0.cmcred_pid
+            }
+
+            /// Returns the user identifier
+            pub fn uid(&self) -> libc::uid_t {
```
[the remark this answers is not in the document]
myfreeweb: We can fill them with `getuid`/etc like on Linux, that's not a problem.. other than inefficiency (wasting time making useless syscalls if it's not going to be read like that).

I guess if we go for not unifying `ScmCredentials` and adding `ScmCreds` for FreeBSD, the `ControlMessage` (non-`Owned`) side would just not take a `UnixCredentials` at all.

src/sys/socket/mod.rs
```diff
@@ -222,8 +232,56 @@ cfg_if! {
                 self.0
             }
         }
-    }
-}
+    } else if #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))] {
```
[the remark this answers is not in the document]
myfreeweb: OpenBSD does not have this anymore, I guess I could try macOS since there's CI, but I don't have a NetBSD box ready to test.

src/sys/socket/mod.rs
```diff
@@ -655,6 +728,11 @@ impl<'a> ControlMessage<'a> {
             ControlMessage::ScmCredentials(creds) => {
                 &creds.0 as *const libc::ucred as *const u8
             }
+            #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+            ControlMessage::ScmCreds => {
+                // The kernel overwrites the data, don't worry
+                return
```
asomers, line 734: Could this result in `cmsg_data` being uninitialized?  That way might lead to UB.  I think that you should zero it instead.

GuillaumeDIDIER:
The libc PR (https://github.com/rust-lang/libc/pull/1740) has apparently been merged, if that may be of use to this one.

asomers:
ping @myfreeweb .  It looks like the libc PR has been merged.  Can you finish this one now?

myfreeweb:
It's still not released to crates.io.. Okay, switched to git for now.

PR #1222, opened 2020-04-26
Add Ipv{4,6}PacketInfo support to ControlMessage for send{m,}msg.

isomer:
This adds Ipv4PacketInfo and Ipv6PacketInfo to ControlMessage, allowing these to be used with sendmsg/sendmmsg.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (#[1233](https://github.com/nix-rust/nix/pull/1233))
 - Added `EventFilter` bitflags for `EV_DISPATCH` and `EV_RECEIPT` on OpenBSD.
   (#[1252](https://github.com/nix-rust/nix/pull/1252))
+- Added support for `Ipv4PacketInfo` and `Ipv6PacketInfo` to `ControlMessage`.
+  (#[1222](https://github.com/nix-rust/nix/pull/1222))
 
 ### Changed
 - Changed `fallocate` return type from `c_int` to `()` (#[1201](https://github.com/nix-rust/nix/pull/1201))
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -717,6 +717,25 @@ pub enum ControlMessage<'a> {
     /// following one by one, and the last, possibly smaller one.
     #[cfg(target_os = "linux")]
     UdpGsoSegments(&'a u16),
+
+    /// Configure the sending addressing and interface for v4
+    ///
+    /// For further information, please refer to the
+    /// [`ip(7)`](http://man7.org/linux/man-pages/man7/ip.7.html) man page.
+    #[cfg(any(target_os = "linux",
+              target_os = "macos",
+              target_os = "netbsd"))]
+    Ipv4PacketInfo(&'a libc::in_pktinfo),
+
+    /// Configure the sending addressing and interface for v6
+    ///
+    /// For further information, please refer to the
+    /// [`ipv6(7)`](http://man7.org/linux/man-pages/man7/ipv6.7.html) man page.
+    #[cfg(any(target_os = "linux",
+              target_os = "macos",
+              target_os = "netbsd",
+              target_os = "freebsd"))]
+    Ipv6PacketInfo(&'a libc::in6_pktinfo),
 }
 
 // An opaque structure used to prevent cmsghdr from being a public type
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   receive offload (GRO) ([#1209](https://github.com/nix-rust/nix/pull/1209))
 - Added support for `sendmmsg` and `recvmmsg` calls
   (#[1208](https://github.com/nix-rust/nix/pull/1208))
+- Added support for `Ipv4PacketInfo` and `Ipv6PacketInfo` to `ControlMessage`.
```
asomers, line 28: Provide a link to the PR, please.

src/sys/socket/mod.rs
```diff
@@ -644,6 +644,20 @@ pub enum ControlMessage<'a> {
     /// following one by one, and the last, possibly smaller one.
     #[cfg(target_os = "linux")]
     UdpGsoSegments(&'a u16),
+
+    /// Configure the sending addressing and interface for v4
+    ///
+    /// For further information, please refer to the
+    /// [`ip(7)`](http://man7.org/linux/man-pages/man7/ip.7.html) man page.
+    #[cfg(target_os = "linux")]
```
asomers, line 652: It looks like OSX and NetBSD also support this feature.

src/sys/socket/mod.rs
```diff
@@ -644,6 +644,20 @@ pub enum ControlMessage<'a> {
...
+    Ipv4PacketInfo(&'a libc::in_pktinfo),
+
+    /// Configure the sending addressing and interface for v6
+    ///
+    /// For further information, please refer to the
+    /// [`ipv6(7)`](http://man7.org/linux/man-pages/man7/ipv6.7.html) man page.
+    #[cfg(target_os = "linux")]
+    Ipv6PacketInfo(&'a libc::in6_pktinfo),
```
asomers, line 660: This feature appears to be supported by NetBSD, OSX, and FreeBSD.

CHANGELOG.md
isomer: Done.

src/sys/socket/mod.rs
isomer: Done.

isomer:
> Can you please add tests for the new functionality?

Ah, it took me a while to figure out where the tests were.  Duh.  Now I've found them, I've added some simple tests.

These features are most useful when there are multiple IP addresses on the machine (to force which IP address is used), but I don't think I can assume that the test hosts will have any IPs other than localhost.  So I've forced localhost (despite the kernel probably choosing localhost anyway).

At least under Linux, the kernel is very picky about these options, so if something is wrong, it will generally fail with EINVAL, so I think this should be fine.

So, tl;dr: Done.

isomer:
I'm still learning rust.  I don't understand why the freebsd test is failing – it should be [cfg]'d to not even compile that section of the code on freebsd.

It's late here, so I'm going to bed, I'll pick this up again and try and figure out what's going on tomorrow evening.

test/sys/test_socket.rs
```diff
@@ -667,6 +667,103 @@ pub fn test_af_alg_aead() {
     assert_eq!(decrypted[(assoc_size as usize)..(payload_len + (assoc_size as usize))], payload[(assoc_size as usize)..payload_len + (assoc_size as usize)]);
 }
 
+// Verify `ControlMessage::Ipv4PacketInfo` for `sendmsg`.
+// This creates a (udp) socket bound to localhost, then sends a message to
+// itself but uses Ipv4PacketInfo to force the source address to be localhost.
+//
+// This would be a more interesting test if we could assume that the test host
```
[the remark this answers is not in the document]
isomer: I don't think a good demo here is particularly useful.  These are low level APIs that people shouldn't be using directly.  Somewhere there should be a crate that abstracts over all of this and provides something like `set_local_address()` and figures out what the correct underlying implementation to use (of which these are only one implementation).  For example, for FreeBSD, `Ipv6PacketInfo` is the correct answer, but there's a different solution for IPv4.

I just thought I'd document why the test seems to be particularly inane.  The test should be useful, if anything goes wrong, the kernel will reject it and the test will catch that.

isomer:
> It's lame that Travis doesn't have any IPv6 support. According to their docs, they only have that for arm64, ppc64le, and s390x. So you'll have to conditionally disable that test.

I'm surprised they don't even have support for the loopback interface.  I've changed the code to detect the error when there is no IPv6 loopback and skip the test.   I hope over time the environments will eventually gain IPv6 support, so this will start be tested there when that happens without a human having to keep track of what is going on.

isomer:
There's one failing test, which I think is an unrelated spurious flake - it appears to be to do with aio on one platform.  I can't figure out how to get travis to retry, and unfortunately I've run out of day again.

asomers:
I believe that aio failure is caused by a bug in glibc.  I've restarted it.

asomers:
I think this PR looks good.  If the Travis test passes, could you please squash your commits?

asomers:
Sorry, but merging master into your branch screwed up the PR.  Now other peoples' commits appear to be part of this PR.  You'll have to rebase, not merge.

isomer:
With some help from tazjan I've finally managed to get this properly cleaned up. Please Take Another Look?

PR #1229, opened 2020-05-02
Build Linux aarch64 on Travis's new arm64 hardware

asomers:


asomers:
This sounded like a nice idea.  Unfortunately, Travis's arm64 cluster is either completely down, or so oversubscribed that I couldn't get a single build all weekend.  In either case, it's not usable for us.

Issue #1178, opened 2020-01-12
readlink uses PATH_MAX-sized buffer, but PATH_MAX can't be relied upon


SolraBizna:
The current nix implementation of readlink, which is a rather nice improvement over the previous version, is:

```rust
fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
    match Errno::result(res) {
        Err(err) => Err(err),
        Ok(len) => {
            unsafe { v.set_len(len as usize) }
            Ok(OsString::from_vec(v.to_vec()))
        }
    }
}

pub fn readlink<P: ?Sized + NixPath>(path: &P) -> Result<OsString> {
    let mut v = Vec::with_capacity(libc::PATH_MAX as usize);
    let res = path.with_nix_path(|cstr| {
        unsafe { libc::readlink(cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.capacity() as size_t) }
    })?;

    wrap_readlink_result(&mut v, res)
}
```

The Linux man page for `readlink` has this to say:

> Using a statically sized buffer might not provide enough room for the symbolic link contents. The required size for the buffer can be obtained from the `stat.st_size` value returned by a call to `lstat(2)` on the link. ... Dynamically allocating the buffer for `readlink()` and `readlinkat()` also addresses a common portability problem when using `PATH_MAX` for the buffer size, as this constant is not guaranteed to be defined per POSIX if the system does not have such limit.

Per the above, the fixed-size buffer should be replaced by one of the following approaches:

- Use `lstat` to read the size of the link, allocate a buffer that large plus one, `readlink`. If the return value isn't exactly `v.capacity()-1`, start over. (An approach like this is recommended in the Linux manpage, but supposedly some filesystems' `stat` implementations report a size of zero bytes for all symlinks, so...)
- Starting with a reasonable minimum size (`PATH_MAX.max(128)` or something?) and increasing each time, repeatedly `readlink` until the returned size is *strictly less than* the buffer size. (maybe fall back to this if `lstat` returns zero size?)

Also, we should probably guard against `readlink` returning a size larger than the buffer. I don't know of any actual UNIXoid that does this, but... this is Rust after all.

I would take a crack at this myself, but I'm still recovering from some nastiness. Just composing this issue took me six hours, drifting in and out of dizzy semi-consciousness. Not the best state to write important code in.

asomers:
[...]

PR #1231, opened 2020-05-03
Add support for reading symlinks longer than `PATH_MAX` to `readlink` and `readlinkat`

SolraBizna:
This is in response to issue #1178.

The new logic uses the following approach.

- At any time, if `readlink` returns an error, or a value ≥ 0 and < (not ≤!) the buffer size, we're done.
- Attempt to `readlink` into a `PATH_MAX` sized buffer. (This will almost always succeed, and saves a system call over calling `lstat` first.)
- Try to `lstat` the link. If it succeeds and returns a sane value, allocate the buffer to be that large plus one byte. Otherwise, allocate the buffer to be `PATH_MAX.max(128) << 1` bytes.
- Repeatedly attempt to `readlink`. Any time its result is ≥ (not >!) the buffer size, double the buffer size and try again.

While testing this, I discovered that ext4 doesn't allow creation of a symlink > 4095 (Linux's `PATH_MAX` minus one) bytes long. This is in spite of Linux happily allowing paths in other contexts to be longer than this—including on ext4! This was probably instated to avoid breaking programs that assume `PATH_MAX` will always be enough, but ironically hindered my attempt to test support for *not* assuming. I tested the code using an artificially small `PATH_MAX` and (separately) a wired-to-fail `lstat`. `strace` showed the code behaving precisely as expected. Unfortunately, I can't add an automatic test for this.

Other changes made by this PR:
- `wrap_readlink_result` now calls `shrink_to_fit` on the buffer before returning, potentially reclaiming kilobytes of memory per call. This could be very important if the returned buffer is long-lived.
- `readlink` and `readlink_at` now both call an `inner_readlink` function that contains the bulk of the logic, avoiding copy-pasting of code. (This is much more important now that the logic is more than a few lines long.)

Notably, this PR does *not* add support for systems that don't define `PATH_MAX` at all. As far as I know, I don't have access to any POSIX-ish OS that doesn't have `PATH_MAX`, and I suspect it would have other compatibility issues with `nix` anyway.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -180,33 +180,80 @@ pub fn renameat<P1: ?Sized + NixPath, P2: ?Sized + NixPath>(old_dirfd: Option<Ra
     Errno::result(res).map(drop)
 }
 
-fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
-    match Errno::result(res) {
-        Err(err) => Err(err),
-        Ok(len) => {
-            unsafe { v.set_len(len as usize) }
-            Ok(OsString::from_vec(v.to_vec()))
+fn wrap_readlink_result(mut v: Vec<u8>, len: ssize_t) -> Result<OsString> {
+    unsafe { v.set_len(len as usize) }
+    v.shrink_to_fit();
+    Ok(OsString::from_vec(v.to_vec()))
+}
+
+fn readlink_maybe_at<P: ?Sized + NixPath>(dirfd: Option<RawFd>, path: &P,
+                                          v: &mut Vec<u8>)
+                                          -> Result<libc::ssize_t> {
+    path.with_nix_path(|cstr| {
+        unsafe {
+            match dirfd  {
+                Some(dirfd) => libc::readlinkat(dirfd, cstr.as_ptr(),
+                                                v.as_mut_ptr() as *mut c_char,
+                                                v.capacity() as size_t),
+                None => libc::readlink(cstr.as_ptr(),
+                                       v.as_mut_ptr() as *mut c_char,
+                                       v.capacity() as size_t),
+            }
+        }
+    })
+}
+
+fn inner_readlink<P: ?Sized + NixPath>(dirfd: Option<RawFd>, path: &P)
+                                       -> Result<OsString> {
+    let mut v = Vec::with_capacity(libc::PATH_MAX as usize);
+    // simple case: result is strictly less than `PATH_MAX`
+    let res = readlink_maybe_at(dirfd, path, &mut v)?;
+    let len = Errno::result(res)?;
+    debug_assert!(len >= 0);
+    if (len as usize) < v.capacity() {
+        return wrap_readlink_result(v, res);
+    }
+    // Uh oh, the result is too long...
+    // Let's try to ask lstat how many bytes to allocate.
+    let reported_size = super::sys::stat::lstat(path)
+        .and_then(|x| Ok(x.st_size)).unwrap_or(0);
+    let mut try_size = if reported_size > 0 {
+        // Note: even if `lstat`'s apparently valid answer turns out to be
+        // wrong, we will still read the full symlink no matter what.
+        reported_size as usize + 1
+    } else {
+        // If lstat doesn't cooperate, or reports an error, be a little less
+        // precise.
+        (libc::PATH_MAX as usize).max(128) << 1
+    };
+    loop {
+        v.reserve_exact(try_size);
+        let res = readlink_maybe_at(dirfd, path, &mut v)?;
+        let len = Errno::result(res)?;
+        debug_assert!(len >= 0);
+        if (len as usize) < v.capacity() {
+            break wrap_readlink_result(v, res);
+        }
+        else {
+            // Ugh! Still not big enough!
+            match try_size.checked_shl(1) {
+                Some(next_size) => try_size = next_size,
+                // It's absurd that this would happen, but handle it sanely
+                // anyway.
+                None => break Err(super::Error::Sys(Errno::ENAMETOOLONG))
+            }
         }
     }
 }
 
 pub fn readlink<P: ?Sized + NixPath>(path: &P) -> Result<OsString> {
-    let mut v = Vec::with_capacity(libc::PATH_MAX as usize);
-    let res = path.with_nix_path(|cstr| {
-        unsafe { libc::readlink(cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.capacity() as size_t) }
-    })?;
-
-    wrap_readlink_result(&mut v, res)
+    inner_readlink(None, path)
 }
 
 
-pub fn readlinkat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P) -> Result<OsString> {
-    let mut v = Vec::with_capacity(libc::PATH_MAX as usize);
-    let res = path.with_nix_path(|cstr| {
-        unsafe { libc::readlinkat(dirfd, cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.capacity() as size_t) }
-    })?;
-
-    wrap_readlink_result(&mut v, res)
+pub fn readlinkat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P)
+                                       -> Result<OsString> {
+    inner_readlink(Some(dirfd), path)
 }
 
 /// Computes the raw fd consumed by a function of the form `*at`.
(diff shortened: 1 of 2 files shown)


src/fcntl.rs
```diff
@@ -185,28 +185,70 @@ fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
...
 
-pub fn readlink<P: ?Sized + NixPath>(path: &P) -> Result<OsString> {
+fn inner_readlink<P: ?Sized + NixPath>(dirfd: Option<RawFd>, path: &P) -> Result<OsString> {
     let mut v = Vec::with_capacity(libc::PATH_MAX as usize);
+    // simple case: result is strictly less than `PATH_MAX`
     let res = path.with_nix_path(|cstr| {
-        unsafe { libc::readlink(cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.capacity() as size_t) }
+        unsafe {
+            match dirfd  {
+                Some(dirfd) => libc::readlinkat(dirfd, cstr.as_ptr(), v.as_mut_ptr() as *mut c_char, v.capacity() as size_t),
```
asomers, line 200: Wrap the long lines, please.

src/fcntl.rs
```diff
@@ -185,28 +185,70 @@ fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
...
     })?;
+    let len = Errno::result(res)?;
+    debug_assert!(len >= 0);
+    if (len as usize) < v.capacity() {
+        return wrap_readlink_result(&mut v, res);
+    }
+    // Uh oh, the result is too long...
+    // Let's try to ask lstat how many bytes to allocate.
+    let reported_size = super::sys::stat::lstat(path).and_then(|x| Ok(x.st_size)).unwrap_or(0);
+    let mut try_size = if reported_size > 0 { reported_size as usize + 1 }
```
asomers, line 213: As a matter of style, I prefer
```rust
if condition {
    something
} else {
   something_else
}
```
rather than cramming everything onto two lines

src/fcntl.rs
```diff
@@ -185,28 +185,70 @@ fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
...
+    // If lstat doesn't cooperate, or reports an error, be a little less precise.
+    else { (libc::PATH_MAX as usize).max(128) << 1 };
+    loop {
+        v.reserve_exact(try_size);
```
asomers, line 217: Interesting.  I think this may be the first good use of `reserve_exact` I've seen.

src/fcntl.rs
```diff
@@ -185,28 +185,70 @@ fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
...
+            }
+        })?;
+        let len = Errno::result(res)?;
+        debug_assert!(len >= 0);
+        if (len as usize) < v.capacity() {
+            break wrap_readlink_result(&mut v, res);
+        }
+        else {
+            // Ugh! Still not big enough!
+            let next_size = try_size << 1;
```
asomers, line 233: I don't think we need to gracefully handle symlinks with paths greater than 32 bits.  A panic would suffice.  You can do
```rust
try_size = try_size.checked_mul(2).unwrap();
```

src/fcntl.rs
```diff
@@ -185,28 +185,70 @@ fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
...
+    }
+    // Uh oh, the result is too long...
+    // Let's try to ask lstat how many bytes to allocate.
+    let reported_size = super::sys::stat::lstat(path).and_then(|x| Ok(x.st_size)).unwrap_or(0);
+    let mut try_size = if reported_size > 0 { reported_size as usize + 1 }
+    // If lstat doesn't cooperate, or reports an error, be a little less precise.
+    else { (libc::PATH_MAX as usize).max(128) << 1 };
+    loop {
+        v.reserve_exact(try_size);
+        let res = path.with_nix_path(|cstr| {
```
asomers, line 218: These 13 lines are duped.  Do you think you could deduplicate them?

src/fcntl.rs
asomers, line 213: Instead of comparing `reported_size` to 0, you may as well compare it to `PATH_MAX`.  After all, `lstat` reports less than what we already tried, then we shouldn't trust `lstat`'s output.

src/fcntl.rs
SolraBizna: What is this project's wrap width? (I normally wrap to 80 columns, but this project clearly has a wider convention.)

src/fcntl.rs
SolraBizna: I think it's better to return an error in this case (as below), but if you ask me to I'll replace that logic with a checked multiply.

src/fcntl.rs
SolraBizna: Trusting `lstat` after a failed `readlink` handles a race condition where the link size changes between the first `readlink` and the first `lstat`.

src/fcntl.rs
SolraBizna: I think the wrong lines got selected. Do you mean L197-L204 and L218-L225?

SolraBizna:
> I just checked, and I can't create a too-long symlink on FreeBSD, NetBSD, OpenBSD, or Solaris. Is there any operating system that _does_ allow it?

On Linux, as far as I can tell, it's up to the particular filesystem. None of [ext*, XFS, btrfs] support very long symlinks, though.

SolraBizna:
Thank you for the thorough review. I have implemented every requested change except the panic on absurdly long symlinks. I'll implement that too while I'm "in the code" if you like.

(I do agree that a >4GiB symlink is absolutely, without a doubt useless, but I also subscribe to the GNU convention of avoiding arbitrary limits where possible, even reasonable ones.)

src/fcntl.rs
```diff
@@ -180,33 +180,88 @@ pub fn renameat<P1: ?Sized + NixPath, P2: ?Sized + NixPath>(old_dirfd: Option<Ra
     Errno::result(res).map(drop)
 }
 
-fn wrap_readlink_result(v: &mut Vec<u8>, res: ssize_t) -> Result<OsString> {
+fn wrap_readlink_result(mut v: Vec<u8>, res: ssize_t) -> Result<OsString> {
     match Errno::result(res) {
         Err(err) => Err(err),
```
asomers, line 185: This line is dead code.  We only ever call `wrap_readlink_result` when `res` is >= 0, so you may as well eliminate the `Errno::result` part.  I think that issued predated your changes, but since you're here would you please fix it?

src/fcntl.rs
```diff
@@ -180,33 +180,88 @@ pub fn renameat<P1: ?Sized + NixPath, P2: ?Sized + NixPath>(old_dirfd: Option<Ra
...
+    let res = readlink_maybe_at(dirfd, path, &mut v)?;
+    let len = Errno::result(res)?;
+    debug_assert!(len >= 0);
+    if (len as usize) < v.capacity() {
+        return wrap_readlink_result(v, res);
+    }
+    // Uh oh, the result is too long...
+    // Let's try to ask lstat how many bytes to allocate.
+    let reported_size = super::sys::stat::lstat(path)
+        .and_then(|x| Ok(x.st_size)).unwrap_or(0);
```
asomers, line 224: You could simply this line into `.map_or(0, |x| x.st_size)`

src/fcntl.rs
asomers: In fact, I think this code will still panic if the size overflows 32 bits.  I think `<<` , like most operators, will panic on overflow in debug builds (but wrap on release builds).  If you really want to handle the overflow without panicing, you need to use `checked_shl`.

SolraBizna:
> In fact, I think this code will still panic if the size overflows 32 bits. I think `<<` , like most operators, will panic on overflow in debug builds (but wrap on release builds).

It seems that, barring obviously foolish cases caught by the compiler, bit shifting wraps in both debug mode and release mode. Guess Rust's designers figured that overflow in bit shifting is likely to be intentional. Regardless, I replaced that code with `checked_shl` just because it's clearer and more idiomatic that way.

Thanks again for the feedback. I've pushed the requested changes. If you think of any other tweaks you'd like me to make, let me know.

SolraBizna:
Huh. Using `map_or` broke the build on FreeBSD. Is it using an old `rustc` or something? (`map_or` seems to have been added in 1.41, which was released this last January.)

asomers:
> Huh. Using `map_or` broke the build on FreeBSD. Is it using an old `rustc` or something? (`map_or` seems to have been added in 1.41, which was released this last January.)

Yes.  Nix's MSRV is currently 1.36.0.  We're pretty conservative with MSRV because we're consumed by platforms that are slow to update, like OpenBSD and embedded systems.  Could you please remove `map_or`?  Fie on me for suggesting it.

SolraBizna:
Done, and now all the checks pass again.

asomers:
Great!  Now all it needs is a squash.

SolraBizna:
Should I be squashing it on my end? I assumed this would be a case for the "Squash and merge" button.

asomers:
Yes you should squash it.  Because we use the bors merge bot, we cannot squash on our end.

SolraBizna:
There. Hopefully I did that properly.

PR #1235, opened 2020-05-11
Limit internal termios API to pub(crate)

### LINKED ISSUE #1115: Termios needs a safe constructor
The only constructor method for `Termios` is the unsafe and undocumented `default_uninit`.  All of the examples use that function.  But not only is it `unsafe`, it also uses `std::mem::uninitialized`, which will be deprecated in Rust 1.38.0.  This will soon be the last remaining use of `std::mem::uninitialized`.

Susurrus:
Fixes #1115

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Removed `sys::socket::addr::from_libc_sockaddr` from the public API.
   (#[1215](https://github.com/nix-rust/nix/pull/1215))
+- Removed `sys::termios::{get_libc_termios, get_libc_termios_mut, update_wrapper`
+  from the public API. These were previously hidden in the docs but still usable
+  by downstream.
+  (#[1235](https://github.com/nix-rust/nix/pull/1235))
 
 - Nix no longer implements `NixPath` for `Option<P> where P: NixPath`.  Most
   Nix functions that accept `NixPath` arguments can't do anything useful with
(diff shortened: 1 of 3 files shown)


Susurrus:
Needs a release note, but let's see how it goes through review and CI then I'll add one.

asomers:
Just like integration tests, I don't think that doc tests are allowed to access private methods.  That's why the tests are failing.

Susurrus:
Just removed it entirely replacing it with a `mem::zerod()` call. Not a great solution, but considering the use case is just for documentation examples, I'm fine with it. Also added a CHANGELOG entry, so we should be good to go here.

Random thought: I just realized that we don't put the update libc version in the CHANGELOG even though it's part of our public API. We should probably add that as a Changed entry when we do releases.

asomers:
> Just removed it entirely replacing it with a `mem::zerod()` call. Not a great solution, but considering the use case is just for documentation examples, I'm fine with it. Also added a CHANGELOG entry, so we should be good to go here.
> 
> Random thought: I just realized that we don't put the update libc version in the CHANGELOG even though it's part of our public API. We should probably add that as a Changed entry when we do releases.

Is the libc version really part of the public API?  A consuming crate can always pull in multiple versions of libc if it needs to.

Susurrus:
> Is the libc version really part of the public API? A consuming crate can always pull in multiple versions of libc if it needs to.

I'm not certain if we directly expose `libc` types in any of our functions or structs anymore, but we do re-export the entire library, so it's part of our API. So while downstream can use their own `libc`, their code may still break if `libc` changes underneath us and they use the version we re-export. I don't think this is a big deal for us, but I think because of it we should list the `libc` version for every `nix` release.

PR #1239, opened 2020-05-13
Replace void crate with Rust standard lib Infallible type

### LINKED ISSUE #1238: Replace void's Void with std Infallible

oherrala:
[`std::convert::Infallible`](https://doc.rust-lang.org/stable/std/convert/enum.Infallible.html) has been available since Rust 1.34 and nix currently targets Rust 1.36 or later so this should not cause problems.

Fixes #1238


oherrala:
@asomers Changelog fixed. Thanks!

PR #1242, opened 2020-05-16
Don't implement `NixPath` for `Option<&P> where P: NixPath`

asomers:
Most Nix functions that accept `NixPath` arguments can't do anything
useful with `None`.  The exceptions (`mount` and `quotactl_sync`)
already take explicitly optional arguments.

Also, this changes the behavior of `mount` with `None` arguments.
Previously, it would call mount(2) with empty strings for those
arguments.  Now, it will use null pointers.


asomers:
Supersedes #1236

baloo:
Can confirm it works as expected:
```
[pid 109788] mount("proc", "/proc", "proc", MS_NOSUID|MS_NODEV|MS_NOEXEC, NULL) = 0
```

asomers:
The bors build failed because the VM got usurped.

bors retry

PR #1244, opened 2020-05-16
Clippy cleanup

asomers:
Reported-by: Clippy

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -221,6 +221,12 @@ cfg_if! {
             }
         }
 
+        impl Default for UnixCredentials {
+            fn default() -> Self {
+                Self::new()
+            }
+        }
+
         impl From<libc::ucred> for UnixCredentials {
             fn from(cred: libc::ucred) -> Self {
                 UnixCredentials(cred)
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1606,16 +1606,19 @@ pub fn getsockname(fd: RawFd) -> Result<SockAddr> {
     }
 }
 
-/// Return the appropriate `SockAddr` type from a `sockaddr_storage` of a certain
-/// size.  In C this would usually be done by casting.  The `len` argument
+/// Return the appropriate `SockAddr` type from a `sockaddr_storage` of a
+/// certain size.
+///
+/// In C this would usually be done by casting.  The `len` argument
 /// should be the number of bytes in the `sockaddr_storage` that are actually
 /// allocated and valid.  It must be at least as large as all the useful parts
 /// of the structure.  Note that in the case of a `sockaddr_un`, `len` need not
 /// include the terminating null.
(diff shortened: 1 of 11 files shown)


asomers:
@kamalmarhubi @Susurrus @posborne could one of you please review this PR?  I especially would like some review for the commits that make previously unsafe functions safe.  This PR eliminates almost all of the outstanding clippy issues.  It would be great to enable clippy in CI.

kamalmarhubi:
I took a quick look, but I think realistically I want to do a closer read, especially since it's been a while since I worked in this codebase. I'll aim to do that in the next few days, by the end of the week at the latest. Does that work?

asomers:
Yeah; there's no urgency.

src/sys/ptrace/linux.rs
```diff
@@ -173,6 +173,7 @@ libc_bitflags! {
     since="0.10.0",
     note="usages of `ptrace()` should be replaced with the specialized helper functions instead"
 )]
+#[allow(clippy::missing_safety_doc)]    // It's deprecated anyway
```
Susurrus, line 176: Any reason we shouldn't just remove this API at this point? It's been 2.5 years at this point. I'd imagine if we've had at least two releases and it's been at least a year we can probably remove deprecated APIs.

asomers: Good point.  I'll take it out.

asomers: Ok, I removed it.  But I decided to remove some other stuff, too, and created a new PR: #1255 .

src/sys/socket/mod.rs
```diff
@@ -1565,49 +1565,68 @@ pub fn getsockname(fd: RawFd) -> Result<SockAddr> {
...
 /// should be the number of bytes in the `sockaddr_storage` that are actually
 /// allocated and valid.  It must be at least as large as all the useful parts
 /// of the structure.  Note that in the case of a `sockaddr_un`, `len` need not
 /// include the terminating null.
-pub unsafe fn sockaddr_storage_to_addr(
+pub fn sockaddr_storage_to_addr(
     addr: &sockaddr_storage,
     len: usize) -> Result<SockAddr> {
 
+    assert!(len <= mem::size_of::<sockaddr_un>());
```
Susurrus, line 1586: Or `debug_assert!`? Not certain when you should use which. If we expect this to be on a hot-path might be worth using `debug_assert` here instead. Otherwise `assert` is probably fine.

asomers: My usual habit is to use `debug_assert` for inputs that came from inside of the module, and `assert` for inputs that cam from outside.

asomers:
bors retry

asomers:
redoxer hanged again.  I think I'll disable the Redox build if we can't get that issue fixed.

bors retry

asomers:
bors retry

asomers:
rebased to fix merge conflict and also fix a formatting error in CHANGELOG

PR #1259, opened 2020-06-16
implement ttyname and ttyname_r

### LINKED ISSUE #1204: Support for ttyname and ttyname_r
POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/functions/ttyname.html

```Rust
pub unsafe fn ttyname(fd: RawFd) -> Result<String>;
pub fn ttyname_r(fd: RawFd) -> Result<String>;
```

`ttyname` and `ttyname_r` are basically `ptsname` that works on any TTY fds, not just those associated with a PTY slave device. The latter is a pretty useful portable alternative to `ptsname_r`.

The functions are found on basically every nix platform; see `libc` crate for details about a macOS compliance fix it has.

doy:
this fixes #1204.

i know this at least works on linux, but i'm not super sure about other platforms - happy to add conditionals if it makes sense.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -2774,3 +2774,20 @@ impl Group {
         })
     }
 }
+
+/// Get the name of the terminal device that is open on file descriptor fd
+/// (see [`ttyname(3)`](http://man7.org/linux/man-pages/man3/ttyname.3.html)).
+pub fn ttyname(fd: RawFd) -> Result<PathBuf> {
+    const PATH_MAX: usize = libc::PATH_MAX as usize;
+    let mut buf = vec![0_u8; PATH_MAX];
+    let c_buf = buf.as_mut_ptr() as *mut libc::c_char;
+
+    let ret = unsafe { libc::ttyname_r(fd, c_buf, buf.len()) };
+    if ret != 0 {
+        return Err(Error::Sys(Errno::from_i32(ret)));
+    }
+
+    let nul = buf.iter().position(|c| *c == b'\0').unwrap();
+    buf.truncate(nul);
+    Ok(OsString::from_vec(buf).into())
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (#[1222](https://github.com/nix-rust/nix/pull/1222))
 - `CpuSet` and `UnixCredentials` now implement `Default`.
   (#[1244](https://github.com/nix-rust/nix/pull/1244))
+- Added `unistd::ttyname`
+  (#[1259](https://github.com/nix-rust/nix/pull/1259))
 
 ### Changed
 - Changed `fallocate` return type from `c_int` to `()` (#[1201](https://github.com/nix-rust/nix/pull/1201))
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -7,6 +7,8 @@ use nix::unistd::ForkResult::*;
 use nix::sys::signal::{SaFlags, SigAction, SigHandler, SigSet, Signal, sigaction};
 use nix::sys::wait::*;
 use nix::sys::stat::{self, Mode, SFlag};
+#[cfg(not(target_os = "redox"))]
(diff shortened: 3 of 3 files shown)


doy:
(i'd really like to see this change, but i don't have access to a dev machine with osx - is there a way forward here?)

src/unistd.rs
```diff
@@ -2778,3 +2778,42 @@ impl Group {
...
+/// (see [`ttyname(3)`](http://man7.org/linux/man-pages/man3/ttyname.3.html)).
+///
+/// # Safety
+///
+/// `ttyname()` mutates global variables and is *not* threadsafe.
+/// Mutating global variables is always considered `unsafe` by Rust and this
+/// function is marked as `unsafe` to reflect that.
+///
+/// For a threadsafe and non-`unsafe` alternative, see `ttyname_r()`.
+#[inline]
```
asomers, line 2792: Don't use `#[inline]` for new code.  The compiler can figure it out, at least when LTO is enabled.

src/unistd.rs
```diff
@@ -2778,3 +2778,42 @@ impl Group {
...
+    const PATH_MAX: usize = libc::PATH_MAX as usize;
+    let mut buf = [0_u8; PATH_MAX];
+    let c_buf = buf.as_mut_ptr() as *mut libc::c_char;
+
+    let ret = unsafe { libc::ttyname_r(fd, c_buf, buf.len()) };
+    if ret != 0 {
+        return Err(Error::Sys(Errno::last()));
+    }
+
+    let nul = buf.iter().position(|c| *c == b'\0').unwrap();
```
asomers, line 2817: Instead of manually searching for NUL, why not just use `CStr::from_ptr`?  In fact, you should technically be using `OsString` here instead of `CString`, because the string comes from the operating system and not from another programming language.  So you should declare the buf as a `Vec`, and then use `OsString::from_vec` followed by `to_string_lossy`.

src/unistd.rs
```diff
@@ -2778,3 +2778,42 @@ impl Group {
...
+///
+/// # Safety
+///
+/// `ttyname()` mutates global variables and is *not* threadsafe.
+/// Mutating global variables is always considered `unsafe` by Rust and this
+/// function is marked as `unsafe` to reflect that.
+///
+/// For a threadsafe and non-`unsafe` alternative, see `ttyname_r()`.
+#[inline]
+pub unsafe fn ttyname(fd: RawFd) -> Result<String> {
```
asomers, line 2793: Is there a good reason why anybody would ever want `ttyname`?  Since the invention of `ttyname_r`, I think the latter should be used in all cases.  If you don't know of a use case where `ttyname_r` does not suffice, then I'd just as soon we don't bind `ttyname` at all.

src/unistd.rs
doy: :+1:

src/unistd.rs
doy: yeah, good point - i was mainly copying from the `ptsname` interface, but i agree that it's really not necessary.

src/unistd.rs
doy: i think returning a PathBuf is actually better than String here, and using OsString for that makes sense. that does mean that i do have to still search for the nul myself though, since OsString won't do that (it'll just create a string with a bunch of trailing nuls). going through CStr (and then being forced to allocate a copy into a second Vec) doesn't seem worth it to avoid that?

doy:
alright, i'll play around with travis some more.

doy:
looks like it was just an issue with the way i was structuring the test - `ttyname` seems to fail on osx when called on a pty master for some reason, but that's basically never a useful thing to do, so it's probably fine to just test something more along the lines of something people would actually use this for. there are still a couple failing builds, but it looks like every pull request for the past few weeks has been failing those builds, so i don't think they are related.

asomers:
The powerpc64 build has only been failing for a few days, and it's fixed on Nix's master branch.  You'll have to rebase.

doy:
done

doy:
rebased again

doy:
test_aio_drop also seems flaky https://travis-ci.org/github/nix-rust/nix/jobs/704771690

PR #1261, opened 2020-06-22
Adding an implementation for timerfd.

vdagonneau-anssi:
Hello,

I have encountered a situation where I needed to use timerfd in addition to other fd with nix implementation of epoll. 

Here is a basic implementation of this feature, I also added some documentation and tests.

Note: currently testing works by pausing for 1-3 seconds. Obviously it is pretty bad if we want tests to run quickly but that was my best idea at the time. Feel free to suggest any other way.

Thank you!

### DIFF
--- a/src/sys/timerfd.rs
+++ b/src/sys/timerfd.rs
@@ -1,0 +1,272 @@
+//! Timer API via file descriptors.
+//!
+//! Timer FD is a Linux-only API to create timers and get expiration
+//! notifications through file descriptors.
+//!
+//! For more documentation, please read [timerfd_create(2)](http://man7.org/linux/man-pages/man2/timerfd_create.2.html).
+//!
+//! # Examples
+//!
+//! Create a new one-shot timer that expires after 1 second.
+//! ```
+//! # use std::os::unix::io::AsRawFd;
+//! # use nix::sys::timerfd::{TimerFd, ClockId, TimerFlags, TimerSetTimeFlags,
+//! #    Expiration};
+//! # use nix::sys::time::{TimeSpec, TimeValLike};
+//! # use nix::unistd::read;
+//! #
+//! // We create a new monotonic timer.
+//! let timer = TimerFd::new(ClockId::CLOCK_MONOTONIC, TimerFlags::empty())
+//!     .unwrap();
+//!
+//! // We set a new one-shot timer in 1 seconds.
+//! timer.set(
+//!     Expiration::OneShot(TimeSpec::seconds(1)),
+//!     TimerSetTimeFlags::empty()
+//! ).unwrap();
+//!
+//! // We wait for the timer to expire.
+//! timer.wait().unwrap();
+//! ```
+use crate::sys::time::TimeSpec;
+use crate::unistd::read;
+use crate::{errno::Errno, Error, Result};
+use bitflags::bitflags;
+use libc::c_int;
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+
+/// A timerfd instance. This is also a file descriptor, you can feed it to
+/// other interfaces consuming file descriptors, epoll for example.
+#[derive(Debug, Clone, Copy)]
+pub struct TimerFd {
+    fd: RawFd,
+}
+
+impl AsRawFd for TimerFd {
+    fn as_raw_fd(&self) -> RawFd {
+        self.fd
+    }
+}
+
+impl FromRawFd for TimerFd {
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        TimerFd { fd }
+    }
+}
+
+libc_enum! {
+    /// The type of the clock used to mark the progress of the timer. For more
+    /// details on each kind of clock, please refer to [timerfd_create(2)](http://man7.org/linux/man-pages/man2/timerfd_create.2.html).
+    #[repr(i32)]
+    pub enum ClockId {
+        CLOCK_REALTIME,
+        CLOCK_MONOTONIC,
+        CLOCK_BOOTTIME,
+        CLOCK_REALTIME_ALARM,
+        CLOCK_BOOTTIME_ALARM,
+    }
+}
+
+libc_bitflags! {
+    /// Additional flags to change the behaviour of the file descriptor at the
+    /// time of creation.
+    pub struct TimerFlags: c_int {
+        TFD_NONBLOCK;
+        TFD_CLOEXEC;
+    }
+}
+
+bitflags! {
+    /// Flags that are used for arming the timer.
+    pub struct TimerSetTimeFlags: libc::c_int {
+        const TFD_TIMER_ABSTIME = libc::TFD_TIMER_ABSTIME;
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+struct TimerSpec(libc::itimerspec);
+
+impl TimerSpec {
+    pub fn none() -> Self {
+        Self(libc::itimerspec {
+            it_interval: libc::timespec {
+                tv_sec: 0,
+                tv_nsec: 0,
+            },
+            it_value: libc::timespec {
+                tv_sec: 0,
+                tv_nsec: 0,
+            },
+        })
+    }
+}
+
+impl AsRef<libc::itimerspec> for TimerSpec {
+    fn as_ref(&self) -> &libc::itimerspec {
+        &self.0
+    }
+}
+
+impl From<Expiration> for TimerSpec {
+    fn from(expiration: Expiration) -> TimerSpec {
+        match expiration {
+            Expiration::OneShot(t) => TimerSpec(libc::itimerspec {
+                it_interval: libc::timespec {
+                    tv_sec: 0,
+                    tv_nsec: 0,
+                },
+                it_value: *t.as_ref(),
+            }),
+            Expiration::IntervalDelayed(start, interval) => TimerSpec(libc::itimerspec {
+                it_interval: *interval.as_ref(),
+                it_value: *start.as_ref(),
+            }),
+            Expiration::Interval(t) => TimerSpec(libc::itimerspec {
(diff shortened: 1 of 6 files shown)


test/sys/test_timerfd.rs
```diff
@@ -0,0 +1,49 @@
...
+    fd.settime(
+        TimerSetTimeFlags::empty(),
+        Expiration::OneShot(TimeSpec::seconds(1)),
+    )
+    .unwrap();
+
+    read(fd.as_raw_fd(), &mut buffer).unwrap();
+
+    let millis = before.elapsed().as_millis();
+    assert!(millis < 1100 && millis > 900);
```
asomers-ax, line 25: You cannot assert that `millis < 1100`.  There is no guarantee that the timer will expire promptly, or that this process will run immediately after the timer expires.

test/sys/test_timerfd.rs
```diff
@@ -0,0 +1,49 @@
...
+    fd.settime(
+        TimerSetTimeFlags::empty(),
+        Expiration::IntervalDelayed(TimeSpec::seconds(1), TimeSpec::seconds(2)),
+    )
+    .unwrap();
+
+    read(fd.as_raw_fd(), &mut buffer).unwrap();
+
+    let start_delay = before.elapsed().as_millis();
+    assert!(start_delay < 1100 && start_delay > 900);
```
asomers-ax, line 43: Ditto as above.  You can't assert an upper bound for the delay.

src/sys/timerfd.rs
```diff
@@ -0,0 +1,124 @@
...
+//!
+//! Timer FD is a Linux-only API to create timers and get expiration notifications through file
+//! descriptors.
+//!
+//! For more documentation, please read [timerfd_create(2)](http://man7.org/linux/man-pages/man2/timerfd_create.2.html).
+//!
+//! # Examples
+//!
+//! Create a new one-shot timer that expires after 5 seconds.
+//! ```no_run
```
asomers-ax, line 11: Why the `no_run`?  You might as well run it, though maybe decrease the timer to 1 second.  Also, the example would be more useful if there were a `read` at the end.

src/sys/timerfd.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum Expiration {
+    OneShot(TimeSpec),
+    IntervalDelayed(TimeSpec, TimeSpec),
+    Interval(TimeSpec),
+}
+
+impl TimerFd {
+    pub fn new(clockid: ClockId, flags: TimerFlags) -> Result<Self> {
+        let clid = match clockid {
```
asomers-ax, line 101: I think you can skip this block if you define ClockId using the `libc_enum` macro.

src/sys/timerfd.rs
```diff
@@ -0,0 +1,124 @@
...
+            ClockId::Monotonic => libc::CLOCK_MONOTONIC,
+            ClockId::BootTime => libc::CLOCK_BOOTTIME,
+            ClockId::RealtimeAlarm => libc::CLOCK_REALTIME_ALARM,
+            ClockId::BootTimeAlarm => libc::CLOCK_BOOTTIME_ALARM,
+        };
+
+        Errno::result(unsafe { libc::timerfd_create(clid, flags.bits()) }).map(|fd| Self { fd })
+    }
+
+    pub fn settime(&self, flags: TimerSetTimeFlags, expiration: Expiration) -> Result<()> {
```
asomers-ax, line 112: Could we have docs for all public functions?

src/sys/timerfd.rs
```diff
@@ -0,0 +1,124 @@
+//! Timer API via file descriptors.
+//!
+//! Timer FD is a Linux-only API to create timers and get expiration notifications through file
+//! descriptors.
+//!
+//! For more documentation, please read [timerfd_create(2)](http://man7.org/linux/man-pages/man2/timerfd_create.2.html).
```
asomers-ax, line 6: Wrap to 80 cols, please.

test/sys/test_timerfd.rs
```diff
@@ -0,0 +1,49 @@
...
+use nix::sys::time::{TimeSpec, TimeValLike};
+use nix::sys::timerfd::{ClockId, Expiration, TimerFd, TimerFlags, TimerSetTimeFlags};
+use nix::unistd::read;
+use nix::Error;
+use std::os::unix::io::AsRawFd;
+use std::time::Instant;
+
+#[test]
+pub fn test_timerfd_oneshot() {
+    let mut buffer = [0u8; 64];
```
asomers-ax, line 11: If buffer is never read, can it be a 0-length slice?

test/sys/test_timerfd.rs
vdagonneau-anssi: You are right, I removed the upper bound. Do you think setting a lower bound could be problematic in some corner cases ? I could not think of any "practical" problems with that.

test/sys/test_timerfd.rs
vdagonneau-anssi: Same as above.

src/sys/timerfd.rs
vdagonneau-anssi: Ok I removed the no_run, decreased the timer and added a read at the end.

src/sys/timerfd.rs
vdagonneau-anssi: Ok refactored ClockId to use libc_enum.

test/sys/test_timerfd.rs
vdagonneau-anssi: So, no I cannot make it a 0-length slice, the syscall requires a buffer that is no less that 8-bytes long. If you don't you get an EINVAL.

That said 64-bytes is way too big. I changed it to use the minimum required size of 8-bytes.

src/sys/timerfd.rs
vdagonneau-anssi: Ok, wrapped everything to 80 cols.

vdagonneau-anssi:
I addressed most of your comments in the latest commit I think. 

As for having a sleep method or equivalent, I think it is a fairly `high level` interface and I will need to think about how to implement it properly.

It is worth noting that reading on the `fd` has different meanings depending on how you created the timerfd in the first place (or modified it with fcntl). Specifically, the call to read can either be blocking or not.

If we indeed need a higher level interface then I would like to move away from the method names I have right now to make it clear that the semantics are different from the underlying syscalls.

test/sys/test_timerfd.rs
asomers: I don't know of any case where the lower bound could be a problem, unless the read gets interrupted by a signal.  What does Linux do in that case?  If it's possible for the read to return early with EINTR, then you should retry the read in your higher-level method.

src/sys/timerfd.rs
```diff
@@ -1,33 +1,45 @@
...
 //! #
 //! // We create a new monotonic timer.
-//! let fd = TimerFd::new(ClockId::Monotonic, TimerFlags::empty()).unwrap();
+//! let fd = TimerFd::new(ClockId::CLOCK_MONOTONIC, TimerFlags::empty())
+//!     .unwrap();
 //!
 //! // We set a new one-shot timer in 5 seconds.
-//! fd.settime(TimerSetTimeFlags::empty(), Expiration::OneShot(TimeSpec::seconds(5))).unwrap();
+//! fd.settime(
+//!     Expiration::OneShot(TimeSpec::seconds(5)),
```
asomers, line 24: Be sure to set the time to just 1 second.  5 seconds is too long for a doc test.

src/sys/timerfd.rs
```diff
@@ -1,33 +1,45 @@
...
 use crate::errno::Errno;
 use crate::sys::time::TimeSpec;
 use crate::Result;
 use bitflags::bitflags;
 use libc::c_int;
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 
+/// A timerfd instance. This is also a file descriptor, you can feed it to
+/// other interfaces consuming file descriptors, epoll for example.
```
asomers, line 39: Ahh, nice addition here.  I didn't realize that timerfd could be used with epoll.  That's probably why it was created, I'm guessing.

src/sys/timerfd.rs
vdagonneau-anssi: Ok, fixed it.

src/sys/timerfd.rs
vdagonneau-anssi: Yes, that is indeed why it was created.

vdagonneau-anssi:
I pushed a new set of changes including a slightly higher level interface.

You now have 5 methods for the TimerFd: `new`, `set`, `get`, `unset`and `wait`.

The idea is that you use `new` to create a new timerfd, it is created without any expiration. You can then `set` a new expiration and `get` it back. If you wish you can also `unset` any  previous expiration that has been set. Finally, you can "acknowledge"/"wait" the timer with `wait`.

I still feel somewhat uneasy about the `wait` name as it really doesn't make a lot of sense in the context of `epoll`.

Typically in the case of `epoll`, you would create a timerfd, set an expiration, feed it to epoll. Once the timer is triggered, epoll will notify you that the timerfd has become readable and you need to "acknowledge" it (a.k.a. read from it) otherwise epoll is going to keep notifying you about it.

The name `wait` works well in a conventional approach where you create the timerfd, set an expiration and the read from it. In that case `wait` will indeed wait until timerfd becomes readable.

vdagonneau-anssi:
Ok, I added it to the changelog.

asomers:
Only one more problem: you'll have to rebase to fix the test failure.

vdagonneau-anssi:
Ok, I rebased on top of master.

PR #1264, opened 2020-06-27
Update the FreeBSD test image to 11.4-release

asomers:
And temporarily disable travis targets

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,12 +1,12 @@
 freebsd_instance:
-  image_family: freebsd-11-3-snap
+  image: freebsd-11-4-release-amd64
 
 # Test FreeBSD in a full VM on cirrus-ci.com.  Test the i686 target too, in the
 # same VM.  The binary will be built in 32-bit mode, but will execute on a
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -1,12 +1,12 @@
 freebsd_instance:
-  image_family: freebsd-11-3-snap
+  image_family: freebsd-11-4-release
```
fkorotkov, line 2: You either should use `image` for `freebsd-11-4-release` or:

```suggestion
  image_family: freebsd-11-4
```

asomers: Thanks for the tip, @fkorotkov .

asomers:
@fkorotkov do you know why the build fails?  When trying to start a FreeBSD instance on Cirrus, it fails with the error "Failed to start an instance: INVALID_ARGUMENT: Bad Request".

fkorotkov:
My bad, the image should be `freebsd-11-4-release-amd64`

asomers:
> My bad, the image should be `freebsd-11-4-release-amd64`

Thanks for the tip.  It's annoying that Google doesn't make it easy to discover the correct image name through the web UI.

PR #1270, opened 2020-07-05
Support nullable timeout in ppoll

MikailBag:


MikailBag:
As stated in Linux man: 

> If tmo_p is specified as NULL, then ppoll() can block indefinitely.

I updated documentation and changelog.

MikailBag:
[Dragonfly man](https://man.dragonflybsd.org/?command=ppoll&section=2): 
> A null pointer may be passed to indicate that ppoll() should wait indefinitely. 

[FreeBSD man](https://man.dragonflybsd.org/?command=ppoll&section=2): 
> A null pointer may be passed to indicate that ppoll() should wait indefinitely.

src/poll.rs
```diff
@@ -130,16 +130,21 @@ pub fn poll(fds: &mut [PollFd], timeout: libc::c_int) -> Result<libc::c_int> {
...
 /// `ppoll` behaves like `poll`, but let you specify what signals may interrupt it
-/// with the `sigmask` argument.
+/// with the `sigmask` argument. If you want `ppoll` to block indefinitely,
+/// specify `None` as `timeout` (it is like `timeout = -1` for `poll`).
 ///
 #[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd", target_os = "linux"))]
-pub fn ppoll(fds: &mut [PollFd], timeout: TimeSpec, sigmask: SigSet) -> Result<libc::c_int> {
-
+pub fn ppoll(fds: &mut [PollFd], timeout: Option<TimeSpec>, sigmask: SigSet) -> Result<libc::c_int> {
+    let timeout = &timeout;
```
asomers, line 138: Why do you need this line?

MikailBag: You're right, line can be deleted.

asomers:
@MikailBag that test failure looks relevant.  Could you please investigate?

MikailBag:
Hmm, it seems that without `let timeout = &timeout` we had a dangling pointer (in match, `t.as_ref()` created *temporary* reference and converted it into *temporary* pointer). So it happened that `timeout` started pointing to some negative value, causing EINVAL.

I think now there is no UB :smile:

asomers:
Good.  Can you squash it now?

MikailBag:
I've squashed.

PR #1274, opened 2020-07-23
Add `sockaddr_ll` to `sockaddr_storage_to_addr`

isomer:
This adds support for `sockaddr_ll` to `sockaddr_storage_to_addr`.  This is needed for dealing with `AF_PACKET` sockets when using `recvmsg`.


test/sys/test_socket.rs
```diff
@@ -1462,3 +1462,38 @@ pub fn test_vsock() {
...
+    // in C, sockaddr_* are as if sockaddr_storage was a union
+    let mut storage : sockaddr_storage = unsafe { std::mem::zeroed::<sockaddr_storage>() };
+    let mut ll = unsafe { &mut *(&mut storage as *mut _ as *mut libc::sockaddr_ll) };
+    // NOTE: libc has libc::AF_PACKET as cint, instead of u16
+    ll.sll_family = libc::AF_PACKET as u16;
+    // NOTE: libc has libc::ETH_P_IP as cint, instead of u16
+    ll.sll_protocol = (libc::ETH_P_IP as u16).to_be();
+    ll.sll_ifindex = 1i32;
+    ll.sll_hatype = libc::ARPHRD_ETHER.to_be();
+    ll.sll_pkttype = 0u8; // PACKET_HOST (constants not in libc crate)
```
jsoref, line 1479:
```suggestion
    ll.sll_pkttype = 0u8; // PACKET_HOST (constant not in libc crate)
```

test/sys/test_socket.rs
```diff
@@ -1462,3 +1462,38 @@ pub fn test_vsock() {
...
+    ll.sll_halen = 6;
+    // 00:00:5E:53:00 is reserved for documentation.
+    ll.sll_addr = [0x02, 0x00, 0x5E, 0x00, 0x53, 0x00, 0x00, 0x00];
```
jsoref, line 1482: Are you talking about this:
>      00-00-5E-00-53-00 through 00-00-5E-00-53-FF: assigned for use in
>         documentation.
https://tools.ietf.org/html/rfc7042#section-2.1.1

asomers: That leading `0x02` should be a `0x00`, right?

test/sys/test_socket.rs
isomer: Fixed.

test/sys/test_socket.rs
```diff
@@ -1462,3 +1462,38 @@ pub fn test_vsock() {
...
     thr.join().unwrap();
 }
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[test]
+pub fn test_sockaddr_storage_to_datalink() {
+    use nix::sys::socket::sockaddr_storage;
+    // in C, sockaddr_* are as if sockaddr_storage was a union
+    let mut storage : sockaddr_storage = unsafe { std::mem::zeroed::<sockaddr_storage>() };
+    let mut ll = unsafe { &mut *(&mut storage as *mut _ as *mut libc::sockaddr_ll) };
```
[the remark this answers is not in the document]
isomer: Ah thank you.  I'm still learning how all the unsafe {} rules work.  I've think I've implemented your suggestion correctly.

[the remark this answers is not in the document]
asomers: No, `ll` is still a reference.  You need to do something like this:
```rust
let ll: *mut libc::sockaddr_ll = &mut storage as *mut _ as *mut libc::sockaddr_ll;
unsafe {
    ll.sll_family = ...
}
```

test/sys/test_socket.rs
isomer: Ah, I was very careful about transcribing the array, and screwed up the comment!  Thanks for noticing that. Fixed.

MAC addresses are sent on the wire Least Significant Bit first.  So the first few bits (lowest few bits) of the mac have special meanings.  0x01 means it's a "group" address (used by multicast or broadcast packets), and 0x02 means "Not globally unique."

Flipping 0x00 to 0x02 is probably not necessary here (people know that the documentation prefixes are not globally unique), but it's habit I have when I'm manually assigning addresses.

The 0x00 padding at the end is annoying, but necessary based on the types.

test/sys/test_socket.rs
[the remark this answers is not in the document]
isomer: Ah, okay.  Again, still learning, thank you for your patience.  Fixed.

test/sys/test_socket.rs
```diff
@@ -1462,3 +1462,40 @@ pub fn test_vsock() {
...
+    unsafe {
+        // NOTE: libc has libc::AF_PACKET as cint, instead of u16
+        (*ll).sll_family = libc::AF_PACKET as u16;
+        // NOTE: libc has libc::ETH_P_IP as cint, instead of u16
+        (*ll).sll_protocol = (libc::ETH_P_IP as u16).to_be();
+        (*ll).sll_ifindex = 1i32;
+        (*ll).sll_hatype = libc::ARPHRD_ETHER.to_be();
+        (*ll).sll_pkttype = 0u8; // PACKET_HOST (constant not in libc crate)
+        (*ll).sll_halen = 6;
+        // 00:00:5E:00:53:00 is reserved for documentation.
```
asomers, line 1482: This comment still doesn't match the code.

isomer: Hmm, I'd argue that it's the 00: form that is technically reserved, but yes, you're right this is confusing.  Fixed.

asomers: No, it's not fixed.  What did you even change?  I can't tell, because you force pushed.

isomer: Urgh, my humblest apologies.

I'm still trying to learn rust, and the github workflows, please bear with me.   Last time you wanted everything squashed before commit, which I managed to completely screw up multiple times, so this time I've been trying to keep the commit simplified as I go along so hopefully I don't quite screw it up as badly...

I think this is correct this time.

asomers:
Ok, that looks good now.  You can squash and then I'll merge.

To be clear, my preferred workflow is to push separate commits until review is complete, and then squash once at the end.

PR #1278, opened 2020-07-27
Allow both &[CString] and &[&CStr] for sequence args in unistd

youknowone:
I am new to this project, so I wonder this kind of change is acceptable for the project or not.

I agree taking `&[&CStr]` is the ideal way for the functions. But unfortunately, when users generate `CString`s from `&str`,  they may get `&[CString]` rather than `&[&CStr]`
Unlike single `CString` to `&CStr` argument, there is no easy way to convert `&[CString]` to `&[&CStr]` without creating new vector.

By changing the type from `&[&CStr]` to `&[AsRef<CStr>]`, the args now can take both `&[CString]` and `&[&CStr]` without changing that much from nix.


youknowone:
If you don't mind, I want to change them also take `OsString` and `OsStr` (which conceptionally more fit). I am using nix as abstract layer of unix but using `CStr` for unix-like and using `OsStr` for windows for similar function make a feel awkward.

youknowone:
I fixed `test_unistd.rs` to check both types of input

youknowone:
@asomers would you review this PR again?

PR #1280, opened 2020-07-28
Add Fuchsia support to nix

amanda-tait:
Fuchsia developers need to depend upon nix. This change adds the
necessary conditional compilation attributes, and increments the version
of libc depended upon, both of which allow allow nix to work on Fuchsia.

Change-Id: If8c11f9fb5455438f1e7ca470e3bbdab763fd67c

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ exclude     = [
 ]
 
 [dependencies]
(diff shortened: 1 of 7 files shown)


amanda-tait:
Found a problem in the tests module on Fuchsia while resolving the CI failure. Consequently, this PR isn't yet ready for review, so I'm closing it until I resolve the underlying problem.

PR #1281, opened 2020-08-01
Added clock_gettime, clock_getres, clock_settime, clock_getcpuclockid

### LINKED ISSUE #1275: Add support for clock_gettime/clock_getcpuclockid
I'm interested in adding those, if there are no concerns with that.

xonatius:
Picked up #1100 and added `clock_getcpuclockid` call as well. Credits to @kevinwern for the initial version.

https://www.man7.org/linux/man-pages/man2/clock_gettime.2.html
https://www.man7.org/linux/man-pages/man3/clock_getcpuclockid.3.html

Closes #1275

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -1,0 +1,258 @@
+use crate::sys::time::TimeSpec;
+#[cfg(any(
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "linux",
+    target_os = "android",
+    target_os = "emscripten",
+))]
+use crate::{unistd::Pid, Error};
+use crate::{Errno, Result};
+use libc::{self, clockid_t};
+use std::mem::MaybeUninit;
+
+/// Clock identifier
+///
+/// Newtype pattern around `clockid_t` (which is just alias). It pervents bugs caused by
+/// accidentally passing wrong value.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct ClockId(clockid_t);
+
+impl ClockId {
+    /// Creates `ClockId` from raw `clockid_t`
+    pub fn from_raw(clk_id: clockid_t) -> Self {
+        ClockId(clk_id)
+    }
+
+    /// Returns `ClockId` of a `pid` CPU-time clock
+    #[cfg(any(
+        target_os = "freebsd",
+        target_os = "dragonfly",
+        target_os = "linux",
+        target_os = "android",
+        target_os = "emscripten",
+    ))]
+    pub fn pid_cpu_clock_id(pid: Pid) -> Result<Self> {
+        clock_getcpuclockid(pid)
+    }
+
+    /// Returns resolution of the clock id
+    pub fn res(self) -> Result<TimeSpec> {
+        clock_getres(self)
+    }
+
+    /// Returns the current time on the clock id
+    pub fn now(self) -> Result<TimeSpec> {
+        clock_gettime(self)
+    }
+
+    /// Sets time to `timespec` on the clock id
+    #[cfg(not(any(
+        target_os = "macos",
+        target_os = "ios",
+        all(
+            not(any(target_env = "uclibc", target_env = "newlibc")),
+            any(target_os = "redox", target_os = "hermit",),
+        ),
+    )))]
+    pub fn set_time(self, timespec: TimeSpec) -> Result<()> {
+        clock_settime(self, timespec)
+    }
+
+    /// Gets the raw `clockid_t` wrapped by `self`
+    pub fn as_raw(self) -> clockid_t {
+        self.0
+    }
+
+    #[cfg(any(
+        target_os = "fuchsia",
+        all(
+            not(any(target_env = "uclibc", target_env = "newlib")),
+            any(target_os = "linux", target_os = "android", target_os = "emscripten"),
+        )
+    ))]
+    pub const CLOCK_BOOTTIME: ClockId = ClockId(libc::CLOCK_BOOTTIME);
+    #[cfg(any(
+        target_os = "fuchsia",
+        all(
+            not(any(target_env = "uclibc", target_env = "newlib")),
+            any(target_os = "linux", target_os = "android", target_os = "emscripten")
+        )
+    ))]
+    pub const CLOCK_BOOTTIME_ALARM: ClockId = ClockId(libc::CLOCK_BOOTTIME_ALARM);
+    pub const CLOCK_MONOTONIC: ClockId = ClockId(libc::CLOCK_MONOTONIC);
+    #[cfg(any(
+        target_os = "fuchsia",
+        all(
+            not(any(target_env = "uclibc", target_env = "newlib")),
+            any(target_os = "linux", target_os = "android", target_os = "emscripten")
+        )
+    ))]
+    pub const CLOCK_MONOTONIC_COARSE: ClockId = ClockId(libc::CLOCK_MONOTONIC_COARSE);
+    #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+    pub const CLOCK_MONOTONIC_FAST: ClockId = ClockId(libc::CLOCK_MONOTONIC_FAST);
+    #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+    pub const CLOCK_MONOTONIC_PRECISE: ClockId = ClockId(libc::CLOCK_MONOTONIC_PRECISE);
+    #[cfg(any(
+        target_os = "fuchsia",
+        all(
+            not(any(target_env = "uclibc", target_env = "newlib")),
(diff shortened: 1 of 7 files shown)


xonatius:
@asomers @kamalmarhubi @posborne

Friendly ping for code review, thanks!

src/time.rs
```diff
@@ -0,0 +1,242 @@
...
+impl ClockId {
+    /// Creates `ClockId` from raw `clockid_t`
+    pub fn from_raw(clk_id: clockid_t) -> Self {
+        ClockId(clk_id)
+    }
+
+    /// Returns `ClockId` of a `pid` CPU-time clock
+    #[cfg(target_os = "linux")]
+    pub fn pid_cpu_clock_id(pid: Pid) -> Result<Self> {
+        clock_getcpuclockid(pid)
```
asomers, line 26: This function is supported by FreeBSD, too.

src/time.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /// Returns resolution of the clock id
+    pub fn res(self) -> Result<TimeSpec> {
+        clock_getres(self)
+    }
+
+    /// Returns time on the clock id
+    pub fn time(self) -> Result<TimeSpec> {
```
asomers, line 35: Since this method is basically the same as `std::time::Instant::now`, it would be good if the API could be similar.  Do you agree?  The `time` method could be renamed `now`, and we could add convertability between `TimeSpec` and `Instant`.

src/time.rs
xonatius: Indeed, although it's not in rust-lang/libc bindings for FreeBSD. I'll try to add it and update the PR.

src/time.rs
xonatius: Agreed. Wasn't sure if `now()` makes sense for pid-based clocks, but I guess it sort of does as we get the time on the pid-based clock at the current moment. I'll update the PR.

src/time.rs
xonatius: [Added to more targets](https://github.com/rust-lang/libc/pull/1867) in libc waiting for a [version bump](https://github.com/rust-lang/libc/pull/1879) to update this PR.

src/time.rs
xonatius: Looking more, it seems like `std::time::Instant` is pretty limited in its usage and seems to be designed to just get the current monotonically increasing time, rather than all purpose type to represent an arbitrary point in time (i.e. the only way to construct `std::time::Instant` is to call `Instant::now()`).

The other problem is that I don't think it makes sense to convert the time process has been running on CPU to a particular point in time since unix epoch. Looking on C `timespec` [documentation](https://en.cppreference.com/w/c/chrono/timespec) I think `TimeSpec` is closer to `std::time::Duration` rather than `std::time::Instant` (or even `std::time::SystemTime`), as it represents the interval rather than particular point in time.

If that makes sense, I can add convertability between `TimeSpec` and `Duration`.

asomers: Hm, good point.  I agree.  Duration is more appropriate.

asomers (changes requested):
Could you please fix the test failures?

xonatius:
Looks like with [a recent change](https://github.com/rust-lang/libc/pull/1870) to libc to change some functions from unsafe to safe on linux platform, the usage of those function is now broken across platforms. I'll look into fixing those, probably the best way is to make functions safe on other platforms as well...

asomers:
PR #1294 will fix the unused unsafe warnings.  However, I like your change better, because it's more finely tuned.  Would you consider opening a separate PR just for those?

xonatius:
Looks like that PR is updated to be more finely tuned, so I guess a separate PR is no longer needed. I'll merge and update my changes once that PR is merged.

On a side note, once https://github.com/rust-lang/libc/pull/1891 gets into new release the `unsafe` blocks and `allow(unused_unsafe)` won't be needed anymore.

xonatius:
Clicked the wrong button :/

vi:
> On a side note, once rust-lang/libc#1891 gets into new release the unsafe blocks and allow(unused_unsafe) won't be needed anymore.

`allow(unused_unsafe)` with unsafe blocks can still be beneficial to extend the range of supported `libc` versions.

xonatius:
> `allow(unused_unsafe)` with unsafe blocks can still be beneficial to extend the range of supported `libc` versions.

Sure, but that probably extends only by 1 version, as this change already bumps up the `libc` version to the latest one and can't be built without it.

asomers:
You'll have to rebase; don't merge master into your branch.

xonatius:
Done, let me know if you prefer me to squash changes into a single commit.

src/time.rs
```diff
@@ -0,0 +1,258 @@
...
+use crate::{Errno, Result};
+use libc::{self, clockid_t};
+use std::mem::MaybeUninit;
+
+/// Clock identifier
+///
+/// Newtype pattern around `clockid_t` (which is just alias). It pervents bugs caused by
+/// accidentally passing wrong value.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct ClockId(clockid_t);
```
asomers, line 19: Should this use `libc_enum!`?

src/time.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+impl std::fmt::Display for ClockId {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        std::fmt::Display::fmt(&self.0, f)
+    }
+}
+
+/// Get the resolution of the specified clock, (see
+/// [clock_getres(2)](https://www.man7.org/linux/man-pages/man2/clock_getres.2.html)).
```
asomers, line 206: Better to link to the portable man page. https://pubs.opengroup.org/onlinepubs/7908799/xsh/clock_settime.html

src/time.rs
```diff
@@ -0,0 +1,258 @@
...
+pub fn clock_getres(clock_id: ClockId) -> Result<TimeSpec> {
+    let mut c_time: MaybeUninit<libc::timespec> = MaybeUninit::uninit();
+    let ret = unsafe { libc::clock_getres(clock_id.as_raw(), c_time.as_mut_ptr()) };
+    Errno::result(ret)?;
+    let res = unsafe { c_time.assume_init() };
+    Ok(TimeSpec::from(res))
+}
+
+/// Get the time of the specified clock, (see
+/// [clock_gettime(2)](https://www.man7.org/linux/man-pages/man2/clock_gettime.2.html)).
```
asomers, line 216: Better to link to the portable man page. https://pubs.opengroup.org/onlinepubs/7908799/xsh/clock_settime.html

src/time.rs
```diff
@@ -0,0 +1,258 @@
...
+pub fn clock_gettime(clock_id: ClockId) -> Result<TimeSpec> {
+    let mut c_time: MaybeUninit<libc::timespec> = MaybeUninit::uninit();
+    let ret = unsafe { libc::clock_gettime(clock_id.as_raw(), c_time.as_mut_ptr()) };
+    Errno::result(ret)?;
+    let res = unsafe { c_time.assume_init() };
+    Ok(TimeSpec::from(res))
+}
+
+/// Set the time of the specified clock, (see
+/// [clock_settime(2)](https://www.man7.org/linux/man-pages/man2/clock_settime.2.html)).
```
asomers, line 226: Better to link to the portable man page. https://pubs.opengroup.org/onlinepubs/7908799/xsh/clock_settime.html

src/time.rs
```diff
@@ -0,0 +1,258 @@
...
+        any(target_os = "redox", target_os = "hermit",),
+    ),
+)))]
+pub fn clock_settime(clock_id: ClockId, timespec: TimeSpec) -> Result<()> {
+    let ret = unsafe { libc::clock_settime(clock_id.as_raw(), timespec.as_ref()) };
+    Errno::result(ret).map(drop)
+}
+
+/// Get the clock id of the specified process id, (see
+/// [clock_getcpuclockid(3)](https://www.man7.org/linux/man-pages/man3/clock_getcpuclockid.3.html)).
```
asomers, line 241: Better to link to the portable man page: https://pubs.opengroup.org/onlinepubs/009695399/functions/clock_getcpuclockid.html

asomers (changes requested):

src/time.rs
xonatius: The enum won't work for `clockid_t` returned by `clock_getcpuclockid`, that's why I had to go with struct.

src/time.rs
xonatius: Thanks, done!

src/time.rs
asomers: You're worried that the OS may add a new clockid that Nix doesn't know about?  Ok, that's a fair reason.

asomers:
LGTM!  Just squash the commits now.

xonatius:
Done!

asomers (approved):
bors r+

PR #1284, opened 2020-08-05
Expose SeekData and SeedHole on all Linux targets

Amanieu:
These were previously missing on musl and mips targets because of
missing definitions in the libc crate.


Amanieu:
Musl doesn't define these constants in its headers, which is why they are not defined by the libc crate. See https://github.com/rust-lang/libc/pull/319.

PR #1285, opened 2020-08-05
Add fuchsia support

amanda-tait:
This change adds support for the Fuchsia operating system to the nix
crate. Fuchsia developers have a use case for nix, particularly its safe
interfaces to the recvmsg(2) and sendmsg(2). Adding support requires:
* incrementing the libc dependency to 0.2.74
* conditionally not compiling nix functionality which depends on libc functionality
  that does not exist for Fuchsia

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -116,7 +116,7 @@ task:
         TARGET: x86_64-unknown-linux-musl
         TOOLCHAIN: 1.40.0
   container:
-    image: rust:1.36
+    image: rust:1.40
   setup_script:
     - rustup toolchain install $TOOLCHAIN
     - rustup target add --toolchain $TOOLCHAIN $TARGET
@@ -161,7 +161,7 @@ task:
       env:
         TARGET: x86_64-unknown-netbsd
   container:
-    image: rust:1.36
+    image: rust:1.40
   setup_script:
     - rustup target add $TARGET
   script:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -160,6 +160,10 @@ task:
     - name: NetBSD x86_64
       env:
         TARGET: x86_64-unknown-netbsd
+    - name: Fuchsia x86_64
+      env:
+        TARGET: x86_64-fuchsia
+        CHECK_TESTS: true
   container:
     image: rust:1.40
   setup_script:
@@ -167,6 +171,7 @@ task:
   script:
     - cargo +$TOOLCHAIN check --target $TARGET
     - cargo +$TOOLCHAIN check --target $TARGET --release
+    - 'if [ "$CHECK_TESTS" == true ]; then cargo +$TOOLCHAIN check --all-targets --target $TARGET; fi'
     # TODO: check the tests, too.  The old Travis CI setup didn't do that, so
     # they don't build on all platforms.
   before_cache_script: rm -rf $CARGO_HOME/registry/index
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `mremap` (#[1306](https://github.com/nix-rust/nix/pull/1306))
 
 - Added `personality` (#[1331](https://github.com/nix-rust/nix/pull/1331))
+
+- Added limited Fuchsia support (#[1285](https://github.com/nix-rust/nix/pull/1285))
+
 ### Fixed
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ targets = [
   "x86_64-unknown-openbsd",
   "x86_64-unknown-netbsd",
   "x86_64-unknown-dragonfly",
+  "x86_64-fuchsia",
   "x86_64-unknown-redox"
 ]
 
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Tier 2:
   * x86_64-unknown-netbsd
 
 Tier 3:
+  * x86_64-fuchsia
   * x86_64-unknown-redox
 
 ## Usage
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -20,7 +20,8 @@ cfg_if! {
         }
     } else if #[cfg(any(target_os = "linux",
                         target_os = "redox",
-                        target_os = "dragonfly"))] {
+                        target_os = "dragonfly",
+                        target_os = "fuchsia"))] {
         unsafe fn errno_location() -> *mut c_int {
             libc::__errno_location()
         }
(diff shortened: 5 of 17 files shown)


amanda-tait:
@asomers @kamalmarhubi @posborne 

Friendly bump for code review. Thank you for taking a look.

posborne:
@amanda-tait The change looks reasonable to me, is there any reasonable path for adding fuchsia support to CI.  Without that, I fear it is pretty easy for other contributors to easily introduce regressions.  I do not see CI support (or missed it) within libc either.

tamird:
@posborne CI support is something we'd like to have, but Fuchsia does not yet provide tools for external CI.

We're OK with taking on the burden of fixing regressions.

asomers:
The change looks good, but please add a CHANGELOG entry.

tamird:
@asomers added a CHANGELOG entry and resolved the merge conflict.

tamird:
@asomers I think cross compiling to Fuchsia would be great, do you have a suggestion on how to do that?

tamird:
@asomers I added a pair of commits to address upstream deprecations in libc, and this is now green and includes cross compilation to Fuchsia. Can you take a look?

.travis.yml
```diff
@@ -106,6 +106,16 @@ matrix:
...
         - cargo build --all-targets
 
+    # Fuchsia doesn't have external CI support, so we'll just make sure things compile
+    - language: generic
+      name: fuchsia
+      script:
+        - curl --proto '=https' --tlsv1.2 -sSf --output rustup.sh https://sh.rustup.rs
+        - sh rustup.sh -y --profile=minimal --default-toolchain 1.36.0 --target x86_64-fuchsia
+        - . $HOME/.cargo/env
+        - cargo build --all-targets
```
asomers, line 116: You need to add `--target x86_64-fuchsia` to these two lines.  (you probably copied this from the Redox section, which makes the same mistake; I'll fix that separately).

asomers (changes requested):
Your PR now conflicts with #1329.  But that PR requires your musl fix in order to build.  Could you please split this PR into two?  One for the musl fix, one for Fuchsia, and discard the af_alg_iv parts.  Also, add Fuchsia to Tier 3 in the README.

.travis.yml
tamird: I don't think this is incorrect; the rustup invocation only installs the fuchsia target, so this does the right thing. The same is true for redox.

tamird:
I can make the changes you requested, but I think the approach taken in #1329 is incorrect. As I mentioned in that PR, the deprecation is informational, and the approach taken there is far more brittle without any upside.

tamird:
Add Fuchsia to Tier 3 in the README.

tamird:
@asomers rebased now that #1332 is in.

tamird:
CI is passing.

.travis.yml
asomers: It is incorrect.  I just checked in a fresh environment.  The rustup invocation installs both the native toolchain and the one that you request with `--target`.  So you need to do one of tho things:
* Use `--default-host` instead of `--target`, or
* Add `--target x86_64-fuchsia` to the `cargo build` line, as I suggested.

asomers (changes requested):

tamird:
@asomers I can't seem to reply to your inline comment, but I've made the changes you requested and added Fuchsia to the list added in #1337.

asomers:
Looks like Nix's MSRV is too old for Fuchsia.  You can push a separate commit to raise it to 1.38.0.

tamird:
Done.

asomers:
Why do you need Rust 1.40.0?  What breaks with 1.38.0 ?

tamird:
I explained in the commit message - 1.40.0 or higher is required for doctest support, which is used by a
dependency (remove_dir_all).

asomers:
But Nix doesn't use doctest, and remove_dir_all is not a dependency.  So please explain, what breaks with 1.38.0?  Exact command, please.

tamird:
What had failed was the FreeBSD CI. That run is here: https://cirrus-ci.com/task/6071807595773952.

tamird:
Looks like the dependency is transitive through tempfile:

https://github.com/nix-rust/nix/blob/062bec356dc115fb7c971a3f543c55ac32ef98df/Cargo.toml#L44

and

https://github.com/Stebalien/tempfile/blob/7c099e9a22255b7888d8be49b8118bc0ba2f083c/Cargo.toml#L23

asomers:
But why does it only fail for Fuchsia?  All operating systems use tempfile, and none of them had a problem on 1.36.0.

tamird:
It _doesn't_ just fail for Fuchsia. It fails on all systems - the failure was on FreeBSD.

The cause seems to be that cfg(doctest) was only added in 1.37.0 (https://github.com/rust-lang/rust/pull/62213), which is why this problem didn't occur with the previous MSRV of 1.36.0.

That said, there seem to be additional issues with 1.40.0 (failure to link with Rust) that I'm looking into.

tamird:
OK - sorry for the churn. The linking failures should be addressed now (by having removed the `cargo test` invocation); I've updated the comments in `.travis.yml` to explain why it isn't possible to run `cargo test`, and added a `cargo check --all-targets` invocation to make sure the tests at least pass the compiler (but not the linker).

asomers (approved):
This looks good.  I'm going to optimistically approve it, even though CI hasn't finished yet.  Maybe, just maybe, Travis will burn through its backlog by morning.

bors r+

asomers:
bors retry

tamird:
Had a test failure under linux-musl; hopefully fixed now.

tamird:
@asomers would you mind approving again?

kamalmarhubi:
bors retry

kamalmarhubi:
bors retry

tamird:
Can I get another retry? Finally made it through travis (some unrelated mac failures though)

asomers:
I've never seen that failure on OSX before.  I wonder if it's a new OSX regression.  I've restarted those builds.

tamird:
I've seen it pretty consistently on this PR. I wonder if it's associated
with the MSRV bump.

asomers:
You should be able to fix the CI problems by rebasing now.

tamird:
Done. Looks like there are failures on OSX and musl unrelated to this PR.

tamird:
@asomers ping.

asomers:
Sorry.  The switch to Cirrus has exposed some weaknesses in our tests.  I'm trying to fix them in a different PR.

asomers:
Ok, try rebasing again.

tamird:
Rebased. Still see a failure on OSX, but otherwise green.

tamird:
@asomers ping.

kamalmarhubi:
bors retry

🤞🏼

asomers:
I want to merge #1358 first, since I've learned the root cause of the musl problem.

bors r-

asomers:
Ok, try rebasing again.  Thanks for being patient.

tamird:
Done.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Added `mremap` (#[1306](https://github.com/nix-rust/nix/pull/1306))
 
+### Changed
+- Minimum supported Rust version is now 1.40.0.
+  ([#1285](https://github.com/nix-rust/nix/pull/1285))
```
asomers, line 12: This section is duped.  You can remove it.

.cirrus.yml
```diff
@@ -116,7 +116,7 @@ task:
         TARGET: x86_64-unknown-linux-musl
         TOOLCHAIN: 1.40.0
   container:
-    image: rust:1.36
+    image: rust:1.40
```
asomers, line 119: Oh crap.  I had a search-and-replace fail on my last PR.  At least this instance had no effect.

.cirrus.yml
```diff
@@ -161,7 +161,7 @@ task:
       env:
         TARGET: x86_64-unknown-netbsd
   container:
-    image: rust:1.36
+    image: rust:1.40
```
asomers, line 168: But this instance did have an effect.  Oops.

asomers (changes requested):
When you rebased, your 1.40.0 commit should've failed to apply.  But it didn't, because I missed two places in PR #1358 .  Would you mind fixing your commit message to clarify what the commit actually does?  And fix the CHANGELOG, too.

tamird:
Done.

tamird:
Another flake.

```
---- test_fcntl::linux_android::test_vmsplice stdout ----
thread 'test_fcntl::linux_android::test_vmsplice' panicked at 'assertion failed: `(left == right)`
  left: `6`,
 right: `3`', test/test_fcntl.rs:198:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

asomers:
> Another flake.
> 
> ```
> ---- test_fcntl::linux_android::test_vmsplice stdout ----
> thread 'test_fcntl::linux_android::test_vmsplice' panicked at 'assertion failed: `(left == right)`
>   left: `6`,
>  right: `3`', test/test_fcntl.rs:198:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
> ```

Sigh.  This looks like a kernel bug.  Or possibly a documentation bug.  Is `vmsplice` allowed to write fewer bytes than were requested?  The docs explicitly say that syscalls like write and sendfile may do that, but the man page for vmsplice is silent on the matter.  In any case, it's certainly unrelated to this PR.  For now, I'll try to rerun the check.

tamird:
Thanks, looks like rerunning it worked. Is this good to land?

asomers:
No, not yet.  You still have that duped entry in the changelog.

tamird:
Ah, sorry. It ended up in the second commit somehow. Done.

tamird:
This is green again and I've made all the changes requested. Please have another look.

asomers (approved):
bors r+

tamird:
Is something wedged? Looks like this has been "waiting in queue" for 2 days (and I can't see the bors details page).

asomers:
Cirrus is fine, but Bors is crashing.
https://github.com/bors-ng/bors-ng/issues/1106
Hopefully fixed by #1365 .

tamird:
Looks like that worked. Rebased.

tamird:
@asomers looks like a network flake this time
```
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-1.40.0.toml.sha256' to '/tmp/home/.rustup/tmp/4t0mcegb2twtxqn5_file'
```
Would you mind retrying it and submitting this once more?

Also, I'm not able to edit the PR description, but you might be able to -- could you edit it to reflect the second commit's message?

asomers:
> Also, I'm not able to edit the PR description, but you might be able to -- could you edit it to reflect the second commit's message?

What's wrong with the PR description?

tamird:
The libc version is not modified.

asomers (approved):
bors r+

PR #1147, opened 2019-10-30
Expose the RawFd wrapped by struct PollFd

mlang:

PR #1286, opened 2020-08-06
Add method to retreive file descriptor from PollFd

### LINKED ISSUE #1146: Expose RawFd in impl PollFd
nix::poll::PollFd currently only exposes revents().  However, there is no way to get to the actual fd because libc::pollfd is private.  I think a function like:

impl PollFd {
  pub fn fd(&self) -> RawFd { self.pollfd.fd }
}
or similar would be useful.
As it stands, I dont really know how to figure out which event belongs to which fd.

IntrepidPig:
This fixes #1146.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Added `poll::PollFd::fd` method to expose the underlying file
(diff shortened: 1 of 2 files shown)


IntrepidPig:
Oops, I just noticed this is a duplicate of #1147. Sorry about that.

SteveLauC:
Thanks for your contributing!

That issue has already been fixed, so I am gonna close this PR, I am sorry about this.

Issue #1288, opened 2020-08-08
Netlink `SockProtocol`s


sinkingpoint:
Currently the `SockProtocol` enum: https://github.com/nix-rust/nix/blob/4b6b14a0c2fabb0cdedf415aa0eece2752adc47e/src/sys/socket/mod.rs#L81-L95 is rather scarce. I'm looking into doing some Netlink work with Rust and the man page defines a number of protocols not currently supported by this enum (https://www.man7.org/linux/man-pages/man7/netlink.7.html). As far as I can tell this should just be a matter of adding them to it.

PR #1289, opened 2020-08-08
Add Netlink Protocol Families to SockProtocol enum

sinkingpoint:
Currently the SockProtocol enum is rather scarce. This commit adds the Netlink protocols
defined in netlink(7) to the SockProtocol enum allowing us to use the Nix socket library for
more in-depth Netlink work. 

Because of the way libc sockets work, this is a bit more annoying than it has any right to be - the Socket Family and the Socket protocol are intertwined in that a given c_int protocol can represent different things depending on the family. This means that as we add values to this enum we're going to get collisions (Indeed, I've left out a few values that collided with the existing TCP/UDP Protocols). I'm not sure how we would want to handle this going forward - maybe different Protocol enums for different Familys, with `Into<SockProtocol>` methods?


sinkingpoint:
Addresses #1288

asomers:
This seems simple enough.  And yes, `libc_enum` is certainly a better way to do this.  Could you add a CHANGELOG entry?

sinkingpoint:
Done @asomers - I figure changing to libc_enum involves changing the enum members names (Making them conform with libc naming standards) so would be a breaking change. Better to leave it out of this PR I think and address in a major release at some point. Thanks!

PR #1290, opened 2020-08-11
use subordinate terminal device for termios calls

jclulow:
The pseudo-terminal device handling tests in some places make
tcgetattr(3C) and tcsetattr(3C) calls using the control/manager file
descriptor rather than the subordinate terminal descriptor.  This works
on some systems, but not all; others such as illumos (and presumably
Solaris) are more strict and require the termios requests be made
against the terminal descriptor only.

### DIFF
--- a/test/sys/test_termios.rs
+++ b/test/sys/test_termios.rs
@@ -22,7 +22,7 @@ fn test_tcgetattr_pty() {
     let _m = crate::PTSNAME_MTX.lock().expect("Mutex got poisoned by another test");
 
     let pty = openpty(None, None).expect("openpty failed");
-    assert!(termios::tcgetattr(pty.master).is_ok());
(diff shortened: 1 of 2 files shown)


jclulow:
In working on illumos support I hit this.  I figured I'd stage it as a separate change to reduce the size of that eventual change.

This passes `cargo test --all` on my Linux system with stable Rust.

asomers:
@Susurrus could you review this?  You're the termios expert.

jclulow:
Hi!  Is there something I can do to help this along?  Thanks.

jclulow:
Hello again.   Is there something I can do to help this along? Thanks.

asomers (approved):
bors r+

jclulow:
Thank you!

PR #1292, opened 2020-08-22
Support `TCP_MAXSEG` TCP Maximum Segment Size socket options

eaon:
Found myself in a situation where I needed to advertise a specific MSS. Thought this crate is the appropriate place for this feature (especially since the `TCP_MAXSEG` const was exposed through it before), so here goes.

Actual change is of course tiny but it comes with a test. Happy to make changes, especially since I need to do more research on which platforms this works.

### DIFF
--- a/test/sys/test_sockopt.rs
+++ b/test/sys/test_sockopt.rs
@@ -70,6 +70,57 @@ fn test_so_buf() {
     assert!(actual >= bufsize);
 }
 
+#[test]
+fn test_so_tcp_maxseg() {
+    use std::net::SocketAddr;
+    use std::str::FromStr;
+    use nix::sys::socket::{accept, bind, connect, listen, InetAddr, SockAddr};
+    use nix::unistd::{close, write};
+
+    let std_sa = SocketAddr::from_str("127.0.0.1:4001").unwrap();
+    let inet_addr = InetAddr::from_std(&std_sa);
+    let sock_addr = SockAddr::new_inet(inet_addr);
+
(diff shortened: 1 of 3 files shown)


test/sys/test_sockopt.rs
```diff
@@ -70,6 +70,57 @@ fn test_so_buf() {
...
+                .unwrap();
+    bind(rsock, &sock_addr).unwrap();
+    listen(rsock, 10).unwrap();
+    let initial = getsockopt(rsock, sockopt::TcpMaxSeg).unwrap();
+    // Initial MSS is expected to be 536 (https://tools.ietf.org/html/rfc879#section-1) but some
+    // platforms keep it even lower. This might fail if you've tuned your initial MSS to be larger
+    // than 700
+    cfg_if! {
+        if #[cfg(any(target_os = "android", target_os = "linux"))] {
+            let segsize: u32 = thread_rng().gen_range(700..1300);
```
asomers, line 94: Sorry for my inattentiveness!  I made a comment when you first opened the PR, but forgot to hit "submit review".

Random numbers in tests are usually bad, because they lead to non-reproducible failures. The exception is for stress tests, but this is not one.  Please use a deterministic segsize instead.

asomers (changes requested):

test/sys/test_sockopt.rs
eaon: Makes sense, I chose a number that's (at least currently) definitely not anywhere near the MSS defaults of Linux/Android. Hope this works for you! And thanks for your work on this great crate!

asomers (approved):
bors r+

PR #1294, opened 2020-09-12
Add allow(unused_unsafe) to some functions.

vi:


vi:
Or shall a bunch of `unsafe{}` blocks be removed instead? Will it cause incompatibilities for users with legacy rustc or deps?

src/lib.rs
```diff
@@ -11,6 +11,7 @@
 #![cfg_attr(test, deny(warnings))]
 #![recursion_limit = "500"]
 #![deny(unused)]
+#![allow(unused_unsafe)]
```
RalfJung, line 14: Where are these unused unsafe blocks? Can't you add the lint more locally?

vi: I expect it to be [this](https://github.com/rust-lang/libc/pull/1870).

Shall it be fixed just for for this use case or also for possible future unsafe-to-safe conversions?

Shall it be fixed by just removing `unsafe{}` or by allowing `unused_unsafe` selectively?

asomers: Please do it selectively, just where we need it.  And BTW we'll have to make a 0.18.1 release just for this change.

asomers (changes requested):

src/lib.rs
vi: Why a point release? Shouldn't lint caps hide those errors when nix is used as a crates.io dependency?

vi:
Is this pull request still needed in face of #1281 which supposedly also includes similar changes?

src/lib.rs
asomers: What are "lint caps"?

RalfJung: Cargo builds dependencies by calling `rustc --cap-lints allow`. This means lints are turned off when building dependencies.

vi:
@asomers What more to be done for this?

asomers (approved):
LGTM!

bors r+

PR #1296, opened 2020-09-17
DragonFlyBSD: use __errno_location now provided by the libc crate

cmusser:
Dragonfly recently (in Dfly 5.8) added an `__errno_location()` function to make it easy to access the thread-local `errno` variable. This has been exposed in rust-libc as of 0.2.77. This PR uses that functionality instead of the locally compiled C language shim, which the PR removes. One issue is backwards compatibilty. It requires 0.2.77 libc and DragonFly 5.8. Not sure how to gracefully handle older DragonFly versions, although I'm also not sure doing so is worth it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ exclude     = [
 ]
 
 [dependencies]
-libc = { version = "0.2.73", features = [ "extra_traits" ] }
+libc = { version = "0.2.77", features = [ "extra_traits" ] }
 bitflags = "1.1"
 cfg-if = "0.1.10"
 
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,0 @@
-#[cfg(target_os = "dragonfly")]
-fn main() {
-    cc::Build::new()
-        .file("src/errno_dragonfly.c")
-        .compile("liberrno_dragonfly.a");
-}
-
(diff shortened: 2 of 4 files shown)


asomers:
LGTM, but you'll have to rebase after #1294 merges.

cmusser:
OK I rebased and pushed my branch. The tests pass, but I want to make sure that my merge didn't disrupt anything it shouldn't have. In the diff, a few functions in ` src/sys/wait.rs` and ` src/sys/quota.rs` now are prefixed with the `allow_unused_unsafe`
 annotation.

asomers:
It looks like you merged master into your branch, rather than rebased your branch on top of master.  That's why the diff looks screwed up.  If you rebase on top of master there won't be that problem.

cmusser:
I rebased on top of master and the diff looks sensible now. The tests fail for aarch64-unknown-linux-gnu and x86_64-unknown-linux-gnu. There may be other pending commits that have to be merged to resolve those.

asomers:
The failure fo x86_64-unknown-linux-gnu is an old one.  I think it's a bug in glibc itself, not in Nix.  The failure for aarch64 is new; I haven't seen it before.  It also doesn't make much sense.  It could be a bug in QEMU.  We've had those before.

asomers (approved):
bors r+

PR #1297, opened 2020-09-17
Add mkdtemp

DBLouis:


src/unistd.rs
```diff
@@ -1754,6 +1754,40 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+/// use nix::unistd;
+///
+/// match unistd::mkdtemp("/tmp/tempdir_XXXXXX") {
+///     Ok(_path) => {
+///         // do something with directory
+///     }
+///     Err(e) => panic!("mkdtemp failed: {}", e)
+/// };
+/// ```
+#[inline]
```
asomers, line 1777: You should remove the `#[inline]`

src/unistd.rs
```diff
@@ -1754,6 +1754,40 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
...
+///
+/// * `template`: a path whose 6 rightmost characters must be X, e.g. `/tmp/tmpdir_XXXXXX`
+/// * returns: filename
+///
+/// Err is returned either if no temporary filename could be created or the template doesn't
+/// end with XXXXXX
+///
+/// See also [mkstemp(2)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mkdtemp.html)
+///
+/// ```rust
```
asomers, line 1767: The "rust" part is unnecessary.  Rustdoc is smart enough to figure that out ;)

src/unistd.rs
```diff
@@ -1754,6 +1754,40 @@ pub fn mkstemp<P: ?Sized + NixPath>(template: &P) -> Result<(RawFd, PathBuf)> {
     Ok((fd, PathBuf::from(pathname)))
 }
 
+/// Creates a directory which persists even after process termination
+///
+/// * `template`: a path whose 6 rightmost characters must be X, e.g. `/tmp/tmpdir_XXXXXX`
```
asomers, line 1759: Why 6?  mkdtemp(3) imposes no such limitation.

asomers (changes requested):
Don't forget to add a CHANGELOG entry

src/unistd.rs
DBLouis: It says in the manual "at least 6" like mkstemp

src/unistd.rs
DBLouis: I did the same as for the mkstemp function

src/unistd.rs
DBLouis: Same as for mkstemp, I just tried to match it. If you think this and and the inline marker are useless, maybe they should all be removed in a separate commit

src/unistd.rs
asomers: Surprising.  I didn't realize that was part of POSIX.  But FreeBSD, NetBSD, and OSX all allow fewer than six X's, so you should still change the docs.

src/unistd.rs
asomers: Yes, the should be removed.  But don't introduce new ones.

src/unistd.rs
DBLouis: OpenBSD [docs](https://man.openbsd.org/mkstemp) says also says "at least six". The FreeBSD one (and OSX) is not very clear but still gives 6 Xs as an example. Are you sure this is not a minimum in all implementations? Moreover the docs for mkstemp is the same.

asomers: Yep, there's definitely no requirement for a minimum number of Xs'.

DBLouis: What should I put in the doc then?

asomers: I would just say "a path whose rightmost characters contain some number of 'Xs'.  The 'Xs' will be replaced with a unique alphanumeric combination."  And in the next paragraph: "Err is returned is no temporary file could be created or if `path` had insufficient 'Xs'".

asomers:
Still needs a CHANGELOG entry.

SteveLauC:
Gentle ping on the author @DBLouis, I am sorry about the late response, would you like to finish this PR, just need to rebase your branch, discard the changes made to the `CHANGELOG.md` file and add a new file under the `changelog` directory, then we can merge it:)

> file name would be `1297.added.md`

```md
 Added `mkdtemp` wrapper
```

SteveLauC (approved):
Thanks!

bjackman:
I noticed that the new `mkdtemp` is behind the `feature` feature. Is that indended? That feature already existed so seems like it may have been deliberate, but I also can't see what it's intended to mean. 

I'm a rust noob, in case that helps contextualise the question.

PR #1190, opened 2020-03-06
Add getrlimit(2) and setrlimit(2)

sushisharkjl:
This is a continuation of #879. I've added comments and removed usage of the now-deprecated std::mem::uninitialized() function.

PR #1302, opened 2020-10-03
Add getrlimit and setrlimit

LMJW:
This PR is based on the previous PR https://github.com/nix-rust/nix/pull/1190, which has not been updated for a very long time.

I have fixed the segfault and compilation error of the original PR and rebased the changes to the latest master branch.

Please let me know if anything need to been changed.

### DIFF
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -1,0 +1,195 @@
+//! Configure the process resource limits.
+use cfg_if::cfg_if;
+
+use crate::errno::Errno;
+use crate::Result;
+pub use libc::rlim_t;
+use std::mem;
+
+cfg_if! {
+    if #[cfg(all(target_os = "linux", target_env = "gnu"))]{
+        use libc::{__rlimit_resource_t, rlimit, RLIM_INFINITY};
+    }else if #[cfg(any(
+        target_os = "freebsd",
+        target_os = "openbsd",
+        target_os = "netbsd",
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "android",
+        target_os = "dragonfly",
+        all(target_os = "linux", not(target_env = "gnu"))
+    ))]{
+        use libc::{c_int, rlimit, RLIM_INFINITY};
+    }
+}
+
+libc_enum! {
+    /// The Resource enum is platform dependent. Check different platform
+    /// manuals for more details. Some platform links has been provided for
+    /// earier reference (non-exhaustive).
+    ///
+    /// * [Linux](https://man7.org/linux/man-pages/man2/getrlimit.2.html)
+    /// * [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=setrlimit)
+
+    // linux-gnu uses u_int as resource enum, which is implemented in libc as
+    // well.
+    //
+    // https://gcc.gnu.org/legacy-ml/gcc/2015-08/msg00441.html
+    // https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/gnu/mod.rs
+    #[cfg_attr(all(target_os = "linux", target_env = "gnu"), repr(u32))]
+    #[cfg_attr(any(
+            target_os = "freebsd",
+            target_os = "openbsd",
+            target_os = "netbsd",
+            target_os = "macos",
+            target_os = "ios",
+            target_os = "android",
+            target_os = "dragonfly",
+            all(target_os = "linux", not(target_env = "gnu"))
+        ), repr(i32))]
+    #[non_exhaustive]
+    pub enum Resource {
+        #[cfg(not(any(target_os = "netbsd", target_os = "freebsd")))]
+        RLIMIT_AS,
+        RLIMIT_CORE,
+        RLIMIT_CPU,
+        RLIMIT_DATA,
+        RLIMIT_FSIZE,
+        RLIMIT_NOFILE,
+        RLIMIT_STACK,
+
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_KQUEUES,
+
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        RLIMIT_LOCKS,
+
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "openbsd", target_os = "linux"))]
+        RLIMIT_MEMLOCK,
+
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        RLIMIT_MSGQUEUE,
+
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        RLIMIT_NICE,
+
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "openbsd", target_os = "linux"))]
+        RLIMIT_NPROC,
+
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_NPTS,
+
+        #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "openbsd", target_os = "linux"))]
+        RLIMIT_RSS,
+
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        RLIMIT_RTPRIO,
+
+        #[cfg(any(target_os = "linux"))]
+        RLIMIT_RTTIME,
+
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        RLIMIT_SIGPENDING,
+
+        #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+        RLIMIT_SBSIZE,
+
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_SWAP,
+
+        #[cfg(target_os = "freebsd")]
+        RLIMIT_VMEM,
+    }
+}
+
+/// Get the current processes resource limits
+///
+/// A value of `None` indicates the value equals to `RLIM_INFINITY` which means
+/// there is no limit.
+///
+/// # Parameters
+///
+/// * `resource`: The [`Resource`] that we want to get the limits of.
+///
+/// # Examples
+///
+/// ```
+/// # use nix::sys::resource::{getrlimit, Resource};
+///
+/// let (soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+/// println!("current soft_limit: {:?}", soft_limit);
+/// println!("current hard_limit: {:?}", hard_limit);
+/// ```
+///
+/// # References
+///
+/// [getrlimit(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html#tag_16_215)
+///
+/// [`Resource`]: enum.Resource.html
+pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)> {
+    let mut old_rlim = mem::MaybeUninit::<rlimit>::uninit();
+
+    cfg_if! {
+        if #[cfg(all(target_os = "linux", target_env = "gnu"))]{
+            let res = unsafe { libc::getrlimit(resource as __rlimit_resource_t, old_rlim.as_mut_ptr()) };
+        }else{
+            let res = unsafe { libc::getrlimit(resource as c_int, old_rlim.as_mut_ptr()) };
+        }
+    }
+
+    Errno::result(res).map(|_| {
+        let rlimit { rlim_cur, rlim_max } = unsafe { old_rlim.assume_init() };
+        (Some(rlim_cur), Some(rlim_max))
+    })
+}
+
+/// Set the current processes resource limits
+///
+/// # Parameters
+///
+/// * `resource`: The [`Resource`] that we want to set the limits of.
+/// * `soft_limit`: The value that the kernel enforces for the corresponding
+///   resource. Note: `None` input will be replaced by constant `RLIM_INFINITY`.
+/// * `hard_limit`: The ceiling for the soft limit. Must be lower or equal to
+///   the current hard limit for non-root users. Note: `None` input will be
+///   replaced by constant `RLIM_INFINITY`.
+///
+/// > Note: for some os (linux_gnu), setting hard_limit to `RLIM_INFINITY` can
+/// > results `EPERM` Error. So you will need to set the number explicitly.
+///
+/// # Examples
+///
+/// ```
+/// # use nix::sys::resource::{setrlimit, Resource};
+///
+/// let soft_limit = Some(1024);
+/// let hard_limit = Some(1048576);
+/// setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit).unwrap();
+/// ```
+///
+/// # References
+///
+/// [setrlimit(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html#tag_16_215)
+///
+/// [`Resource`]: enum.Resource.html
+///
+/// Note: `setrlimit` provides a safe wrapper to libc's `setrlimit`.
+pub fn setrlimit(
+    resource: Resource,
+    soft_limit: Option<rlim_t>,
+    hard_limit: Option<rlim_t>,
+) -> Result<()> {
+    let new_rlim = rlimit {
+        rlim_cur: soft_limit.unwrap_or(RLIM_INFINITY),
+        rlim_max: hard_limit.unwrap_or(RLIM_INFINITY),
+    };
+    cfg_if! {
+        if #[cfg(all(target_os = "linux", target_env = "gnu"))]{
+            let res = unsafe { libc::setrlimit(resource as __rlimit_resource_t, &new_rlim as *const rlimit) };
+        }else{
+            let res = unsafe { libc::setrlimit(resource as c_int, &new_rlim as *const rlimit) };
+        }
+    }
+
+    Errno::result(res).map(drop)
+}
--- a/test/test.rs
+++ b/test/test.rs
@@ -24,6 +24,7 @@ mod test_mq;
 #[cfg(not(target_os = "redox"))]
 mod test_net;
 mod test_nix_path;
+mod test_resource;
 mod test_poll;
 #[cfg(not(any(target_os = "redox", target_os = "fuchsia")))]
 mod test_pty;
--- a/test/test_resource.rs
+++ b/test/test_resource.rs
@@ -1,0 +1,23 @@
+#[cfg(not(any(target_os = "redox", target_os = "fuchsia", target_os = "illumos")))]
+use nix::sys::resource::{getrlimit, setrlimit, Resource};
+
+/// Tests the RLIMIT_NOFILE functionality of getrlimit(), where the resource RLIMIT_NOFILE refers
+/// to the maximum file descriptor number that can be opened by the process (aka the maximum number
+/// of file descriptors that the process can open, since Linux 4.5).
+///
+/// We first fetch the existing file descriptor maximum values using getrlimit(), then edit the
+/// soft limit to make sure it has a new and distinct value to the hard limit. We then setrlimit()
+/// to put the new soft limit in effect, and then getrlimit() once more to ensure the limits have
+/// been updated.
+#[test]
+#[cfg(not(any(target_os = "redox", target_os = "fuchsia", target_os = "illumos")))]
+pub fn test_resource_limits_nofile() {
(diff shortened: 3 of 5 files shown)


asomers:
You need to fix the test failures.  First of all, `use libc::self` is unnecessary in Rust 2018.  Just remove it.  I don't know if that's related to the other test failures.

LMJW:
> You need to fix the test failures. First of all, `use libc::self` is unnecessary in Rust 2018. Just remove it. I don't know if that's related to the other test failures.

LMJW:
Clicked the wrong button... Will fix the issues.

LMJW:
Fixed the test errors. Please review. Thank you @asomers

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+/// * `resource`: The [`Resource`] that we want to set the limits of.
+/// * `soft_limit`: The value that the kernel enforces for the corresponding
+///   resource. Note: `None` input will be replaced by constant `RLIM_INFINITY`.
+/// * `hard_limit`: The ceiling for the soft limit. Must be lower or equal to
+///   the current hard limit for non-root users. Note: `None` input will be
+///   replaced by constant `RLIM_INFINITY`.
+///
+/// # Examples
+///
+/// ```no_run
```
asomers, line 180: Why `no_run`?  Doc tests each get their own process, so this shouldn't interfere with other tests.

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+    let new_rlim = rlimit {
+        rlim_cur: soft_limit.unwrap_or(RLIM_INFINITY),
+        rlim_max: hard_limit.unwrap_or(RLIM_INFINITY),
+    };
+    cfg_if! {
+        if #[cfg(all(target_os = "linux", target_env = "gnu"))]{
+            // the below implementation is mimicing the similar implementation in golang
+            // https://go-review.googlesource.com/c/sys/+/230478/2/unix/syscall_linux_arm64.go#176
+            // seems for some of the architectures, we prefer to use prlimit instead of {g,s}etrlimit
+            let res = unsafe { libc::prlimit(0, resource as __rlimit_resource_t, &new_rlim as *const _, std::ptr::null_mut()) };
```
asomers, line 207: I think this is just GoLang being silly.  Go religiously avoids using libc at all, prefering to make syscalls directly.  Rust doesn't do that.  You should just use setrlimit.

test/test.rs
```diff
@@ -41,12 +41,10 @@ macro_rules! skip_if_jailed {
     ($name:expr) => {
         use ::sysctl::CtlValue;
 
-        if let CtlValue::Int(1) = ::sysctl::value("security.jail.jailed")
-            .unwrap()
-        {
+        if let CtlValue::Int(1) = ::sysctl::value("security.jail.jailed").unwrap() {
```
asomers, line 44: Please don't mix formatting changes with content changes.

asomers (changes requested):
Rather than define `Resource` twice for different platforms, which doesn't even handle all platform-specificities, I think you should define `Resource` only once, and gate each enum item with an appropriate `#[cfg(...)]`.

I haven't reviewed the tests yet.  ENOTIME.

test/test.rs
LMJW: I see. I will revert the commit that changes the style and will keep the format unchanged.

src/sys/resource.rs
LMJW: The reason why I change to this it because in some of the architecture, using `setrlimit` in `libc` fails for unit test of set `RLIMIT_STACK` (test `test_resource_limits_stack`). To make all the architecture working, I did some research and found this implementation in golang. The failed architecture was `aarch64` and `arm` (see the [failure build](https://travis-ci.com/github/LMJW/nix/builds/187996674)). I am not quite sure why `setrlimit` is not working for some architectures. But using `prlimit` seems working for all the architectures. That's why I choose this implementation.

src/sys/resource.rs
LMJW: That's what was written in previous PR. I will remove this `no_run`

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+/// [`Resource`]: enum.Resource.html
+pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)> {
+    let mut old_rlim = rlimit {
+        rlim_cur: 0,
+        rlim_max: 0,
+    };
+
+    cfg_if! {
+        if #[cfg(all(target_os = "linux", target_env = "gnu"))]{
+            let res = unsafe { libc::getrlimit(resource as __rlimit_resource_t, &mut old_rlim) };
```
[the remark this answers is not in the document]
LMJW: Thank you for pointing this out @asomers . I think this issue originated in [here](https://gcc.gnu.org/legacy-ml/gcc/2015-08/msg00441.html) (at least the resource I can find). So for linux/gnu `libc` bindings, it will use `u32` instead of `i32`. This is implemented in [libc](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/gnu/mod.rs) as well.  Other linux env like musl will still remains as `i32` and that's why I used `cfg_if!`

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }else if #[cfg(any(
+        target_os = "freebsd",
+        target_os = "openbsd",
+        target_os = "netbsd",
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "android",
+        target_os = "dragonfly",
+        target_os = "bitrig",
```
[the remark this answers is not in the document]
LMJW: Will do in the new revision

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+        target_os = "linux", // target_env != "gnu"
+    ))]{
+        use libc::{c_int, rlimit, RLIM_INFINITY};
+
+        libc_enum! {
+            #[repr(i32)]
+            pub enum Resource {
+                /// See detail of each Resource https://man7.org/linux/man-pages/man2/getrlimit.2.html
```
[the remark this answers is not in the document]
LMJW: Hmm. Good question. The original idea was because I think the man page was a good reference that most are similar to this section. And it is easy to find(apologize for my lazyness). But I totoally agree that you are right. This link will be removed since it is for linux.

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+/// # use nix::sys::resource::{getrlimit, Resource};
+///
+/// let (soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+/// println!("current soft_limit: {:?}", soft_limit);
+/// println!("current hard_limit: {:?}", hard_limit);
+/// ```
+///
+/// # References
+///
+/// [getrlimit(2)](https://linux.die.net/man/2/getrlimit)
```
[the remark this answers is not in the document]
LMJW: > Rather than define `Resource` twice for different platforms, which doesn't even handle all platform-specificities, I think you should define `Resource` only once, and gate each enum item with an appropriate `#[cfg(...)]`.
> 
> I haven't reviewed the tests yet. ENOTIME.

Thank you for the suggestion. But I am not sure when enum can be both `#[repr(u32)]` and `#[repr(i32)]`, how could I do this. (The reason why we have both `u32` and `i32` is explained in one of comments below) The closest case I can think of is using something like this:

```rust
libc_enum!{
    pub enum ResourceDemo {
        #[cfg(all(target_os = "linux", target_env = "gnu"))] #[repr(u32)]
        #[cfg(any(
            target_os = "freebsd",
            target_os = "openbsd",
            target_os = "netbsd",
            target_os = "macos",
            target_os = "ios",
            target_os = "android",
            target_os = "dragonfly",
            target_os = "linux"))] #[repr(i32)]
        RLIMIT_AS,

        #[cfg(all(target_os = "linux", target_env = "gnu"))] #[repr(u32)]
        #[cfg(any(
            target_os = "freebsd",
            target_os = "openbsd",
            target_os = "netbsd",
            target_os = "macos",
            target_os = "ios",
            target_os = "android",
            target_os = "dragonfly",
            target_os = "linux"))] #[repr(i32)]
        RLIMIT_CORE,
        RLIMIT_CPU,
        RLIMIT_DATA,
        RLIMIT_FSIZE
    }
}
```
But this will have compile error for linux-musl series toolchain as both cfg will be compiled thus causing the compile error. The idea I have came out is using `cfg_if`, but if there is better ways to do this. I am happy to adopt. Thank you very much.

LMJW:
> Rather than define `Resource` twice for different platforms, which doesn't even handle all platform-specificities, I think you should define `Resource` only once, and gate each enum item with an appropriate `#[cfg(...)]`.
> 
> I haven't reviewed the tests yet. ENOTIME.

Thank you for the suggestion. But I am not sure when enum can be both `#[repr(u32)]` and `#[repr(i32)]`, how could I do this. (The reason why we have both `u32` and `i32` is explained in one of comments above) The closest case I can think of is using something like this:

```rust
libc_enum!{
    pub enum ResourceDemo {
        #[cfg(all(target_os = "linux", target_env = "gnu"))] #[repr(u32)]
        #[cfg(any(
            target_os = "freebsd",
            target_os = "openbsd",
            target_os = "netbsd",
            target_os = "macos",
            target_os = "ios",
            target_os = "android",
            target_os = "dragonfly",
            target_os = "linux"))] #[repr(i32)]
        RLIMIT_AS,

        #[cfg(all(target_os = "linux", target_env = "gnu"))] #[repr(u32)]
        #[cfg(any(
            target_os = "freebsd",
            target_os = "openbsd",
            target_os = "netbsd",
            target_os = "macos",
            target_os = "ios",
            target_os = "android",
            target_os = "dragonfly",
            target_os = "linux"))] #[repr(i32)]
        RLIMIT_CORE,
        RLIMIT_CPU,
        RLIMIT_DATA,
        RLIMIT_FSIZE
    }
}
```
But this will have compile error for linux-musl series toolchain as both cfg will be compiled thus causing the compile error. The idea I have came out is using `cfg_if`, but if there is better ways to do this. I am happy to adopt. Thank you very much.

src/sys/resource.rs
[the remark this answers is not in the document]
LMJW: Will change this in new revision

src/sys/resource.rs
```diff
@@ -0,0 +1,230 @@
...
+
+pub use libc::rlim_t;
+
+cfg_if! {
+    if #[cfg(all(target_os = "linux", target_env = "gnu"))]{
+        use libc::{__rlimit_resource_t, rlimit, RLIM_INFINITY};
+        libc_enum!{
+            #[repr(u32)]
+            pub enum Resource {
+                /// See detail of each Resource https://man7.org/linux/man-pages/man2/getrlimit.2.html
```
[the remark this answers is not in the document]
LMJW: will change this in new revision

asomers:
> Thank you for the suggestion. But I am not sure when enum can be both `#[repr(u32)]` and `#[repr(i32)]`, how could I do this. 
You can do it like this:
```rust
#[cfg_attr(any(target_os="linux", ...), repr(u32))]
#[cfg_attr(not(any(target_os="linux", ...), repr(u32)))]
```

LMJW:
Thanks for given me suggestions. It is very helpful for me. The new PR will be pushed out shortly. However, there are few concerns about the API of the current `setrlimit` and `getrlimit` function.

LMJW:
Also, I feel like the previous implementation does not have enough test. But I am not sure if it is necessary to test each flag(eg `RLIMIT_AS`, `RLIMIT_CORE` ...)

LMJW:
Looks like the test failure is a fake failure. Can we rerun it?

```
install.sh: Installing to: /home/travis/.cargo/bin
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

asomers:
> Thanks for given me suggestions. It is very helpful for. The new PR will be pushed out shortly. However, I have few doubts about the API of the current `setrlimit` and `getrlimit` function.
> 
> The existing api for `getrlimit` and `setrlimit` is like this
> 
> ```rust
> pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>;
> pub fn setrlimit(resource: Resource, soft_limit: Option<rlim_t>, hard_limit: Option<rlim_t>) -> Result<()>;
> ```
> 
> My current implementation is inherit from previous implementation, which treate `None` as `RLIM_INFINITY`. However, for some os environment, like `linux_gnu`, use `setrlimit(Resource::RLIMIT_NOFILE, Some(1024), None)` will results an `EPERM` error whereas other os environment will treat it is okay (see [build](https://travis-ci.com/github/LMJW/nix/builds/188837910)). The `EPERM` error is due to `None` converted to `RLIM_INFINITY` and it is bigger than the default value`1048576` , causing kernel to throw `EPERM` error.
> 
> The question I have is: should we treat `None` as `RLIM_INFINITY` or we should treate `None` as not change this value. If we decide `None` as not change the current value, we will need to get the orginal value first, which means we will need to call `getrlimit` in `setrlimit` method.
> 
> I feel the later case is more natural for me. But it might be my personal preference.

I agree that treating `None` as "no change" would be most natural. That would be similar to how `chown(2)` treats an `owner` or `group` of `-1`.  HOWEVER, that is not the API that `settlimit` and `getrlimit` provide.  There is no race-free way to use `setrlimit` to change one limit but not the other.  The goal of Nix is to provide idiomatically Rusty access to libc, but nothing more.  So I don't think we should try to provide an interface like `chown` here.

LMJW:
Hi @asomers. I think I have addressed your review comments. Please let me know if you feel like further improvements. 

Thank you.

src/sys/resource.rs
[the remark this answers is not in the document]
asomers: Ok, so it's a Linux man page bug.

src/sys/resource.rs
asomers: I'm not satisfied.  Nix's goal is to make the system's C library Rusty.  Cross-platform portability and platform-specific workarounds are beyond our remit.  We've gotten in trouble in the past by trying to add this kind of compatibility code.  Please just use setrlimit alone.

asomers (changes requested):

Nugine:
https://github.com/rust-lang/libc/blob/bc30283752bae5918f40e6d63f858203c81303c7/src/unix/bsd/apple/mod.rs#L1914

```rust
pub const RLIMIT_RSS: ::c_int = RLIMIT_AS;
```

`RLIMIT_RSS` is same with `RLIMIT_AS` on macos and ios. It means that the two variants will conflict when targeting apple.

Is it correct to select only one of them?

LMJW:
> https://github.com/rust-lang/libc/blob/bc30283752bae5918f40e6d63f858203c81303c7/src/unix/bsd/apple/mod.rs#L1914
> 
> ```rust
> pub const RLIMIT_RSS: ::c_int = RLIMIT_AS;
> ```
> 
> `RLIMIT_RSS` is same with `RLIMIT_AS` on macos and ios. It means that the two variants will conflict when targeting apple.
> 
> Is it correct to select only one of them?

The reason why I select one of them is because they have the same value. If I select two of them, there will be a compile error like the following:

> error[E0425]: cannot find value `RLIMIT_AS` in module `libc`
>   --> src/sys/resource.rs:53:17
>    |
> 53 |                 RLIMIT_AS,
>    |                 ^^^^^^^^^
> help: a constant with a similar name exists
>    |
> 53 |                 RLIMIT_RSS,
>    |                 ^^^^^^^^^^
> help: possible candidate is found in another module, you can import it into scope
>    |
> 2  | use crate::sys::resource::Resource::RLIMIT_AS;
>    |
> error[E0081]: discriminant value `0` already exists
>    --> src/macros.rs:164:26
>     |
> 164 |                   $entry = libc::$entry,
>     |                            ^^^^
>     |                            |
>     |                            first use of `0`
>     |                            enum already has `0`
>     | 
>    ::: src/sys/resource.rs:48:9
>     |
> 48  | /         libc_enum! {
> 49  | |             #[repr(i32)]
> 50  | |             pub enum Resource {
> 51  | |                 /// See detail of each Resource https://man7.org/linux/man-pages/man2/getrlimit.2.html
> ...   |
> 103 | |             }
> 104 | |         }
>     | |_________- in this macro invocation
> error: aborting due to 2 previous errors

LMJW:
@asomers Do you have any further suggestions on the improvement?

Nugine:
Eight months have passed. Is there any progress?

LMJW:
I am happy to keep work on it. Just needs more inputs on the changes

src/sys/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+libc_enum! {
+    /// The Resource enum is platform dependent. Check different platform
+    /// manuals for more details. Some platform links has been provided for
+    /// earier reference (non-exhaustive).
+    ///
+    /// * [Linux](https://man7.org/linux/man-pages/man2/getrlimit.2.html)
+    /// * [freeBSD](https://www.freebsd.org/cgi/man.cgi?query=setrlimit)
```
asomers, line 22: The preferred spelling is "FreeBSD" (note the capital F)

src/sys/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+        #[cfg(not(any(target_os = "netbsd", target_os = "freebsd")))]
+        RLIMIT_AS,
+        RLIMIT_CORE,
+        RLIMIT_CPU,
+        RLIMIT_DATA,
+        RLIMIT_FSIZE,
+        RLIMIT_NOFILE,
+        RLIMIT_STACK,
+
+        // platform specific
```
asomers, line 50: Why is this comment here?  It seems like it should apply to every entry that has a `#[cfg(...)]`.  But `#[cfg(...)]` is self-explanatory.  Better just to leave it out.

src/sys/resource.rs
```diff
@@ -0,0 +1,192 @@
...
+/// ```
+///
+/// # References
+///
+/// [setrlimit(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html#tag_16_215)
+///
+/// [`Resource`]: enum.Resource.html
+///
+/// Note: `setrlimit` provides a safe wrapper to libc's `setrlimit`. For the
+/// platform that are not compatible, try using `prlimit` to set rlimit.
```
asomers, line 173: Is this comment obsolete now?  prlimit is no longer included in the PR.

test/test_resource.rs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// We first fetch the existing file descriptor maximum values using getrlimit(), then edit the
+/// soft limit to make sure it has a new and distinct value to the hard limit. We then setrlimit()
+/// to put the new soft limit in effect, and then getrlimit() once more to ensure the limits have
+/// been updated.
+#[test]
+pub fn test_resource_limits_nofile() {
+    let (soft_limit, hard_limit) = getrlimit(Resource::RLIMIT_NOFILE).unwrap();
+
+    // make sure the soft limit and hard limit are not equal
```
asomers, line 15: What ensures that the soft limit is not equal to the hard limit?

asomers (changes requested):

src/sys/resource.rs
LMJW: Thanks for the comments. Corrected :)

src/sys/resource.rs
LMJW: Removed

src/sys/resource.rs
LMJW: Updated the obsolete comments.

test/test_resource.rs
LMJW: This is actually obsolete comments from the very early version of the test, see https://github.com/LMJW/nix/commit/4f8b468db6ff2388afc0051032acbc1e662afcf2

For this particular test, I believe we don't need to have soft limit to equal to hard limit

LMJW:
Hi @asomers. I have made some updates. Please let me know if you have any concerns. Thanks

asomers (commented):
Ok, it looks good.  Just add a CHANGELOG entry and you'll be done.

LMJW:
Hi @asomers. I have updated the CHANGELOG. Thanks for reviewing the pull request :)

asomers (approved):
bors r+

yihuaf:
When would the next release be that includes this change? Thanks.

PR #1306, opened 2020-10-05
Add wrapper for mremap

### LINKED ISSUE #1295: Add wrapper for libc::mremap
I need this in https://github.com/shadow/shadow. Happy to take a stab at it :)

sporksmith:
Fixes #1295

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -139,6 +139,17 @@ libc_bitflags!{
     }
 }
 
+#[cfg(target_os = "linux")]
+libc_bitflags!{
+    /// Options for `mremap()`.
+    pub struct MRemapFlags: c_int {
+        /// Permit the kernel to relocate the mapping to a new virtual address, if necessary.
+        MREMAP_MAYMOVE;
+        /// Place the mapping at exactly the address specified in `new_address`.
+        MREMAP_FIXED;
+    }
+}
+
 libc_enum!{
     /// Usage information for a range of memory to allow for performance optimizations by the kernel.
     ///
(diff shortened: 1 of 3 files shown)


sporksmith:
It looks like the libc crate doesn't have mremap in the FreeBSD CI environment:

```
error[E0425]: cannot find function `mremap` in module `libc`
   --> src/sys/mman.rs:330:21
    |
330 |     let ret = libc::mremap(addr, old_size, new_size, flags.bits(), new_address);
    |                     ^^^^^^
help: a function with a similar name exists
    |
330 |     let ret = libc::mmap(addr, old_size, new_size, flags.bits(), new_address);
    |                     ^^^^
help: possible candidate is found in another module, you can import it into scope
    |
1   | use crate::sys::mman::mremap;
    |
```

I assume I need to config-guard this wrapper, but I'm not sure how to tell which configs ought to include it. I don't see any such config-guard in the libc crate itself: https://docs.rs/libc/0.2.68/src/libc/unix/linux_like/linux/mod.rs.html#2910

Any ideas?

sporksmith:
> I assume I need to config-guard this wrapper, but I'm not sure how to tell which configs ought to include it. I don't see any such config-guard in the libc crate itself: https://docs.rs/libc/0.2.68/src/libc/unix/linux_like/linux/mod.rs.html#2910

Added a guard to make it Linux-only, since the man page notes that it's Linux-specific.

sporksmith:
Hmm, I'm fairly certain the CI failure is spurious. It doesn't look related to this PR, and it previously passed on 7a46b52, which should be exactly the same.

I'm happy to rerun if desired; is there a nicer way to force a rerun than pushing an empty commit?

src/sys/mman.rs
```diff
@@ -315,6 +315,28 @@ pub unsafe fn mmap(addr: *mut c_void, length: size_t, prot: ProtFlags, flags: Ma
...
+/// # Safety
+///
+/// See the `mremap(2)` man page for detailed requirements.
+#[cfg(target_os = "linux")]
+pub unsafe fn mremap(
+    addr: *mut c_void,
+    old_size: size_t,
+    new_size: size_t,
+    flags: MapFlags,
+    new_address: size_t,
```
asomers, line 329: Shouldn't `new_address` be optional?

sporksmith: Yes; I wasn't sure the best way to handle this since Rust doesn't support variadic functions. The argument is ignored `MAP_FIXED` isn't in `flags`; i.e. if `MAP_FIXED` isn't set the caller can put anything there, and it'll be ignored.

Maybe it'd be a bit clearer to make it an `Option<size_t>`?

sporksmith: I just realized the flags enum is wrong; will fix that and change new_address to be an `Option<size_t>` (but lmk if there's another solution you recommend)

sporksmith: Done. Also added a few tests, though in hindsight maybe these should just be doc tests (and drop the second mremap test)?

src/sys/mman.rs
```diff
@@ -315,6 +326,28 @@ pub unsafe fn mmap(addr: *mut c_void, length: size_t, prot: ProtFlags, flags: Ma
     }
 }
 
+/// Expands (or shrinks) an existing memory mapping, potentially moving it at the same time.
+///
+/// # Safety
+///
+/// See the `mremap(2)` man page for detailed requirements.
```
asomers, line 333: How about a URL to a man page?

asomers (changes requested):
Nice tests.

src/sys/mman.rs
sporksmith: Done

asomers:
Looks good.  Now it just needs a squash.

sporksmith:
> Looks good. Now it just needs a squash.

Done

asomers-ax (approved):
bors r+

asomers (approved):
bors r+

PR #1313, opened 2020-10-12
Update to cfg-if 1.0

mbrubeck:


asomers:
Why?  What is the benefit?

mbrubeck:
The only change in this release is alexcrichton/cfg-if#37, which removes some syntax that was inconsistent with the standard `cfg` macro and attributes.

There's no urgent reason to update right away, but once everyone has updated to latest version, it will reduce the need for downstream crates to download and compile two different versions of `cfg-if`.  It will also let users of this library automatically benefit from fixes in any future 1.x release of `cfg-if`.

asomers (approved):
bors r+

asomers:
bors retry

lopopolo:
@asomers is it possible to get an 0.19 point release that contains this dep bump? I'm trying to update all of my transitive dependencies to cfg-if v1.

asomers:
Yes.  In fact, I would really like to do a point release just for #1337

asomers:
https://crates.io/crates/nix/0.19.1

PR #1316, opened 2020-10-16
update require_kernel_version string to handle "_" in version string

ritzk:
test code breaks on fc333

```
$ cargo test
failures:

---- sys::test_socket::recvfrom::udp_offload::gro stdout ----
thread 'sys::test_socket::recvfrom::udp_offload::gro' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError("Extra junk after valid version: _64")', test/sys/test_socket.rs:292:13
```

this is due underscore in release string( arch/x86_64), which is not supported by semver.
```
$ uname -r
5.8.14-300.fc33.x86_64

```

### DIFF
--- a/test/test.rs
+++ b/test/test.rs
@@ -95,7 +95,11 @@ cfg_if! {
 
                 let uname = nix::sys::utsname::uname();
 
-                let mut version = Version::parse(uname.release()).unwrap();
+                // Linux may report version as 4.18.el8_2.x86_64 or 5.18.200-fc33.x86_64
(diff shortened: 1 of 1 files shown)


asomers:
So the problem here is that we're parsing the Linux kernel version as a Semver string, even though it isn't.  And `_` isn't allowed anywhere in a Semver string.  Your solution is reasonable.  But could you please add a comment explaining why?

asomers (approved):
bors r+

asomers:
bors retry

PR #1317, opened 2020-10-18
test_af_alg_aead waits indefinitely

ritzk:
Starting with linux kernel 4.9, the crypto interface changed slightly such that the authentication tag memory is only needed in the output buffer for encryption and in the input buffer for decryption. Thus, we have fewer bytes to read than the buffer size. Do not block on read.

alternatively,  we can adjust the decrypted buffer size based on kernel version ( ">= 4.9") to not include auth_size .
```
     if kernel_version >= "4.9":
         let mut decrypted = vec![0u8; payload_len + (assoc_size as usize) ];
```

before 
```
test sys::test_socket::test_af_alg_aead ... test sys::test_socket::test_af_alg_aead has been running for over 60 seconds
```
after
```
test sys::test_socket::test_af_alg_aead ... ok
```

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -583,6 +583,7 @@ pub fn test_af_alg_cipher() {
 #[test]
 pub fn test_af_alg_aead() {
     use libc::{ALG_OP_DECRYPT, ALG_OP_ENCRYPT};
+    use nix::fcntl::{fcntl, FcntlArg, OFlag};
     use nix::sys::uio::IoVec;
     use nix::unistd::{read, close};
     use nix::sys::socket::{socket, sendmsg, bind, accept, setsockopt,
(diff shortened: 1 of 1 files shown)


asomers:
So if the caller tries to retrieve more data than is available with read(2), the kernel will block?  That seems very stupid.  Shouldn't it only block if there are 0 bytes available?

ritzk:
> So if the caller tries to retrieve more data than is available with read(2), the kernel will block? That seems very stupid. Shouldn't it only block if there are 0 bytes available?

On Linux, read will blocks until it can fill the requested nbytes of buffer or if the connection terminates.

cemeyer:
I also encountered this trying to run the full test suite on Linux 5.13.12.

asomers (approved):
It seems like a stupid interface if it blocks because the user's buffer is too big.  But Nix can't fix that.  Let's merge your hack.

bors r+

cemeyer:
> It seems like a stupid interface if it blocks because the user's buffer is too big.

Totally agree.

>But Nix can't fix that. Let's merge your hack.

Thanks!

PR #1318, opened 2020-10-18
KEEPCNT/KEEPINTVL aren't available on OpenBSD.

ltratt:
Without this commit, nix doesn't currently compile on OpenBSD. Some keepalive settings can be set globally, but not per-socket (see e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=970550 for some digging into this). Since it seems that NetBSD and DragonflyBSD have these settings, it makes more sense to only exclude them on OpenBSD rather than include them on (pretty much every) other operating systems.


asomers (approved):
bors r+

PR #1320, opened 2020-10-20
Add an optional interface argument to Ipv6MembershipRequest's constru…

simias:
This is my proposed change for #323 

This is a breaking change, existing code will have to add an additional `None` argument to the constructor to get the previous behaviour.


simias:
Apparently on Android the interface is signed even though `if_nametoindex` still returns a `c_uint`. Odd.

simias:
I tried to update this PR for current master but I see that the tests have failed, however I'm not sure if my change is in cause.

PR #1321, opened 2020-10-20
Add an `any` constructor to Ipv6Addr

simias:
This mirrors the behaviour of `Ipv4Addr::any()` but for IPV6 addresses.

Unless I missed something, `libc` doesn't export anything resembling `INADDR_ANY` for IPV6, so the zeroes are hardcoded. I can't imagine that this would ever cause issues however.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -477,6 +477,10 @@ impl Ipv6Addr {
         Ipv6Addr(libc::in6_addr{s6_addr: to_u8_array!(a,b,c,d,e,f,g,h)})
     }
 
+    pub fn any() -> Ipv6Addr {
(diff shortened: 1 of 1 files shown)


simias:
The error appears to be unrelated to the change.

asomers:
In C, this is called IN6ADDR_ANY_INIT.  You could try adding that to libc.

simias:
Well there are two initializers in C since IPV6 uses structs instead of a 32bit int, there's the `IN6ADDR_ANY_INIT` macro and the `in6_addr_any` global. I suspected that not having them in the libc crate was intentional since they're not very useful in Rust but maybe I'm wrong. I'll open an issue in the libc repo to ask what they think.

simias:
I opened an issue in the libc crate: https://github.com/rust-lang/libc/issues/1950

PR #1324, opened 2020-10-21
Make `nix::sys::termios::Termios` implement `Sync` and `Copy`.

jimblandy:
It's kind of tough that `Termios` doesn't implement `Sync`, because then you
can't use them with the `signal_hook` crate's `register` function, or any other
signal-handling wrapper that tries to be safe.

This approach probably entails a bit more copying bits around, but I'd argue the
convenience of `Sync` and `Copy` implementations is worth it.


jimblandy:
(I tried to fix the cfg-ed out code, too, but if I got it all right I'll be surprised...)

jimblandy:
Yeah, I missed a case, and that's causing the FreeBSD failure.

The `test-aio-drop` failures don't seem to have anything to do with this push.

jimblandy:
@asomers: It doesn't seem like @susurrus is available; is there anyone else who can review?

src/sys/termios.rs
```diff
@@ -187,39 +191,19 @@ impl Termios {
...
-            let mut termios = self.inner.borrow_mut();
-            termios.c_iflag = self.input_flags.bits();
-            termios.c_oflag = self.output_flags.bits();
-            termios.c_cflag = self.control_flags.bits();
-            termios.c_lflag = self.local_flags.bits();
-            termios.c_cc = self.control_chars;
-        }
-        self.inner.as_ptr()
+    pub(crate) fn get_libc_termios(&self) -> libc::termios {
+        let mut termios = self.inner.clone();
```
asomers, line 195: You shouldn't need `clone` if libc::termios is already `Copy`.  And if `libc::termios` is not already `Copy`, then you can't derive `Copy`.  Or is it only `Copy` on some platforms?

jimblandy: Nothing subtle, I just missed it. I believe it is clone on all platforms.

jimblandy: Pushed a version without the clone.

jimblandy:
The CI failure is due to `libc`'s deprecation of the `af_alg_iv` struct:
```
error: use of deprecated struct `libc::af_alg_iv`: this struct has unsafe trait implementations that will be removed in the future
```

This could be addressed by either #1329 or #1285.

PR #1329, opened 2020-11-09
Avoid using libc::af_alg_iv to get rid of deprecation warnings

coolreader18:
rust-lang/libc#1501

Not sure if this is the best approach, or whether we should just define our own version? Though I know that's generally avoided in nix. Alternatively, we could just slap a `#[allow(deprecated)]` on it and figure it out later, but I just wanted to do something because it's causing CI to fail.


tamird:
Upstream's deprecation warning is informational, it is intended to let folks know that the type is soon changing. I've addressed this as part of #1285.

asomers (approved):

asomers:
Fixed by #1332 instead.

PR #1333, opened 2020-11-15
Dir: Implement `IntoIterator` for `Dir`

wmanley:
This is useful to allow returning an iterator based on a directory iterator
without needing a self-referential struct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `personality` (#[1331](https://github.com/nix-rust/nix/pull/1331))
 - Added limited Fuchsia support (#[1285](https://github.com/nix-rust/nix/pull/1285))
 - Added `getpeereid` (#[1342](https://github.com/nix-rust/nix/pull/1342))
+- Implemented `IntoIterator` for `Dir`
+  (#[1333](https://github.com/nix-rust/nix/pull/1333)).
+### Changed
 ### Fixed
 - Define `*_MAGIC` filesystem constants on Linux s390x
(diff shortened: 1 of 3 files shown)


asomers:
I don't understand the motivation for this PR.  Can you please show an example?  What does this enable that couldn't as easily be done with the existing API?

wmanley:
Here's where I use it: https://github.com/wmanley/ostree-oxide/blob/d3f38da61deb614be67d3705519df0235e4d2e84/ostree-repo/src/lib.rs#L253-L281

This allows me to return an `Iterator` that owns the `Dir`. This is important because I open the `Dir` in the function that returns the `Iterator`.  I couldn't work out how to do this with the existing API short of self-referential structs.

asomers:
Would it be possible to add a usage example?

wmanley:
> Would it be possible to add a usage example?

Done

wmanley:
Ping: Are there any other changes you'd like to make this suitable to go in?

asomers (changes requested):
I think this looks good.  It just needs three small things:
1) CHANGELOG entry (may require a rebase to avoid conflicts)
2) Squash
3) Doc comment for `OwningIter`.  It can simply say "The return type of [`Dir::into_iter`]".

wmanley:
> I think this looks good. It just needs three small things:

Thanks for the review. I've done these three things.

CI on "OSX x86_64" failed with:

> warning Agent process on a persistent worker exited unexpectedly!

but I can't imagine that the failure was caused by this PR.

asomers (approved):
The OSX failure looks like perhaps GCE yanked the VM out from under us.  That happens sometimes.

bors r+

PR #1336, opened 2020-11-17
Allow Android to use timerfd

zachoverflow:
Hey there! I'm from the Android Bluetooth team, and we're using timerfd for our timers (even in our old C/C++ code) since they can wake the device. Here's a small patch that opens up timerfd to the Android config too.


asomers:
Looks good, assuming it works (we don't run Android tests in CI).  But could you also please add a CHANGELOG entry?  And since you obviously have an Android environment, could you please try running the test suite and report what happens?

rtzoeller:
Superseded by #1509.

PR #1337, opened 2020-11-17
Add docs.rs target list

myfreeweb:
ref: https://blog.rust-lang.org/2020/03/15/docs-rs-opt-into-fewer-targets.html

Makes sense to build docs for all supported OSes (BSD, Redox) rather than completely empty docs for Windows :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,19 @@ exclude     = [
   "/bors.toml"
 ]
 
+[package.metadata.docs.rs]
+targets = [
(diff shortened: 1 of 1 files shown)


asomers:
Oh, this is cool!  I didn't know that doc.rs could do this.  I thought that it gave you just Windows, OSX, and Linux and that was it.

asomers (approved):
bors r+

asomers:
bors retry

PR #1338, opened 2020-11-22
Add a passwd field to Group

blinsay:
Adds a `passwd` field to unistd::Group. The `gr_passwd` field exists on `libc::group` and wasn't exposed.

I didn't see tests for `from(libc:group)` for either `User` or `Password`. Let me know if there are other tests I should add!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+
+- Added a `passwd` field to `Group` (#[1338](https://github.com/nix-rust/nix/pull/1338))
 - Added `mremap` (#[1306](https://github.com/nix-rust/nix/pull/1306))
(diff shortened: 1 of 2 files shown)


asomers (commented):
What systems actually use the group password?  I for one have never seen it actually used.

blinsay:
I haven't actually seen it used either, but I'm working on a backend to nscd and glibc seems to expect [a group password as part of the response](https://sourceware.org/git/?p=glibc.git;a=blob;f=nscd/nscd_getgr_r.c;h=94d418df08170464d88bd3250fd436692adcd517;hb=HEAD#l183).

Given that passwd is mentioned [in the man page](https://linux.die.net/man/5/group), is there a reason to not include it here?

blinsay:
@asomers given this any more thought?

asomers:
Yes, I think it seems reasonable.  But could you add a CHANGELOG entry and rebase?

blinsay:
Done! Sorry for the delay.

asomers (approved):
bors r+

Issue #1325, opened 2020-10-21
Recvmmsg can return fewer than specified number of messages


gmilleramilar:
The recvmmsg system call can result in fewer than the specified number of message buffers being filled.  For example if the system call is interrupted after some data has been received.  However the implementation of `nix::sys::socket::recvmmsg()` does not allow for this and unconditionally attempts to process all of the message buffers, and crashes with a failed assertion.

Fortunately, I believe the fix is simple.

```diff
diff --git a/src/sys/socket/mod.rs b/src/sys/socket/mod.rs
index 5e5fb8d..98203bf 100644
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1219,7 +1219,7 @@ pub fn recvmmsg<'a, I>(
     let r = Errno::result(ret)?;

     Ok(output
-        .into_iter()
+        .into_iter().take(ret as usize)
         .zip(addresses.iter().map(|addr| unsafe{addr.assume_init()}))
         .zip(results.into_iter())
         .map(|((mmsghdr, address), (msg_controllen, cmsg_buffer))| {
```

codeslinger:
Probably can close this one now, right?

asomers:
Fixed by #1341

PR #1341, opened 2020-11-25
Fix recvmmsg(2) implementation

codeslinger:
There were two problems discovered with the `recvmmsg(2)` implementation
that this changeset attempts to fix:

1. As mentioned in nix-rust/issues/1325, `recvmmsg(2)` can return fewer
   messages than requested, and
2. Passing the return value of `recvmmsg(2)` as the number of bytes in
   the messages received is incorrect.

This changeset incorporates the proposed fix from nix-rust/issues/1325,
as well as passing the correct value (`mmsghdr.msg_len`) for the number
of bytes in a given message.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1217,17 +1217,18 @@ pub fn recvmmsg<'a, I>(
 
     let ret = unsafe { libc::recvmmsg(fd, output.as_mut_ptr(), output.len() as _, flags.bits() as _, timeout) };
 
-    let r = Errno::result(ret)?;
+    let _ = Errno::result(ret)?;
 
     Ok(output
         .into_iter()
+        .take(ret as usize)
         .zip(addresses.iter().map(|addr| unsafe{addr.assume_init()}))
         .zip(results.into_iter())
         .map(|((mmsghdr, address), (msg_controllen, cmsg_buffer))| {
             unsafe {
                 read_mhdr(
                     mmsghdr.msg_hdr,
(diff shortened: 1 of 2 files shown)


codeslinger:
I don't quite understand why the CI didn't pass on stable. On my Linux machine, the `test-aio-drop` test passes. It appears to be failing on other open PRs against this repo, as well, without any associated changes to anything that would affect that test.

asomers:
Do you know of any way to force the condition described in #1325, so we can test for it?

asomers:
BTW, the failure of test-aio-drop is not related to this PR.  I suspect it's due to a bug in glibc.  But I've never been able to reproduce it locally, so I'm not sure.

codeslinger:
Re: testing: as far as the actual problem that existed prior to this PR, you should be able to:

1. Create a UDP socket
2. In one thread, send the socket some number of messages, `X`
3. In another thread, call `recvmmsg` with a number of `mmsghdr`s > `X` and a `flags` value of `MSG_DONTWAIT` (or `MSG_WAITFORONE` on Linux)

I can add a test for that to this PR if that seems acceptable.

codeslinger:
Added a test for short reads when using `MSG_DONTWAIT`.

asomers:
Looks good!  Could you also please add a CHANGELOG entry?

asomers:
Actually, in the interest of expediency, I'm going to merge this now and add a CHANGELOG entry later.

asomers (approved):
bors r+

PR #1327, opened 2020-10-28
Disable clock_getres on redox

coolreader18:

PR #1349, opened 2020-12-02
Correctly build for redox in CI

tamird:
Per discussion with @asomers in #1285.


asomers:
This is going to fail without other changes.  Let's just cancel this PR and fix Redox in PR #1327 , which is already open.

PR #1350, opened 2020-12-03
Switch all builds from Travis to Cirrus

asomers:
Travis has been super-slow lately (> 6 hours per build).  Cirrus is much
faster: about 20 minutes.  Cirrus also has slightly better test
coverage, mainly because it doesn't use SECCOMP.
    
Also,
   
* Fix the Redox CI build.  The old Travis configuration didn't actually
  build for Redox, so we never noticed that Redox can't be built with a
  stable compiler.  Thanks to @coolreader18 for finding this.
    
* Disable the udp_offload tests on cross-tested platforms.  These tests
  are failing with ENOPROTOOPT in Cirrus-CI.  I suspect it's due to a
  lack of support in QEMU.  These tests were skipped on Travis because
  its kernel was too old.
    
* Fix require_kernel_version on Cirrus-CI.  Cirrus reports the Linux
  kernel version as 4.19.112+, which the semver crate can't handle.
    
* Fix test_setfsuid on Cirrus.  When run on Cirrus, it seems like the
  file in /tmp gets deleted as soon as it's closed.  Probably an
  overzealous temporary file cleaner.  Use /var/tmp, because no
  temporary file cleaner should run in there.
    
* Skip mount tests on Cirrus.  They fail for an unknown reason.
Issue #1351
    
* Skip the AF_ALG tests on Cirrus-CI
Issue #1352


kamalmarhubi (approved):
wow this is amazing 😮

posborne (approved):
Nice work!  That's more than a little faster!

asomers-ax:
Grr, but now musl is failing, and it's consistent.  I guess I have to fix that one, too.

PR #1353, opened 2020-12-09
Fix intermittency in some tests

asomers:
These tests have recently begun to fail intermittently in CI.

### DIFF
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -19,7 +19,6 @@ use std::ffi::CString;
 use std::fs::DirBuilder;
 use std::fs::{self, File};
 use std::io::Write;
-use std::mem;
 use std::os::unix::prelude::*;
 #[cfg(not(target_os = "redox"))]
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


test/test_unistd.rs
```diff
@@ -466,9 +465,7 @@ fn test_fchown() {
     fchown(fd, uid, gid).unwrap();
     fchown(fd, uid, None).unwrap();
     fchown(fd, None, gid).unwrap();
-
-    mem::drop(path);
-    fchown(fd, uid, gid).unwrap_err();
+    fchown(999999999, uid, gid).unwrap_err();
```
kamalmarhubi, line 468: what's this change doing?

kamalmarhubi (approved):
one question but don't block on it

test/test_unistd.rs
asomers: Here's what I put in the commit message:

Fix a race in the test_unistd::test_fchown test
    
The test was assuming that once dropped, a temporary file's file
descriptor would be invalid.  But it might not be, because another file
might be opened with the same file descriptor.  Instead, use an
obviously invalid file descriptor.

kamalmarhubi: ah my bad: didn't look at the individual commits. sorry!

PR #1358, opened 2020-12-11
Raise MSRV to 1.40.0

asomers:
This will shortly be needed for Fuchsia support.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,7 +12,7 @@ task:
     image: freebsd-11-4-release-amd64
   setup_script:
     - fetch https://sh.rustup.rs -o rustup.sh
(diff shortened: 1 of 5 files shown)


tamird:
I saw that same musl issue in my PR (https://github.com/nix-rust/nix/pull/1285/files#diff-c2454cd83dd7725daae35caff3a062f117176b9d4922393e97307b2a91e2aa12L1073).

I guess this is the cause of the mac failures. Thanks for investigating. I'll watch this PR.

asomers:
@posborne @kamalmarhubi do either of you have a Mac and want to try debugging the OSX failure?  It fails every time.  But when I add a line of debugging just above the `libc::aio_suspend`, it passes.  So it seems like a timing issue.

kamalmarhubi (approved):
unfortunately I don't have a mac :/

Issue #1360, opened 2020-12-15
How to use `SockLevel` for `setsockopt` and IP_MULTICAST_IF


keepsimple1:
https://github.com/nix-rust/nix/blob/d922c72a765640c5e190fb01129f712e6ac513df/src/sys/socket/mod.rs#L1585

I cannot find how `SockLevel` is used in the code. For example, `setsockopt` did not have a level argument: 

https://github.com/nix-rust/nix/blob/d922c72a765640c5e190fb01129f712e6ac513df/src/sys/socket/mod.rs#L1639

How can we use `SockLevel`? I have a use case that needs to do something like this (not real code):

```
setsockopt(IPPROTO_IP, IP_MULTICAST_IF, <value>);
```
I cannot find `IP_MULTICAST_IF` in Nix either. Did I miss it somewhere? Thanks.

asomers:
Ooh, great question!  `SockLevel` is actually dead code.  AFAICT, its use was eliminated way back in PR #133 .  So we should delete it.  But what you probably ought to do is submit a PR that adds this socket option using `sockopt_impl!` to src/sys/socket/sockopt.rs.

keepsimple1:
thanks @asomers for opening PR #1362 to remove `SockLevel`. I will submit a PR to add `IP_MULTICAST_IF` .

PR #1363, opened 2020-12-15
add IP_MULTICAST_IF for sockopt

keepsimple1:
IP_MULTICAST_IF is not supported by `sockopt` in `nix`. See issue #1360 .

According to Linux [man page](https://man7.org/linux/man-pages/man7/ip.7.html), IP_MULTICAST_IF is used in this way: 

quote:
```
IP_MULTICAST_IF (since Linux 1.2)
              Set the local device for a multicast socket.  The argument for
              setsockopt(2) is an ip_mreqn or (since Linux 3.5) ip_mreq
              structure similar to IP_ADD_MEMBERSHIP, or an in_addr struc‐
              ture.  (The kernel determines which structure is being passed
              based on the size passed in optlen.)  For getsockopt(2), the
              argument is an in_addr structure.
```

Please review this patch.  Also I have a question:  how do we add automated testing for such `sockopt` in `nix`? I've added a test case in my own code that uses `nix` and it works, but don't know about the practice of testing inside `nix`. Thanks.

Update: 12/24/2020

The original approach of defining an Enum did not work well because `#[repr(transparent)]` does not support Enum. Moreover, my use cases only care about `in_addr` value, which is the most common AFAIK. So I refactored the patch to address the major cases I actually understand, and added a test case.

Also, based on the Github Actions results, I excluded the patch from the (OS, env, arch) that failed with error `ENOPROTOOPT`.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -385,8 +385,9 @@ cfg_if! {
                          target_arch = "powerpc64"))))] {
         /// Request for multicast socket's outgoing interface
         ///
-        /// The value must be a `libc::in_addr` structure, which is the most common.
-        /// `ip_mreqn` or `ip_mreq` structure are not supported. One reason is that
+        /// The value is a `libc::in_addr` structure, which is the most common.
+        /// `ip_mreqn` or `ip_mreq` structure are not supported at this point, one
+        /// reason is that if we use enum to represent all possible structures,
         /// `#[repr(transparent)]` does not support enum.
         #[repr(transparent)]
         #[derive(Clone, Copy, Debug, Eq, PartialEq)]
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -375,18 +375,28 @@ impl Ipv6MembershipRequest {
     }
 }
 
-/// Request for multicast socket's outgoing interface
-///
-/// The value must be a `libc::in_addr` structure, which is the most common.
-/// `ip_mreqn` or `ip_mreq` structure are not supported. One reason is that
-/// `#[repr(transparent)]` does not support enum.
-#[repr(transparent)]
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub struct IpMulticastIfInAddr(libc::in_addr);
+cfg_if! {
+    if #[cfg(not(all(target_os = "linux",
+                     target_env = "gnu",
+                     any(target_arch = "mips",
+                         target_arch = "mips64",
+                         target_arch = "aarch64",
+                         target_arch = "arm",
+                         target_arch = "powerpc64"))))] {
+        /// Request for multicast socket's outgoing interface
+        ///
+        /// The value must be a `libc::in_addr` structure, which is the most common.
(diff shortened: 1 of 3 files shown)


asomers:
Nix doesn't currently use anything fancy like mocking for tests.  Just try to exercise the socket option.  Don't write an exhaustive test suite for it; that's a job for the kernel maintainers.  Nix's test suite should do just enough to ensure that the bindings are correct.

keepsimple1:
Thanks @asomers for your answers. I've updated the diff and the PR Description, adding a test case. Please take a look when you have time.

src/sys/socket/sockopt.rs
```diff
@@ -233,6 +233,17 @@ cfg_if! {
         sockopt_impl!(SetOnly, Ipv6DropMembership, libc::IPPROTO_IPV6, libc::IPV6_LEAVE_GROUP, super::Ipv6MembershipRequest);
     }
 }
+cfg_if! {
```
asomers, line 236: `cfg_if` is helpful when you expect to have `else` or `else if` blocks.  But in cases like this, it's simpler to do:
```rust
#[cfg(...)]
sockopt_impl!(...)
```

src/sys/socket/mod.rs
```diff
@@ -375,6 +375,32 @@ impl Ipv6MembershipRequest {
...
+                         target_arch = "aarch64",
+                         target_arch = "arm",
+                         target_arch = "powerpc64"))))] {
+        /// Request for multicast socket's outgoing interface
+        ///
+        /// The value is a `libc::in_addr` structure, which is the most common.
+        /// `ip_mreqn` or `ip_mreq` structure are not supported at this point, one
+        /// reason is that if we use enum to represent all possible structures,
+        /// `#[repr(transparent)]` does not support enum.
+        #[repr(transparent)]
```
asomers, line 392: Does the option work with IPv6?  If so you should use an enum type.  Or maybe define different Nix sockopts, one for each supported address type.

asomers (changes requested):
My man page says `The use of IP_MULTICAST_IF is not recommended, as multicast memberships are scoped to each individual interface.  It is supported for legacy use only...`.  Given that, we probably should't add Nix support for it.  Is there something you need it for that can't be done any other way?

keepsimple1:
> My man page says `The use of IP_MULTICAST_IF is not recommended, as multicast memberships are scoped to each individual interface. It is supported for legacy use only...`. Given that, we probably should't add Nix support for it. Is there something you need it for that can't be done any other way?

Could you please share where I can find this man page you referred to? Thanks.

asomers:
https://www.freebsd.org/cgi/man.cgi?query=ip&apropos=0&sektion=4&manpath=FreeBSD+12.2-RELEASE+and+Ports&arch=default&format=html

keepsimple1:
> https://www.freebsd.org/cgi/man.cgi?query=ip&apropos=0&sektion=4&manpath=FreeBSD+12.2-RELEASE+and+Ports&arch=default&format=html

Thanks.  Looks like FreeBSD deems it legacy (while Linux seems not).

And I found a workaround in my project at this point. I will close this PR for now until someone found its use again.

PR #1365, opened 2020-12-18
Fix duplicate entry in bors.toml

asomers:
Also, sort the entries alphabetically, since that's how Github displays
them.


asomers:
Let's see if bors tries to use the bors.toml from this PR or from the master branch.

bors r+

PR #1366, opened 2020-12-27
x32 port

nabijaczleweli:
I played pretty loose with the statfs bit, hoping for CI to tell me if I broke something.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -157,6 +157,10 @@ task:
     - name: Linux s390x
       env:
         TARGET: s390x-unknown-linux-gnu
+    - name: Linux x32
+      env:
+        TARGET: x86_64-unknown-linux-gnux32
+        CHECK_TESTS: true
     - name: NetBSD x86_64
       env:
         TARGET: x86_64-unknown-netbsd
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Implemented `IntoIterator` for `Dir`
   (#[1333](https://github.com/nix-rust/nix/pull/1333)).
 ### Changed
+
 ### Fixed
 - Define `*_MAGIC` filesystem constants on Linux s390x
   (#[1372](https://github.com/nix-rust/nix/pull/1372))
+- mqueue, sysinfo, timespec, statfs, test_ptrace_syscall() on x32
+  (#[1366](https://github.com/nix-rust/nix/pull/1366))
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ tempfile = "3.0.5"
 semver = "0.9.0"
 
 [target.'cfg(any(target_os = "android", target_os = "linux"))'.dev-dependencies]
-caps = "0.3.1"
+caps = "0.5.1"
 
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
 sysctl = "0.1"
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Tier 2:
 Tier 3:
   * x86_64-fuchsia
   * x86_64-unknown-redox
+  * x86_64-unknown-linux-gnux32
 
 ## Usage
 
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -5,7 +5,7 @@
 use crate::Result;
 use crate::errno::Errno;
 
-use libc::{self, c_char, c_long, mqd_t, size_t};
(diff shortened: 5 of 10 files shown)


nabijaczleweli:
Should be g2g, will wait for review to squash.

Oddly, the `test_fcntl::linux_android::test_vmsplice` test that failed passes on amd64 for me?

asomers:
Interesting.  Are you actually using x32 with Rust?  Last I heard, Linux was talking about dropping it.

nabijaczleweli:
Well, I am running a full Debian x32 userland, with the only foreign packages being the kernel.

As for using it "with Rust", well, I only recently [found](https://twitter.com/nabijaczleweli/status/1343201883335364608) the target in rustup, did a bare-bones port of nix, and [hit a codegen bug](https://twitter.com/nabijaczleweli/status/1343221650259636224) almost immediately; LLVM (and, well, Rust: https://github.com/rust-lang/rust/issues/59220) is still very much baby mode on x32, but I would love to be able to just™ target Rust there. It's currently not really viable, but I also see no reason why it wouldn't become viable in future.

As for Linux dropping it, I neither know nor care; it's a (low priority, but still) supported target of the Rust platform as it stands.

nabijaczleweli:
Rebased, squashed, includes @hvdijk's test fixes from https://github.com/nix-rust/nix/issues/1384#issuecomment-774708486.

nabijaczleweli:
pogue, it even builds on CI. @hvdijk mind testing on your set-up?

hvdijk:
> pogue, it even builds on CI. @hvdijk mind testing on your set-up again?

It passes testing here, except for unrelated issues, nice.

For completeness, the unrelated issues are:
- `test_vsock` fails here because `SockAddr::new_vsock(libc::VMADDR_CID_LOCAL, port)` succeeds, just as it does on x86_64-linux-gnu.
- `test_af_alg_aead` hangs here during `read()`. After attaching gdb and immediately detaching, the test passes. I do not know what is happening here, but it also happens here for x86_64-unknown-linux-gnu.
- `unistd::Group::from_gid` fails because my system has no group named `root`.

asomers:
@nabijaczleweli why don't you go ahead and add x32 to .cirrus-ci.yml and to the Tier 3 target list in the README?

nabijaczleweli:
The tests [fail typecheck](https://cirrus-ci.com/task/4876934426394624?command=main#L73) at `CAPGET` and `CAPSET`, which are fixed in caps HEAD (https://github.com/lucab/caps-rs/commit/f6178d0ab650d68b64afb11eb8f530d4ff7ba148). I've [bugged](https://github.com/lucab/caps-rs/pull/66#issuecomment-778586570) the caps maintainer to push out a new version, but

hvdijk:
Just a heads up, tests are going to fail for x32 even with the updated caps-rs unless you have also patched rust's LLVM to include the x32 codegen fixes that are going to be in LLVM 12.

nabijaczleweli:
Hence why I only marked it for typecheck

asomers:
Ouch.  Well, we can't merge it like that.  You can remove the `CHECK_TESTS` line from .cirrus-ci.yml, which will fix the caps failure.  But what's this about a toolchain problem?  At what point will that cause a failure?

hvdijk:
> But what's this about a toolchain problem? At what point will that cause a failure?

LLVM has x32 codegen bugs that are only fixed in LLVM 12, but Rust, even nightly, is currently still based on LLVM 11. These codegen bugs affect Rust code badly. If you stick with what you get from `rustup`, then (after fixing the other problems) `cargo test --target=x86_64-unknown-linux-gnux32` fails with:

```
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/home/harald/rust-nix/target/x86_64-unknown-linux-gnux32/debug/deps/nix-0354ce9290eb8626` (signal: 11, SIGSEGV: invalid memory reference)
```

LLVM 12 already has its first release candidate out, so I do not expect it to take too long till we have an official version of Rust that generates correct x32 code. Until then, for actually running Rust code on x32, a built-from-source version of Rust is needed with LLVM 12's fixes applied.

nabijaczleweli:
At the point where you run it and it fails to properly relocate and segfaults in a Funny Way:tm:, at least that's what happened that time I built one of my programs.

I got rid of `CHECK_TESTS`, should I post another small PR adding it back when caps gets updated later, then?

asomers:
Yes, please add a PR for that when it's ready.  In the meantime, how about a CHANGELOG entry?

nabijaczleweli:
Added CHANGELOG entry and [re-added the test typecheck](https://github.com/lucab/caps-rs/pull/68#issuecomment-779376294) with the bumped caps.

asomers (approved):
bors r+

PR #1367, opened 2020-12-30
Add Conversion: Signal into c_int

szdytom:
**Background**
Some APIs (such as `nix::sched::clone`) requires a signal field with type `c_int` but not `nix::sys::signal::Signal`.
But I didn't find any functions can convert a `Signal` enum into its id in `int`.

**Description**
impl the `Into<c_int>` trait for `nix::sys::signal::Signal`.

**Other infomation**
The tests has passed on my computer(`Ubuntu 20.10`, `rustc 1.50.0-nightly`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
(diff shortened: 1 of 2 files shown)


asomers:
You don't need any special functions.  You can just do `Signal::SIGHUP as libc::c_int`.  An even better change would be to fix `clone` to take a `Signal` instead of a `c_int`.

szdytom:
Thank you for reviewing this! As there're better solotions, I'll close this pr.

PR #1375, opened 2021-01-12
Add safe wrappers for block device ioctl calls

### LINKED ISSUE #1374: Add protable wrappers for the block device IOCTLs
Add a portable and safe way of getting things like the size of block device (e.g. USB sticks or SD cards).

Other people (#1006 & #573) have had issue using the ioctl directly so it would be nice to have a safe simple wrapper.

It should have:
- a way to get the size of a block (BLKSSZGET)
- a way to get the size of the disc in bytes (BLKGETSIZE64) 
- a way to get the number of block

It would be nice to support the other BLK* ioctls
- re-read partion table (BLKRRPART)
- discard a range of blocks (BLKDISCARD)
- zero a range of block (BLKZEROOUT)
- check if the underlying medai supports something like TRIM that zeros when a block is discarded (BLKDISCARDZEROS)

aj-bagwell:
This add safe warpper for the BLK* ioctls using an extention trait of the File

It fixes #1374


aj-bagwell:
Agreed that nix is not the right place for this, closing unmerged.

Issue #1379, opened 2021-01-29
File descriptor not closed in TimerFd


crdumoul:
As specified in https://man7.org/linux/man-pages/man2/timerfd_create.2.html the file descriptor for a timerfd should be closed when it is no longer required. In `nix::sys::timerfd::TimerFd` there is no `Drop` implementation and the file descriptor is not automatically closed. This leads to a file descriptor leak.

The `TimerFd` struct either needs to properly own the file descriptor and close it when it's dropped, or make it explicit through documentation that the user of `TimerFd` needs to close the file descriptor themselves.

asomers:
Good catch!  Would you be willing to submit a PR to fix it?

crdumoul:
I'll do that shortly.

PR #1381, opened 2021-02-01
Close file descriptor on drop in TimerFd

crdumoul:
This change closes the TimerFd file descriptor on drop. Note that the TimerFd will no longer be `Clone` or `Copy`. Since it has a destructor it can't be `Copy`, and if it were `Clone` you could end up trying to use a closed TimerFd, or double-closing the file descriptor.

Addresses #1379.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added limited Fuchsia support (#[1285](https://github.com/nix-rust/nix/pull/1285))
 - Added `getpeereid` (#[1342](https://github.com/nix-rust/nix/pull/1342))
 ### Fixed
+- `TimerFd` now closes the underlying fd on drop.
+  ([#1381](https://github.com/nix-rust/nix/pull/1381))
 ### Changed
 
 - Minimum supported Rust version is now 1.40.0.
@@ -25,6 +27,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Removed `SockLevel`, which hasn't been used for a few years
   (#[1362](https://github.com/nix-rust/nix/pull/1362))
+- Removed both `Copy` and `Clone` from `TimerFd`.
+  ([#1381](https://github.com/nix-rust/nix/pull/1381))
 
 ## [0.19.1] - 28 November 2020
 ### Fixed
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
Good catch about `Clone`.  I think there may be some other structs that have that problem; I'll check for them.

A few requests:
* The drop implementation should panic on `EBADF`, because that usually indicates a double-close.
* Document somewhere that this type closes on drop.
* Don't forget to add a CHANGELOG entry!

crdumoul:
I believe I've addressed the suggested changes.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added limited Fuchsia support (#[1285](https://github.com/nix-rust/nix/pull/1285))
 - Added `getpeereid` (#[1342](https://github.com/nix-rust/nix/pull/1342))
 ### Fixed
+- TimerFd now closes the underlying fd on drop.
```
asomers, line 13: Also, in the "Removed" section you should note that is no longer Clone and Copy.

src/sys/timerfd.rs
```diff
@@ -270,3 +270,14 @@ impl TimerFd {
...
     }
 }
+
+impl Drop for TimerFd {
+    fn drop(&mut self) {
+        let result = Errno::result(unsafe {
+            libc::close(self.fd)
+        });
+        if let Err(Error::Sys(Errno::EBADF)) = result {
+            panic!("close of TimerFd encountered EBADF");
```
asomers, line 280: One problem: panicking during `drop` is usually bad, because `drop` itself gets called during panic.  A double-panic is difficult to debug.  Instead, you should only panic if the thread isn't already panicking.  In fact, you don't even need to `close` when panicking, because the process is about to abort anyway.  So I would write this as
```rust
if !std::thread::panicking() {
    //close the file descriptor, panicking on EBADF
}
```

asomers (changes requested):

src/sys/timerfd.rs
crdumoul: The file descriptor is now only closed if the thread isn't already panicking.

CHANGELOG.md
crdumoul: Done

asomers:
Would you mind squashing your commits?  We can't do this with Github's merge button because we use bors.

crdumoul:
> Would you mind squashing your commits? We can't do this with Github's merge button because we use bors.

This is done.

asomers (approved):
bors r+

PR #1382, opened 2021-02-02
Don't implement Clone on Dir, SignalFd, and PtyMaster

asomers:
Since they close their file descriptors on Drop, it's almost impossible
to use Clone without creating a double-close situation.

Also, check for EBADF in SignalFd::drop and Dir::drop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


asomers:
bors retry

coolreader18:
This is causing problems for me - I don't think drop implementations should panic on a syscall error. File et al in `std` just ignore the result of the close() call, since I think it can be assumed that if there is an error like EBADF then that was achieved through `FromRawFd` or calling into `libc` or something like that. At the very least, I'd appreciate an option to close a Dir without it panicking on error.

PR #1390, opened 2021-02-20
pty: Make forkpty() unsafe

### LINKED ISSUE #1388: forkpty() should probably be unsafe
`fork()` [is unsafe](https://docs.rs/nix/0.19.1/nix/unistd/fn.fork.html) (for good reason; you're not allowed to do much in the child if the parent was multithreaded).  OTOH, `forkpty()` [is not unsafe](https://docs.rs/nix/0.19.1/nix/pty/fn.forkpty.html) even though the same reasoning should apply.

tavianator:
After the child returns from a fork() of a multi-threaded process, it is
undefined behaviour to call non-async-signal-safe functions according to
POSIX.  Since forkpty() is implemented in terms of fork(), those
restrictions should apply to it too.

Fixes #1388

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+
 ### Changed
+- Made `forkpty` unsafe, like `fork`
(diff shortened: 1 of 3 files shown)


asomers:
Thanks for the PR!  This looks good, but you will have to rebase.  Otherwise git will mismerge the CHANGELOG.

asomers (approved):
bors r+

PR #1392, opened 2021-02-23
During CI, use just 1 cpu for QEMU builds

asomers:
https://github.com/cirruslabs/cirrus-ci-docs/issues/741

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -52,12 +52,11 @@ task:
 
 # Use cross for QEMU-based testing
 # cross needs to execute Docker, so we must use Cirrus's Docker Builder task.
-docker_builder:
+task:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -89,6 +88,12 @@ docker_builder:
     - name: Linux powerpc64le
       env:
         TARGET: powerpc64le-unknown-linux-gnu
+  compute_engine_instance:
+    image_project: cirrus-images
+    image: family/docker_builder
```
fkorotkov, line 93: Should be

```suggestion
    image: family/docker-builder
```

asomers: Grahh, thanks for the tip.

PR #1395, opened 2021-03-02
Allow sockaddr_ll size mismatch

internetionals:
Apparently the Linux kernel can return smaller sizes when the value in
the last element of sockaddr_ll (`sll_addr`) is smaller than the
declared size of that field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (#[1390](https://github.com/nix-rust/nix/pull/1390))
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


asomers (approved):
Interesting.  Did you see this happen on Ethernet, or with some other link layer?

Also, you'll have to rebase after #1399 merges.

internetionals:
> Interesting. Did you see this happen on Ethernet, or with some other link layer?

I saw this happen on Ethernet. I don't recall on which system I tested the original, might have been some virtual fabric.

> Also, you'll have to rebase after #1399 merges.

Ok

asomers (approved):
bors r+

PR #1397, opened 2021-03-04
A little bit of cleanup in errno.rs

asomers:


asomers:
Ack, my eyeballs deceived me.  There actually was a difference in those two lines; I just didn't see it.

PR #1402, opened 2021-03-20
Support TIMESTAMPNS

WiSaGaN:
This adds support of linux TIMESTAMPNS.
The code is mostly copied paste from https://github.com/nix-rust/nix/pull/663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Added TIMESTAMPNS support for linux
+  (#[1402](https://github.com/nix-rust/nix/pull/1402))
 
 ### Changed
 - Made `forkpty` unsafe, like `fork`
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { version = "0.2.82", features = [ "extra_traits" ] }
+libc = { version = "0.2.93", features = [ "extra_traits" ] }
 bitflags = "1.1"
 cfg-if = "1.0"
 
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -7,6 +7,8 @@ use libc::{self, c_void, c_int, iovec, socklen_t, size_t,
         CMSG_FIRSTHDR, CMSG_NXTHDR, CMSG_DATA, CMSG_LEN};
 use std::{mem, ptr, slice};
 use std::os::unix::io::RawFd;
+#[cfg(all(target_os = "linux"))]
+use crate::sys::time::TimeSpec;
 use crate::sys::time::TimeVal;
 use crate::sys::uio::IoVec;
 
@@ -554,6 +556,11 @@ pub enum ControlMessageOwned {
     /// # }
     /// ```
     ScmTimestamp(TimeVal),
+    /// Nanoseconds resolution timestamp
+    ///
+    /// [Further reading](https://www.kernel.org/doc/html/latest/networking/timestamping.html)
+    #[cfg(all(target_os = "linux"))]
+    ScmTimestampns(TimeSpec),
     #[cfg(any(
         target_os = "android",
         target_os = "ios",
@@ -645,6 +652,11 @@ impl ControlMessageOwned {
                 let tv: libc::timeval = ptr::read_unaligned(p as *const _);
                 ControlMessageOwned::ScmTimestamp(TimeVal::from(tv))
             },
+            #[cfg(all(target_os = "linux"))]
+            (libc::SOL_SOCKET, libc::SCM_TIMESTAMPNS) => {
+                let ts: libc::timespec = ptr::read_unaligned(p as *const _);
+                ControlMessageOwned::ScmTimestampns(TimeSpec::from(ts))
+            }
             #[cfg(any(
                 target_os = "android",
                 target_os = "freebsd",
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -1535,3 +1535,111 @@ pub fn test_vsock() {
     close(s1).unwrap();
     thr.join().unwrap();
 }
+
+// Disable the test on emulated platforms because it fails in Cirrus-CI.  Lack of QEMU
+// support is suspected.
+#[cfg_attr(not(any(target_arch = "x86_64")), ignore)]
+#[cfg(all(target_os = "linux"))]
+#[test]
+fn test_recvmsg_timestampns() {
+    use nix::sys::socket::*;
+    use nix::sys::uio::IoVec;
+    use nix::sys::time::*;
+    use std::time::*;
+
+    // Set up
+    let message = "Ohayō!".as_bytes();
+    let in_socket = socket(
+        AddressFamily::Inet,
+        SockType::Datagram,
+        SockFlag::empty(),
+        None).unwrap();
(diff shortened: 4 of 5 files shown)


asomers (changes requested):
It looks pretty good so far.  But why did you decide to write the test without the usual test harness?  It's preferable to use the regular test harness if possible.  In fact, right now your test won't even be compiled, because you didn't add it to Cargo.toml.

WiSaGaN:
> It looks pretty good so far. But why did you decide to write the test without the usual test harness? It's preferable to use the regular test harness if possible. In fact, right now your test won't even be compiled, because you didn't add it to Cargo.toml.

Thanks for the review! Is there a way to use the macro `cmsg_space` in unit tests? It contains `use unix::...`, which I can't make it work within a unit test.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Added linux TIMESTAMPNS support for x86/x64 gnu
```
asomers, line 8: Why only x86_64 and x86?  It looks like that symbol is defined for many other arches, too.

asomers (commented):
I assume you mean `cmsg_space!`?  It's currently used by the functional tests in test/sys/test_socket.rs.  There aren't any unit tests that currently use it, but I don't see why there couldn't be.  What problem are you having?

WiSaGaN:
> I assume you mean `cmsg_space!`? It's currently used by the functional tests in test/sys/test_socket.rs. There aren't any unit tests that currently use it, but I don't see why there couldn't be. What problem are you having?

It contains self-referential "use unix::*" statement, which can't be used in unit tests.

CHANGELOG.md
WiSaGaN: Even now arch like `Linux i686` is still failing. I am wondering why.

asomers:
> > I assume you mean `cmsg_space!`? It's currently used by the functional tests in test/sys/test_socket.rs. There aren't any unit tests that currently use it, but I don't see why there couldn't be. What problem are you having?
> 
> It contains self-referential "use unix::*" statement. Apparently, when in "nix" crate, you can only reference it as "use crate::..." but not "use nix:...".

Well, that sounds like something you can fix.

WiSaGaN:
> > > I assume you mean `cmsg_space!`? It's currently used by the functional tests in test/sys/test_socket.rs. There aren't any unit tests that currently use it, but I don't see why there couldn't be. What problem are you having?
> > 
> > 
> > It contains self-referential "use unix::*" statement. Apparently, when in "nix" crate, you can only reference it as "use crate::..." but not "use nix:...".
> 
> Well, that sounds like something you can fix.

I think it would be better to do that in a separate pull request. In the mean time, I can add a todo of moving the integration test to unit test after that is done.

Cargo.toml
```diff
@@ -75,3 +75,7 @@ harness = false
 [[test]]
 name = "test-ptymaster-drop"
 path = "test/test_ptymaster_drop.rs"
+
+[[test]]
+name = "test-receive-timestampns"
```
asomers, line 80: Why put it in its own test program, as opposed to the main "test" executable?  Only tests that require their own process should get their own test programs.  It's fine for the test to be a functional test; I don't think there's any reason to prefer a unit test instead.  But it shouldn't have its own test program.

CHANGELOG.md
WiSaGaN: It seems that even i686 linux does not have that symbol. Neither did mips, etc...

I have restricted to x86_64, and now the tests pass.

src/sys/socket/mod.rs
```diff
@@ -554,6 +556,11 @@ pub enum ControlMessageOwned {
     /// # }
     /// ```
     ScmTimestamp(TimeVal),
+    /// Nanoseconds resolution timestamp
+    ///
+    /// [Further reading](https://www.kernel.org/doc/html/latest/networking/timestamping.html)
+    #[cfg(all(target_os = "linux", any(target_arch = "x86_64"), target_env = "gnu"))]
```
asomers, line 562: You should not restrict the target_arch to just x86_64.  libc defines SCM_TIMESTAMPNS for many architectures.

asomers (changes requested):

Cargo.toml
WiSaGaN: This is the CI result when the conditional compilation was turned off: https://cirrus-ci.com/build/5978723427024896

I think there are cases where the symbols are added, but it was not supported: https://cirrus-ci.com/task/5721153533640704?command=main#L510

src/sys/socket/mod.rs
WiSaGaN: This is the CI result when the conditional compilation was turned off: https://cirrus-ci.com/build/5978723427024896

I think there are cases where the symbols are added, but it was not supported: https://cirrus-ci.com/task/5721153533640704?command=main#L510

asomers: It may be that Linux supports that socket option, but QEMU does not.  Looking at the Linux source code, I think it should be supported everywhere.

WiSaGaN: I can add more arch if you have concrete ones in mind. Otherwise I think it makes sense to merge this as MVP since x86_64 covers major use cases, and follow up with other archs.

asomers: You should add every arch that is supported by libc.  Don't be deterred by the fact that you can't run it in CI; some architectures don't have any CI coverage at all.  You have a working test case and there is no reason to believe that the feature would work differently on other architectures.

WiSaGaN: Adding those would break CI, which prevents us from getting breaking warnings for future changes for those archs. Are you sure that is what we want?

asomers: I'm not asking you to break CI.  I'm asking you to get the feature working on all platforms.  But if QEMU doesn't support it, then disable the test (not the feature) on those platforms.  There are already plenty of examples of tests that are disabled on QEMU platforms.

WiSaGaN:
`libc` recently updated its socketopt consts: https://github.com/rust-lang/libc/commit/682eba61169005a6d904fc53fc749788f7e3cf9b

We will wait for the release of this to have a more updated lists of arch here.

WiSaGaN:
After upgrade libc to 0.2.93, now the support has been added to all linux platform. Tests however are restricted to x86-64, non musl due to suspected QEMU issue.

WiSaGaN:
Please feel free to squash all the commits into one if it is ready to merge.

test/sys/test_socket.rs
```diff
@@ -1535,3 +1535,114 @@ pub fn test_vsock() {
     close(s1).unwrap();
     thr.join().unwrap();
 }
+
+// Disable the test on emulated platforms because it fails in Cirrus-CI.  Lack of QEMU
+// support is suspected.
+#[cfg_attr(any(not(any(target_arch = "x86_64")), any(target_env = "musl")), ignore)]
```
asomers, line 1541: This can't be right.  CI doesn't use QEMU for x86-64-linux-musl.  What error do you get on that platform?

WiSaGaN: Fixed the leftover from before the libc upgrade.

asomers:
Ok, this looks good now.  But you'll have to squash yourself.  Nix uses the bors merge bot, which is incompatible with Github's squash-and-merge button.

WiSaGaN:
Sqaushed. Now it is ready to merge.

asomers (approved):
bors r+

Issue #1403, opened 2021-03-21
test_vsock fails on Linux under QEMU


asomers:
Sometime in between 6-March-2021 and 16-March-2021, `sys::test_socket::test_vsock` has begin to fail for the QEMU architectures (but it still passes on x86_64 Linux).

For example, from https://github.com/nix-rust/nix/pull/1401/checks?check_run_id=2124256635 ,
```
---- sys::test_socket::test_vsock stdout ----
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some(Sys(EADDRNOTAVAIL))`', test/sys/test_socket.rs:1517:5
```
@stefano-garzarella @qwandor-google as the only two known users of Vsock with Nix, can one of you please debug this?

asomers:
Possibly the same root cause as #1310

stefano-garzarella:
@asomers sure! I can take a look!

stefano-garzarella:
The test is failing here:
```rust
    let sockaddr = SockAddr::new_vsock(libc::VMADDR_CID_LOCAL, port);
    assert_eq!(bind(s1, &sockaddr).err(),
               Some(Error::Sys(Errno::EADDRNOTAVAIL)));
```

Starting from Linux 5.6, `VMADDR_CID_LOCAL` is supported to do local communication (loopback device).
Before this it was reserved, but probably the test runner now has a newer kernel and `VMADDR_CID_LOCAL` is supported.

I'll send a patch right away to remove that check and re-enable the test.

asomers:
Fixed by #1407 .

Issue #1310, opened 2020-10-10
Test sys::test_socket::test_vsock fails on Ubuntu 20.10/AMD64 beta


tommythorn:
Tested nix@566e2d7dad025c9b2bef98115b584328c145675d using rust 1.47 (stable) on Ubuntu 20.04/AMD64 (up to date at time of testing)

`cargo test` hangs on `test_af_alg_aead` (to be file after this one).  Once commented out, it fails with
```
failures:

---- sys::test_socket::test_vsock stdout ----
thread 'sys::test_socket::test_vsock' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some(Sys(EADDRNOTAVAIL))`', test/sys/test_socket.rs:1440:5
```

SteveLauC:
Closing as this has been fixed.

PR #1407, opened 2021-03-22
Fix test_vsock failure on VMADDR_CID_LOCAL testing and re-enable it

stefano-garzarella:
Starting from Linux 5.6, `VMADDR_CID_LOCAL` is supported to do local communication (loopback device).

Before Linux 5.6 it was called `VMADDR_CID_RESERVED` and was not supported, so we could expect an `EADDRNOTAVAIL`, but now this address is supported and handled by the 'vsock_loopback' kernel module loaded automatically if no other vsock transports are loaded.

Issue #1310
Issue #1403

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -1508,16 +1508,10 @@ pub fn test_vsock() {
                     SockFlag::empty(), None)
              .expect("socket failed");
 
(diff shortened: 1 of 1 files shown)


qwandor-google (approved):

asomers (approved):
Thanks for the quick fix!

bors r+

asomers:
bors retry

PR #1409, opened 2021-03-25
unistd: Increase maximum passwd/group buffer to 1MB

geofft:
We have one UNIX group that contains most of our users whose size is
about 20 kB, so `Group::from_name` is failing with ERANGE.

The discussion on PR #864 suggests that 1 MB is a reasonable maximum -
it follows what FreeBSD's libc does. (glibc appears to have no maximum
on the _r function and will just double the buffer until malloc fails,
but that's not particularly Rusty.)

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -2641,10 +2641,10 @@ impl User {
               libc::size_t,
               *mut *mut libc::passwd) -> libc::c_int
     {
-        let buflimit = 16384;
+        let buflimit = 1048576;
         let bufsize = match sysconf(SysconfVar::GETPW_R_SIZE_MAX) {
             Ok(Some(n)) => n as usize,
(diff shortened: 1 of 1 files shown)


geofft:
Happy to add a changelog entry if this needs one.

Also, the way this code currently works is that if the platform supports `sysconf(_GETPW_R_SIZE_MAX)`, it will use that as the initial buffer size, otherwise it will use the maximum. So on platforms without `_GETPW_R_SIZE_MAX`, we'll start out allocating a 1MB buffer. If that's unwanted, I can change it to default to start allocating a 16kB buffer, doubling until it hits 1MB.

asomers (changes requested):
Yeah, we probably don't want to start out at 1MB.  That could take a little bit of time to malloc.

asomers (approved):
bors r+

PR #1410, opened 2021-03-26
add libc::IP6T_SO_ORIGINAL_DST to socket opt

Icemic:
In order to support ipv6 address

ex.

```rust
let addr = getsockopt(fd.as_raw_fd(), Ip6tOriginalDst)
```


asomers (changes requested):
Needs a CHANGELOG entry

PR #1414, opened 2021-04-02
Fix corrupted sendmmsg() call

eaufavor:
Before this fix, the buffer that holds cmsgs may move due to the resize()
call. That causes msg_hdr pointing to invalid memory, which ends up
breaking the sendmmsg() call, resulting in EINVAL.

This change fixes it by avoiding re-allocating the buffers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Allow `sockaddr_ll` size, as reported by the Linux kernel, to be smaller then it's definition
   (#[1395](https://github.com/nix-rust/nix/pull/1395))
+- Fix spurious errors using `sendmmsg` with multiple cmsgs
+  (#[1414](https://github.com/nix-rust/nix/pull/1414))
 
 ### Removed
 
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1101,23 +1101,22 @@ pub fn sendmmsg<'a, I, C>(
 
     let mut output = Vec::<libc::mmsghdr>::with_capacity(reserve_items);
 
-    let mut cmsgs_buffer = vec![0u8; 0];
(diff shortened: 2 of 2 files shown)


asomers-ax (commented):
Good find!  Would you mind adding a CHANGELOG entry?

eaufavor:
Added, thanks.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Allow `sockaddr_ll` size, as reported by the Linux kernel, to be smaller then it's definition
   (#[1395](https://github.com/nix-rust/nix/pull/1395))
+- Fix corrupted `sendmmsg` call
+  (#[1395](https://github.com/nix-rust/nix/pull/1414))
```
asomers, line 17: Typo in the link.

asomers (changes requested):

CHANGELOG.md
eaufavor: copypasta, fixed, thanks.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Allow `sockaddr_ll` size, as reported by the Linux kernel, to be smaller then it's definition
   (#[1395](https://github.com/nix-rust/nix/pull/1395))
+- Fix corrupted `sendmmsg` call
```
asomers, line 16: Could you be more specific?  Something like "Fix spurious errors using `sendmmsg` with multiple and/or long cmsgs`?

asomers (commented):
Looks good, but please squash your commits.

CHANGELOG.md
eaufavor: updated

eaufavor:
Squashed with the updated commit message

asomers (approved):
bors r+

asomers:
bors retry

PR #1418, opened 2021-04-06
Add new symbol for linux IP_MULTICAST_ALL

### LINKED ISSUE #1199: Support for IP_MULTICAST_ALL
I would like to request a friendly binding for the socket option IP_MULTICAST_ALL.

This was added in version 2.67 of [libc](https://github.com/rust-lang/libc), in the following [libc PR](https://github.com/rust-lang/libc/pull/1696/commits/b7cb4d2c7b988ca0ce82fc43ee6484035a578cf0)

anth1y:
this closes #1199 
/cc @asomers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added TIMESTAMPNS support for linux
   (#[1402](https://github.com/nix-rust/nix/pull/1402))
 
+- Add new symbol for linux IP_MULTICAST_ALL
+  (#[1418](https://github.com/nix-rust/nix/pull/1418))
+
+- Added TIMESTAMPNS support for linux
+  (#[1402](https://github.com/nix-rust/nix/pull/1402))
+
+
 ### Changed
 - Made `forkpty` unsafe, like `fork`
   (#[1390](https://github.com/nix-rust/nix/pull/1390))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Add new symbol for linux IP_MULTICAST_ALL
   (#[1418](https://github.com/nix-rust/nix/pull/1418))
+- Added TIMESTAMPNS support for linux
+  (#[1402](https://github.com/nix-rust/nix/pull/1402))
 
 ### Changed
 - Made `forkpty` unsafe, like `fork`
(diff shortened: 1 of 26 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -238,6 +238,17 @@ cfg_if! {
 }
 sockopt_impl!(Both, IpMulticastTtl, libc::IPPROTO_IP, libc::IP_MULTICAST_TTL, u8);
 sockopt_impl!(Both, IpMulticastLoop, libc::IPPROTO_IP, libc::IP_MULTICAST_LOOP, bool);
+cfg_if! {
```
asomers, line 241: You shouldn't use `cfg_if` if you're only going to have one clause.  Instead, just do
```rust
#[cfg(target_os = "linux")]
whatever
```

src/sys/socket/sockopt.rs
```diff
@@ -238,6 +238,17 @@ cfg_if! {
...
+	 if #[cfg(any(target_os = "linux"))] {
```
asomers, line 242: The symbol is defined on Android, too.

asomers (changes requested):
You also need a CHANGELOG entry.  And if possible, a test.  I don't know what this option does.  Is it practical to add a test for it in Nix?

anth1y:
I dont think this needs a test but i'll defer to your judgement here:

```
       IP_MULTICAST_ALL (since Linux 2.6.31)
              This option can be used to modify the delivery policy of
              multicast messages to sockets bound to the wildcard
              INADDR_ANY address.  The argument is a boolean integer
              (defaults to 1).  If set to 1, the socket will receive
              messages from all the groups that have been joined
              globally on the whole system.  Otherwise, it will deliver
              messages only from the groups that have been explicitly
              joined (for example via the IP_ADD_MEMBERSHIP option) on
              this particular socket.
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Add new symbol for linux IP_MULTICAST_ALL
+  (#[1411](https://github.com/nix-rust/nix/pull/1411))
```
asomers, line 9: Looks like you copy/pasted the wrong PR number.

src/sys/socket/sockopt.rs
asomers: Did you forget to push this change?

asomers (changes requested):

asomers (commented):
Better.  Can you please squash your commits?  Then we'll be good to go.

anth1y:
going to close this my branch is super polluted ...

PR #1420, opened 2021-04-08
Support linux TIMESTAMPING

WiSaGaN:


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Added TIMESTAMPNS support for linux
   (#[1402](https://github.com/nix-rust/nix/pull/1402))
+- Added TIMESTAMPNING support for linux
+  (#[1402](https://github.com/nix-rust/nix/pull/1420))
```
asomers, line 11: Fix the PR number

src/sys/socket/mod.rs
```diff
@@ -562,6 +562,11 @@ pub enum ControlMessageOwned {
     /// [Further reading](https://www.kernel.org/doc/html/latest/networking/timestamping.html)
     #[cfg(all(target_os = "linux"))]
     ScmTimestampns(TimeSpec),
+    /// Configurable nanoseconds resolution timestamps
```
asomers, line 565: Could you please explain what the wrapped type is?  Why are there three timespecs?

WiSaGaN: https://www.kernel.org/doc/html/latest/networking/timestamping.html#scm-timestamping-records

asomers: A link to an external website is not sufficient documentation.  Why is the wrapped value an array?  What's special about "3"?

WiSaGaN: The link was specifically for your question. Have you checked it out?

asomers: No; I'm asking for Nix's own documentation to describe why the value is a 3-member array.  You can leave the external link in there, similar to how many Nix functions have man page links, but Nix's own documentation should have at least basic information about what the type is.

WiSaGaN: Ok. Added some documentation.

src/sys/socket/mod.rs
```diff
@@ -563,6 +563,8 @@ pub enum ControlMessageOwned {
     #[cfg(all(target_os = "linux"))]
     ScmTimestampns(TimeSpec),
     /// Configurable nanoseconds resolution timestamps
+    /// Three `TimeSpec`s are returned in the control message. At least one field is non-zero at
+    /// any time. Most timestamps are passed in ts[0]. Hardware timestamps are passed in ts[2].
```
asomers, line 567: So what is ts[1] for?  Also, the second line of a multiline doc comment like this should be blank.  That way, rustdoc will interpret the first line as a summary and the rest as extended details.

WiSaGaN: I don't think it is necessary to specify every nuances in this short description. User of this feature needs to read the original linux documentation to have a reasonable idea in how to set and what to set in order to get either software and hardware timestamp (in hardware case, user also need to initiate in different way according to different hardware), and to know the historical reason for `ts[1]` being deprecated. I don't see the point to just copy large paragraphs of the original text here.

asomers: ts[1] is deprecated?  In that case, there isn't any reason to expose it to the user at all.  You should remove it in `ControlMessageOwned::decode_from`.

WiSaGaN: This is a pretty highly specific interface. It is more likely the user has learned the original linux interface first and try to use it through `nix` crate than people directly learn it through `nix` itself. Thus, preserving the original interface as much as possible (close to 1-1 map) can reduce the amount of surprise and avoid possible footguns.

FloLo100:
Hi! Was there any particular reason this pull request was closed? I was looking forward to using the timestamping-interface to get access to outgoing timestamps. Would be glad to help with any outstanding problems! (we're a team of 3 Devs, but kinda new to Rust)

WiSaGaN:
@FloLo100 I don't see this pull request getting merged any time soon. But feel free to reuse/copy any of the code here if you think they are useful.

FloLo100:
@asomers is there anything beyond the documentation that's blocking the merge? We will try to fork this and merge it locally for now.

PR #1279, opened 2020-07-28
Add safe wrapper for `PTRACE_INTERRUPT`

GabrielMajeri:
Adds a new `ptrace:interrupt` function for stopping the running traced process. Also exposes the `PTRACE_EVENT_STOP` variant from the updated `libc`.

PR #1422, opened 2021-04-11
Add PTRACE_INTERRUPT call as `ptrace::interrupt(pid)` 

blaind:
I've based the test on `fn test_ptrace_cont`. Removed some parts, but not 100% sure what's the proper way of testing in nix context.

From ptrace-man page:

```
       PTRACE_INTERRUPT (since Linux 3.4)
              Stop a tracee.  If the tracee is running or sleeping in
              kernel space and PTRACE_SYSCALL is in effect, the system
              call is interrupted and syscall-exit-stop is reported.
              (The interrupted system call is restarted when the tracee
              is restarted.)  If the tracee was already stopped by a
              signal and PTRACE_LISTEN was sent to it, the tracee stops
              with PTRACE_EVENT_STOP and WSTOPSIG(status) returns the
              stop signal.  If any other ptrace-stop is generated at the
              same time (for example, if a signal is sent to the
              tracee), this ptrace-stop happens.  If none of the above
              applies (for example, if the tracee is running in user
              space), it stops with PTRACE_EVENT_STOP with
              WSTOPSIG(status) == SIGTRAP.  PTRACE_INTERRUPT only works
              on tracees attached by PTRACE_SEIZE.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `sendfile64` (#[1439](https://github.com/nix-rust/nix/pull/1439))
 - Added `MS_LAZYTIME` to `MsFlags`
   (#[1437](https://github.com/nix-rust/nix/pull/1437))
+- Added `ptrace::interrupt` method for platforms that support `PTRACE_INTERRUPT`
+  (#[1422](https://github.com/nix-rust/nix/pull/1422))
 
 ### Changed
 - Made `forkpty` unsafe, like `fork`
--- a/test/sys/test_ptrace.rs
+++ b/test/sys/test_ptrace.rs
@@ -114,6 +114,48 @@ fn test_ptrace_cont() {
     }
 }
 
+#[cfg(target_os = "linux")]
+#[test]
+fn test_ptrace_interrupt() {
+    use nix::sys::ptrace;
+    use nix::sys::signal::Signal;
+    use nix::sys::wait::{waitpid, WaitPidFlag, WaitStatus};
+    use nix::unistd::fork;
+    use nix::unistd::ForkResult::*;
+    use std::thread::sleep;
+    use std::time::Duration;
+
+    require_capability!(CAP_SYS_PTRACE);
+
+    let _m = crate::FORK_MTX.lock().expect("Mutex got poisoned by another test");
+
+    match unsafe{fork()}.expect("Error: Fork Failed") {
+        Child => {
+            loop {
+                sleep(Duration::from_millis(1000));
+            }
+
+        },
+        Parent { child } => {
+            ptrace::seize(child, ptrace::Options::PTRACE_O_TRACESYSGOOD).unwrap();
+            ptrace::interrupt(child).unwrap();
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::PtraceEvent(child, Signal::SIGTRAP, 128)));
+            ptrace::syscall(child, None).unwrap();
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::PtraceSyscall(child)));
+            ptrace::detach(child, Some(Signal::SIGKILL)).unwrap();
+            match waitpid(child, None) {
+                Ok(WaitStatus::Signaled(pid, Signal::SIGKILL, _)) if pid == child => {
+                    let _ = waitpid(child, Some(WaitPidFlag::WNOHANG));
+                    while ptrace::cont(child, Some(Signal::SIGKILL)).is_ok() {
+                        let _ = waitpid(child, Some(WaitPidFlag::WNOHANG));
+                    }
+                }
+                _ => panic!("The process should have been killed"),
+            }
+        },
+    }
+}
+
 // ptrace::{setoptions, getregs} are only available in these platforms
 #[cfg(all(target_os = "linux",
           any(target_arch = "x86_64",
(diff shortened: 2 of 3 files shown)


test/sys/test_ptrace.rs
```diff
@@ -115,6 +115,45 @@ fn test_ptrace_cont() {
...
+        Child => {
+            loop {
+                sleep(Duration::from_millis(1000));
+            }
+
+        },
+        Parent { child } => {
+            ptrace::seize(child, ptrace::Options::PTRACE_O_TRACESYSGOOD).unwrap();
+            ptrace::interrupt(child).unwrap();
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::PtraceEvent(child, Signal::SIGTRAP, 128)));
```
blaind, line 140: This I verified, interrupt will fire the PtraceEvent. Not sure about the third parameter 128, is it same over all systems?

test/sys/test_ptrace.rs
```diff
@@ -115,6 +115,45 @@ fn test_ptrace_cont() {
...
+            ptrace::syscall(child, None).unwrap();
+            assert_eq!(waitpid(child, None), Ok(WaitStatus::PtraceSyscall(child)));
```
blaind, line 142: ptrace::syscall and waitpid with waitstatus::ptracesyscall are "extraneous", but could not get the kill/detach working without adding these

test/sys/test_ptrace.rs
```diff
@@ -115,6 +115,49 @@ fn test_ptrace_cont() {
     }
 }
 
+// ptrace::interrupt is only available in these platforms
+#[cfg(all(target_os = "linux", not(any(target_arch = "mips", target_arch = "mips64"))))]
```
asomers, line 119: Really?  libc defines it for all architectures.

blaind: See https://github.com/nix-rust/nix/blob/68d01f0426a82b485b22c31425c0e53b5c372a7d/src/sys/ptrace/linux.rs#L105

asomers: Yes I know.  But you added that line.  So why did you leave out mips?

blaind: I mean, my rationale for adding the line (at `test/sys/test_ptrace.rs` is because this pull request depends on `nix::Request::PTRACE_INTERRUPT` to be available. 

The test would fail without me adding that limitation (which is similar than for PTRACE_INTERRUPT). For why that has been added for Request::PTRACE_INTERRUPT initially, I have no idea.

asomers: My apologies.  I saw that you added `fn interrupt`, and didn't realize that `PTRACE_INTERRUPT` already existed.  But none the less, I can see no reason to restrict it on mips.  Could you please remove the restriction?

asomers:
@blaind could you please fix the compile failures?

blaind:
> @blaind could you please fix the compile failures?

I had a look, the builds are failing now because of missing variables on various platforms. On android, `libc::{PTRACE_SEIZE, PTRACE_INTERRUPT};` are missing, and on *BSD (mips?) a lot more failures.

I guess one way to fix is to go back with feature gate (` #[cfg(all(target_os = "linux", not(any(target_arch = "mips", target_arch = "mips64"))))] `). Any other ideas?

asomers:
It's failing on non-Linux platforms because stuff like PTRACE_SEIZE is Linux-specific.  That's why they were feature gated to begin with.  I only asked about removing the mips restriction, not the OS restriction.

blaind:
> It's failing on non-Linux platforms because stuff like PTRACE_SEIZE is Linux-specific. That's why they were feature gated to begin with. I only asked about removing the mips restriction, not the OS restriction.

I mistakenly took architecture to mean OS as well. Anyway, just pushed a new commit that might fix it. Testing is semiwhat slow, as I can not test locally.

test/sys/test_ptrace.rs
asomers: Good question.  Where did you copy the 128 parameter from?

blaind: I took it from the `waitpid(...)` return state. Seems that internally it converts `status` through
```
fn stop_additional(status: i32) -> c_int {
    (status >> 16) as c_int
}
```

ranweiler:
Note, this is a duplicate of #1279 (which omits tests). Whichever PR gets merged, the other should be closed.

asomers (changes requested):
Could you please rebase onto master to fix the conflicts?

blaind:
> Could you please rebase onto master to fix the conflicts?

Done

asomers (changes requested):
Ok, everything looks good now.  Would you mind squashing your commits?

CHANGELOG.md
```diff
@@ -90,6 +90,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `sendfile64` (#[1439](https://github.com/nix-rust/nix/pull/1439))
 - Added `MS_LAZYTIME` to `MsFlags`
   (#[1437](https://github.com/nix-rust/nix/pull/1437))
+- Added `ptrace::interrupt` method for platforms that support `PTRACE_INTERRUPT`
```
asomers, line 93: Shoot.  I just noticed that your CHANGELOG entry got moved into the wrong part of the file when you rebased.  You'll have to move it up to the top.

asomers (changes requested):
bors r+

CHANGELOG.md
asomers: Whoops, I forgot to stop bors when I found the CHANGELOG.  problem.  Don't worry; I'll take care of it.

blaind: Oops, sorry about that. Did not notice when rebasing with master. Thanks for fixing!

PR #1426, opened 2021-04-15
Support GETFPREGS and SETFPREGS ptrace calls

nemarci:
Added two new functions to the `ptrace` module: `getfpregs` and `setfpregs` gets and sets the floating point registers of the tracee.

### DIFF
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -16,7 +16,7 @@ pub type AddressType = *mut ::libc::c_void;
             any(target_env = "gnu", target_env = "musl")),
         all(target_arch = "x86", target_env = "gnu"))
 ))]
-use libc::user_regs_struct;
+use libc::{user_regs_struct, user_fpregs_struct};
 
 cfg_if! {
     if #[cfg(any(all(target_os = "linux", target_arch = "s390x"),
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
Would it be possible to add a test?

nemarci:
> Would it be possible to add a test?

Yes, sure. What kind of test do you have in mind? I see that most ptrace related tests just call `ptrace` and then check that it doesn't return `Error::UnsupportedOperation`. Would a similar check be sufficient, or would you like some more thorough testing? (e.g. creating a subprocess, then modify one of the XMM reigsters with `setfpregs`, and reading it with `getfpregs` to check that we successfully modified it).

asomers:
> > Would it be possible to add a test?
> 
> Yes, sure. What kind of test do you have in mind? I see that most ptrace related tests just call `ptrace` and then check that it doesn't return `Error::UnsupportedOperation`. Would a similar check be sufficient, or would you like some more thorough testing? (e.g. creating a subprocess, then modify one of the XMM reigsters with `setfpregs`, and reading it with `getfpregs` to check that we successfully modified it).

That would be perfect.

nemarci:
Two of the automatic tests failed, but as far as I understand, these failures are unrelated to my patch. Could you verify if this is indeed true? And if I'm wrong, could you give some advice how to fix the failures?

asomers:
Thanks for the very good tests!  As for the test failures, both should be fixed if you rebase your changes onto master.

asomers:
Looks good, but I forgot one thing: this needs a CHANGELOG entry.  And you can go ahead and squash while you're at it.

nemarci:
Closing in favor of https://github.com/nix-rust/nix/pull/1844

PR #1303, opened 2020-10-03
Renable the ScmTimestamp test on FreeBSD i386

asomers:
The relevant bug was fixed in 11.3-RELEASE

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=222039

PR #1427, opened 2021-04-19
Fix timestamp doc

WiSaGaN:
This fixed doc test issue here: https://docs.rs/nix/0.20.0/nix/sys/socket/enum.ControlMessageOwned.html#variant.ScmTimestamp


asomers (approved):
Ahh, good catch.  Looks like it's my fault, from PR #1303 .

bors r+

PR #1428, opened 2021-04-19
Add support for TTL control messages

pmalmgren:
This patch adds support for reading the IP `ttl` header from a control message if `IP_RECVTTL` is set on the socket.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -609,6 +609,11 @@ pub enum ControlMessageOwned {
     #[cfg(target_os = "linux")]
     UdpGroSegments(u16),
 
+    /// IP Time-to-live (IP_TTL) is the value of the 8-bit TTL header
+    /// in an ipv4 packet.
+    #[cfg(target_os = "linux")]
+    IpTtl(u8),
+
     /// Catch-all variant for unimplemented cmsg types.
     #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


asomers (commented):
So this patch adds support for two new types of cmsg?  How does one request receipt of those cmsg types?

pmalmgren:
@asomers Correct, it adds support for two new types of cmsg. They are requested with `setsockopt` and require new socket options. I'll add a commit for those to this PR.

cemeyer:
N.B., #1161 also attempts to add IP_RECVERR support.  I have not looked at either PR's changes yet, and the other PR doesn't attempt to cover IP_RECVTTL.

src/sys/socket/mod.rs
```diff
@@ -430,6 +431,12 @@ macro_rules! cmsg_space {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct RecvErr {
+    pub ee: sock_extended_err,
+    pub offender: sockaddr_in,
```
cemeyer, line 437: N.B., the similar `IPV6_RECVERR` option produces `sockaddr_in6`es on IPv6 sockets.  Regardless of which patch we take for this, it should probably have IPv6 <-> v4 parity.

PR #1430, opened 2021-04-23
Add getresuid() and getresgid() to unistd

jerylvaz:


test/test_unistd.rs
```diff
@@ -624,6 +624,27 @@ fn test_sysconf_unsupported() {
     assert!(open_max.expect("sysconf failed").is_none())
 }
 
+
+#[cfg(target_os = "linux")]
```
asomers, line 628: libc also implements these functions on Android.  But they're actually implemented on some of the BSDs, too.  Best would be to make a PR to libc that adds bindings everywhere they're implemented, then broaden your `cfg` restriction in Nix appropriately.

test/test_unistd.rs
```diff
@@ -624,6 +624,27 @@ fn test_sysconf_unsupported() {
...
+#[test]
+fn test_getresuid() {
+    let resuid = getresuid().unwrap();
+    // Assume that tests are never run as root
```
asomers, line 632: LOL!  Of course the tests sometimes run as root!  Some tests only work when run as root!

asomers (changes requested):

jerylvaz:
@asomers  What should be done with the unrelated test failure?

src/unistd.rs
```diff
@@ -2524,6 +2527,51 @@ mod setres {
     }
 }
 
+#[cfg(any(target_os = "android", target_os = "linux", target_os = "openbsd"))]
```
asomers, line 2530: This will fail to compile for OpenBSD.  You need to open a PR with https://github.com/rust-lang/libc first.

asomers (changes requested):

src/unistd.rs
```diff
@@ -2524,6 +2527,51 @@ mod setres {
...
+    ///
+    /// ([see getresuid(2)](http://man7.org/linux/man-pages/man2/getresuid.2.html))
+    ///
+    /// #Returns
+    ///
+    /// - `Ok((Uid, Uid, Uid))`: tuple of real, effective and saved uids on success.
+    /// - `Err(x)`: libc error code on failure.
+    ///
+    #[inline]
+    pub fn getresuid() -> Result<(Uid, Uid, Uid)> {
```
asomers, line 2546: Returning a tuple is a very old fashioned and failure-prone technique.  Better to return a structure.  That way the meaning is obvious even without consulting the documentation.  Something like:
```rust
struct ResUid {
    real: Uid,
    effective: Uid,
    saved: uid
}
```

asomers:
Looks good, but you'll have to rebase to fix the CI failure on FreeBSD.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - ReleaseDate
+### Added
```
asomers, line 7: Please fix the CHANGELOG.  There was a merge error, and now there are two "Added" sections.

asomers (changes requested):

asomers (approved):
bors r+

PR #1433, opened 2021-05-11
Use fstatat to check long path sizes in fcntl::readlinkat

zxzax:
Without this, the `lstat` call ignores the `dirfd` and checks the wrong file. I wanted to write a unit test for this but I'm not sure how to do it reliably, AFAIK the maximum path size can vary between filesystems.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -252,7 +252,18 @@ fn inner_readlink<P: ?Sized + NixPath>(dirfd: Option<RawFd>, path: &P) -> Result
     }
     // Uh oh, the result is too long...
     // Let's try to ask lstat how many bytes to allocate.
-    let reported_size = super::sys::stat::lstat(path)
+    let reported_size = match dirfd {
+        #[cfg(target_os = "redox")]
+        Some(_) => unreachable!(),
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        Some(dirfd) => {
+            let flags = if path.is_empty() { AtFlags::AT_EMPTY_PATH } else { AtFlags::empty() };
+            super::sys::stat::fstatat(dirfd, path, flags | AtFlags::AT_SYMLINK_NOFOLLOW)
+        },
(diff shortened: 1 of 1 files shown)


asomers:
Can you explain in a little more detail what the problem is?  Or better yet, show that unit test you came up with, even if it doesn't work everywhere.

zxzax:
The `readlink` and `readlinkat` syscalls read out a symlink to a fixed size buffer. If the buffer size is too small, then the syscall has to be performed again with a larger buffer. As a first step to mitigate this, `inner_readlink` will try to call `lstat` to read the size of the symlink from the filesystem. If the `lstat` call fails then it will fall back to enlarging the buffer in a loop and retrying until it succeeds.

This will work fine if we called it as `fcntl::readlink`, but the problem comes in if this is a call to `fcntl::readlinkat`. The `readlinkat` syscall expects a dirfd and a relative path. If we only call `lstat` then the dirfd is ignored and `lstat` will be trying to access the wrong file. On second thought, it's not really feasible to write a unit test for this since this can never cause the whole function to fail, it just causes it to be slower in some circumstances, since it will always enter the loop if the `lstat` call fails. The solution to this is: When we have a dirfd, use the corresponding stat-family syscall that takes a dirfd, which is `fstatat`.

There is another special case on Linux, where `readlinkat` will read the symlink straight from the fd if it's passed an empty string as a path. `fstatat` has similar behavior, but only if it's passed an `AT_EMPTY_PATH` flag, so we have to reconcile that here too.

Reference:
- https://man7.org/linux/man-pages/man2/readlinkat.2.html
- https://man7.org/linux/man-pages/man2/fstatat.2.html

asomers (approved):
This looks good. I just need to figure out why CI is hanging.

asomers:
CI is hanging due to a regression in rustup. https://github.com/rust-lang/rustup/issues/2774

asomers:
If you rebase on main, CI should be fixed.

asomers (approved):
bors r+

PR #1434, opened 2021-05-14
Rewrite links to use https:// instead of http://

rtzoeller:
All of the `http://` links in the repository are also available as `https://, and have been updated.

The sole exception to this trivially rewrite is a dead-link to `http://rust-lang.org/COPYRIGHT`, which has been replaced with `https://www.rust-lang.org/policies/licenses`.


asomers (approved):
Thanks!

bors r+

PR #1435, opened 2021-05-15
Add `sched_setscheduler` and associated functions

rtzoeller:
Add support for the following functions on Linux and Android:
- [`sched_setscheduler(2)`](https://man7.org/linux/man-pages/man2/sched_setscheduler.2.html)
- [`sched_getscheduler(2)`](https://man7.org/linux/man-pages/man2/sched_getscheduler.2.html)
- [`sched_setparam(2)`](https://man7.org/linux/man-pages/man2/sched_setparam.2.html)
- [`sched_getparam(2)`](https://man7.org/linux/man-pages/man2/sched_getparam.2.html)
- [`sched_get_priority_min(2)`](https://man7.org/linux/man-pages/man2/sched_get_priority_min.2.html)
- [`sched_get_priority_max(2)`](https://man7.org/linux/man-pages/man2/sched_get_priority_max.2.html)

`SCHED_DEADLINE` is defined for Android for `SchedFlags` completeness, however [`sched_setattr(2)`](https://man7.org/linux/man-pages/man2/sched_setattr.2.html) support is not included in this change.


src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
...
+            SCHED_RESET_ON_FORK;
+        }
+    }
+
+    #[repr(C)]
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+    pub struct SchedParam(libc::sched_param);
+
+    impl SchedParam {
+        pub fn new(priority: i32) -> Self {
```
rtzoeller, line 241: I'm fairly confident priorities can't be negative (`-1` is used as an error value in some functions that return priorities), but I haven't seen it explicitly stated.

I considered using a `u32` here (and am open to doing so if that's the direction the maintainers would like to go), but switching to a `u32` adds a new failure mode for values larger than `i32::MAX` that we'd need to manually handle.

src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
...
+            SCHED_BATCH;
+            SCHED_IDLE;
+            #[cfg(target_os = "android")]
+            SCHED_DEADLINE;
+            #[cfg(target_os = "linux")]
+            SCHED_RESET_ON_FORK;
+        }
+    }
+
+    #[repr(C)]
```
asomers, line 236: The distinction is subtle, but this should be `repr(transparent)` instead.  See https://doc.rust-lang.org/1.26.2/unstable-book/language-features/repr-transparent.html .

src/sched.rs
asomers: No, i32 is better, because that's what the C functions return.

src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
...
+            sched_param.sched_priority = priority;
+            SchedParam(sched_param)
+        }
+
+        pub fn priority(&self) -> i32 {
+            self.0.sched_priority
+        }
+    }
+
+    impl Default for SchedParam {
```
asomers, line 252: You should be able to omit this part, and just add `#[derive(Default)]` to the struct.

src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
...
+        Errno::result(res)
+    }
+
+    /// Set thread's scheduling parameters
+    ///
+    /// `pid` is the thread ID to update.
+    /// If `pid` is zero, then the parameters for the calling thread are set.
+    ///
+    /// See also [`sched_setparam(2)`](https://man7.org/linux/man-pages/man2/sched_setparam.2.html)
+    pub fn sched_setparam(pid: Pid, sched_param: &SchedParam) -> Result<()> {
```
asomers, line 282: Since `SchedParam` is `Copy` and a mere 32-bits, it is more efficient to pass by value than by reference.

asomers, line 282: If pid 0 has a special meaning for the C function, then a Rustier way to bind is is for the argument to be of type `Option<Pid>`.  See for example https://docs.rs/nix/0.20.0/nix/unistd/fn.getsid.html .

src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
...
+    ///
+    /// `pid` is the thread ID to check.
+    /// If `pid` is zero, then the parameters for the calling thread are retrieved.
+    ///
+    /// See also [`sched_getparam(2)`](https://man7.org/linux/man-pages/man2/sched_getparam.2.html)
+    pub fn sched_getparam(pid: Pid) -> Result<SchedParam> {
+        let mut sched_param = mem::MaybeUninit::uninit();
+        let res = unsafe { libc::sched_getparam(pid.into(), sched_param.as_mut_ptr()) };
+
+        Errno::result(res).and(Ok(unsafe { SchedParam(sched_param.assume_init()) }))
```
asomers, line 298: I think the `and(Ok` could be shortened to `map(`

asomers (changes requested):

src/sched.rs
rtzoeller: Should `nix::sched::CpuSet` also be updated to use `repr(transparent)`? I based the implementation for `SchedParam` off of it.

If so, I'm happy to open a separate PR making that change.

src/sched.rs
rtzoeller: I agree that's a Rustier way to do it. That said, other functions in `nix::sched` such as [`sched_setaffinity`](https://docs.rs/nix/0.20.0/nix/sched/fn.sched_setaffinity.html) take raw `Pid` types -- I based most of the implementation for the new `sched_*` functions off of the existing ones.

I've got no problem switching to `Option<Pid>` here, but thought I'd point out the inconsistency before making changes.

src/sched.rs
asomers: Yeah, pretty much every Newtype should use `repr(transparent)`.  It wasn't available in older compilers, which is why some parts of Nix don't use it.

src/sched.rs
rtzoeller: `libc::unix::linux_like::sched_param` does not implement `std::default::Default`, so this fails with a compiler error.

src/sched.rs
rtzoeller: Is this something you'd accept a PR for? I.e. find-and-replacing `repr(c)` with `repr(transparent)`?

src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
 
         Errno::result(res).map(drop)
     }
+
+    libc_bitflags! {
+        pub struct SchedFlags: c_int {
```
rtzoeller, line 220: This isn't a straightforward bitfield, and treating it as such is causing incorrect matches. Instead, the entries aside from `SCHED_RESET_ON_FORK` are a standard enum, but the enum can be ORed with `SCHED_RESET_ON_FORK`. Need to figure out how to represent this.

rtzoeller: @asomers I am revisiting this and would like some input on how to best expose this type in Rust. The lower bits of the type in libc are effectively an enum, and the upper bits are a bitfield.

I've considered the following options:
  1. Create 4-byte `SchedType` enum and `SchedFlags` bitfield types, and have the user pass them separately. nix can then OR them together internally, and hide this as an implementation detail.
  2. Same as 1., but create an 8-byte struct to hold these two fields.
  3. Create 2-byte `SchedType` enum and `SchedFlags` bitfield types, and have the user pass them separately. nix can then combine them via bitshifting to form a 4-byte value, and hide this as an implementation detail.
  4. Same as 3., but create a 4-byte struct to hold these two fields.
  5. Create a completely custom 4-byte type with accessors to properly manipulate the data.


--- 

1\. and 2. are wasteful in terms of space, but don't make assumptions about the number of bits assigned to the enum and flag portions of the type.

3\. and 4. place restrictions on the number of future bits in the enum and flag portions of the type, but aren't wasteful in terms of space.

5\. is the most work, and adds complexity to using the scheduler functions.


Thoughts?

src/sched.rs
```diff
@@ -215,6 +215,112 @@ mod sched_linux_like {
...
+
+    impl Default for SchedParam {
+        fn default() -> Self {
+            SchedParam::new(0)
+        }
+    }
+
+    /// Get minimum priority value for policy
+    ///
+    /// See also [`sched_get_priority_min(2)`](https://man7.org/linux/man-pages/man2/sched_get_priority_min.2.html)
```
asomers, line 260: This and the other functions in this PR are actually standardized by POSIX.  So you should like to the OpenGroup's man page, rather than the Linux manpage.  But if the POSIX man page doesn't make sense for some OS-specific options, then you can link to both.  Or if you're ambitious, you can selectively link to one or the other, like this:
```rust
#[cfg_attr(target_os = "linux", doc = "whatever")]
#[cfg_attr(not(target_os = "linux"), doc = "something else")]
```

test/test_sched.rs
```diff
@@ -1,4 +1,4 @@
-use nix::sched::{sched_getaffinity, sched_setaffinity, CpuSet};
+use nix::sched::{sched_getaffinity, sched_setaffinity, sched_getparam, sched_getscheduler, sched_setscheduler, CpuSet, SchedFlags, SchedParam};
```
asomers, line 1: Wrap to 80 cols, please.

src/sched.rs
asomers: 1 and 3 sound the most ergonomic.  I wouldn't worry about the size of the field, because Nix doesn't currently guarantee ABI stability across releases, and the field's size won't be expressed in the API at all.  But there is another important question: do any of the bitfield-like bits apply on non-Linux operating systems?  If not, then options 1-4 would seem awfully odd on non-Linux platforms.  5 would be the best in that case.

rtzoeller: The bitfield-like bits seem to only apply on Linux and Android (so... Linux). 

Does it make sense to wrap them in a struct (so 2 or 4), but have the second member of the struct conditionally defined only on Linux and Android? I.e. the struct would have a single enum member on non-Linux platforms, and an enum and bitfield on Linux + Android. Maybe that's just 5.

asomers: Yeah, I like that idea.  Even better if the struct's most common contructors don't involve the bitffield-like bits, making them portable.

rtzoeller: I was working on implementing these changes, but realized I wouldn't be satisfied with the API if it didn't support at least one platform without `SCHED_RESET_ON_FORK`. I picked FreeBSD (and will probably try to pick up other Tier 1 targets provided I can validate them locally), but have now run into the issue of libc not being up-to-date for some of these functions + platforms.

I'm working on some contributions to libc to expand support for `sched_*` functions, and will get back to this PR as those changes are submitted.

rtzoeller: @asomers I've put up the changes implementing this new approach; I don't love the type names, but I couldn't come up with anything better. Ditto for the namespace.

I kept this new commit separate for now, but will squash before this is actually submitted.

The dependency on libc was bumped to pick up the following changes:
 - https://github.com/rust-lang/libc/commit/b68bd77e3130115a5312d856ce881e0ec8533b0e
 - https://github.com/rust-lang/libc/commit/f148bfd673823b07bb7bdee7369cb0f88d368f5d
 - https://github.com/rust-lang/libc/commit/10eff9941f3002f31e18a2befe4bfcb86868140e
 - https://github.com/rust-lang/libc/commit/3c17b02bbec9a2db73cff14b225f3627bf07ac0a

asomers:
The test failures on FreeBSD and Illumos look like they're related to this PR.

rtzoeller:
> The test failures on FreeBSD and Illumos look like they're related to this PR.

The fix for Illumos will require additional [changes to libc](https://github.com/rust-lang/libc/pull/2350). I'm looking into the FreeBSD test failure as well.

src/sched.rs
rtzoeller: Decided to just refer to the POSIX documentation for now.

src/sched.rs
rtzoeller: Bumped libc again to pick up https://github.com/rust-lang/libc/commit/0904320c928ec98b3afea7874b29c7902490c836.

test/test_sched.rs
```diff
@@ -30,3 +38,82 @@ fn test_sched_affinity() {
     // Finally, reset the initial CPU set
     sched_setaffinity(Pid::from_raw(0), &initial_affinity).unwrap();
 }
+
+#[test]
+#[cfg(not(any(target_env = "musl", target_os = "freebsd")))]
+fn test_sched_scheduler_not_realtime() {
+    let initial_scheduler = sched_getscheduler(None).unwrap();
```
asomers, line 45: These two tests could interfere with each other.  To prevent that, you should create a SCHED_MTX in test/test.rs and grab it during these tests.

rtzoeller:
> The test failures on FreeBSD and Illumos look like they're related to this PR.

I've fixes these test failures, but am now faced with another batch of test failures.

I'm seeing an `EPERM` error on `test_sched_scheduler_realtime`, which should only execute when run as root. It behaves correctly on my local machine, but seems to fail on a number of Linux test environments despite apparently running as root.

My guess is that the tests are failing due to the container itself lacking the correct permissions, so even root within the container can't set an RT scheduler.

Any ideas on how to disable this test when running in the container? Should I just skip on `EPERM`, or is there a better way?

src/sched.rs
```diff
@@ -225,6 +225,221 @@ mod sched_linux_like {
...
+    }
+
+    #[repr(transparent)]
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+    pub struct SchedParam(libc::sched_param);
+
+    impl SchedParam {
+        pub fn new(priority: i32) -> Self {
+            let mut sched_param: libc::sched_param = unsafe { mem::zeroed() };
+            sched_param.sched_priority = priority;
```
asomers, line 344:
```suggestion
            let sched_param = unsafe { libc::sched_param {
                sched_priority: priority,
                .. mem::zeroed()
            }};
```
```

**Review by asomers — CHANGES_REQUESTED** [review]

File: test/test_sched.rs (hunk 9 above)
Comment (rtzoeller) [reply test/test_sched.rs:47]: Good catch; they can also interfere with the other two new tests if they change the scheduler + params between reads. I've added the mutex to all four.

File: src/sched.rs (hunk 10 above)
Comment (rtzoeller) [reply src/sched.rs]: Hmm, looks like Illumos broke because of this. Illumos defines `libc::sched_param` as 

```rust
    pub struct sched_param {
        pub sched_priority: ::c_int,
        sched_pad: [::c_int; 8]
    }
```

Should I revert this change?

asomers:
> > The test failures on FreeBSD and Illumos look like they're related to this PR.
> 
> I've fixes these test failures, but am now faced with another batch of test failures.
> 
> I'm seeing an `EPERM` error on `test_sched_scheduler_realtime`, which should only execute when run as root. It behaves correctly on my local machine, but seems to fail on a number of Linux test environments despite apparently running as root.
> 
> My guess is that the tests are failing due to the container itself lacking the correct permissions, so even root within the container can't set an RT scheduler.
> 
> Any ideas on how to disable this test when running in the container? Should I just skip on `EPERM`, or is there a better way? Alternatively, can we request this permission for the container?

You might need `require_capability!(CAP_SYS_NICE)` and/or `require_capability!(CAP_SYS_ADMIN)`.

src/sched.rs
asomers: Yeah.  Arguably, this is a bug in libc.  But I don't want to start that argument.  So instead you should do it like this: https://github.com/nix-rust/nix/blob/dab7332eabed8646f6d01a0d0688b4d1438accb4/src/sys/socket/mod.rs#L1388-L1401

rtzoeller: Think I have this working as expected now.

rtzoeller:
@asomers I've squashed the commits for this PR; the build is passing and I believe it's ready for another (hopefully final) review.

rtzoeller:
#1499 will break this, but it's a trivial fix to add the function name to the `require_capability` macro call.

test/test_sched.rs
```diff
@@ -1,7 +1,15 @@
+use nix::sched::{
+    sched_get_priority_min, sched_getparam, sched_getscheduler,
+    sched_setscheduler, SchedParam, SchedPolicy, SchedType,
+};
+#[cfg(not(any(target_os = "dragonfly", target_os = "freebsd")))]
 use nix::sched::{sched_getaffinity, sched_setaffinity, CpuSet};
 use nix::unistd::Pid;
 
+use crate::*;
```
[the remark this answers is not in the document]
rtzoeller: IIRC I added it to support the `require_capability!(CAP_SYS_NICE);` call in this same file. Other files using `require_capability` used the glob import, and in the moment it was easier than tracking the correct import.

I need to add documentation to this PR thanks to #1508 (which is a good thing!), so I can clean up this import at the same time.

rtzoeller:
Closing this PR for now; I need to rework it fix conflicts, and also want to get proper `SCHED_DEADLINE` support.

PR #1445, opened 2021-06-04
Add a wrapper around if_nameindex

coolreader18:


asomers (changes requested):
Looks pretty good!  Don't forget to add a CHANGELOG entry.

src/net/if_.rs
```diff
@@ -267,3 +267,147 @@ libc_bitflags!(
...
+                    self.ptr = self.ptr.add(1);
+                    Some(ret)
+                }
+            }
+        }
+    }
+
+    /// Retrieve a list of the network interfaces available on the local system.
+    ///
+    /// ```no_run
```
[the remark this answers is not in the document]
coolreader18: I thought it might error in CI or something, but in retrospect it's probably fine.

src/net/if_.rs
```diff
@@ -267,3 +267,147 @@ libc_bitflags!(
...
+    /// # fn main() -> nix::Result<()> {
```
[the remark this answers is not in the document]
coolreader18: Ah, yep, I thought so but didn't realize you still needed the `Ok(())`. Fixed.

src/net/if_.rs
```diff
@@ -267,3 +267,146 @@ libc_bitflags!(
...
+    }
+
+    /// Retrieve a list of the network interfaces available on the local system.
+    ///
+    /// ```
+    /// let interfaces = nix::net::if_::if_nameindex()?;
+    /// for iface in &interfaces {
+    ///     println!("Interface #{} is called {}", iface.index(), iface.name().to_string_lossy());
+    /// }
+    /// # nix::Result::Ok(())
```
asomers, line 392: I guess this works, but it would be in keeping with the style of the other doc tests if you return () here, and do an `.unwrap()` above instead of `?`.

asomers (changes requested):

src/net/if_.rs
coolreader18: Ah, alright, hadn't noticed that pattern.

coolreader18: Oh, though note that the `#` means that line will be hidden from the docs; readers won't see that it returns `Ok(())`

asomers (approved):
LGTM!

bors r+

PR #1447, opened 2021-06-08
Expose SockAddr::from_raw_sockaddr

coolreader18:
I also noticed the `SockAddr/InetAddr::to_str` functions were entirely redundant - `ToString` exists for that, & has a blanket impl on `T: Display`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1490](https://github.com/nix-rust/nix/pull/1490))
 - Added the `PTRACE_EVENT_STOP` variant to the `sys::ptrace::Event` enum
   (#[1335](https://github.com/nix-rust/nix/pull/1335))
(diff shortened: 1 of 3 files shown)


asomers (changes requested):
Could you please split these two changes up, since they're unrelated?  Also, it would be good to deprecate `to_str` for a few releases before removing it entirely.

coolreader18:
@asomers Could you take a look at this again? I split out the other change into #1495 as well.

asomers (approved):
bors r+

coolreader18:
Huh. I think the failure was just cause of the conflict in CHANGELOG?

asomers:
bors retry

coolreader18:
Ah, shoot, just realized the functionality I was looking for was in `sockaddr_storage_to_addr` - however, I'm pretty sure that should also be unsafe and take a `*const sockaddr` instead of an `&sockaddr_storage`, since it seems like there are some `assume_init` calls that might not be valid as the full extent of the sockaddr_storage might not actually get initialized if the address is only a couple bytes in size.

PR #1448, opened 2021-06-12
Add documentation for sched::clone

Mandragorian:
Add a note in the documentation for sched::clone to point out that the
stack pointer does not neet to be a reference to the highest address of
the stack.

Users who simply read the manpages for clone(2) might assume that they
will need to use unsafe pointer arithmetics in order to create a
reference to the highest address of their buffer, rather than providing
their buffer directly.

### DIFF
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -180,11 +180,10 @@ mod sched_linux_like {
     /// ([`clone(2)`](https://man7.org/linux/man-pages/man2/clone.2.html))
     ///
     /// `stack` is a reference to an array which will hold the stack of the new
-    /// process. Contrary to colling `clone(2)` from C, where the provided
-    /// stack address must be the highest address of the memory region, this is
-    /// not needed here. This requirement will be taken care of by `nix` and
-    /// the user only needs to provide a reference to a normally allocated
-    /// buffer.
+    /// process.  Unlike when calling `clone(2)` from C, the provided stack
+    /// address need not be the highest address of the region.  Nix will take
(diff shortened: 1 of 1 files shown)


src/sched.rs
```diff
@@ -176,6 +176,15 @@ mod sched_linux_like {
...
 
+    /// `clone` create a child process
+    /// ([`clone(2)`](https://man7.org/linux/man-pages/man2/clone.2.html))
+    ///
+    /// `stack` is a reference to an array which will hold the stack of the new
+    /// process. Contrary to colling `clone(2)` from C, where the provided
+    /// stack address must be the highest address of the memory region, this is
+    /// not needed here. This requirement will be taken care of by `nix` and
+    /// the user only needs to provide a reference to a normally allocated
+    /// buffer.
```
asomers, line 187:
```suggestion
    /// process.  Unlike when calling `clone(2)` from C, the provided stack
    /// address need not be the highest address of the region.  Nix will take
    /// care of that requirement.  The user only needs to provide a reference to
    /// a normally allocated buffer.
```

asomers (commented):
χαίρε, Κόστε!  Thanks for the tip.  I've just tweaked the wording a little bit to make it sound more natural.  Does it still look accurate to you?

src/sched.rs
Mandragorian: Thanks for the reply! Yes this reads much better.

asomers (approved):
bors r+

PR #1456, opened 2021-06-23
Add IpFreebind socket option

nox:


asomers (approved):
Looks simple enough.  I think we can merge as-is.

bors r+

PR #1462, opened 2021-07-09
Fix fork test and enable doc test

sporksmith:
Replaces `println!`  with raw `libc::write`. `println!` isn't guaranteed
to be async-signal-safe, and almost certainly *isn't* due to internal
buffering and locking.

Adds a call to `libc::_exit` in the child arm, so that it doesn't fall
through and start executing the parent code.

Adds a call to `waitpid` in the parent arm, to clean up the child
process.

Removes the `no_run` directive, so that it's run in the doc tests.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -200,14 +200,19 @@ impl ForkResult {
 /// be created that are identical with the exception of their pid and the
 /// return value of this function.  As an example:
 ///
-/// ```no_run
-/// use nix::unistd::{fork, ForkResult};
+/// ```
+/// use nix::{sys::wait::waitpid,unistd::{fork, ForkResult, write}};
 ///
 /// match unsafe{fork()} {
 ///    Ok(ForkResult::Parent { child, .. }) => {
 ///        println!("Continuing execution in parent process, new child has pid: {}", child);
+///        waitpid(child, None).unwrap();
 ///    }
-///    Ok(ForkResult::Child) => println!("I'm a new child process"),
+///    Ok(ForkResult::Child) => {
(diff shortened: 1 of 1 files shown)


src/unistd.rs
```diff
@@ -200,14 +200,19 @@ impl ForkResult {
...
+/// use nix::{sys::wait::waitpid,unistd::{fork, ForkResult, write}};
 ///
 /// match unsafe{fork()} {
 ///    Ok(ForkResult::Parent { child, .. }) => {
 ///        println!("Continuing execution in parent process, new child has pid: {}", child);
+///        waitpid(child, None).unwrap();
+///    }
+///    Ok(ForkResult::Child) => {
+///        // Unsafe to use `println!` (or `unwrap`) here. See Safety.
+///        write(libc::STDOUT_FILENO, "I'm a new child process\n".as_bytes()).ok();
```
asomers, line 213: Should you explicitly nul-terminate this string?

sporksmith: No, I don't think so. The [underlying syscall](https://docs.rs/libc-nnsdk/0.0.1/libc/fn.write.html) takes an explicit length parameter (which the nix wrapper [provides for us](https://docs.rs/nix/0.7.0/src/nix/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.7.0/src/unistd.rs.html#311-315))

asomers: Yeah, that's right.  Because you're writing to stdout, there's no need to nul-terminate.  stdout never ends.

asomers (approved):
bors r+

PR #1465, opened 2021-07-13
Enable creating a const TimeSpec

danieldulaney:
Previously, there was no way to create a `TimeSpec` in a `const` context because all creation was through traits. This adds two utility functions to create a `const TimeSpec` from a `libc::timespec` or an `std::time::Duration`.

An alternative approach would be to make the inner `timespec` field `pub`, which would not require any additional functions but would expose some (potentially unwanted) implementation details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
I think this looks good.  Would you mind adding a CHANGELOG entry?

danieldulaney:
I think this is the correct way to add an entry -- is this the first change for this version?

asomers (approved):
Yep, that's correct.

asomers (approved):
bors r+

Issue #1466, opened 2021-07-13
0.22.0 includes breaking changes


sporksmith:
Mostly FYI; fwiw I'm happy with the API change and it wasn't a big deal to fix in our code, but if I understand semver correctly this should be a major version number bump? (I know nobody wants to go from 0 to 1 ;))

Here's our broken build after bumping from 0.21.0 to 0.22.0: https://github.com/shadow/shadow/pull/1508/checks?check_run_id=3057359519

And our fixes: https://github.com/shadow/shadow/pull/1512

```
error[E0425]: cannot find value `EWOULDBLOCK` in module `nix::errno`
  --> host/descriptor/pipe.rs:82:29
   |
82 |             Err(nix::errno::EWOULDBLOCK.into())
   |                             ^^^^^^^^^^^ not found in `nix::errno`
   |
help: consider importing this constant
   |
1  | use libc::EWOULDBLOCK;
   |

error[E0425]: cannot find value `EWOULDBLOCK` in module `nix::errno`
   --> host/descriptor/pipe.rs:115:29
    |
115 |             Err(nix::errno::EWOULDBLOCK.into())
    |                             ^^^^^^^^^^^ not found in `nix::errno`
    |
help: consider importing this constant
    |
1   | use libc::EWOULDBLOCK;
    |

error[E0425]: cannot find value `EWOULDBLOCK` in module `nix::errno`
   --> host/syscall/unistd.rs:150:34
    |
150 |     if result == Err(nix::errno::EWOULDBLOCK.into()) && !file_flags.contains(FileFlags::NONBLOCK) {
    |                                  ^^^^^^^^^^^ not found in `nix::errno`
    |
help: consider importing this constant
    |
1   | use libc::EWOULDBLOCK;
    |

error[E0425]: cannot find value `EWOULDBLOCK` in module `nix::errno`
   --> host/syscall/unistd.rs:219:34
    |
219 |     if result == Err(nix::errno::EWOULDBLOCK.into()) && !file_flags.contains(FileFlags::NONBLOCK) {
    |                                  ^^^^^^^^^^^ not found in `nix::errno`
    |
help: consider importing this constant
    |
1   | use libc::EWOULDBLOCK;
    |

error[E0425]: cannot find value `ENOTSUP` in module `nix::errno`
   --> host/syscall_types.rs:324:39
    |
324 |             let default = nix::errno::ENOTSUP;
    |                                       ^^^^^^^ not found in `nix::errno`
    |
[...]

PR #1467, opened 2021-07-14
Add more errno definitions for better backwards compat with 0.21.0

asomers:
Fixes #1464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1464](https://github.com/nix-rust/nix/pull/1464))
 
 ### Fixed
+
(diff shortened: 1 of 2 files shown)


asomers:
@sporksmith would the changes in this PR make the 0.21.0 -> 0.22.X upgrade easier for projects like yours?

sporksmith:
Yes, it looks like this would've turned some of the errors into warnings. In our particular case there were few enough that I just fixed all the errors and warnings at once, but I imagine this could be helpful for other projects.

sporksmith:
> Fixes #1464

Did you mean #1466? Btw there was another kind of error that looks like wasn't included in the truncated error list in the compiler output I shared there; will add on to #1466 since I think it'd be orthogonal to this PR.

PR #1468, opened 2021-07-14
Add raw timerfd wrappers

danieldulaney:
Add raw wrappers for the timerfd functions (`timerfd_create`, `timerfd_settime`, and `timerfd_gettime`) and refactor `TimerFd` to use those wrappers safely. Additionally, add extra wrappers and conversion functions to convert between `Expiration` and `TimerSpec` (pulling out some of the logic explicitly).

It can be useful to handle timerfds in a "raw" way, and sometimes the `TimerFd` abstraction gets in the way. I encountered this while passing raw FDs around for epoll operations, and this interface is inspired by the `epoll` module: 1-to-1 wrappers over the libc function calls.

The `TimerFd` API remains identical, but it is implemented as a thin shim over the new functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Added raw `timerfd` wrappers (#[1468](https://github.com/nix-rust/nix/pull/1468))
+
 ### Changed
 ### Fixed
 ### Removed
--- a/src/sys/timerfd.rs
+++ b/src/sys/timerfd.rs
@@ -99,11 +99,24 @@ impl TimerSpec {
             },
         })
     }
+
+    pub fn is_none(&self) -> bool {
+        self.0.it_interval.tv_sec == 0
+            && self.0.it_interval.tv_nsec == 0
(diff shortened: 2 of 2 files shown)


asomers (commented):
These new functions don't really do anything except call the libc FFI functions.  All they do is check errno.  Do you really need Nix for that?  Why not just use libc directly?

danieldulaney:
I was trying to leave `unsafe` out of my code as much as possible, and having Rust-friendly errno checking plus a nice enum-based interface for setting times is nicer than dropping down to the C interface. That said, it's not *that* many LOC and I could just have it as a utility module in my crate.

I felt the need for these because I ran into places where I needed to pass the FD through epoll, but without actually dropping (and therefore closing) the `TimerFd`. I ended up having several places where I was using `ManuallyDrop`, as well as several more where I needed the unsafe `FromRawFd` call to re-build the `TimerFd` abstraction. Having the low-level calls directly (and safely) available was useful.

That said, it's definitely not required, and I understand if it's not a good fit for nix.

asomers:
Yeah, I don't think it's a good fit for Nix's API, so I'll close it.  As for epoll, note that Nix's `TimerFd` already implements `AsRawFd`, which might be good enough for you.

PR #1469, opened 2021-07-15
Fix CONTRIBUTING.md formating

rusty-snake:
Guess the label prefixes should be rendered bold and not with literal stars.


asomers (approved):
bors r+

PR #1470, opened 2021-07-16
Add the IPV6_V6ONLY sockopt

danieldulaney:
`IPV6_V6ONLY` allows the user to select between dual-stack and IPv6-only sockets.

Platform support seems to be good; all platforms with IPv6 support seem to have this sockopt.


asomers (approved):
bors r+

PR #1471, opened 2021-07-18
unistd: Add conversion from User to libc::passwd

orvij:
Add `From<User> for libc::passwd` trait implementation to convert a `User` into a `libc::passwd`

Implementation consumes the `User` struct, giving ownership over the internal members to the `libc::passwd` struct

Exposes the `User::gecos` field to 64-bit Android builds, since `libc::passwd::pw_gecos` is supported on those builds

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -2629,7 +2629,7 @@ pub struct User {
     /// Group ID
     pub gid: Gid,
     /// User information
-    #[cfg(not(target_os = "android"))]
+    #[cfg(not(all(target_os = "android", target_pointer_width = "32")))]
     pub gecos: CString,
     /// Home directory
     pub dir: PathBuf,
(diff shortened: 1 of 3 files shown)


src/unistd.rs
```diff
@@ -2629,7 +2629,7 @@ pub struct User {
     /// Group ID
     pub gid: Gid,
     /// User information
-    #[cfg(not(target_os = "android"))]
+    #[cfg(not(all(target_os = "android", target_pointer_width = "32")))]
```
asomers, line 2632: This looks like an extra change that was not mentioned in the commit message.  Could you please mention it?

asomers (changes requested):
Just add a CHANGELOG entry, and don't forget to mention the change to the gecos field.

asomers (changes requested):
Some kind of weird merge error screwed up the changelog.  There are three "Added" sections now.

orvij:
> Some kind of weird merge error screwed up the changelog. There are three "Added" sections now.

Sorry about that, when I rebased some things went weird. Fixed

asomers (approved):
bors r+

PR #1473, opened 2021-07-22
sys/stat: add a safe wrapper for mknodat(2)

lucab:
This introduces a new `mknodat` helper.

Ref: https://pubs.opengroup.org/onlinepubs/9699919799/functions/mknod.html


otavio (approved):

asomers (changes requested):
Looks good.  How about a test?

lucab:
I was hoping to copy+adjust the `mknod` test, but there was none; I thus added a basic test for both.
I initially started writing a chardev test, but then realized it could be problematic on `nodev` mountpoints so I went for a simpler regular-file one instead (which is not supported on freebsd though).
Also rebased on current master to solve changelog conflict. CI is green. This is ready for another review pass now.

asomers (approved):
LGTM.

bors r+

PR #1474, opened 2021-07-24
Mark most C-derived enums as non_exhaustive

### LINKED ISSUE #1182: Use #[non_exhaustive] after updating MSRV to 1.40.0
The `#[non_exhaustive]` attribute, introduced by Rust 1.40.0, should be used for most enums that come from C, since new values could be added by any release.  This attribute isn't valuable enough to justify a MSRV bump by itself.  But we should still make use of it whenever we update the MSRV for any other reason.

asomers:
Since libc may add new variants at any time, Nix's consumers should not
use exhaustive match patterns.

Fixes #1182


asomers (commented):
bors r+

PR #1475, opened 2021-07-24
Improve the efficiency of cmsg_space!

asomers:
Compute the required space at compile time rather than at runtime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ targets = [
 ]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


asomers:
So apparently stable compilers don't support this yet.  https://github.com/rust-lang/rust/issues/64926 .

asomers:
Rust and libc have advanced to the point where this PR could be completed.  However, as that isn't possible.  As explained in #2282 , there are cases where we can't compute cmsg_space at compile time, because of subtleties involving generics.  Closing this PR.

PR #1483, opened 2021-08-08
Remove some actually unsupported termios iflags on redox

coolreader18:
Related to rust-lang/libc#2327

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc", rev = "9c1489fa8", features = [ "extra_traits" ] }
+libc = { git = "https://github.com/rust-lang/libc", rev = "f5e31f208", features = [ "extra_traits" ] }
 bitflags = "1.1"
 cfg-if = "1.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc", rev = "9c1489fa8", features = [ "extra_traits" ] }
+libc = { git = "https://github.com/rust-lang/libc", features = [ "extra_traits" ] }
```
asomers, line 28: It's good to always specific a `rev` when using git dependencies with Cargo.  Otherwise, you force Cargo to fetch the repository each and every time.

asomers (changes requested):

Cargo.toml
coolreader18: Alright, fixed!

asomers (changes requested):
Ok.  Two more things. First, would you please squash your commits?  Second, this probably deserves a mention in the Fixed section of the  CHANGELOG.

asomers (approved):
bors r+

PR #1486, opened 2021-08-10
Relax assertions in sockaddr_storage_to_addr to match the documentation.

### LINKED ISSUE #1479: sockaddr_storage_to_addr is documented as being more lenient with the size than it is in practice
The documentation says

```
The len argument should be the number of bytes in the sockaddr_storage that are actually allocated and valid. It must be at least as large as all the useful parts of the structure.
```

This suggests that passing in a size larger than what is actually used (say, `size_of::<sockaddr_storage>()`) should be fine. But in the implementation, both `AF_INET` and `AF_INET6` assert that the size passed is exactly `size_of::<sockaddr_in>()`/`size_of::<sockaddr_in6>()`. Ideally IMO the implementation would match the documented behavior.

khuey:
Fixes #1479

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -1,12 +1,13 @@
-use nix::sys::socket::{AddressFamily, InetAddr, UnixAddr, getsockname};
+use nix::sys::socket::{AddressFamily, InetAddr, SockAddr, UnixAddr, getsockname, sockaddr, sockaddr_in6, sockaddr_storage_to_addr};
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
+use std::mem::{self, MaybeUninit};
 use std::net::{self, Ipv6Addr, SocketAddr, SocketAddrV6};
 use std::os::unix::io::RawFd;
 use std::path::Path;
 use std::slice;
 use std::str::FromStr;
-use libc::c_char;
+use libc::{c_char, sockaddr_storage};
 #[cfg(any(target_os = "linux", target_os= "android"))]
 use crate::*;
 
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
The tests you modified are now doing two completely different things.  Better to split them up, testing only the minimal amount of functionality in each test.

test/sys/test_socket.rs
```diff
@@ -54,6 +78,35 @@ pub fn test_inetv6_addr_to_sock_addr() {
...
+    let actual = SocketAddr::V6(SocketAddrV6::new(ip, port, flowinfo, scope_id));
+    let addr = InetAddr::from_std(&actual);
+    let sockaddr = SockAddr::new_inet(addr);
+
+    let (storage, ffi_size) = {
+        let mut storage = MaybeUninit::<sockaddr_storage>::zeroed();
+        let storage_ptr = storage.as_mut_ptr().cast::<sockaddr_in6>();
+        let (ffi_ptr, ffi_size) = sockaddr.as_ffi_pair();
+        // XXX this does not pass.
+        // assert_eq!(mem::size_of::<sockaddr>(), ffi_size as usize);
```
asomers, line 97: You should delete this comment

asomers (changes requested):

test/sys/test_socket.rs
khuey: Done.

asomers:
You'll need to rebase to fix the test failures.

khuey:
Done.

asomers:
Just one more thing: would you mind squashing your commits?

khuey:
Done.

asomers (approved):
bors r+

asomers:
bors retry

khuey:
Thanks.

PR #1487, opened 2021-08-10
Update to libc 0.2.99

coolreader18:


asomers:
Why?  Does some other PR require a newer libc, or are you doing this in preparation for a release?

coolreader18:
Oh, I suppose in prep for release, I noticed there was a `staging` branch pointing to HEAD. Also makes it easier to use as a (git, for now) dependency since you don't have to add a separate `[patch.'https://github.com/rust-lang/libc']` in Cargo.toml

PR #1489, opened 2021-08-11
Don't use mem::transmute in SignalFd

asomers:
There was a better case for using it before mem::uninitialized was
    available, but not great.  Even before then, mem::zeroed could've been
    used instead.
    
    Issue #373


asomers:
bors retry

PR #1492, opened 2021-08-13
Fix the build with bitflags-1.3.0 and newer

### LINKED ISSUE #1491: nix doesn't compile with bitflags-1.3.1 on rustc-1.45
`bitflags` version 1.3.1 was recently released. `nix` specifies a dependency on `bitflags` with version constraint `"1.1"`. Without a previously working `Cargo.lock` file `cargo` is now free to choose `bitflags` version 1.3.1 to satisfy `nix`'s dependency. However, with that `bitflags` version, `nix` doesn't compile on `rustc` version `1.45`. Here is one of the many errors I get:

```
error[E0658]: `if` is not allowed in a `const fn`
  --> src/fcntl.rs:28:1
   |
28 | / libc_bitflags! {
29 | |     pub struct AtFlags: c_int {
30 | |         AT_REMOVEDIR;
31 | |         AT_SYMLINK_FOLLOW;
...  |
39 | |     }
40 | | }
   | |_^
   |
   = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

Since `nix` claims to build on `rustc-1.41` and upwards, I think this is a bug/misconfiguration.

asomers:
Bitflags raised its MSRV in a minor version, forcing all consumers to follow suit.
    
Fixes #1491

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,7 +15,7 @@ task:
     image: freebsd-11-4-release-amd64
   setup_script:
     - fetch https://sh.rustup.rs -o rustup.sh
-    - sh rustup.sh -y --profile=minimal --default-toolchain 1.41.0
+    - sh rustup.sh -y --profile=minimal --default-toolchain 1.46.0
     - $HOME/.cargo/bin/rustup target add i686-unknown-freebsd
   amd64_test_script:
     - . $HOME/.cargo/env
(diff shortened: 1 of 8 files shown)


lopopolo:
> Bitflags raised its MSRV in a minor version, forcing all consumers to follow suit.

Not to be combative, but an alternative way to say this is that `nix` is holding the ecosystem back from upgrading.

Leaving the existing constraint in place and generating a lockfile with minimal versions would allow applications building with older `rustc` to continue to compile even in the absence of an already-generated lockfile and allow applications building with newer `rustc` to use the latest `bitflags`.

```console
cargo +nightly -Z minimal-versions generate-lockfile
```

lopopolo:
Oh wow, I did not look at the code in this PR at all, sorry for that!

It looks like this PR _does_ raise the MSRV for `nix`. I didn't pick that up from either this PR description or the linked ticket.

Sorry. I'm embarrassed.

asomers:
Yes, that's right.  Nix raised its MSRV on the master branch, but locked bitflags to < 1.3.0 in the point releases.

PR #1496, opened 2021-08-22
Rework UnixAddr to fix soundness issues

### LINKED ISSUE #1494: Unsoundness in UnixPath
```rust
println!("{}", UnixAddr(sockaddr_un { sun_family: AF_UNIX as _, sun_path: [0; 108] }, 10000));
```
Output;
```
"@<\u{0} x 108><9892 bytes of garbage>"
```
Also in the PartialEq and Hash impls and `as_abstract`; anywhere that uses `sun_path()` raw without doing what `path()` does and calling `strnlen` (and even `path()` is stiill unsound, it still calls `sun_path()` and creates a slice with a length longer than it should have)

coolreader18:
Fixes #1494

I went with making `sun_path` always nul-terminated since that just seems to make things easier, since (at least according to linux man pages) `sockaddr_un`s returned by the kernel will always be nul-terminated.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -515,15 +515,42 @@ impl fmt::Display for Ipv6Addr {
 }
 
 /// A wrapper around `sockaddr_un`.
-///
-/// This also tracks the length of `sun_path` address (excluding
-/// a terminating null), because it may not be null-terminated.  For example,
-/// unconnected and Linux abstract sockets are never null-terminated, and POSIX
-/// does not require that `sun_len` include the terminating null even for normal
-/// sockets.  Note that the actual sockaddr length is greater by
-/// `offset_of!(libc::sockaddr_un, sun_path)`
 #[derive(Clone, Copy, Debug)]
-pub struct UnixAddr(pub libc::sockaddr_un, pub usize);
+pub struct UnixAddr {
+    // INVARIANT: sun & path_len are valid as defined by docs for from_raw_parts
+    sun: libc::sockaddr_un,
+    path_len: usize,
+}
+
+// linux man page unix(7) says there are 3 kinds of unix socket:
+// pathname: addrlen = offsetof(struct sockaddr_un, sun_path) + strlen(sun_path) + 1
+// unnamed: addrlen = sizeof(sa_family_t)
+// abstract: addren > sizeof(sa_family_t), name = sun_path[..(addrlen - sizeof(sa_family_t))]
+//
+// what we call path_len = addrlen - offsetof(struct sockaddr_un, sun_path)
+#[derive(PartialEq, Eq, Hash)]
+enum UnixAddrKind<'a> {
+    Pathname(&'a Path),
+    Unnamed,
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    Abstract(&'a [u8]),
+}
+impl<'a> UnixAddrKind<'a> {
+    /// Safety: sun & path_len must be valid
+    unsafe fn get(sun: &'a libc::sockaddr_un, path_len: usize) -> Self {
+        if path_len == 0 {
+            return Self::Unnamed;
+        }
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        if sun.sun_path[0] == 0 {
+            let name =
+                slice::from_raw_parts(sun.sun_path.as_ptr().add(1) as *const u8, path_len - 1);
+            return Self::Abstract(name);
+        }
+        let pathname = slice::from_raw_parts(sun.sun_path.as_ptr() as *const u8, path_len - 1);
+        Self::Pathname(Path::new(OsStr::from_bytes(pathname)))
+    }
+}
 
 impl UnixAddr {
     /// Create a new sockaddr_un representing a filesystem path.
@@ -537,7 +564,7 @@ impl UnixAddr {
 
                 let bytes = cstr.to_bytes();
 
-                if bytes.len() > ret.sun_path.len() {
+                if bytes.len() >= ret.sun_path.len() {
                     return Err(Error::from(Errno::ENAMETOOLONG));
                 }
 
@@ -545,7 +572,7 @@ impl UnixAddr {
                                          ret.sun_path.as_mut_ptr() as *mut u8,
                                          bytes.len());
 
-                Ok(UnixAddr(ret, bytes.len()))
+                Ok(UnixAddr::from_raw_parts(ret, bytes.len() + 1))
             }
         })?
     }
@@ -564,7 +591,7 @@ impl UnixAddr {
                 .. mem::zeroed()
             };
 
-            if path.len() + 1 > ret.sun_path.len() {
+            if path.len() >= ret.sun_path.len() {
                 return Err(Error::from(Errno::ENAMETOOLONG));
             }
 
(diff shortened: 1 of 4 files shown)


src/sys/socket/addr.rs
```diff
@@ -515,15 +515,44 @@ impl fmt::Display for Ipv6Addr {
...
+#[derive(PartialEq, Eq, Hash)]
+enum UnixAddrKind<'a> {
+    Pathname(&'a Path),
+    Unnamed,
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    Abstract(&'a [u8]),
+}
+impl<'a> UnixAddrKind<'a> {
+    /// Safety: sockaddr+size must be valid
+    #[allow(unused_unsafe)] // #[deny(unsafe_op_in_unsafe_fn)]
```
asomers, line 540: This `allow(unused_unsafe)` is a red flag.  Why did you make the function unsafe?  Either the function itself should be unsafe, or there should be unsafe blocks around `slice::from_raw_parts`, but not both.

src/sys/socket/addr.rs
```diff
@@ -574,28 +603,31 @@ impl UnixAddr {
...
-    fn sun_path(&self) -> &[u8] {
-        unsafe { slice::from_raw_parts(self.0.sun_path.as_ptr() as *const u8, self.1) }
+    /// Create a UnixAddr from a raw `sockaddr_un` struct and a size.
+    ///
+    /// # Safety
+    /// This pair of sockaddr_un+size must be a valid unix addr, which means:
+    /// - sun_len <= sockaddr_un.sun_path.len()
+    /// - if this is a unix addr with a fs path, sun.sun_path is a nul-terminated fs path and
+    ///   sun.sun_path[sun_len - 1] == 0
+    pub unsafe fn from_raw_parts(sun: libc::sockaddr_un, sun_len: usize) -> UnixAddr {
```
asomers, line 617: Why `pub`?

src/sys/socket/addr.rs
```diff
@@ -515,15 +515,44 @@ impl fmt::Display for Ipv6Addr {
...
-/// unconnected and Linux abstract sockets are never null-terminated, and POSIX
-/// does not require that `sun_len` include the terminating null even for normal
-/// sockets.  Note that the actual sockaddr length is greater by
-/// `offset_of!(libc::sockaddr_un, sun_path)`
 #[derive(Clone, Copy, Debug)]
-pub struct UnixAddr(pub libc::sockaddr_un, pub usize);
+pub struct UnixAddr {
+    // INVARIANT: sun+sun_len is valid as defined by docs for from_raw_parts
+    sun: libc::sockaddr_un,
+    sun_len: usize
```
asomers, line 522: This is a pretty confusing name, since it's not the same as `sun.sun_len`.

asomers (changes requested):

src/sys/socket/addr.rs
coolreader18: I was sort of following [RFC 2585](https://github.com/rust-lang/rfcs/blob/master/text/2585-unsafe-block-in-unsafe-fn.md) (more accidentally than anything), but it would probably just be easier if I just get rid of the unsafe blocks.

src/sys/socket/addr.rs
coolreader18: I didn't want to remove the functionality that was there previously, creating a `UnixAddr` from a raw `sockaddr_un + usize`, but it should be unsafe now.

src/sys/socket/addr.rs
coolreader18: Ah, I didn't realize that there was a `sun.sun_len` in BSD, I see that now. What would you recommend? `addrlen` would be the name for the length including the `sa_family`, so maybe `path_len` or `buflen`?

src/sys/socket/addr.rs
asomers: Ahh, I wasn't aware of that RFC.  Is it implemented yet?

coolreader18: Yes, and I did use the uncommented `unsafe_op_in_unsafe_fn` but it was stabilized fairly recently so it didn't pass CI.

src/sys/socket/addr.rs
```diff
@@ -574,28 +601,33 @@ impl UnixAddr {
...
+    /// Create a UnixAddr from a raw `sockaddr_un` struct and a size. `path_len` is the "addrlen"
+    /// of this address, but minus `offsetof(struct sockaddr_un, sun_path)`. Basically the length
+    /// of the data in `sun_path`.
+    ///
+    /// # Safety
+    /// This pair of sockaddr_un & path_len must be a valid unix addr, which means:
+    /// - path_len <= sockaddr_un.sun_path.len()
+    /// - if this is a unix addr with a pathname, sun.sun_path is a nul-terminated fs path and
+    ///   sun.sun_path[path_len - 1] == 0
+    pub unsafe fn from_raw_parts(sun: libc::sockaddr_un, path_len: usize) -> UnixAddr {
```
asomers, line 617: Even though previous releases of Nix technically made this possible, I doubt that there's a good reason for it.  If it were backwards-compatible, we should deprecate it.  But since it's not, we should remove it instead.  Please remove the `pub`, and note this in the CHANGELOG.

coolreader18:
I also just changed the documentation of `sockaddr_storage_to_addr`.

coolreader18:
@asomers could you take a look?

src/sys/socket/addr.rs
asomers: There probably wasn't a good reason for anybody to manually construct a `UnixAddr` before.  Let's mark this new constructor as `deprecated`.

src/sys/socket/addr.rs
asomers: What exactly will be the difference between this field and `sun_len` ( on platforms where it exists)?  pathlen would probably be fine.

asomers (changes requested):
Changing the meaning of `len` in sockaddr_storage_to_addr is a big deal.  Can you make that backwards compatible instead?

src/sys/socket/addr.rs
```diff
@@ -605,39 +637,53 @@ impl UnixAddr {
...
-            path.display().fmt(f)
-        } else {
-            let display = String::from_utf8_lossy(&self.sun_path()[1..]);
-            write!(f, "@{}", display)
+        match self.kind() {
+            UnixAddrKind::Pathname(path) => path.display().fmt(f),
+            UnixAddrKind::Unnamed => f.pad("<unbound UNIX socket>"),
+            #[cfg(any(target_os = "android", target_os = "linux"))]
+            UnixAddrKind::Abstract(name) => {
+                write!(f, "@{}", String::from_utf8_lossy(name))
```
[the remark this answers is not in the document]
coolreader18: That's what it did already for `Display`. I can change it though.

src/sys/socket/mod.rs
```diff
@@ -1774,7 +1774,7 @@ pub fn getsockname(fd: RawFd) -> Result<SockAddr> {
 /// In C this would usually be done by casting.  The `len` argument
 /// should be the number of bytes in the `sockaddr_storage` that are actually
 /// allocated and valid.  It must be at least as large as all the useful parts
-/// of the structure.  Note that in the case of a `sockaddr_un`, `len` need not
+/// of the structure.  Note that in the case of a `sockaddr_un`, `len` should
```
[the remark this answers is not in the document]
coolreader18: Sure, the only thing with that is it would basically just ignore the `len` argument and do `strlen(sun_path.as_ptr(), sun_path.len())`. Unless it should *just* check?:
```rust
if sun_path.get(len) == Some(b'\0') {
    len += 1
} else {
    assert!(sun_path[len - 1] == b'0'); // maybe?
}
```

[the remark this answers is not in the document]
coolreader18: Alright, I think I fixed it.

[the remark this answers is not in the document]
asomers: What exactly did you change?  It's hard to tell since you force-pushed and github forgets about your older commits.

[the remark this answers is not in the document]
coolreader18: In `UnixAddr::from_raw_parts`, it increments len by 1 if `path[len - 1] != 0`

asomers (approved):
bors r+

PR #1498, opened 2021-08-22
Conditionally enable functionality with feature flags

asomers:
This PR satisfies a long-running request: to slim down the Nix crate by conditionally enabling certain functionality.  By default everything is enabled, but consumers can reduce that by selecting a subset of features.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,6 +20,8 @@ build: &BUILD
     - $TOOL +$TOOLCHAIN $BUILD $ZFLAGS --target $TARGET --all-targets
     - $TOOL +$TOOLCHAIN doc $ZFLAGS --no-deps --target $TARGET
     - $TOOL +$TOOLCHAIN clippy $ZFLAGS --target $TARGET -- $CLIPPYFLAGS
+    - if [ -z "$NOHACK" ]; then $TOOL install cargo-hack; fi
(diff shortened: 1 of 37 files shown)


asomers:
I think I addressed all of your comments.  HOWEVER, rustdoc's behavior just changed, which will require revisiting every part of this PR.  I'm working on that now.
https://github.com/rust-lang/rust/pull/89596

rtzoeller:
One other note: if you've considered updating the codebase to use `rustfmt`/`cargo fmt`, doing it immediately following the submission of this PR might be a good idea. This change will likely cause most existing PRs to require resolving conflicts anyways.

rtzoeller:
@asomers I think we should consider whether or not we expect tests with just some features enabled to work.

E.g. should `cargo test --no-default-features --features fs` build successfully, and run just the `fs`-related tests?

I'm also fine with saying "ideally yes, but not in this PR".

asomers:
> @asomers I think we should consider whether or not we expect tests with just some features enabled to work.
> 
> E.g. should `cargo test --no-default-features --features fs` build successfully, and run just the `fs`-related tests?
> 
> I'm also fine with saying "ideally yes, but not in this PR".

It would be nice.  OTOH it would clutter up the test files with a lot of `#[cfg()]` directives, which is why I didn't do it originally.  I don't think it's necessary, because I can't see any way that disabling features could cause the tests to fail, as long as they compile.  Can you?

PR #1505, opened 2021-08-26
Fix build for DragonFlyBSD

rtzoeller:
The build for DragonFlyBSD is currently failing, due to a missing field when initializing `passwd`.

Fix warnings due to types being referenced which are marked as deprecated in libc. These constants are unused by DragonFlyBSD and can be removed: https://github.com/rust-lang/libc/commit/e0d27d01051f80e1a54df2d7e73a3e2b0e1a594d

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -65,8 +65,8 @@ libc_bitflags!{
         MAP_LOCKED;
         /// Do not reserve swap space for this mapping.
         ///
-        /// This was removed in FreeBSD 11.
-        #[cfg(not(target_os = "freebsd"))]
+        /// This was removed in FreeBSD 11 and is unused in DragonFlyBSD.
+        #[cfg(not(any(target_os = "dragonfly", target_os = "freebsd")))]
         MAP_NORESERVE;
         /// Populate page tables for a mapping.
         #[cfg(any(target_os = "android", target_os = "linux"))]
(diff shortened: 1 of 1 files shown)


rtzoeller:
@asomers no good ideas about running tests for DragonFlyBSD in CI, but I think we could cross-compile for it from a Linux host with 

```
cargo build -Zbuild-std --target x86_64-unknown-dragonfly
```

I just recently submitted a [fix](https://github.com/rust-lang/rust/commit/0d1d9788e5a98268472f6c8538f4c0d18c4986c6) to the rustc to fix the nightly compiler, and the above command now works with the latest nightly builds.

asomers:
Does it not require that the x86_64-unknown-dragonfly target be available from rustup?  And does it not also take a _very_ long time?

rtzoeller:
> Does it not require that the x86_64-unknown-dragonfly target be available from rustup?

It does not appear so.

```
$ rustup target list --installed
x86_64-unknown-linux-gnu
```

I did have to run the following (which cargo instructed me to do after the first build attempt failed):

```
rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
```

---

> And does it not also take a _very_ long time?

It takes longer, but IMHO not prohibitively (at least on my laptop).

```
$ # Reinstalling the toolchain to try to minimize any caching
$ rustup toolchain remove nightly
$ rustup toolchain add nightly
$ rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
$ cargo clean
$ time cargo build -Zbuild-std --target x86_64-unknown-dragonfly
real	0m21.213s
user	0m43.172s
sys 	0m3.359s
$ cargo clean
$ time cargo build --target x86_64-unknown-linux-gnu
real	0m6.588s
user	0m8.877s
sys 	0m0.884s
```

Building for `x86_64-unknown-dragonfly` compiles 42 crates, while building for `x86_64-unknown-linux-gnu` builds 10.

PR #1509, opened 2021-08-31
Allow Android to use timerfd

rtzoeller:
This is a continuation of #1336 which also enables the timerfd tests.

```
running 3 tests
test sys::test_timerfd::test_timerfd_unset ... ok
test sys::test_timerfd::test_timerfd_oneshot ... ok
test sys::test_timerfd::test_timerfd_interval ... ok
```

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -135,6 +135,6 @@ pub mod wait;
 #[allow(missing_docs)]
 pub mod inotify;
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "android", target_os = "linux"))]
 #[allow(missing_docs)]
 pub mod timerfd;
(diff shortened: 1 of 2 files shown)


rtzoeller:
@asomers I'm looking into the current status of the rest of the tests on Android (as requested on the original PR), but I likely won't be able to complete that investigation for a week or two due to limited availibility.

I'm running the tests in Termux (without root), and it is subject to a number of limitations by the OS which a more privileged process might not be. For example, I can't `ls /`! I'm trying to narrow down if the tests are failing because the core functionality is broken, or because the test needs to be updated to account for the lack of permissions, but this could take some time.

asomers:
It looks good, @rtzoeller .  Do you actually use Nix on Android?

asomers (approved):
bors r+

PR #1511, opened 2021-09-04
Document more things

asomers:
Also, test rustdoc in CI, and demote missing_docs from a deny to a
warning (but still deny it in CI).


barracuda156:
macOS may not have `preadv`/`pwritev`, which breaks linking in that case.

PR #1514, opened 2021-09-06
Add support for IP_RECVERR

cemeyer:
This pull request adds support for setting the `IP_RECVERR` (and corresponding `IPV6_RECVERR`) sockopt, and then decoding the resulting control messages from `recvmsg()`.  It is a Linux-specific API.

This PR extends #1161 by:
* Making the address associated with a `IpvXRecvErr` control message optional (`recvmsg` documentation claims it is not provided for some kinds of error), per @WGH-.
* Adding basic tests for both IPv4 and IPv6 (blat a UDP packet at a hopefully-unoccupied port on localhost and observe ICMP port unreachable).
* Adding a Changelog entry.

I added some trivial doc comments on the `ControlMessageOwned` variants, but I'm not sure exactly what documentation you had in mind earlier on #1161.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1503](https://github.com/nix-rust/nix/pull/1503))
 - Enabled `pwritev` and `preadv` for more operating systems.
   (#[1511](https://github.com/nix-rust/nix/pull/1511))
+- Added `Ipv4RecvErr` and `Ipv6RecvErr` sockopts and associated control messages.
+  (#[1514](https://github.com/nix-rust/nix/pull/1514))
 
 ### Changed
 
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -653,6 +653,13 @@ pub enum ControlMessageOwned {
     #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
     RxqOvfl(u32),
 
+    /// Socket error queue control messages read with the `MSG_ERRQUEUE` flag.
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    Ipv4RecvErr(libc::sock_extended_err, Option<sockaddr_in>),
+    /// Socket error queue control messages read with the `MSG_ERRQUEUE` flag.
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    Ipv6RecvErr(libc::sock_extended_err, Option<sockaddr_in6>),
+
     /// Catch-all variant for unimplemented cmsg types.
     #[doc(hidden)]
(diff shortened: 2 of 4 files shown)


cemeyer:
[This failure is fun](https://api.cirrus-ci.com/v1/task/5351339917049856/logs/test.log).  What we've discovered is that `libc::ECONNREFUSED` is wrong on Linux-MIPS.  (It's 111 on x86, but 146 in the MIPS ABI.)

```
failures:

---- sys::test_socket::linux_errqueue::test_recverr_v4 stdout ----
  left: `111`,
 right: `146`', test/sys/test_socket.rs:1931:9
```

cemeyer:
The failing ARM tests ("Linux arm gnueabi," "Linux armv7 gnueabihf") are hitting some totally bogus Address Family 25951, along with "Linux mipsel."  But other QEMU targets pass (including 64-bit MIPS and BE 32-bit MIPS).  I guess it's somewhat plausible that some QEMU targets are more broken than others?

WGH-:
Is it qemu-system or qemu-user? If it's the latter, I'm not particularly surprised.

cemeyer:
> Is it qemu-system or qemu-user? If it's the latter, I'm not particularly surprised.

My guess is user, although I'm not sure how to tell from the cirrus config.

So, I think both QEMU emulation bugs relate to `do_sendrecvmsg_locked()` (QEMU user): https://github.com/qemu/qemu/blob/master/linux-user/syscall.c#L3282-L3289

Both problems appears to be bugs in `host_to_target_cmsg()`:
* `sa_family` for sockaddrs associated with `IP_RECVERR` control messages is being butchered on some platforms (looks like maybe 32- / 64-bit issue).
* `ee_errno` is not being correctly translated from host to target on platforms where `sa_family` was fine.

The `sa_family` one is tricky -- we do *appear* to do `host_to_target_sockaddr` translation.  Perhaps that isn't fully functional?  Alternatively, [if the control message does not contain an offender address, or if it does not exactly match the size we expect, we just skip straight to 'unimplemented' and do a memcpy instead of translating](https://github.com/qemu/qemu/blob/211364c21e7f757ae1acf4e72b5df39c498fb88b/linux-user/syscall.c#L1958-L1969).  Different alignment requirements between 32/64-bit might cause that, although I'm not seeing it immediately.

[`ee_errno` bug is pretty obvious here (and same for ipv6)](https://github.com/qemu/qemu/blob/211364c21e7f757ae1acf4e72b5df39c498fb88b/linux-user/syscall.c#L1970).

asomers:
Oof, that's a complicated test.  And the QEMU bug doesn't suprise me; there are plenty more where it came from.  This PR LGTM.  Just squash your commits and I'll merge it.

cemeyer:
Thanks!  Squashed.

asomers (approved):
bors r+

PR #1515, opened 2021-09-07
Add IP_TTL/IPV6_UNICAST_HOPS SockOpts

cemeyer:
Test: `cargo test --test test test_ttl_opts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1514](https://github.com/nix-rust/nix/pull/1514))
 - Added `AsRawFd` implementation on `PollFd`.
   (#[1516](https://github.com/nix-rust/nix/pull/1516))
+- Added `Ipv4Ttl` and `Ipv6Ttl` sockopts.
+  (#[1515](https://github.com/nix-rust/nix/pull/1515))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


cemeyer:
(Oops, initial version of the test was not `cfg`-limited to the same targets as the `SockOpt`s.)

asomers (commented):
Is there anything you can do to test that the sockopts actually work?

cemeyer:
> Is there anything you can do to test that the sockopts actually work?

Like, unit tests, or have I verified it works?

For unit tests, I'm not sure we can (or want to) rely on any particular internet host, and I don't know of any good way to simulate it.

Manually, yes, I've written [a little traceroute program](https://github.com/cemeyer/tracer_t) [using this new TTL functionality](https://github.com/cemeyer/tracer_t/blob/db1b7cadb81907249a65debfb4a48e10ed8667fa/src/main.rs#L60-L66) and it works:

```
tracer_t to 8.8.8.8, 12 hops max
 1  192.168.0.1: EHOSTUNREACH, false.
 2  63.231.10.66: EHOSTUNREACH, false.
 3  63.226.198.9: EHOSTUNREACH, false.
 4  67.14.41.158: EHOSTUNREACH, false.
 5  72.14.221.110: EHOSTUNREACH, false.
 6  10.252.161.62: EHOSTUNREACH, false.
 7  142.251.55.203: EHOSTUNREACH, false.
 8  8.8.8.8: ECONNREFUSED, true.
```

I observe, e.g., ICMP "Time Exceeded - TTL expired in transit" errors from hops 1-7 and "Unreachable - Port unreachable" from hop 8.  All of the UDP packets are sent to the destination (8.8.8.8), but the ICMP responses from all other addresses in the path are due to TTL expiration.

cemeyer:
(Rebased over #1516, no changes.)

asomers (approved):
bors r+

cemeyer:
Thanks!

PR #1516, opened 2021-09-07
Implement AsRawFd for PollFd

### LINKED ISSUE #1146: Expose RawFd in impl PollFd

cemeyer:
Implement the trait on PollFd, providing an `as_raw_fd()` accessor to
get the RawFd associated with the PollFd.

Subsumes #1147, #1286.  Closes #1146.

Test: `cargo test --test test test_pollfd_fd`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1511](https://github.com/nix-rust/nix/pull/1511))
 - Added `Ipv4RecvErr` and `Ipv6RecvErr` sockopts and associated control messages.
   (#[1514](https://github.com/nix-rust/nix/pull/1514))
+- Added `AsRawFd` implementation on `PollFd`.
+  (#[1516](https://github.com/nix-rust/nix/pull/1516))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


cemeyer:
Will have a CHANGELOG.md conflict with #1515, which is one of the annoying aspects of maintaining a Changelog file in-repo.

asomers (approved):
Yes, CHANGELOG conflicts are common and annoying.  The saltstack project does something cool, which is to create a separate file for the changelog entry of every PR.  Then when they make a release they run a script to merge those changelog entries into the main CHANGELOG.  I think something like that would work well for Nix.

bors r+

cemeyer:
Another changelog idea I've seen is formatting commits such that a tool can construct the document from them.  E.g., https://www.conventionalcommits.org/en/v1.0.0/ .

asomers:
I've seen that one.  But I don't think it would work.  Enforcing consistent commit message is harder than enforcing consistent changelog entries, especially on a project with so many different contributors.

cemeyer:
I think it should be enforceable via bors / pre-commit CI check, but I haven't investigated very hard.

PR #1518, opened 2021-09-08
Make Errno::result() inline

linkmauve:
This makes the success case a single `cmpl $-1, %eax` followed with a non-taken conditional jump, instead of the current `callq` which is always taken.

I have not benchmarked the difference, but it is an obvious improvement.


asomers (approved):
Thanks for this.  I usually frown on `#[inline]`, but if you've actually checked the assembly then it's definitely worthwhile.

bors r+

PR #1519, opened 2021-09-10
Update version in Cargo.toml to match crates.io version

coolreader18:
It looks like 0.22.1 was released on crates.io, but the in-tree Cargo.toml wasn't updated. Using `[patch]` in Cargo.toml is easier when the git version is up to date with the registry version.


asomers:
That's because the 0.22.1 release wasn't prepared from the master branch, it was prepared from the r0.22 branch.  What's now the master branch will (shortly, I hope) get released as 0.23.

PR #1520, opened 2021-09-12
mman module netbsd additions.

devnexen:


src/sys/mman.rs
```diff
@@ -351,6 +363,29 @@ pub unsafe fn mremap(
     }
 }
 
+#[cfg(target_os = "netbsd")]
```
asomers, line 366: This function already exists on Linux, and it looks to me like the only different is the operand order.  In the interest of portability, could you please share a single signature for the two operating systems?  You can use `cfg_if!` around the libc syscall to handle the different operand orders.

src/sys/mman.rs
```diff
@@ -150,6 +152,16 @@ libc_bitflags!{
...
```
asomers, line 155: You should share this definition with the Linux one.

src/sys/mman.rs
```diff
@@ -136,6 +136,8 @@ libc_bitflags!{
         MAP_NOCACHE;
         #[cfg(any(target_os = "ios", target_os = "macos"))]
         MAP_JIT;
+        #[cfg(target_os = "freebsd")]
+        MAP_ALIGNED_SUPER;
```
asomers, line 140: This seems unrelated to the mremap stuff.  Better to put it in its own PR.

asomers (changes requested):
Could you please add a test?  Probably you can just enable the existing `test_mremap_grow` and `test_mremap_shrink` for NetBSD?

asomers (approved):
LGTM, assuming you ran the tests on NetBSD.  CI doesn't.

bors r+

rtzoeller:
@asomers while testing #1624 I found that `test_mremap_shrink` was failing with `ENOMEM` on NetBSD. After investigating it seems like this test was never actually run on NetBSD when this PR was authored. 😞  The documentation added for `MAP_FIXED` is also incorrect, and the confusion seems to be why the test is failing.

I'll submit a PR (another day) to clean all this up.

---

The NetBSD man page for `mremap` states

```
     MAP_FIXED          newp is tried and mremap() fails if that address can't
                        be used as new base address for the range.  Otherwise,
                        oldp and newp are used as hints for the position,
                        factoring in the given alignment.
```

Whereas nix's documentation for the corresponding parameter `new_address` states

```
Permits to use the old and new address as hints to relocate the mapping.
```

I believe nix's documentation is completely backwards. If `MAP_FIXED` is _not_ specified, `newp` is just a hint. But if `MAP_FIXED` is specified, NetBSD attempts to use `newp` and fails if it can't. 

The test passes `None` for `newp`/`new_address`,  which is converted to `NULL` which will never work.

On a related note, I have a NetBSD VM now.

devnexen:
> @asomers while testing #1624 I found that `test_mremap_shrink` was failing with `ENOMEM` on NetBSD. 

What happens if you do `ulimit -c unlimited` ?

rtzoeller:
@devnexen no change, test still fails.

devnexen:
in my netbsd laptop 3 times out of 4 it passes however if indeed I put a address in the last argument it always pass.

PR #1521, opened 2021-09-12
Prevent buffer over-read in getgroups()

blyxxyz:
An edge case I found in another wrapper around `getgroups()`. @jhscheer pointed me to this implementation which had the same issue.


asomers (commented):
Good find!  Can you think of a way to add a good test case for this bug, or is it just too racy?

blyxxyz:
The only thing I can think of is spamming `setgroups()` in another thread and hoping that it happens over the course of a hundred calls, but that requires elevated permissions and it's dirty in lots of ways.

asomers (approved):
Yeah, it seems like a pretty tight race to hit.  I'll go ahead and merge it.

bors r+

PR #1522, opened 2021-09-12
mman mod adding MAP_ALIGNED_SUPER flag for freebsd.

devnexen:


asomers (changes requested):
Could we have a doc comment?  I no the other enum values don't, but we're trying to fix that.

asomers (changes requested):
CHANGELOG entry, please.

asomers (approved):
bors r+

PR #1525, opened 2021-09-17
mman: add MAP_EXCL flag for freebsd.

devnexen:


src/sys/mman.rs
```diff
@@ -40,6 +40,9 @@ libc_bitflags!{
         MAP_PRIVATE;
         /// Place the mapping at exactly the address specified in `addr`.
         MAP_FIXED;
+        /// To be used with MAP_FIXED context and will only succeeds if there is no existing mapping.
```
asomers, line 43:
```suggestion
        /// To be used with `MAP_FIXED`, to forbid the system to select a different address than the one specified.
```

asomers (changes requested):
Could you please add a CHANGELOG entry too?

asomers (approved):
bors r+

PR #1530, opened 2021-09-19
mman: Add mincore binding fixes #1528

### LINKED ISSUE #1528: Add mincore binding
It'd be nice to have `mincore` (https://man7.org/linux/man-pages/man2/mincore.2.html) bindings in this crate, which may be useful when we inspect page cache usage.

If no objection, I would like to send a patch later.

ocadaruma:
## Summary
- This PR adds `mincore` binding for Tier1 targets (linux, freebsd, macos)
- Also make `mman` tests runnable by adding it in parent module, with fixing some previous outdated mman tests

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -334,6 +334,7 @@ pub type mincore_vec_char_t = std::os::raw::c_char;
 /// # Safety
 ///
 /// `addr` and `vec` must meet all the requirements described in the `mincore(2)` man page.
+#[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "macos"))]
 pub fn mincore(addr: *mut c_void, length: size_t, vec: *mut mincore_vec_char_t) -> Result<()> {
     unsafe { Errno::result(libc::mincore(addr, length, vec)) }.map(drop)
 }
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -324,6 +324,20 @@ pub fn munlockall() -> Result<()> {
     unsafe { Errno::result(libc::munlockall()) }.map(drop)
 }
 
+#[cfg(any(target_os = "linux"))]
+pub type mincore_vec_char_t = std::os::raw::c_uchar;
+#[cfg(any(target_os = "freebsd", target_os = "macos"))]
+pub type mincore_vec_char_t = std::os::raw::c_char;
+
+/// Inspect page residency in memory at the address `addr` continuing for `length` bytes.
+///
+/// # Safety
+///
+/// `addr` and `vec` must meet all the requirements described in the `mincore(2)` man page.
+pub fn mincore(addr: *mut c_void, length: size_t, vec: *mut mincore_vec_char_t) -> Result<()> {
+    unsafe { Errno::result(libc::mincore(addr, length, vec)) }.map(drop)
+}
+
 /// allocate memory, or map files or devices into memory
 ///
 /// # Safety
--- a/test/sys/mod.rs
+++ b/test/sys/mod.rs
@@ -43,3 +43,4 @@ mod test_pthread;
 mod test_ptrace;
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod test_timerfd;
+mod test_mman;
--- a/test/sys/mod.rs
+++ b/test/sys/mod.rs
@@ -43,4 +43,8 @@ mod test_pthread;
 mod test_ptrace;
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod test_timerfd;
+#[cfg(any(target_os = "linux",
+          target_os = "netbsd",
+          target_os = "freebsd",
+          target_os = "macos"))]
 mod test_mman;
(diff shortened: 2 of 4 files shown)


src/sys/mman.rs
```diff
@@ -324,6 +324,21 @@ pub fn munlockall() -> Result<()> {
...
 }
 
+#[cfg(any(target_os = "linux"))]
+pub type mincore_vec_char_t = std::os::raw::c_uchar;
+#[cfg(any(target_os = "freebsd", target_os = "macos"))]
+pub type mincore_vec_char_t = std::os::raw::c_char;
+
+/// Inspect page residency in memory at the address `addr` continuing for `length` bytes.
+///
+/// # Safety
```
asomers, line 334: `Safety` sections normally only apply to `unsafe` functions.  Should this function be `unsafe`?  What would happen if the user supplies invalid arguments?

src/sys/mman.rs
```diff
@@ -324,6 +324,21 @@ pub fn munlockall() -> Result<()> {
...
+///
+/// `addr` and `vec` must meet all the requirements described in the `mincore(2)` man page.
+#[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "macos"))]
+pub fn mincore(addr: *mut c_void, length: size_t, vec: *mut mincore_vec_char_t) -> Result<()> {
```
asomers, line 338: Returning via a mutable pointer is not very Rusty.  It's also `unsafe`, because mincore has no way to know the lifetime of the pointed-to data.  Instead, you should return the data as a Vec.  You should also define an enum for the contents, instead of using c_char.  Finally, there's no reason why `addr` needs to be mutable.  So the signature should look like this:
```rust
libc_bitflags! {
    pub struct MincoreFlags {
        MINCORE_INCORE,
        ...
    }
}
pub fn mincore(addr: *const c_void, length: usize) -> Result<Vec<MincoreFlags>>
```

test/sys/test_mman.rs
```diff
@@ -1,16 +1,14 @@
...
+#[cfg(any(target_os = "linux", target_os = "netbsd"))]
 use nix::sys::mman::{mremap, MRemapFlags};
 
 #[test]
 fn test_mmap_anonymous() {
     let ref mut byte = unsafe {
         let ptr = mmap(std::ptr::null_mut(), 1,
                        ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
-                       MapFlags::MAP_PRIVATE | MapFlags::MAP_ANONYMOUS, -1, 0)
+                       MapFlags::MAP_PRIVATE | MapFlags::MAP_ANON, -1, 0)
```
asomers, line 11: Was there a need for this change or do you just prefer the shorter symbol?

test/sys/mod.rs
```diff
@@ -43,3 +43,8 @@ mod test_pthread;
 mod test_ptrace;
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod test_timerfd;
+#[cfg(any(target_os = "linux",
```
asomers, line 46: Oops.  Good catch.

asomers (changes requested):
Would you please mind splitting up this PR into two?  One for mincore, and one for fixing the mman tests.

test/sys/test_mman.rs
ocadaruma: The reason of the change was, `MAP_ANONYMOUS` is defined on only specific target (`android,linux,freebsd`) while `MAP_ANON` is defined for every os.

src/sys/mman.rs
ocadaruma: > So the signature should look like this

I see, the new signature looks much safe and easy to use.
Let me try.

src/sys/mman.rs
ocadaruma: In new suggested signature, I suppose there's no unsafety in this method. I'll fix doc comment.
https://github.com/nix-rust/nix/pull/1530#discussion_r711755282

ocadaruma:
> Would you please mind splitting up this PR into two

Sure. I'll submit another PR for fixing existing mman tests.

asomers:
> > Would you please mind splitting up this PR into two
> 
> Sure. I'll submit another PR for fixing existing mman tests.

Don't bother; I just did it myself.  I thought you might be asleep already.

ocadaruma:
Ah just noticed you already made a patch and I sent another PR. (#1535)
I'll close it so please go ahead

ocadaruma:
Hm.
I tried to refine the signature to be like this (https://github.com/nix-rust/nix/pull/1530#discussion_r711755282) but it turns out that two additional library call will be required for it.
The code will be roughly looks like below:
```rust
# to allocate sufficient memory for mincore vec, we have to get page size
let page_size = sysconf(PAGESIZE);

# allocate memory for mincore vector
let vec_mem = alloc(Layout::(...));

libc::mincore(...);
```

It sound like doing too much things implicitly, so I'm not sure if we should provide such "high-level" API in nix.

However, if we don't provide "high-level" signature for `mincore`, we have to provide allocate and pass mincore vec to get page residency info, it sounds like almost same thing as calling `libc::mincore` directly.

As the conclusion, I'll close the PR and the issue for now.
Thanks for your review.

PR #1535, opened 2021-09-19
mman: Fix mman tests to be runnable

ocadaruma:
# Summary
- (This change is submitted as #1530 originally and separated as new PR)
- Make mman tests runnable by adding it in parent module, with fixing some previous outdated mman tests


ocadaruma:
Duplicated as https://github.com/nix-rust/nix/pull/1534

PR #1536, opened 2021-09-19
Lower limits in setrlimit example

rtzoeller:
The current values causes the doc test to fail on Fedora 34 for unprivileged users. The values can be lowered without meaningfully changing the example.

Previously the example failed with EPERM.

### DIFF
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -200,8 +200,8 @@ pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>
 /// ```
 /// # use nix::sys::resource::{setrlimit, Resource};
 ///
-/// let soft_limit = Some(1024);
-/// let hard_limit = Some(1048576);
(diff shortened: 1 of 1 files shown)


rtzoeller:
Obtained the new value by building and running the following C program as a non-root user:

```c
#include <stdio.h>
#include <sys/resource.h>

int main() {
    struct rlimit rlimit;
    int retval = getrlimit(RLIMIT_NOFILE, &rlimit);
    printf("%d\n", rlimit.rlim_cur);
    printf("%d\n", rlimit.rlim_max);
}
```

Results in

```
1024
524288
```

rtzoeller:
Alternatively, this code block could be marked as `no_run`.

asomers (changes requested):
I think you should make the limit way smaller, like maybe 512/1024 or something.  There's no reason it should be this high.

asomers (approved):

asomers:
I guess I forgot to order bors around earlier.

bors r+

PR #1538, opened 2021-09-23
posix_fadvise doesn't return -1 as sentinel value

ocadaruma:
## Summary
- `posix_fadvise(2)` does return error number directly (i.e. not through `errno`)
  * refs: https://man7.org/linux/man-pages/man2/posix_fadvise.2.html , https://man7.org/linux/man-pages/man2/posix_fadvise.2.html
- However `posix_fadvise`-binding uses `Errno::result` to translate the error now, which is mis-use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,11 +83,15 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- `posix_fadvise` now returns errors in the conventional way, rather than as a
+  non-zero value in `Ok()`.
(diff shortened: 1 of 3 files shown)


asomers (commented):
Thanks, this looks like a good change.  But why just Linux and FreeBSD?  I think it probably returns directly on every OS.

ocadaruma:
> I think it probably returns directly on every OS

H-m, sounds correct. Let me update.

asomers (approved):
LGTM.

PR #1539, opened 2021-09-25
pthread_setname_np/pthread_getname_np wrappers.

devnexen:
linux support only for now.


src/sys/pthread.rs
```diff
@@ -36,3 +38,22 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
     let res = unsafe { libc::pthread_kill(thread, sig) };
     Errno::result(res).map(drop)
 }
+
+/// Obtain the name of the thread
+///
+/// On linux the name cannot exceed 16 length including null terminator
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
```
asomers, line 45: NetBSD and FreeBSD also support these functions.  You should include them, though first you'll have to add FreeBSD's to libc.

src/sys/pthread.rs
```diff
@@ -36,3 +38,22 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
...
+pub fn pthread_getname_np(thread: Pthread) -> Result<String> {
+    let mut name = [0u8; 16];
```
asomers, line 47: The 16 bytes thing is Linux specific.  On NetBSD it's controlled by the `PTHREAD_MAX_NAMELEN_NP` constant, currently 32 bytes.  On FreeBSD it's not clearly documented, but the limit is `MAXCOMLEN + 1`, which is 20.

src/sys/pthread.rs
```diff
@@ -36,3 +38,22 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
...
+    unsafe { libc::pthread_getname_np(thread, name.as_mut_ptr() as _, name.len()) };
+    let cname = unsafe { CStr::from_ptr(name.as_ptr() as _) };
+    Ok(cname.to_owned().to_string_lossy().to_string())
+}
+
+/// Set the name of the thread
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
+pub fn pthread_setname_np(thread: Pthread, name: String) {
+    let cname = CString::new(name).expect("name failed");
```
asomers, line 56: Better to return an error here instead of panic.

src/sys/pthread.rs
```diff
@@ -36,3 +38,22 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
...
+pub fn pthread_getname_np(thread: Pthread) -> Result<String> {
+    let mut name = [0u8; 16];
+    unsafe { libc::pthread_getname_np(thread, name.as_mut_ptr() as _, name.len()) };
+    let cname = unsafe { CStr::from_ptr(name.as_ptr() as _) };
+    Ok(cname.to_owned().to_string_lossy().to_string())
+}
+
+/// Set the name of the thread
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
+pub fn pthread_setname_np(thread: Pthread, name: String) {
```
asomers, line 55: You could be a little more generic here and accept any `T: Into<Vec<u8>>`.  That way the caller could provide, for example, a byte string `b"nix-rust"`.

src/sys/pthread.rs
```diff
@@ -36,3 +38,22 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
     let res = unsafe { libc::pthread_kill(thread, sig) };
     Errno::result(res).map(drop)
 }
+
+/// Obtain the name of the thread
+///
+/// On linux the name cannot exceed 16 length including null terminator
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
+pub fn pthread_getname_np(thread: Pthread) -> Result<String> {
```
asomers, line 46: I suggesting returning `CString` and letting the caller convert it if he wants to.

asomers (changes requested):

src/sys/pthread.rs
devnexen: I know but I wanted to see first it would appeal ;-) I added missing calls earlier anyway.

src/sys/pthread.rs
devnexen: very true but as mentionned in the commit message I focus on Linux for now.

devnexen:
before moving further (libc had been updated since) what do you think ?

src/sys/pthread.rs
```diff
@@ -43,17 +43,24 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
 ///
 /// On linux the name cannot exceed 16 length including null terminator
 #[cfg(all(target_os = "linux", not(target_env = "musl")))]
-pub fn pthread_getname_np(thread: Pthread) -> Result<String> {
-    let mut name = [0u8; 16];
+pub fn pthread_getname_np(thread: Pthread) -> Result<CString> {
+    let mut name: Vec<u8> = Vec::with_capacity(16);
+    unsafe { name.set_len(16); }
```
asomers, line 48: This is technically UB.  You shouldn't call `set_len` until after the the FFI function returns successfully, and you should only set it to as many bytes as the FFI function actually wrote.  In this case, `pthread_getname_np` doesn't actually return the length of the name, so you'll have to fully initialize the buffer anyway.

src/sys/pthread.rs
```diff
@@ -43,17 +43,24 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
...
     unsafe { libc::pthread_getname_np(thread, name.as_mut_ptr() as _, name.len()) };
-    let cname = unsafe { CStr::from_ptr(name.as_ptr() as _) };
-    Ok(cname.to_owned().to_string_lossy().to_string())
+    let zi = name.iter().position(|x| *x == b'\0').unwrap();
```
asomers, line 50: This three lines can be simplified.  If you preinitialize the buffer to zero, then all you'll have to do is `CString::new(name)`.

src/sys/pthread.rs
```diff
@@ -36,3 +38,29 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
...
+/// Set the name of the thread
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
+pub fn pthread_setname_np<T: Into<Vec<u8>>>(thread: Pthread, name: T) -> Result<()> {
+    match CString::new(name.into()) {
+        Ok(cname) => {
+            let nameptr = cname.as_ptr();
+            let res = unsafe { libc::pthread_setname_np(thread, nameptr) };
+            Errno::result(res).map(drop)
+        },
+        Err(_) => Ok(()),
```
asomers, line 64: Don't blindly ignore errors.  You should return an error in this case.

src/sys/pthread.rs
```diff
@@ -36,3 +38,29 @@ pub fn pthread_kill<T: Into<Option<Signal>>>(thread: Pthread, signal: T) -> Resu
...
+pub fn pthread_getname_np(thread: Pthread) -> Result<CString> {
+    let mut name: Vec<u8> = Vec::with_capacity(16);
+    unsafe { name.set_len(16); }
+    unsafe { libc::pthread_getname_np(thread, name.as_mut_ptr() as _, name.len()) };
+    let zi = name.iter().position(|x| *x == b'\0').unwrap();
+    name.truncate(zi);
+    Ok(unsafe { CString::from_vec_unchecked(name) })
+}
+
+/// Set the name of the thread
```
asomers, line 55: A doc test would be good here, especially because it's not obvious what the type of the argument should be.

test/sys/test_pthread.rs
```diff
@@ -20,3 +20,16 @@ fn test_pthread_kill_none() {
     pthread_kill(pthread_self(), None)
         .expect("Should be able to send signal to my thread.");
 }
+
+#[test]
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
+fn test_pthread_name_np() {
+    let tid = pthread_self();
+    let name = b"nix-rust";
```
asomers, line 28: The fact that it takes three lines to put `name` into the right format suggests that the function's interface isn't well-designed.  The goal is to be able to call it in any convenient way, like this:
```rust
pthread_setname_np(tid, "foo");    // Rust string
pthread_setname_np(tid, b"foo");    // Byte vector
pthread_setname_np(tid, CString::new("foo"));    // C String
```

test/sys/test_pthread.rs
```diff
@@ -20,3 +20,16 @@ fn test_pthread_kill_none() {
...
+    let namestr = std::str::from_utf8(name).unwrap();
+    pthread_setname_np(tid, name.to_vec()).expect("Should pass");
```
asomers, line 30: `expect` should only be used in combination with a useful message.  If you're just going to say "Should pass", then you might as well just use `unwrap()`.

asomers (changes requested):

Issue #1544, opened 2021-09-28
Change Sockaddr to a union


asomers:
PR #1504 addresses a few safety and soundness issues with our `Sockaddr` type, which all boil down to "We should never create a `&T` reference if the entire T isn't initialized".  But in reviewing it, I realized another problem: many Nix functions force the user to allocate a struct large enough for any type of sockaddr, even if they know exactly which type they need.  For example, `nix::sys::socket::bind` takes an `enum Sockaddr` as its argument, which can be very large, even though all it does is immediately transform that argument into a pointer.  Many Nix users undoubtedly know which type of Sockaddr they need before they call bind, but we force them to allocate a whole `enum Sockaddr`.  And that begs the question: why use an enum at all?  These types are basically already an enum in C.  They can all be cast to and from sockaddr and sockaddr_storage, using the s*_family field as a discriminator.  So here is my proposal for how to fix the problems raised by #1504 , stop requiring such large allocations, and make the sockaddr types more ergonomic too:

* Create a `SockaddrLike` trait, mostly because for the `as_ffi_pair` method.  The trait will be sealed, because it requires the `sa_family_t` and `sa_len` fields to be located at fixed offsets.
* Create newtypes for every specific sockaddr type, all with `#[repr(transparent)]`.
* Create a Rust union for `SockaddrStorage`.  Thanks to the `sa_family_t` tag, this union will be safely, falliably convertible to the other sockaddr types.
* Change all functions like `bind` to accept a generic parameter of any `SockaddrLike` type.
* Deprecate `enum Sockaddr` and related functions.

@coolreader18 what do you think of this idea?  Here's an outline of the code:
```rust
/// Anything that, in C, can be cast back and forth to `sockaddr`.
pub trait SockaddrLike: private::Sealed {
    const FAMILY: AddressFamily;

    // Unsafe constructor from a variable length source
    // Some C APIs from provide `len`, and others do not.  If it's provided it
    // will be validated.  If not, it will be guessed based on the family.
    unsafe from_raw(addr: *const libc::sockaddr, len: Option<libc::socklen_t>)
        -> Result<Self> {...}

    // The next three methods have the same implementation for all concrete
    // types, since the trait is sealed!  This works because all implementors
    // are #[repr(transparent)] and the socklen and sa_family fields are located
    // at the same offset for all structs
    fn family(&self) -> AddressFamily {...}
    fn socklen(&self) -> socklen_t {...}
    // Used for many syscalls
[...]

PR #1546, opened 2021-09-29
Revert "Expose SockAddr::from_raw_sockaddr"

asomers:
This reverts commit ed43d2c65e65dd68c9cf2dcf06f5ec45a44aaccd.

As discussed in #1544 the API of this function needs to change.  For
now, revert the PR that made it public, because it has not yet been
included in any release.


asomers:
cc @coolreader18

PR #1549, opened 2021-10-03
unistd porting gettid to other oses

devnexen:


asomers (changes requested):
I'm going to have to say no to this PR.  Nix really isn't the right place for cross-platform abstractions.  It's too low level.  We've tried doing that in the past, but inevitably the different implementations aren't exactly equivalent, which causes confusion.  Best to leave this stuff to higher level crates.

devnexen:
understood.

Issue #601, opened 2017-05-04
Add prctl(2) support


cyphar:
I'm working on an [alternative container init and signal forwarder](https://github.com/cyphar/initrs), and a nice feature would be to be able to run it as non-pid1. However, that requires being able to set PR_SET_CHILD_SUBREAPER from within Rust.

While it's possible to do this:

```rust
use libc::{prctl, PR_SET_CHILD_SUBREAPER};
unsafe { prctl(PR_SET_CHILD_SUBREAPER, 1, 0, 0); }
```

The `unsafe` is unsightly.

Susurrus:
Indeed, it seems reasonable to add `prctl(2)`. Would you be willing to submit a PR for this?

cyphar:
I'm not a very proficient user of Rust yet (I just started picking it up last weekend), but I believe that [the `prctl` crate](https://github.com/viraptor/prctl-rs) could be a reasonable starting point. The biggest thing that makes `prctl(2)` ugly and unsafe is that the arguments are sometimes read and sometimes written to by the kernel.

Susurrus:
If you want to have a go at it, I'll mentor you. This crate does very low level abstractions, so basically what you already wrote puts you half way there I'd think. We wouldn't try to offer a more comprehensive high-level API as that crate seems to be aiming for. Otherwise I don't know how likely it is for someone to tackle this in the near future.

SteveLauC:
Close as completed in #1550

PR #1550, opened 2021-10-03
Add the Linux prctl syscall

DOhlsson:
I have added support for the prctl syscall.

However, I am not sure that this is the right way to do it. Another way would be to wrap each possible flag to prctl as it's own function, that way it would be more similar to how the [prctl](https://github.com/viraptor/prctl-rs) crate does it. Wrapping the different options would be safer, since some options to prctl return data through pointers.

What do you think?

### DIFF
--- a/src/sys/prctl.rs
+++ b/src/sys/prctl.rs
@@ -1,0 +1,208 @@
+//! prctl is a Linux-only API for performing operations on a process or thread.
+//!
+//! Note that careless use of some prctl() operations can confuse the user-space run-time
+//! environment, so these operations should be used with care.
+//!
+//! For more documentation, please read [prctl(2)](https://man7.org/linux/man-pages/man2/prctl.2.html).
+
+use crate::errno::Errno;
+use crate::sys::signal::Signal;
+use crate::Result;
+
+use libc::{c_int, c_ulong};
+use std::convert::TryFrom;
+use std::ffi::{CStr, CString};
+
+libc_enum! {
+    /// The type of hardware memory corruption kill policy for the thread.
+
+    #[repr(i32)]
+    #[non_exhaustive]
+    #[allow(non_camel_case_types)]
+    pub enum PrctlMCEKillPolicy {
+        /// The thread will receive SIGBUS as soon as a memory corruption is detected.
+        PR_MCE_KILL_EARLY,
+        /// The process is killed only when it accesses a corrupted page.
+        PR_MCE_KILL_LATE,
+        /// Uses the system-wide default.
+        PR_MCE_KILL_DEFAULT,
+    }
+    impl TryFrom<i32>
+}
+
+fn prctl_set_bool(option: c_int, status: bool) -> Result<()> {
+    let res = unsafe { libc::prctl(option, status as c_ulong, 0, 0, 0) };
+    Errno::result(res).map(drop)
+}
+
+fn prctl_get_bool(option: c_int) -> Result<bool> {
+    let res = unsafe { libc::prctl(option, 0, 0, 0, 0) };
+    Errno::result(res).map(|res| res != 0)
+}
+
+/// Set the "child subreaper" attribute for this process
+pub fn set_child_subreaper(attribute: bool) -> Result<()> {
+    prctl_set_bool(libc::PR_SET_CHILD_SUBREAPER, attribute)
+}
+
+/// Get the "child subreaper" attribute for this process
+pub fn get_child_subreaper() -> Result<bool> {
+    // prctl writes into this var
+    let mut subreaper: c_int = 0;
+
+    let res = unsafe { libc::prctl(libc::PR_GET_CHILD_SUBREAPER, &mut subreaper, 0, 0, 0) };
+
+    Errno::result(res).map(|_| subreaper != 0)
+}
+
+/// Set the dumpable attribute which determines if core dumps are created for this process.
+pub fn set_dumpable(attribute: bool) -> Result<()> {
+    prctl_set_bool(libc::PR_SET_DUMPABLE, attribute)
+}
+
+/// Get the dumpable attribute for this process.
+pub fn get_dumpable() -> Result<bool> {
+    prctl_get_bool(libc::PR_GET_DUMPABLE)
+}
+
+/// Set the "keep capabilities" attribute for this process. This causes the thread to retain
+/// capabilities even if it switches its UID to a nonzero value.
+pub fn set_keepcaps(attribute: bool) -> Result<()> {
+    prctl_set_bool(libc::PR_SET_KEEPCAPS, attribute)
+}
+
+/// Get the "keep capabilities" attribute for this process
+pub fn get_keepcaps() -> Result<bool> {
+    prctl_get_bool(libc::PR_GET_KEEPCAPS)
+}
+
+/// Clear the thread memory corruption kill policy and use the system-wide default
+pub fn clear_mce_kill() -> Result<()> {
+    let res = unsafe { libc::prctl(libc::PR_MCE_KILL, libc::PR_MCE_KILL_CLEAR, 0, 0, 0) };
+
+    Errno::result(res).map(drop)
+}
+
+/// Set the thread memory corruption kill policy
+pub fn set_mce_kill(policy: PrctlMCEKillPolicy) -> Result<()> {
+    let res = unsafe {
+        libc::prctl(
+            libc::PR_MCE_KILL,
+            libc::PR_MCE_KILL_SET,
+            policy as c_ulong,
+            0,
+            0,
+        )
+    };
+
+    Errno::result(res).map(drop)
+}
+
+/// Get the thread memory corruption kill policy
+pub fn get_mce_kill() -> Result<PrctlMCEKillPolicy> {
+    let res = unsafe { libc::prctl(libc::PR_MCE_KILL_GET, 0, 0, 0, 0) };
+
+    match Errno::result(res) {
+        Ok(val) => Ok(PrctlMCEKillPolicy::try_from(val)?),
+        Err(e) => Err(e),
+    }
+}
+
+/// Set the parent-death signal of the calling process. This is the signal that the calling process
+/// will get when its parent dies.
+pub fn set_pdeathsig<T: Into<Option<Signal>>>(signal: T) -> Result<()> {
+    let sig = match signal.into() {
+        Some(s) => s as c_int,
+        None => 0,
+    };
+
+    let res = unsafe { libc::prctl(libc::PR_SET_PDEATHSIG, sig, 0, 0, 0) };
+
+    Errno::result(res).map(drop)
+}
+
+/// Returns the current parent-death signal
+pub fn get_pdeathsig() -> Result<Option<Signal>> {
+    // prctl writes into this var
+    let mut sig: c_int = 0;
+
+    let res = unsafe { libc::prctl(libc::PR_GET_PDEATHSIG, &mut sig, 0, 0, 0) };
+
+    match Errno::result(res) {
+        Ok(_) => Ok(match sig {
+            0 => None,
+            _ => Some(Signal::try_from(sig)?),
+        }),
+        Err(e) => Err(e),
+    }
+}
+
+/// Set the name of the calling thread. Strings longer than 15 bytes will be truncated.
+pub fn set_name(name: &CStr) -> Result<()> {
+    let res = unsafe { libc::prctl(libc::PR_SET_NAME, name.as_ptr(), 0, 0, 0) };
+
+    Errno::result(res).map(drop)
+}
+
+/// Return the name of the calling thread
+pub fn get_name() -> Result<CString> {
+    // Size of buffer determined by linux/sched.h TASK_COMM_LEN
+    let buf = [0u8; 16];
+
+    let res = unsafe { libc::prctl(libc::PR_GET_NAME, &buf, 0, 0, 0) };
+
+    let len = buf.iter().position(|&c| c == 0).unwrap_or(buf.len());
+    let name = CStr::from_bytes_with_nul(&buf[..=len]).map_err(|_| Errno::EINVAL)?;
+
+    Errno::result(res).map(|_| name.to_owned())
+}
+
+/// Sets the timer slack value for the calling thread. Timer slack is used by the kernel to group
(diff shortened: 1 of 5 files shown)


asomers (changes requested):
The first question is: "if the prctl crate is good enough, then why does Nix need these bindings too?"
Secondly, I don't think the proposed API is very good.  Those four `c_ulong` arguments look like they have different meanings for different PrctlOption values, and that's not very Rusty.  One alternative would be to specify the option as an enum with a variable number arguments for each value.  Another, as you suggested, would be to turn each prctl option into its own function.  Not being familiar with prctl myself, I can't say which is best.

DOhlsson:
I forgot to mention the related issue (#601) in my first post.

To answer your first question, I think that it would be nice to have prctl implemented in Nix. There are other crates for inotify and epoll too, but if I can get all three of them in one crate then that is my preferred option.

I agree that my pull request isn't very good. But I am willing to put in the work to bring it up to Nix's standards.

The weird thing about prctl is that some options can return data through the function call and other options return data to the location pointed to by one of the arguments, this behaviour makes it quite unsafe. Hence why I think that making wrapper functions is the best option.

I'll work a bit on this PR and come back with the better solution.

DOhlsson:
What is your opinion on the solution now? I have not implemented all the options yet as I wanted to hear your opinion first.

Possibly the prctl function and PrctlOption enum do not have to be public anymore, if all possibilities are implemented as separate functions.

test/sys/mod.rs
```diff
@@ -45,3 +45,5 @@ mod test_pthread;
 mod test_ptrace;
 #[cfg(any(target_os = "android", target_os = "linux"))]
 mod test_timerfd;
+#[cfg(target_os = "linux")]
```
asomers, line 48:
```suggestion
#[cfg(target_os = "linux")]
#[cfg(feature = "process")]
```

src/sys/prctl.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/// Manipulates various aspects of the behaviour of the calling process or thread.
+///
+/// # Safety
+/// Note that careless use of some prctl() operations can confuse the user-space run-time
+/// environment, so these operations should be used with care.
+///
+/// For more information, see [prctl(2)](https://man7.org/linux/man-pages/man2/prctl.2.html).
+pub unsafe fn prctl(
```
asomers, line 221: I don't think there's any reason to leave this in here.  It's no better than calling libc directly.

src/sys/prctl.rs
```diff
@@ -0,0 +1,229 @@
...
+    Errno::result(res).map(drop)
+}
+
+/// Get the thread memory corruption kill policy
+pub fn get_mce_kill() -> Result<PrctlMCEKillPolicy> {
+    let res = unsafe {
+        libc::prctl(PR_MCE_KILL_GET as c_int, 0, 0, 0, 0)
+    };
+
+    Errno::result(res).map(|res| PrctlMCEKillPolicy::try_from(res).unwrap())
```
asomers, line 211: This unwrap could panic in the future if the kernel adds a new kill policy.  Better to return an error in that case.

test/sys/test_prctl.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    prctl::set_mce_kill(PR_MCE_KILL_LATE).unwrap();
+    let mce = prctl::get_mce_kill().unwrap();
+    assert_eq!(mce, PR_MCE_KILL_LATE);
+
+    prctl::clear_mce_kill().unwrap();
+    let mce = prctl::get_mce_kill().unwrap();
+    assert_eq!(mce, PR_MCE_KILL_DEFAULT);
+
+
```
asomers, line 45:
```suggestion
```

asomers (commented):
The new methods look a lot safer.  That's what I had in mind.  Also, in the tests, don't forget to clean up after yourself!

DOhlsson:
Thank you for your feedback! I have a couple more of the options to add (and fix the tests), but I hope to get some time to get it done this week.

What is your opinion on the more exotic options that are available through prctl, should I implement them as well in this PR? Like architecture specific options, that I don't have the hardware to test with (powerpc, ia, mips etc.) and options to control capabilities where there are recommended higher-level libraries like libcap.

I have implemented the process-control options and left out the more exotic options for now.

src/sys/prctl.rs
DOhlsson: Agreed, I have removed it now.

src/sys/prctl.rs
DOhlsson: Fixed it :+1:

DOhlsson:
@asomers I can't seem to figure out what is going wrong in the builds. Do you have any idea? Could you point me in the right direction?

The errors are in architectures I don't have access to, namely MIPS, MIPS64, mipsel, arm gnueabi and armv7 gnueabihf. It appears to be pointer related, as the error EFAULT from prctl indicates that it is an invalid address.

I pass the pointer to prctl as a `&mut c_int` cast as `c_ulong` so there might be something architecture specific that I do not know.

Do you know of any examples in the rest of the Nix code where a pointer is passed to and written by the syscall?

asomers:
uclibceabihf is failing through no fault of your own, and hopefully it will soon be fixed on the master branch.  arm gnueabi and armv7 gnueabihf are tested using QEMU.  QEMU doesn't implement all syscalls.  If prctl is one, then you can add `#[cfg_attr(qemu, ignore)]` to those test cases.

DOhlsson:
@asomers I have fixed the tests and completed a bunch more options. I think this PR is ready if you are satisfied with it.

DOhlsson:
@asomers What do you think? Is this PR ready or is there anything I should change?

ColonelThirtyTwo:
I'm interested in this too, particularly setting no new privileges.

test/sys/test_prctl.rs
```diff
@@ -0,0 +1,106 @@
...
+    let long_name = String::from("0123456789abcdefghijklmn");
+    prctl::set_name(&long_name).unwrap();
+    let res = prctl::get_name().unwrap();
+    assert_eq!(&long_name[..15], res); // name truncated by kernel
+
+    let short_name = String::from("01234567");
+    prctl::set_name(&short_name).unwrap();
+    let res = prctl::get_name().unwrap();
+    assert_eq!(short_name, res);
+}
```
asomers, line 71: Needs cleanup

test/sys/test_prctl.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[cfg_attr(qemu, ignore)]
+#[test]
+fn test_get_set_timerslack() {
+    let slack = 60_000;
+    prctl::set_timerslack(slack).unwrap();
+    let res = prctl::get_timerslack().unwrap();
+    assert_eq!(slack, res as u64);
+}
```
asomers, line 80: Needs cleanup

test/sys/test_prctl.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[test]
+fn test_get_set_no_new_privs() {
+    let res = prctl::get_no_new_privs().unwrap();
+    assert_eq!(res, false);
+
+    prctl::set_no_new_privs().unwrap();
+    let res = prctl::get_no_new_privs().unwrap();
+    assert_eq!(res, true);
+}
```
asomers, line 96: Needs cleanup

test/sys/test_prctl.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[test]
+fn test_get_set_thp_disable() {
+    let res = prctl::get_thp_disable().unwrap();
+    assert_eq!(res, false);
+
+    prctl::set_thp_disable(true).unwrap();
+    let res = prctl::get_thp_disable().unwrap();
+    assert_eq!(res, true);
+}
```
asomers, line 106: Needs cleanup

asomers (changes requested):
All of the tests seem to assume that the default settings will be present.  Are you sure that will be true, or could it depend on the environment? Also, is it possible for any of them to interfere with other tests, including in other modules?  It makes me nervous that they all change global settings.

src/sys/prctl.rs
```diff
@@ -0,0 +1,213 @@
...
+    let name = CString::new(name.as_bytes()).unwrap();
+
+    let res = unsafe { libc::prctl(libc::PR_SET_NAME, name.as_ptr(), 0, 0, 0) };
+
+    Errno::result(res).map(drop)
+}
+
+/// Return the name of the calling thread
+pub fn get_name() -> Result<String> {
+    let buf = [32u8; 16];
```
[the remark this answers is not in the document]
DOhlsson: I was confused about CStr, but now I see that 0u32 also works.

The manpage for prctl says the buffer should be 16 bytes. I looked into it and it is because TASK_COMM_LEN is defined as 16 in `include/linux/sched.h` in the kernel. It's the length of the comm field in task_struct. Sadly not defined in libc.

src/sys/prctl.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    let res = unsafe { libc::prctl(libc::PR_GET_NAME, &buf, 0, 0, 0) };
+
+    let ptr = buf.as_ptr() as *const c_char;
+    let cstr = unsafe { CStr::from_ptr(ptr) };
```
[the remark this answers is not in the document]
DOhlsson: Good suggestion, now I don't need the unsafe `from_ptr`.

test/sys/test_prctl.rs
```diff
@@ -0,0 +1,106 @@
...
+    let sig = prctl::get_pdeathsig().unwrap();
+    assert_eq!(sig, Some(Signal::SIGUSR1));
+}
+
+#[test]
+fn test_get_set_name() {
+    let long_name = String::from("0123456789abcdefghijklmn");
+    prctl::set_name(&long_name).unwrap();
+    let res = prctl::get_name().unwrap();
+    assert_eq!(&long_name[..15], res); // name truncated by kernel
```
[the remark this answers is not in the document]
DOhlsson: See my reply on the function definition. It's defined in the kernel.

test/sys/test_prctl.rs
DOhlsson: This one can't easily be cleaned up. It is an irreversible operation. It acts on the thread so I don't think it should affect other tests.

DOhlsson:
> All of the tests seem to assume that the default settings will be present. Are you sure that will be true, or could it depend on the environment?

I have changed the tests so that they do not assume that the defaults are set. And I've also added cleanup that resets the settings to what they were before the test.

> Also, is it possible for any of them to interfere with other tests, including in other modules? It makes me nervous that they all change global settings.

Some of these affect the process and not only the thread, so it might be possible that these tests affect other tests. 
I'll move the tests into their own test target.

test/sys/test_prctl.rs
asomers: But if `--test-threads=1`, then Rust's test harness doesn't create a new thread per test.  It just reuses one.  So it might affect other tests.  I suggest that you either explicitly spawn a thread for this test, or else create its own test program.

asomers:
I see, you did move them into their own process.  That looks good.  Would you mind squashing your commits now?

DOhlsson:
@asomers I have squashed the commits. I also added a line to the changelog.

test/sys/test_prctl.rs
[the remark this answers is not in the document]
DOhlsson: I also added a comment explaining where the magic number comes from. Is that enough or would you suggest I do something else about the magic number?

DOhlsson:
@asomers is it ready to merge or is there anything else I should do?

src/sys/prctl.rs
```diff
@@ -0,0 +1,211 @@
...
+    let buf = [0u8; 16];
+
+    let res = unsafe { libc::prctl(libc::PR_GET_NAME, &buf, 0, 0, 0) };
+
+    let len = buf.iter().position(|&c| c == 0).unwrap_or(buf.len());
+    let cstr = CStr::from_bytes_with_nul(&buf[..=len]).map_err(|_| Errno::EINVAL)?;
+    let name = cstr.to_str().map_err(|_| Errno::EINVAL)?;
+
+    Errno::result(res).map(|_| name.to_owned())
+}
```
eric-seppanen, line 161: If `set_name` were to switch to `&[u8]` then perhaps this should change the return value from `String` to `Box<[u8]>`?

Then the caller could decide how to handle non-UTF-8 data.

eric-seppanen: On second thought, `Vec<u8>` might be better than `Box<[u8]>`, to make it easier for the caller to do `String::from_utf8(name)` afterward.

src/sys/prctl.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Set the name of the calling thread. Strings longer than 15 bytes will be truncated.
+pub fn set_name(name: &str) -> Result<()> {
+    let name = CString::new(name.as_bytes()).unwrap();
+
+    let res = unsafe { libc::prctl(libc::PR_SET_NAME, name.as_ptr(), 0, 0, 0) };
+
+    Errno::result(res).map(drop)
+}
```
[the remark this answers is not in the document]
DOhlsson: @eric-seppanen Changing to `&[u8]` would not prevent characters to be cut off in the middle? If the user just does an `as_bytes()` on their string it could still truncate in the middle of a character.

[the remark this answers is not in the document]
eric-seppanen: That's true. My concern was that `&str` implies that unicode will be handled appropriately, which isn't a promise this function can make.

Maybe `OsString` would be the right choice? It implies that there's a string, but clearly expresses to the caller that the handling of the bytes is os-specific.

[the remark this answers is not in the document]
DOhlsson: I don't think there is any convenient way to convey that the input string could be truncated. In the end I settled on CStr because it hints at the underlying C code and because it's the type of string least likely to be UTF-8.

src/sys/prctl.rs
DOhlsson: Same here as on the previous comment. I settled for CString because then the user will be forced to handle the utf8 conversion themselves.

DOhlsson:
@asomers I have rebased this branch onto latest commit from master. Can this PR be merged as is, or would you like me to change anything else?

asomers (approved):

PR #1560, opened 2021-10-15
Vendor the bitflags crate

### LINKED ISSUE #1555: Cannot have nix `0.20` and `0.23` in same package due to conflicting `bitflags` requirement
In `Cargo.toml`

```toml
[dependencies]
nix = "0.23.0"
zbus = "1.9.2"
```

This leads to

```rust
error: failed to select a version for `bitflags`.
    ... required by package `nix v0.20.2`
    ... which is depended on by `zbus v1.9.2`
    ... which is depended on by `whyyy v0.1.0 (/private/tmp/whyyy)`
versions that meet the requirements `>=1.1.0, <1.3.0` are: 1.2.1, 1.2.0, 1.1.0

all possible versions conflict with previously selected packages.

  previously selected package `bitflags v1.3.1`
    ... which is depended on by `nix v0.23.0`
    ... which is depended on by `whyyy v0.1.0 (/private/tmp/whyyy)`

failed to select a version for `bitflags` which could resolve this conflict
```

Because `v0.23` needs `bitflags > 1.3` but `v0.20` need `bitflags < 1.3`

asomers:
Bitflags has a policy of changing the MSRV in minor releases.  This is a
breaking change, which semver does not allow in minor releases.  Nix
adjusts by locking bitflags to a specific minor version.  But Cargo does
not allow multiple semver-compatible versions of a crate to be included
in the same build, with the result that crates which use both Nix 0.23
and also use bitflags 1.2 cannot compile.
    
Fix the problem by vendoring the bitflags crate, removing the external
dependency completely.

Fixes #1555


asomers:
@matklad that's an interesting idea.  It's the first real alternative to "everybody should run stable" that I've heard.  And it until Cargo gets an msrv-aware resolver, it looks like it could work.  Would you be willing to submit a PR to set it up?  The biggest headache, I think, will be maintaining that file.  Are you aware of any tools that can help?

matklad:
Yeah, let's try this: https://github.com/nix-rust/nix/pull/1592 (note: that's intentionally not based on the current master, to verify that the trick could have prevented bitflags breaking the CI). 

>Are you aware of any tools that can help?

`cargo +nightly generate-lockfile -Zminimal-versions` gets the job done, provided that you and deps are minimal-version correct. If that' not the case, I just manually downgrade the offending packages with `cargo update -p offending-package --precise an.older.version`.

asomers:
Sadly, `-Zminimal-versions` doesn't work for most complex crates, because it requires all of your deps to be minimal-version correct, too.  There's been talk about fixing that by making it downgrade all immediate deps to the minimal versions, but use the latest version for transitive deps, but I don't think anybody has implemented that yet.

asomers:
Closing in favor of #1593

PR #1566, opened 2021-10-16
Correct definition of MADV_SOFT_OFFLINE on ppc

asomers:
Rust has no "ppc" target_arch.  It should be "powerpc" or "powerpc64".

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -210,7 +210,8 @@ libc_enum!{
             all(target_os = "linux", any(
                 target_arch = "aarch64",
                 target_arch = "arm",
-                target_arch = "ppc",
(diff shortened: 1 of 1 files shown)


asomers:
This should be merged_after_ #1498 or bad merge collisions will result.

rtzoeller:
@asomers is there a way we can lint for this kind of typo?

asomers:
I'm surprised that existing tools didn't already catch it.  Rustdoc clearly knows that `ppc` wasn't a valid `target_os`.  Clippy could, and probably should, check for this.  It's similar to https://github.com/rust-lang/rust-clippy/pull/5506 .  I'll open an issue over there, but I don't know Clippy well enough to implement it myself.

PR #1571, opened 2021-10-17
Remove broken newlib support

asomers:
Nix has never supported a newlib target, but there were a few cfg checks
for it in our codebase.  Some of them were misspelled (newlibc vs
newlib), and some of the checks were wrong.  Removing them makes the
code much more readable.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -51,10 +51,8 @@ impl ClockId {
     #[cfg(not(any(
         target_os = "macos",
         target_os = "ios",
-        all(
-            not(any(target_env = "uclibc", target_env = "newlibc")),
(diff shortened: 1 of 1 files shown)


src/time.rs
```diff
@@ -51,10 +51,8 @@ impl ClockId {
     #[cfg(not(any(
         target_os = "macos",
         target_os = "ios",
-        all(
-            not(any(target_env = "uclibc", target_env = "newlibc")),
```
rtzoeller, line 55: I'm fine with dropping `uclibc` here since we don't support it at the moment, but it's worth noting #1603 may need to put it back.

rtzoeller (approved):

PR #1574, opened 2021-10-18
Add 'From' implementations for 'SockAddr'

### LINKED ISSUE #1573: Support From trait for SockAddr
It would be useful if nix provided `From` implementations for `SockAddr`. For example, `impl From<InetAddr> for SockAddr`, `impl From<UnixAddr> for SockAddr`, etc.

stevenengler:
Closes #1573.


asomers:
Let's do this as part of issue #1544 instead.

PR #1581, opened 2021-10-22
Add fdatasync for missing platforms

rtzoeller:
These platforms stub out `fdatasync()` by just calling `fsync()`, but that is a conforming implementation.

DragonFly recently provided [an implementation](https://github.com/DragonFlyBSD/DragonFlyBSD/commit/74fa2560ac77f9db4a34b2a7c72450126fec4ed6), but it has not made it into a release or rust-lang/libc.


asomers:
Yep yep yep.  BTW, FreeBSD has had an fdatasync since 11.1, but it's equivalent to fsync on ZFS, which is the only file system many users care about.

asomers (approved):

asomers:
@rtzoeller time for a rebase and CHANGELOG notice?

rtzoeller:
bors retry

PR #1583, opened 2021-10-23
Remove unsafe in with_nix_path() for [u8]

rtzoeller:
Remove use of `unsafe` in the implementation of `with_nix_path()` for `[u8]`. This also comes with a nice determinism win across input sizes, and is fairly performance neutral (slightly slower for small strings, much faster for large strings).

I suspect the performance degradation in the existing implementation is related to the following note in the `CStr::from_ptr()` documentation:

> Note: This operation is intended to be a 0-cost cast but it is currently implemented with an up-front calculation of the length of the string. This is not guaranteed to always be the case.

---

Tested with `cargo 1.57.0-nightly (7fbbf4e8f 2021-10-19)`, with variations of the following benchmarking code:

```rs
#[bench]
fn bench_with_nix_path_1024(b: &mut test::Bencher) {
    let bytes = std::hint::black_box([70u8; 1024]);
    b.iter(|| {
        bytes.with_nix_path(|cstr| {
            std::hint::black_box(&cstr);
        }).unwrap();
    })
}
```

| Length |     Before Change     |      After Change     |
|--------|-----------------------|-----------------------|
|    16  |   37 ns/iter (+/- 0)  |   44 ns/iter (+/- 0)  |
|    64  |   39 ns/iter (+/- 0)  |   44 ns/iter (+/- 0)  |
|   256  |   84 ns/iter (+/- 0)  |   48 ns/iter (+/- 0)  |
|  1024  |  232 ns/iter (+/- 1)  |   50 ns/iter (+/- 1)  |
|  4095  |  796 ns/iter (+/- 8)  |   62 ns/iter (+/- 2)  |

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,9 +78,9 @@ pub mod unistd;
  *
  */
 
-use libc::{c_char, PATH_MAX};
+use libc::PATH_MAX;
 
-use std::{ptr, result};
+use std::result;
 use std::ffi::{CStr, OsStr};
 use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


asomers (approved):
Looks great!

PR #1584, opened 2021-10-23
wait: implement waitid()

NeoRaider:
waitid() has a number of additional features that waitpid() is missing:

- WNOWAIT is only accepted for waitid() on Linux (and possibly other platforms)
- Support for waiting on PID file descriptors on Linux

For now support is added for all platforms with waitid() that have proper siginfo_t support in libc. NetBSD support is currently a work in progress [1].

Tests for the signal/exit code are currently skipped on MIPS platforms due to multiple bugs in qemu-user in the translation of siginfo_t (one fixed in January [2], one currently under review [3]).

[1] rust-lang/libc#2476
[2] https://lists.nongnu.org/archive/html/qemu-devel/2021-01/msg04810.html
[3] https://lists.nongnu.org/archive/html/qemu-devel/2021-10/msg05433.html

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -6,6 +6,11 @@ use crate::Result;
 use cfg_if::cfg_if;
 use libc::{self, c_int};
 use std::convert::TryFrom;
+#[cfg(any(
+    target_os = "android",
+    all(target_os = "linux", not(target_env = "uclibc")),
+))]
+use std::os::unix::io::RawFd;
 
 libc_bitflags!(
     /// Controls the behavior of [`waitpid`].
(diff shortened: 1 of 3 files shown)


src/sys/wait.rs
```diff
@@ -260,3 +317,54 @@ pub fn waitpid<P: Into<Option<Pid>>>(pid: P, options: Option<WaitPidFlag>) -> Re
...
+    PGid(Pid),
+    /// Wait for the child referred to by the given PID file descriptor
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    PIDFd(RawFd),
+}
+
+/// Wait for a process to change status
+///
+/// See also [waitid(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html)
+#[cfg(any(
```
asomers, line 346: libc defines this function for almost all OSes.  I think it might be everything except Redox.  Any reason not to enable it for all of them in Nix?

src/sys/wait.rs
```diff
@@ -225,6 +227,61 @@ impl WaitStatus {
...
+    /// siginfo_t is actually a union, not all fields may be initialized.
+    /// The functions si_pid() and si_status() must be valid to call on
+    /// the passed siginfo_t.
+    #[cfg(any(
+        target_os = "android",
+        target_os = "freebsd",
+        target_os = "haiku",
+        target_os = "linux",
+    ))]
+    pub unsafe fn from_siginfo(siginfo: &libc::siginfo_t) -> Result<WaitStatus> {
```
asomers, line 248: Why does this need to be `pub`?

asomers (changes requested):
Don't forget a CHANGELOG entry too

src/sys/wait.rs
NeoRaider: I started with the list of platforms for which `WEXITED` is defined in `WaitPidFlag` and then removed all that had some obvious issue:
- NetBSD has broken siginfo_t in libc (see issue linked in PR description)
- Apple/IOS are missing CLD_EXITED/KILLED/DUMPED in libc (and I don't really want to open *another* libc PR for a system that I do not use and that I'm not familiar with...)

src/sys/wait.rs
NeoRaider: I thought it might be useful with sigwaitinfo or sigaction as well, but maybe it would make more sense to create wrappers for these functions as well if somebody is interested in using them. (sigaction's sa_sigaction is not represented well in libc either, as the struct contains a union in C on most platforms...)

asomers: Yeah, I don't think it's quite useful enough as is.  Please make it private.

NeoRaider: Done

rtzoeller:
@NeoRaider can you rebase this? It isn't conflicting on GitHub, but local testing shows the changelog will merge into the `0.23.1` patch's section, rather than the current one.

NeoRaider:
Rebased and updated commit message to reflect the current status of the mentioned qemu-user issue.

NeoRaider:
Hmm, seems I'll have to update the test code as well.

rtzoeller:
Another PR went in first and modified the changelog in the same spot. Can you rebase?

PR #1590, opened 2021-11-14
Disable pwritev and preadv for 'macos' (fixes #1588)

### LINKED ISSUE #1588: MacOS: Undefined symbols for architecture x86_64
= note: Undefined symbols for architecture x86_64:
            "_preadv", referenced from:
                nix::sys::uio::preadv::h8c66d9fb8b619a73 in libnix-1b7d7137231b30ff.rlib(nix-1b7d7137231b30ff.nix.1b5c8824-cgu.0.rcgu.o)
            "_pwritev", referenced from:
                nix::sys::uio::pwritev::hb39f620af1eb34cc in libnix-1b7d7137231b30ff.rlib(nix-1b7d7137231b30ff.nix.1b5c8824-cgu.0.rcgu.o)
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)

rivy:
Fixes #1588.

Please merge; we're using a patched vendored version until you're able to fix this...

### DIFF
--- a/src/sys/uio.rs
+++ b/src/sys/uio.rs
@@ -30,7 +30,7 @@ pub fn readv(fd: RawFd, iov: &mut [IoVec<&mut [u8]>]) -> Result<usize> {
 /// or an error occurs. The file offset is not changed.
 ///
 /// See also: [`writev`](fn.writev.html) and [`pwrite`](fn.pwrite.html)
-#[cfg(not(target_os = "redox"))]
+#[cfg(not(any(target_os = "macos", target_os = "redox")))]
 pub fn pwritev(fd: RawFd, iov: &[IoVec<&[u8]>],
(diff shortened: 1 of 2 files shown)


rivy:
I've added matching OS restrictions to the matching tests to help surface any future breakage.

asomers (changes requested):
Removing those functions is not the solution to whatever problem you're experiencing.  Those functions do exist on OSX, at least according to libc.  If they in fact don't, or if they need to be linked with a different library, then you need to fix this issue in libc.  We can't merge this PR until you've root-caused the issue.

rivy:
I believe that you have already "root-caused" the issue by noting that the current version fails for OSX v10.15 (see #1588).

PR #1591, opened 2021-11-16
Fix CI

asomers:
* Update Redox compiler
* Fix "unknown lint clippy::if_then_panic"

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -223,9 +223,9 @@ task:
       env:
         TARGET: x86_64-unknown-redox
         # Redox requires a nightly compiler.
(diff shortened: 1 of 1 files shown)


rtzoeller:
The dfly failure is due to a change I submitted to libc. The libc policy on breaking changes is to deprecate the item in a release prior to actually making the change (and removing the deprecation).

It's listed as a tier 3 target, so I think the correct move is for the failure to be ignored at the build level, but we could also trivially disable deprecation warnings on dfly until the change to libc is submitted.

asomers:
Yeah, since there isn't a non-deprecated alternative, I think we should silence those deprecation warnings for now.

PR #1592, opened 2021-11-16
Ensure that MSRV tests do not regress

matklad:
EDIT: this is testing PR, see https://github.com/nix-rust/nix/pull/1593 for the real thing. 

When testing with older versions of rustc, there's a CI failure mode
when our dependency, which used to be compatible with that Rust version,
publishes a new version which can't be compiled using that old Rust
anymore. That's pretty unfortunate, as that means that third parties can
break our CI without any changes to the code in this repo.

To protect against that, let's use Cargo.lock on CI, but only when
testing against older versions. For stable Rust, we continue to generate
Cargo.lock with freshest dependencies. Implementation wise, we save
known-good Cargo.lock as `Cargo.lock.msrv`, and just copy that in every
relevant CI job.

To get a working `Cargo.lock.msrv`, I run `cargo +nightly
generate-lockfile -Zminimal-versions` -- that is guaranteed to support
the oldest Rust version we can.

I've submitted the PR not against the current master, but against the commit before `bitflags` upgrade, just to check that the thing works, as I am not familiar with .cirrus.yml.


matklad:
Seems to work:

```
Linux x86_64:
Compiling bitflags v1.1.0

Rust Stable:
Compiling bitflags v1.3.2
```

matklad:
Ok, I submitted the version of this against master: https://github.com/nix-rust/nix/pull/1593. Won't close this PR immediateely so that build logs don't go away.

PR #1595, opened 2021-11-17
Include tests in the crate again

keszybz:
…so it is possible to run checks on the crate again.

Fixup for 7a2fe9099fc226cf5010b450f473ca9a0db5bafc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ authors     = ["The nix-rust Project Developers"]
 repository  = "https://github.com/nix-rust/nix"
 license     = "MIT"
 categories  = ["os::unix-apis"]
-include = ["src/**/*", "LICENSE", "README.md", "CHANGELOG.md"]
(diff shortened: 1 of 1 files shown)


asomers:
What exactly are you trying to do?

keszybz:
I'm making an update of the distro package in Fedora (https://src.fedoraproject.org/rpms/rust-nix/pull-request/4). We generally run `cargo test` as part of the build. This is especially important for a package like `nix` which is closely tied to the details of the operating system.

asomers:
And you download the crate from crates.io rather than from github?

keszybz:
Yes, we use the following URL: `https://crates.io/api/v1/crates/nix/0.23.0/download#/nix-0.23.0.crate` (defined through some macro). We want to use creates.io to be able to handle all packages uniformly.

asomers (approved):
Weird.  I had no idea that anybody used crates.io that way.

bors r+

PR #1598, opened 2021-11-23
Update doc comment to match type `Option<Signal>`

dario23:
This has previously been done for `killpg`, but not `kill` it seems.


src/sys/signal.rs
```diff
@@ -868,8 +868,8 @@ pub fn sigprocmask(how: SigmaskHow, set: Option<&SigSet>, oldset: Option<&mut Si
 ///     is sent to all processes exclusing system processes.
 ///   - If less than `-1`, the signal is sent to all processes whose
 ///     process group ID is equal to the absolute value of `pid`.
-/// * `signal` - Signal to send.  If 0, error checking if performed but no
-///              signal is actually sent.
+/// * `signal` - Signal to send. If `None`, error checking if performed
```
rtzoeller, line 871: Thanks! Can you amend your commit to fix the other typo in this line as well? "if" -> "is".

asomers (approved):

PR #1599, opened 2021-11-28
Avoid lock poisoning by using parking_lot

rtzoeller:
The synchronization primitives in the standard library are not panic-safe, however the `parking_lot` crate provides drop-in replacement primitives which are.

Add the `parking_lot` crate as a dev dependency and update the tests to use it. This should make it easier to narrow down which test(s) in a group of tests are actually failing, since the entire group will no longer fail due to poisoning.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ cc = "1"
 [dev-dependencies]
 assert-impl = "0.1"
 lazy_static = "1.2"
+parking_lot = "0.11.2"
 rand = "0.8"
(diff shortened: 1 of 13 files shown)


rtzoeller:
@asomers I'm not particularly attached to these changes, but they would have simplified some debugging I did at one point. The changes are also easy enough to recreate, should we want to rebase it onto other changes (like #1498).

asomers (approved):
Cool!  I didn't know that about parking_lot.

bors r+

PR #1603, opened 2021-12-07
uclibc support

skrap:
uclibc is a libc alternative (peer to glibc and musl) which is used in low-resource embedded linux applications.

It's supported in rust as the `target_env` of several tier 3 targets, but `nix` currently doesn't build.  This patch provides a few customizations to get uclibc building.

To test:
* Get nightly rust
* Follow directions for getting a cross toolchain and env setup here: https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/armv7-unknown-linux-uclibceabihf.md

Thanks for your consideration!

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Tier 2:
   * x86_64-unknown-netbsd
 
 Tier 3:
+  * armv7-unknown-linux-uclibceabihf
   * x86_64-fuchsia
   * x86_64-unknown-dragonfly
   * x86_64-unknown-linux-gnux32
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -11,6 +11,7 @@ use std::str::FromStr;
 #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
 use std::os::unix::io::RawFd;
 use std::ptr;
+use cfg_if::cfg_if;
 
 #[cfg(not(any(target_os = "openbsd", target_os = "redox")))]
 #[cfg(any(feature = "aio", feature = "signal"))]
@@ -420,10 +421,15 @@ pub const SIGPOLL : Signal = SIGIO;
 /// Alias for [`SIGSYS`]
 pub const SIGUNUSED : Signal = SIGSYS;
 
-#[cfg(not(target_os = "redox"))]
-type SaFlags_t = libc::c_int;
-#[cfg(target_os = "redox")]
-type SaFlags_t = libc::c_ulong;
+cfg_if! {
+    if #[cfg(target_os = "redox")] {
+        type SaFlags_t = libc::c_ulong;
+    } else if #[cfg(target_env = "uclibc")] {
+        type SaFlags_t = libc::c_ulong;
+    } else {
+        type SaFlags_t = libc::c_int;
+    }
+}
 }
 
 #[cfg(feature = "signal")]
@@ -1045,6 +1051,8 @@ mod sigevent {
                 #[cfg(target_os = "freebsd")]
                 SigevNotify::SigevThreadId{..} => libc::SIGEV_THREAD_ID,
                 #[cfg(all(target_os = "linux", target_env = "gnu", not(target_arch = "mips")))]
+                SigevNotify::SigevThreadId{..} => libc::SIGEV_THREAD_ID,
+                #[cfg(all(target_os = "linux", target_env = "uclibc"))]
                 SigevNotify::SigevThreadId{..} => libc::SIGEV_THREAD_ID,
                 #[cfg(any(all(target_os = "linux", target_env = "musl"), target_arch = "mips"))]
                 SigevNotify::SigevThreadId{..} => 4  // No SIGEV_THREAD_ID defined
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -120,10 +120,10 @@ pub enum AddressFamily {
     #[cfg(any(target_os = "android", target_os = "linux"))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
     Llc = libc::AF_LLC,
-    #[cfg(target_os = "linux")]
+    #[cfg(all(target_os = "linux", not(target_env = "uclibc")))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
     Ib = libc::AF_IB,
-    #[cfg(target_os = "linux")]
+    #[cfg(all(target_os = "linux", not(target_env = "uclibc")))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
     Mpls = libc::AF_MPLS,
     #[cfg(any(target_os = "android", target_os = "linux"))]
--- a/test/sys/mod.rs
+++ b/test/sys/mod.rs
@@ -7,7 +7,7 @@ mod test_signal;
 // cases on DragonFly.
 #[cfg(any(target_os = "freebsd",
           target_os = "ios",
-          target_os = "linux",
+          all(target_os = "linux", not(target_env = "uclibc")),
           target_os = "macos",
           target_os = "netbsd"))]
 mod test_aio;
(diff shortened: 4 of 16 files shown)


rtzoeller:
Thanks! Changes LGTM, although I didn't run the tests. A few thoughts....

- Is `armv7-unknown-linux-uclibceabihf` the primary target you are intending to support with these changes? I tried building with some of the other `uclibc` targets and had mixed success: only `armv7-unknown-linux-uclibceabihf` and `armv5te-unknown-linux-uclibceabi` build successfully, while the `mips`, `mipsel`, and `x86_64` targets fail to build.
  - If yes we might consider adding `armv7-unknown-linux-uclibceabihf` to the list of targets we build for, similar to how we support other tier 3 targets like `x86_64-unknown-dragonfly` using `-Zbuild-std`. It seems like the effort to support `uclibc` is likely low for any individual PR, but that the build could easily break if we aren't validating it.
- We'll want to squash/rebase the commits before merging. This doesn't necessarily need to happen now, as I imagine we'll hold off on submitting this until after the 0.23.1 release, and more changes could be required.

skrap:
Thanks!
> Is armv7-unknown-linux-uclibceabihf the primary target you are intending to support with these changes? 
Yes, and I am doing so with corporate backing for a long-term product need.  I would be glad to have this build running in CI to catch any errors early.
> We'll want to squash/rebase the commits before merging.
Yes, that is fine.  I can do that once the main branch is ready for these changes.  Want to let me know with a comment here? Or some other way?

Thanks again!

src/sys/termios.rs
```diff
@@ -776,7 +776,7 @@ libc_bitflags! {
         ALTWERASE;
         IEXTEN;
         #[cfg(not(target_os = "redox"))]
-        EXTPROC;
+        EXTPROC as tcflag_t; // uclibc gets the type wrong
```
asomers, line 779: Is this due to a bug in libc?

asomers (commented):
I thought uclibc was used with RTOSes and microcontrollers, not Linux.  Is it used with both?

rtzoeller:
> I thought uclibc was used with RTOSes and microcontrollers, not Linux. Is it used with both?

I believe it is used with embedded devices which run Linux. https://gogs.waldemar-brodkorb.de/oss/uclibc-ng

rust-lang/libc has it alongside glibc and musl: https://github.com/rust-lang/libc/tree/master/src/unix/linux_like/linux

src/sys/termios.rs
skrap: Yes, it seems that `libc`'s `uclibc` module doesn't have the right type for this item.  We could accept this change here, pending a fix in `libc`, or I could try to get a `libc` change through first.  Preferences?

asomers: Let's fix it in libc first.

rtzoeller: https://github.com/rust-lang/libc/pull/2592

skrap: Just checking in here- I'm pushing through the changes in `libc` but it's a bit slowed by some larger sibling changes.  I am still definitely pursuing this PR :)

lancethepants:
Also looking forward to uclibc support as I am trying to bring up `armv7-unknown-linux-uclibceabi` as new tier3 target.
https://github.com/rust-lang/rust/pull/92383
Tested this patch and it enabled me to compile my application. Not sure if I tested before or after https://github.com/nix-rust/nix/pull/1571, but hopefully @skrap will keep on it.

rtzoeller:
@skrap with https://github.com/rust-lang/libc/pull/2615 being submitted, I believe this can likely be rebased and squashed.

Here's an example of how to build tier 3 targets in nix's CI: https://github.com/nix-rust/nix/blob/18d1f62b150177ca4c40e2cbe21a98ddba80ecf7/.cirrus.yml#L280-L282

skrap:
For now, I'm expecting CI to fail until:
* `libc` cuts a new version to depend upon
* I can track down the last few failures, which may be related to qemu user-space emulation inconsistencies

skrap:
OK, this PR should be good at this point!  (Tempting fate by saying this before CI has completed...)

skrap:
The CI failures seem independent of these changes.  Perhaps it's a new clippy lint?

rtzoeller:
> The CI failures seem independent of these changes. Perhaps it's a new clippy lint?

Was just responding about that. I'll investigate and fix it independently of this PR, and will try to look at this in the next day or two.

README.md
```diff
@@ -83,6 +83,7 @@ Tier 3:
   * x86_64-unknown-linux-gnux32
   * x86_64-unknown-openbsd
   * x86_64-unknown-redox
+  * armv7-unknown-linux-uclibceabihf
```
rtzoeller, line 86: Let's put this at the top of this section, to keep them sorted alphabetically.

src/sys/signal.rs
```diff
@@ -1046,6 +1052,8 @@ mod sigevent {
                 SigevNotify::SigevThreadId{..} => libc::SIGEV_THREAD_ID,
                 #[cfg(all(target_os = "linux", target_env = "gnu", not(target_arch = "mips")))]
                 SigevNotify::SigevThreadId{..} => libc::SIGEV_THREAD_ID,
+                #[cfg(all(target_os = "linux", target_env = "uclibc"))]
```
rtzoeller, line 1055: All of the references to `SIGEV_THREAD_ID` on Linux can probably be consolidated, but it's outside the scope of this PR. I see `SIGEV_THREAD_ID` defined by musl, but not in rust-lang/libc's redefinitions, so it'd need to be updated there first.

Not necessary to do it here, just calling it out.

src/sys/socket/addr.rs
```diff
@@ -120,10 +120,10 @@ pub enum AddressFamily {
     #[cfg(any(target_os = "android", target_os = "linux"))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
     Llc = libc::AF_LLC,
-    #[cfg(target_os = "linux")]
+    #[cfg(all(target_os = "linux",not(target_env = "uclibc")))]
```
rtzoeller, line 123: Nit: be consistent with spaces following commas, at least for new code.

test/test_fcntl.rs
```diff
@@ -30,6 +30,9 @@ use std::os::unix::fs;
 
 #[test]
 #[cfg(not(target_os = "redox"))]
+// QEMU does not handle openat well enough to satisfy this test
```
rtzoeller, line 33: What's the status of running the tests locally for you?

skrap: The tests all pass for me under qemu-arm, having ignored `test_openat` and the two OFD flock tests.  (The OFD flock fcntls are not supported on uclibc.)

I have run a subset of the tests on a physical device, but I don't have a device which runs a new enough uclibc & kernel to run every test.  Those which can run have passed.

skrap: Oh, and the `test_openat` test does succeed on actual hardware.  :+1:

src/sys/socket/addr.rs
skrap: Fixed.

src/sys/signal.rs
skrap: :+1:

README.md
skrap: Fixed.

test/sys/mod.rs
```diff
@@ -28,6 +28,7 @@ mod test_termios;
 #[cfg(not(any(target_os = "redox", target_os = "fuchsia")))]
 mod test_ioctl;
 mod test_wait;
+#[cfg(not(target_env = "uclibc"))]
```
asomers, line 31: Why disable this entire module?

test/test_fcntl.rs
asomers: Wow, that's some surprisingly basic functionality to be broken in user-mode emulation.

skrap:
Thanks for the review and the merge!

PR #1610, opened 2021-12-19
Re-enable deprecation warnings on DragonFly

rtzoeller:
Reverts 0df82ded4704caf78c32df941c170cdacc9688dc and ac159d9c4e4165d74af25dc8b4913e6b7efe7272.

### DIFF
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -38,12 +38,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cc"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7db2f146208d7e0fbee761b09cd65a7f51ccc38705d4e7262dad4d73b12a76b1"
-
-[[package]]
 name = "cfg-if"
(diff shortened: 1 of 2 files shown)


Cargo.lock.msrv
```diff
@@ -104,9 +104,9 @@ checksum = "a374c89b9db55895453a74c1e38861d9deec0b01b405a82516e9d5de4820dea1"
 
 [[package]]
 name = "libc"
-version = "0.2.102"
+version = "0.2.112"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2a5ac8f984bfcf3a823267e5fde638acc3325f6496633a5da6bb6eb2171e103"
```
rtzoeller, line 109: @asomers I ran `cargo update -Zminimal-versions` and copied the generated Cargo.lock file over this one. Is that the expected workflow?

asomers (changes requested):
Now that DragonflyBSD doesn't have any more special settings, you should be able to combine it with OpenBSD's task again.

asomers (approved):
bors r+

PR #1611, opened 2021-12-21
feature-gate most Nix functions

asomers:
Using features reduces build time and size for consumer crates.  By
default all features are enabled.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -5,13 +5,15 @@ use std::ffi::OsString;
 use std::os::raw;
 use std::os::unix::ffi::OsStringExt;
 use std::os::unix::io::RawFd;
-use crate::sys::stat::Mode;
-use crate::{NixPath, Result};
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 use std::ptr; // For splice and copy_file_range
-#[cfg(any(target_os = "android", target_os = "linux"))]
-use crate::sys::uio::IoVec; // For vmsplice
+#[cfg(feature = "fs")]
+use crate::{
+    NixPath,
+    Result,
+    sys::stat::Mode
+};
 
 #[cfg(any(
     target_os = "linux",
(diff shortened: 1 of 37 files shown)


asomers:
Supersedes #1498 .  I'm creating a new PR rather than update the previous one because I didn't want to erase some of its history by force-pushing.

src/fcntl.rs
```diff
@@ -99,30 +113,37 @@ libc_bitflags!(
...
                   target_os = "redox"))]
+        #[cfg_attr(docsrs, doc(cfg(all())))]
         O_FSYNC;
         /// Allow files whose sizes can't be represented in an `off_t` to be opened.
         #[cfg(any(target_os = "android", target_os = "linux"))]
+        #[cfg_attr(docsrs, doc(cfg(all())))]
         O_LARGEFILE;
         /// Do not update the file last access time during `read(2)`s.
         #[cfg(any(target_os = "android", target_os = "linux"))]
+        #[cfg_attr(docsrs, doc(cfg(all())))]
```
SteveLauC, line 124: Hi @asomers, looking at our source code, I am curious why this `#[cfg_attr(docsrs, doc(cfg(all())))]` is needed?

From my obervation:

1. We are not enabling the [`doc_auto_cfg`](https://doc.rust-lang.org/rustdoc/unstable-features.html#doc_auto_cfg-automatically-generate-doccfg) feature, which means we won't get a `#[doc(cfg(xx))]` with `#[cfg(xxx)]`, so no need to overwrite a non-existing value
2. The `doc_cfg` feature does not seem to work with assocaiated item:

    ```rs
        #![feature(doc_cfg)]

    #[cfg(target_os = "linux")]
    #[doc(cfg(target_os = "linux"))]
    pub struct Foo;

    impl Foo {
        #[cfg(target_os = "linux")]
        #[doc(cfg(target_os = "linux"))]
        pub const LINUX_SPECIFIC_CONST: usize = 1;
    }
   ```
  
   Run:

   ```sh
   $ cargo +nightly doc --target x86_64-unknown-linux-gnu --open
   ```

   That `Available on Linux only.` banner will be set for `Foo` only, not for `LINUX_SPECIFIC_CONST`

And some types in our code are partially annocated with this attribute, e.g., `enum Request` defined in `src/sys/ptrace/linux.rs`, which makes me confused

asomers: I think they're there because I originally developed the PR with the `doc_auto_cfg` feature (see https://github.com/nix-rust/nix/pull/1498/commits/83ea2f34f1010762fc6fd8783d04191623e6abaf).  When that's on, constants like `O_NOCTTY` get a "This is supported on non-Redox only" banner.  Those banners are both annoyingly verbose and deceptive (see the linked commit message for the explanation for why they're deceptive).  But then I must've removed `doc_auto_cfg` before I posted the PR, leaving the `doc(cfg(all()))` attributes useless.  So we can remove them now.

PR #1615, opened 2021-12-24
Add sendfile(2) for DragonFly

rtzoeller:
The code is copied from the Mac OS and FreeBSD implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1564](https://github.com/nix-rust/nix/pull/1564))
 - Added POSIX per-process timer support
   (#[1622](https://github.com/nix-rust/nix/pull/1622))
+- Added `sendfile` on DragonFly.
+  (#[1615](https://github.com/nix-rust/nix/pull/1615))
 
 ### Changed
 ### Fixed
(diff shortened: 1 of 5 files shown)


asomers (commented):
The signatures in libc look the same.  Is the only difference that none of the flags are supported on Dragonfly?

rtzoeller:
The primary difference is the flags field (which FreeBSD actually packs flags and an int into). From the DragonFly man page:

> The flags argument is currently undefined and should be specified as 0.

There are also some documentation differences, notably that FreeBSD supports reading from a shared memory object, while the DragonFly man page indicates it must be a regular file.

asomers:
Just for consistency's sake, do you think it would make sense to define the API identically on FreeBSD and Dragonfly, but require that the flags field be zero on Dragonfly?  That would make it easier to port programs from FreeBSD to Dragonfly.  OTOH, it's not very hard with the current API either.  I'm kind of on the fence about it.

rtzoeller:
I considered it, but my reservation is DragonFly doesn't currently assign meaning to the flags parameter other than requiring that it must be zero. It could be repurposed to something other than a flags bitfield while maintaining ABI compatibility. Consolidating them would also require passing zero for the `readahead` parameter, which doesn't currently apply to DragonFly (and may never).

Given other targets like Linux have different signatures already, I think keeping it separate is _probably_ (?) the best idea for now, while acknowledging they may be consolidated in the future.

We can defer merging this until after some of the other PRs are submitted, to avoid them needing to rebased yet again for changelog reasons. I'd rather rebase than make someone else need to, and it gives us time to change our minds on the implementation.

asomers (approved):
Sounds good.

PR #1616, opened 2021-12-24
Disable mknod and mknodat tests on DragonFly

rtzoeller:
Like FreeBSD, DragonFly does not support creating regular files (i.e. `S_IFREG`) with `mknod` or `mknodat`. These tests should be disabled as they always fail.

There are still a few other failing tests on DragonFly, but it is unclear to me if they are due to faulty/inapplicable test logic or a bug in DragonFly.


asomers (approved):
bors r+

PR #1618, opened 2021-12-26
Refactor UnixAddr

asomers:
* Within UnixAddr, replace the path_len variable (length of the sun_path
    field) with sun_len (length of the whole structure).  This is more
    similar to how other sockaddr types work, and it's the same way that the
    BSDs use the sun_len field.
    
    Also, don't require that sun_path be nul-terminated. The OS doesn't
    require it.

* On BSD-derived operating systems, struct sockaddr has a sa_len field
    that holds the length of the structure.  UnixAddr's path_len field is
    redundant.  Remove path_len on BSD-derived OSes, retaining it only for
    Illumos and Linux-based OSes.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -1,4 +1,5 @@
 use super::sa_family_t;
+use cfg_if::cfg_if;
 use crate::{Result, NixPath};
 use crate::errno::Errno;
 use memoffset::offset_of;
@@ -580,7 +581,13 @@ pub struct UnixAddr {
     sun: libc::sockaddr_un,
     /// The length of the valid part of `sun`, including the sun_family field
     /// but excluding any trailing nul.
-    sun_len: u8,
+    // On the BSDs, this field is built into sun
+    #[cfg(any(target_os = "android",
+              target_os = "fuchsia",
+              target_os = "illumos",
+              target_os = "linux"
+    ))]
+    sun_len: u8
 }
 
 // linux man page unix(7) says there are 3 kinds of unix socket:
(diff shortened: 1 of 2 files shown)


asomers:
cc @coolreader18

coolreader18:
Oh this looks good! I think it's kinda what I was going for (though it's a little hard to remember at this point) but obviously you've looked more into the issue

asomers:
So the test failures on Linux are due to a Linux kernel bug.  `getsockname` returns a length that includes the path's trailing NUL, even if the length passed to `bind` did not include that trailing NUL, and EVEN if the path passed to `bind` was not nul-terminated.  I'm not sure if I should adjust the getsockname test to account for that, or if I should explicitly trim the trailing NUL in UnixAddr::from_raw_parts.

The Illumos failure is my fault and easy to fix.

src/sys/socket/addr.rs
```diff
@@ -605,8 +617,19 @@ impl<'a> UnixAddrKind<'a> {
                 slice::from_raw_parts(sun.sun_path.as_ptr().add(1) as *const u8, path_len - 1);
             return Self::Abstract(name);
         }
-        let pathname = slice::from_raw_parts(sun.sun_path.as_ptr() as *const u8, path_len - 1);
-        Self::Pathname(Path::new(OsStr::from_bytes(pathname)))
+        let pathname = slice::from_raw_parts(sun.sun_path.as_ptr() as *const u8, path_len);
+        if pathname.last() == Some(&0) {
+            // A trailing NUL is not considered part of the path, and it does
```
rtzoeller, line 622: Nit: you're using both `NUL` and `nul` for `b'\0'`. Pick one.

src/sys/socket/mod.rs
```diff
@@ -1934,10 +1934,10 @@ pub fn sockaddr_storage_to_addr(
             Ok(SockAddr::Inet(InetAddr::V6(sin6)))
         }
         libc::AF_UNIX => {
-            let pathlen = len - offset_of!(sockaddr_un, sun_path);
             unsafe {
                 let sun = *(addr as *const _ as *const sockaddr_un);
-                Ok(SockAddr::Unix(UnixAddr::from_raw_parts(sun, pathlen)))
+                let sun_len = len.try_into().unwrap();
```
rtzoeller, line 1939: Unwrapping here makes me a little concerned, but I'm not familiar with the previous behavior. Should we return an `Err` instead of panicking?

rtzoeller (approved):
Looks good to me, just a few minor nitpicks/questions. 

Tests are passing on DragonFly.

rtzoeller:
@asomers GitHub says you live in Longmont, so stay safe re: wildfires!

src/sys/socket/mod.rs
asomers: The unwrap is guaranteed to pass, because of the assertion on line 1914.

asomers:
> @asomers GitHub says you live in Longmont, so stay safe re: wildfires!

Fortunately the big one is downwind of me.  There's a closer one upwind, but it's controlled now, and didn't do much damage.  And oh yeah, it's SNOWING NOW!!!

bors r+

PR #1619, opened 2021-12-26
Added getters for the MqAttr struct

fpagliughi:
With the existing code, if you call `mq_getattr()`, there does not appear to be a way to get any of the attributes from the returned `MqAttr` struct, other than the flags. This adds getter functions to retrieve the size parameters of the queue, and the current number of messages in the queue.

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -64,6 +64,21 @@ impl MqAttr {
     pub const fn flags(&self) -> mq_attr_member_t {
         self.mq_attr.mq_flags
     }
+
+    /// The max number of messages that can be held by the queue
+    pub const fn maxmsg(&self) -> mq_attr_member_t {
+        self.mq_attr.mq_maxmsg
+    }
+
+    /// The maximum size of each message (in bytes)
+    pub const fn msgsize(&self) -> mq_attr_member_t {
+        self.mq_attr.mq_msgsize
+    }
+
+    /// The number of messages currently held in the queue
+    pub const fn curmsgs(&self) -> mq_attr_member_t {
+        self.mq_attr.mq_curmsgs
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/mqueue.rs
```diff
@@ -64,6 +64,21 @@ impl MqAttr {
...
+
+    /// The maximum size of each message (in bytes)
+    pub const fn msg_size(&self) -> mq_attr_member_t {
+        self.mq_attr.mq_msgsize
+    }
+
+    /// The number of messages currently held in the queue
+    pub const fn current_msg(&self) -> mq_attr_member_t {
+        self.mq_attr.mq_curmsgs
+    }
```
rtzoeller, line 81: I'm not sure how I feel about the naming here. `cur` is expanded to `current`, but we don't expand `msg`, and the underscores aren't present in the `libc` definitions.

For consistency with the rest of nix (example below), I'd lean towards using the same names as the underlying fields, minus the `mq_` prefix. @asomers thoughts?

https://github.com/nix-rust/nix/blob/a392647f450f88bf4a162135712ea023e8da3b1a/src/sys/socket/addr.rs#L1270-L1273

rtzoeller (changes requested):
Thanks for the PR! Can you update the changelog to include these changes?

I actually ran into the absence of these fields while looking into a failing test on DragonFly (and presumably NetBSD), so this change will also help clean up some tests in the future!

src/mqueue.rs
fpagliughi: I'm good with that. Seems more consistent. I can fix it.

asomers:
LGTM.  Just squash your commits and I'll merge it.

rtzoeller (approved):

PR #1622, opened 2021-12-28
Introduce `timer_*` support 

### LINKED ISSUE #1424: Add timer_* bindings
Feature request for adding bindings for the following POSIX timer functions to the crate:
* [ ] timer_create
* [ ] timer_delete
* [ ] timer_settime
* [ ] timer_gettime
* [ ] timer_getoverrun

See e.g. https://www.man7.org/linux/man-pages/man2/timer_create.2.html

blt:
This commit adds support for the signal timer mechanism in POSIX, the mirror to timerfd on Linux. I wasn't _quite_ sure of how to fit into the project organization but hopefully this patch isn't too far off.

Resolves #1424

Signed-off-by: Brian L. Troutwine <brian@troutwine.us>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1619](https://github.com/nix-rust/nix/pull/1619))
 - Added the `TxTime` sockopt and control message.
   (#[1564](https://github.com/nix-rust/nix/pull/1564))
+- Added POSIX per-process timer support
+  (#[1622](https://github.com/nix-rust/nix/pull/1622))
 
 ### Changed
 ### Fixed
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -201,3 +201,18 @@ feature! {
     #[allow(missing_docs)]
     pub mod timerfd;
 }
+
+#[cfg(all(
+    any(
+        target_os = "freebsd",
+        target_os = "illumos",
+        target_os = "linux",
+        target_os = "netbsd"
+    ),
+    feature = "time",
+    feature = "signal"
+))]
+feature! {
+    #![feature = "time"]
+    pub mod timer;
+}
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -1085,6 +1085,11 @@ mod sigevent {
         pub fn sigevent(&self) -> libc::sigevent {
             self.sigevent
         }
+
+        /// Returns a mutable pointer to the `sigevent` wrapped by `self`
+        pub fn as_mut_ptr(&mut self) -> *mut libc::sigevent {
+            &mut self.sigevent
+        }
     }
 
     impl<'a> From<&'a libc::sigevent> for SigEvent {
--- a/src/sys/time.rs
+++ b/src/sys/time.rs
@@ -5,6 +5,129 @@ use libc::{timespec, timeval};
 #[cfg_attr(target_env = "musl", allow(deprecated))] // https://github.com/rust-lang/libc/issues/1848
 pub use libc::{time_t, suseconds_t};
 
+#[cfg(any(
+    all(feature = "time", any(target_os = "android", target_os = "linux")),
+    all(
+        any(
+            target_os = "freebsd",
+            target_os = "illumos",
+            target_os = "linux",
+            target_os = "netbsd"
+        ),
+        feature = "time",
+        feature = "signal"
+    )
+))]
+pub(crate) mod timer {
+    use crate::sys::time::TimeSpec;
+    use bitflags::bitflags;
+
+    #[derive(Debug, Clone, Copy)]
+    pub(crate) struct TimerSpec(libc::itimerspec);
+
+    impl TimerSpec {
+        pub const fn none() -> Self {
+            Self(libc::itimerspec {
+                it_interval: libc::timespec {
+                    tv_sec: 0,
+                    tv_nsec: 0,
+                },
+                it_value: libc::timespec {
+                    tv_sec: 0,
+                    tv_nsec: 0,
+                },
+            })
+        }
+    }
+
+    impl AsMut<libc::itimerspec> for TimerSpec {
+        fn as_mut(&mut self) -> &mut libc::itimerspec {
+            &mut self.0
+        }
+    }
+
+    impl AsRef<libc::itimerspec> for TimerSpec {
+        fn as_ref(&self) -> &libc::itimerspec {
+            &self.0
+        }
+    }
+
+    impl From<Expiration> for TimerSpec {
+        fn from(expiration: Expiration) -> TimerSpec {
+            match expiration {
+                Expiration::OneShot(t) => TimerSpec(libc::itimerspec {
+                    it_interval: libc::timespec {
+                        tv_sec: 0,
+                        tv_nsec: 0,
+                    },
+                    it_value: *t.as_ref(),
+                }),
+                Expiration::IntervalDelayed(start, interval) => TimerSpec(libc::itimerspec {
+                    it_interval: *interval.as_ref(),
+                    it_value: *start.as_ref(),
+                }),
+                Expiration::Interval(t) => TimerSpec(libc::itimerspec {
+                    it_interval: *t.as_ref(),
+                    it_value: *t.as_ref(),
+                }),
+            }
+        }
+    }
+
+    /// An enumeration allowing the definition of the expiration time of an alarm,
+    /// recurring or not.
+    #[derive(Debug, Clone, Copy, PartialEq)]
+    pub enum Expiration {
+        /// Alarm will trigger once after the time given in `TimeSpec`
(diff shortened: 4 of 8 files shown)


src/sys/timer.rs
```diff
@@ -0,0 +1,116 @@
...
+            } else {
+                Some(timerspec.into())
+            }
+        })
+    }
+
+    /// Return the number of timers that have overrun
+    ///
+    /// An overrun timer is one which expires while its related signal is still
+    /// queued. TODO explain better
```
asomers, line 101: Let's not forget to fix this TODO before merging.

asomers (changes requested):
It's a good start, especially the documentation.  In addition to the inline comments, could you please add some tests and an example?

blt:
> It's a good start, especially the documentation. In addition to the inline comments, could you please add some tests and an example?

I believe this is ready for re-review. I've expanded OS coverage to include everything libc covers and added a `no_run` example to the timer module. With regard to tests, I'm not totally sure what's approachable with unit tests in the things I've added here, but if you have anything you'd like to see covered I'd be glad to add it.

src/sys/signal.rs
```diff
@@ -3,21 +3,21 @@
 
 //! Operating system signals.
 
-use crate::{Error, Result};
```
asomers, line 6: It looks like you ran rustfmt on this file.  Were there any content changes?  If not, please revert it.  If so, revert the formatting changes so we can see the content changes.

Cargo.toml
```diff
@@ -70,7 +70,7 @@ sched = ["process"]
 signal = ["process"]
 socket = []
 term = []
-time = []
+time = ["signal"]
```
asomers, line 73: Rather than make all of time depend on signal, I suggest you just conditionalize the timer module on `#[cfg(all(feature = "time", feature = "signal"))]`.

src/sys/timer.rs
```diff
@@ -0,0 +1,171 @@
...
+pub struct Timer(libc::timer_t);
+
+impl Timer {
+    /// Creates a new timer based on the clock defined by `clockid`. The details
+    /// of the signal and its handler are defined by the passed `sigevent`.
+    pub fn new(clockid: ClockId, mut sigevent: SigEvent) -> Result<Self> {
+        // On Linux and FreeBSD timer_t is a *mut c_void. On Illumos and NetBSD
+        // it is a c_int.
+        #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+        let mut timer_id: libc::timer_t = mem::MaybeUninit::uninit().as_mut_ptr();
```
asomers, line 75: By dropping the return value of `mem::MaybeUninit::uninit()` the resulting pointer will be dangling.  You need to save that value.  Also, was there really no way to make this work using type inference so as not to need the `#[cfg]` stuff?

asomers (changes requested):

src/sys/signal.rs
blt: Oops, yes, not intended.

src/sys/timer.rs
blt: > By dropping the return value of `mem::MaybeUninit::uninit()` the resulting pointer will be dangling. You need to save that value. 

Ah yes, indeed. 

> Also, was there really no way to make this work using type inference so as not to need the `#[cfg]` stuff?

Not that I could find, no. I'd be happy lose all the `cfg` mess, perhaps addressing the first part this comment I'll think of something.

blt:
I believe this is ready again. I did manage to tidy up the cfg mess in `Timer::new`, happily.

src/sys/mod.rs
```diff
@@ -201,3 +201,13 @@ feature! {
...
     pub mod timerfd;
 }
+
+#[cfg(all(any(target_os = "freebsd",
+              target_os = "illumos",
+              target_os = "linux",
+              target_os = "netbsd"),
+          feature = "time", feature = "signal"))]
+feature! {
+    #![feature = "time"]
```
asomers, line 211:
```suggestion
    #![all(feature = "signal", feature = "time")]
```

src/sys/time.rs
```diff
@@ -5,6 +5,130 @@ use libc::{timespec, timeval};
 #[cfg_attr(target_env = "musl", allow(deprecated))] // https://github.com/rust-lang/libc/issues/1848
 pub use libc::{time_t, suseconds_t};
 
+#[cfg(any(
+    all(feature = "time", any(target_os = "android", target_os = "linux")),
+    all(
+        any(
+            target_os = "android",
```
asomers, line 12: Should Android really be here, even though it's also on line 9?

src/sys/time.rs
```diff
@@ -5,6 +5,130 @@ use libc::{timespec, timeval};
...
+            const TFD_TIMER_ABSTIME = libc::TFD_TIMER_ABSTIME;
+        }
+    }
+    #[cfg(any(target_os = "freebsd", target_os = "netbsd", target_os = "dragonfly", target_os = "illumos"))]
+    bitflags! {
+        /// Flags that are used for arming the timer.
+        pub struct TimerSetTimeFlags: libc::c_int {
+            const TFD_TIMER_ABSTIME = libc::TIMER_ABSTIME;
+        }
+    }
```
asomers, line 104:
```suggestion
    bitflags! {
        /// Flags that are used for arming the timer.
        pub struct TimerSetTimeFlags: libc::c_int {
            #[cfg(any(target_os = "android", target_os = "linux"))]
            const TFD_TIMER_ABSTIME = libc::TFD_TIMER_ABSTIME;
            #[cfg(any(target_os = "freebsd", target_os = "netbsd", target_os = "dragonfly", target_os = "illumos"))]
            const TFD_TIMER_ABSTIME = libc::TIMER_ABSTIME;
        }
    }
```

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1547](https://github.com/nix-rust/nix/pull/1547))
 - Added getter methods to `MqAttr` struct
   (#[1619](https://github.com/nix-rust/nix/pull/1619))
+- Added `timer` support
```
asomers, line 28: "timer" is a bit vague.  "timer_create" would be more descriptive, or even "POSIX per-process timers".

asomers (changes requested):
Did you ever try to write a test?  Something like `test_alarm` in test/test_unistd.rs might be a good template to follow.

blt:
> Did you ever try to write a test? Something like `test_alarm` in test/test_unistd.rs might be a good template to follow.

I didn't, no, as mentioned [here](https://github.com/nix-rust/nix/pull/1622#issuecomment-1002288142) I was open to suggestions considering the low surface area introduced in this patch. I'll look at what's in `test/test_unistd.rs`, thanks.

src/sys/time.rs
blt: I think there's probably a handful of different ways to write this logic. Is there an alternate form you would prefer to see?

asomers: Why is it that the signal feature is only required on BSD and illumos, anyway?

src/sys/time.rs
blt: Introducing this change caused a build failure, see [here](https://github.com/nix-rust/nix/pull/1622/checks?check_run_id=4663863629), to explain why I'm backing out the suggestion.

src/sys/time.rs
blt: Right, yeah. So the `timer` submodule here is required by both the `timer` and `timerfd` mechanism. The `timerfd` bit of the cfg here is what you see on line 9, the `timer` bit starts on 10. Android could actually be removed from the part that starts on 10. The need for signal is a consequence of something hack turned up. If we flag this whole common module on by OS if signal is flagged off then `TimerSpec::none` is unused on the BSDs.

blt:
Aside from this [outstanding thread](https://github.com/nix-rust/nix/pull/1622#discussion_r776546963) I've incorporated feedback and think this is ready for review again. Of note I've added a `tests/test_timer` that is patterned on the example program in the timer docs and `test_alarm` from `test_unistd`.

src/sys/time.rs
asomers: Ok.  Weird, but I get it.

test/test_timer.rs
```diff
@@ -0,0 +1,92 @@
...
+    let starttime = Instant::now();
+    loop {
+        thread::sleep(Duration::from_millis(500));
+        if ALARM_CALLED.load(Ordering::Acquire) {
+            break;
+        }
+        if starttime.elapsed() > Duration::from_secs(3) {
+            panic!("Timeout waiting for SIGALRM");
+        }
+        yield_now()
```
asomers, line 83: You don't need yield_now.  The loop is about to go to sleep, which yields anyway.

blt: Ah yeah, left from when I didn't have the sleep. Removed.

asomers (approved):
bors r+

PR #1624, opened 2022-01-02
Fix mq tests on NetBSD and DragonFly

rtzoeller:
NetBSD (and DragonFly, which borrows its implementation) include additional flags beyond O_NONBLOCK in MqAttr, such as the flags passed to mq_open(). Modify the mq tests to validate _at least_ the expected flags are set, but don't require strict equality on these platforms.

Verified these tests pass on DragonFly and NetBSD locally.

### DIFF
--- a/test/test_mq.rs
+++ b/test/test_mq.rs
@@ -1,3 +1,4 @@
+use cfg_if::cfg_if;
 use std::ffi::CString;
 use std::str;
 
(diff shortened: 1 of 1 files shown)


test/test_mq.rs
```diff
@@ -79,28 +97,33 @@ fn test_mq_setattr() {
...
 
-    let new_attr_get = mq_getattr(mqd).unwrap();
-    // The following tests make sense. No changes here because according to the Linux man page only
+    // No changes here because according to the Linux man page only
     // O_NONBLOCK can be set (see tests below)
-    assert_ne!(new_attr_get, new_attr);
+    #[cfg(not(any(target_os = "dragonfly", target_os = "netbsd")))]
+    {
+        let new_attr_get = mq_getattr(mqd).unwrap();
+        assert_ne!(new_attr_get, new_attr);
```
rtzoeller, line 109: Presumably we could enable this assert for NetBSD and DragonFly as well, but writing the negation of `assert_attr_eq` is just painful enough (and IMO low-value enough) that I'd prefer not to do it. :)

rtzoeller:
Once https://github.com/rust-lang/libc/pull/2610 is submitted (and it makes it into a libc release used by nix), the enabled tests should be 100% passing on DragonFly. 🎉

asomers (approved):

rtzoeller:
bors retry

rtzoeller:
```
Caused by:
  process didn't exit successfully: `/tmp/cirrus-ci-build/target/x86_64-unknown-freebsd/debug/deps/test-9b88884763bbe54f` (signal: 14, SIGALRM: alarm clock)
```

@asomers any ideas on this FreeBSD failure?

rtzoeller:
AFAICT the failure is due to some sort of race condition between the newly minted `test_timer::alarm_fires` and the existing alarm tests. It also reproduces intermittently on NetBSD locally when running the tests in the same test run.

asomers:
I can't reproduce the failure locally, possibly because my dev system is too fast.  The CI systems are slower, and have slower storage.  But I can guess what the problem is: the `alarm_fires` test disables the signal handler _before_ dropping the `Timer`, and `Timer::drop` is what disables the timer in the OS.  I'll create a PR.

rtzoeller:
bors retry

PR #1625, opened 2022-01-02
Correct MAP_FIXED documentation for NetBSD

rtzoeller:
The previous documentation described the default behavior, rather than the behavior when the flag was set.

Also fix a test which is failing due to passing this flag erroneously.


rtzoeller:
@asomers are documentation-only corrections like this something you'd expect to see in the changelog?

asomers:
No, I don't think that doc fixes are worth a CHANGELOG entry.

asomers (approved):
LGTM, once the FreeBSD test failure is fixed.

PR #1626, opened 2022-01-02
Fix intermittency in test_timer::alarm_fires

asomers:
The test was disabling the signal handler before disabling the timer.
Fix intermittent failures but reversing the cleanup order.

Also, speed up the timer to make the test suite complete faster.

### DIFF
--- a/test/test_timer.rs
+++ b/test/test_timer.rs
@@ -23,6 +23,7 @@ fn alarm_fires() {
     // Avoid interfering with other signal using tests by taking a mutex shared
     // among other tests in this crate.
     let _m = crate::SIGNAL_MTX.lock();
+    const TIMER_PERIOD: Duration = Duration::from_millis(100);
 
     //
     // Setup
@@ -44,7 +45,7 @@ fn alarm_fires() {
         si_value: 0,
     });
     let mut timer = Timer::new(clockid, sigevent).expect("failed to create timer");
-    let expiration = Expiration::Interval(Duration::from_millis(250).into());
+    let expiration = Expiration::Interval(TIMER_PERIOD.into());
     let flags = TimerSetTimeFlags::empty();
     timer.set(expiration, flags).expect("could not set timer");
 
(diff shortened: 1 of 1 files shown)


asomers:
cc @blt

test/test_timer.rs
```diff
@@ -82,9 +82,12 @@ fn alarm_fires() {
...
 
-    // Replace the old signal handler now that we've completed the test. If the
-    // test fails this process panics, so the fact we might not get here is
-    // okay.
+    // Cleanup:
+    // 1) deregister the OS's timer.
+    // 2) Replace the old signal handler now that we've completed the test. If
+    //    the test fails this process panics, so the fact we might not get here
+    //    is okay.
+    drop(timer);
```
rtzoeller, line 90: This doesn't seem to address the issue on my NetBSD VM. I'm not entirely sure why, but I think it might be related to the following section of the timer_delete man page:

> Pending notification events (signals) may or may not be delivered.

rtzoeller: Adding a 10ms delay after dropping but before removing the signal handler seems to provide enough time for things to resolve on my VM.

asomers: Interesting.  From the FreeBSD man page: "Pending signals for the deleted timer are cleared." and from OpenGroup: "The disposition of pending signals for the deleted timer is unspecified."  So I guess OpenGroup is the lowest common denominator.  How do we ensure that all pending signals are dealt with?

rtzoeller: I tried `thread::yield_now()` without success, so I think you're current timing solution is as good as we'll get for now.

I'm going to leave the tests running for a bit, and if they don't fail with the current timeout for a few minutes I'll merge this.

rtzoeller (approved):

test/test_timer.rs
blt: Oh, hmm. I suppose this implementation could destroy the timer, replace the old signal handler and then call `sigpending`. If the result of `sigpending` turns up empty the previously deleted timer didn't leave any trailing signals behind on systems that don't clean up. I'm not totally convinced that the OpenGroup wording doesn't allow for some unexpected ordering of operations in that proposed implementation, that said. The implementation that was merged up here makes good sense to me. 

Apologies for not noticing this myself.

asomers: I don't think `sigpending` will work, because of this line from the man page:
```
Signals may be pending because they are currently
masked, or transiently before delivery (although the latter case is not
normally detectable).
```

blt: Ah ha, I believe you are correct.

PR #1630, opened 2022-01-11
Change port used by test_txtime to avoid conflict

rtzoeller:
The socket tests request specific ports, and `test_timestamping` and `test_txtime` are currently conflicting in the port they request.

This leads to the second of the tests failing with `EADDRINUSE` when run locally.


asomers (approved):
bors r+

PR #1632, opened 2022-01-13
sys/termios: add cross-platform API for arbitrary baud rates

tmplt:
This commit adds

    pub struct ArbitraryBaudRate(pub u32);
    impl TryFrom<ArbitraryBaudRate> for BaudRate { ... }

such that arbitrary baud rates can be specified on systems that
otherwise do not support it (e.g. Linux) via abstraction.

The one problem I see with this implementation is the duplication of the
BaudRate enum. Removing the duplication would require a new macro, which
may be too complex for this feature to warrant implementation.


asomers (commented):
Why add the `ArbitraryBaudRate` structure?  I think implementing `TryFrom<u32>` directly on `BaudRate` should be fine.

tmplt:
Alan Somers ***@***.***> writes:
> Why add the `ArbitraryBaudRate` structure?  I think implementing
> `TryFrom<u32>` directly on `BaudRate` should be fine.
TryFrom<u32> is already implemented but the `u32` is the C
representation of the underlying `B*` enum. E.g.
`assert_eq!(BaudRate::try_from(14), Ok(Baudrate::B115200))`, IIRC.

PR #1635, opened 2022-01-16
Remove EventFlag::EV_SYSFLAG

rtzoeller:
It is not stable across OpenBSD versions and is reserved by the system on FreeBSD and NetBSD.

This should fix the nix build for OpenBSD, which is failing due to https://github.com/rust-lang/libc/pull/2596. Bump the libc dependency to uptake that change.


rtzoeller:
bors retry

rtzoeller:
bors retry

PR #1638, opened 2022-01-21
Suppress clippy::not_unsafe_ptr_arg_deref in mqueue, ptrace

rtzoeller:
Fix the BSD builds on nightly.

### DIFF
--- a/src/sys/ptrace/bsd.rs
+++ b/src/sys/ptrace/bsd.rs
@@ -9,8 +9,8 @@ use crate::Result;
 pub type RequestType = c_int;
 
 cfg_if! {
-    if #[cfg(any(target_os = "dragonfly", 
-                 target_os = "freebsd", 
+    if #[cfg(any(target_os = "dragonfly",
+                 target_os = "freebsd",
                  target_os = "macos",
                  target_os = "openbsd"))] {
         #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


src/sys/ptrace/bsd.rs
```diff
@@ -176,6 +177,7 @@ pub fn read(pid: Pid, addr: AddressType) -> Result<c_int> {
 }
 
 /// Writes a word into the processes memory at the given address
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub fn write(pid: Pid, addr: AddressType, data: c_int) -> Result<()> {
```
rtzoeller, line 181: ptrace::write is `unsafe` on Linux. Rather than suppress this warning, do we want to just make the ptrace API consistently unsafe?

asomers: I think technically this doesn't violate Rust's safety rules.  `libc::ptrace` doesn't dereference those addresses.  Instead, it passes them to the kernel as-is.  So from a safety perspective, this is similar to a Rust function that accepts a raw pointer argument but does nothing more than print its debugging form.
OTOH, you could make the argument that since any foreign function call is `unsafe`, and we don't really know what ptrace does, that the entire ptrace API ought to be unsafe.
I usually hew to the narrow technical definition of unsafe, but in this case I think making the entire API unsafe would be sensible.  It would probably match users' expectations.

PR #1639, opened 2022-01-22
Fix Clippy warnings on FreeBSD with the latest nightly

asomers:
* Better type safety for mqueue
* Suppress clippy::not_unsafe_ptr_arg_deref warnings in ptrace on BSD

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,6 +40,7 @@ task:
   freebsd_instance:
     image: freebsd-12-2-release-amd64
   setup_script:
+    - kldload mqueuefs
     - fetch https://sh.rustup.rs -o rustup.sh
     - sh rustup.sh -y --profile=minimal --default-toolchain $TOOLCHAIN
     - . $HOME/.cargo/env
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,11 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1636](https://github.com/nix-rust/nix/pull/1636))
 
 ### Changed
+
+- `mqueue` functions now operate on a distinct type, `nix::mqueue::MqdT`.
+  Accessors take this type by reference, not by value.
+  (#[1639](https://github.com/nix-rust/nix/pull/1639))
+
 ### Fixed
 ### Removed
 
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -1,12 +1,40 @@
 //! Posix Message Queue functions
 //!
+//! # Example
+//!
+// no_run because a kernel module may be required.
+//! ```no_run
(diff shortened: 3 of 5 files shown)


asomers:
Supersedes #1638

lucab:
POSIX does explicitly mention that `A message queue descriptor may be implemented using a file descriptor`, which I think is the case at least on Linux.
Before this, it used to be possible to check an existing open file-descriptor via `getattr` and then convert it to a message queue descriptor. Would it be possible to get that functionality back, on specific platforms? How should that look like?

asomers:
@lucab could you give an example in Rust, C, or pseudocode?  How would that feature work?

lucab:
Sure, and thanks for the prompt reply!
The initial context here is `systemd`, which is a Linux-only service manager.
It supports listening on a message queue (or socket/FIFO/etc.) on behalf of a service and activating it on demand, see <https://www.freedesktop.org/software/systemd/man/systemd.socket.html#ListenMessageQueue=>.
The activation mechanism works in a way such that systemd performs the `mq_open()`, then later passes the open file-descriptor to the spawned service/process.
The spawned process here would be a Rust binary, and we have a library method `is_mq()` to check the inherited FD. It is built on top of the previous design of `nix::mqueue::mq_getattr()`, see https://docs.rs/libsystemd/latest/libsystemd/activation/trait.IsType.html#method.is_mq.
Overall I'm not particularly attached to the existing code/design, but I feel that we are now missing primitives for fallible conversion between a file-descriptor and a mqueue-descriptor.

rtzoeller:
> I feel that we are now missing primitives for fallible conversion between a file-descriptor and a mqueue-descriptor.

I think it's worth noting that the conversion itself was infallible. You could end up with an invalid mqueue-descriptor, as `libsystemd` handled, but the conversion still succeeded. I don't think nix can make the conversion fallible without performing a similar syscall, and don't think it needs to.

[`std::os::unix::io::FromRawFd`](https://doc.rust-lang.org/nightly/std/os/unix/io/trait.FromRawFd.html) is surprising unhelpful for you, as it takes ownership and `is_mq()` takes an immutable reference. Consequently I think we'll need to implement at least [`std::convert::From`](https://doc.rust-lang.org/std/convert/trait.From.html).

Implementing [`std::os::unix::io::AsRawFd`](https://doc.rust-lang.org/nightly/std/os/unix/io/trait.AsRawFd.html) and [`std::os::unix::io::IntoRawFd`](https://doc.rust-lang.org/nightly/std/os/unix/io/trait.IntoRawFd.html) gets you from an `MqdT` to a `RawFd` (which isn't what you need), but IMO is worth doing for symmetry.

We should probably implement all four traits on Linux and any other platforms that use FDs instead of pointers for `mqd_t`, unless @asomers sees a reason not to.

PR #1640, opened 2022-01-23
Add fspacectl on FreeBSD

asomers:


src/fcntl.rs
```diff
@@ -697,6 +697,139 @@ pub fn fallocate(
...
+            fd,
+            libc::SPACECTL_DEALLOC, // Only one command is supported ATM
+            &rqsr,
+            0,                      // No flags are currently supported
+            &mut rqsr
+    )};
+    Errno::result(res).map(|_| SpacectlRange(rqsr.r_offset, rqsr.r_len))
+}
+
+/// Like [`fspacectl`], but will never return incomplete.
```
rtzoeller, line 780: I'm struggling to see a scenario where a user would want to allow the operation to return incomplete. Do you have an example?

I'm wondering if we should make `fspacectl` the looping version, since it seems like that's the common use case.

src/fcntl.rs
```diff
@@ -697,6 +697,139 @@ pub fn fallocate(
...
+/// # use nix::fcntl::*;
+/// # use tempfile::tempfile;
+/// const INITIAL: &[u8] = b"0123456789abcdef";
+/// let mut f = tempfile().unwrap();
+/// f.write_all(INITIAL).unwrap();
+/// let mut range = SpacectlRange(3, 6);
+/// while (!range.is_empty()) {
+///     let r = fspacectl(f.as_raw_fd(), range);
+///     # if r == Err(nix::Error::ENOSYS) {
+///     #     // not supported until FreeBSD 14.0
```
rtzoeller, line 758: Can we write a test for this (even if it isn't executed on PR builds due to requiring FreeBSD 14.0)?

src/fcntl.rs
asomers: If you're deallocating something really big, like an entire disk image, then fspacectl can block for a long time.  Making it interruptible allows the process to receive signals, including SIGINT.  And it allows reporting progress.

src/fcntl.rs
asomers: What should I put in the test that isn't covered by the doc test?

rtzoeller: Doh, nothing. Just not thinking about the doc test properly.

asomers:
rebased

asomers:
bors retry

PR #1642, opened 2022-01-23
InetAddr::from_std should set sin_len/sin6_len on the BSDs

### LINKED ISSUE #1246: InetAddr::from_std fails to initialize len fields
`InetAddr::from_std`  creates `libc::sockaddr_in` and `libc::sockaddr_in6` structs, but doesn't set the `len` field.  That field is only present on BSD-based operating systems.

rtzoeller:
Resolves #1246.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -330,6 +330,11 @@ impl InetAddr {
         match *std {
             net::SocketAddr::V4(ref addr) => {
                 InetAddr::V4(libc::sockaddr_in {
+                    #[cfg(any(target_os = "dragonfly", target_os = "freebsd",
+                              target_os = "haiku", target_os = "hermit",
+                              target_os = "ios", target_os = "macos",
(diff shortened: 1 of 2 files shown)


asomers (approved):

src/sys/socket/addr.rs
```diff
@@ -330,6 +330,11 @@ impl InetAddr {
         match *std {
             net::SocketAddr::V4(ref addr) => {
                 InetAddr::V4(libc::sockaddr_in {
+                    #[cfg(any(target_os = "dragonfly", target_os = "freebsd",
+                              target_os = "haiku", target_os = "hermit",
```
rtzoeller, line 334: I pulled the list of platforms which define these from rust-lang/libc. Although nix doesn't support Haiku or Hermit, there are other `cfg` blocks which reference them in nix, so I included them here.

I omitted newlib and VxWorks, as nix doesn't currently have any `cfg` blocks for them.

Issue #1371, opened 2021-01-06
Possible UB since preadv accepts &[IoVec<&mut [u8]>] 


4lDO2:
In Rust, attempting to obtain a mutable reference from an immutable reference is immediately Undefined Behavior, unless UnsafeCell is used, which includes mutable references behind immutable references (`&&mut [u8]`, or in this case `&[IoVec<&mut [u8]>]`). This has also been extended to regular pointers, that originate from stack variables or regular references, now that Stacked Borrows is about to get integrated into the compiler. While the Rust compiler may not be ever able to detect UB within preadv, as syscalls are opaque and could have other side-effects, it could theoretically optimize subsequent code under the assumption that the iovec is aliased, and hence cannot allow its data to be mutated for the lifetime of the system call. Not only that, but this behavior could also change arbitrarily in future compiler versions.

What I suggest we do to fix this potential issue, is to revert the function signature in `preadv` (changed in https://github.com/nix-rust/nix/pull/914) to take `&mut [IoVec<&mut [u8]>]`, which is something `libstd` already does, as well as the regular `readv` call in `sys::uio`. Additionally, we will need to change the implementations of `readv` and `preadv` to take `as_mut_ptr` rather than `as_ptr`, since Rust allows no mutation from any pointer coming from `as_ptr` (which borrows &self immutably).

sporksmith:
Came here to file the same issue for `process_vm_readv`, which has the same problem. IANA expert, but I started digging into this when I was writing a wrapper for `process_vm_readv`, and found I couldn't have it take `&&mut [u8]` without using unsafe to coerce pointers, and doing so would cause Miri to flag it as UB.

The language docs on interior mutability seem pretty clear that the current implementation is UB. https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#interior-mutability

sporksmith:
Example in the Rust playground that (I think) implements the same behavior as these methods. It works with `unsafe`, but don't pass the Miri checks. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8cb8a4507c42d86f6efd8a81d31f89d6

PR #1643, opened 2022-01-23
Replace the IoVec struct with IoSlice and IoSliceMut from the standard library

### LINKED ISSUE #1637: Add an "advance" function for the `iovec` type.
In Rust's standard library, the [`IoSlice`](https://doc.rust-lang.org/std/io/struct.IoSlice.html) type has methods "[`advance`](https://doc.rust-lang.org/std/io/struct.IoSlice.html#method.advance)" and "[`advance_slices`](https://doc.rust-lang.org/std/io/struct.IoSlice.html#method.advance_slices)", which are used to "advance" the slices by a set number of bytes. This is usually done in response to a `read_vectored` call.

However, the [`iovec`](https://docs.rs/nix/latest/nix/sys/uio/struct.IoVec.html) type, necessary for the `[read/write]v` and `[send/recv]msg` calls, does not have an equivalent pair of functions. This makes it difficult to call, say `readv` in such a way that you can be sure that *all* of the bytes have been written.

If there is interest in this, I am willing to write a pull request for it.

notgull:
As per discussion in #1637, the `IoVec<&[u8]>` and `IoVec<&mut [u8]>` types have been replaced with `std::io::IoSlice` and `IoSliceMut`, respectively. Notable changes made in this pull request include:

- The complete replacement of `IoVec` with `IoSlice*` types in both public API, private API, and tests.
- Replacing `IoVec` with `IoSlice` in docs.
- Replacing `&[IoVec<&mut [u8]>]` with `&mut [IoSliceMut]`, note that the slice requires a mutable reference now. This is how it's done in the standard library, and there might be a soundness issue in doing it the other way.

Resolves #1637

### DIFF
--- a/src/mount/bsd.rs
+++ b/src/mount/bsd.rs
@@ -3,7 +3,6 @@ use crate::{
     Errno,
     NixPath,
     Result,
-    sys::uio::IoVec
 };
 use libc::{c_char, c_int, c_uint, c_void};
 use std::{
@@ -11,7 +10,7 @@ use std::{
     ffi::{CString, CStr},
     fmt,
     io,
-    ptr
+    marker::PhantomData,
 };
 
 
@@ -198,13 +197,45 @@ pub type NmountResult = std::result::Result<(), NmountError>;
 #[cfg_attr(docsrs, doc(cfg(all())))]
 #[derive(Debug, Default)]
 pub struct Nmount<'a>{
-    iov: Vec<IoVec<&'a [u8]>>,
+    // n.b. notgull: In reality, this is a list that contains
+    //               both mutable and immutable pointers.
+    //               Be careful using this.
+    iov: Vec<libc::iovec>,
     is_owned: Vec<bool>,
+    marker: PhantomData<&'a ()>,
 }
 
 #[cfg(target_os = "freebsd")]
 #[cfg_attr(docsrs, doc(cfg(all())))]
 impl<'a> Nmount<'a> {
+    /// Helper function to push a slice onto the `iov` array.
+    fn push_slice(&mut self, val: &'a [u8], is_owned: bool) {
+        self.iov.push(libc::iovec {
+            iov_base: val.as_ptr() as *mut _,
+            iov_len: val.len(),
+        });
+        self.is_owned.push(is_owned);
+    }
+
+    /// Helper function to push a pointer and its length onto the `iov` array.
+    fn push_pointer_and_length(&mut self, val: *const u8, len: usize, is_owned: bool) {
+        self.iov.push(libc::iovec {
+            iov_base: val as *mut _,
+            iov_len: len,
+        });
+        self.is_owned.push(is_owned);
+    }
+
+    /// Helper function to push a `nix` path as owned.
+    fn push_nix_path<P: ?Sized + NixPath>(&mut self, val: &P) {
+        val.with_nix_path(|s| {
+            let len = s.to_bytes_with_nul().len();
+            let ptr = s.to_owned().into_raw() as *const u8;
+
+            self.push_pointer_and_length(ptr, len, true);
+        }).unwrap(); 
+    }
+
     /// Add an opaque mount option.
     ///
     /// Some file systems take binary-valued mount options.  They can be set
@@ -239,10 +270,8 @@ impl<'a> Nmount<'a> {
         len: usize
     ) -> &mut Self
     {
-        self.iov.push(IoVec::from_slice(name.to_bytes_with_nul()));
-        self.is_owned.push(false);
-        self.iov.push(IoVec::from_raw_parts(val, len));
-        self.is_owned.push(false);
+        self.push_slice(name.to_bytes_with_nul(), false);
+        self.push_pointer_and_length(val.cast(), len, false);
         self
     }
 
@@ -258,10 +287,8 @@ impl<'a> Nmount<'a> {
     ///     .null_opt(&read_only);
     /// ```
     pub fn null_opt(&mut self, name: &'a CStr) -> &mut Self {
-        self.iov.push(IoVec::from_slice(name.to_bytes_with_nul()));
-        self.is_owned.push(false);
-        self.iov.push(IoVec::from_raw_parts(ptr::null_mut(), 0));
-        self.is_owned.push(false);
+        self.push_slice(name.to_bytes_with_nul(), false);
+        self.push_slice(&[], false);
         self
     }
 
@@ -283,17 +310,8 @@ impl<'a> Nmount<'a> {
     /// ```
     pub fn null_opt_owned<P: ?Sized + NixPath>(&mut self, name: &P) -> &mut Self
     {
-        name.with_nix_path(|s| {
-            let len = s.to_bytes_with_nul().len();
-            self.iov.push(IoVec::from_raw_parts(
-                // Must free it later
-                s.to_owned().into_raw() as *mut c_void,
-                len
-            ));
-            self.is_owned.push(true);
-        }).unwrap();
-        self.iov.push(IoVec::from_raw_parts(ptr::null_mut(), 0));
-        self.is_owned.push(false);
+        self.push_nix_path(name);
+        self.push_slice(&[], false);
         self
     }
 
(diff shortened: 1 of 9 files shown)


notgull:
Checks appear to be failing because `IoSlice` does not implement `Hash`, `PartialEq`, and some other common traits. Not sure how it would be preferred to respond to this.

src/sys/sendfile.rs
```diff
@@ -68,24 +68,24 @@ cfg_if! {
                  target_os = "freebsd",
                  target_os = "ios",
                  target_os = "macos"))] {
-        use crate::sys::uio::IoVec;
+        use std::io::IoSlice;
 
         #[derive(Clone, Debug, Eq, Hash, PartialEq)]
```
asomers, line 73: This is causing failures, because IoSlice doesn't implement some of these traits.  But `SendfileHeaderTrailer` isn't `pub`, so there's no penalty to remove some of its trait impls.  What happens if you remove the offending trait impls?  If it breaks something else, then you'll probably have to manually implement them, rather than derive them.

src/sys/uio.rs
```diff
@@ -50,8 +59,9 @@ pub fn pwritev(fd: RawFd, iov: &[IoVec<&[u8]>],
 /// See also: [`readv`](fn.readv.html) and [`pread`](fn.pread.html)
 #[cfg(not(target_os = "redox"))]
 #[cfg_attr(docsrs, doc(cfg(all())))]
-pub fn preadv(fd: RawFd, iov: &[IoVec<&mut [u8]>],
+pub fn preadv(fd: RawFd, iov: &mut [IoSliceMut<'_>],
```
asomers, line 62: See #1371 .

src/sys/uio.rs
```diff
@@ -88,7 +98,7 @@ pub fn pread(fd: RawFd, buf: &mut [u8], offset: off_t) -> Result<usize>{
 /// A slice of memory in a remote process, starting at address `base`
 /// and consisting of `len` bytes.
 ///
-/// This is the same underlying C structure as [`IoVec`](struct.IoVec.html),
+/// This is the same underlying C structure as `IoVec`,
```
asomers, line 101:
```suggestion
/// This is the same underlying C structure as `IoSlice`,
```

src/sys/uio.rs
```diff
@@ -177,59 +187,3 @@ pub fn process_vm_readv(
...
-    #[cfg(all(feature = "mount", target_os = "freebsd"))]
-    pub(crate) fn from_raw_parts(base: *mut c_void, len: usize) -> Self {
-        IoVec(libc::iovec {
-            iov_base: base,
-            iov_len: len
-        }, PhantomData)
-    }
-
-    /// Create an `IoVec` from a Rust slice.
-    pub fn from_slice(buf: &'a [u8]) -> IoVec<&'a [u8]> {
```
asomers, line 214: IoSlice does not have a method by this name.  So you should probably mention in CHANGELOG that users should replace IoVec::from_slice with IoSlice::new.

src/sys/uio.rs
```diff
@@ -177,59 +187,3 @@ pub fn process_vm_readv(
...
 }
-
-/// A vector of buffers.
-///
-/// Vectored I/O methods like [`writev`] and [`readv`] use this structure for
-/// both reading and writing.  Each `IoVec` specifies the base address and
-/// length of an area in memory.
-#[repr(transparent)]
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub struct IoVec<T>(pub(crate) libc::iovec, PhantomData<T>);
```
asomers, line 188: It might help users update their codebases if you leave this struct definition in place, but mark it and all of its methods as `#[deprecated]`.

asomers (changes requested):
Don't forget a CHANGELOG, too!

notgull:
The CI errors appear to be due to a clippy error unrelated to this pull request.

asomers:
> The CI errors appear to be due to a clippy error unrelated to this pull request.

I don't think so.  I think those errors were introduced by this PR.  nmount uses IoVec/IoSlice.

notgull:
Remainder of CI failures appear to be due to the aforementioned clippy error

asomers:
The ptrace warning is fixed in master.  You need to rebase.

notgull:
Is there anything else that I need to do here before I can merge this request?

src/mount/bsd.rs
```diff
@@ -371,19 +371,27 @@ impl<'a> Nmount<'a> {
     pub fn nmount(&mut self, flags: MntFlags) -> NmountResult {
         // nmount can return extra error information via a "errmsg" return
```
asomers, line 372: This comment got separated from the code it describes.  It should be directly above line 380.

src/mount/bsd.rs
```diff
@@ -371,19 +371,27 @@ impl<'a> Nmount<'a> {
...
         // argument.
+        
+        // N.B. notgull: the last entry here is technically an IoSliceMut.
+        //               this is a workaround to keep soundness in place
+        let mut iovecs: Vec<libc::iovec> =
```
asomers, line 377: Instead of doing a const transmutation, should self.iov be a vec of IoSliceMut instead?

asomers (commented):
Sorry for the slow response, I've been working on other things lately.  This looks 99% good; I just have a few comments about nmount.

src/mount/bsd.rs
asomers: I guess the problem is that the C API isn't const-correct.  All of nmount's arguments are _supposed_ to be const, except for the errmsg argument.  So a transmutation like you had is probably Nix's best option.

src/mount/bsd.rs
```diff
@@ -198,13 +197,48 @@ pub type NmountResult = std::result::Result<(), NmountError>;
...
+    iov: Vec<libc::iovec>,
     is_owned: Vec<bool>,
+    marker: PhantomData<&'a ()>,
 }
 
 #[cfg(target_os = "freebsd")]
 #[cfg_attr(docsrs, doc(cfg(all())))]
 impl<'a> Nmount<'a> {
+    /// Helper function to push a slice onto the `iov` array.
+    #[inline]
```
asomers, line 212: There's no need to specify `#[inline]` for private methods.

asomers (commented):
I guess whether you do a mem::transmute or a pointer cast, it's equally ugly, and it's C's fault.  This PR looks almost good to me, but I suggest getting rid of the `#[inline]` on private methods.

asomers:
Could you please rebase?  You've got some conflicts now, and a rebase will also fix the test failure on uclibc.

asomers:
@notgull are you sure you rebased to the current master?  It doesn't look like it.

SUPERCILEX:
Should this be blocking the 0.24 release? It'd be nice to have the feature splits out the door.

rtzoeller:
@asomers will you have a chance to look at this in the next week or so? If not I can take a look; I agree with @SUPERCILEX that it'd be nice to get a `0.24.0` out soon (even if it means releasing a smaller `0.25.0` shortly thereafter).

src/sys/socket/mod.rs
```diff
@@ -14,10 +14,9 @@ use std::net;
 #[cfg(feature = "uio")]
 use crate::sys::time::TimeSpec;
 #[cfg(feature = "uio")]
-use crate::sys::{
-    time::TimeVal,
-    uio::IoVec
-};
+use crate::sys::time::TimeVal;
+#[cfg(feature = "uio")]
```
asomers, line 18: This line shouldn't be necessary.  It was only necessary before your change because it gated a `use crate` import.
```suggestion
```

src/sys/uio.rs
```diff
@@ -108,13 +122,70 @@ pub struct RemoteIoVec {
     pub len: usize,
 }
 
+/// A vector of buffers.
+///
+/// Vectored I/O methods like [`writev`] and [`readv`] use this structure for
+/// both reading and writing.  Each `IoVec` specifies the base address and
+/// length of an area in memory.
+#[deprecated = "`IoVec` is no longer used in the public interface, use `IoSlice` or `IoSliceMut` instead"]
```
asomers, line 130:
```suggestion
#[deprecated(
    since = "0.24.0",
    note = "`IoVec` is no longer used in the public interface, use `IoSlice` or `IoSliceMut` instead"
)]
```

asomers (changes requested):
It LGTM, apart from a few minor nits.

asomers:
Ok, it looks like you've made all requested changes and the tests pass.  Would you please squash now? Then we'll merge it.

notgull:
Squashed and ready to rumble!

asomers (approved):
bors r+

SUPERCILEX:
Thanks everyone for helping get 0.24 out! :)

PR #1644, opened 2022-01-23
Make memoffset dependency optional

rtzoeller:
Only the socket feature depends on memoffset. Allow clients to skip pulling memoffset in as a dependency if they don't need it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ bitflags = "1.1"
 cfg-if = "1.0"
 
 [target.'cfg(not(target_os = "redox"))'.dependencies]
-memoffset = "0.6.3"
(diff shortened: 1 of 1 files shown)


rtzoeller:
@asomers I wasn't sure if this was worth mentioning in the changelog, or if the existing note about breaking nix down into features was sufficient.

asomers:
This is going to make `memoffset` show up as a crate feature.  I don't know of any way to prevent that for optional dependencies.  Do you?

rtzoeller:
I'm not aware of a way to prevent that, unfortunately. I don't see it as an issue though, as it's true for all crates with optional dependencies and is an expected quirk.

asomers (approved):
bors r+

PR #1660, opened 2022-02-09
Add statx

### LINKED ISSUE #1649: Feature request: wrap Linux's `statx`
It has additional fourth "Birth time" for files.

vi:
Resolves #1649.

In current state this breaks the build for Musl targets.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -889,3 +889,6 @@ pub fn posix_fallocate(fd: RawFd, offset: libc::off_t, len: libc::off_t) -> Resu
     }
 }
 }
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub mod statx;
--- a/src/fcntl/statx.rs
+++ b/src/fcntl/statx.rs
@@ -1,0 +1,371 @@
+use std::{mem, os::unix::prelude::RawFd};
+
+use libc::{self, c_int, c_uint, mode_t};
+
+use crate::{errno::Errno, NixPath, Result, sys::stat::{SFlag, Mode}, unistd::{Uid, Gid}};
+
+libc_bitflags!(
+    /// Configuration options for statx.
+    pub struct Flags: c_int {
+        /// Allow empty `pathname`, to receive status of the specified file descriptor itself (not a directory entry) instead.
+        AT_EMPTY_PATH;
+
+        /// Do not automount the last component of the path being queried.
+        AT_NO_AUTOMOUNT;
+
+        /// Do not follow symlink in the last component of the path being queried.
+        AT_SYMLINK_NOFOLLOW;
+
+        /// Do whatever usual, non-`x` stat does. The default.
+        AT_STATX_SYNC_AS_STAT;
+
+        /// Force returned attributes to be up to date (synchonized). May trigger data writeback.
+        /// May involve additional roudtrips for networked filesystems.
+        AT_STATX_FORCE_SYNC;
+
+        /// Don't synchronize, use cached data if possible. This may lead to the returned data to be approximate.
+        AT_STATX_DONT_SYNC;
+    }
+);
+
+libc_bitflags!(
+    /// Configuration options for statx.
+    pub struct Mask: c_uint {
+        /// Want stx_mode & S_IFMT
+        STATX_TYPE;
+        /// Want stx_mode & ~S_IFMT
+        STATX_MODE;
+        /// Want stx_nlink
+        STATX_NLINK;
+        /// Want stx_uid
+        STATX_UID;
+        /// Want stx_gid
+        STATX_GID;
+        /// Want stx_atime
+        STATX_ATIME;
+        /// Want stx_mtime
+        STATX_MTIME;
+        /// Want stx_ctime
+        STATX_CTIME;
+        /// Want stx_ino
+        STATX_INO;
+        /// Want stx_size
+        STATX_SIZE;
+        /// Want stx_blocks
+        STATX_BLOCKS;
+        /// [All of the above]
+        STATX_BASIC_STATS;
+        /// Want stx_btime
+        STATX_BTIME;
+        /// Want stx_mnt_id (since Linux 5.8);
+        STATX_MNT_ID;
+        /// [All currently available fields]
+        STATX_ALL;
+    }
+);
+
+/*
+libc_bitflags!(
+    pub struct FileAttributes: c_int {
+        /// The file is compressed by the filesystem and may take extra resources to access.
+        STATX_ATTR_COMPRESSED;
+
+        /// The file cannot be modified: it cannot be deleted or renamed, no hard links can be created to this file and no data can be written to it.
+        STATX_ATTR_IMMUTABLE;
+
+        /// The file can only be opened in append mode for writing.  Random access writing is not permitted.
+        STATX_ATTR_APPEND;
+
+
+        /// File is not a candidate for backup when a backup program such as `dump` is run.
+        STATX_ATTR_NODUMP;
+
+        /// A key is required for the file to be encrypted by the filesystem.
+        STATX_ATTR_ENCRYPTED;
+
+        // ///  (since Linux 5.5)
+        // /// The  file  has fs-verity enabled.  It cannot be written to, and all reads from it will be verified
+        // /// against a cryptographic hash that covers the entire file
+        // /// (e.g., via a Merkle tree).
+        // STATX_ATTR_VERITY;
+
+        // /// (since Linux 5.8)
+        // /// The file is in the DAX (cpu direct access) state.
+        // STATX_ATTR_DAX;
+    }
+);
+*/
+
+/// Attempt to retrieve stats of `pathname`. If `pathname` is relative, `dirfd` is used as starting directory for lookups. If `dirfs` is None, current directory is used.
+/// `mask` determines what stats entries should be filled in, but kernel can actually fill more or less than requested.
+/// `statx` is not atomic: if attributes are being changed at the time of `statx` is called, the returned attributes set may have items from different moments of time.
+pub fn statx<P: ?Sized + NixPath>(
+    dirfd: Option<RawFd>,
+    pathname: &P,
+    flags: Flags,
+    mask: Mask,
+) -> Result<Statx> {
+    let dirfd = dirfd.unwrap_or(libc::AT_FDCWD);
+
+    let mut dst = mem::MaybeUninit::uninit();
+    let res = pathname.with_nix_path(|cstr| unsafe {
+        libc::statx(
+            dirfd,
+            cstr.as_ptr(),
+            flags.bits() as libc::c_int,
+            mask.bits() as libc::c_uint,
+            dst.as_mut_ptr(),
+        )
+    })?;
+
+    Errno::result(res)?;
+
+    Ok(Statx {
+        inner: unsafe { dst.assume_init() },
+    })
+}
+
+/// Get stats of specified file descriptor (not nesessarily directory) without resolving the path.
+/// Automaitcally sets the `AT_EMPTY_PATH` flag.
+pub fn statx_without_path(dirfd: RawFd, mut flags: Flags, mask: Mask) -> Result<Statx> {
+    flags.set(Flags::AT_EMPTY_PATH, true);
+    let mut dst = mem::MaybeUninit::uninit();
+    let res = unsafe {
+        libc::statx(
+            dirfd,
+            b"\0" as *const u8 as *const libc::c_char,
+            flags.bits() as libc::c_int,
+            mask.bits() as libc::c_uint,
+            dst.as_mut_ptr(),
+        )
+    };
+
+    Errno::result(res)?;
+
+    Ok(Statx {
+        inner: unsafe { dst.assume_init() },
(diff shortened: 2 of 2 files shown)


src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+    Errno::result(res)?;
+
+    Ok(Statx {
+        inner: unsafe { dst.assume_init() },
+    })
+}
+
+/// Get stats of specified file descriptor (not nesessarily directory) without resolving the path.
+/// Automaitcally sets the `AT_EMPTY_PATH` flag.
+pub fn statx_without_path(dirfd: RawFd, mut flags: Flags, mask: Mask) -> Result<Statx> {
```
asomers, line 130: Is this function really necessary?  It seems like the user can accomplish the same thing with 
```rust
statx(Some(fd), "", Flags::AT_EMPTY_path, MASK)
```

src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+#[derive(Debug,Copy,Clone)]
+pub struct Statx {
+    pub inner: libc::statx,
+}
+
+impl Statx {
+    /// Retrieve file type, if it has been returned by kernel
+    pub fn r#type(&self) -> Option<SFlag> {
```
asomers, line 157: This method name will be hard for users to type.  I suggest `filetype`.

src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+        if Mask::STATX_NLINK.bits() & self.inner.stx_mask > 0 {
+            Some(self.inner.stx_nlink)
+        } else {
+            None
+        }
+    }
+
+    /// Retrieve uid (owner, user ID), if it has been returned by kernel.
+    pub fn uid(&self) -> Option<Uid> {
+        if Mask::STATX_NLINK.bits() & self.inner.stx_mask > 0 {
```
asomers, line 185:
```suggestion
        if Mask::STATX_UID.bits() & self.inner.stx_mask > 0 {
```

src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+            Some(Uid::from_raw(self.inner.stx_uid))
+        } else {
+            None
+        }
+    }
+
+    /// Retrieve gid (group ID), if it has been returned by kernel.
+    pub fn gid(&self) -> Option<Gid> {
+        if Mask::STATX_NLINK.bits() & self.inner.stx_mask > 0 {
```
asomers, line 194:
```suggestion
        if Mask::STATX_GID.bits() & self.inner.stx_mask > 0 {
```

src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+            Some(Gid::from_raw(self.inner.stx_uid))
+        } else {
+            None
+        }
+    }
+
+    /// Retrieve file access time, if it has been returned by kernel
+    pub fn atime(&self) -> Option<libc::statx_timestamp> {
```
asomers, line 202: Returning the raw libc type isn't very useful.  Is it identical to a `struct timespec`?  If so, I suggest you convert it into a `nix::sys::time::TimeSpec`.

src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+    pub fn blocks(&self) -> Option<u64> {
+        if Mask::STATX_BLOCKS.bits() & self.inner.stx_mask > 0 {
+            Some(self.inner.stx_blocks)
+        } else {
+            None
+        }
+    }
+
+    /// Retrieve file size as a number of blocks, if it has been returned by kernel
+    pub fn blksize(&self) -> Option<u64> {
```
asomers, line 265: Is this identical to `blocks`?

src/fcntl/statx.rs
```diff
@@ -0,0 +1,371 @@
...
+        if Mask::STATX_BASIC_STATS.bits() & self.inner.stx_mask == Mask::STATX_BASIC_STATS.bits() {
+            Some((self.inner.stx_dev_major, self.inner.stx_dev_minor))
+        } else {
+            None
+        }
+    }
+
+    /// Retrieve pointed-to device major and minor numbers (first and second elements of the tuple respectively) of this character or block device file, if it has been returned by kernel.
+    /// Note that this function does not check for the file type. It would return `Some` even for regular files.
+    pub fn rdev_major_minor(&self) -> Option<(u32, u32)> {
```
asomers, line 297: What's the difference compared to the previous method?

asomers (changes requested):

src/fcntl/statx.rs
vi: Yes, it is a helper function to call more generic `statx`. The syscall does different things based on parameters, so I expect it to be more readable and obvious if multiple functions represented different cases. For example, it is easy for forget about `Flags::AT_EMPTY_PATH` or wonder what to put inside the path argument, forcing user to stop coding and start looking at docs.
Also I expect even empty path should be terminated be zero byte, so just `""` may allocate (unless `with_nix_path` have some special optimisation).

src/fcntl/statx.rs
vi: No, probably I just forgot to fill in new doccomment and implementation after copying neighbouring method.

src/fcntl/statx.rs
vi: For example, `/dev/null` character special file lives in a `/dev/` directory of mounted filesystem addressed by a device number `5h/5d`. But `/dev/null` itself _points_ to another, non-filesystem device of type `1,3`. If you `cp -a /dev/null /root/null`, the resulting file would have different `dev_major_minor` (as it is now on another filesystem), but would point to the same device (`1,3`), so the same `rdev_major_minor`. `dev_major_minor` is relevant for all fiels on Linux, `rdev_major_minor` is relevant only for block and character special files. `dev_major_minor` + `ino` = sort of a numeric identifier of a specific file on Linux.

vi: But, as usual, there may be copy&paste sloppiness here, so implementation may return wrong number.

How such syscall wrappers are supposed to be tested? Shall unit tests create some tricky files (as root) to be able to properly test all those things? Making Linux kernel to return some newer attributes may be nontrivial.

src/fcntl/statx.rs
asomers: I'm leery of providing convenience methods like this.  Historically, whenever Nix goes beyond the minimum of "provide Rusty bindings" and invent a convenience API, we get it wrong.  Absent a more compelling reason, I would prefer to remove this function.

src/fcntl/statx.rs
asomers: What is it supposed to do then?  Because it's just returning the same stx_blocks variable.

src/fcntl/statx.rs
asomers: Ok, it must be a copy/paste error then.  Please fix it.

As for testing, you can add tests in the test directory.  You don't need to worry about testing every feature advertised by the kernel; that's a job for the kernel test suite.  You just need to test enough features to prove that the Rust bindings work.

asomers (changes requested):

src/fcntl/statx.rs
vi: Done, also added the appropriate `From` implementation within the `statx` module. Or should it be moved to the place where other `TimeSpec` `From` implementations live with a duplicated `#[cfg]` line?

src/fcntl/statx.rs
vi: Fixed to return `blocksize`, fixed the doccomment.

src/fcntl/statx.rs
```diff
@@ -0,0 +1,350 @@
...
+        STATX_BTIME;
+        /// Want stx_mnt_id (since Linux 5.8);
+        STATX_MNT_ID;
+        /// [All currently available fields]
+        STATX_ALL;
+    }
+);
+
+/// Attempt to retrieve stats of `pathname`. If `pathname` is relative, `dirfd`
+/// is used as starting directory for lookups. If `dirfs` is None, current
```
asomers, line 71:
```suggestion
/// is used as starting directory for lookups. If `dirfd` is None, current
```

src/fcntl/statx.rs
```diff
@@ -0,0 +1,350 @@
...
+/// directory is used.
+/// `pathname` may be empty string. But instead of specifying empty string
+/// literal, you are adviced to use zero-terminated `CStr` to avoid extra allocation
```
asomers, line 74: How about adding some examples to show usage?  That would be especially helpful for the "empty pathname" case.

src/fcntl/statx.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    Ok(Statx {
+        inner: unsafe { dst.assume_init() },
+    })
+}
+
+#[derive(Debug,Copy,Clone)]
+pub struct Statx {
+    pub inner: libc::statx,
+}
```
asomers, line 109: Since it has only one member, it can be a NewType.
```suggestion
#[repr(transparent)]
pub struct Statx(libc::statx);
```

src/fcntl/statx.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// None means kernel does not indicate this attrbiute is supported by the filesystem
+    pub fn encrypted(&self) -> Option<bool> {
+        if self.inner.stx_attributes_mask & libc::STATX_ATTR_ENCRYPTED as u64 > 0 {
+            Some(self.inner.stx_attributes & libc::STATX_ATTR_ENCRYPTED as u64 > 0)
+        } else {
+            None
+        }
+    }
+
+    /*
```
asomers, line 329: Why are these two methods commented out?

vi: Because of current version of `libc` crate does not expose e.g. `STATX_ATTR_VERITY`. See my question at https://github.com/nix-rust/nix/issues/1649#issuecomment-1033938084.
Android target's libc lacks even more consts in libc.

src/fcntl/statx.rs
vi: Good idea. Shall doc examples be marked as testable code (and assert something about typical external files e.g. `/dev/null`)?

src/fcntl/statx.rs
vi: Done (inner tuple field is still `pub`, unlike in the suggested change).

SteveLauC:
Hey @vi :) What is the status of this PR? I really hope this can be merged

src/fcntl.rs
```diff
@@ -889,3 +889,6 @@ pub fn posix_fallocate(fd: RawFd, offset: libc::off_t, len: libc::off_t) -> Resu
     }
 }
 }
+
+#[cfg(all(feature="fs", any(target_os = "android", target_os = "linux")))]
```
SteveLauC, line 893:
```rust
#[cfg_attr(all(
    feature = "fs",
    any(
        target_os = "android",
        all(target_os = "linux", not(target_env = "musl"))
    )
))]
```
Since this is not available on `musl`, this configuration should work

src/fcntl/statx.rs
SteveLauC: I submitted a [PR](https://github.com/rust-lang/libc/pull/2941) to add the missing constants on GNU/Linux

src/fcntl/statx.rs
```diff
@@ -0,0 +1,346 @@
...
+    dirfd: Option<RawFd>,
+    pathname: &P,
+    flags: Flags,
+    mask: Mask,
+) -> Result<Statx> {
+    let dirfd = dirfd.unwrap_or(libc::AT_FDCWD);
+
+    let mut dst = mem::MaybeUninit::uninit();
+    let res = pathname.with_nix_path(|cstr| unsafe {
+        libc::statx(
```
SteveLauC, line 90: This wrapper function is only [exposed](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/gnu/mod.rs#L1189~L1195) on GNU/Linux as glibc added it. 

What about making this `statx` module temporarily exclusive to GNU/Linux? When the missing stuff is added, we can easily add support for other platforms by changing a few lines of code.

vi:
> What is the status of this PR?

It is waiting for answers for questions:

* What to do with the missing constants? Based on your pull request to libc, supposed answer is "wait for libc support", then "wait when `nix` starts depending on new libc".
* Shall we get ahead of musl upstream and provide `statx` even on musl by using hacks?
* Is overall design of the feature is OK and in line with `nix`? If code review comments e.g. about `statx_without_path` are resolved, and points above are resolved, will it be OK for merge or there will be just another round of problems?

SteveLauC:
> Shall we get ahead of musl upstream and provide statx even on musl by using hacks?

IMHO, this is not a good idea as `musl` hasn't provided a wrapper in their library. As I said in this [comment](https://github.com/nix-rust/nix/pull/1660#discussion_r987500799), I would like to temporarily just support GNU/Linux.

AppleSheeple:
Is this waiting on any blockers, abandoned, or forgotten?

vi:
@AppleSheeple , You can claim it if you want.

PR #1664, opened 2022-02-21
also implement Read and Write for &PtyMaster

doy:
align with std::fs::File which also does this because the underlying calls are just syscalls which are safe to run concurrently


rtzoeller:
@doy can you add a CHANGELOG.md entry (squashed with the current commit)?

doy:
done

PR #1668, opened 2022-03-05
Change getrlimit and setrlimit to use rlim_t directly.

### LINKED ISSUE #1666: nix::sys::resource::getrlimit doesn't actually return None
Documentation says:

https://github.com/nix-rust/nix/blob/6123083a4702002e1958229735a4fd70cb326257/src/sys/resource.rs#L176-L177

but code unconditionally returns `Some()`:

https://github.com/nix-rust/nix/blob/6123083a4702002e1958229735a4fd70cb326257/src/sys/resource.rs#L209-L212

This documentation is leftover from the old PR code [which did return `None`:](https://github.com/nix-rust/nix/pull/1190/files#diff-cd0992ba58f628e359dd65444caf718d326a635698ef9f893eb50a7c4e05ee2eR215-R218)

```rust
    Errno::result(res).map(|_| {
        (if rlim.rlim_cur != RLIM_INFINITY { Some(rlim.rlim_cur) } else { None },
         if rlim.rlim_max != RLIM_INFINITY { Some(rlim.rlim_max) } else { None })
    })
```

... but the new PR [removed that branch.](https://github.com/nix-rust/nix/pull/1302#issuecomment-706115931)

It would be nice to not return `Option` from `getrlimit` at least, since it makes the caller code more annoying when we only want to call `setrlimit` when we *need* to, which requires comparing the returned soft limit vs the returned hard limit. And if `getrlimit` is changed to not special-case `INFINITY`, the special-case might as well be removed from `setrlimit` too so that one doesn't need to `Some()`-wrap the values when passing them from `getrlimit` to `setrlimit`.

Arnavion:
Fixes #1666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Removed the the `PATH_MAX` restriction from APIs accepting paths. Paths
   will now be allocated on the heap if they are too long. In addition, large
   instruction count improvements (~30x) were made to path handling.
+- Changed `getrlimit` and `setrlimit` to use `rlim_t` directly
+  instead of `Option<rlim_t>`.
+  (#[1668](https://github.com/nix-rust/nix/pull/1668))
 
 ### Fixed
 
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -8,7 +8,7 @@ use std::mem;
 
 cfg_if! {
     if #[cfg(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")))]{
-        use libc::{__rlimit_resource_t, rlimit, RLIM_INFINITY};
+        use libc::{__rlimit_resource_t, rlimit};
     } else if #[cfg(any(
         target_os = "freebsd",
         target_os = "openbsd",
@@ -19,7 +19,7 @@ cfg_if! {
         target_os = "dragonfly",
         all(target_os = "linux", not(target_env = "gnu"))
     ))]{
-        use libc::{c_int, rlimit, RLIM_INFINITY};
+        use libc::{c_int, rlimit};
     }
 }
 
@@ -173,8 +173,8 @@ libc_enum! {
 
 /// Get the current processes resource limits
 ///
-/// A value of `None` indicates the value equals to `RLIM_INFINITY` which means
-/// there is no limit.
+/// The special value `RLIM_INFINITY` indicates that no limit will be
+/// enforced.
 ///
 /// # Parameters
 ///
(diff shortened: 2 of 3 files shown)


asomers (approved):
LGTM.  @rtzoeller do you have any thoughts on this one?

rtzoeller:
@Arnavion can you add back a reference to `RLIM_INFINITY` in the documentation for both the getter and setter? Even if nix isn't handling it, I'd expect that users might need to.

We can paraphrase from this part of the documentation:

> If a call to getrlimit() returns RLIM_INFINITY for a resource, it means the implementation shall not enforce limits on that resource. Specifying RLIM_INFINITY as any resource limit value on a successful call to setrlimit() shall inhibit enforcement of that resource limit.

Otherwise the changes look good to me.

Arnavion:
Done.

src/sys/resource.rs
```diff
@@ -218,21 +218,22 @@ pub fn getrlimit(resource: Resource) -> Result<(Option<rlim_t>, Option<rlim_t>)>
...
+///   resource.
 /// * `hard_limit`: The ceiling for the soft limit. Must be lower or equal to
-///   the current hard limit for non-root users. Note: `None` input will be
-///   replaced by constant `RLIM_INFINITY`.
+///   the current hard limit for non-root users.
 ///
-/// > Note: for some os (linux_gnu), setting hard_limit to `RLIM_INFINITY` can
-/// > results `EPERM` Error. So you will need to set the number explicitly.
+/// The special value `RLIM_INFINITY` indicates that no limit will be
+/// enforced. For some OSes (linux_gnu), setting the hard limit to
```
asomers, line 226: I don't think that Linux returns EPERM because the user tried to set RLIM_INFINITY specifically.  Rather, I think that happens because the user tries to raise the limit.  That's disallowed on every OS.

Arnavion: Yes, I wasn't sure why the comment was initially worded the way it was. I can fix it.

Arnavion: Actually, is there a point in fixing it? It would just be duplicating the spec, that's already linked as a reference. Should I just delete that sentence?

Arnavion: @asomers Ping.

CHANGELOG.md
```diff
@@ -54,6 +54,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1665](https://github.com/nix-rust/nix/pull/1665))
 - Implemented `Read` and `Write` for `&PtyMaster`
   (#[1664](https://github.com/nix-rust/nix/pull/1664))
+- Changed `getrlimit` and `setrlimit` to use `rlim_t` directly
```
rtzoeller, line 57: This should go in the "Changed" section, rather than "Added".

If you squash and rebase, this LGTM.

src/sys/resource.rs
asomers: Removing it should be ok, but we shouldn't leave the wrong explanation in place.

Arnavion: Done.

CHANGELOG.md
Arnavion: Done.

asomers (approved):
bors r+

rtzoeller:
bors retry

PR #1669, opened 2022-03-06
Fix complation on riscv32

kraj:
Signed-off-by: Khem Raj <raj.khem@gmail.com>


asomers (approved):
LGTM.

Note that we aren't currently running CI for riscv.

bors r+

PR #1672, opened 2022-03-10
Make `uname` always safe

koute:
Currently `uname` doesn't check for errors and just blindly assumes that it always succeeds. According to the manpage this function can fail, even though no actual errors are defined:

```
RETURN VALUE
       Upon successful completion, a non-negative value shall be returned.  Otherwise, -1 shall be returned and errno set to indicate the error.

ERRORS
       No errors are defined.

       The following sections are informative.
```

Looking at [the glibc's sources](https://github.com/bminor/glibc/blob/b92a49359f33a461db080a33940d73f47c756126/posix/uname.c#L29) we can see that it indeed could fail if the internal `gethostname` call fails for some reason.

This code also assumes that every field of `utsname` is going to be initialized by the call to `uname`, which apparently is also not true. Even though the interface doesn't expose this field so it's not a problem in practice (although it might be UB since we do call `assume_init` on the whole struct) [the `utsname` does have a `domainname` field](https://docs.rs/libc/0.2.119/libc/struct.utsname.html) which glibc doesn't initialize.

The code also assumes that every field is a valid UTF-8 string, which is also technically not guaranteed.

The code also assumes that every field will be null terminated, which might not be true if any of the strings are too long (since glibc uses `strncpy` which will *not* null-terminate the string if it ends up running out of space).

This PR should fix all of these problems.

This is a breaking change.

### DIFF
--- a/src/sys/utsname.rs
+++ b/src/sys/utsname.rs
@@ -1,8 +1,9 @@
 //! Get system identification
 use std::mem;
-use libc::{self, c_char};
-use std::ffi::CStr;
-use std::str::from_utf8_unchecked;
+use std::os::unix::ffi::OsStrExt;
+use std::ffi::OsStr;
+use libc::c_char;
+use crate::{Errno, Result};
 
 /// Describes the running system.  Return type of [`uname`].
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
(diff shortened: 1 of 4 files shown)


src/sys/utsname.rs
```diff
@@ -1,55 +1,53 @@
...
-    pub fn machine(&self) -> &str {
-        to_str(&(&self.0.machine as *const c_char ) as *const *const c_char)
+    pub fn machine(&self) -> &[u8] {
+        cast_and_trim(&self.0.machine)
     }
 }
 
 /// Get system identification
-pub fn uname() -> UtsName {
+pub fn uname() -> Result<UtsName, Errno> {
```
asomers, line 39:
```suggestion
pub fn uname() -> Result<UtsName> {
```

src/sys/utsname.rs
```diff
@@ -1,55 +1,53 @@
 //! Get system identification
 use std::mem;
-use libc::{self, c_char};
-use std::ffi::CStr;
-use std::str::from_utf8_unchecked;
+use libc::c_char;
+use crate::Errno;
```
asomers, line 4:
```suggestion
use crate::Result;
```

asomers (changes requested):
Do you know of any cases where the strings _aren't_ valid UTF-8 ?  It is certainly more convenient for the user if we return a `&str` than a `&[u8]`.  Perhaps do a checked conversion to utf8?

koute:
> Do you know of any cases where the strings _aren't_ valid UTF-8 ?

Not specifically, but e.g. someone could set a non-UTF8 hostname with `sethostname` which will end up in the `nodename` returned here.

> It is certainly more convenient for the user if we return a `&str` than a `&[u8]`. Perhaps do a checked conversion to utf8?

Well, I see three possibilities here:

1. Return a `&[u8]`.
2. Return a `&str` and panic if it can't be converted.
3. Return a `Result<&str, Utf8Error>`.

Converting the `&[u8]` to string on the user's side is just a matter of changing `utsname.nodename()` to `std::str::from_utf8(utsname.nodename())?` so it's not *that* big of a deal, and it's not like this is a super frequently used function. I'd rather not go for (2) because I think libraries shouldn't panic on errors, and (3) feels like a halfway measure.

asomers:
What if it returns an `OsString`?  Those are allowed to be non-UTF8.  See fcntl::readlink for an example.  Would that work?

koute:
> What if it returns an `OsString`? Those are allowed to be non-UTF8. See fcntl::readlink for an example. Would that work?

`OsString` allocates so that's not great, but I guess `&OsStr` should indeed work. Done!

rtzoeller:
LGTM; can you squash and rebase?

koute:
> LGTM; can you squash and rebase?

Done!

asomers (approved):
bors r+

PR #1675, opened 2022-03-13
Fix a panic in Linkaddr::addr

asomers:
The function assumed something about the values of the sockaddr_dl's
fields.  But because the inner type is public, we musn't do that.  The
only solution is to change the function's signature to return an Option.


rtzoeller (approved):
bors r+

asomers:
bors retry

PR #1677, opened 2022-03-14
Use the same signature for LinkAddr::addr on all platforms

asomers:
This should've been done as part of #1675


rtzoeller:
Doh, good catch.

bors r+

rtzoeller:
bors retry

PR #1679, opened 2022-03-16
WIP implement some GETREGSET functionality for x64 and aarch64

Gankra:
The CFGs are kinda all over the place... I don't have a good sense for how fragmented the different env's are.

Note that this depends on https://github.com/rust-lang/libc/pull/2719 for the definitions of user_fpsimd_struct

### DIFF
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -4,19 +4,34 @@ use cfg_if::cfg_if;
 use std::{mem, ptr};
 use crate::Result;
 use crate::errno::Errno;
-use libc::{self, c_void, c_long, siginfo_t};
+use libc::{self, c_void, c_long, siginfo_t, NT_PRSTATUS, NT_PRFPREG, iovec};
 use crate::unistd::Pid;
 use crate::sys::signal::Signal;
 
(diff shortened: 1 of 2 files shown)


src/sys/ptrace/linux.rs
```diff
@@ -233,6 +268,33 @@ fn ptrace_get_data<T>(request: Request, pid: Pid) -> Result<T> {
...
+
+    let res = unsafe {
+        libc::ptrace(request as RequestType,
+                     libc::pid_t::from(pid),
+                     regset as usize as *const c_void,
+                     &mut output as *mut _ as *mut c_void)
+    };
+    // If we get the struct definition wrong for a platform, linux can just scribble all over
+    // our memory. In some sense it's "too late" because it could have written past the bounds
+    // of our `data` value, but it's better to try to crash now than keep the program going!
```
jasonwhite, line 291: I don't think this comment is accurate. From what I understand reading the ptrace docs, it won't write past the end of our buffer since it knows that from the input `iov_len`. It sets the length of `iov_len` so that we can know if our buffer is too small or too large for the full set of registers. So, there shouldn't be any stack corruption here as far as I can tell (but this assertion is still great to have).

PR #1685, opened 2022-03-22
Deprecate IpAddr, Ipv4Addr, and Ipv6Addr

### LINKED ISSUE #1681: Deprecate Ipv4Addr, Ipv6Addr, and IpAddr
These three types from `nix::sys::socket::addr` are identical to `std::net::Ipv4Addr`, `std::net::Ipv6Addr` and `std::net::IpAddr`, respectively.  They should be deprecated and replaced by the standard types.  There are a few other Nix functions that use them internally however, and those must be adapted.

asomers:
Because they're redundant with types in the standard library.

Fixes #1681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Deprecated `InetAddr` and `SockAddr` in favor of `SockaddrIn`, `SockaddrIn6`,
   and `SockaddrStorage`.
   (#[1684](https://github.com/nix-rust/nix/pull/1684))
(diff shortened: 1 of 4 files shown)


asomers:
Ugh, apparently you can't deference a pointer in a const fn until rustc 1.57.0.  Changing it to a mem::transmute only lowers the compatibility threshold to 1.56.0.

rtzoeller (approved):
I don't see a way to make `ipv4addr_to_libc` const in 1.46 either. The generated assembly is concise though, so I'm not overly concerned. If it is a bottleneck for someone we can evaluate providing a const version where available.

LGTM, just needs a squash.

asomers:
bors r=rtzoeller

PR #1686, opened 2022-03-23
[skip ci] better docs for SockaddrLike::from_raw

### LINKED ISSUE #1680: Unclear documentation for `sockaddr_storage_to_addr()`
In the latest master,  the documentation for `sockaddr_storage_to_addr()` says:

> The `len` argument should be the number of bytes in the `sockaddr_storage` that are actually allocated and valid.  It must be at least as large as all the useful parts of the structure.  Note that in the case of a `sockaddr_un`, `len` need not include the terminating null.

For a unix sockaddr, the "[The len] must be at least as large as all the useful parts of the structure" makes it sound like the len is allowed to be *larger* than the useful part of the structure. It's also unclear what "valid" and "useful" mean in this context. For example, it's common for C code to initialize something like `sockaddr_un { family: AF_UNIX, path: b"tmp.sock\0" }`, but use a length of `size_of(sockaddr_un)` when passed to syscalls. But `sockaddr_storage_to_addr()` calls `from_raw_parts()` which has the documentation:

> `sun_len` is the size of the valid portion of the struct, excluding any trailing NUL

So the `len` argument to `sockaddr_storage_to_addr()` must *not* be larger than `offsetof(addr.path) + strlen(addr.path)` for pathname unix sockaddrs.

This issue causes a panic in the latest nix release (0.23.1) due to https://github.com/nix-rust/nix/blob/afba7c5a33dd4fd62b047e64089487cf822ccec2/src/sys/socket/addr.rs#L619-L622

asomers:
Fixes #1680

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -975,10 +975,22 @@ pub trait SockaddrLike: private::SockaddrLikePriv {
     /// Some C APIs from provide `len`, and others do not.  If it's provided it
     /// will be validated.  If not, it will be guessed based on the family.
     ///
+    /// # Arguments
+    ///
+    /// - `addr`:   raw pointer to something that can be cast to a
+    ///             `libc::sockaddr`. For example, `libc::sockaddr_in`,
+    ///             `libc::sockaddr_in6`, etc.
+    /// - `len`:    For fixed-width types like `sockaddr_in`, it will be
+    ///             validated if present and ignored if not.  For variable-width
+    ///             types it is required and must be the total length of valid
+    ///             data.  For example, if `addr` points to a
+    ///             named `sockaddr_un`, then `len` must be the length of the
+    ///             structure up to but not including the trailing NUL.
+    ///
     /// # Safety
     ///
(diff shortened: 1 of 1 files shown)


asomers:
cc @stevenengler

src/sys/socket/addr.rs
```diff
@@ -975,10 +975,22 @@ pub trait SockaddrLike: private::SockaddrLikePriv {
     /// Some C APIs from provide `len`, and others do not.  If it's provided it
     /// will be validated.  If not, it will be guessed based on the family.
     ///
+    /// # Arguments
+    ///
+    /// - `addr`:   raw pointer to something that can be cast to a
+    ///             `libc::sockaddr`. For example, `libc::sockaddr_in`,
+    ///             `libc::sockaddr_in`, etc.
```
rtzoeller, line 982: Should the second `libc::sockaddr_in` be `libc::sockaddr_in6`?

asomers: Yep.

PR #1692, opened 2022-04-05
Add IP_DONTFRAG and IPV6_DONTFRAG SockOpts

junhochoi:
IP_DONTFRAG: iOS, macOS
IPV6_DONTFRAG: android, iOS, linux and macOS

Test: `cargo test --test test dontfrag_opts`

Some CI tests running ENOPROTOOPT are disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1670](https://github.com/nix-rust/nix/pull/1670))
 - Added `waitid`.
   (#[1584](https://github.com/nix-rust/nix/pull/1584))
+- Added `Ipv6DontFrag` for android, iOS, linux and macOS.
+- Added `IpDontFrag` for iOS, macOS.
+  (#[1692](https://github.com/nix-rust/nix/pull/1692))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


test/sys/test_sockopt.rs
```diff
@@ -196,3 +196,41 @@ fn test_ttl_opts() {
...
+#[test]
+#[cfg(all(
+    any(target_arch = "x86", target_arch = "x86_64"),
+    any(
+        target_os = "android",
+        target_os = "ios",
+        target_os = "linux",
+        target_os = "macos",
+    )
+))]
```
rtzoeller, line 224: This logic is incorrect - why is the target architecture limited?

There aren't any x86 iOS targets, and the test passes when I run it locally on Android via Termux.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased] - ReleaseDate
 ### Added
 
+- Added `Ipv6DontFrag` for android, iOS, linux and macOS.
+- Added `IpDontFrag` for iOS, macOS.
+  (#[1692](https://github.com/nix-rust/nix/pull/1692))
```
rtzoeller, line 11: Please move this lower in the "Added" section - I'd prefer to keep #1611 at the top.

junhochoi: Moved. thanks

test/sys/test_sockopt.rs
junhochoi: This is because some CI test failed. Let me take a look again.

junhochoi: Updated to exclude this test on qemu-based CI.

rtzoeller (approved):

PR #1693, opened 2022-04-08
Document aliases for functions like getuid()

### LINKED ISSUE #1673: Add `doc(alias)` to functions
I had some trouble finding `User::from_uid` and `Uid::current` when looking for `getpwuid` and `getuid` replacements. It would be nice to add some [doc aliases](https://doc.rust-lang.org/rustdoc/advanced-features.html#add-aliases-for-an-item-in-documentation-search) for the underlying C function names so that a library user can search the C function name and find the corresponding Rust function.

Unfortunately this feature was stabilized in 1.48 so we will have to wait for a MSRV bump to do it.

rtzoeller:
Add the autocfg crate as a build dependency, and introduce `has_doc_alias` as a conditional compilation symbol.

Closes #1673.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,9 @@ pin-utils = { version = "0.1.0", optional = true }
 [target.'cfg(not(target_os = "redox"))'.dependencies]
 memoffset = { version = "0.6.3", optional = true }
 
+[build-dependencies]
+autocfg = "1.1.0"
+
 [features]
 default = [
   "acct", "aio", "dir", "env", "event", "feature", "fs",
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,7 @@
+fn main() {
+    let cfg = autocfg::new();
+
+    if cfg.probe_rustc_version(1, 52) {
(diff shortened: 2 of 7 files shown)


rtzoeller:
@asomers I'm not sure it's worth adding `autocfg` just for this, but I think it's a potential option for addressing some of the MSRV-related TODOs without actually bumping the MSRV. Thoughts?

asomers:
I guess it would work, but I'm not sure that it's worth the effort.  Are any of those MSRV TODOs really very important?

rtzoeller:
I don't think any are particularly critical - I came across this while revisiting trying to make `ipv4addr_to_libc` const.

#1673 is essentially the same to do with and without the MSRV check (and you can subsequently clean it up with a regex), so the marginal cost to do it now is low.

asomers (approved):
LGTM.

bors r+

rtzoeller:
I'll fix the cross issues that are popping up.

rtzoeller:
bors r=asomers

PR #1697, opened 2022-04-14
Add ptrace::read_user and ptrace::write_user

nbaksalyar:


nbaksalyar:
Hmm, the `Linux armv7` build failure seems to be unrelated to changes made in this PR.

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `Ipv6DontFrag` for android, iOS, linux and macOS.
 - Added `IpDontFrag` for iOS, macOS.
   (#[1692](https://github.com/nix-rust/nix/pull/1692))
+- Added `ptrace::read_user` and `ptrace::write_user` for Linux.
```
rtzoeller, line 67: Can you rebase such that this is in the latest unreleased section (the current changes it's near are part of the `0.24.0` release) and also add a link back to this PR, like the other items in the CHANGELOG?

src/sys/ptrace/linux.rs
```diff
@@ -481,3 +481,24 @@ pub unsafe fn write(
 {
     ptrace_other(Request::PTRACE_POKEDATA, pid, addr, data).map(drop)
 }
+
+/// Reads a word from a user area at `offset`.
+/// The user struct definition can be found in `/usr/include/sys/user.h`.
+pub fn read_user(pid: Pid, offset: AddressType) -> Result<c_long> {
+    ptrace_peek(Request::PTRACE_PEEKUSER, pid, offset, ptr::null_mut())
+}
```
rtzoeller, line 489: @asomers what are your thoughts on having separate `read` and `read_user` functions (and the same for write), vs. combining them and taking a third parameter for "data" or "user"?

Same for write.

asomers: In this case I think it's better to have separate functions.  IIUC the PEEKDATA and PEEKUSER don't really do the same thing at all, and it's unlikely that a consumer's crate would have a single code path for both.

CHANGELOG.md
nbaksalyar: Done. Thanks!

test/sys/test_ptrace.rs
```diff
@@ -197,15 +197,36 @@ fn test_ptrace_syscall() {
...
+            #[cfg(target_arch = "x86")]
+            let get_rax_offset = |user_struct_ptr: *const libc::user| {
+                unsafe { &(*user_struct_ptr).regs.orig_eax as *const _ }
+            };
+
+            let get_syscall_from_user_area = || {
+                // Find the offset of `user.regs.rax` (or `eax` for x86)
+                let user_struct = std::mem::MaybeUninit::<libc::user>::uninit();
+                let user_struct_ptr = user_struct.as_ptr();
+                let rax_offset = get_rax_offset(user_struct_ptr) as usize - user_struct_ptr as usize;
```
rtzoeller, line 214: Nix already depends on the `memoffset` crate conditionally, and we can use it in our tests (may need to update Cargo.toml to do so).

It should be possible to use [`memoffset::offset_of`](https://docs.rs/memoffset/latest/memoffset/macro.offset_of.html) to simplify this.

nbaksalyar: Great suggestion, that should be done :+1:

rtzoeller (approved):
bors r+

PR #1708, opened 2022-04-25
Upgrade dev-dependencies to the latest versions

rtzoeller:
Don't change how the versions are specified (i.e. whether a patch version is listed).

parking_lot cannot be upgraded due to 0.12.0 bumping the MSRV to 1.49.

In many cases, nix was one of the only consumers of the current versions (i.e. we are late to upgrade).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,14 +78,14 @@ zerocopy = ["fs", "uio"]
 
 [dev-dependencies]
 assert-impl = "0.1"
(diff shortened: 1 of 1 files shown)


rtzoeller:
Follow on to #1707; none of these should require any code changes.

asomers (approved):
Makes sense. I don't think there are very many reasons to prefer a lower version of a dev dependency.

bors r+

PR #1715, opened 2022-05-07
Log compiler version in Cirrus CI

rtzoeller:


.cirrus.yml
```diff
@@ -16,6 +16,8 @@ env:
 build: &BUILD
   build_script:
     - . $HOME/.cargo/env || true
+    - $TOOL +$TOOLCHAIN -Vv
+    - rustc +$TOOLCHAIN -Vv
```
rtzoeller, line 20: @asomers any idea if it's possible to ask cargo for the rustc version, instead of hard-coding it like this? 

Feels a bit weird to be using `$TOOL` everywhere for `cargo`, but hard-code `rustc`.

asomers: Nope.  AFAIK invoking `rustc` is the best way to do it.

asomers (approved):

rtzoeller:
bors r=asomers

PR #1717, opened 2022-05-07
Define FsType constants for musl

jirutka:
These constants are provided by Linux (linux/magic.h), not libc.

See rust-lang/libc#2639.


asomers (approved):
LGTM.

bors r+

PR #1724, opened 2022-05-27
Return None if user does not exists when looking for users

lmm-git:
getpwuid_r and getpwnam_r does return ENOENT if a user does not exists. This might lead to the unexpected case where the outer Result may fail even if only the user does not exists.

This commit changes this behavior to return a successful Result but with an empty Option. It unifies the behavior with older systems.


lmm-git:
Nevermind, apparently a nss module caused the ENOENT, which shouldn't be passed via the api calls in question. After fixing the nss module everything works now as expected. ENOENT shouldn't be an error, which is raised by the used API call.

PR #1727, opened 2022-05-30
Add From<uid_t> and From<gid_t>

vkkoskie:
Conversions to/from primitive uid_t and gid_t to newtype Uid and Gid types are valid and infallible, but are only implemented in one direction. This provides the counterparts in the other direction.

These conversions are identical in behavior to the from_raw methods. However, using the more idiomatic From trait enables easier interoperability with other crates (e.g., deserialization with serde)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `aio_writev` and `aio_readv`.
   (#[1713](https://github.com/nix-rust/nix/pull/1713))
 
+- impl `From<uid_t>` for `Uid` and `From<gid_t>` for `Gid`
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `aio_writev` and `aio_readv`.
   (#[1713](https://github.com/nix-rust/nix/pull/1713))
 
+- impl `From<uid_t>` for `Uid` and `From<gid_t>` for Gid
```
rtzoeller, line 12: Nit: Can you add backticks around `Gid` like you have them for `Uid`?

rtzoeller (approved):
Thanks! Can you address a minor CHANGELOG nitpick and then squash everything into one commit?

vkkoskie:
All comments addressed. Good to go.

PR #1729, opened 2022-05-31
Enable SockaddrStorage::{as_link_addr, as_link_addr_mut} on Linux.

### LINKED ISSUE #1728: No way to obtain a `LinkAddr` on Linux
On Linux, there is no method `SockaddrStorage::as_link_addr`, so it is not possible to obtain a MAC address for an interface using safe Rust. I cooked up the following, which seems to work, but it feels like this should be possible using only safe code:

```rust
fn sockaddr_link(sockaddr: &SockaddrStorage) -> Option<LinkAddr> {
    if let Some(AddressFamily::Packet) = sockaddr.family() {
        // Safety: Checked the address family
        unsafe { LinkAddr::from_raw(sockaddr.as_ptr(), None) }
    } else {
        None
    }
}
```

asomers:
This was an oversight from #1684.

Fixes #1728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1713](https://github.com/nix-rust/nix/pull/1713))
 
 ### Fixed
+
+- Enabled `SockaddrStorage::{as_link_addr, as_link_addr_mut}` for Linux-like
+  operating systems.
+  (#[1729](https://github.com/nix-rust/nix/pull/1729))
+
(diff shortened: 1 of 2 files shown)


rtzoeller (approved):
@asomers given that the next minor release of nix will have the aio breaking changes, do you think it's worth backporting this change to a 0.24.2 release? I don't mind doing the legwork.

bors r+

asomers:
I do think that a patch release would be worthwhile.  But there's one other issue that ought to go into it.  I'll go ping the reporter again.

PR #1731, opened 2022-06-04
Allow setting kevent_flags on struct sigevent

asomers:
Depends on https://github.com/rust-lang/libc/pull/2813
Blocks https://github.com/tokio-rs/tokio/issues/4728

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { version = "0.2.147", features = ["extra_traits"] }
+libc = { git = "https://github.com/asomers/libc.git", rev = "d81b318adeae923142c0b5640417cd90509e02ca", features = [ "extra_traits" ] }
 bitflags = "2.3.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/asomers/libc.git", rev = "d81b318adeae923142c0b5640417cd90509e02ca", features = [ "extra_traits" ] }
+libc = { version = "0.2.147", features = ["extra_traits"] }
 bitflags = "2.3.1"
 cfg-if = "1.0"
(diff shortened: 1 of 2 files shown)


asomers:
Back when I originally defined `struct sigevent`, Rust could not represent C unions.  So I cheated and defined it as a struct, replacing the union field with what I thought would be its most important member.  But now I need to use a different union member.  So in the linked libc PR I defined sigevent the correct way, using unions.  This PR both fixes the field names and adds a backwards-compatible way to set the new field.
A search through every crate on crates.io shows that Nix was the only currently-maintained crate using the union-as-struct field.  So I think we can handle the backwards-compatibility issues in libc by merging the libc PR, then quickly releasing patch releases of Nix.

rtzoeller:
> So I think we can handle the backwards-compatibility issues in libc by merging the libc PR, then quickly releasing patch releases of Nix.

This seems reasonable. A few procedural questions:

 - How far back to we want to patch?
 
 	I pulled the download numbers for each recent minor version over the last 14 days:
	| Version | Count |
	|---|---|
	| 0.18 | 18143 |
	| 0.19 | 40880 |
	| 0.20 | 71998 |
	| 0.21 | 21651 |
	| 0.22 | 146464 |
	| 0.23 | 274245 |
	| 0.24 | 160180 |

	IMO this supports patching back to 0.22, and only back to 0.20 by request.

- Should we combine this with the existing plans for a 0.24.2 patch, or address this subsequently as a 0.24.3?

asomers:
I've rebased.  But so far, the libc maintainers still haven't merged the PR there.  Trying to be both backwards-compatible AND Rust 1.13.0 compatible AND satisfy that project's style checker is very difficult.  If they refuse to merge it, I'm thinking about just copying the FFI definitions into Nix, so we won't be blocked anymore.  What do you think of that, @rtzoeller ?

asomers:
Ok, I pushed a version that inlines the C structures into Nix.  But only for FreeBSD, because that's the only platform that really needs this change.

asomers:
Still no word from upstream.  I'll merge this now.  Once the upstream PR merges, we can revert the 2nd commit.

SteveLauC:
Re-link the updated libc PR: https://github.com/rust-lang/libc/pull/3630

Issue #1709, opened 2022-04-26
`SockaddrStorage::as_sockaddr_in` incorrectly returns None sometime 


poliorcetics:
On `nix 0.24.x` the following is necessary to always manage to extract the IP address (and maybe something similar is needed for IPv6, but I have not encountered the problem

```rust
/// Converts a socket addr into an IP address when possible
fn try_into_ip_addr(s: &SockaddrStorage) -> Option<std::net::IpAddr> {
    // Normal method
    let or_else = s
        .as_sockaddr_in()
        .map(|sa| sa.ip().to_be_bytes().into())
        .or_else(|| s.as_sockaddr_in6().map(|sa| sa.ip().into()));

    // Start of the fallback
    match initial_method {
        im @ Some(_) => return im,
        None => (),
    }

    let sock_addr = unsafe { s.as_ptr().as_ref()? };
    if i32::from(sock_addr.sa_family) == nix::libc::AF_INET {
        dbg!(&s);
        let sock_addr = unsafe { *s.as_ptr().cast::<nix::libc::sockaddr_in>() };
        Some(Ipv4Addr::from(sock_addr.sin_addr.s_addr).into())
    } else {
        None
    }
}
```

Maybe something about the size ? 

I'm on macOS 12 but CI also fails for every type of Linux we have, both GNU and MUSL

asomers:
What code did you try first, and what error did you get?

asomers:
Also, is that `IpAddr` a `std::net::IpAddr` or a `nix::sys::socket::IpAddr`?

poliorcetics:
We have an `InterfaceAddress` we originally got from `getifaddrs()`. We try to get the IP of its `netmask` member using `try_into_ip_addr`.

I modified the original code to add a `dgb!` call, I get this in the branch:

```rust
&s = SockaddrStorage {
    ss: sockaddr_storage {
        ss_len: 7,
        ss_family: 2,
        __ss_pad1: [
            0,
            0,
            255,
            255,
            255,
            0,
        ],
        __ss_align: 2882523706792346128,
    },
}
```

`ss_family: 2` is `AF_INET`, which should be caught by `as_sockaddr_in` earlier but is not. Going down in `SockaddrIn::from_raw`, I think this is a size problem, but I have no idea why

asomers:
You said that the original version of `try_into_ip_addr` didn't work.  What was that original version, and what error did you get?

poliorcetics:
[...]

PR #1736, opened 2022-06-07
Fix socket address family checks

qwandor:
The `SockaddrLike::from_raw` implementations for `VsockAddr` and `SysControlAddr` were checking against the wrong address family constant. This PR makes them consistent with the values matched against in `SockaddrStorage::from_raw`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Enabled `SockaddrStorage::{as_link_addr, as_link_addr_mut}` for Linux-like
   operating systems.
   (#[1729](https://github.com/nix-rust/nix/pull/1729))
+- Fixed `SockaddrLike::from_raw` implementations for `VsockAddr` and
+  `SysControlAddr`.
+  (#[1736](https://github.com/nix-rust/nix/pull/1736))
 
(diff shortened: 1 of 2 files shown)


qwandor:
Done.

rtzoeller:
@asomers any idea what's going on with the FreeBSD 14 timeouts? I've noticed that on a few recent PRs. 

I have a fix submitted for Haiku, we're just waiting on a new nightly rustc to be available.

asomers:
No, I saw that too but I haven't had time to investigate yet.

asomers:
It looks like `test_acct` is the culprit.  When I run it locally its runtime varies anywhere from a few seconds to 3m.  It's just a matter of when does the kernel decide to flush the accounting file, and I don't think userland has any control over that.

rtzoeller:
@asomers this seems like another patch candidate.

rtzoeller:
@asomers this seems like another patch candidate.

qwandor:
Hi, any idea when you plan to push a release with this fix included?

asomers:
I think the only remaining issue that we should consider for a patch release is #1709 , if indeed that turns out to be a bug in Nix.

rtzoeller:
@qwandor this should be in a 0.24.2 patch release now.

PR #1737, opened 2022-06-08
Fix the test_acct test

asomers:
It has never actually executed its command, so the only reason that it
ever worked is that on most systems there are usually processes starting
and exiting all the time.


rtzoeller (approved):

rtzoeller:
bors retry

PR #1738, opened 2022-06-08
Add support for loongarch64

zhangwenlong8911:


rtzoeller (changes requested):
It's difficult for the nix maintainers to validate changes for platforms which are not supported by `rustc`, including ensuring they don't immediately regress in support as part of other changes.  I see that there is a [PR](https://github.com/rust-lang/rust/pull/96971) up adding support to `rustc` for `loongarch64-unknown-linux-gnu`; this change is likely gated on that PR and until `loongarch64-unknown-linux-gnu` is supported as at least a Tier 3 target by the Rust compiler.

Once `loongarch64-unknown-linux-gnu` is supported as a Tier 3 target by `rustc`, I'd like to see it added to our CI pipeline as part of this PR, so we can ensure the support doesn't regress. #1603 is a good example of the changes needed. Note that it'd be added to the Tier 3 section of `README.md`, not Tier 1.

zhangwenlong8911:
> It's difficult for the nix maintainers to validate changes for platforms which are not supported by `rustc`, including ensuring they don't immediately regress in support as part of other changes. I see that there is a [PR](https://github.com/rust-lang/rust/pull/96971) up adding support to `rustc` for `loongarch64-unknown-linux-gnu`; this change is likely gated on that PR and until `loongarch64-unknown-linux-gnu` is supported as at least a Tier 3 target by the Rust compiler.
> 
> Once `loongarch64-unknown-linux-gnu` is supported as a Tier 3 target by `rustc`, I'd like to see it added to our CI pipeline as part of this PR, so we can ensure the support doesn't regress. #1603 is a good example of the changes needed. Note that it'd be added to the Tier 3 section of `README.md`, not Tier 1.

thank you very much for your suggestion, I have moved  `loongarch64-unknown-linux-gnu`  to the Tier 3 section of `README.md`

.cirrus.yml
```diff
@@ -288,6 +288,9 @@ task:
     - name: Haiku x86_64
       env:
         TARGET: x86_64-unknown-haiku
+    - name: Loongnix
```
xen0n, line 291: No. Loongnix is old-world but here it's 100% sure this should be running on a new-world distro. This could be just `Linux loongarch64 lp64d`, and any adaptations for Loongnix should be done in your downstream fork specifically for Loongnix, not here.

xen0n (changes requested):
And please just wait until https://github.com/rust-lang/rust/pull/96971 is merged and released, otherwise the code here would be still useless at best, or require further changes if things change at the Rust side.

zhangwenlong8911:
> And please just wait until [rust-lang/rust#96971](https://github.com/rust-lang/rust/pull/96971) is merged and released, otherwise the code here would be still useless at best, or require further changes if things change at the Rust side.

Thans for your review, i just rebase the code, I whill wait until https://github.com/rust-lang/rust/pull/96971 is merged and released

.cirrus.yml
```diff
@@ -288,6 +288,9 @@ task:
     - name: Haiku x86_64
       env:
         TARGET: x86_64-unknown-haiku
+    - name: Linux loongarch64 lp64d
+      env:
+        TARGET: loongarch64-unknown-linux-gnuf64
```
xen0n, line 293: We're back to using `loongarch64-unknown-linux-gnu` in the final iteration of rust-lang/rust#96971, so the tuple should get restored here.

README.md
```diff
@@ -86,6 +86,7 @@ Tier 3:
   * x86_64-unknown-linux-gnux32
   * x86_64-unknown-openbsd
   * x86_64-unknown-redox
+  * loongarch64-unknown-linux-gnuf64
```
xen0n, line 89: We're now *Tier 2 with host tools*, so please move the (corrected) tuple here to the appropriate location too.

xen0n (commented):
Enough time has passed that LoongArch is already pretty well supported by Rust 1.72.0, so this PR could get a rebase?

xen0n:
Ah, so an equivalent change was already merged in #2045. This can be safely closed then.

PR #1741, opened 2022-06-08
SigSet: A new unsafe helper method to create a SigSet from a sigset_t

germag:
Currently,  the only way to create a `SigSet` from a `sigset_t` object
is by using pointer casts, like:

```
unsafe {
    let sigset = *(&sigset as *const libc::sigset_t as *const SigSet)
};
```

This is un-ergonomic for library creators with interfaces to C.
So, let's add a new unsafe method that creates a `SigSet` from a 
`libc::sigset_t` object.

We can't implement `From` since converting from `libc::sigset_t` to
`SigSet` is unsafe, because objects of type `libc::sigset_t` must be
initialized by calling either `sigemptyset(3)` or `sigfillset(3)`
before being used. In other case, the results are undefined.
We can't implement `TryFrom` either, because there is no way to check
if an object of type `libc::sigset_t` is initialized.

Signed-off-by: German Maglione <gmaglione@redhat.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1747](https://github.com/nix-rust/nix/pull/1747))
 - Added the `DontRoute` SockOpt
   (#[1752](https://github.com/nix-rust/nix/pull/1752))
+- Added `signal::SigSet::from_sigset_t_unchecked()`.
+  (#[1741](https://github.com/nix-rust/nix/pull/1741))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1739](https://github.com/nix-rust/nix/pull/1739))
 - Changed `gethostname` to return an owned `OsString`.
   (#[1745](https://github.com/nix-rust/nix/pull/1745))
+- `signal:SigSet` is now marked as `repr(transparent)`.
+  (#[1741](https://github.com/nix-rust/nix/pull/1741))
 
 ### Fixed
 
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -465,6 +465,9 @@ use std::iter::FromIterator;
 use std::iter::IntoIterator;
 
 /// Specifies a set of [`Signal`]s that may be blocked, waited for, etc.
+// We are using `transparent` here to be super sure that `SigSet`
+// is represented exactly like the `sigset_t` struct from C.
+#[repr(transparent)]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct SigSet {
     sigset: libc::sigset_t
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -567,6 +567,19 @@ impl SigSet {
             Signal::try_from(signum.assume_init()).unwrap()
         })
     }
+
+    /// Converts a `libc::sigset_t` object to a [`SigSet`] without checking  whether the
+    /// `libc::sigset_t` is already initialized.
+    ///
+    /// # Safety
+    ///
+    /// The `sigset` passed in must be a valid an initialized `libc::sigset_t` by calling either
+    /// [`sigemptyset(3)`](https://man7.org/linux/man-pages/man3/sigemptyset.3p.html) or
+    /// [`sigfillset(3)`](https://man7.org/linux/man-pages/man3/sigfillset.3p.html).
+    /// Otherwise, the results are undefined.
+    pub unsafe fn from_sigset_t_unchecked(sigset: libc::sigset_t) -> SigSet {
+        SigSet { sigset }
+    }
 }
 
 impl AsRef<libc::sigset_t> for SigSet {
@@ -1310,5 +1323,22 @@ mod tests {
         })
         .join()
         .unwrap();
+    }
+
+    #[test]
+    fn test_from_sigset_t_unchecked() {
+        let src_set = SigSet::empty();
+        let set = unsafe { SigSet::from_sigset_t_unchecked(src_set.sigset) };
+
+        for signal in Signal::iterator() {
+            assert!(!set.contains(signal));
+        }
+
+        let src_set = SigSet::all();
+        let set = unsafe { SigSet::from_sigset_t_unchecked(src_set.sigset) };
+
+        for signal in Signal::iterator() {
+            assert!(set.contains(signal));
+        }
     }
 }


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1703](https://github.com/nix-rust/nix/pull/1703))
 - Added `ptrace::read_user` and `ptrace::write_user` for Linux.
   (#[1697](https://github.com/nix-rust/nix/pull/1697))
+- Added `from_sigset_t_unchecked()` to `signal::SigSet`.
```
rtzoeller, line 22: Can you add a link to this PR, similar to other items (squashed into the current commit)?

src/sys/signal.rs
```diff
@@ -619,6 +619,13 @@ impl SigSet {
             Signal::try_from(signum.assume_init()).unwrap()
         })
     }
+
+    /// Converts a sigset_t object to a SigSet without checking  whether the sigset_t is already
```
rtzoeller, line 623: Can you add links to the functions referenced in this documentation, or at least surround them with backticks so they render as code with `cargo doc`?

Thanks!

germag:
v2:
- Fix links and ticks in comments
- Add link to PR in Changelog
- Fix typo and ticks in commit message

src/sys/signal.rs
```diff
@@ -619,6 +619,15 @@ impl SigSet {
             Signal::try_from(signum.assume_init()).unwrap()
         })
     }
+
+    /// Converts a `libc::sigset_t` object to a [`SigSet`] without checking  whether the `libc::sigset_t`
+    /// is already initialized. Objects of type`libc::sigset_t` must be initialized by calling either
+    /// [`sigemptyset(3)`](https://man7.org/linux/man-pages/man3/sigemptyset.3p.html) or
+    /// [`sigfillset(3)`](https://man7.org/linux/man-pages/man3/sigfillset.3p.html)
+    /// before being used. In other case, the results are undefined.
+    pub unsafe fn from_sigset_t_unchecked(sigset: libc::sigset_t) -> SigSet {
```
asomers, line 628: It's good practice to have a `# Safety` section for every unsafe pub fun.  Could you please add one?  Also, why does this need to be `unsafe`?  `libc::sigset` doesn't contain any pointers, so it's not obvious what invariant might be violated.

asomers (changes requested):
The FreeBSD test failure should be fixed by rebasing.

germag:
v3:
- Add `# Sefety` section
- Rebase

src/sys/signal.rs
germag: > Also, why does this need to be `unsafe`? `libc::sigset` doesn't contain any pointers, so it's not obvious what invariant might be violated.

Good point, it's not related to memory safety or ownership. I marked it as `unsafe`, because In theory it should not be assumed the implementation of signal set, and  the results are undefined if the `sigset_t` it's uninitialized according to posix. 

But, it's also true that to get an uninitialized `libc::sigset_t` it is necessary to use `unsafe`, so perhaps it's better to implement `From`, which is much more convenient than `SigSet::from_sigset_t_unchecked()`. @asomers what do you think, should I change this PR and implement `From` instead?

germag: btw, `UContext::sigmask()` and `UContext::sigmask_mut()` both use the pointer cast I mention in the commit message, but I just realize the `SigSet` is not `#[repr(transparent)]`

rtzoeller: I think leaving it marked as `unsafe` is helpful, as it helps alert the user about a potential footgun in the future regarding ABI compatibility. Here's what POSIX says:

> The implementation of the sigemptyset() (or [sigfillset()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigfillset.html)) function could quite trivially clear (or set) all the bits in the signal set. Alternatively, it would be reasonable to initialize part of the structure, such as a version field, to permit binary-compatibility between releases where the size of the set varies. For such reasons, either sigemptyset() or [sigfillset()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigfillset.html) must be called prior to any other use of the signal set, even if such use is read-only (for example, as an argument to [sigpending()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigpending.html)).

---

> I just realize the SigSet is not #[repr(transparent)]

Can you include this change as part of the PR? I see no reason it can't be marked transparent.

Thanks!

germag:
v4:
- Add `repr(transparent)`attribute to `SigSet` struct

germag:
I know it's unrelated to this PR, but `SigSet` derives `Eq`, I assume because `libc::sigset_t` does it too, but at least in linux, `sigfillset(3)` leaves bits uninitialized (IIRC  it only initializes the first 64 bits), so two `SigSet`s can be different even though they have the same set of signals

src/sys/signal.rs
asomers: If `SigSet` must have the invariant "the inner `sigset` must be initialized with either `sigemptyset` or `sigfillset`", then I suggest eliminating this function entirely, and replacing it with a pair of safe `SigSet::empty() -> Self` and `SigSet::full() -> Self` functions instead.

asomers:
> I know it's unrelated to this PR, but `SigSet` derives `Eq` and `PartialEq`, I assume because `libc::sigset_t` does it too, but at least in linux, `sigfillset(3)` leaves bits uninitialized (IIRC it only initializes the first 64 bits), so two `SigSet`s can be different even though they have the same set of signals

Ooh, good catch.  Could you open an issue for this?

src/sys/signal.rs
rtzoeller: `SigSet::empty` and `SigSet::all` already exist. I believe the intent is to accept a value which was initialized with `sigemptyset` or `sigfillset` and then modified, likely all external to Rust code.

germag: > `SigSet::empty` and `SigSet::all` already exist. I believe the intent is to accept a value which was initialized with `sigemptyset` or `sigfillset` and then modified, likely all external to Rust code.

Yes exactly, our use case is in libraries written in rust that will be used from C

germag:
> Ooh, good catch. Could you open an issue for this?

Done

germag:
it has been suggested to me that `new_unchecked()` may be a better name than `from_sigset_t_unchecked()`,  WDYT?

CHANGELOG.md
```diff
@@ -32,6 +34,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1713](https://github.com/nix-rust/nix/pull/1713))
 - `nix::poll::ppoll`: `sigmask` parameter is now optional.
   (#[1739](https://github.com/nix-rust/nix/pull/1739))
+- `signal:SigSet` it's now marked as `repr(trasparent)`.
```
rtzoeller, line 37: There's a minor typo or two here, but we can clean it up in a later commit before the next release. No reason for this to stay open.

rtzoeller (approved):
bors r+

rtzoeller:
@germag can you update the commit adding the test to fix the formatting issue, and also fix the typos in the changelog (e.g. in transparent), since we're updating things anyways?

germag:
v5:
- Remove empty line
- Fix Changelog typos
- Rebase

germag:
> @germag can you update the commit adding the test to fix the formatting issue, and also fix the typos in the changelog (e.g. in transparent), since we're updating things anyways?

Done, thanks

rtzoeller (approved):
bors r+

PR #1744, opened 2022-06-13
reimplement sendmmsg/recvmmsg with better API

### LINKED ISSUE #1602: recvmmsg is strange
`recvmmsg` takes `data` which is almost but not quite a mutable array reference to prepared buffers and returns a vector of `RecvMsg<'a>` which contains a reference to `data`, at least for as long as lifetimes are concerned. At this moment information about received packets is contained in two places: whatever is used to create `data` - payload we got and a vector that `recvmmsg` returns. To do anything with a packet we need both - payload from `data` and size of payload from the vector, but as long as lifetimes are concerned we can't do anything to `data` until we release the reference to the vector.

Even the test case seem to cheat here by size of the data sent rather than data received and changing it to do something similar to this
```rust
         for (buf, r) in msgs.iter().zip(res.iter()) {
                ....
          } 
```
results in
```
cannot borrow `msgs` as immutable because it is also borrowed as mutable
```

And it seems like the only way to actually use it is to extract anything from the resulting vector first (by doing allocations and not holding any references to it), drop it and only then process the actual payload - more allocations, bigger and slower code which you are probably trying to avoid by using `recvmmsg` in the first place.

So questions:

1. Is there anything that prevents recvmmsg from accepting an array (or a slice) and producing an iterator with references to this slice to avoid allocations?
2. Should produced iterator (or vector) also contain references to payload buffers?

pacak:
Motivation is explained in #1602, new version allows to receive data without performing allocations inside the receive loop and to use received data without extra copying.

This pull request contains a breaking change to API `recvmmsg` (obviously) and also affects `recvmsg` - new version does not set length of control message buffer if one is passed. Later change can be avoided with a bit more copy-paste.

Fixes #1602

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1,21 +1,23 @@
 //! Socket interface functions
 //!
 //! [Further reading](https://man7.org/linux/man-pages/man7/socket.7.html)
-use cfg_if::cfg_if;
-use crate::{Result, errno::Errno};
-use libc::{self, c_void, c_int, iovec, socklen_t, size_t,
-        CMSG_FIRSTHDR, CMSG_NXTHDR, CMSG_DATA, CMSG_LEN};
-use std::convert::TryInto;
-use std::{mem, ptr, slice};
-use std::os::unix::io::RawFd;
-#[cfg(feature = "net")]
-use std::net;
 #[cfg(target_os = "linux")]
 #[cfg(feature = "uio")]
 use crate::sys::time::TimeSpec;
 #[cfg(feature = "uio")]
 use crate::sys::time::TimeVal;
+use crate::{errno::Errno, Result};
+use cfg_if::cfg_if;
+use libc::{
+    self, c_int, c_void, iovec, size_t, socklen_t, CMSG_DATA, CMSG_FIRSTHDR,
+    CMSG_LEN, CMSG_NXTHDR,
+};
+use std::convert::TryInto;
 use std::io::{IoSlice, IoSliceMut};
+#[cfg(feature = "net")]
+use std::net;
+use std::os::unix::io::RawFd;
+use std::{mem, ptr, slice};
 
 #[deny(missing_docs)]
 mod addr;
@@ -32,60 +34,44 @@ pub use self::addr::{SockaddrLike, SockaddrStorage};
 
 #[cfg(not(any(target_os = "illumos", target_os = "solaris")))]
 #[allow(deprecated)]
-pub use self::addr::{
-    AddressFamily,
-    SockAddr,
-    UnixAddr,
-};
-#[allow(deprecated)]
-#[cfg(not(any(target_os = "illumos", target_os = "solaris", target_os = "haiku")))]
-#[cfg(feature = "net")]
-pub use self::addr::{
-    InetAddr,
-    IpAddr,
-    Ipv4Addr,
-    Ipv6Addr,
-    LinkAddr,
-    SockaddrIn,
-    SockaddrIn6
-};
+pub use self::addr::{AddressFamily, SockAddr, UnixAddr};
 #[cfg(any(target_os = "illumos", target_os = "solaris"))]
 #[allow(deprecated)]
-pub use self::addr::{
-    AddressFamily,
-    SockAddr,
-    UnixAddr,
-};
+pub use self::addr::{AddressFamily, SockAddr, UnixAddr};
 #[allow(deprecated)]
-#[cfg(any(target_os = "illumos", target_os = "solaris", target_os = "haiku"))]
+#[cfg(not(any(
+    target_os = "illumos",
+    target_os = "solaris",
+    target_os = "haiku"
+)))]
 #[cfg(feature = "net")]
 pub use self::addr::{
-    InetAddr,
-    IpAddr,
-    Ipv4Addr,
-    Ipv6Addr,
-    SockaddrIn,
-    SockaddrIn6
+    InetAddr, IpAddr, Ipv4Addr, Ipv6Addr, LinkAddr, SockaddrIn, SockaddrIn6,
+};
+#[allow(deprecated)]
+#[cfg(any(
+    target_os = "illumos",
+    target_os = "solaris",
+    target_os = "haiku"
+))]
+#[cfg(feature = "net")]
+pub use self::addr::{
+    InetAddr, IpAddr, Ipv4Addr, Ipv6Addr, SockaddrIn, SockaddrIn6,
 };
 
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub use crate::sys::socket::addr::alg::AlgAddr;
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub use crate::sys::socket::addr::netlink::NetlinkAddr;
 #[cfg(any(target_os = "ios", target_os = "macos"))]
 #[cfg(feature = "ioctl")]
 pub use crate::sys::socket::addr::sys_control::SysControlAddr;
-#[cfg(any(target_os = "android", target_os = "linux"))]
-pub use crate::sys::socket::addr::netlink::NetlinkAddr;
-#[cfg(any(target_os = "android", target_os = "linux"))]
-pub use crate::sys::socket::addr::alg::AlgAddr;
 #[cfg(any(target_os = "android", target_os = "linux"))]
 pub use crate::sys::socket::addr::vsock::VsockAddr;
 
 #[cfg(feature = "uio")]
 pub use libc::{cmsghdr, msghdr};
-pub use libc::{
-    sa_family_t,
-    sockaddr,
-    sockaddr_storage,
-    sockaddr_un,
-};
+pub use libc::{sa_family_t, sockaddr, sockaddr_storage, sockaddr_un};
 #[cfg(feature = "net")]
 pub use libc::{sockaddr_in, sockaddr_in6};
 
@@ -245,7 +231,7 @@ libc_bitflags! {
     }
 }
 
-libc_bitflags!{
+libc_bitflags! {
     /// Additional socket options
     pub struct SockFlag: c_int {
         /// Set non-blocking mode on the new socket
@@ -280,7 +266,7 @@ libc_bitflags!{
     }
 }
 
-libc_bitflags!{
+libc_bitflags! {
     /// Flags for send/recv and their relatives
     pub struct MsgFlags: c_int {
         /// Sends or requests out-of-band data on sockets that support this notion
@@ -462,7 +448,7 @@ cfg_if! {
     }
 }
 
-cfg_if!{
+cfg_if! {
     if #[cfg(any(
                 target_os = "dragonfly",
                 target_os = "freebsd",
@@ -1645,7 +1631,7 @@ pub fn recvmmsg<'a, I, S>(
             }
         );
 
-        (msg_controllen, &mut d.cmsg_buffer)
+        msg_controllen as usize
     }).collect();
 
     let timeout = if let Some(mut t) = timeout {
@@ -1663,40 +1649,31 @@ pub fn recvmmsg<'a, I, S>(
         .take(ret as usize)
         .zip(addresses.iter().map(|addr| unsafe{addr.assume_init()}))
         .zip(results.into_iter())
-        .map(|((mmsghdr, address), (msg_controllen, cmsg_buffer))| {
-            // The cast is not unnecessary on all platforms.
-            #[allow(clippy::unnecessary_cast)]
+        .map(|((mmsghdr, address), msg_controllen)| {
             unsafe {
                 read_mhdr(
                     mmsghdr.msg_hdr,
                     mmsghdr.msg_len as isize,
                     msg_controllen,
                     address,
-                    cmsg_buffer
                 )
             }
         })
         .collect())
 }
 
-unsafe fn read_mhdr<'a, 'b, S>(
+unsafe fn read_mhdr<'a, S>(
     mhdr: msghdr,
     r: isize,
     msg_controllen: usize,
     address: S,
-    cmsg_buffer: &'a mut Option<&'b mut Vec<u8>>
-) -> RecvMsg<'b, S>
+) -> RecvMsg<'a, S>
     where S: SockaddrLike
 {
     // The cast is not unnecessary on all platforms.
     #[allow(clippy::unnecessary_cast)]
     let cmsghdr = {
         if mhdr.msg_controllen > 0 {
-            // got control message(s)
-            cmsg_buffer
-                .as_mut()
-                .unwrap()
-                .set_len(mhdr.msg_controllen as usize);
             debug_assert!(!mhdr.msg_control.is_null());
             debug_assert!(msg_controllen >= mhdr.msg_controllen as usize);
             CMSG_FIRSTHDR(&mhdr as *const msghdr)
@@ -1827,10 +1804,9 @@ pub fn recvmsg<'a, 'outer, 'inner, S>(fd: RawFd, iov: &'outer mut [IoSliceMut<'i
 
     let r = Errno::result(ret)?;
 
-    Ok(unsafe { read_mhdr(mhdr, r, msg_controllen, address.assume_init(), &mut cmsg_buffer) })
+    Ok(unsafe { read_mhdr(mhdr, r, msg_controllen, address.assume_init()) })
 }
 }
-
 
 /// Create an endpoint for communication
 ///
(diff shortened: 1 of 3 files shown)


rtzoeller (commented):
Thanks @pacak! Overall the changes look very reasonable, although I've not done an entirely thorough review just yet.

A few requests:
 - IMHO it makes sense to make a similar change for `sendmmsg` in the same release as this PR, to avoid splitting up the breaking changes into multiple releases (thus prolonging any associated pains). Would you be willing to make a similar change for `sendmmsg`?
 - Do you have an example of retrofitting an existing caller to use the new API you can share? In places where we make nontrivial API changes, I'd like to maintain a list of example changes to help other clients migrate.

pacak:
> Would you be willing to make a similar change for `sendmmsg`?

I can but it would take some time. It would be great if https://github.com/nix-rust/nix/pull/1748 was merged before that. Currently making any changes to `socket/mod.rs` is unnecessary painful with the editor configured to automatically format everything.

> Do you have an example of retrofitting an existing caller to use the new API you can share?

Sort of. I updated the tests in `test/sys/test_socket.rs` to use new API, nothing crazy. Just one extra parameters allocated in advance, passed into `recvmmsg` and implicit collect is replaced with explicit one. Other than that I can't really imagine how would you use existing API. https://github.com/nix-rust/nix/issues/1602 contains the explanation of what's wrong with the old API and it also contains some "me too" messages, plus a few more of "me too" got removed.

rtzoeller:
> I can but it would take some time. It would be great if https://github.com/nix-rust/nix/pull/1748 was merged before that. Currently making any changes to socket/mod.rs is unnecessary painful with the editor configured to automatically format everything.

Fair enough. The alternative would be to just disable the auto-formatting, but I agree that's annoying to do for one repo (when the proper fix is apparent). I've been reluctant to submit any bulk reformatting changes up until now to avoid causing merge conflicts for all outstanding PRs, but I think we've reached a cross-over point where most PRs already have merge conflicts anyways so we should just do it. @asomers thoughts?

asomers:
Yeah, I think it's time to do it.  I for one don't have any big outstanding PRs.

pacak:
> Would you be willing to make a similar change for `sendmmsg`?

Ended up implementing it. Tests seem to work. Don't have use cases in real code so can't test it properly.

rtzoeller:
@pacak can you rebase and reformat? Thanks!

pacak:
Seems to work.

rroohhh:
Something seems to be wrong here. This testcase: https://gist.github.com/rroohhh/acb6247f7634e4e68d53604c4c3f88b1
fails in debug mode with 
```
  left: `[207]`,
 right: `[170]`', src/bin/testcase.rs:52:9
```
and in release mode with
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: EFAULT', src/bin/testcase.rs:38:6
```

pacak:
> Something seems to be wrong here. This testcase: https://gist.github.com/rroohhh/acb6247f7634e4e68d53604c4c3f88b1

Interesting. Moving buffers allocation out from lines 33 and 45 fixes the problem for me, looks like something holds an invalid pointer in both cases... Thank you for a great test case, I'll see how to fix the problem tomorrow.

pacak:
A good test case.

A temporary array created on lines 33/45 used to be stored  inside a pointer in `MultHdrs` structure, I fixed it by requiring a reference with appropriate lifetime, now the test case fails to compile.

rtzoeller:
Looks like nightly clippy introduced some new warnings. I'll look at fixing them in the next day or two.

rroohhh:
@pacak thank you for the quick fix

rtzoeller:
@pacak I tried implementing these changes for the `tokio-udt` crate, but ran into issues regarding `MultHdrs<_>` not implementing `Send`:

> help: within `[mmsghdr]`, the trait `Send` is not implemented for `*mut iovec`

Can you try getting `tokio-udt` building with your changes, and see if anything else is needed?

src/sys/socket/mod.rs
```diff
@@ -1536,134 +1538,338 @@ pub fn sendmmsg<'a, I, C, S>(
...
+    msg_controllen: usize,
+}
+
+#[cfg(any(
+    target_os = "linux",
+    target_os = "android",
+    target_os = "freebsd",
+    target_os = "netbsd",
+))]
+impl<S> MultHdrs<S> {
```
rtzoeller, line 1559: For consistency with `MultiResults`, should this be `MultiHeaders`?

amatissart:
> @pacak I tried implementing these changes for the `tokio-udt` crate, but ran into issues regarding `MultHdrs<_>` not implementing `Send`:
> 
> > help: within `[mmsghdr]`, the trait `Send` is not implemented for `*mut iovec`
> 
> Can you try getting `tokio-udt` building with your changes, and see if anything else is needed?

Thanks for the ping, and sorry for the late response.

I made it compile on this branch: https://github.com/amatissart/tokio-udt/compare/nix-mmsg-upgrade
It seems that `MultHdrs<_>` doesn't need to implement `Send` (in this use case at least), as the headers have to be preallocated before each call to `recvmmsg` anyway. Do I understand that correctly?

Note that "tokio-udt" is still very experimental. My usage of sendmmsg/recvmmsg might not be completely relevant in the first place.

pacak:
> as the headers have to be preallocated before each call to `recvmmsg` anyway. Do I understand that correctly?

The idea is that you preallocate them once then keep reusing it between calls. Preallocating them before each call gives you the behavior similar to the old one. If you want to use it with async you can either preallocate before each call or try to share the same set of buffers using mutex.

amatissart:
> If you want to use it with async you can either preallocate before each call or try to share the same set of buffers using mutex.

Ok, that makes sense. But as pointed by @rtzoeller, using a `Mutex<>` to reuse the headers does not seem to be possible, as `MultHdrs<_>` is not `Send`.

src/sys/socket/mod.rs
pacak: Sure, fixed.

pacak:
I'm not sure if preallocated headers can be made send - types from libc are not and the whole idea for this change is to avoid extra allocations between calls - if you care about making less system calls you want to avoid allocations as well. If you really want to use it from multiple threads you can stick to allocating and immediately using them, after this change you can consume payload from iterator instead of a vector.

rtzoeller (approved):
bors r+

pacak:
> [Linux aarch64](https://github.com/nix-rust/nix/runs/7800429661)

Hmm... Is it not running in qemu? The test that is fails is one that checks the timestamps added by my previous pull request. I though it might not work when it's running in qemu. is aarch64 specific in some other way?

rtzoeller:
> > [Linux aarch64](https://github.com/nix-rust/nix/runs/7800429661)
> 
> Hmm... Is it not running in qemu? The test that is fails is one that checks the timestamps added by my previous pull request. I though it might not work when it's running in qemu. is aarch64 specific in some other way?

It is not running in qemu; Cirrus has a native ARM container.

https://github.com/nix-rust/nix/blob/5252f7e1bd1361b47a655ae46a02c32d604483ee/.cirrus.yml#L145-L153

rtzoeller:
@pacak curiously, it passed on your original commit pre-bors. I'm going have bors try again (but not submit), to get a feel for how flaky this is. I'm not familiar with any aarch64 quirks here.

bors try

rtzoeller:
@asomers any ideas on this flaky test?

asomers:
No, I don't have any quick guesses.

rtzoeller:
@pacak I dug out an old Raspberry Pi 3 and confirm that this failure reproduces on real hardware. I'm not sure why yet, but I'll try to look into it over the next few days.

pacak:
> I dug out an old Raspberry Pi 3 and confirm that this failure reproduces on real hardware. I'm not sure why yet, but I'll try to look into it over the next few days.

Thank you for the update, let me know if I can help you with anything.

rtzoeller:
@pacak a few observations; I've not had too much time to dig in:

1. I can bang on `test_recvmm2` indefinitely on Linux AMD64 with no issues.
2. Anecdotally, there appears to be a "clumpiness" to the failures. Run in a loop until failure, I might see the test fail on the first iteration on three or four attempts in a row, and then on the next run take dozens of iterations to fail. Not sure what to make of this - perhaps an alignment issue or something similar where we get "lucky" frequently, in a streaky way?
3. Here are the values of `recv` and `rmsg` for the passing and failing cases on Linux aarch64. Perhaps you can make something out of them - it seems like we might be getting back data but not the control message?

<details>
  <summary>Passing</summary>

```
[src/sys/socket/mod.rs:1827] &recv = MultiResults {
    rmm: MultiHeaders {
        items: [
            mmsghdr {
                msg_hdr: msghdr {
                    msg_name: 0x0000000000000000,
                    msg_namelen: 0,
                    msg_iov: 0x0000ffffa4001290,
                    msg_iovlen: 2,
                    msg_control: 0x0000ffffa40013f0,
                    msg_controllen: 64,
                    msg_flags: 0,
                },
                msg_len: 400,
            },
            mmsghdr {
                msg_hdr: msghdr {
                    msg_name: 0x0000000000000000,
                    msg_namelen: 0,
                    msg_iov: 0x0000ffffa4001350,
                    msg_iovlen: 2,
                    msg_control: 0x0000ffffa4001430,
                    msg_controllen: 64,
                    msg_flags: 0,
                },
                msg_len: 0,
            },
        ],
        addresses: [
            core::mem::maybe_uninit::MaybeUninit<()>,
            core::mem::maybe_uninit::MaybeUninit<()>,
        ],
        cmsg_buffers: Some(
            [
                64,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                1,
                0,
                0,
                0,
                37,
                0,
                0,
                0,
                11,
                73,
                4,
                99,
                0,
                0,
                0,
                0,
                84,
                54,
                93,
                30,
                0,
                0,
				<zeroes>,
                0,
                0,
            ],
        ),
        msg_controllen: 64,
    },
    current_index: 0,
    received: 1,
}
[src/sys/socket/mod.rs:1832] &rmsg = RecvMsg {
    bytes: 400,
    cmsghdr: Some(
        cmsghdr {
            cmsg_len: 64,
            cmsg_level: 1,
            cmsg_type: 37,
        },
    ),
    address: Some(
        (),
    ),
    flags: (empty),
    iobufs: PhantomData,
    mhdr: msghdr {
        msg_name: 0x0000000000000000,
        msg_namelen: 0,
        msg_iov: 0x0000ffffa4001290,
        msg_iovlen: 2,
        msg_control: 0x0000ffffa40013f0,
        msg_controllen: 64,
        msg_flags: 0,
    },
}
```

</details>


<details>
  <summary>Failing</summary>

```
[src/sys/socket/mod.rs:1827] &recv = MultiResults {
    rmm: MultiHeaders {
        items: [
            mmsghdr {
                msg_hdr: msghdr {
                    msg_name: 0x0000000000000000,
                    msg_namelen: 0,
                    msg_iov: 0x0000ffffb4001290,
                    msg_iovlen: 2,
                    msg_control: 0x0000ffffb40013f0,
                    msg_controllen: 0,
                    msg_flags: 0,
                },
                msg_len: 400,
            },
            mmsghdr {
                msg_hdr: msghdr {
                    msg_name: 0x0000000000000000,
                    msg_namelen: 0,
                    msg_iov: 0x0000ffffb4001350,
                    msg_iovlen: 2,
                    msg_control: 0x0000ffffb4001430,
                    msg_controllen: 64,
                    msg_flags: 0,
                },
                msg_len: 0,
            },
        ],
        addresses: [
            core::mem::maybe_uninit::MaybeUninit<()>,
            core::mem::maybe_uninit::MaybeUninit<()>,
        ],
        cmsg_buffers: Some(
            [
                0,
                0,
                <zeroes>,
                0,
                0,
            ],
        ),
        msg_controllen: 64,
    },
    current_index: 0,
    received: 1,
}
[src/sys/socket/mod.rs:1832] &rmsg = RecvMsg {
    bytes: 400,
    cmsghdr: None,
    address: Some(
        (),
    ),
    flags: (empty),
    iobufs: PhantomData,
    mhdr: msghdr {
        msg_name: 0x0000000000000000,
        msg_namelen: 0,
        msg_iov: 0x0000ffffb4001290,
        msg_iovlen: 2,
        msg_control: 0x0000ffffb40013f0,
        msg_controllen: 0,
        msg_flags: 0,
    },
}
```

</details>

pacak:
Yea, looks like we are getting the data but not the control header.

Things to try in no particular order:
1. Try to change the test to transmit a bunch of messages at once and see if it fails in the first iteration or it can fail in the non first one - where it should perform no allocations.
2. Try to backport this test to the old API in some way. It's a new test I added, maybe there's a problem with recvmmsg on aarch64

pacak:
Rebased the branch, backported the test in https://github.com/nix-rust/nix/pull/1837

pacak:
@rtzoeller Can you try running both this and backported tests on your Pi?

rtzoeller:
bors try

pacak:
```console
ubuntu@ip-10-90-3-99:~/nix$ cat /proc/cpuinfo  | head -8
processor       : 0
BogoMIPS        : 243.75
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer : 0x41
CPU architecture: 8
CPU variant     : 0x3
CPU part        : 0xd0c
CPU revision    : 1
```

```console
ubuntu@ip-10-90-3-99:~/nix$ lsb_release  -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
```

```console
ubuntu@ip-10-90-3-99:~/nix$ rustc -vV
rustc 1.56.1 (59eed8a2a 2021-11-01)
binary: rustc
commit-hash: 59eed8a2aac0230a8b53e89d4e99d55912ba6b35
commit-date: 2021-11-01
host: aarch64-unknown-linux-gnu
release: 1.56.1
LLVM version: 13.0.0
```

```console
ubuntu@ip-10-90-3-99:~/nix$ for i in `seq 1 10000` ;do cargo test test_recvmm2 --lib ; done 2> /dev/null > a
ubuntu@ip-10-90-3-99:~/nix$ cat a | sort | uniq -c
  30000 
  10000 running 1 test
  10000 test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 60 filtered out; finished in 0.00s
  10000 test sys::socket::test::test_recvmm2 ... ok
```

```
thread 'sys::test_socket::test_af_alg_cipher' panicked at 'bind failed: ENOENT', test/sys/test_socket.rs:861:27
```

Something tells me the problem is in the image or container CI uses.

rtzoeller:
> Something tells me the problem is in the image or container CI uses.

I would agree, except it is reproducing on the physical hardware in front me of me as well. 😭

---

Tired of being the bottleneck here, I spent some additional time on this today with the intent of finding a root cause. I had the idea today of using `strace` to look for differences in syscalls in the runs that pass and the runs that fail.

The only problem with that strategy is that using `strace` causes the runs to always succeed - I can reliably reproduce the failure by invoking the command directly (with `--test-threads=1`), but running it under e.g. `strace -e trace=recvmmsg` leads to the test always passing...

Which leads me to conclude that there is probably some sort of race condition going on, but that it's probably outside the scope of this code (i.e. in the kernel) and that it's only hitting very slow devices. It reproduces on the CI, and on my RPi3 with both the 5.17 and 5.19 kernels on Fedora 36, but I wouldn't expect either to be particularly quick. Alternatively, there is a bug in this code, but it's sufficiently subtle that we are unlikely to characterize it here without additional reports from other users.

For good measure, I confirmed that the bug also reproduces on my hardware under the `aarch64-unknown-linux-musl` target but fails to reproduce under `strace` (as expected).

@pacak **thank you for your effort and patience on this**. Can you disable `test_recvmm2` on Linux aarch64, clean up the git history (either all squashed or combining the format + clippy fixes into the previous commits, whichever you prefer) and then I'll submit this?

pacak:
> I would agree, except it is reproducing on the physical hardware in front me of me as well.

It could be related to the kernel version, I used Ubuntu 22.04 - latest LTS which uses 5.15. Also not shown in what I posted but I did the same test on 4 different instances - starting from the fast 64 core bare metal ones and down to slow dual core virtual one.

> Can you disable test_recvmm2 on Linux aarch64, clean up the git history

Will do. The test itself mostly works on aarch64, I'll just disable the assert about the timestamps on that platform.

Do you think it's worth mentioning a potential problem on aarch64 in the docs? Or at least in comments? Or the conditions are rare enough so it's unlikely? It needs to be a combination of aarch64,  slow hardware,  `recvmmsg` and an attempt to collect kernel timestamps - later to you usually do when you care about the best performance and accurate measurements only with a workaround being to use `recvmsg`...

rtzoeller:
> Do you think it's worth mentioning a potential problem on aarch64 in the docs? Or at least in comments?

A comment would be a good idea - I don't think we need to call it out in the public docs.

pacak:
Added a comment. Also if we are talking race conditions - backported tests might be passing on the old version of `recvmmsg` because it's less efficient and performs multiple allocations between calls.

pacak:
Hmmm.... And we can test this theory by adding a sleep or some redundant allocations in the new test between calling recvmmsg and using the results - line 1847 or so. Well, you can because I can't reproduce the problem on my side.

PR #1745, opened 2022-06-15
Change gethostname to use a buffer of MaybeUninit values

nathaniel-daniel:
Changing `gethostname` to accept a buffer of `MaybeUninit` bytes allows the user to avoid needlessly initializing a buffer. This is a breaking API change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1713](https://github.com/nix-rust/nix/pull/1713))
 - `nix::poll::ppoll`: `sigmask` parameter is now optional.
   (#[1739](https://github.com/nix-rust/nix/pull/1739))
+- Changed `gethostname` to use a buffer of `MaybeUninit` values.
+  (#[1745](https://github.com/nix-rust/nix/pull/1745))
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


asomers:
Certainly an improvement.  But wouldn't it be even better if it just returned an owned string, like what `ptsname` does?

nathaniel-daniel:
I actually am using this for in a wrapper that returns an owned string. I wasn't sure if changing this to only work for owned strings would be an improvement, as its technically possible to take the user-created buffer and unsafely make it owned with the length data from the returned c-string. This method also allows users to use buffers  they may already have.

Making this wrapper work only for owned strings would be the most convenient for me, but it would break use cases where the user already has a buffer. I wasn't sure if such a change would be accepted, so I opted to make the simplest, least controversial change. I would be willing to change my PR so that it only works for owned buffers if that is deemed important.

asomers:
I think owned strings would be most convenient for everybody.  It wouldn't be the most performant option, but how many people rely on maximum performance for `gethostname` ?

nathaniel-daniel:
Fair enough. I've updated the PR to return an `OsString`. I'm fairly certain hostnames need to be ascii, but I wasn't able to find a solid source for that. Using an `OsString` also mirrors `sethostname`.

nathaniel-daniel:
CI failures seem to be due to `cross v0.2.2` now requiring edition 2021.

rtzoeller:
@nathaniel-daniel it should be fixed if you rebase!

nathaniel-daniel:
Rebased

asomers (approved):
bors r+

PR #1746, opened 2022-06-18
Fix typo and minimise the use of `unsafe` blocks inside the `pipe` function

costinsin:
Some of the operations inside the pipe function are safe and should not be included inside an unsafe block.


rtzoeller (approved):
Thanks!

bors r+

PR #1747, opened 2022-06-19
Add getrusage wrapper

kov:
Includes an enum to specify what to get resource usage for, and a new
struct that provides a more readable view into libc::rusage, including
using TimeVal for user and system CPU time.

Signed-off-by: Gustavo Noronha Silva <gustavo@noronha.dev.br>

### DIFF
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -1,7 +1,9 @@
 //! Configure the process resource limits.
 use cfg_if::cfg_if;
+use libc::{c_int, c_long, rusage};
 
 use crate::errno::Errno;
+use crate::sys::time::TimeVal;
 use crate::Result;
 pub use libc::rlim_t;
 use std::mem;
@@ -19,7 +21,7 @@ cfg_if! {
         target_os = "dragonfly",
         all(target_os = "linux", not(target_env = "gnu"))
     ))]{
-        use libc::{c_int, rlimit};
+        use libc::rlimit;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/sys/resource.rs
```diff
@@ -261,3 +260,177 @@ pub fn setrlimit(
...
+mod test {
+    use super::{getrusage, UsageWho};
+
+    #[test]
+    pub fn test_self_cpu_time() {
+        // Make sure some CPU time is used. FIXME: replace assert with blackbox once stabilized.
+        let mut numbers: Vec<i32> = (1..1_000_000).collect();
+        numbers.iter_mut().for_each(|item| *item *= 2);
+
+        assert_eq!(numbers[100..200].iter().sum::<i32>(), 30_100);
```
rtzoeller, line 426: Is this assertion the one referenced by the `FIXME` above? I think you're just trying to busy wait and force the CPU to do some work - is that correct?

You might move the above comment closer to this assertion to make things clearer.

rtzoeller (commented):
Thanks! Can you add a note to the changelog linking back to this PR (squashed into the current commit)?

src/sys/resource.rs
kov: Exactly! I ran it on a FreeBSD VM and it failed to show any CPU time, so I thought this was a flaky test in the making and decided to add this 'busy wait'. I wanted to use test::black_box here to make sure it does not get optimized out, but it is nightly-only for now.

rtzoeller:
We use a tool called bors to ensure PRs merged at the same time don't break the build. I believe this was what you were referring to.

The CHANGELOG.md updates are made manually though; can you modify CHANGELOG.md, similar to [this PR](https://github.com/nix-rust/nix/pull/1727/files)? Ideally the changelog is modified in the same commit as the existing one, because bors won't squash the commit history when merging.

kov:
Oh, I thought you meant the commit message, didn't notice there was a changelog. On it!

rtzoeller (approved):
bors r+

PR #1752, opened 2022-06-28
Add missing DontRoute SockOpt

leoleoasd:
I'm not sure, but according to [unix standard](https://pubs.opengroup.org/onlinepubs/7908799/xns/setsockopt.html), SO_DONTROUTE should appear in all unix systems.
Does this need a test?


asomers (changes requested):
No, it doesn't need a test.  But could you please add a CHANGELOG entry?

asomers:
You need to rebase to fix the CI failures.

asomers (approved):
bors r+

PR #1759, opened 2022-07-10
More docs for dir and mqueue

asomers:
Add doc comments for the `dir` and `mqueue` modules.  Also, delete dead code in `mqueue`


rtzoeller (approved):
bors r+

PR #1760, opened 2022-07-11
Add const constructors for TimeSpec and TimeVal

asomers:
These are basically the same as From<libc::timespec> and
From<libc::timeval>, but they're const and require less typing.


rtzoeller (approved):
bors r+

PR #1764, opened 2022-07-15
Fix description of fchownat

valdaarhun:
Based on the man page for `fchownat` and `lchown`, I have got the impression that with `FchownatFlags::NoFollowSymlink`, fchownat and lchown are identical. I couldn't find any documentation on `lchmod`.


asomers (approved):
Yep, looks like a copypasta.

bors r+

PR #1770, opened 2022-07-23
Fix SockaddrLike::from_raw with unaligned inputs

asomers:
The major users of this function are functions like gethostname, which
will always properly align their buffers.  But out-of-crate consumers
could manually construct an unaligned buffer.  Handle that correctly.

Enable Clippy's cast_ptr_alignment lint.  It's disabled by default as it
reports many false positives, but it would've caught this problem.

Reported-by:	Miri
Fixes:		1769


rtzoeller (approved):

asomers:
bors r=rtzoeller

rtzoeller:
@asomers I think this is worth considering for a future 0.24.3 patch.

asomers:
> @asomers I think this is worth considering for a future 0.24.3 patch.

Can you think of any way that it might be exposed to a user?  The main use for this method is in functions like `getsockname`, which already aligned the pointer correctly.  And the test that revealed the problem was doing something that few ordinary users would.

rtzoeller:
Sorry, I could have been clearer. I don't think this justifies a patch, just that if we're already patching in the future for some reason, I'd consider pulling this in.

PR #1772, opened 2022-07-23
Add support for RecvOrigDstAddr on Linux

### LINKED ISSUE #1767: Ipv4RecvDstAddr recvmsg stuff disabled on Linux
It looks like definitions needed to find the real IP destination of a TPROXY redirected stream are disabled on Linux. Such as:

```rust
nix::sys::socket::sockopt::Ipv4RecvDstAddr
ControlMessageOwned::Ipv4RecvDstAddr
```

And the equivalent IPv6 versions.

https://powerdns.org/tproxydoc/tproxy.md.html

These are valid on Linux and should be enabled.

e.g. the code has:

```rust
#[cfg(any(
    target_os = "freebsd",
    target_os = "ios",
    target_os = "macos",
    target_os = "netbsd",
    target_os = "openbsd",
))]
#[cfg(feature = "net")]
sockopt_impl!(
    /// The `recvmsg(2)` call will return the destination IP address for a UDP
    /// datagram.
    Ipv4RecvDstAddr, Both, libc::IPPROTO_IP, libc::IP_RECVDSTADDR, bool);

```

I believe this explains why I cannot use the on Linux.

brianmay:
Fixes #1767

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1752](https://github.com/nix-rust/nix/pull/1752))
 - Added `signal::SigSet::from_sigset_t_unchecked()`.
   (#[1741](https://github.com/nix-rust/nix/pull/1741))
+- Added IP_ORIGDSTADDR using Ipv4OrigDstAddr in setsockopt and recvmsg.
+  (#[1772](https://github.com/nix-rust/nix/pull/1772))
+- Added IPV6_ORIGDSTADDR using Ipv6OrigDstAddr in setsockopt and recvmsg.
+  (#[1772](https://github.com/nix-rust/nix/pull/1772))
 
 ### Changed
 
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -757,6 +757,14 @@ pub enum ControlMessageOwned {
     #[cfg(feature = "net")]
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
     Ipv4RecvDstAddr(libc::in_addr),
+    #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+    #[cfg(feature = "net")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    Ipv4OrigDstAddr(libc::sockaddr_in),
+    #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+    #[cfg(feature = "net")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    Ipv6OrigDstAddr(libc::sockaddr_in6),
 
     /// UDP Generic Receive Offload (GRO) allows receiving multiple UDP
     /// packets from a single sender.
@@ -916,6 +924,12 @@ impl ControlMessageOwned {
                 let dl = ptr::read_unaligned(p as *const libc::in_addr);
                 ControlMessageOwned::Ipv4RecvDstAddr(dl)
             },
+            #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+            #[cfg(feature = "net")]
+            (libc::IPPROTO_IP, libc::IP_ORIGDSTADDR) => {
+                let dl = ptr::read_unaligned(p as *const libc::sockaddr_in);
+                ControlMessageOwned::Ipv4OrigDstAddr(dl)
+            },
             #[cfg(target_os = "linux")]
             #[cfg(feature = "net")]
             (libc::SOL_UDP, libc::UDP_GRO) => {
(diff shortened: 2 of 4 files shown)


brianmay:
Still trying to work out why this doesn't work. But is a start.

Also not sure if I should use `RecvOrigDstAddr` or `OrigDstAddr`.

brianmay:
Don't understand why I am getting errors on non-Linux platforms, everything I added was Linux only.

asomers (changes requested):
These new symbols are also defined for Android and FreeBSD.  You should enable them there.

brianmay:
@asomers Will do so, thanks.

I am not actually sure what these functions are suppose to be returning for non-TPROXY connections. Seem to be getting random addresses, which is causing the tests to fail. Not sure if that is normal or not.

Setting up TPROXY firewall rules probably an overkill for these tests.

brianmay:
Also for some reason this works:

```rust
let mut space = cmsg_space!(libc::in_addr, [u8; 0]);
```

But this fails with `MSG_CTRUNC`:

```rust
let mut space = cmsg_space!(libc::in_addr);
```

I have no idea what is going on here.

brianmay:
I am also a bit puzzled in that some Linux platforms don't appear to be running the `test_recvif_ipv6` test, and some don't appear to be running the `test_recvif_ipv4` test either.

asomers:
> Also for some reason this works:
> 
> ```rust
> let mut space = cmsg_space!(libc::in_addr, [u8; 0]);
> ```
> 
> But this fails with `MSG_CTRUNC`:
> 
> ```rust
> let mut space = cmsg_space!(libc::in_addr);
> ```
> 
> I have no idea what is going on here.

That's because `in_addr` is the wrong datatype.  According to the FreeBSD man page at least, `IP_ORIGDSTADDR` sends its message as a `struct sockaddr_in`.

brianmay:
OK, fixed that, was confused reading an existing test.

Trying to run a local test with TPROXY, but first need to debug why my TPROXY UDP  firewall rules don't appear to be working...

src/sys/socket/mod.rs
```diff
@@ -757,6 +757,14 @@ pub enum ControlMessageOwned {
     #[cfg(feature = "net")]
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
     Ipv4RecvDstAddr(libc::in_addr),
+    #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+    #[cfg(feature = "net")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    Ipv4RecvOrigDstAddr(libc::in_addr),
```
asomers, line 763: This is still wrong.  It should be `libc::sockaddr_in`.

src/sys/socket/mod.rs
```diff
@@ -757,6 +757,14 @@ pub enum ControlMessageOwned {
...
+    #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+    #[cfg(feature = "net")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    Ipv6RecvOrigDstAddr(libc::in6_addr),
```
asomers, line 767: And this should be `sockaddr_in6`

asomers (changes requested):

src/sys/socket/mod.rs
brianmay: Fixed.

src/sys/socket/mod.rs
brianmay: Fixed.

brianmay:
All tests pass now when run locally....

brianmay:
Some of the Linux targets are failing tests. Wonder if this is a Linux compile time option.

brianmay:
Confirmed it works with a really hackish TPROXY UDP client now.

Looks like values need network byte ordering conversion, was expecting that.

asomers:
The `ENOPROTOOPT` errors are probably due to lack of support in QEMU user mode emulation.  There are a bunch of tests that have to be disabled due to that.  You can add `#[cfg_attr(qemu, ignore)]` to those tests.

brianmay:
Done.

asomers (changes requested):
Mostly looks good, but for the naming.  And don't forget to add a CHANGELOG entry.

src/sys/socket/mod.rs
```diff
@@ -757,6 +757,14 @@ pub enum ControlMessageOwned {
     #[cfg(feature = "net")]
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
     Ipv4RecvDstAddr(libc::in_addr),
+    #[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+    #[cfg(feature = "net")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    Ipv4RecvOrigDstAddr(libc::sockaddr_in),
```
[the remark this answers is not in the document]
brianmay: Done.

brianmay:
Added changelog entry.

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1752](https://github.com/nix-rust/nix/pull/1752))
 - Added `signal::SigSet::from_sigset_t_unchecked()`.
   (#[1741](https://github.com/nix-rust/nix/pull/1741))
+- Added IP_ORIGDSTADDR using Ipv4OrigDstAddr in setsockopt and recvmsg.
+- Added IPV6_ORIGDSTADDR using Ipv6OrigDstAddr in setsockopt and recvmsg.
```
asomers, line 34: Needs a PR link.

asomers (changes requested):

brianmay:
> Needs a PR link. 

Done.

CHANGELOG.md
brianmay: Done.

asomers (approved):
bors r+

asomers:
BTW, I just added IPV6_ORIGDSTADDR to the ip6(4) man page on FreeBSD.

PR #1773, opened 2022-07-25
Add memfd for target_os = "android"

flxo:
Memory fds (`memfd`) are implemented and exported by Androids bionic.
Export the `memfd` module if the target os is `android`.

https://cs.android.com/android/platform/superproject/+/master:bionic/libc/include/sys/mman.h;drc=23c7543b8e608ebcbb38b952761b54bb56065577;bpv=1;bpt=1;l=182


asomers (approved):
bors r+

PR #1774, opened 2022-07-26
Fix typo guarding memfd export on android.

detly:
Extremely minor but crucial typo on import guard from #1773. @flxo


flxo:
Thanks @detly!

asomers (approved):
Ahh, good.  Thanks for the fix.

bors r+

PR #1776, opened 2022-08-01
Add support for the IP_SENDSRCADDR control message

matttpt:
This control message is available on FreeBSD, NetBSD, and OpenBSD. When used with `sendmsg`, it sets the IPv4 source address. This adds support through a new `ControlMessage::Ipv4SendSrcAddr` variant that complements `ControlMessageOwned::Ipv4RecvDstAddr`.

A few notes:
* `IP_SENDSRCADDR` is actually just an alias for `IP_RECVDSTADDR` (though the code doesn't depend on this).
* On NetBSD, `IP_PKTINFO` can be used to accomplish the same thing and is already supported by nix. On FreeBSD and OpenBSD, though, `IP_SENDSRCADDR` is the only method I'm aware of.
* The accompanying test binds a UDP socket to all local interfaces (0.0.0.0). If this is not acceptable, please let me know; however, FreeBSD requires this to use `IP_SENDSRCADDR`.

I'll add a change-log entry once I see the PR number.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1772](https://github.com/nix-rust/nix/pull/1772))
 - Added IPV6_ORIGDSTADDR using Ipv6OrigDstAddr in setsockopt and recvmsg.
   (#[1772](https://github.com/nix-rust/nix/pull/1772))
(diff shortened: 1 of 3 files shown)


asomers (changes requested):
Looks like DragonflyBSD supports it too.  Can you add that please?  Also, the test failures should be fixed if you rebase.

rtzoeller (approved):
bors r+

PR #1778, opened 2022-08-04
fix clippy assertions_on_result_states

pacak:
https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_result_states

Fixes broken CI for https://github.com/nix-rust/nix/pull/1744

### DIFF
--- a/src/sys/signalfd.rs
+++ b/src/sys/signalfd.rs
@@ -148,14 +148,17 @@ mod tests {
     fn create_signalfd() {
         let mask = SigSet::empty();
         let fd = SignalFd::new(&mask);
-        assert!(fd.is_ok());
+        fd.expect("assert failed");
     }
 
     #[test]
     fn create_signalfd_with_opts() {
         let mask = SigSet::empty();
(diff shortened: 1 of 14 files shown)


rtzoeller (approved):
bors r+

rtzoeller:
Thanks!

src/sys/signalfd.rs
```diff
@@ -148,14 +148,17 @@ mod tests {
     fn create_signalfd() {
         let mask = SigSet::empty();
         let fd = SignalFd::new(&mask);
-        assert!(fd.is_ok());
+        fd.expect("assert failed");
```
asomers, line 151: "assert failed" isn't a helpful error message.  Could you replace them with something more useful?  Or if not, then just use `.unwrap()`.

asomers (changes requested):

src/sys/signalfd.rs
pacak: https://github.com/nix-rust/nix/pull/1786

The idea was to be able to tell that this is an assert instead of a random unwrap, but I don't mind either way.

PR #1779, opened 2022-08-04
Upgrade libc to 0.2.127

rtzoeller:
This is the last version of libc which will support Rust 1.46, per https://github.com/rust-lang/libc/pull/2845.

I think we should ship nix 0.25.0 against this version of libc (at some point, not necessarily now), as it will allow us to maintain an MSRV of 1.46, and subsequently upgrade to a later libc immediately after releasing 0.25.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


rtzoeller:
We'll need to figure out a new MSRV strategy once `libc` picks one (or just adopt theirs).

asomers (approved):
bors r+

PR #1788, opened 2022-08-10
Workaround XNU bug in getifaddrs netmasks

### LINKED ISSUE #1709: `SockaddrStorage::as_sockaddr_in` incorrectly returns None sometime

roblabla:
Fixes #1709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Fix `User::from_uid` and `User::from_name` crash on Android platform.
   ([#1824](https://github.com/nix-rust/nix/pull/1824))
 - Workaround XNU bug causing netmasks returned by `getifaddrs` to misbehave.
-  (#1709)
+  ([#1788](https://github.com/nix-rust/nix/pull/1788))
   
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#1815](https://github.com/nix-rust/nix/pull/1815))
 - Fix `User::from_uid` and `User::from_name` crash on Android platform.
   ([#1824](https://github.com/nix-rust/nix/pull/1824))
-
+- Workaround XNU bug causing netmasks returned by `getifaddrs` to misbehave.
+  (#1709)
+  
 ### Removed
 
 - Removed deprecated error constants and conversions.
--- a/src/ifaddrs.rs
+++ b/src/ifaddrs.rs
@@ -4,6 +4,8 @@
 //! of interfaces and their associated addresses.
 
 use cfg_if::cfg_if;
+#[cfg(any(target_os = "ios", target_os = "macos"))]
+use std::convert::TryFrom;
 use std::ffi;
 use std::iter::Iterator;
 use std::mem;
@@ -42,11 +44,50 @@ cfg_if! {
     }
 }
 
+/// Workaround a bug in XNU where netmasks will always have the wrong size in
+/// the sa_len field due to the kernel ignoring trailing zeroes in the structure
+/// when setting the field. See https://github.com/nix-rust/nix/issues/1709#issuecomment-1199304470
+///
+/// To fix this, we stack-allocate a new sockaddr_storage, zero it out, and
+/// memcpy sa_len of the netmask to that new storage. Finally, we reset the
+/// ss_len field to sizeof(sockaddr_storage). This is supposedly valid as all
+/// members of the sockaddr_storage are "ok" with being zeroed out (there are
+/// no pointers).
+#[cfg(any(target_os = "ios", target_os = "macos"))]
+unsafe fn workaround_xnu_bug(info: &libc::ifaddrs) -> Option<SockaddrStorage> {
+    let src_sock = info.ifa_netmask;
+    if src_sock.is_null() {
+        return None;
+    }
+
+    let mut dst_sock = mem::MaybeUninit::<libc::sockaddr_storage>::zeroed();
+
+    // memcpy only sa_len bytes, assume the rest is zero
+    std::ptr::copy_nonoverlapping(
+        src_sock as *const u8,
+        dst_sock.as_mut_ptr() as *mut u8,
+        (*src_sock).sa_len.into(),
+    );
+
+    // Initialize ss_len to sizeof(libc::sockaddr_storage).
+    (*dst_sock.as_mut_ptr()).ss_len =
+        u8::try_from(mem::size_of::<libc::sockaddr_storage>()).unwrap();
+    let dst_sock = dst_sock.assume_init();
+
+    let dst_sock_ptr =
+        &dst_sock as *const libc::sockaddr_storage as *const libc::sockaddr;
+
+    SockaddrStorage::from_raw(dst_sock_ptr, None)
+}
+
 impl InterfaceAddress {
     /// Create an `InterfaceAddress` from the libc struct.
     fn from_libc_ifaddrs(info: &libc::ifaddrs) -> InterfaceAddress {
         let ifname = unsafe { ffi::CStr::from_ptr(info.ifa_name) };
         let address = unsafe { SockaddrStorage::from_raw(info.ifa_addr, None) };
+        #[cfg(any(target_os = "ios", target_os = "macos"))]
+        let netmask = unsafe { workaround_xnu_bug(info) };
+        #[cfg(not(any(target_os = "ios", target_os = "macos")))]
         let netmask =
             unsafe { SockaddrStorage::from_raw(info.ifa_netmask, None) };
         let mut addr = InterfaceAddress {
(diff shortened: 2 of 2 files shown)


src/ifaddrs.rs
```diff
@@ -42,11 +42,47 @@ cfg_if! {
...
+    }
+    let family = unsafe {
+        (*info.ifa_netmask).sa_family
+    };
+
+    let fixup_size = match i32::from(family) {
+        libc::AF_INET => Some(std::mem::size_of::<libc::sockaddr_in>()),
+        libc::AF_INET6 => Some(std::mem::size_of::<libc::sockaddr_in6>()),
+        _ => None
+    };
```
poliorcetics, line 64:
```suggestion
    let fixup_size = match i32::from(family) {
        libc::AF_INET => std::mem::size_of::<libc::sockaddr_in>(),
        libc::AF_INET6 => std::mem::size_of::<libc::sockaddr_in6>(),
        _ => return
    };
```

src/ifaddrs.rs
```diff
@@ -42,11 +42,47 @@ cfg_if! {
...
+
+    if let Some(fixup_size) = fixup_size {
+        assert!(fixup_size < u8::MAX.into());
+        let sock = info.ifa_netmask as *mut libc::sockaddr;
+        unsafe {
+            if (*sock).sa_len < fixup_size as u8 {
+                (*sock).sa_len = fixup_size as u8;
+            }
+        }
+    }
```
poliorcetics, line 74:
```suggestion
    assert!(fixup_size < u8::MAX.into());
    let sock = info.ifa_netmask as *mut libc::sockaddr;
    unsafe { (*sock).sa_len = (*sock).sa_len.max(fixup_size as u8) }
```

Also, seems like `ifa_netmask` is already of the type you're casting it to ? https://docs.rs/libc/latest/x86_64-apple-darwin/libc/struct.ifaddrs.html#structfield.ifa_netmask

src/ifaddrs.rs
```diff
@@ -42,11 +42,47 @@ cfg_if! {
     }
 }
 
+/// Workaround a bug in XNU where netmasks will always have the wrong size in
+/// the sa_len field due to the kernel ignoring trailing zeroes in the future
```
poliorcetics, line 46:
```suggestion
/// the `sa_len` field due to the kernel ignoring trailing zeroes in the structure
```

src/ifaddrs.rs
```diff
@@ -42,11 +42,47 @@ cfg_if! {
...
+/// when setting the field. See https://github.com/nix-rust/nix/issues/1709#issuecomment-1199304470
+///
+/// To fix this, we overwrite sa_len to have the known (static) value of the
```
poliorcetics, line 49:
```suggestion
/// To fix this, we overwrite `sa_len` to have the known (static) value of the
```

roblabla:
CI breakage seems unrelated to my MR (looks like the latest serde release broke compatibility with the version of rustc/cargo used by CI).

src/ifaddrs.rs
```diff
@@ -42,11 +42,43 @@ cfg_if! {
...
+        libc::AF_INET => std::mem::size_of::<libc::sockaddr_in>(),
+        libc::AF_INET6 => std::mem::size_of::<libc::sockaddr_in6>(),
+        _ => return
+    };
+
+    assert!(fixup_size < u8::MAX.into());
+    let sock = info.ifa_netmask;
+    unsafe {
+        (*sock).sa_len = (*sock).sa_len.max(fixup_size as u8);
+    }
```
poliorcetics, line 70:
```suggestion
    let sock = info.ifa_netmask;
    if sock.is_null() {
        return
    }
    let family = unsafe {
        (*sock).sa_family
    };

    let fixup_size = match i32::from(family) {
        libc::AF_INET => std::mem::size_of::<libc::sockaddr_in>(),
        libc::AF_INET6 => std::mem::size_of::<libc::sockaddr_in6>(),
        _ => return
    };

    assert!(fixup_size < u8::MAX.into());
    unsafe {
        (*sock).sa_len = (*sock).sa_len.max(fixup_size as u8)
    };
```

Uses `sock` everywhere so it's clearer we always look at the same memnber

poliorcetics: Uh, it's highlighting changes I haven't made ? Sorry for that, I don't know what's happening

asomers:
> CI breakage seems unrelated to my MR (looks like the latest serde release broke compatibility with the version of rustc/cargo used by CI).

The CI breakage was due to a Cargo bug that has since been fixed.  However, serde needs to yank their most recent release and publish a new one before things will work again.  https://github.com/serde-rs/serde/issues/2255 .

src/ifaddrs.rs
```diff
@@ -42,11 +42,43 @@ cfg_if! {
     }
 }
 
+/// Workaround a bug in XNU where netmasks will always have the wrong size in
+/// the sa_len field due to the kernel ignoring trailing zeroes in the structure
+/// when setting the field. See https://github.com/nix-rust/nix/issues/1709#issuecomment-1199304470
+///
+/// To fix this, we overwrite `sa_len` to have the known (static) value of the
+/// structure.
+#[cfg(target_os = "macos")]
```
asomers, line 51: Does the bug affect iOS too?

src/ifaddrs.rs
```diff
@@ -42,11 +42,43 @@ cfg_if! {
...
+    let fixup_size = match i32::from(family) {
+        libc::AF_INET => std::mem::size_of::<libc::sockaddr_in>(),
+        libc::AF_INET6 => std::mem::size_of::<libc::sockaddr_in6>(),
+        _ => return
+    };
+
+    assert!(fixup_size < u8::MAX.into());
+    unsafe {
+        (*sock).sa_len = (*sock).sa_len.max(fixup_size as u8);
+    }
```
asomers, line 70: You can avoid an extra explicit assertion like this:
```suggestion
    unsafe {
        (*sock).sa_len = (*sock).sa_len.max(u8::try_from(fixup_size).unwrap());
    }
```
However, this isn't safe.  How do you know that the memory past sa_len is actually allocated and unused?  You can't.  Instead, you should fix it by copying it to a new location.

src/ifaddrs.rs
```diff
@@ -42,11 +42,43 @@ cfg_if! {
...
 }
 
+/// Workaround a bug in XNU where netmasks will always have the wrong size in
+/// the sa_len field due to the kernel ignoring trailing zeroes in the structure
+/// when setting the field. See https://github.com/nix-rust/nix/issues/1709#issuecomment-1199304470
+///
+/// To fix this, we overwrite `sa_len` to have the known (static) value of the
+/// structure.
+#[cfg(target_os = "macos")]
+fn workaround_xnu_bug(info: &libc::ifaddrs) {
```
asomers, line 52: Needs a mutable reference because you're changing the structure.  The compiler doesn't warn you because the change happens via a raw pointer, but you still need a mutable reference for safety.
```suggestion
fn workaround_xnu_bug(info: &mut libc::ifaddrs) {
```

src/ifaddrs.rs
```diff
@@ -144,4 +176,21 @@ mod tests {
...
+                sock
+            } else {
+                continue
+            };
+            if sock.family() == Some(crate::sys::socket::AddressFamily::Inet) {
+                let _ = sock.as_sockaddr_in().unwrap();
+                return;
+            }
+        }
+        panic!("No address?");
```
asomers, line 194: This will panic on an IPv6-only system.  Don't do that.

src/ifaddrs.rs
```diff
@@ -144,4 +176,21 @@ mod tests {
     fn test_getifaddrs() {
         let _ = getifaddrs();
     }
+
+    #[test]
```
asomers, line 180: To a casual reader, this test looks pretty pointless.  Please add a comment explaining the rationale.

asomers (changes requested):
It really sticks in my craw that we have to do this, but I guess we do.  We can't fix OSX.  Don't forget to add a CHANGELOG entry.

src/ifaddrs.rs
roblabla: I suppose so, given the kernel is the same, but I have no way to test - I don't own an iDevice.

asomers:
Rebasing will fix ci

src/ifaddrs.rs
roblabla: It shouldn't? There's localhost. But I can always do Inet || Inet6.

roblabla:
I rewrote the workaround to be safer - at least, I think it's safer. It now copies the data to a new zeroed sockaddr_storage.

roblabla:
Bump. This has been used in prod for a little while, and is, as far as I can tell, sound.

asomers:
@roblabla you still need to rebase this, both to fix CI and to fix the conflict.

asomers:
Looks good, but still needs a CHANGELOG entry.

CHANGELOG.md
```diff
@@ -45,7 +45,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#1815](https://github.com/nix-rust/nix/pull/1815))
 - Fix `User::from_uid` and `User::from_name` crash on Android platform.
   ([#1824](https://github.com/nix-rust/nix/pull/1824))
-
+- Workaround XNU bug causing netmasks returned by `getifaddrs` to misbehave.
+  (#1709)
```
asomers, line 49:
```suggestion
  ([#1788](https://github.com/nix-rust/nix/pull/1788))
```

asomers (changes requested):

asomers (approved):
bors r+

PR #1793, opened 2022-08-11
Disable cargo-hack in CI

asomers:
Serde accidentally raised its MSRV to 1.51.0 in a patch release.  They
don't intent to fix it. Nix uses Serde via cargo-hack in CI.  Disable it
so we can publish a final release at 1.46.0.


rtzoeller (approved):
bors r+

rtzoeller:
Looks like something is up with `cross` as well?

asomers:
bors r=rtzoeller

PR #1801, opened 2022-08-19
fix microsecond calculation for TimeSpec

StephanvanSchaik:
The implementation of `num_microseconds()` for `TimeSpec` returns the number of seconds not microseconds, as it divided by the wrong factor. The number of nanoseconds should be divided by 1000 to get the number of microseconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


rtzoeller:
@StephanvanSchaik thanks! Can you add an entry to CHANGELOG.md, ideally squashed into the current commit?

StephanvanSchaik:
@rtzoeller Done!

rtzoeller (approved):
bors r+

Issue #1802, opened 2022-08-19
Termios `c_line` field


tertsdiepraam:
Hi! I've been working on an `stty` implementation in Rust for uutils using `nix` and I noticed that the `nix::Termios` struct does not have a `c_line` field, even though the `libc::termios` does on Linux. Would it be possible to add it for Linux only?

I could give this a go if this is something you want to include.

rtzoeller:
I think adding this makes sense; it looks like Haiku and possibly a few other platforms also support it (at least it's present in rust-lang/libc). 

> I could give this a go if this is something you want to include.

Please do! Don't forget to add an entry to the CHANGELOG.md as part of that commit.

PR #1805, opened 2022-08-20
Add `line` field to `Termios` struct

tertsdiepraam:
Fixes https://github.com/nix-rust/nix/issues/1802

I have to admit I'm not really sure how to test this properly, so if that's necessary I require some help :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased] - ReleaseDate
 ### Added
 
+- Added `line_discipline` field to `Termios` on Linux, Android and Haiku
+    ([#1805](https://github.com/nix-rust/nix/pull/1805))
+
 ### Changed
 
 - The MSRV is now 1.56.1
--- a/src/sys/termios.rs
+++ b/src/sys/termios.rs
@@ -181,6 +181,15 @@ pub struct Termios {
     pub local_flags: LocalFlags,
     /// Control characters (see `termios.c_cc` documentation)
     pub control_chars: [libc::cc_t; NCCS],
+    /// Line discipline (see `termios.c_line` documentation)
+    #[cfg(any(
+        target_os = "linux",
+        target_os = "android",
+    ))]
+    pub line_discipline: libc::cc_t,
+    /// Line discipline (see `termios.c_line` documentation)
+    #[cfg(target_os = "haiku")]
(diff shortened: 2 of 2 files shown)


src/sys/termios.rs
```diff
@@ -181,6 +181,16 @@ pub struct Termios {
     pub local_flags: LocalFlags,
     /// Control characters (see `termios.c_cc` documentation)
     pub control_chars: [libc::cc_t; NCCS],
+    /// Line number (see `termios.c_line` documentation)
```
rtzoeller, line 184: `c_line` is described on Linux as "line discipline", not "line number". I'm guessing it's the same for other platforms.

There's an enumeration of the disciplines on Linux [here](https://github.com/torvalds/linux/blob/master/include/uapi/linux/tty.h). I don't think we need to add them to Nix necessarily, but I do see that they are also missing from https://github.com/rust-lang/libc; you might consider adding them there.

tertsdiepraam: Ah of course. I'll update the PR. Shall I add the possible values to the docstring instead of making an `enum`?

tertsdiepraam: And do you think the field should be called `line_discipline` given that the other fields also have more descriptive names than `libc`?

rtzoeller: I do think `line_discipline` is a better name.

I don't have a strong preference between using a numeric type vs. an `enum`. If you want to go the `enum` route, we'd want to derive our `enum` values from the `libc` definition, which we do using the `libc_bitflags!` macro. This would necessitate adding them to `libc` before introducing them here. I'll leave that choice up to you.

If you're sticking with a numeric type, I don't see a need to enumerate them fully in Nix's documentation. I think referring users to their platform-specific documentation is fine.

tertsdiepraam: I'll stick with numbers for now. In existing `stty` implementations the number is also used directly. Thank you for your feedback!

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased] - ReleaseDate
 ### Added
 
+- Added `line` field to `Termios` on Linux, Android and Haiku
```
rtzoeller, line 9: Can you update this to `line_discipline`?

rtzoeller: Otherwise things LGTM.

rtzoeller (approved):
bors r+

PR #1807, opened 2022-08-22
Add pthread_jit_write_protect api subset into sys::pthread

devnexen:


asomers (commented):
It doesn't seem like these bindings add a whole lot of value.  Is it really necessary to have them in Nix?

devnexen:
In fact, those are necessary on mac arm64 (not much on amd64) if you do mapping with execution and write bits.

asomers:
But it is necessary for those functions to be `in Nix` as opposed to remaining in libc.  Because all your bindings do is call the libc function, which a consumer could do just as well.

devnexen:
good point.

PR #1808, opened 2022-08-23
expose memfd on freebsd

### LINKED ISSUE #1775: Expose memfd on (Free?)BSD
According to [the FreeBSD man pages](https://www.freebsd.org/cgi/man.cgi?query=memfd_create&sektion=3&n=1) `memfd_create()` has been present since 13.0. Should it be exported like it is for Android and Linux?

i509VCB:
Resolves #1775

### DIFF
--- a/src/sys/memfd.rs
+++ b/src/sys/memfd.rs
@@ -1,6 +1,8 @@
 //! Interfaces for managing memory-backed files.
 
 use std::os::unix::io::RawFd;
+use cfg_if::cfg_if;
+
 use crate::Result;
 use crate::errno::Errno;
 use std::ffi::CStr;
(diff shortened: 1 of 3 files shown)


src/sys/memfd.rs
```diff
@@ -40,7 +42,15 @@ libc_bitflags!(
 /// [`memfd_create(2)`]: https://man7.org/linux/man-pages/man2/memfd_create.2.html
 pub fn memfd_create(name: &CStr, flags: MemFdCreateFlag) -> Result<RawFd> {
     let res = unsafe {
-        libc::syscall(libc::SYS_memfd_create, name.as_ptr(), flags.bits())
```
asomers, line 43: I would rather not bend over backwards to cater to the obscure arches.  If they don't define the libc symbol, then I think we should just not bind it.  We can do it like this:
`#[cfg(any(target_os = "freebsd", target_env = "gnu", target_env = "musl"))]`

i509VCB: Seem that it's just armv7 uclibceabihf that's failing

i509VCB: I have some solution but it's a more ugly cfg block.

Essentially use the libc symbol if it's freebsd or gnu/musl linux and otherwise use the syscall function.

Or was the intention to gate memfd_create from all linuxes that aren't using a gnu or musl libc?

asomers (approved):
bors r+

i509VCB:
conflicts are fixed

asomers (approved):
bors r+

i509VCB:
The failures appear to be unrelated to this pull request.

asomers:
The build should be fixed if you rebase.

PR #1815, opened 2022-09-11
Handle unacceptable name gracefully in {User,Group}::from_name

magicant:
Calling `unwrap` on the result of `CString::new` may cause the current thread to panic, which is a bit surprising undocumented behavior. It would be more reasonable to treat the erroneous name as a non-existing user or group.


asomers (approved):
Thanks!

bors r+

asomers:
The build should be fixed if you rebase.

asomers:
Actually, maybe bors is smart enough to do that.

bors r+

Issue #1811, opened 2022-08-30
API Request: expose getpwent and getgrent


SteveLauC:
These two syscalls sequentially scan the `/etc/passwd/ and `/etc/group` databases and return all the entries in those files one by one.

#### Signature
```rust
pub fn getpwent() -> Result<Vec<User>, Errno>;
pub fn getgrent() -> Result<Vec<Group>, Errno>;
```
#### Implementation
On some OSes, we have non-reentrant functions `getpwent_r()` and `getgrent_r()`. If so, they will be used. Otherwise, will use `getpwent()` and `getgrent()`.

Here is  draft implementation of `getgrent` under Linux:
```rust
pub fn getgrent() -> Result<Vec<Group>, Errno> {
    let mut v: Vec<Group> = Vec::new();

    let mut gr_buf: group = unsafe { zeroed() };
    let mut gr_str_buf: [c_char; 4096] = [0; 4096];
    unsafe { setgrent() };

    loop {
        let mut result: *mut group = null_mut();
        let res: c_int = unsafe {
            getgrent_r(
                &mut gr_buf as *mut group,
                &mut gr_str_buf as *mut c_char,
                4096,
                &mut result as *mut *mut group,
            )
        };

        // no more entries
        if res == ENOENT {
            unsafe { endgrent() };
            break;
        }

        // error
        if res == ERANGE {
            return Err(Errno::from_i32(ERANGE));
        }

        // man page for `getgrent_r` only lists the two above erroneous cases.
        // In other cases, it should be successful.
        assert_eq!(res, 0);
        v.push(Group::from(&gr_buf));
    }

    Ok(v)
}
```

#### Note
1. The above implementation calls `setgrent` and `endgrent` inside, which actually changes the semantics of this syscall.
2. Due to Note `1`, perhaps we can give a better name to those syscalls, like:

    ```rust
    impl User {
        pub fn all_users() -> Result<Vec<User>, Errno>;
    }
    impl Group {
        pub fn getgrent() -> Result<Vec<Group>, Errno>;
    }
    ```

zseri:
I think it would be clearer if `unsafe { endgrent() };` would be moved below the loop.

SteveLauC:
> I think it would be clearer if `unsafe { endgrent() };` would be moved below the loop.

Good catch! It should be.

asomers:
Hm.  This would work.  But I suggest changing the interface to avoid collecting into a Vec.  More like this:
```rust
[...]

PR #1817, opened 2022-09-11
fix issue1814 add domainname field of utsname on android and linux

SteveLauC:
Closes #1814

PR #1820, opened 2022-09-16
expose getpwent and getgrent

SteveLauC:
This PR is for issue #1811.

On OSes where `getpwent_r/getgrent_r` is available, it is used. On other OSes, use `getpwent/getgrent`. However, `illumos` and `solaris`, they do have `getpwent_r/getgrent_r`, but the definitions exposed by `libc` are [wrong](https://github.com/rust-lang/libc/issues/2908), so I currently use `getpwent/getgrent` on them.

The implementations of `AllUsers::next` and `AllGroups::next` on platforms where `getpwent_r/getgrent_r` is available are kind of ugly, this is because the return values of these functions are so different:

* freebsd: [getpwent_r man page](https://www.freebsd.org/cgi/man.cgi?query=getpwent&sektion=3)
  1. On success: return 0, `result` will be non-NULL
  2. On error: return the error number, `result` will be NULL
  3. End: return 0, `result` will be NULL. 

* dragonfly: [getpwent_r man page](https://leaf.dragonflybsd.org/cgi/web-man?command=getpwent_r&section=ANY)
  1. On success: return 0 and `result` will be non-NULL
  2. On error: return the error number, `result` will be NULL
  3. End: return 0, `result` will be NULL
 
> The return values on freebsd and dragonfly are the same.

* netbsd: [getpwent_r man page](https://man.netbsd.org/getpwent_r.3)
  1. On success: return 0, `result` will be non-NULL, 
  2. On error: return a non-zero value (errno is set), `result` will be NULL
  2. End: return 0, `result` will be NULL

* Linux: [getpwent_r man page](https://man7.org/linux/man-pages/man3/getpwent_r.3.html)
  1. On success: return 0, `result` will be non-NULL
  2. On error: return the error number (not ENOENT), `result` will be NULL
  3. End: return ENOENT and `result` will be NULL.

Another thing I am not sure about is the return value of `AllUsers::next()/AllGroups::next()`. In my implementation, it is:

```rust
fn next(&mut self) -> Option<Result<User/Group, Errno>>
```
The inner `Result` is used to expose the `errno` value, the drawback of such a design is that the user has to stop iterating when an error occurs.

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -3151,6 +3151,139 @@ impl User {
             unsafe { libc::getpwnam_r(name.as_ptr(), pwd, cbuf, cap, res) }
         })
     }
+
+    /// Construct an iterator over all users on the system.
+    ///
+    /// Call [`setpwent(3)`](https://pubs.opengroup.org/onlinepubs/7908799/xsh/setpwent.html),
+    /// [`getpwent(3)`](https://pubs.opengroup.org/onlinepubs/7908799/xsh/getpwent.html),
+    /// [`getpwent_r(3)`](https://man7.org/linux/man-pages/man3/getpwent_r.3.html) and
+    /// [`endpwent(3)`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/endpwent.html)
+    /// under the hood.
+    ///
+    /// When encounters an error, [`AllUsers::next()`] will yield a
+    /// `Some(Err(Errno))`, stop iterating if this happens.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked as `unsafe` cause `setpwent()` and `endpwent()`
+    /// will modify a global state, which will result in a data race if we have
+    /// multiple instances running at the same time.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use nix::unistd::User;
+    /// for opt_user in unsafe { User::all_users() } {
+    ///     match opt_user {
+    ///         Ok(user) => println!("{:?}", user),
+    ///         Err(errno) => {
+    ///             eprintln!("{:?}", errno);
+    ///             break;
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    // RedoxFS does not support passwd and android does not have /etc/passwd.
+    #[cfg(not(any(target_os = "redox", target_os = "android")))]
+    pub unsafe fn all_users() -> AllUsers {
+        libc::setpwent();
+
+        AllUsers
+    }
+}
+
+/// An iterator over the users of this system
+///
+/// This struct is constructed by [`User::all_users()`].
+// RedoxFS does not support passwd and android does not have /etc/passwd.
+#[cfg(not(any(target_os = "redox", target_os = "android")))]
+#[derive(Debug)]
+pub struct AllUsers;
+
+#[cfg(not(any(target_os = "redox", target_os = "android")))]
+impl Iterator for AllUsers {
+    type Item = Result<User>;
+
+    #[cfg(any(
+        target_os = "freebsd",
+        target_os = "netbsd",
+        target_os = "dragonfly",
+        all(target_os = "linux", target_env = "gnu"),
+    ))]
+    // On these OSes, call `getpwent_r()`
+    fn next(&mut self) -> Option<Self::Item> {
+            #[cfg(target_os = "netbsd")]
+            Errno::clear();
+
+            let mut pwd = mem::MaybeUninit::<libc::passwd>::uninit();
+            let str_buf_size = match sysconf(SysconfVar::GETPW_R_SIZE_MAX) {
+                Ok(Some(n)) => n as usize,
+                Err(_) | Ok(None) => 4096,
+            };
+            let mut str_buf = Vec::with_capacity(str_buf_size);
+            let mut res: *mut libc::passwd = ptr::null_mut();
+            let ret = unsafe{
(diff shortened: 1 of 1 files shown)


SteveLauC:
Haven't added the changes to `CHANGELOG.md`. 

Will add it, maybe after #1817 is merged, or there will be a merge conflict.

SteveLauC:
The CI for fuchsia failed because `setgrent/getgrent/endgrent` are currently not exposed in `libc`.

src/unistd.rs
```diff
@@ -3151,6 +3151,139 @@ impl User {
...
+    /// [`getpwent_r(3)`](https://man7.org/linux/man-pages/man3/getpwent_r.3.html) and
+    /// [`endpwent(3)`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/endpwent.html)
+    /// under the hood.
+    ///
+    /// When encounters an error, [`AllUsers::next()`] will yield a
+    /// `Some(Err(Errno))`, stop iterating if this happens.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked as `unsafe` cause `setpwent()` and `endpwent()`
```
asomers, line 3168:
```suggestion
    /// This function is marked as `unsafe` because `setpwent()` and `endpwent()`
```

src/unistd.rs
```diff
@@ -3151,6 +3151,139 @@ impl User {
...
+#[derive(Debug)]
+pub struct AllUsers;
+
+#[cfg(not(any(target_os = "redox", target_os = "android")))]
+impl Iterator for AllUsers {
+    type Item = Result<User>;
+
+    #[cfg(any(
+        target_os = "freebsd",
+        target_os = "netbsd",
```
asomers, line 3209: What about OpenBSD?

src/unistd.rs
```diff
@@ -3151,6 +3151,139 @@ impl User {
...
+    /// [`endpwent(3)`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/endpwent.html)
+    /// under the hood.
+    ///
+    /// When encounters an error, [`AllUsers::next()`] will yield a
+    /// `Some(Err(Errno))`, stop iterating if this happens.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked as `unsafe` cause `setpwent()` and `endpwent()`
+    /// will modify a global state, which will result in a data race if we have
```
asomers, line 3169: What kind of data race?  Skipping over some users, or duplicating some, is not considered `unsafe`.  Only something like reading a struct with some entries from one source and some from another would be considered `unsafe`.

src/unistd.rs
```diff
@@ -3151,6 +3151,139 @@ impl User {
...
+    ///         Err(errno) => {
+    ///             eprintln!("{:?}", errno);
+    ///             break;
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    // RedoxFS does not support passwd and android does not have /etc/passwd.
+    #[cfg(not(any(target_os = "redox", target_os = "android")))]
+    pub unsafe fn all_users() -> AllUsers {
```
asomers, line 3188: I suggest that a more natural API would be to call this function `AllUsers::new()`

src/unistd.rs
```diff
@@ -3151,6 +3151,139 @@ impl User {
...
+            }
+
+            // Now, all we need to do is to distinguish the successful case and
+            // the case where there are no more entries. Luckily, we can do this
+            // through `res`.
+
+            // no more entries
+            if res.is_null() {
+                None
+            }else {
```
asomers, line 3250:
```suggestion
            } else {
```

asomers (commented):
This `setpwent` business sucks.  Very un-Rusty.  I wonder if it would better just to use the raceless `fgetpwent_r` function instead.  I think that's only available on GNU/Linux, but at least it's thread-safe.

src/unistd.rs
SteveLauC: OpenBSD does not have [`getpwent_r`](https://man.openbsd.org/getpwent_r) and [`getgrent_r`](https://man.openbsd.org/getgrent_r)

src/unistd.rs
SteveLauC: There are two issues here I think:
1. The pointer returned by `getpwent()` points to static memory, and thus can be overwritten by subsequent calls like `getpwent`, `getpwuid`, and `getpwnam`. So in the implementation using `getpwent()`, when executing `Some(Ok(User::from(unsafe{&*pwd})))` in a thread (say thread A), there happens to be another thread (thread B) calling `getpwent/getpwuid/getpwnam`, then the static memory will be overwritten due to the call in thread B, making thread A get a wrong result. This is the data race.
2. `setpwent/endpwent` will modify a global state, if two instances are running at the same time, the state set by the first instance can be reset by the second one, then the first instance will get duplicate entries. 

Sorry for the unclarity in my documentation, will update it stating both issues.

SteveLauC:
> This `setpwent` business sucks. Very un-Rusty. I wonder if it would better just to use the raceless `fgetpwent_r` function instead. I think that's only available on GNU/Linux, but at least it's thread-safe.

Yes, this API sucks. What are we going to do about this pr? Change the implementation for GNU/Linux to `fgetpwent_r`, what about the other platforms?

voidc:
This would be useful. Is there anything blocking this PR apart from a rebase?

SteveLauC:
> This would be useful. Is there anything blocking this PR apart from a rebase?

This: https://github.com/nix-rust/nix/pull/1820#issuecomment-1250171451

PR #1821, opened 2022-09-17
Fix UB in the SO_TYPE sockopt

### LINKED ISSUE #1819: `impl GetSockOpt for nix::sys::socket::sockopt::SockType` causes UB with unknown socket type
## Description

I noticed that the implementation of `GetSockOpt` for `nix::sys::socket::sockopt::SockType` [simply assumes](https://docs.rs/nix/0.25.0/src/nix/sys/socket/sockopt.rs.html#91) that the socket type returned by the kernel is one of those defined in `enum nix::sys::socket::SockType` without checking. If that assumption is violated, you get undefined behavior.

Linux has at least one such socket type already (`SOCK_PACKET`), and operating systems are free to add new ones at any time.

Other socket options probably suffer from a similar problem, but I haven't tested that.

## Alternative

[`socket2::Type`](https://docs.rs/socket2/0.4.7/socket2/struct.Type.html) deals with this problem by being a newtype wrapper around a `c_int` instead of an `enum`.

## Proof of concept

This works on Linux only.

```rust
use nix::sys::socket::{getsockopt, sockopt};
use std::process::ExitCode;

fn main() -> ExitCode {
	let sockfd = unsafe { libc::socket(libc::AF_PACKET, libc::SOCK_PACKET, 0) };

	if sockfd == -1 {
		eprintln!("Error opening socket: {}", nix::Error::last());
		return ExitCode::FAILURE;
	}

	let socktype = match getsockopt(sockfd, sockopt::SockType) {
		Ok(ok) => ok,
		Err(error) => {
			eprintln!("Error getting socket type: {error}");
			return ExitCode::FAILURE
		}
	};

	eprintln!("The socket's type is: {socktype:?}");
	ExitCode::SUCCESS
}
```

Output:

```
$ target/debug/nix-ub 
The socket's type is: Segmentation fault
```

Note that creating a `SOCK_PACKET` socket is a privileged operation, so the executable needs to be granted that privilege first:

```
$ cargo build
$ sudo setcap CAP_NET_RAW+eip target/debug/nix-ub
```

asomers:
When reading a value into an enum from getsockopt, we must validate it. Failing to do so can lead to UB for example with SOCK_PACKET on Linux.

Perform the validation in GetSockOpt::get.  Currently SockType is the only type that requires validation.

Fixes #1819

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -12,7 +12,7 @@ use libc::{
     self, c_int, c_void, iovec, size_t, socklen_t, CMSG_DATA, CMSG_FIRSTHDR,
     CMSG_LEN, CMSG_NXTHDR,
 };
-use std::convert::TryInto;
+use std::convert::{TryFrom, TryInto};
 use std::io::{IoSlice, IoSliceMut};
 #[cfg(feature = "net")]
 use std::net;
@@ -106,6 +106,24 @@ pub enum SockType {
     /// guarantee ordering.
     #[cfg(not(any(target_os = "haiku")))]
     Rdm = libc::SOCK_RDM,
+}
+// The TryFrom impl could've been derived using libc_enum!.  But for
+// backwards-compatibility with Nix-0.25.0 we manually implement it, so as to
+// keep the old variant names.
+impl TryFrom<i32> for SockType {
+    type Error = crate::Error;
+
+    fn try_from(x: i32) -> Result<Self> {
+        match x {
+            libc::SOCK_STREAM => Ok(Self::Stream),
+            libc::SOCK_DGRAM => Ok(Self::Datagram),
+            libc::SOCK_SEQPACKET => Ok(Self::SeqPacket),
+            libc::SOCK_RAW => Ok(Self::Raw),
+            #[cfg(not(any(target_os = "haiku")))]
+            libc::SOCK_RDM => Ok(Self::Rdm),
+            _ => Err(Errno::EINVAL)
+        }
+    }
 }
 
 /// Constants used in [`socket`](fn.socket.html) and [`socketpair`](fn.socketpair.html)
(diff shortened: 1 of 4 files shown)


src/sys/socket/mod.rs
```diff
@@ -121,6 +121,24 @@ pub enum SockType {
     #[cfg(not(any(target_os = "haiku")))]
     Rdm = libc::SOCK_RDM,
 }
+// The TryFrom impl could've been derived using libc_enum!.  But for
+// backwards-compatibility with Nix-0.25.0 we manually implement it, so as to
```
rtzoeller, line 125: @asomers I'm interpreting this as driving a patch, is that the intent?

rtzoeller (approved):

src/sys/socket/mod.rs
asomers: You mean so we can build patch releases?  Yes, I think that's worthwhile.  But even for new releases too I think it's annoying to force people to change SockType::Stream to SockType::SOCK_STREAM without great reason.

asomers:
I fixed the formatting.  And rather than figure out how to determine whether a Fuchsia process can create raw sockets, I'm just skipping that test on Fuchsia.

asomers:
@ahcodedthat does this patch solve your original problem?

ahcodedthat:
Yes, that works. The example program now just fails cleanly with `EINVAL` instead of segfaulting. 👍

I suggest mentioning in the documentation for `nix::sys::socket::sockopt::SockType` that it will fail with `EINVAL` if the socket type is unknown.

It might also be wise to use a different error code, since the `getsockopt` system call can fail with `EINVAL` for other reasons too. `ESOCKTNOSUPPORT` seems best, but [it doesn't exist on `cfg(target_os = "haiku")`](https://github.com/nix-rust/nix/blob/be36bf05478ac40f92a01e1e8ec6aa11fad65d1f/src/errno.rs#L239), so consider using `EPROTONOSUPPORT` instead.

asomers:
@rtzoeller I stupidly forgot about this PR when I made the 0.26.0 release a few minutes ago.

rtzoeller (approved):
Sometimes it's important to stress test `cargo release` by releasing multiple times in quick succession. Glad you caught it now.

PR #1825, opened 2022-09-23
Add a `sched_getcpu` wrapper

jonas-schievink:


src/sched.rs
```diff
@@ -140,6 +140,13 @@ mod sched_linux_like {
 
         Errno::result(res).map(drop)
     }
+
+    /// Determines the CPU on which the calling thread is running.
+    pub fn sched_getcpu() -> Result<usize> {
```
rtzoeller, line 145: It looks like this is supported on other platforms, including [FreeBSD](https://github.com/freebsd/freebsd-src/blob/8b43388c745ebb42414496f2925f47acac662c2f/include/sched.h#L45) and [DragonFly](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/39b691836ec8de6e84bc13067f04db947dd032f6/sys/sys/sched.h#L84).

Can you move this to the `sched_affinity` section below (since it seems related and has these platforms already?

rtzoeller:
Thanks! Can you modify the existing test [here](https://github.com/nix-rust/nix/blob/a45e9f8fe8e18be48cd40de941d867581e3c6ac0/test/test_sched.rs) to validate that setting the affinity actually has an effect by calling `sched_getcpu`?

jonas-schievink:
Thanks for the review! I've addressed your comments.

The Redox CI failure looks spurious.

jonas-schievink:
Thanks for the review! I've addressed the comments.

rtzoeller:
@jonas-schievink can you squash the current changes into a single commit? Otherwise this looks great.

jonas-schievink:
Okay, I've squashed the commits

rtzoeller (approved):
bors r+

Issue #1826, opened 2022-09-23
Feature request: wrap Linux's `getrandom` syscall


stefins:
```c 
ssize_t getrandom(void *buf, size_t buflen, unsigned int flags);
```

PR #1827, opened 2022-09-23
Added wrapper around `getrandom` syscall

stefins:
Hi,
Added wrapper around Linux's `getrandom` syscall as mentioned in #1826.
```rust
pub fn getrandom(buffer: &mut [u8], mode: RandomMode) -> Result<isize>
```
returns number of bytes written
 
Thanks :)
Signed-off-by: Stefin <stefin@pm.me>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ default = [
   "hostname", "inotify", "ioctl", "kmod", "mman", "mount", "mqueue",
   "net", "personality", "poll", "process", "pthread", "ptrace", "quota",
   "reboot", "resource", "sched", "signal", "socket", "term", "time",
-  "ucontext", "uio", "user", "zerocopy", "getrandom",
+  "ucontext", "uio", "user", "zerocopy", "random",
 ]
 
 acct = []
@@ -66,7 +66,7 @@ ptrace = ["process"]
 quota = []
 process = []
 reboot = []
-getrandom = []
+random = []
 resource = []
 sched = ["process"]
 signal = ["process"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ default = [
   "hostname", "inotify", "ioctl", "kmod", "mman", "mount", "mqueue",
   "net", "personality", "poll", "process", "pthread", "ptrace", "quota",
   "reboot", "resource", "sched", "signal", "socket", "term", "time",
-  "ucontext", "uio", "user", "zerocopy",
+  "ucontext", "uio", "user", "zerocopy", "getrandom",
 ]
 
 acct = []
@@ -66,6 +66,7 @@ ptrace = ["process"]
 quota = []
 process = []
 reboot = []
+getrandom = []
 resource = []
 sched = ["process"]
 signal = ["process"]
--- a/src/sys/getrandom.rs
+++ b/src/sys/getrandom.rs
@@ -1,6 +1,6 @@
 //! Wrapper around getrandom.
 
-use crate::Result;
+use crate::{Errno, Result};
 use libc;
 
 libc_enum! {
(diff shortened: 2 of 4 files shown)


src/sys/getrandom.rs
```diff
@@ -0,0 +1,48 @@
...
+        /// from the urandom source, it blocks if the entropy pool has
+        /// not yet been initialized. If the GRND_NONBLOCK flag is
+        /// set, then getrandom() does not block in these cases
+        GRND_NONBLOCK,
+    }
+}
+
+/// Returns a vector of random bytes
+pub fn getrandom(size: usize, mode: RandomMode) -> Result<Vec<u8>> {
+    let mut buffer = vec![0; size];
```
rtzoeller, line 26: Rather than allocate internally, we should allow the user to pass in a buffer as a mutable slice.

src/sys/getrandom.rs
```diff
@@ -0,0 +1,48 @@
...
+    unsafe {
+        assert_eq!(
```
rtzoeller, line 28: The Linux man page lists a few ways that `getrandom` can fail - we should propagate these up via the `Result`, rather than asserting.

We also will want to expose the number of bytes actually read, since `GRND_NONBLOCK` may cause this not to write all of them.

rtzoeller: Perhaps the signature should be something like

```rust
pub fn getrandom(buffer: &mut [u8], mode: RandomMode) -> Result<usize> 
```

where the result indicates the number of bytes written?

rtzoeller (changes requested):

Cargo.toml
```diff
@@ -66,6 +66,7 @@ ptrace = ["process"]
 quota = []
 process = []
 reboot = []
+getrandom = []
```
rtzoeller, line 69: I'd prefer if this was just "random".

src/sys/getrandom.rs
```diff
@@ -0,0 +1,48 @@
...
+use libc;
+
+libc_enum! {
+    /// How the random bytes should be filled
+    #[repr(u32)]
+    #[non_exhaustive]
+    pub enum RandomMode{
+        /// If  this  bit is set, then random bytes are drawn from the
+        /// random source (i.e., the same source as the /dev/random device)
+        ///  instead of the urandom source.
```
rtzoeller, line 13: We should not copy directly from the man page, which may have license restrictions, but instead summarize.

src/sys/getrandom.rs
```diff
@@ -0,0 +1,48 @@
...
+        /// blocks if no random bytes are available, and when reading
+        /// from the urandom source, it blocks if the entropy pool has
+        /// not yet been initialized. If the GRND_NONBLOCK flag is
+        /// set, then getrandom() does not block in these cases
+        GRND_NONBLOCK,
+    }
+}
+
+/// Returns a vector of random bytes
+pub fn getrandom(size: usize, mode: RandomMode) -> Result<Vec<u8>> {
```
rtzoeller, line 25: It looks like FreeBSD and possibly others support this same (or a similar) system call. We should support these where possible.

rtzoeller:
@stefins actually, is there a compelling reason to include these changes in `nix`, instead of just depending on the `getrandom` crate?

stefins:
@rtzoeller apart from adding a wrapper around `getrandom` this doesn't have any other compelling reason. Since `libc` support `getrandom` I thought this will be a good addition for the `nix` crate

rtzoeller:
> @rtzoeller apart from adding a wrapper around `getrandom` this doesn't have any other compelling reason. Since `libc` support `getrandom` I thought this will be a good addition for the `nix` crate

The `getrandom` crate already provides a wrapper around the `libc` function, including aggregating similar functions from multiple platforms. In this case I'd rather not attempt to duplicate their work, which is excellent.

If a compelling reason comes up for why the `getrandom` crate cannot be used and a separate wrapper is strongly preferable, I'd be happy to see this PR re-opened. I'm going to close it for now though.

PR #1829, opened 2022-09-26
statfs: add namespace fs magic

tammela:
Namespace filesystem magic is missing from FsType constants.


rtzoeller:
Thanks!

@tammela can you add a note to the changelog, linking back to this PR and squashed into the current commit?

rtzoeller:
@tammela it looks like a 0.2.134 release of libc was cut, so this can be updated.

rtzoeller (approved):
bors r+

PR #1837, opened 2022-10-06
backport a test to see if timestamps are working in aarch64

pacak:
Don't merge, this MR is to run the CI to see if we are getting the same errors as in https://github.com/nix-rust/nix/pull/1744

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -565,6 +565,9 @@ mod recvfrom {
             None,
         )
         .expect("send socket failed");
+        setsockopt(rsock, sockopt::ReceiveTimestampns, &true).unwrap();
+        let mut cmsgspace1 = nix::cmsg_space!(nix::sys::time::TimeSpec);
+        let mut cmsgspace2 = nix::cmsg_space!(nix::sys::time::TimeSpec);
 
         let send_thread = thread::spawn(move || {
             for _ in 0..NUM_MESSAGES_SENT {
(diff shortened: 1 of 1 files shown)


rtzoeller:
bors try

rtzoeller:
@pacak the test you're modifying currently builds for a few different platforms, but it looks like `ScmTimestampns` is only defined on Linux. You can ignore the failures on other platforms.

rtzoeller:
Those checkout failures are odd though.

bors retry

pacak:
The idea for this is to check if we can replicate the same problems on aarch64 as in https://github.com/nix-rust/nix/pull/1744 so I'm only interested in one result - aarch64. Or you think it makes sense to merge 
#1837 anyway?

rtzoeller:
No, I wasn't suggesting merging this PR - I just was offering a suggestion to get rid of the other build failures to make it easier to parse the results.

pacak:
If it's important - sure, I'll see what I can do

rtzoeller:
This seems to be passing just fine on my aarch64 target. :/

pacak:
> This seems to be passing just fine on my aarch64 target. :/

:thinking: and the new one still fails? I guess I'll go over the new branch one more time - maybe I'm doing something stupid and just managing to get away with it on non aarch64.

rtzoeller:
> > This seems to be passing just fine on my aarch64 target. :/
> 
> thinking and the new one still fails? I guess I'll go over the new branch one more time - maybe I'm doing something stupid and just managing to get away with it on non aarch64.

I'm going to do another clean build against your new branch just to be sure.... just to make sure something didn't get screwed up during my uptaking of your latest changes. I'll let you know shortly.

rtzoeller:
Yeah, confirmed the other branch (`pacak/master`) is still failing. Tested against the 5.19 kernel.

rtzoeller:
Hmm, it looks like AWS Graviton has a free trial right now (with a fairly high compute limit, at least for this). I wonder if you'd be able to reproduce it there?

pacak:
> I wonder if you'd be able to reproduce it there?

Probably, need to figure out what to do with it first :)

PR #1839, opened 2022-10-08
Fix clippy warnings on nightly

rtzoeller:
Clippy is now smarter about detecting unnecessary casts and useless conversions, which means we need to be more explicit about when the conversions are needed for a subset of platforms.

Required changes found by repeatedly running the following command against a list of the supported platforms.

`xargs -t -I {} sh -c "cargo clippy -Zbuild-std --target {} --all-targets -- -D warnings || exit 255"`

I removed the casts it complained about, and then restored them with an `#[allow]` if a later target needed the cast.

### DIFF
--- a/src/sys/pthread.rs
+++ b/src/sys/pthread.rs
@@ -28,6 +28,7 @@ feature! {
 /// won't send any signal.
 ///
 /// [`pthread_kill(3)`]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_kill.html
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 #[cfg(not(target_os = "redox"))]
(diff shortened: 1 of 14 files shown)


src/sys/pthread.rs
```diff
@@ -28,6 +28,7 @@ feature! {
 /// won't send any signal.
 ///
 /// [`pthread_kill(3)`]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_kill.html
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
```
rtzoeller, line 31: This is the only non-conversion related warning I needed to allow.

asomers (approved):

rtzoeller:
bors r=asomers

PR #1840, opened 2022-10-09
Upgrade parking_lot dev-dependency to 0.12

rtzoeller:
We were previously stuck on 0.11 for MSRV reasons, but since bumping nix's MSRV to Rust 1.56 this is no longer the case.


asomers (approved):

rtzoeller:
bors r=asomers

PR #1841, opened 2022-10-10
add faccessat on illumos

SteveLauC:
This binding has been added in [libc#2953](https://github.com/rust-lang/libc/pull/2953)


asomers (approved):
bors r+

PR #1847, opened 2022-10-14
add listxattr/getxattr/setxattr/removexattr on Linux and Android

SteveLauC:
#### What this PR does:
This PR adds `listxattr`, `getxattr`, `setxattr`, `removexattr` and their variants on Linux and Android

```rust
pub fn listxattr<P: ?Sized + NixPath>(path: &P) -> Result<Vec<OsString>>
pub fn llistxattr<P: ?Sized + NixPath>(path: &P) -> Result<Vec<OsString>>
pub fn flistxattr(fd: RawFd) -> Result<Vec<OsString>>

pub fn getxattr<P, S>(path: &P, name: S) -> Result<Vec<u8>>
pub fn lgetxattr<P, S>(path: &P, name: S) -> Result<Vec<u8>>
pub fn fgetxattr<S>(fd: RawFd, name: S) -> Result<Vec<u8>>

pub fn setxattr<P, S, B>(
    path: &P,
    name: S,
    value: B,
    flags: SetxattrFlag,
) -> Result<()>
pub fn lsetxattr<P, S, B>(
    path: &P,
    name: S,
    value: B,
    flags: SetxattrFlag,
) -> Result<()>
pub fn fsetxattr<S, B>(
    fd: RawFd,
    name: S,
    value: B,
    flags: SetxattrFlag,
) -> Result<()>

pub fn removexattr<P, S>(path: &P, name: S) -> Result<()>
pub fn lremovexattr<P, S>(path: &P, name: S) -> Result<()>
pub fn fremovexattr<S>(fd: RawFd, name: S) -> Result<()>
```

#### man pages

* [listxattr man page](https://man7.org/linux/man-pages/man2/listxattr.2.html)
* [getxattr man page](https://man7.org/linux/man-pages/man2/getxattr.2.html)
* [setxattr man page](https://man7.org/linux/man-pages/man2/lsetxattr.2.html)
* [removexattr man page](https://man7.org/linux/man-pages/man2/removexattr.2.html)

#### What about other platforms

`NetBSD`  should also be added in this PR, but [their current `libc` bindings are wrong](https://github.com/rust-lang/libc/pull/2961).

Update: I just found that `NetBSD` has another collection of `xattr` APIs, which is compatible with `FreeBSD`, and the `libc` bindings are added in [libc#1114](https://github.com/rust-lang/libc/pull/1114). Then I prefer to put these two BSD OSes together.

#### Module
These APIs are placed in `sys/xattr.rs`

#### Tests

Tests are placed in `test/sys/test-xattr.rs`. Symbolic link versions of APIs are not covered.

> If they should be covered, I will add them.

```rust
pub fn llistxattr<P: ?Sized + NixPath>(path: &P) -> Result<Vec<OsString>>
pub fn lgetxattr<P, S>(path: &P, name: S) -> Result<Vec<u8>>
pub fn lsetxattr<P, S>(
    path: &P,
    name: S,
    value: B,
    flags: SetxattrFlag,
) -> Result<()>
pub fn lremovexattr<P, S>(path: &P, name: S) -> Result<()>
```

In the tests, temporary directory are created in the current working directory instead of under `/tmp`, this is needed because we need to test `user` EA, `tmpfs` does not support this, see [this answer](https://stackoverflow.com/a/46598063/14092446).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#1841](https://github.com/nix-rust/nix/pull/1841))
 - Added `eaccess()` on FreeBSD, DragonFly and Linux (glibc and musl).
   ([#1842](https://github.com/nix-rust/nix/pull/1842))
+- Added `listxattr/setxattr/getxattr/removexattr` and their variants on Linux 
+  and Android.
+  ([#1847](https://github.com/nix-rust/nix/pull/1847))
   
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ default = [
   "hostname", "inotify", "ioctl", "kmod", "mman", "mount", "mqueue",
   "net", "personality", "poll", "process", "pthread", "ptrace", "quota",
   "reboot", "resource", "sched", "signal", "socket", "term", "time",
-  "ucontext", "uio", "user", "zerocopy",
+  "ucontext", "uio", "user", "zerocopy", "xattr"
 ]
 
 acct = []
@@ -76,6 +76,7 @@ ucontext = ["signal"]
 uio = []
 user = ["feature"]
 zerocopy = ["fs", "uio"]
+xattr = []
 
 [dev-dependencies]
 assert-impl = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ default = [
   "hostname", "inotify", "ioctl", "kmod", "mman", "mount", "mqueue",
   "net", "personality", "poll", "process", "pthread", "ptrace", "quota",
   "reboot", "resource", "sched", "signal", "socket", "term", "time",
-  "ucontext", "uio", "user", "zerocopy", "xattr"
+  "ucontext", "uio", "user", "xattr", "zerocopy"
 ]
 
 acct = []
@@ -75,8 +75,8 @@ time = []
 ucontext = ["signal"]
 uio = []
 user = ["feature"]
-zerocopy = ["fs", "uio"]
 xattr = []
+zerocopy = ["fs", "uio"]
 
 [dev-dependencies]
 assert-impl = "0.1"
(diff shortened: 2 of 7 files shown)


asomers:
I haven't read the whole PR yet.  But does this have any advantages over the existing xattr crate?

SteveLauC:
> I haven't read the whole PR yet. But does this have any advantages over the existing xattr crate?

Yes, it has some minor advantages, see [Why another crate for EA? Any difference from xattr](https://github.com/SteveLauC/extattr#why-another-crate-for-ea-any-difference-from-xattr).


As you can see, I have made a dedicated crate for this, and have almost all the platforms supported . If you want it to be merged into `nix`, I am happy to contribute. If you don't, feel free to close this PR.

asomers:
> Yes, it has some minor advantages, see [Why another crate for EA? Any difference from xattr](https://github.com/SteveLauC/extattr#why-another-crate-for-ea-any-difference-from-xattr).
> 
> As you can see, I have made a dedicated crate for this, and have almost all the platforms supported. If you want it to be merged into `nix`, I am happy to contribute. If you don't, feel free to close this PR.

And this PR merges the entirety of your extattr crate into Nix?  Are you intending to shut down the other one?

SteveLauC:
> And this PR merges the entirety of your `extattr` crate into `Nix`?

I have no strong view on this, it depends on if `nix` wants it or not. It does not matter whether these bindings exist in `nix` or `extattr` as long as Rust users have such an option. Though I do tend to get this merged into `nix` as `nix` has some great facilities for handling UNIX syscalls.


>  Are you intending to shut down the other one?

If `nix` wants it, I won't shut it down immediately, adding some deprecation messages to let users (if I have) migrate to `nix` would be the option.

asomers:
Ok, but I don't have time to review it now.  Postponing to the next release.

asomers:
I just got back around to reviewing this, because I'm going to make a new release soon.  I looked at your extattr crate, and I like it.  In fact, I think I have another project that could benefit from it.  However, Nix, as you may have noticed, has been getting rather bloated.  And since you've already got this stuff wrapped up in a neatly self-contained crate, I propose that we keep it there rather than merge into Nix.  Do you agree with that?

SteveLauC:
> I just got back around to reviewing this, because I'm going to make a new release soon.

Happy to hear that there will be a new release!

> I propose that we keep it there rather than merge into Nix. Do you agree with that?

Yeah, Nix is indeed some kind of bloated, so i am ok with this proposal, I will close this PR.

PR #1857, opened 2022-11-07
Add better support for unnamed unix socket addrs

### LINKED ISSUE #1585: No longer able to create an unnamed `UnixAddr` as of nix 23.0
It seems that the [`UnixAddr` type](https://docs.rs/nix/0.23.0/nix/sys/socket/struct.UnixAddr.html) no longer provides a way to generate an unnamed unix address (a unix address with an `addrlen` of 2, or `UnixAddrKind::Unnamed`). This used to be possible in [nix 22.0](https://docs.rs/nix/0.22.0/nix/sys/socket/struct.UnixAddr.html) when the internal struct fields were public.

stevenengler:
This adds the following 2 functions/methods: `UnixAddr::new_unnamed` and `UnixAddr::is_unnamed`.

Closes #1585

unix(7) on Linux:

> unnamed: A stream socket that has not been bound to a pathname using bind(2) has no name. Likewise, the two sockets created by socketpair(2) are unnamed. When the address of an unnamed socket is returned, its length is `sizeof(sa_family_t)`, and `sun_path` should not be inspected.

**Edit:** This currently isn't working on BSD, but I see why. Will fix it shortly.

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -1561,7 +1561,6 @@ pub fn test_unnamed_unixdomain() {
 
     let addr_1: UnixAddr = getsockname(fd_1).expect("getsockname failed");
     assert!(addr_1.is_unnamed());
-    assert_eq!(addr_1, UnixAddr::new_unnamed());
 
     close(fd_1).unwrap();
     close(fd_2).unwrap();
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -223,7 +223,7 @@ pub fn test_addr_equality_abstract() {
 }
 
 // Test getting/setting abstract addresses (without unix socket creation)
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "android", target_os = "linux"))]
 #[test]
 pub fn test_abstract_uds_addr() {
     let empty = String::new();
@@ -245,6 +245,7 @@ pub fn test_abstract_uds_addr() {
 }
 
 // Test getting an unnamed address (without unix socket creation)
+#[cfg(any(target_os = "android", target_os = "linux"))]
 #[test]
 pub fn test_unnamed_uds_addr() {
     use crate::nix::sys::socket::SockaddrLike;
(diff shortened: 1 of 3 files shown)


stevenengler:
I decided to only enable these on Linux/Android, since unnamed unix sockets seem to behave differently on FreeBSD and I can't find any documentation for its behaviour. For example when running `getsockname()` on the fd returned from `socketpair()`, it returns an addrlen of 16 bytes while `SUN_LEN(&addr)` returns 2. I'm not sure that the existing code in `UnixAddrKind::get()` handles this correctly either.

The current test failures seem to be unrelated, probably from a new clippy version.

test/sys/test_socket.rs
```diff
@@ -1527,6 +1543,60 @@ pub fn test_unixdomain() {
...
+        AddressFamily::Unix,
+        SockType::Stream,
+        None,
+        SockFlag::empty(),
+    )
+    .expect("socketpair failed");
+
+    let addr_1: UnixAddr = getsockname(fd_1).expect("getsockname failed");
+    assert!(addr_1.is_unnamed());
+    assert_eq!(addr_1, UnixAddr::new_unnamed());
```
asomers, line 1564: Is this part guaranteed to work, according to any Linux documentation, or is it just luck?  I'm not sure we should have this assertion.

asomers (changes requested):
I think that defining this for Linux only is appropriate, because AFAIK no BSD OS documents such a thing.  But don't forget the CHANGELOG.  Also, I'd like to merge this before PR #1871.

asomers:
Also, rebasing should fix the CI failures.

test/sys/test_socket.rs
stevenengler: Linux says "When the address of an unnamed socket is returned, its length is `sizeof(sa_family_t)`", so the only possible unnamed unix socket address it could return would be `sockaddr_un { sun_family: AF_UNIX, ... }` with length 2. But I agree that the documentation around this still isn't great, so I removed this assertion.

asomers (approved):
LGTM, but I violated my own request and merged #1871 first.  So I'll fix the subtle conflict myself.

bors r+

PR #1859, opened 2022-11-10
Implement pidfd APIs for Linux

oxalica:
PID file descriptors are file descriptors that refers to processes. This is a Linux specific API.
The use case is described in [the manual of pidfd_open](https://man7.org/linux/man-pages/man2/pidfd_open.2.html).

Added syscall wrappers:
- `pidfd_open(2)`
- `pidfd_getfd(2)`
- `pidfd_send_signal(2)`


JonathanWoollett-Light:
Seems we had the same idea https://github.com/nix-rust/nix/pull/1868

PR #1860, opened 2022-11-12
Nuke deprecated Errno flags

SUPERCILEX:
@rtzoeller It's been quite a few releases since these have been deprecated, more bloat removal.

PR #1855, opened 2022-11-06
Nuke IoVec

SUPERCILEX:
A release cycle went by... it's be nice to remove some bloat.

Closes https://github.com/nix-rust/nix/issues/1647, closes https://github.com/nix-rust/nix/issues/1371, closes https://github.com/nix-rust/nix/issues/305

PR #1861, opened 2022-11-12
Nuke deprecated net APIs

SUPERCILEX:
This one's a little more questionable since it's only been one release, but it removes a ton of code which is sweet.


rtzoeller:
@SUPERCILEX barring a compelling reason to get this PR and #1855 merged ASAP, I'm tempted to submit #1860 and hold these until after the next release (which I would like to publish after working through the latest series of PRs, which I've admittedly fallen behind on). Practically speaking that would only delay these PRs from being submitted by a few weeks, and since they are largely technical debt I don't see a need for them to land in a release urgently. **Any reason to deviate from that?**

There is still a large percentage of users on versions [earlier than 0.24.0](https://crates.io/crates/nix), so waiting a bit longer feels reasonable.

@asomers how do you feel about a `0.26.0` release in a week or two?

SUPERCILEX:
That sounds super reasonable to me, happy to let these sit until the next release. :)

asomers:
I agree.  Let's do 0.26.0 after #1871 merges, and put this PR into the subsequent release.

rtzoeller:
@SUPERCILEX can you rebase this? I'm happy to submit it whenever it's updated.

SUPERCILEX:
Sweet, done!

SUPERCILEX:
Lol, that was fast. I think the macOS aarch64 job borked itself for unrelated reasons.

rtzoeller:
> Lol, that was fast. I think the macOS aarch64 job borked itself for unrelated reasons.

Yeah, it seems unhappy for reasons unrelated to this PR. I'll re-kick that in a few hours and see if anything has changed.

While we're here though, would it be possible to get a changelog note added to this and #1855? I'd like to be clear that we're gutting the deprecated APIs.

SUPERCILEX:
Oops, yep, fixed!

src/sys/socket/mod.rs
```diff
@@ -32,32 +32,28 @@ pub mod sockopt;
 
 pub use self::addr::{SockaddrLike, SockaddrStorage};
 
-#[cfg(not(any(target_os = "illumos", target_os = "solaris")))]
-#[allow(deprecated)]
-pub use self::addr::{AddressFamily, SockAddr, UnixAddr};
 #[cfg(any(target_os = "illumos", target_os = "solaris"))]
 #[allow(deprecated)]
```
asomers, line 36: Do we still need the `#[allow(deprecated)]` attribute here and below?

SUPERCILEX: Removed. Not quite sure why they were suppressing deprecations anyway.

asomers: Thanks.  I think they were there because `pub use` counts as "using" a deprecated item, even though it's from the same crate.

asomers (approved):
bors r+

Issue #1850, opened 2022-10-22
Change the interface of `openat(dirfd: RawFd, ...)` to `openat(dirfd: Option<RawFd>, ...)`


SteveLauC:
Almost every `xxxat()` function takes an optional `dirfd`, except [`fcntl::openat(2)`](https://docs.rs/nix/latest/nix/fcntl/fn.openat.html) and [`Dir::openat()`](https://docs.rs/nix/latest/nix/dir/struct.Dir.html#method.openat)

```rust
pub fn openat<P: ?Sized + NixPath>(
    dirfd: RawFd,
    path: &P,
    oflag: OFlag,
    mode: Mode,
) -> Result<RawFd> 
```
```rust
pub fn openat<P: ?Sized + NixPath>(dirfd: RawFd, path: &P, oflag: OFlag, mode: sys::stat::Mode) -> Result<Self> {
    let fd = fcntl::openat(dirfd, path, oflag, mode)?;
    Dir::from_fd(fd)
}
```

Any reason why it does not take an `Option<RawFd>`?

SUPERCILEX:
We should just do something similar to https://docs.rs/rustix/latest/src/rustix/fs/cwd.rs.html#26-32 (but as a const) while implementing https://github.com/nix-rust/nix/issues/1750.

asomers:
Git archaeology doesn't show any good reason why the functions take a `RawFd`.  I agree that an Option would be better.

SteveLauC:
Close as completed in #2139

PR #1863, opened 2022-11-13
Convert common syscalls to safe I/O types

SUPERCILEX:
Works towards https://github.com/nix-rust/nix/issues/1750. Fixes https://github.com/nix-rust/nix/issues/1850.

I went the lazy route and pulled the plug... trying to feature gate everything seemed way too complicated.

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -6,6 +6,7 @@ use crate::sys;
 use crate::{Error, NixPath, Result};
 use cfg_if::cfg_if;
 use std::ffi;
+use std::os::unix::io::AsFd;
 use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
 use std::ptr;
 
@@ -39,18 +40,18 @@ impl Dir {
         mode: sys::stat::Mode,
     ) -> Result<Self> {
         let fd = fcntl::open(path, oflag, mode)?;
-        Dir::from_fd(fd)
+        Dir::from_fd(fd.into_raw_fd())
     }
 
     /// Opens the given path as with `fcntl::openat`.
-    pub fn openat<P: ?Sized + NixPath>(
-        dirfd: RawFd,
+    pub fn openat<Fd: AsFd, P: ?Sized + NixPath>(
+        dirfd: &Fd,
         path: &P,
         oflag: OFlag,
         mode: sys::stat::Mode,
     ) -> Result<Self> {
         let fd = fcntl::openat(dirfd, path, oflag, mode)?;
-        Dir::from_fd(fd)
+        Dir::from_fd(fd.into_raw_fd())
     }
 
     /// Converts from a descriptor-based object, closing the descriptor on success or failure.
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -2,11 +2,10 @@
 
 use crate::errno::Errno;
 use crate::fcntl::{self, OFlag};
-use crate::sys;
+use crate::{sys, AsDirFd};
 use crate::{Error, NixPath, Result};
 use cfg_if::cfg_if;
 use std::ffi;
-use std::os::unix::io::AsFd;
 use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
 use std::ptr;
 
@@ -44,8 +43,8 @@ impl Dir {
     }
 
     /// Opens the given path as with `fcntl::openat`.
-    pub fn openat<Fd: AsFd, P: ?Sized + NixPath>(
-        dirfd: &Fd,
+    pub fn openat<Fd: AsDirFd, P: ?Sized + NixPath>(
+        dirfd: Fd,
         path: &P,
         oflag: OFlag,
         mode: sys::stat::Mode,
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -9,6 +9,8 @@ use std::os::unix::io::FromRawFd;
 use std::os::unix::io::OwnedFd;
 use std::os::unix::io::RawFd;
 
+#[cfg(not(target_os = "redox"))]
+use crate::AsDirFd;
 #[cfg(feature = "fs")]
 use crate::{sys::stat::Mode, NixPath, Result};
 #[cfg(any(target_os = "android", target_os = "linux"))]
@@ -211,15 +213,15 @@ pub fn open<P: ?Sized + NixPath>(
 // The conversion is not identical on all operating systems.
 #[allow(clippy::useless_conversion)]
 #[cfg(not(target_os = "redox"))]
-pub fn openat<Fd: AsFd, P: ?Sized + NixPath>(
-    dirfd: &Fd,
+pub fn openat<Fd: AsDirFd, P: ?Sized + NixPath>(
+    dirfd: Fd,
     path: &P,
     oflag: OFlag,
     mode: Mode,
 ) -> Result<OwnedFd> {
     let fd = path.with_nix_path(|cstr| unsafe {
         libc::openat(
-            dirfd.as_fd().as_raw_fd(),
+            dirfd.as_dir_fd(),
             cstr.as_ptr(),
             oflag.bits(),
             mode.bits() as c_uint,
@@ -230,18 +232,18 @@ pub fn openat<Fd: AsFd, P: ?Sized + NixPath>(
 }
 
 #[cfg(not(target_os = "redox"))]
-pub fn renameat<Fd1: AsFd, P1: ?Sized + NixPath, Fd2: AsFd, P2: ?Sized + NixPath>(
-    old_dirfd: &Fd1,
+pub fn renameat<Fd1: AsDirFd, P1: ?Sized + NixPath, Fd2: AsDirFd, P2: ?Sized + NixPath>(
+    old_dirfd: Fd1,
     old_path: &P1,
-    new_dirfd: &Fd2,
+    new_dirfd: Fd2,
     new_path: &P2,
 ) -> Result<()> {
     let res = old_path.with_nix_path(|old_cstr| {
         new_path.with_nix_path(|new_cstr| unsafe {
             libc::renameat(
-                old_dirfd.as_fd().as_raw_fd(),
+                old_dirfd.as_dir_fd(),
                 old_cstr.as_ptr(),
-                new_dirfd.as_fd().as_raw_fd(),
+                new_dirfd.as_dir_fd(),
                 new_cstr.as_ptr(),
             )
         })
@@ -264,19 +266,19 @@ libc_bitflags! {
 feature! {
 #![feature = "fs"]
 #[cfg(all(target_os = "linux", target_env = "gnu"))]
-pub fn renameat2<Fd1: AsFd, P1: ?Sized + NixPath, Fd2: AsFd, P2: ?Sized + NixPath>(
-    old_dirfd: &Fd1,
+pub fn renameat2<Fd1: AsDirFd, P1: ?Sized + NixPath, Fd2: AsDirFd, P2: ?Sized + NixPath>(
+    old_dirfd: Fd1,
     old_path: &P1,
-    new_dirfd: &Fd2,
+    new_dirfd: Fd2,
     new_path: &P2,
     flags: RenameFlags,
 ) -> Result<()> {
     let res = old_path.with_nix_path(|old_cstr| {
         new_path.with_nix_path(|new_cstr| unsafe {
             libc::renameat2(
-                old_dirfd.as_fd().as_raw_fd(),
+                old_dirfd.as_dir_fd(),
                 old_cstr.as_ptr(),
-                new_dirfd.as_fd().as_raw_fd(),
+                new_dirfd.as_dir_fd(),
                 new_cstr.as_ptr(),
                 flags.bits(),
             )
@@ -404,10 +406,10 @@ pub fn readlink<P: ?Sized + NixPath>(path: &P) -> Result<OsString> {
 
 #[cfg(not(target_os = "redox"))]
 pub fn readlinkat<Fd: AsFd, P: ?Sized + NixPath>(
-    dirfd: &Fd,
+    dirfd: Fd,
     path: &P,
 ) -> Result<OsString> {
-    inner_readlink(Some(dirfd), path)
+    inner_readlink(Some(&dirfd), path)
 }
 }
 
@@ -623,9 +625,9 @@ feature! {
 /// returned.
 #[cfg(any(target_os = "android", target_os = "linux"))]
 pub fn copy_file_range<Fd1: AsFd, Fd2: AsFd>(
-    fd_in: &Fd1,
+    fd_in: Fd1,
     off_in: Option<&mut libc::loff_t>,
-    fd_out: &Fd2,
+    fd_out: Fd2,
     off_out: Option<&mut libc::loff_t>,
     len: usize,
 ) -> Result<usize> {
@@ -652,9 +654,9 @@ pub fn copy_file_range<Fd1: AsFd, Fd2: AsFd>(
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub fn splice<Fd1: AsFd, Fd2: AsFd>(
-    fd_in: &Fd1,
+    fd_in: Fd1,
     off_in: Option<&mut libc::loff_t>,
-    fd_out: &Fd2,
+    fd_out: Fd2,
     off_out: Option<&mut libc::loff_t>,
     len: usize,
     flags: SpliceFFlags,
@@ -681,8 +683,8 @@ pub fn splice<Fd1: AsFd, Fd2: AsFd>(
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub fn tee<Fd1: AsFd, Fd2: AsFd>(
-    fd_in: &Fd1,
-    fd_out: &Fd2,
+    fd_in: Fd1,
+    fd_out: Fd2,
     len: usize,
     flags: SpliceFFlags,
 ) -> Result<usize> {
@@ -699,7 +701,7 @@ pub fn tee<Fd1: AsFd, Fd2: AsFd>(
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub fn vmsplice<Fd: AsFd>(
-    fd: &Fd,
+    fd: Fd,
     iov: &[std::io::IoSlice<'_>],
     flags: SpliceFFlags,
 ) -> Result<usize> {
@@ -758,7 +760,7 @@ feature! {
 #[cfg(any(target_os = "linux"))]
 #[cfg(feature = "fs")]
 pub fn fallocate<Fd: AsFd>(
-    fd: &Fd,
+    fd: Fd,
     mode: FallocateFlags,
     offset: libc::off_t,
     len: libc::off_t,
@@ -834,7 +836,7 @@ impl SpacectlRange {
 /// ```
 #[cfg(target_os = "freebsd")]
 pub fn fspacectl<Fd: AsFd>(
-    fd: &Fd,
+    fd: Fd,
     range: SpacectlRange,
 ) -> Result<SpacectlRange> {
     let mut rqsr = libc::spacectl_range {
@@ -885,7 +887,7 @@ pub fn fspacectl<Fd: AsFd>(
 /// ```
 #[cfg(target_os = "freebsd")]
 pub fn fspacectl_all<Fd: AsFd>(
-    fd: &Fd,
+    fd: Fd,
     offset: libc::off_t,
     len: libc::off_t,
 ) -> Result<()> {
@@ -941,7 +943,7 @@ mod posix_fadvise {
     feature! {
     #![feature = "fs"]
     pub fn posix_fadvise<Fd: AsFd>(
-        fd: &Fd,
+        fd: Fd,
         offset: libc::off_t,
         len: libc::off_t,
         advice: PosixFadviseAdvice,
@@ -974,7 +976,7 @@ mod posix_fadvise {
     target_os = "freebsd"
 ))]
 pub fn posix_fallocate<Fd: AsFd>(
-    fd: &Fd,
+    fd: Fd,
     offset: libc::off_t,
     len: libc::off_t,
 ) -> Result<()> {
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1,9 +1,12 @@
 use crate::errno::Errno;
 use libc::{self, c_char, c_int, c_uint, size_t, ssize_t};
 use std::ffi::OsString;
-#[cfg(not(target_os = "redox"))]
-use std::os::raw;
 use std::os::unix::ffi::OsStringExt;
+use std::os::unix::io::AsFd;
+#[cfg(not(target_os = "redox"))]
+use std::os::unix::io::AsRawFd;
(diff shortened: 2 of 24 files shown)


asomers:
There's a lot of stuff here.  Could you please annotate #1750 to say which modules you're converting?

src/unistd.rs
```diff
@@ -478,9 +491,8 @@ fn dup3_polyfill(oldfd: RawFd, newfd: RawFd, flags: OFlag) -> Result<RawFd> {
 /// pages for additional details on possible failure cases.
 #[inline]
 pub fn chdir<P: ?Sized + NixPath>(path: &P) -> Result<()> {
-    let res = path.with_nix_path(|cstr| {
-        unsafe { libc::chdir(cstr.as_ptr()) }
-    })?;
+    let res =
```
asomers, line 494: This looks like purely a formatting change.  Could you please separate formatting changes into a separate commit?

SUPERCILEX: Sounds good, done: https://github.com/nix-rust/nix/pull/1908. Some background: to edit this code, I would temporarily remove the `feature!` macro wrappers which meant everything got reformatted.

asomers: Are you using some kind of fancy editor, yet one that doesn't understand feature flags?

SUPERCILEX: Nah, this one is due to https://github.com/rust-lang/rustfmt/issues/3253. Which actually, I wonder if we can just move the `#![$meta:meta]` into the macro call so that formatting will finally work properly. Lemme try that.

SUPERCILEX: Tried it, didn't work. :(

SUPERCILEX:
Wow that was painful. :sweat_smile: I think this is finally ready for review.

src/unistd.rs
```diff
@@ -438,37 +452,43 @@ pub fn dup(oldfd: RawFd) -> Result<RawFd> {
 /// specified fd instead of allocating a new one.  See the man pages for more
 /// detail on the exact behavior of this function.
 #[inline]
-pub fn dup2(oldfd: RawFd, newfd: RawFd) -> Result<RawFd> {
-    let res = unsafe { libc::dup2(oldfd, newfd) };
+pub fn dup2<Fd: AsFd>(oldfd: &Fd, newfd: &mut OwnedFd) -> Result<()> {
```
asomers, line 455: Clever, taking `OwnedFd` by borrow.  But what if the user wants to supply an integer that doesn't correspond to any currently open file?  That was possible with the old interface.

src/unistd.rs
```diff
@@ -438,37 +452,43 @@ pub fn dup(oldfd: RawFd) -> Result<RawFd> {
...
-    Errno::result(res)
+    Errno::result(res).map(drop)
 }
 
 /// Create a new copy of the specified file descriptor using the specified fd
 /// and flags (see [dup(2)](https://man7.org/linux/man-pages/man2/dup.2.html)).
 ///
 /// This function behaves similar to `dup2()` but allows for flags to be
 /// specified.
-pub fn dup3(oldfd: RawFd, newfd: RawFd, flags: OFlag) -> Result<RawFd> {
```
asomers, line 452: Sigh, we still need to get rid of this function.  It isn't Nix's place to be a C library.

src/unistd.rs
```diff
@@ -527,8 +546,8 @@ pub fn fchdir(dirfd: RawFd) -> Result<()> {
 /// ```
 #[inline]
 pub fn mkdir<P: ?Sized + NixPath>(path: &P, mode: Mode) -> Result<()> {
-    let res = path.with_nix_path(|cstr| {
-        unsafe { libc::mkdir(cstr.as_ptr(), mode.bits() as mode_t) }
+    let res = path.with_nix_path(|cstr| unsafe {
```
asomers, line 549: More formatting-only changes here.

src/unistd.rs
```diff
@@ -1063,32 +1110,28 @@ pub fn gethostname() -> Result<OsString> {
...
+/// use std::os::unix::io::AsFd;
 /// use nix::unistd::close;
 ///
 /// let f = tempfile::tempfile().unwrap();
-/// close(f.into_raw_fd()).unwrap(); // Good.  into_raw_fd consumes f
+/// close(f).unwrap();
 /// ```
-pub fn close(fd: RawFd) -> Result<()> {
-    let res = unsafe { libc::close(fd) };
+pub fn close<Fd: IntoRawFd>(fd: Fd) -> Result<()> {
```
asomers, line 1120: Nice.

asomers (changes requested):
I haven't reviewed the tests yet.  But the src changes look mostly good.  The one big problem I see is that you removed the Option from all of the *at functions.  I don't think we want to do that.

src/dir.rs
```diff
@@ -39,18 +40,18 @@ impl Dir {
         mode: sys::stat::Mode,
     ) -> Result<Self> {
         let fd = fcntl::open(path, oflag, mode)?;
-        Dir::from_fd(fd)
+        Dir::from_fd(fd.into_raw_fd())
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, absolutely. I just didn't have the energy to do it in this PR. :)

src/fcntl.rs
```diff
@@ -193,42 +196,52 @@ feature! {
...
-    Errno::result(fd)
+
+    Errno::result(fd).map(|fd| unsafe { OwnedFd::from_raw_fd(fd) })
 }
 
 #[cfg(not(target_os = "redox"))]
-pub fn renameat<P1: ?Sized + NixPath, P2: ?Sized + NixPath>(
-    old_dirfd: Option<RawFd>,
+pub fn renameat<Fd1: AsFd, P1: ?Sized + NixPath, Fd2: AsFd, P2: ?Sized + NixPath>(
+    old_dirfd: &Fd1,
```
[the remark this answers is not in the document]
SUPERCILEX: What was the rationale for using options anyway when you could use `libc::AT_FDCWD`? The intended usage is that people will pass in a `nix::AT_FDCWD` instead of None.

src/lib.rs
```diff
@@ -182,6 +184,17 @@ pub type Result<T> = result::Result<T, Errno>;
 /// ones.
 pub type Error = Errno;
 
+/// A file descriptor representing the current working directory.
+#[cfg(not(target_os = "redox"))]
+pub const AT_FDCWD: &BorrowedFd<'static> = unsafe {
```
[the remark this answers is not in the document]
SUPERCILEX: > Why

Makes the API clearer IMO. Then you don't need to constantly wrap your FDs with Some.

> For example, there are many functions that can't accept AT_FDCWD.

WDYM? All the old option functions just converted None to AT_FDCWD AFAIK.

src/pty.rs
```diff
@@ -81,13 +81,15 @@ impl Drop for PtyMaster {
 
 impl io::Read for PtyMaster {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        unistd::read(self.0, buf).map_err(io::Error::from)
+        unistd::read(unsafe { &BorrowedFd::borrow_raw(self.0) }, buf)
```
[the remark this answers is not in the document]
SUPERCILEX: Yup.

Also just want to make sure we're on the same page: your wording suggests I'll be writing all the PRs for I/O safety, but I won't have the time to do that. I can definitely put in a few more conversions (next week ish), but other people will need to take the torch. :)

src/unistd.rs
SUPERCILEX: Yeah, I tried having it use libc::dup3, but that's only available on certain platforms. Should I cfg it to just those platforms?

src/unistd.rs
SUPERCILEX: Dang, you're right. Will fix.

src/unistd.rs
SUPERCILEX: This may actually be wrong come to think of it. I think this lets you pass in an owned type which will double close. Need to double check. Solution is to ManuallyDrop before close.

src/fcntl.rs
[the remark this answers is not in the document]
asomers: It's a lot Rustier.  In particular, most functions that accept `RawFd` can't accept `libc::AT_FDCWD` .  Putting `Option` in the signature makes it clear which do.  If anything, the `Option` is even more helpful now that we have I/O safety.

src/unistd.rs
asomers: Let's do it in a separate PR.

src/unistd.rs
asomers: No, this is right.  `into_raw_fd` transfers ownership.
https://doc.rust-lang.org/stable/std/os/unix/io/trait.IntoRawFd.html

SUPERCILEX: Whew, forgot that `into_raw_fd` takes self by value. Then I agree, nice. :)

src/lib.rs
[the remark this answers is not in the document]
SUPERCILEX: Ohhh, excellent point! I'm 100% onboard the "express constraint with the type system" bandwagon, but I still don't like using Option and I think this is a great opportunity to revisit this. What about adding a trait (maybe called `IntoPathFd`) implemented for all `AsFd` and a special `Cwd` marker struct. The interface would be `fn as_raw_fd(&self) -> RawFd`. Then open would be `fn open<Fd: IntoPathFd>(fd: Fd, ...)` and write would be `fn write<Fd: AsFd>(fd: Fd, ...)`. `write(Cwd, ...)` would not compile while `open(Cwd, ...)` would. WDYT?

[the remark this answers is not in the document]
SUPERCILEX: Flushed it out a bit, this seems awesome:
```rust
#![allow(dead_code, unused_variables)]
use std::fs::File;

trait IntoPathFd {
    fn as_raw_fd(&self) -> RawFd;
}

impl<Fd: AsFd> IntoPathFd for Fd {
    fn as_raw_fd(&self) -> RawFd {
        AsRawFd::as_raw_fd(&self.as_fd())
    }
}

struct Cwd;

impl IntoPathFd for Cwd {
    fn as_raw_fd(&self) -> RawFd {
        -1
    }
}

fn open(fd: impl IntoPathFd) {}
fn write(fd: impl AsFd) {}

fn foo() {
    open(&File::open("").unwrap());
    open(Cwd);
    // write(Cwd);
    write(&File::open("").unwrap());
}
```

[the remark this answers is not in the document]
SteveLauC: > WDYM? All the old option functions just converted None to AT_FDCWD AFAIK.

When the `path` argument is an absolute path, `dirfd` will be ignored. So if I wanna pass an absolute path to those `xxxat()` functions, in the previous implementation:

```rust
xxxat(None, "/absolute/path", ...)
```

What should I pass to the first argument with the new interface:

```rust
fn xxxat<Fd: AsFd>(dirfd: &Fd, ...)
```

`AT_FDCWD`? 
```rust
xxxat(AT_FDCWD, "/absolute/path", ...)
```
I know the value of actual `dirfd` passed into the function does not matter, but the **semantics** has been changed, and becomes a little bit weird to the user IMO...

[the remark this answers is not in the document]
SUPERCILEX: I personally think you should be passing in AT_FDCWD since if your absolute path turns out to be relative, cwd is what's going to be used. That said, with the design I presented above, we could just add a type called `Ignored` or even implement IntoPathFd for options so you can pass in None.

SUPERCILEX:
Also at @asomers any chance this PR can be prioritized over the other I/O safety ones? Otherwise I'm going to be stuck constantly resolving merge conflicts.

asomers (commented):
There's another possibility for the *at functions.  The cap-std crate treats them as methods on an already-created file descriptor object, like this:
```rust
let etc = Dir::open_ambient_dir("/etc", ambient_authority())?;
let passwd = etc.open("passwd")?;
```
I think that's better than any of the proposals here.  The only problems with cap-std are that it's too big.  It pulls in too many dependencies, and it tries to provide a platform-independent abstraction layer.  Should we do something similar?

src/lib.rs
```diff
@@ -182,6 +183,30 @@ pub type Result<T> = result::Result<T, Errno>;
 /// ones.
 pub type Error = Errno;
 
+/// A trait that models requirements for path file descriptors.
```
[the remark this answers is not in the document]
SUPERCILEX: Done.

src/lib.rs
```diff
@@ -182,6 +183,30 @@ pub type Result<T> = result::Result<T, Errno>;
...
+    fn as_path_fd(&self) -> RawFd;
+}
+
+impl<Fd: AsFd> AsPathFd for Fd {
+    fn as_path_fd(&self) -> RawFd {
+        self.as_fd().as_raw_fd()
+    }
+}
+
+/// The `AT_FDCWD` marker FD.
```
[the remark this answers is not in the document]
SUPERCILEX: Done.

SUPERCILEX:
> There's another possibility for the *at functions. The cap-std crate treats them as methods on an already-created file descriptor object

I had considered this a while back and think the conclusion was that I didn't like it because it'd pollute the i32 namespace. Now it would pollute `AsFd`s (which is definitely better than all i32s, but still). Over at [rustix](https://github.com/bytecodealliance/rustix/issues/478#issuecomment-1345036331), we're talking about potentially having a `DirFd` type to also prevent using FDs opened for files instead of directories, but I need to do some more research. That would combine nicely with extension methods.

Not sure where that leaves this PR... maybe we can keep `AsDirFd` and punt the discussion for now or I could try pulling the *at changes out of this PR?

asomers:
> > There's another possibility for the *at functions. The cap-std crate treats them as methods on an already-created file descriptor object
> 
> I had considered this a while back and think the conclusion was that I didn't like it because it'd pollute the i32 namespace. Now it would pollute `AsFd`s (which is definitely better than all i32s, but still). Over at [rustix](https://github.com/bytecodealliance/rustix/issues/478#issuecomment-1345036331), we're talking about potentially having a `DirFd` type to also prevent using FDs opened for files instead of directories, but I need to do some more research. That would combine nicely with extension methods.
> 
> Not sure where that leaves this PR... maybe we can keep `AsDirFd` and punt the discussion for now or I could try pulling the *at changes out of this PR?

Yeah, I wasn't thinking about adding a method to `AsFd`, but about adding a newtype for open directories.  However, that would be suboptimal on Illumos, where you can use `openat` on a regular file to edit extended attributes.  I see that you're a Rustix contributor too.  So I've got to ask, why use both Nix and Rustix?  I've generally thought of them as providing equivalent functionality.  And also, why bother with both?

SUPERCILEX:
> Yeah, I wasn't thinking about adding a method to AsFd, but about adding a newtype for open directories.

Gotya, cool. I'm planning on playing around with that tomorrow. Regarding Illumos, I'd prefer doing the nice thing for everybody else and adding a cfg exception for them.

> So I've got to ask, why use both Nix and Rustix?

A month ago, I just used Nix, but over the past few weeks, I've fully migrated over to rustix for the perf improvements of raw syscalls. So why stick around? Originally, it was because I had open PRs here and it felt wrong to close everything and be like, "So long, suckers!" :laughing: Anyway, now I'm sticking around because between you and Dan we're generating some interesting ideas that I don't think would have come about without the cross pollination.

notgull:
This smells like it should be a newtype to me. Make it implement AsFd or From<OwnedFd> and it should work.

src/unistd.rs
```diff
@@ -436,37 +450,65 @@ pub fn dup(oldfd: RawFd) -> Result<RawFd> {
...
-    let res = unsafe { libc::dup2(oldfd, newfd) };
+pub fn dup2<Fd: AsFd>(oldfd: Fd, newfd: &mut OwnedFd) -> Result<()> {
+    let res =
+        unsafe { libc::dup2(oldfd.as_fd().as_raw_fd(), newfd.as_raw_fd()) };
+
+    Errno::result(res).map(drop)
+}
+
+#[inline]
+pub fn dup2_raw<Fd1: AsFd, Fd2: AsRawFd + IntoRawFd>(
```
ids1024, line 461: If this operates on a `RawFd` it should be `unsafe`. Otherwise it could be passed a fd corresponding to a file open elsewhere, and break soundness assumptions in the code using  it.

SUPERCILEX: No, that's why it requires `IntoRawFd`. If the call fails, you had to pass in your FD by value which means it gets dropped inside `dup2_raw`. If it succeeds, we give you back the same fd. Of course if you don't want the safety stuff, then you pass in a `RawFd` and `IntoRawFd` will do nothing.

SUPERCILEX:
@asomers see https://github.com/bytecodealliance/rustix/issues/478#issuecomment-1355851691. I want to try your idea, but not in this PR.

Regarding this PR: please let me know what needs to change for it to be merged (I'll resolve conflicts then). I think AsDirFd should be left in, though I'd be willing to temporarily rip it out. I can also drop the dup commit, but I think as long as the commits are merged as is this PR is fine and we can just do fixups before release if necessary.

Regarding AsDirFd, I'll make a new PR after this one is merged that replaces it with the idea discussed here and in rustix. Regarding making openat an extension method, I think that's a bad idea: there's no symmetry to be found. If we make openat an extension method, why stop there? Open can be an extension on paths. Close can be an extension on FDs. But then what do you do about methods that have multiple paths/fds or mix the two? The whole thing becomes a mess, so let's just keep it simple and use top-level functions.

src/unistd.rs
ids1024: In the current Rust version `RawFd` implements `AsRawFd` and `IntoRawFd`, so this API can easily be used with an arbitrary integer. So you could for instance call `close(dup2_raw(some_fd, 0).unwrap())`. Allowing safe code to close stdin or any other fd. This shouldn't be possible in safe code without an `OwnedFd`.

This function is only safe if the caller either owns fd2, or knows the fd isn't used.

So I think the function should be an `unsafe fn`, with a doc comment describing when it is safe.

SUPERCILEX: Maybe? When to require unsafe is unclear IMO. BTW, your example can just be `close(0)`—no need for dup. I think our goal should be to not break BorrowedFd or OwnedFd. If you want to do weird stuff with RawFd, then I feel like that's up to you.

ids1024: https://rust-lang.github.io/rfcs/3128-io-safety.html

> Functions accepting arbitrary raw I/O handle values ([RawFd](https://doc.rust-lang.org/stable/std/os/unix/io/type.RawFd.html), [RawHandle](https://doc.rust-lang.org/stable/std/os/windows/io/type.RawHandle.html), or [RawSocket](https://doc.rust-lang.org/stable/std/os/windows/io/type.RawSocket.html)) should be unsafe if they can lead to any I/O being performed on those handles through safe APIs.

Basically any function that can operate on a `RawFd` should be unsafe. So to conform to IO safety, `close` should either be unsafe, or it should take a `OwnedFd` and consume it.

This is definitely awkward for a crate like `nix` though, since conforming to this necessitates breaking changes to basically all of the API. Perhaps these functions in `nix` should have been unsafe already, but ideally Rust would have had clear rules about IO safety before 1.0.

SUPERCILEX: > This is definitely awkward for a crate like nix though

Yeah, I think that's the crux of the matter. Nix is trying to be more low-level than the stdlib, so requiring unsafe is probably going to be annoying. I also think you can argue that we're following the definition. It says "can lead to" which to me means that if you store the RawFd somewhere and then use it later (like in a BorrowedFd or File), the function that does the storing should be marked unsafe, but if you immediately use the RawFd then you're good. Otherwise anytime a RawFd is even mentioned you'd need unsafe which doesn't make sense to me.

All of that said, I wouldn't be opposed to marking `close` as unsafe. Interestingly enough, dup is perfectly safe according to "Rust code has I/O safety if it's not possible for that code to cause invalid I/O operations" because anything you throw at dup will work (if the fd doesn't exist, it'll be allocated, if it does exist, it'll be closed).

ids1024: Other than the breaking API changes, and interop with other things using `RawFd`, it seems like it should be fairly ergonomic to work with a safe API that uses `OwnedFd`/`BorrowedFd`/`AsFd` and mostly doesn't expose `RawFd` anywhere in the API (`dup2_raw` is a special case here).

> Interestingly enough, dup is perfectly safe according to "Rust code has I/O safety if it's not possible for that code to cause invalid I/O operations"Interestingly enough, dup is perfectly safe according to "Rust code has I/O safety if it's not possible for that code to cause invalid I/O operations"

This is subtle, but a safe `dup` taking a `RawFd` and returning an `OwnedFd` could be used to trigger memory unsafety with the sort of things envisioned in the IO safety RFC. In particular:

> And in specialized circumstances, violating I/O safety could even lead to violating memory safety. For example, in theory it should be possible to make a safe wrapper around an mmap of a file descriptor created by Linux's [memfd_create](https://man7.org/linux/man-pages/man2/memfd_create.2.html) system call and pass &[u8]s to safe Rust, since it's an anonymous open file which other processes wouldn't be able to access. However, without I/O safety, and without permanently sealing the file, other code in the program could accidentally call write or ftruncate on the file descriptor, breaking the memory-safety invariants of &[u8].

So safe code shouldn't be able to `write` or `truncate` an arbitrary fd, which could break immutability assumptions of something mmapping that fd. `dup` is also a problem, because safe code could write or truncate given an `OwnedFd`, which will have the same issue as write or truncate to the original fd.

SUPERCILEX: Sure, but then everything is unsafe because you can just open `/proc/self/mem`. I think we have to take a pragmatic approach here and making dup unsafe because someone might be messing around with mmap is overkill.

src/fcntl.rs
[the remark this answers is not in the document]
SteveLauC: >  In particular, most functions that accept RawFd can't accept libc::AT_FDCWD

true

But one drawback of using `Option<Fd>` is that one has to specify 2 concrete types when passing `None`:

```rs
renameat::<?, ?>(None, oldpath, None, newpath)
```

src/unistd.rs
SteveLauC: > so I ended up adding dup*_raw variants.

This is something unfortunate, but I think this unfortunate approach is the only way to do it, rustix does not provide something similar to `unsafe fn dup2_raw()`, so one cannot specify a fd that is still not open.

And interestingly, rustix provides 3 helper functions to use `dup2` with `Stdin`, `Stdout` and `Stderr`: https://docs.rs/rustix/latest/rustix/stdio/fn.dup2_stdin.html

src/unistd.rs
SteveLauC: Yeah, I think `dup2_raw()` should be unsafe, it is pretty easy to trigger double close with:


```rs
let file = File::open();
let owned_fd = dup2_raw(fd, file.as_raw_fd());
// double close
```

src/unistd.rs
SteveLauC: > And interestingly, rustix provides 3 helper functions to use dup2 with Stdin, Stdout and Stderr: https://docs.rs/rustix/latest/rustix/stdio/fn.dup2_stdin.html

Now I think this is indeed necessary

SteveLauC:
Close as this has been superseded by other I/O safety PRs: https://github.com/nix-rust/nix/tree/d60f710c31394b3ee7af2a1be0f37a713a9565bd/changelog

PR #1867, opened 2022-11-16
Add routing socket type on macOS

pinkisemils:
This is a small change to add the routing socket type to the list of socket types one can open with `nix`. I've added a smoke test to see that a socket of such type can actually be opened, but I'm not sure if such a test belongs in the codebase here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Add `PF_ROUTE` to `SockType` on macOS, iOS, all of the BSDs, Fuchsia, Haiku, Illumos.
+  ([#1867](https://github.com/nix-rust/nix/pull/1867))
+
 ### Changed
 ### Fixed
 ### Removed
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -80,6 +80,9 @@ pub enum AddressFamily {
     #[cfg(any(target_os = "android", target_os = "linux"))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
     Netlink = libc::AF_NETLINK,
+    /// Kernel interface for interacting with the routing table
+    #[cfg(not(any(target_os = "redox", target_os = "linux", target_os = "android")))]
+    Route = libc::PF_ROUTE,
     /// Low level packet interface (see [`packet(7)`](https://man7.org/linux/man-pages/man7/packet.7.html))
     #[cfg(any(
         target_os = "android",
@@ -421,6 +424,8 @@ impl AddressFamily {
             libc::AF_NETLINK => Some(AddressFamily::Netlink),
             #[cfg(any(target_os = "macos", target_os = "macos"))]
             libc::AF_SYSTEM => Some(AddressFamily::System),
+            #[cfg(not(any(target_os = "redox", target_os = "linux", target_os = "android")))]
+            libc::PF_ROUTE => Some(AddressFamily::Route),
             #[cfg(any(target_os = "android", target_os = "linux"))]
             libc::AF_PACKET => Some(AddressFamily::Packet),
             #[cfg(any(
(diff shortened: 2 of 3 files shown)


asomers (commented):
I think this is reasonable.  But how does one use a routing socket?  Do you need to `bind` and/or `connect` with it?  If so, then Nix support won't be complete without those.  Also, it needs a CHANGELOG entry.

pinkisemils:
There's no need to connect or bind to anything - once the socket is opened, it can send and receive messages to the kernel to read or write to the routing table. It's used in the `route` CLI, e.g. `route get 8.8.8.8` or `route monitor`.  There's no reason to `connect` or `bind` the socket as it can only ever be connected to the kernel to talk to/about the routing table, and it is ready to send and receive messages as soon as it's created via `socket`. 

Having used this for a bit, it seems that there are also some `sockaddr_x` types that are missing, I'll see about adding those too in a separate PR.

pinkisemils:
I now realize that these changes might also be relevant for FreeBSD and friends - but I don't have a FreeBSD VM handy to test my changes.

asomers:
It looks to me like you should enable this for macos, ios, freebsd, dragonfly, openbsd, netbsd, linux, android, illumos, haiku, and fuchsia.  In fact, it's everywhere but Redox.  But on Linux and Android, it's an alias for AF_NETLINK.  Why don't you try enabling it for all of those, and let CI test it for you?  We have full testing in CI for FreeBSD, OSX, and Linux.

pinkisemils:
It seems that it does _just work_ on most OS'es. I elected to not define it on Linux, since the same value is already used for netlink sockets (and they are kind of the same kind of a thing). Should I remove the test that tries to open the routing socket, or should that stay? I think it doesn't really test any logic, just that a socket of this type can be opened.

asomers:
I think it's ok to leave the test.  It helps us validate which platforms have routing sockets.  I'll create another PR to fix the clippy failure.

asomers (approved):
The CI failure on uclibc is fixed on master.

bors r+

pinkisemils:
I've rebased and it passes all the checks now.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `MntFlags` and `unmount` on all of the BSDs.
 - Added `any()` and `all()` to `poll::PollFd`.
   ([#1877](https://github.com/nix-rust/nix/pull/1877))
+- Add `PF_ROUTE` to `SockType` on macOS.
```
asomers, line 25: You've got a duped line here.
```suggestion
```

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
...
+- Add `PF_ROUTE` to `SockType` on macOS, all of the BSDs and Linux.
```
asomers, line 26: This is incorrect, because you _didn't_ add it to Linux.

asomers (changes requested):

CHANGELOG.md
pinkisemils: Ah, you are right. I've updated it now and I do apologize for the sloppy push.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `MntFlags` and `unmount` on all of the BSDs.
 - Added `any()` and `all()` to `poll::PollFd`.
   ([#1877](https://github.com/nix-rust/nix/pull/1877))
+- Add `PF_ROUTE` to `SockType` on macOS, all of the BSDs.
```
asomers, line 25:
```suggestion
- Add `PF_ROUTE` to `SockType` on macOS and the BSDs.
```

asomers: But isn't this still inaccurate?  Because you also added it for iOS, illumos, Haiku, and Fuchsia, I believe.

pinkisemils: You are correct, I also moved the entry up the changelog, since these changes have not been released anywhere yet.

asomers: I think you forgot to push?

asomers (approved):
bors r+

dlon:
@asomers Would you be able to cut a release that includes this patch? We're relying on it now but would like to get rid of the git dependency.

PR #1870, opened 2022-11-19
mmap addr

JonathanWoollett-Light:
Uses `Some<size_t>` instead of `*mut c_void` for the `addr` passed to [`sys::mman::mmap`](https://docs.rs/nix/latest/nix/sys/mman/fn.mmap.html).

In this instance we are not usefully passing a pointer, it will never be dereferenced. We are passing a location which represents where to attach the shared memory to.

In this case `size_t` better represents an address and not a pointer, and `Option<size_t>` better represents an optional argument than `NULLPTR`.

In C since there is no optional type this is a pointer as this allows it be null which is an alias here for `None`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - The MSRV is now 1.56.1
   ([#1792](https://github.com/nix-rust/nix/pull/1792))
+- The `addr` argument of `sys::mman::mmap` is now of type `Option<NonZeroUsize>`.
+  ([#1870](https://github.com/nix-rust/nix/pull/1870))
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


asomers (commented):
This increases the size of the argument.  How about using `Option<std::ptr::NonNull>` instead?  Or `Option<std::num::NonZeroUsize>`?

JonathanWoollett-Light:
Good point, updated to use `Option<std::num::NonZeroUsize>`.

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - The MSRV is now 1.56.1
   ([#1792](https://github.com/nix-rust/nix/pull/1792))
+- Passes `addr` to `sys::mmap` as `Option<NonZeroUsize>` from `*mut c_void`.
```
asomers, line 38: I find this wording confusing.  How about this?
```suggestion
- The `addr` argument to `sys::mman::mmap` is now of type `Option<NonZeroUsize>`.
```

JonathanWoollett-Light: Updated.

asomers (approved):
bors r+

PR #1871, opened 2022-11-19
Fix using SockaddrStorage to store Unix domain addresses on Linux

### LINKED ISSUE #1866: `SockaddrStorage` broken for variable-length addresses
The `SockaddrStorage` type "forgets" the address length, which causes incorrect and unexpected behaviour. I think this is a more fundamental issue with how nix represents socket addresses.

Example:

```rust
use std::os::unix::net::UnixListener;

use nix::sys::socket::{connect, getsockname, socket};
use nix::sys::socket::{AddressFamily, SockFlag, SockType, SockaddrStorage};

let listener = UnixListener::bind("/tmp/example.sock").unwrap();
let listener_fd = listener.as_raw_fd();
let listener_addr: SockaddrStorage = getsockname(listener_fd).unwrap();

let peer_fd = socket(AddressFamily::Unix, SockType::Stream, SockFlag::empty(), None).unwrap();

// this panics
connect(peer_fd, &listener_addr).unwrap();
```

If we look at the strace, we can see that nix is proving the wrong sockaddr length to `connect()`, which represents an incorrect address. This is equally important for abstract named and unnamed unix sockets.

```strace
socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC, 0) = 3
bind(3, {sa_family=AF_UNIX, sun_path="/tmp/example.sock"}, 20) = 0
listen(3, 4294967295)                   = 0
getsockname(3, {sa_family=AF_UNIX, sun_path="/tmp/example.sock"}, [128->20]) = 0
write(1, "SockaddrStorage { ss: sockaddr_s"..., 91SockaddrStorage { ss: sockaddr_storage { ss_family: 1, __ss_align: 8299682679707492728 } }
) = 91
socket(AF_UNIX, SOCK_STREAM, 0)         = 4
connect(4, {sa_family=AF_UNIX, sun_path="/tmp/example.sock"}, 128) = -1 EINVAL (Invalid argument)
```

When you're writing code that has to support many different socket address types, this means that `SockaddrStorage` is unusable. We're currently sticking with the deprecated `nix::sys::socket::SockAddr` for now as it doesn't have this issue.

asomers:
Since it has variable length, the user of a sockaddr_un must keep track of its true length.  On the BSDs, this is handled by the builtin sun_len field.  But on Linux-like operating systems it isn't.  Fix this bug by explicitly tracking it for SockaddrStorage just like we already do for UnixAddr.

Fixes #1866


asomers:
@stevenengler  could you test your code with this PR please?

rtzoeller (approved):

src/sys/socket/addr.rs
```diff
@@ -1537,6 +1537,17 @@ impl SockaddrLike for SockaddrStorage {
...
+                    target_os = "android",
+                    target_os = "fuchsia",
+                    target_os = "illumos",
+                    target_os = "linux"
+                ))]
+                if i32::from(ss.ss_family) == libc::AF_UNIX {
+                    // Safe because we UnixAddr is strictly smaller than
+                    // SockaddrStorage, and we just initialized the structure.
+                    (*(&mut ss as *mut libc::sockaddr_storage as *mut UnixAddr)).sun_len = len as u8;
+                }
```
stevenengler, line 1550: A `libc::sockaddr_storage` was initialized above, but I don't think that means that those bytes represent a valid `UnixAddr`? For example `libc::sockaddr_storage` has [6 padding bytes](https://github.com/rust-lang/libc/issues/3004) on Linux x86-64 that may not be zeroed, and reading those uninitialized bytes as a `UnixAddr` would be UB. I think this would only occur if `len` is less than 8 (`sizeof(u16) + 6`).

stevenengler:
> @stevenengler could you test your code with this PR please?

Yep, my example code works with this PR. Thanks for looking into this!

src/sys/socket/addr.rs
asomers: The docs for `mem::zeroed` say that padding bytes may not be initialized.  But if you look at the source, they are, at least in the current implementation.  So I don't think there's a problem to merge this PR.  But we should still fix the issue you raised with libc.

asomers:
bors r=rtzoeller

src/sys/socket/addr.rs
```diff
@@ -1537,6 +1537,17 @@ impl SockaddrLike for SockaddrStorage {
...
+                #[cfg(any(
+                    target_os = "android",
+                    target_os = "fuchsia",
+                    target_os = "illumos",
+                    target_os = "linux"
+                ))]
+                if i32::from(ss.ss_family) == libc::AF_UNIX {
+                    // Safe because we UnixAddr is strictly smaller than
+                    // SockaddrStorage, and we just initialized the structure.
+                    (*(&mut ss as *mut libc::sockaddr_storage as *mut UnixAddr)).sun_len = len as u8;
```
rtzoeller, line 1549: @asomers do you think nix should release patch versions of 0.24 and 0.25 with this change?

PR #1873, opened 2022-11-19
mmap non-zero length

JonathanWoollett-Light:
When calling [`mmap`](https://man7.org/linux/man-pages/man2/mmap.2.html) the passed length needs to be greater than zero, else an  error is returned:

> EINVAL (since Linux 2.6.12) length was 0.

By specifying the argument as `std::num::NonZeroUsize` we eliminate this error case.

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -418,7 +418,7 @@ pub fn munlockall() -> Result<()> {
 /// [`mmap(2)`]: https://man7.org/linux/man-pages/man2/mmap.2.html
 pub unsafe fn mmap(
     addr: Option<NonZeroUsize>,
-    length: size_t,
+    length: NonZeroUsize,
     prot: ProtFlags,
     flags: MapFlags,
     fd: RawFd,
@@ -428,8 +428,8 @@ pub unsafe fn mmap(
         std::ptr::null_mut(),
         |a| usize::from(a) as *mut c_void
     );
-
-    let ret = libc::mmap(ptr, length, prot.bits(), flags.bits(), fd, offset);
+    
+    let ret = libc::mmap(ptr, length.into(), prot.bits(), flags.bits(), fd, offset);
 
     if ret == libc::MAP_FAILED {
         Err(Errno::last())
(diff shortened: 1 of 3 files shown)


test/sys/test_mman.rs
```diff
@@ -1,11 +1,12 @@
 use nix::sys::mman::{mmap, MapFlags, ProtFlags};
+use std::num::NonZeroUsize;
 
 #[test]
 fn test_mmap_anonymous() {
     unsafe {
         let ptr = mmap(
             std::ptr::null_mut(),
-            1,
+            NonZeroUsize::new(1).unwrap(),
```
asomers, line 9: Do you think it would be worthwhile to bring in the nonzero_ext crate just for these tests?
```suggestion
            nonzero_ext::nonzero!(1)
```

asomers (changes requested):
I think this is a good improvement.  But don't forget a CHANGELOG.  Also, what about
`mremap`, and `munmap`?  The man pages have the same restriction for them.  And for `msync`, a zero-value for the `length` argument has special meaning, so we should probably turn it into `Option<usize>`.

src/sys/mman.rs
```diff
@@ -515,8 +515,9 @@ pub unsafe fn madvise(
 /// # use nix::sys::mman::{mmap, mprotect, MapFlags, ProtFlags};
 /// # use std::ptr;
 /// const ONE_K: size_t = 1024;
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Updated to fix this.

test/sys/test_mman.rs
JonathanWoollett-Light: At this point no, but if `NonZeroUsize` sees wider usage it may become worthwhile.

JonathanWoollett-Light:
> I think this is a good improvement. But don't forget a CHANGELOG. Also, what about `mremap`, and `munmap`? The man pages have the same restriction for them. And for `msync`, a zero-value for the `length` argument has special meaning, so we should probably turn it into `Option<usize>`.

These changes make sense to me. I think they make sense as separate PRs to keep the changes incremental.

test/sys/test_mman.rs
```diff
@@ -79,12 +82,17 @@ fn test_mremap_grow() {
 fn test_mremap_shrink() {
     use nix::libc::{c_void, size_t};
     use nix::sys::mman::{mremap, MRemapFlags};
+    use std::num::NonZeroUsize;
 
     const ONE_K: size_t = 1024;
+    let one_k_non_zero = NonZeroUsize::new(ONE_K).unwrap();
+    let ten_one_k = one_k_non_zero
+        .checked_mul(NonZeroUsize::new(10).unwrap())
+        .unwrap();
```
asomers, line 91: I'd hate to raise the MSRV just for this one line in a test.  How about this instead?
```suggestion
    let ten_one_k = NonZeroUsize::new(10 * 1024).unwrap();
```

JonathanWoollett-Light: Fixed

test/sys/test_mman.rs
asomers: BTW, I took a look through my other crates and found exactly zero places where nonzero_ext would help.  It turns out that `Option<NonZeroXXX>` is a lot more common than `NonZeroXXX`.

asomers (approved):
bors r+

asomers:
bors retry

PR #1874, opened 2022-11-20
signalfd optional file descriptor

JonathanWoollett-Light:
[`sys::signalfd::signalfd`](https://docs.rs/nix/latest/nix/sys/signalfd/fn.signalfd.html) currently takes a `RawFd` for its `fd` argument.

Considering from [the documentation](https://man7.org/linux/man-pages/man2/signalfd.2.html):
> If the fd argument is -1, then the call creates a new file
       descriptor and associates the signal set specified in mask with
       that file descriptor.  If fd is not -1, then it must specify a
       valid existing signalfd file descriptor, and mask is used to
       replace the signal set associated with that file descriptor.

We can better pass the argument as `Option<BorrowedFd>` which encodes the optional nature of this parameter in an option rather than the value being -1 (invalid).

This removes the error case where `fd < -1`.

> EBADF  The fd file descriptor is not a valid file descriptor.

Does this does however require additional changes to produce a cohesive implementation.


SUPERCILEX:
This should use `impl AsFd`.

src/sys/signalfd.rs
```diff
@@ -46,13 +46,14 @@ pub const SIGNALFD_SIGINFO_SIZE: usize = mem::size_of::<siginfo>();
 /// signalfd (the default handler will be invoked instead).
 ///
 /// See [the signalfd man page for more information](https://man7.org/linux/man-pages/man2/signalfd.2.html)
-pub fn signalfd(fd: RawFd, mask: &SigSet, flags: SfdFlags) -> Result<RawFd> {
+pub fn signalfd(fd: Option<BorrowedFd>, mask: &SigSet, flags: SfdFlags) -> Result<OwnedFd> {
```
notgull, line 49: Is there a way we could use the `io-lifetimes` crate here to avoid bumping the MSRV?

SUPERCILEX: Yes since that's what rustix does, but not sure that's worth it without someone piping up to give a compelling reason they need to be able to use old rustcs.

asomers: We're already planning to raise MSRV for the 0.27.0 release.

JonathanWoollett-Light:
> This should use `impl AsFd` instead of a `BorrowedFd`.

In my view this defats the purpose of this PR, which is to restrict the type you can pass to `signalfd` to a valid file descriptor such that the error case

> EBADF The fd file descriptor is not a valid file descriptor.

Cannot be encountered.

SUPERCILEX:
> With impl AsFd a user can still pass an invalid file descriptor.

Do you have a specific example in mind? Also, nothing prevents you from using unsafe to create a BorrowedFd from random garbage.

JonathanWoollett-Light:
Apologies I misread, thought you wrote `AsRawFd`. I will change it to `impl AsFd`.

SUPERCILEX:
Dope :+1:

JonathanWoollett-Light:
It does make specifying `None`a bit more of a pain:
```rust
signalfd(None::<OwnedFd>, mask, flags)?;
```
instead of
```
signalfd(None, mask, flags)?;
```

SteveLauC:
For the type of that file descriptor argument, any reason why not to use `<F: AsFd>(fd: F)`, this is basically equivalent to `impl AsFd` except you can specify the particular type using turbofish

JonathanWoollett-Light:
@SteveLauC I have no strong view either way, I've changed it to match stdlib as you mentioned.

asomers:
You should be able to rebase this now

JonathanWoollett-Light:
> You should be able to rebase this now

Rebased

src/sys/signalfd.rs
```diff
@@ -82,30 +85,30 @@ pub fn signalfd(fd: RawFd, mask: &SigSet, flags: SfdFlags) -> Result<RawFd> {
 ///     Err(err) => (), // some error happend
 /// }
 /// ```
-#[derive(Debug, Eq, Hash, PartialEq)]
-pub struct SignalFd(RawFd);
+#[derive(Debug)]
+pub struct SignalFd(ManuallyDrop<OwnedFd>);
```
asomers, line 89: Why use `ManuallyDrop`?  An `OwnedFd` will close itself on drop.  Relying on that would simplify things.

JonathanWoollett-Light: Previously the closing of the file descriptor was handled such that it would panic if `unistd::close` failed:
```rust
impl Drop for SignalFd {
    fn drop(&mut self) {
        let e = unistd::close(self.0.as_raw_fd());
        if !std::thread::panicking() && e == Err(Errno::EBADF) {
            panic!("Closing an invalid file descriptor!");
        };
    }
}
```
Without `ManuallyDrop` we rely upon how the [standard library `Drop` implementation](https://doc.rust-lang.org/stable/src/std/os/fd/owned.rs.html#167-179) which simply ignore the return value of `close` meaning it would no longer panic if it failed.
```rust
impl Drop for OwnedFd {
    #[inline]
    fn drop(&mut self) {
        unsafe {
            // Note that errors are ignored when closing a file descriptor. The
            // reason for this is that if an error occurs we don't actually know if
            // the file descriptor was closed or not, and if we retried (for
            // something like EINTR), we might close another valid file descriptor
            // opened after we closed ours.
            let _ = libc::close(self.fd);
        }
    }
}
```

SUPERCILEX: By definition, it is UB for an OwnedFd to use an invalid file descriptor, so that code can be removed.

src/sys/signalfd.rs
```diff
@@ -46,13 +44,14 @@ pub const SIGNALFD_SIGINFO_SIZE: usize = mem::size_of::<siginfo>();
 /// signalfd (the default handler will be invoked instead).
 ///
 /// See [the signalfd man page for more information](https://man7.org/linux/man-pages/man2/signalfd.2.html)
-pub fn signalfd(fd: RawFd, mask: &SigSet, flags: SfdFlags) -> Result<RawFd> {
+pub fn signalfd<F: AsFd>(fd: Option<F>, mask: &SigSet, flags: SfdFlags) -> Result<OwnedFd> {
```
SteveLauC, line 47: `pub fn signalfd<F: AsFd>(fd: Option<&F>` probably is the better way to go, the interface exposed from the `stdlib` [is not that great](https://github.com/rust-lang/rust/issues/88989#issuecomment-1035416012), sorry for my previous suggestion on changing the type to `<F: AsFd>(fd: F)`

JonathanWoollett-Light: Given https://github.com/nix-rust/nix/pull/1928#discussion_r1043304050 I think we can resolve this?

src/unistd.rs
```diff
@@ -1078,8 +1078,8 @@ pub fn gethostname() -> Result<OsString> {
 /// let f = tempfile::tempfile().unwrap();
 /// close(f.into_raw_fd()).unwrap(); // Good.  into_raw_fd consumes f
 /// ```
-pub fn close(fd: RawFd) -> Result<()> {
-    let res = unsafe { libc::close(fd) };
+pub fn close<Fd: IntoRawFd>(fd: Fd) -> Result<()> {
```
asomers, line 1081: Is this change really supposed to be part of this PR, or a different one?

JonathanWoollett-Light: You are right, this was not supposed to be in this PR. Removed it.

src/sys/signalfd.rs
asomers: Is it technically UB?  I think the compiler will be perfectly happy with it.  The memory is initialized, after all.

asomers (approved):
bors r+

PR #1862, opened 2022-11-13
Bump MSRV to 1.63 for I/O safety

SUPERCILEX:
Prep for https://github.com/nix-rust/nix/issues/1750

PR #1876, opened 2022-11-21
Safer poll timeout

JonathanWoollett-Light:
Implements a wrapper around `i32` for the `timeout` argument in [`poll::poll`](https://docs.rs/nix/latest/nix/poll/fn.poll.html).

This prevents the error case:

> EINVAL (ppoll()) The timeout value expressed in *ip is invalid
              (negative).

Most of the code additions are simple conversions to/from integer primitives.

An alternative approach to implement the same safety with less code would require something like https://internals.rust-lang.org/t/non-negative-integer-types/17796 and would offer less usability like converting to/from `std::time::Duration`s.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,9 +1,9 @@
 //! Wait for events to trigger on specific file descriptors
 use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
+use std::time::Duration;
 
 use crate::errno::Errno;
 use crate::Result;
-
 /// This is a wrapper around `libc::pollfd`.
 ///
 /// It's meant to be used as an argument to the [`poll`](fn.poll.html) and
@@ -27,13 +27,13 @@ impl<'fd> PollFd<'fd> {
     /// ```no_run
     /// # use std::os::unix::io::{AsFd, AsRawFd, FromRawFd};
     /// # use nix::{
-    /// #     poll::{PollFd, PollFlags, poll},
+    /// #     poll::{PollTimeout, PollFd, PollFlags, poll},
     /// #     unistd::{pipe, read}
     /// # };
     /// let (r, w) = pipe().unwrap();
     /// let pfd = PollFd::new(r.as_fd(), PollFlags::POLLIN);
     /// let mut fds = [pfd];
-    /// poll(&mut fds, -1).unwrap();
+    /// poll(&mut fds, PollTimeout::NONE).unwrap();
     /// let mut buf = [0u8; 80];
     /// read(r.as_raw_fd(), &mut buf[..]);
     /// ```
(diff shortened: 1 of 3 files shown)


asomers:
CI is failing because you're trying to use rust 1.58.0, but the 1.58 container installs 1.58.1.  Change the VERSION to 1.58.1

JonathanWoollett-Light:
> CI is failing because you're trying to use rust 1.58.0, but the 1.58 container installs 1.58.1. Change the VERSION to 1.58.1

Thank you for letting me know.

Why is this the case?

asomers:
> > CI is failing because you're trying to use rust 1.58.0, but the 1.58 container installs 1.58.1. Change the VERSION to 1.58.1
> 
> Thank you for letting me know.
> 
> Why is this the case?

That's just the convention for the rust-lang containers.  They always include the newest patch release.

src/poll.rs
```diff
@@ -112,6 +115,255 @@ libc_bitflags! {
...
+        }
+    }
+}
+impl std::error::Error for TryFromI64Error {}
+
+// These cases implement slightly different conversions that make using generics impossible without
+// specialization.
+impl PollTimeout {
+    /// Blocks indefinitely.
+    pub const NONE: Self = Self(1 << 31);
```
asomers, line 162: Is this supposed to be C's `INFTIM`, and if so shouldn't it be `-1`?

src/poll.rs
```diff
@@ -132,12 +384,12 @@ libc_bitflags! {
 /// in timeout means an infinite timeout.  Specifying a timeout of zero
```
asomers, line 384: You should update the docs for the `timeout` variable.

CHANGELOG.md
```diff
@@ -41,12 +41,14 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
-- The MSRV is now 1.56.1
-  ([#1792](https://github.com/nix-rust/nix/pull/1792))
+- The MSRV is now 1.58.1
+  ([#1876](https://github.com/nix-rust/nix/pull/1876))
```
asomers, line 45: You can delete this part, and all of the other MSRV stuff, since we'll merge #1862 first.

asomers (changes requested):

src/poll.rs
JonathanWoollett-Light: I think I was originally using a unsigned integer type here for some reason e.g. `1 << 31 == -1`. But it makes more sense to use `-1` now, will change this.

CHANGELOG.md
JonathanWoollett-Light: Dropped this commit

JonathanWoollett-Light:
@asomers It seems like the one failing task needs to be re-run.

src/poll.rs
```diff
@@ -132,6 +135,255 @@ libc_bitflags! {
...
+            // EINVAL (ppoll()) The timeout value expressed in *ip is invalid (negative).
+            _ => Err(Errno::EINVAL),
+        }
+    }
+}
+impl TryFrom<PollTimeout> for Duration {
+    type Error = ();
+    fn try_from(x: PollTimeout) -> std::result::Result<Self, ()> {
+        match x.timeout() {
+            // SAFETY: `x.0` is always positive.
```
asomers, line 297:
```suggestion
            // SAFETY: `x.0` is always nonnegative.
```

src/poll.rs
```diff
@@ -132,6 +135,255 @@ libc_bitflags! {
...
+impl PollTimeout {
+    /// Blocks indefinitely.
+    pub const NONE: Self = Self(-1);
+    /// Returns immediately.
+    pub const ZERO: Self = Self(0);
+    /// Blocks for at most [`std::i32::MAX`] milliseconds.
+    pub const MAX: Self = Self(i32::MAX);
+    /// Returns if `self` equals [`PollTimeout::NONE`].
+    pub fn is_none(&self) -> bool {
+        *self == Self::NONE
```
asomers, line 189: Just to program defensively, I recommend:
```suggestion
        *self <= Self::NONE
```

src/poll.rs
```diff
@@ -148,16 +400,16 @@ libc_bitflags! {
...
-/// interval may overrun by a small amount.  Specifying a negative value
-/// in timeout means an infinite timeout.  Specifying a timeout of zero
-/// causes `poll()` to return immediately, even if no file descriptors are
-/// ready.
-pub fn poll(fds: &mut [PollFd], timeout: libc::c_int) -> Result<libc::c_int> {
+/// interval may overrun by a small amount.  Specifying a [`PollTimeout::NONE`]
+/// in timeout means an infinite timeout.  Specifying a timeout of
+/// [`PollTimeout::ZERO`] causes `poll()` to return immediately, even if no file
+/// descriptors are ready.
+pub fn poll(fds: &mut [PollFd], timeout: PollTimeout) -> Result<libc::c_int> {
```
asomers, line 407: I think it would be more usable, and more backwards compatible, to write the signature like this:
```suggestion
pub fn poll<T: Into<PollTimeout>>(fds: &mut [PollFd], timeout: T) -> Result<libc::c_int> {
```

src/poll.rs
JonathanWoollett-Light: You've right, fixed it.

src/poll.rs
JonathanWoollett-Light: Fixed.

src/poll.rs
```diff
@@ -132,6 +135,255 @@ libc_bitflags! {
...
+            Self::Underflow(err) => write!(f, "Underflow: {err}"),
+            Self::Overflow(err) => write!(f, "Overflow: {err}"),
+        }
+    }
+}
+impl std::error::Error for TryFromI128Error {}
+
+/// Error type for [`PollTimeout::try_from::<i68>()`].
+#[derive(Debug, Clone, Copy)]
+pub enum TryFromI64Error {
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Made some changes to the conversions to remove this error type.

src/poll.rs
JonathanWoollett-Light: Makes sense, made the change.

src/poll.rs
```diff
@@ -179,6 +179,180 @@ libc_bitflags! {
...
+impl From<i16> for PollTimeout {
+    fn from(x: i16) -> Self {
+        Self(i32::from(x))
+    }
+}
+impl From<i8> for PollTimeout {
+    fn from(x: i8) -> Self {
+        Self(i32::from(x))
+    }
+}
```
SteveLauC, line 285: The conversions from signed integers seem to be insistent, for `i128/i64`, a negative number would be converted to `PollTimeOut(-1)`, and for `i32/i16/i8`, they are stored as is.

But this won't affect how `poll()` behaves as it will be an infinite timeout as long as it is negative

JonathanWoollett-Light: Yes, these will produce different underlying values. I did this as the conversions from `i32/i16/i8` are able to be slightly more efficient while maintaining the same overall behavior. I can change them to be more like `i128/i64` if you want, I am not strongly opinionated here.

SteveLauC: I agree the current impl is more performant.

My worry is that users can convert a signed integer to a `PollTimeOut`, and they can also convert it back, then such an inconsistence would be discovered, which may confuses them a little bit:

```rs
let orig_value1: i128 = -2;
let orig_value2: i132 = -2;

let timeout1 = PollTimeOut::from(orig_value1);
let timeout2 = PollTimeOut::from(orig_value2);

let value1: i128 = timeout1.into();
let value2: i32 = timeout2.into();

assert_eq!(orig_value1, value1); // assert_eq!(-2, -1);
assert_eq!(orig_value2, value2); // assert_eq!(-2, -2);
```

SteveLauC:
BTW, CI is failing due the an unused import:

```
// test_poll.rs

error: unused import: `close`
 --> test/test_poll.rs:4:14
  |
4 |     unistd::{close, pipe, write},
  |              ^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`

```

JonathanWoollett-Light:
> BTW, CI is failing due the an unused import:
> 
> ```
> // test_poll.rs
> 
> error: unused import: `close`
>  --> test/test_poll.rs:4:14
>   |
> 4 |     unistd::{close, pipe, write},
>   |              ^^^^^
>   |
>   = note: `-D unused-imports` implied by `-D warnings`
> 
> error: could not compile `nix` due to previous error
> ```

Fixed

src/poll.rs
JonathanWoollett-Light: Would documentation on `PollTimeout` be sufficient to alleviate this concern?

SteveLauC: Yeap, a documentation on this behavior would be nice:)

src/poll.rs
```diff
@@ -179,6 +179,180 @@ libc_bitflags! {
...
+    pub fn is_none(&self) -> bool {
+        // > Specifying a negative value in timeout means an infinite timeout.
+        *self <= Self::NONE
+    }
+    /// Returns if `self` does not equal [`PollTimeout::NONE`].
+    pub fn is_some(&self) -> bool {
+        !self.is_none()
+    }
+    /// Returns the timeout in milliseconds if there is some, otherwise returns `None`.
+    pub fn timeout(&self) -> Option<i32> {
```
asomers, line 208: I think naming this method `as_millis` would be more idiomatic.  For example, it would mirror `Duration::as_millis`.  And since it will never return a negative value, should it return `u32` instead of `i32`?

src/poll.rs
asomers: Actually, accepting negative values for `timeout` seems to be a Linuxism.  POSIX only requires that `-1` be interpreted as "forever", and FreeBSD will return `EINVAL` for other negative values.  Please make it impossible to contruct a `PollTimeout` with a negative value other than `-1`.

asomers (changes requested):

src/poll.rs
JonathanWoollett-Light: I've split this into 2 functions
```rust
    /// Returns the timeout in milliseconds if there is some, otherwise returns `None`.
    pub fn as_millis(&self) -> Option<i32> {
        self.is_some().then_some(self.0)
    }
    /// Returns the timeout as a `Duration` if there is some, otherwise returns `None`.
    pub fn timeout(&self) -> Option<Duration> {
        self.as_millis().map(|x|Duration::from_millis(u64::try_from(x).unwrap()))
    }
```

src/poll.rs
JonathanWoollett-Light: It is quite a bit messier to encapsulate this logic as it requires custom `Error` types for many of the `TryFrom` implementations to cover this.

src/poll.rs
JonathanWoollett-Light: I think using `i32` or `u32` each have their own drawbacks here, ideally we could use `u31` with something like https://crates.io/crates/ux but this crate is not actively maintained. I tried my hand at making a better version as an experiment with https://crates.io/crates/ux2 but I am busy so am not able to maintain this alone.

src/poll.rs
SteveLauC: > I do not think the POSIX specification specifies that negative values other than -1 should error. 

I think @asomers was not saying that specifying a negative value other than -1 should be considered an error in the POSIX standard, it is the FreeBSD implementation that treats that case as an error

The behavior of FreeBSD is documented in their [manual](https://man.freebsd.org/cgi/man.cgi?poll#end), and from this issue [cpython: select.poll.poll fails on BSDs with arbitrary negative timeouts](https://github.com/python/cpython/issues/75515), this seems to be a general behavior across BSD OSes

SteveLauC:
BTW, we have changed our changelog mode, please see [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add changelog entries

src/poll.rs
JonathanWoollett-Light: Okay, I have re-worked the conversions to return an error when a non -1 negative is given.

SteveLauC (changes requested):

src/poll.rs
SteveLauC: I agree that `u31` suits best here, but before switching to the `ux` or `ux2` crate, `u32` is the better choice compared to `i32`

SteveLauC:
Left some comments, I think this PR is ready for merge after addressing these comments

src/poll.rs
```diff
@@ -175,6 +175,229 @@ libc_bitflags! {
...
+    /// > Specifying  a timeout of zero causes poll() to return immediately, even if no file
+    /// > descriptors are ready.
+    pub const ZERO: Self = Self(0);
+    /// Blocks for at most [`std::i32::MAX`] milliseconds.
+    pub const MAX: Self = Self(i32::MAX);
+    /// Returns if `self` equals [`PollTimeout::NONE`].
+    pub fn is_none(&self) -> bool {
+        // > Specifying a negative value in timeout means an infinite timeout.
+        *self <= Self::NONE
+    }
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I think this should remain `<=` as this is the safer option. The speedup gained from `==` over `<=` will be a couple nanoseconds at most. I can see a possible bug in the future arising otherwise.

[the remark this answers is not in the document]
SteveLauC: > I can see a possible bug in the future arising otherwise.

Would like to elaborate on it a bit? Maybe give an example?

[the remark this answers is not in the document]
JonathanWoollett-Light: If in the future for some reason on some specific platform other non-negative values are supported because it is useful or necessary, this would then cause this function to be incorrect if not also changed. By sticking close to the documentation it provides the safest approach considering cross-platform and future proofing.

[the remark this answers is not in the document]
SteveLauC: Ah, I see, makes sense to me, would you like to add a comment here stating why we are doing it in this way in case someone will accidentally change it in the future

src/poll.rs
```diff
@@ -175,6 +175,229 @@ libc_bitflags! {
...
+pub struct PollTimeout(i32);
+
+impl PollTimeout {
+    /// Blocks indefinitely.
+    ///
+    /// > Specifying a negative value in timeout means an infinite timeout.
+    pub const NONE: Self = Self(-1);
+    /// Returns immediately.
+    ///
+    /// > Specifying  a timeout of zero causes poll() to return immediately, even if no file
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Fixed.

src/poll.rs
SteveLauC: What about this one?

JonathanWoollett-Light: I am okay to use `u32`.

JonathanWoollett-Light: Updated to u32.

SteveLauC (approved):

SteveLauC:
The CI failure is not related to this PR

asomers:
I can't reproduce the failure.  Could github somehow be using a cached Cargo.lock file?

SteveLauC:
> I can't reproduce the failure. Could github somehow be using a cached Cargo.lock file?

Try [this](https://github.com/nix-rust/nix/pull/2187#issuecomment-1820253849)? I can reproduce it even with an empty hello-world crate

asomers:
Ahh, so it's a toolchain bug.  I thought it was a libc bug.  I'll open a bug upstream.  In the meantime, we'll either have to disable this target, or pin our rustc version.

SteveLauC:
> Ahh, so it's a toolchain bug. I thought it was a libc bug. I'll open a bug upstream. In the meantime, we'll either have to disable this target, or pin our rustc version.

I will pin the nightly toolchain to the version just before that

PR #1878, opened 2022-11-21
EPOLL_CLOEXEC bool

JonathanWoollett-Light:
Replaces `EpollCreateFlags` with a simple bool argument. Bitflags with 1 option should just be a boolean.


notgull (commented):
In the future, the `epoll` API might add more flags to this bitflag set. If so, it would be a breaking change to replace this `bool` with the proper bitflags.

asomers:
@notgull is right.  It's best to leave epoll how it is.

PR #1879, opened 2022-11-21
feat: Add associated constants `UTIME_OMIT` `UTIME_NOW` to `TimeSpec`

### LINKED ISSUE #1834: Improve the doc for `utimensat(3)` and `futimens(3)`
One advantage of `utimensat(3)` and `futimens(3)` over `utimes(2)` is that they can just set **one** timestamp at a time.

Cite the [man page](https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html):
> If the tv_nsec field of a timespec structure has the special value UTIME_NOW, the file's relevant timestamp shall be set to the greatest value supported by the file system that is not greater than the current time. If the tv_nsec field has the special value UTIME_OMIT, the file's relevant timestamp shall not be changed. In either case, the tv_sec field shall be ignored.

Say I wanna update `atime` and leave `mtime` unchanged, if using `utimes(2)`, I have to retrieve the previous `mtime` using `stat(2)` or something similar, but using `utimensat(3)` or `futimens(3)`, I can just easily set the `tv_nsec` field of that timestamp to `UTIME_OMIT`.

So I think this feature deserves to be documented.

SteveLauC:
Since we can add some breaking changes in the following 0.27 release,  I made this PR.

#### What this PR does
1. Change the type of  `atime`, `mtime` arguments of `utimenstat(2)` and `futimens(2)`  to `Option<&TimeSpec>`, when set to `None`, the corresponding timestamp remains unchanged.
2. Add two tests for feature 1
3. Re-export `libc::UTIME_NOW` in `nix::sys::stat` module so that users can directly use it when they want to update timestamps to the special value `Now`
4. Change the URL of `gethostname(2)` in `README.md` to the UNIX one (from Open Group)

Fixes #1834 

cc @sargun


SteveLauC:
`redox` does not support `UTIME_NOW` and `UIME_OMIT`, I will file an issue in their GitLab repo to ask if they have any plan to support this feature.

src/sys/stat.rs
```diff
@@ -402,14 +403,51 @@ pub fn lutimes<P: ?Sized + NixPath>(
...
+) -> [libc::timespec; 2] {
+    let mut times = [
+        libc::timespec {
+            tv_sec: 0,
+            tv_nsec: libc::UTIME_OMIT,
+        },
+        libc::timespec {
+            tv_sec: 0,
+            tv_nsec: libc::UTIME_OMIT,
+        },
```
asomers, line 421:
```suggestion
        libc::timespec {
            tv_sec: 0,
            tv_nsec: libc::UTIME_OMIT,
            ..
        },
        libc::timespec {
            tv_sec: 0,
            tv_nsec: libc::UTIME_OMIT,
            ..
        },
```
See #1886 for the reason why.

asomers (commented):
Do you think it would be more natural to change the arguments into an enum like this:
```rust
pub enum UtimeSpec {
    Omit,
    Now,
    Set(TimeSpec)
}
```
Or would that just be unnecessarily complicated?

src/sys/stat.rs
```diff
@@ -402,12 +402,87 @@ pub fn lutimes<P: ?Sized + NixPath>(
...
+            target_arch = "x86_64",
+            target_pointer_width = "32"
+        ))]
+        TimestampSpec::Now => default_value.tv_nsec = libc::UTIME_NOW as i64,
+        #[cfg(not(all(
+            target_os = "linux",
+            target_arch = "x86_64",
+            target_pointer_width = "32"
+        )))]
+        TimestampSpec::Now => default_value.tv_nsec = libc::UTIME_NOW,
```
asomers, line 450: Let's deuglify this:
```suggestion
        TimestampSpec::Omit => default_value.tv_nsec = libc::UTIME_OMIT as _,
        TimestampSpec::Now => default_value.tv_nsec = libc::UTIME_NOW as _,
```

src/sys/stat.rs
```diff
@@ -402,12 +402,87 @@ pub fn lutimes<P: ?Sized + NixPath>(
...
+    /// Update the timestamp to `Now`
+    Now,
+    /// Update the timestamp to the specific value specified in `TimeSpec`.
+    Set(TimeSpec),
+}
+
+/// A helper function to construct an instance of `libc::timespec` according to
+/// the operation specified in `spec`.
+#[cfg(not(target_os = "redox"))] // Redox does not support `UTIME_OMIT` and `UTIME_NOW`
+fn create_timespec(spec: &TimestampSpec) -> libc::timespec {
```
asomers, line 420: I know it's only an internal function, so this is no big deal.  But it would be a little more ergonomic to use if this were written as
```rust
impl From<TimestampSpec> for libc::timespec {...}
```

src/sys/stat.rs
```diff
@@ -402,12 +402,87 @@ pub fn lutimes<P: ?Sized + NixPath>(
...
+/// descriptor.
+///
+/// # References
+///
+/// [futimens(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html).
+#[cfg(not(target_os = "redox"))] // Redox does not support `UTIME_OMIT` and `UTIME_NOW`
+#[inline]
+pub fn futimens(
+    fd: RawFd,
+    atime: TimestampSpec,
```
asomers, line 470: This structure is large enough that it would be more efficient to pass by reference.

asomers (changes requested):

src/sys/stat.rs
SteveLauC: Thanks for showing me this!

src/sys/time.rs
```diff
@@ -6,7 +6,7 @@ use std::convert::From;
 use std::time::Duration;
 use std::{cmp, fmt, ops};
 
-const fn zero_init_timespec() -> timespec {
+pub(crate) const fn zero_init_timespec() -> timespec {
```
SteveLauC, line 9: Make this function `pub(crate)` so that we can use it internally.

src/sys/stat.rs
```diff
@@ -402,12 +402,69 @@ pub fn lutimes<P: ?Sized + NixPath>(
...
+/// Construct a new [`libc::timespec`] instance according to the value of [`TimestampSpec`].
+///
+/// This is used in the implementations of [`futimens`] and [`utimensat`]. As an
+/// end user, you may NOT find this trait implementation useful.
+///
+/// [`timespec`]: libc::timespec
+#[cfg(not(target_os = "redox"))] // Redox does not support `UTIME_OMIT` and `UTIME_NOW`
+impl From<&TimestampSpec> for libc::timespec {
+    fn from(spec: &TimestampSpec) -> Self {
+        let mut default_value = super::time::zero_init_timespec();
```
SteveLauC, line 426: With the use of this `zero_init_timespec()` function, I guess we don't need to take care of that padding field issue on `musl`?

SteveLauC:
> redox does not support `UTIME_NOW` and `UIME_OMIT`, I will file an issue in their GitLab repo to ask if they have any plan to support this feature.

[Issue](https://gitlab.redox-os.org/redox-os/relibc/-/issues/172) filed.

------

This PR involves two syscalls:
1. `utimensat(2)`
2. `futimens(2)`

Redox does not support `utimensat(2)`. 

For `futimens(2)`, since redox does not have `UTIME_NOW` and `UTIME_OMIT`, I leave the previous implementation unchanged for it:

```rust
#[inline]
#[cfg(target_os = "redox")]
pub fn futimens(fd: RawFd, atime: &TimeSpec, mtime: &TimeSpec) -> Result<()> {
    let times: [libc::timespec; 2] = [*atime.as_ref(), *mtime.as_ref()];
    let res = unsafe { libc::futimens(fd, &times[0]) };
```

src/sys/stat.rs
asomers: Yep

asomers (commented):
This looks good.  But before we merge it, I just have to ask a question: does the `TimestampSpec` abstraction create too much overhead?  I think it's nice and ergonomic.  But a zero-cost alternative would simply be to leave the old interface as it was, and add constants like this:
```rust
pub const UTIME_OMIT: libc::timespec {tv_sec: 0, tv_nsec: libc::UTIME_OMIT}
```
that the user could pass in to `futimens`.

SteveLauC:
> I just have to ask a question: does the TimestampSpec abstraction create too much overhead?

Let's compare these two interfaces:

##### NEW interface
```rust
// To leave timestamps unchanged
futimens(&file, &TimestampSpec::Omit, &TimestampSpec::Omit).unwrap();
// To update timestamps to NOW
futimens(&file, &TimestampSpec::Now, &TimestampSpec::Now).unwrap();
// To set timestamps to specific values, e.g., UNIX epoch
let timespec = TimeSpec::zero();
futimens(
    &file,
    &TimestampSpec::Set(timespec),
    &TimestampSpec::Set(timespec),
)
.unwrap();
```

##### OLD interface, with `UTIME_OMIT` and `UTIME_NOW` defined:

```rust
// To leave timestamps unchanged
futimens(
    &file,
    &TimeSpec::from_timespec(UTIME_OMIT),
    &TimeSpec::from_timespec(UTIME_OMIT),
)
.unwrap();
// To update timestamps to NOW
futimens(
    &file,
    &TimeSpec::from_timespec(UTIME_NOW),
    &TimeSpec::from_timespec(UTIME_NOW),
)
.unwrap();
// To set timestamps to specific values, e.g., UNIX epoch
let timespec = TimeSpec::zero();
futimens(&file, &timespec, &timespec).unwrap();
```

--------

IMO, the new interface looks pretty concise and expressive. The old interface, if we define `UTIME_OMIT` and `UTIME_NOW` like this:

```rust
pub const UTIME_OMIT: TimeSpec = TimeSpec::new(0, libc::UTIME_OMIT);
pub const UTIME_NOW: TimeSpec = TimeSpec::new(0, libc::UTIME_NOW);
```
Then it could be more concise than the new one, but less expressive IMO:

```rust
futimens(&file, &UTIME_OMIT, &UTIME_OMIT).unwrap();
futimens(&file, &UTIME_NOW, &UTIME_NOW).unwrap();
```

SteveLauC:
Update on the that Redox (relibc) issue, [reply](https://gitlab.redox-os.org/redox-os/relibc/-/issues/172#note_24834) from developers of Redox:
> Since they're part of POSIX and don't seem to conflict with any core redox functionality, yes.

tertsdiepraam:
I just ran into this issue, so I'm glad to see it's being worked on :)

A third option for the API might be to define associated constants for `TimeSpec`, which I think would be zero-cost, fairly intuitive and a non-breaking change:
```rust
impl TimeSpec {
    pub const NOW: TimeSpec = TimeSpec::new(0, libc::UTIME_NOW);
    pub const OMIT: TimeSpec = TimeSpec::new(0, libc::UTIME_OMIT);
}
```
It would also be more self-documenting than reexporting `UTIME_NOW/OMIT`, because it would show up in the docs for `TimeSpec`.

It would look like this when used:
```rust
// To leave timestamps unchanged
futimens(&file, &TimeSpec::OMIT, &TimeSpec::OMIT).unwrap();
// To update timestamps to NOW
futimens(&file, &TimeSpec::NOW, &TimeSpec::NOW).unwrap();
// To set timestamps to specific values, e.g., UNIX epoch
let timespec = TimeSpec::zero();
futimens(
    &file,
    &timespec,
    &timespec,
)
.unwrap();
```

SteveLauC:
Revisiting this PR, I think defining 2 types for a function isn't quite suitable, we should define 2 associated constants for them


BTW, this is a duplicate PR of #1103, I am sorry that I didn't notice it while opening this PR

src/sys/stat.rs
```diff
@@ -376,6 +376,9 @@ pub fn lutimes<P: ?Sized + NixPath>(
 
 /// Change the access and modification times of the file specified by a file descriptor.
 ///
+/// If you want to set the timestamp to now, use `TimeSpec::UTIME_NOW`. Use
```
SteveLauC, line 379: Cannot use "[\`TimeSpec::UTIME_NOW\`]" because redox does not have it

asomers (approved):

PR #1880, opened 2022-11-21
Use the new UnixAddr::new_unnamed in the unit tests

asomers:
Use it in the from_sockaddr_un_abstract_unnamed test.  That test and this method were introduced by PRs #1871 and #1857, which crossed each other.


rtzoeller (approved):
bors r+

PR #1882, opened 2022-11-22
Epoll type

JonathanWoollett-Light:
Epoll can be most safely used as a type. This implement a type `Epoll` which supports this.


src/sys/epoll.rs
```diff
@@ -26,6 +26,64 @@ libc_bitflags!(
...
+/// let mut events = [EpollEvent::empty()];
+/// epoll.wait(&mut events, MILLIS as isize);
+/// 
+/// // Assert data correct & timeout didn't occur
+/// assert_eq!(events[0].data(), DATA);
+/// assert!(now.elapsed() < Duration::from_millis(MILLIS));
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug)]
```
asomers, line 60: Should this also derive Send and Sync ?  I think it should.

src/sys/epoll.rs
```diff
@@ -26,6 +26,64 @@ libc_bitflags!(
     }
 );
 
+/// A safe wrapper around [`epoll`](https://man7.org/linux/man-pages/man7/epoll.7.html).
+/// ```
+/// # use nix::sys::{epoll::{Epoll, EpollEvent, EpollFlags, EpollCreateFlags}, eventfd::{eventfd, EfdFlags}};
+/// # use nix::unistd::write;
+/// # use std::os::unix::io::{OwnedFd, FromRawFd, AsRawFd, AsFd};
+/// # use std::time::{Instant, Duration};
+/// # fn main() -> nix::Result<()> {
```
asomers, line 35: Do you need to declare the `main()` function explicitly?  If you leave it out, I think rustdoc will add it automatically.

src/sys/epoll.rs
```diff
@@ -26,6 +26,64 @@ libc_bitflags!(
...
+/// // Assert data correct & timeout didn't occur
+/// assert_eq!(events[0].data(), DATA);
+/// assert!(now.elapsed() < Duration::from_millis(MILLIS));
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug)]
+pub struct Epoll(pub OwnedFd);
+impl Epoll {
+    /// [`epoll_create1`].
```
asomers, line 63: This link isn't very useful.  How about this?
```suggestion
    /// Create a new `Epoll` object.
    /// [`epoll_create1`](https://linux.die.net/man/2/epoll_create1).
```
And similarly for the other methods.

asomers (changes requested):
This looks like a good improvement.  But I think you should deprecate the old interface.  We don't want to provide two APIs for doing the same thing.  Also, it needs a CHANGELOG entry.

src/sys/epoll.rs
JonathanWoollett-Light: Since [`OwnedFd`](https://doc.rust-lang.org/stable/std/os/unix/io/struct.OwnedFd.html) is `Send` and `Sync` these are automatically derived.

https://doc.rust-lang.org/nomicon/send-and-sync.html:

> Send and Sync are also automatically derived traits. This means that, unlike every other trait, if a type is composed entirely of Send or Sync types, then it is Send or Sync. Almost all primitives are Send and Sync, and as a consequence pretty much all types you'll ever interact with are Send and Sync.

src/sys/epoll.rs
JonathanWoollett-Light: Rustdoc automatically adds `main() {}` without a return type. To use `?` requires manually writing it with a return type.

src/sys/epoll.rs
JonathanWoollett-Light: Updated the comments for all the functions.

JonathanWoollett-Light:
> This looks like a good improvement. But I think you should deprecate the old interface. We don't want to provide two APIs for doing the same thing. Also, it needs a CHANGELOG entry.

Is there a policy on deprecation possibly using [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)? Or can I just remove the old implementations?

src/sys/epoll.rs
asomers: Doh, I forgot that part.

asomers:
> Is there a policy on deprecation possibly using [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)? Or can I just remove the old implementations?

Yes, you need to `#[deprecated]` them for at least one release before removing them.

JonathanWoollett-Light:
> > Is there a policy on deprecation possibly using [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)? Or can I just remove the old implementations?
> 
> Yes, you need to `#[deprecated]` them for at least one release before removing them.

Added `#[deprecated]` am I correct in adding `since = 0.26.0`? (since the function will be deprecated in `0.26.0`)

asomers:
> Added `#[deprecated]` am I correct in adding `since = 0.26.0`? (since the function will be deprecated in `0.26.0`)

No.  Due to the MSRV bump, release 0.26.0 won't include any of the I/O Safety stuff.  So you should change it to 0.27.0.

JonathanWoollett-Light:
@asomers Updated it to `0.27.0`

CHANGELOG.md
```diff
@@ -60,8 +60,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
-- The MSRV is now 1.56.1
-  ([#1792](https://github.com/nix-rust/nix/pull/1792))
+- The epoll interface now uses a type.
```
asomers, line 63: You need to rebase on master, and then move the CHANGELOG stuff to the top of the file.

src/sys/epoll.rs
```diff
@@ -70,20 +69,126 @@ impl EpollEvent {
...
+/// # fn main() -> nix::Result<()> {
+/// const DATA: u64 = 17;
+/// const MILLIS: u64 = 100;
+/// 
+/// // Create epoll
+/// let epoll = Epoll::new(EpollCreateFlags::empty())?;
+/// 
+/// // Create eventfd & Add event
+/// let eventfd = unsafe { OwnedFd::from_raw_fd(eventfd(0, EfdFlags::empty())?) };
+/// epoll.add(eventfd.as_fd(), EpollEvent::new(EpollFlags::EPOLLIN,DATA))?;
```
asomers, line 87: If you impl AsFd for Epoll, then you can simplify this statement:
```suggestion
/// epoll.add(&eventfd, EpollEvent::new(EpollFlags::EPOLLIN,DATA))?;
```

src/sys/epoll.rs
```diff
@@ -70,20 +69,126 @@ impl EpollEvent {
...
+    /// instance referred to by `self`. It requests that the operation `op` be performed for the
+    /// target file descriptor, `fd`.
+    /// 
+    /// When possible prefer [`Epoll::add`], [`Epoll::delete`] and [`Epoll::modify`].
+    /// 
+    /// [`epoll_ctl`](https://man7.org/linux/man-pages/man2/epoll_ctl.2.html)
+    pub fn epoll_ctl<'a, T>(
+        &self,
+        op: EpollOp,
+        fd: RawFd,
```
asomers, line 161: In the future, we should probably make this BorrowedFd, but not until that type has had time to trickle through the ecosystem.

src/sys/epoll.rs
```diff
@@ -70,20 +69,126 @@ impl EpollEvent {
...
+            )
+        };
+    
+        Errno::result(res).map(|r| r as usize)
+    }
+    /// This system call is used to add, modify, or remove entries in the interest list of the epoll
+    /// instance referred to by `self`. It requests that the operation `op` be performed for the
+    /// target file descriptor, `fd`.
+    /// 
+    /// When possible prefer [`Epoll::add`], [`Epoll::delete`] and [`Epoll::modify`].
```
asomers, line 155: Is there any reason to make this public?  Any operation that can't be performed with `add`, `delete`, or `modify`?

asomers (changes requested):

CHANGELOG.md
JonathanWoollett-Light: Rebased.

src/sys/epoll.rs
JonathanWoollett-Light: I've updated it to be `Fd` where `Fd: AsFd` matching the other functions like `Epoll::add` etc. I'm not sure about the case of specifically using `BorrowedFd` as this prohibits someone simply passing a reference to an `OwnedFd` and requires `fd.as_fd()` rather than `&fd`.

src/sys/epoll.rs
JonathanWoollett-Light: No, I'll make it private.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - The MSRV is now 1.63
   ([#1862](https://github.com/nix-rust/nix/pull/1862))
+- The MSRV is now 1.64.0.
+  ([#1882](https://github.com/nix-rust/nix/pull/1882))
```
asomers, line 18: You can just delete this entry now.
```suggestion
```

asomers (changes requested):

CHANGELOG.md
JonathanWoollett-Light: Removed.

asomers (approved):

asomers (approved):
bors r+

PR #1885, opened 2022-11-28
Update to memoffset 0.7

sdroege:


rtzoeller:
@asomers I had a hard time tracking down a changelog for `memoffset`, but a quick perusal through the [git history](https://github.com/Gilnaa/memoffset/commits/master) looks fairly innocuous. Mostly doc/test updates, but they did add support for unions.

Any reason we shouldn't pull this in for 0.26.0?

asomers:
Perhaps a bigger question is, why _should_ we bring it in at all?  What's the motivation for the change, @sdroege ?

rtzoeller:
> Perhaps a bigger question is, why _should_ we bring it in at all? What's the motivation for the change, @sdroege ?

We are one of the most-used dependencies on `memoffset` (https://lib.rs/crates/memoffset/rev), and us being pinned to 0.6 may keep others from being able to use a newer version due to the semver being less than 1.0.

asomers:
> We are one of the most-used dependencies on `memoffset` (https://lib.rs/crates/memoffset/rev), and us being pinned to 0.6 may keep others from being able to use a newer version due to the semver being less than 1.0.

That makes sense.  I had hoped to hear the author's motivation.  But I want to get 0.26.0 out soon, and don't want to wait.

asomers (approved):
bors r+

sdroege:
> We are one of the most-used dependencies on `memoffset` (https://lib.rs/crates/memoffset/rev), and us being pinned to 0.6 may keep others from being able to use a newer version due to the semver being less than 1.0.

Yes exactly that. Because of `nix` I had multiple versions of `memoffset` in the dependency tree of a few of my crates.

PR #1886, opened 2022-11-28
Update use of libc::timespec to prepare for future libc version

wesleywiser:

`x86_64-unknown-linux-gnu`).

See also https://github.com/rust-lang/libc/pull/2088

### DIFF
--- a/src/sys/time.rs
+++ b/src/sys/time.rs
@@ -6,6 +6,13 @@ use std::convert::From;
 use std::time::Duration;
 use std::{cmp, fmt, ops};
 
+const fn zero_init_timespec() -> timespec {
+    // `std::mem::MaybeUninit::zeroed()` is not yet a const fn
+    // (https://github.com/rust-lang/rust/issues/91850) so we will instead initialize an array of
+    // the appropriate size to zero and then transmute it to a timespec value.
+    unsafe { std::mem::transmute([0u8; std::mem::size_of::<timespec>()]) }
+}
+
 #[cfg(any(
     all(feature = "time", any(target_os = "android", target_os = "linux")),
     all(
@@ -20,7 +27,7 @@ use std::{cmp, fmt, ops};
     )
 ))]
 pub(crate) mod timer {
-    use crate::sys::time::TimeSpec;
+    use crate::sys::time::{zero_init_timespec, TimeSpec};
     use bitflags::bitflags;
 
     #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


src/sys/time.rs
```diff
@@ -6,6 +6,13 @@ use std::convert::From;
 use std::time::Duration;
 use std::{cmp, fmt, ops};
 
+const fn zero_init_timespec() -> timespec {
+    // `std::mem::zeroed()` is not yet a const fn (https://github.com/rust-lang/rust/issues/62061)
```
asomers, line 10: Another option is `std::mem::MaybeUninit::zeroed()`.  It's also not yet const (https://github.com/rust-lang/rust/issues/91850), but seems to be moving faster.

src/sys/time.rs
```diff
@@ -118,6 +116,7 @@ pub(crate) mod timer {
                         libc::timespec {
                             tv_sec: 0,
                             tv_nsec: 0,
+                            ..
```
asomers, line 119: What's the function of this line?  Are you programming defensively, just in case freebsd, netbsd, dragonfly, or illumos grow some padding fields too?

asomers (commented):
You should rebase on master, which already has the Clippy fixes.  Otherwise, this change looks good.

src/sys/time.rs
wesleywiser: It's required once Rust starts targeting musl 1.2 as `i686-unknown-linux-musl` will have a padding field:

```
$ cargo +e392e7b50913d5cb6e2c08d70cdeb72e56d289f9 build --target i686-unknown-linux-musl
   Compiling libc v0.2.137 (https://github.com/wesleywiser/libc?branch=musl-1.2#3049e59a)
   Compiling nix v0.25.0 (/tmp/nix)
error: pattern requires `..` due to inaccessible fields
   --> src/sys/time.rs:116:25
    |
116 | /                         libc::timespec {
117 | |                             tv_sec: 0,
118 | |                             tv_nsec: 0,
119 | |                         },
    | |_________________________^
    |
help: ignore the inaccessible and unused fields
    |
118 |                             tv_nsec: 0, ..,
    |                                       ++++
```

src/sys/time.rs
wesleywiser: Updated the comment to reflect that.

src/sys/time.rs
asomers: Oh, my bad.  I saw that `#[cfg()]` block on line 100 and thought that it applied here.

asomers (approved):
bors r+

src/sys/time.rs
wesleywiser: No problem, thanks for the quick reviews! 🙂

asomers:
@wesleywiser it looks like `mem::transmute` only became `const` in Rust 1.56.1.  That means we can't backport this PR.  Can you come up with a version that doesn't rely on `mem::transmute`?  CC @rtzoeller .

wesleywiser:
Sorry for the delay! I'll post the backports tomorrow.

PR #1890, opened 2022-11-29
[skip ci] add CHANGELOG entries for old point releases.

asomers:
I opted to preserve explicit entries for backports in both the original feature release and the backported point release, rather than trying to pretend that the releases were actually sequential.

Also, remove some empty subsections from the file.


rtzoeller (approved):

rtzoeller:
@asomers merge when ready; I'm ignoring bors.

PR #1894, opened 2022-11-30
Update use of libc::timespec to prepare for future libc version (0.25 backport)

wesleywiser:
This is a backport of 006fc6f7975b3a6b64329847b780622aab392109. The original commit message follows:

`x86_64-unknown-linux-gnu`).


wesleywiser:
The CI failures seem to be clippy warnings but nothing more severe than that.

rtzoeller:
This was included in v0.25.1.

Issue #1884, opened 2022-11-27
expose AT_EMPTY_PATH flag with linkat / LinkatFlags 


xaverdh:
Currently you can't set this flag..

This is useful in the following situation:
You create a temporary file with `O_TMPFILE` and without `O_EXCL`, then you want to link it somewhere, but have no old file path you could give `linkat`.

asomers:
Yeah, that would be good.  Care to submit a PR?

xaverdh:
Ok, I will give it a try

xaverdh:
What is the policy regarding api breakage? I am tempted to just make the `oldpath` argument an `Option`..
I can only think of two ways to preserve the api:
* introduce a new function with the new api
*  make the argument polymorphic via a trait that the old `LinkatFlags` then implements

xaverdh:
Ok, so the linked pr now just swaps `LinkatFlags` for a `bitflags` based implementation. That will cause breakage of course..

PR #1895, opened 2022-11-30
extend LinkatFlags to allow passing AT_EMPTY_PATH

xaverdh:
pr for https://github.com/nix-rust/nix/issues/1884

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1231,6 +1231,7 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
 libc_bitflags! {
     /// Flags for `linkat` function.
     pub struct LinkatFlags: c_int {
+        #[cfg(not(target_os = "redox"))]
         AT_SYMLINK_FOLLOW;
         #[cfg(any(target_os = "android", target_os = "linux"))]
         AT_EMPTY_PATH;
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1228,11 +1228,13 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
     }
 }
 
-/// Flags for `linkat` function.
-#[derive(Clone, Copy, Debug)]
-pub enum LinkatFlags {
-    SymlinkFollow,
-    NoSymlinkFollow,
+libc_bitflags! {
+    /// Flags for `linkat` function.
+    pub struct LinkatFlags: c_int {
+        AT_SYMLINK_FOLLOW;
+        #[cfg(any(target_os = "android", target_os = "linux"))]
+        AT_EMPTY_PATH;
+    }
 }
 
 /// Link one file to another file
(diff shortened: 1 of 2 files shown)


xaverdh:
Ok this does not work as is obviously because `SymlinkFollow` is not a member of `LinkatFlags` anymore.. is there a way to have them both in the same namespace?

src/unistd.rs
```diff
@@ -1272,7 +1273,7 @@ pub fn linkat<P: ?Sized + NixPath>(
                     oldcstr.as_ptr(),
                     at_rawfd(newdirfd),
                     newcstr.as_ptr(),
-                    atflag.bits() as libc::c_int
+                    flags.bits() as libc::c_int
```
SteveLauC, line 1276: This type cast is unnecessary as `flags.bits()` already returns `libc::c_int`

src/unistd.rs
```diff
@@ -1228,11 +1228,14 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
     }
 }
 
-/// Flags for `linkat` function.
-#[derive(Clone, Copy, Debug)]
-pub enum LinkatFlags {
-    SymlinkFollow,
-    NoSymlinkFollow,
+libc_bitflags! {
```
SteveLauC, line 1231: This type itself should be `#[cfg(not(target_os = "redox"))]` as the file system on redox does not support symlink

src/unistd.rs
```diff
@@ -1228,11 +1228,14 @@ pub fn isatty(fd: RawFd) -> Result<bool> {
...
+    /// Flags for `linkat` function.
+    pub struct LinkatFlags: c_int {
+        #[cfg(not(target_os = "redox"))]
+        AT_SYMLINK_FOLLOW;
+        #[cfg(any(target_os = "android", target_os = "linux"))]
```
SteveLauC, line 1236:
```suggestion
        #[cfg(linux_android)]
```

src/unistd.rs
SteveLauC: Actually, I am not sure if we should define a new type or just use the [`AtFlags`](https://docs.rs/nix/latest/nix/fcntl/struct.AtFlags.html) type.

It is safer to define a new type so that we can prevent users from passing invalid arguments, but this will add our maintainence burgen. We are already using `AtFlags` type for other functions like [`fstatat`](https://docs.rs/nix/latest/nix/sys/stat/fn.fstatat.html), cc @asomers, what do you think?

SteveLauC (changes requested):
Sorry for the late review, I have left some comments

src/unistd.rs
SteveLauC: Link [Add EmptyPath for FchownatFlags](https://github.com/nix-rust/nix/pull/1128) as these 2 PRs are similar

PR #1898, opened 2022-12-01
Update use of libc::timespec to prepare for future libc version (0.24 backport)

wesleywiser:
This is a backport of 006fc6f7975b3a6b64329847b780622aab392109. The original commit message follows:

`x86_64-unknown-linux-gnu`).

### DIFF
--- a/src/sys/time.rs
+++ b/src/sys/time.rs
@@ -5,6 +5,10 @@ use libc::{timespec, timeval};
 #[cfg_attr(target_env = "musl", allow(deprecated))] // https://github.com/rust-lang/libc/issues/1848
 pub use libc::{time_t, suseconds_t};
 
+const TIMESPEC_ZERO: libc::timespec = unsafe {
+    std::mem::transmute([0u8; std::mem::size_of::<libc::timespec>()])
+};
+
 #[cfg(any(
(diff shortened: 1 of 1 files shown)


rtzoeller:
Thanks! 

I plan to get the patch releases handled in the next few hours; I will likely just cherry pick these back though, rather than merging the PRs.

wesleywiser:
Sounds good! I went ahead and provided the backports for all of (`0.25`, `0.24`, `0.23` and `0.22`) in various PRs. All of the backports required resolving some merge conflicts with the base branch so feel free to use them if you want 🙂 

I also tested that each built against current `i686-unknown-linux-musl`, `i686-unknown-linux-musl` with the musl 1.2 changes, `i686-unknown-linux-gnu`, `x86-64-unknown-linux-musl` and `x86-64-unknown-linux-gnu`.

rtzoeller:
This was included in v0.24.3.

rtzoeller:
@wesleywiser we still need to handle 0.23 and 0.22, but 0.25 and 0.24 are done. Thanks for backporting your change - it was helpful not to have to handle the merge conflicts for one of the commits myself! 😃

wesleywiser:
@rtzoeller You're welcome, happy to help out! Thanks so much for including these 🙂

PR #1902, opened 2022-12-03
Enable ucontext module on aarch64-unknown-linux-gnu

rtzoeller:
Untested, but I saw https://github.com/rust-lang/libc/commit/fd32da6e7dfa2afcae86e176904244cf45a90c06 and figured we should uptake it.


asomers (approved):
bors r+

PR #1906, opened 2022-12-04
Expose copy_file_range on FreeBSD and use I/O Safety

asomers:
Depends on #1862


asomers:
@rtzoeller suddenly I'm getting a bunch of errors on musl, android, and qemu.  But I can't figure out why they didn't happen before this PR.  Any guesses?

- On android, `libc::copy_file_range` doesn't exist.  But why didn't that cause problems before?
- Same thing on musl.
- On QEMU, `undefined reference to `copy_file_range'`  at link time.  But again, what changed?

Even before this PR, copy_file_range should've been enabled for all of those targets.

asomers:
Oh, I see.  The old version used `libc::syscall(libc::SYS_copy_file_range, ...)`.  Now I'm trying to call `libc::copy_file_range(...)`.  Silly Linux.  I wonder if it actually worked before?

PR #1907, opened 2022-12-04
Futex

JonathanWoollett-Light:


rtzoeller:
Does https://github.com/m-ou-se/linux-futex meet your needs?

JonathanWoollett-Light:
> Does https://github.com/m-ou-se/linux-futex meet your needs?

This is a useful reference, thank you. I do think this functionality should be included in nix.

rtzoeller:
> I do think this functionality should be included in nix.

I don't disagree in principle, but I'd like to see nix's support provide some value-add compared to prior art. For example, Mara's crate only (appears to) supports Linux futexes. It'd be nice if nix supported the BSD variants as well.

JonathanWoollett-Light:
> > I do think this functionality should be included in nix.
> 
> I don't disagree in principle, but I'd like to see nix's support provide some value-add compared to prior art. For example, Mara's crate only (appears to) supports Linux futexes. It'd be nice if nix supported the BSD variants as well.

I haven't implemented BSD since it is not personally useful to me. I think if someone needs this they can implement it. This is still a significant value add to nix and I would also argue my implementation is better than [linux-futex](https://github.com/m-ou-se/linux-futex).

src/sys/futex.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// Creates a file descriptor associated with the futex.
+    ///
+    /// When [`Futex::wake`] is performed on the futex this file indicates being readable with
+    /// `select`, `poll` and `epoll`.
+    ///
+    /// The file descriptor can be used to obtain asynchronous notifications: if val is nonzero,
+    /// then, when another process or thread executes a FUTEX_WAKE, the caller will receive the
+    /// signal number that was passed in val.
+    ///
+    /// **Because it was inherently racy, this is unsupported from Linux 2.6.26 onward.**
```
rtzoeller, line 160: If  this is racy and has been removed since 2008, why wrap it?

src/sys/futex.rs
```diff
@@ -0,0 +1,363 @@
...
+/// ```
+#[derive(Debug)]
+pub struct Futex<const PRIVATE: bool = false>(UnsafeCell<u32>);
+impl<const PRIVATE: bool> Futex<PRIVATE> {
+    const MASK: i32 = if PRIVATE { libc::FUTEX_PRIVATE_FLAG } else { 0 };
+    /// Constructs new futex with a given `val`.
+    pub fn new(val: u32) -> Self {
+        Self(UnsafeCell::new(val))
+    }
+    /// If the value of the futex:
```
rtzoeller, line 111: The rest of the nix source puts newlines between structs, impls, functions, etc. I'd appreciate if we matched that style.

src/sys/futex.rs
```diff
@@ -0,0 +1,363 @@
...
+/// const TIMEOUT: Duration = Duration::from_millis(500);
+/// const DELTA: Duration = Duration::from_millis(100);
+/// # fn main() -> nix::Result<()> {
+/// let futex: Futex = Futex::new(0);
+///
+/// // If the value of the futex is 0, wait for wake. Since the value is 0 and no wake occurs,
+/// // we expect the timeout will pass.
+///
+/// let instant = Instant::now();
+/// assert_eq!(futex.wait(0, Some(TIMEOUT)),Err(Errno::ETIMEDOUT));
```
rtzoeller, line 49: My familiarity is mostly with the PI variants of futexes, but my understanding is that proper futex use in general involves trying the operation in userspace via atomic operations before falling back to a syscall.

We should make sure the examples/tests demonstrate this, if only to avoid being incredibly confusing about their proper use.

rtzoeller (commented):
> This is still a significant value add to nix and I would also argue my implementation is smaller and simpler than [linux-futex](https://github.com/m-ou-se/linux-futex) and this I would argue makes it better.

I completely agree that it's a simpler/more direct wrapping of the underlying futex operations. However, given that there's an existing implementation which (IMO) provides a nice abstraction over the futex syscall, I'm uncertain that splitting the investment into Rust futex support is actually better outcome.

As I mentioned, I don't object to including futex support in the nix crate if there is a compelling reason to do so. What is the use case where the `linux-futex` crate is insufficient? Missing functionality?

src/sys/futex.rs
JonathanWoollett-Light: Fixed.

src/sys/futex.rs
JonathanWoollett-Light: It is a part of the public futex API. I have no strong opinion either way on keeping or removing it here.

src/sys/mod.rs
```diff
@@ -226,3 +226,7 @@ feature! {
     #![feature = "time"]
     pub mod timer;
 }
+
+/// Fast user-space locking.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+pub mod futex;
```
rtzoeller, line 232: We should probably put this behind a feature (see #1611).

JonathanWoollett-Light: Added a `futex` feature.

src/sys/futex.rs
JonathanWoollett-Light: My understanding is that a `futex` is used like a Rust [`Condvar`](https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html). I'm not sure what you mean by:

> proper futex use in general involves trying the operation in userspace via atomic operations before falling back to a syscall.

or how this would look.

Could you maybe provide an example?

rtzoeller: See https://linux.die.net/man/4/futex for the description (note it's futex(4), not the typical futex(2)).

The point of futexes is that syscalls are slow, but we can make the operations faster in uncontended cases by never leaving userspace. This is accomplished by doing a CAS operation and only performing the syscall if it fails.

See https://github.com/rtzoeller/rust-priority-inheriting-lock/blob/6225e90971d9aa72f070145788669e421cf012d4/src/lib.rs#L48 for an example of the userspace operation and falling back to the syscall for LOCK_PI (incidentally using linux-futex).

JonathanWoollett-Light: I don't think this matter in non-highly contentious cases, not all use cases of a futex are highly contentious and not all are using it to implement a locking mechanism.

To have an example that properly illustrates what you suggest would be seem to need to be including an implementation of a priority-inheriting lock as an example.

Relating to docs, if you have suggestions for additions here please add a suggestion. Perhaps a comment suggesting that if a user is looking for a locking mechanism and does not require handling of priority-inheritance, they should use a mutex?

I'm not sure otherwise what changes you would suggest?

asomers:
I still don't understand what advantage this PR has over the existing linux-futex crate.

JonathanWoollett-Light:
> I still don't understand what advantage this PR has over the existing linux-futex crate.

The point of nix is to provide bindings to Linux.

Providing bindings to Linux except for functionality that is already covered by other crates would mean removing like 80% of already implemented functionality.

asomers:
> The point of nix is to provide wrappers to Unix-like systems..

The point of Nix is to provide idiomatic zero-cost access to Unix system APIs.  But providing access to ALL Unix system APIs isn't a goal.  Nor is it productive to duplicate work that others have already done.  That wastes developer resources and increases our maintenance burden.  If another crate already does a good job, then there's no reason why we should need to, too.

JonathanWoollett-Light:
> > The point of nix is to provide wrappers to Unix-like systems..
> 
> The point of Nix is to provide idiomatic zero-cost access to Unix system APIs. But providing access to ALL Unix system APIs isn't a goal. Nor is it productive to duplicate work that others have already done. That wastes developer resources and increases our maintenance burden. If another crate already does a good job, then there's no reason why we should need to, too.

If this is really the path Nix pursues it should be documented:

> Nix seeks to provide friendly bindings to various *nix platform APIs (Linux, Darwin, ...) except where existing safe bindings exist in the eco-system.

I do not agree with this, I see the reasoning but I think it is shallow and misguided.

I could for instance post an issue on half the functionality on Nix saying it should be removed since decent alternatives exist.

I could also leave this comment on most currently open PRs saying a decent alternative exists that support this.

PR #1910, opened 2022-12-04
Added I/O safety to sys::aio

asomers:


asomers:
Before merging this I want to check that it works with some downstream crates.

src/fcntl.rs
```diff
@@ -571,6 +571,9 @@ libc_bitflags! {
 feature! {
 #![feature = "zerocopy"]
 
+    pub fn xxx() {
+        libc::xdfsersesefsfe();
```
rtzoeller, line 575: Seems legit.

asomers: Ahh, crap.  I've been jumping between too many branches today.

asomers:
So this branch works as-is.  However, after a few months of on-off trying, I've been unable to make the downstream crates work with it.  The fundamental problem is that all of AIO's interesting applications require heap-allocating the aiocb.  For example, when using a reactor like Tokio.  And heap allocation requires that the object be `'static`.  Speaking of Tokio, that project doesn't intend to ever adopt I/O Safety.  And since it's the major downstream consumer of Nix's `aio` module, that limits the usefuless of I/O Safety here.

So for now I'm going to close this PR.  If Tokio ever changes their mind, or if a different consumer decides to use I/O Safety, then we can reopen it.  Making I/O Safety work with `'static` aiocbs would probably require implementing both `from_fd` and `from_rawfd` methods, and doing something like this following in a downstream crate:
```rust
struct AiocbWrapper {
    struct nix::sys::aio::AioWrite<'static>,
    f: Arc<OwnedFd>
}
```

PR #1913, opened 2022-12-05
feat: I/O safety for 'sys/inotify'

SteveLauC:
#### What this PR does:

1. Changes the `fd` field of `struct Inotify` from `RawFd` to `OwnedFd`
2. Changes the interfaces of functions in the `impl Inotify {}`
   
    > The type of `self` changes from `Self` to `&mut Self`. 
    
    From:

   ```rust
   pub fn add_watch<P: ?Sized + NixPath>(
         self,
         path: &P,
         mask: AddWatchFlags,
   ) -> Result<WatchDescriptor> 

   pub fn rm_watch(self, wd: WatchDescriptor) -> Result<()>

   pub fn read_events(self) -> Result<Vec<InotifyEvent>>
   ```

   To:

   ```rust
   pub fn add_watch<P: ?Sized + NixPath>(
        &mut self,
        path: &P,
        mask: AddWatchFlags,
    ) -> Result<WatchDescriptor>

   pub fn rm_watch(&mut self, wd: WatchDescriptor) -> Result<()>

   pub fn read_events(&mut self) -> Result<Vec<InotifyEvent>>
   ```
  

   In the previous implementation, these functions can take `self` by value as `struct Inotify` [was `Copy`](https://docs.rs/nix/latest/nix/sys/inotify/struct.Inotify.html#impl-Copy-for-Inotify). With the changes in `1` applied, `struct Inotify` is no longer `Copy`, so we have to take `self` by reference.

-------

Blocks until the merge of #1863 as this PR needs `read(2)` to be I/O-safe.

### DIFF
--- a/src/sys/inotify.rs
+++ b/src/sys/inotify.rs
@@ -32,7 +32,7 @@ use libc::{c_char, c_int};
 use std::ffi::{CStr, OsStr, OsString};
 use std::mem::{size_of, MaybeUninit};
 use std::os::unix::ffi::OsStrExt;
-use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, OwnedFd, RawFd};
 use std::ptr;
 
 libc_bitflags! {
@@ -101,9 +101,9 @@ libc_bitflags! {
 
 /// An inotify instance. This is also a file descriptor, you can feed it to
 /// other interfaces consuming file descriptors, epoll for example.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug)]
 pub struct Inotify {
-    fd: RawFd,
+    fd: OwnedFd,
 }
 
 /// This object is returned when you create a new watch on an inotify instance.
@@ -143,7 +143,7 @@ impl Inotify {
     pub fn init(flags: InitFlags) -> Result<Inotify> {
         let res = Errno::result(unsafe { libc::inotify_init1(flags.bits()) });
 
-        res.map(|fd| Inotify { fd })
+        res.map(|fd| Inotify { fd: unsafe { OwnedFd::from_raw_fd(fd) } })
     }
 
     /// Adds a new watch on the target file or directory.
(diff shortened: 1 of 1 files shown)


src/sys/inotify.rs
```diff
@@ -235,14 +235,8 @@ impl Inotify {
     }
 }
 
-impl AsRawFd for Inotify {
-    fn as_raw_fd(&self) -> RawFd {
-        self.fd
-    }
-}
-
```
SteveLauC, line 243: This trait impl extracts the `RawFd` from `struct Inofity`, IMO, this is used to `close(2)` the file descriptor representing the `inotify` instance, which is no longer needed as `OwnedFd` will drop it automatically.

asomers:
The CI failure looks directly related to this PR.

SteveLauC:
> The CI failure looks directly related to this PR.

Yes, that's why this PR needs to be blocked until the merge of #1863

SteveLauC:
@asomers I have fixed the error (by using the old `RawFd` `read(2)` interface) and this is ready for review.

src/sys/inotify.rs
```diff
@@ -152,12 +152,12 @@ impl Inotify {
     ///
     /// For more information see, [inotify_add_watch(2)](https://man7.org/linux/man-pages/man2/inotify_add_watch.2.html).
     pub fn add_watch<P: ?Sized + NixPath>(
-        self,
+        &mut self,
```
asomers, line 155: Why `&mut self` and not `&self`?

SteveLauC: Semantically, inotify_add_watch modifies the existing inotify instance, though this can not be reflected in the implementation of this binding.

[Man 2 inotify_add_watch](https://man7.org/linux/man-pages/man2/inotify_add_watch.2.html)

asomers: I consider this is a case of "internal mutability". The correctness of the the method doesn't depend on it having exclusive access to the the structure, so we can remove the `mut`.

SteveLauC: Get it, this can also be applied to rm_watch and read_events

asomers (approved):
bors r+

VorpalBlade:
It seems this broke monitoring Inotify with Epoll since it removed the AsRawFd but didn't implement any other way to combine it will Epoll (or plain Poll for that matter). This makes it impossible to monitor both inotify files and some other FDs at the same time.

See also issue #1998

PR #1914, opened 2022-12-05
Added CanAddr for binding sockets for CANbus access

fpagliughi:
This adds a `CanAddr` type for CANbus access. It would allow an application to `bind()` a SocketCAN socket.

I have no idea if I did this properly; I just tried some copy-pasta from the `NetlinkAddr`. But it does appear to wok and I can use it to successfully bind to a can interface.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -1763,6 +1763,89 @@ pub mod netlink {
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 #[cfg_attr(docsrs, doc(cfg(all())))]
+pub mod can {
+    use super::*;
+    use crate::sys::socket::addr::AddressFamily;
+    use libc::{sa_family_t, sockaddr_can, c_int, c_uint};
+    use std::{fmt, mem};
+
+    /// Address for the Linux SocketCAN implementation for the Controller Area Network
+    /// automotive bus protocol.
+    ///
+    /// # References
+    ///
+    /// [SocketCAN](https://docs.kernel.org/networking/can.html)
+    #[derive(Copy, Clone)]
+    #[allow(missing_debug_implementations)]
+    #[repr(transparent)]
+    pub struct CanAddr(pub(in super::super) sockaddr_can);
+
+    impl CanAddr {
+        /// Construct a new CAN socket address from its interface index.
+        pub fn new(ifindex: c_uint) -> Self {
+            let mut addr: sockaddr_can = unsafe { mem::zeroed() };
+            addr.can_family = AddressFamily::Can as sa_family_t;
+            addr.can_ifindex = ifindex as c_int;
+
+            Self(addr)
+        }
+
+        /// Return the socket's CAN interface index.
+        pub const fn ifindex(&self) -> c_uint {
+            self.0.can_ifindex as c_uint
+        }
+    }
+
+    impl private::SockaddrLikePriv for CanAddr {}
+    impl SockaddrLike for CanAddr {
+        unsafe fn from_raw(
+            addr: *const libc::sockaddr,
+            len: Option<libc::socklen_t>,
+        ) -> Option<Self>
+        where
+            Self: Sized,
+        {
+            if let Some(l) = len {
+                if l != mem::size_of::<libc::sockaddr_can>() as libc::socklen_t {
+                    return None;
+                }
+            }
+            if (*addr).sa_family as i32 != libc::AF_CAN {
+                return None;
+            }
+            Some(Self(ptr::read_unaligned(addr as *const _)))
+        }
+    }
+
+    impl AsRef<libc::sockaddr_can> for CanAddr {
+        fn as_ref(&self) -> &libc::sockaddr_can {
+            &self.0
+        }
+    }
+
+    impl fmt::Display for CanAddr {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            write!(f, "ifindex: {}", self.ifindex())
+        }
+    }
+
+    impl PartialEq for CanAddr {
+        fn eq(&self, other: &CanAddr) -> bool {
+            self.ifindex() == other.ifindex()
+        }
+    }
+
+    impl Eq for CanAddr {}
+
+    impl Hash for CanAddr {
+        fn hash<H: Hasher>(&self, s: &mut H) {
+            self.ifindex().hash(s)
+        }
+    }
+}
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[cfg_attr(docsrs, doc(cfg(all())))]
 pub mod alg {
     use super::*;
     use libc::{c_char, sockaddr_alg, AF_ALG};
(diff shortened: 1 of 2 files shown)


src/sys/socket/addr.rs
```diff
@@ -1761,6 +1761,89 @@ pub mod netlink {
...
+        }
+    }
+
+    impl fmt::Display for CanAddr {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            write!(f, "ifindex: {}", self.ifindex())
+        }
+    }
+
+    impl PartialEq for CanAddr {
```
asomers, line 1832: This implementation ignores the union field.  Does the union field really not matter?  It's usually not possible to implement `PartialEq` for anything with union fields, unless you have some way to know which field is valid.

asomers (changes requested):

src/sys/socket/addr.rs
```diff
@@ -1761,6 +1761,89 @@ pub mod netlink {
...
+        pub fn new(ifindex: c_uint) -> Self {
+            let mut addr: sockaddr_can = unsafe { mem::zeroed() };
+            addr.can_family = AddressFamily::Can as sa_family_t;
+            addr.can_ifindex = ifindex as c_int;
+
+            Self(addr)
+        }
+
+        /// Return the socket's CAN interface index.
+        pub const fn ifindex(&self) -> c_uint {
```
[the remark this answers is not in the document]
fpagliughi: TLDR: A valid index can only be >=0 (zero for "bind all" and >=1 for a specific index). The `new()` function takes a uint and the getter returns that uint as the same type. The weird this is that  the `sockaddr_can` struct takes it as signed int, though a valid one can never be less than zero. So keeping the public API as uint seems the most consistent.

More details :

This is actually a weird one. 

The C function [if_nametoindex](https://man7.org/linux/man-pages/man3/if_nametoindex.3.html) is used to look up the index number from a name (like "can0", "can1", "vcan0", etc). Very strangely, it returns an `unsigned` with zero indicating an error (instead of the normal C API convention of returning -1 on error) and >=1 giving a valid index number.

The nix call to get the interface number, [if_nametoindex](https://docs.rs/nix/latest/nix/net/if_/fn.if_nametoindex.html) returns a `Result<c_uint>`, which makes perfect sense.

But then, for some reason, the `sockaddr_can` struct takes a signed `int` vlaue (`c_int`). But it really can only be an int >= 1.

src/sys/socket/addr.rs
fpagliughi: The most common base raw socket address does not use the union at all for anything. It just needs to specify the family (always `AF_CAN`) and the interface number. This is all that the current and next major releases of the `socketcan-rs` crate will cover.

This is all that I was thinking this `CanAddr` is meant to model.

The union is used for higher level protocols like J1939 or ISO-TP. I was thinking that those might be better modeled as separate address types like, maybe, `CanJ1939Addr`.  I don't have a really strong opinion on this, but considering that the protocols aren't typically mixed, and are implemented by separate Rust crates, having different targeted address structs at this level makes some sense.

But if we want to make this one implement all the different possibilities in one struct, I could take a stab at it.

src/sys/socket/addr.rs
```diff
@@ -1761,6 +1761,89 @@ pub mod netlink {
...
+
+    impl PartialEq for CanAddr {
+        fn eq(&self, other: &CanAddr) -> bool {
+            self.ifindex() == other.ifindex()
+        }
+    }
+
+    impl Eq for CanAddr {}
+
+    impl Hash for CanAddr {
```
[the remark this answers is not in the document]
fpagliughi: As stated above, this implementation works for the common/base CAN address in which the union is zeroed out and ignored; and the only thing that could change from one address to another is the value of the index.

But if we wanted to implement the two other protocols in this struct (J1939 and ISO-TP), then, since the struct  is Sized, could we just do a bitwise comparison for  `PartialEq`? And a hash of the bytes as a slice?

Would that work?

src/sys/socket/addr.rs
fpagliughi: As an alternate, more complete implementation, we can make new functions for the additional protocols:

https://github.com/fpagliughi/nix/blob/88df2c37b54a7444257d8f3ae7ad928e3bea4aea/src/sys/socket/addr.rs#L1784-L1808

Then we could implement `PartialEq` as a bitwise binary comparison of the whole struct, and `Hash` by considering the struct similarly as a slice of bytes over the size of the underlying sockaddr struct.

https://github.com/fpagliughi/nix/blob/88df2c37b54a7444257d8f3ae7ad928e3bea4aea/src/sys/socket/addr.rs#L1849-L1879

This works insofar as two addresses created with the same parameters to any of the `new...()` functions will report as equal, and any two addresses that are equal will produce the same hash.

There is the argument that addresses created for the different protocols could, with certain parameters, produce the same binary address value, even though they were logically different as they were meant for different protocols. If that argument were to prevail, I think it would indicate that these are actually different address structures and should be different types.

I don't have a strong opinion. I could go either way. They both work equally well for my use case.

src/sys/socket/addr.rs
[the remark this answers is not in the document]
asomers: What if you get a sockaddr from a function like `getsockname`?  Then it might match any of these union variants, or even none of them.  Implementing `Hash` by casting to a byte slice would work only if you can guarantee that all of the memory will always be initialized.  That should be the case for constructors like `getsockname`, but what about manual construction?

src/sys/socket/addr.rs
fpagliughi: I agree this isn't a general way to compare unions. But all `sockaddr` types are like one big de-facto union in C, and the typical convention has always been to zero out the whole struct. This is enforced here with the `new...()` functions, but since users can create these types from underlying structs, they can bypass the conventions and create incompatible values.

But what bothers me more about this implementation is that users can compare, say, a J1939 address to a raw address or an ISO-TP address. These really aren't the same thing at all. They're actually different protocols.

So it makes me think these should be different types. So we could do something like this (yet another implementation):

https://github.com/fpagliughi/nix/blob/2276af8daa455a4e7762617872c40248fc629d8f/src/sys/socket/addr.rs#L1764-L2111

It's a bit verbose, and I wonder if there isn't some way to consolidate the common code between them. But it's clean in keeping the types separate, and the `PartialEq` and `Hash` implementations should be valid for each separate type.

src/sys/socket/addr.rs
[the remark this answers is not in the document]
fpagliughi: Yes, again, I was assuming this is not a general solution for all enums, but acceptable for this type of `sockaddr` where the convention is to zero out the struct.

But agree this is a cheat. The alternate implementation of separate implementations for each enum possibility would fix this.

src/sys/socket/addr.rs
asomers: If you split the type into three, how would `SockaddrLike::from_raw` work?

PR #1915, opened 2022-12-05
Add I/O safety to sockopt and some socket functions

asomers:
* socket
* socketpair
* listen
* setsockopt
* getsockopt

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -18,7 +18,7 @@ use libc::{
 use std::io::{IoSlice, IoSliceMut};
 #[cfg(feature = "net")]
 use std::net;
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, RawFd, OwnedFd};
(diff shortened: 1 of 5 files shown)


src/sys/socket/mod.rs
```diff
@@ -2076,7 +2079,13 @@ pub fn socket<T: Into<Option<SockProtocol>>>(
...
     let res = unsafe { libc::socket(domain as c_int, ty, protocol) };
 
-    Errno::result(res)
+    match res {
+        -1 => Err(Errno::last()),
+        fd => {
+            // Safe because libc::socket returned success
+            unsafe { Ok(OwnedFd::from_raw_fd(fd)) }
+        }
+    }
```
JonathanWoollett-Light, line 2088: I think this can be more safely and succinctly handled like:
```suggestion
    match res {
        -1 => Err(Errno::last()),
        // Safe because libc::socket returned success
        fd @ 0.. => unsafe { Ok(OwnedFd::from_raw_fd(fd)) },
        _ => unreachable!()
    }
```

PR #1916, opened 2022-12-05
Use I/O safety in sys::mman

asomers:


rtzoeller (approved):
Approved, but with a question: how do we want to handle the changelog for these changes, which do modify API signatures?

asomers:
> Approved, but with a question: how do we want to handle the changelog for these changes, which do modify API signatures?

Since all of these changes are related, I think we should roll them up into one big CHANGELOG entry at the top of the file.  The awesomest option would be if we could convince Cargo to print suggestions for how to fix users' code, at least for the common cases.  But I don't know how.

rtzoeller:
> The awesomest option would be if we could convince Cargo to print suggestions for how to fix users' code, at least for the common cases. But I don't know how.

I'll do some digging to see if there are attributes we can use.

bors r+

PR #1919, opened 2022-12-06
feat: I/O safety for 'sys/statfs'

SteveLauC:
### What this PR does:

1. Adds I/O safety for module `sys/statfs`. 

This PR is pretty small as all we need to do is to change the interface of `fstatfs(2)`:

from:

```rust
pub fn fstatfs<T: AsRawFd>(fd: &T) -> Result<Statfs> 
```

to:
```rust
pub fn fstatfs<Fd: AsFd>(fd: &Fd) -> Result<Statfs>
```

------

~Besides from the changes in module `sys/statfs`, there are two extra places where care needs to be taken:~

```shell
$ cd nix

# Search for the usage of `fstatfs(2)` in `nix`
$ rg "fstatfs\("
test/test_fcntl.rs
386:        let statfs = nix::sys::statfs::fstatfs(&tmp).unwrap();
424:        let statfs = nix::sys::statfs::fstatfs(&tmp).unwrap();

CHANGELOG.md
849:- Now functions `statfs()` and `fstatfs()` return result with `Statfs` wrapper

src/sys/statfs.rs
769:        check_fstatfs("/tmp");
770:        check_fstatfs("/dev");
771:        check_fstatfs("/run");
772:        check_fstatfs("/");
775:    fn check_fstatfs(path: &str) {
781:        let fs = fstatfs(&file).unwrap();
830:        let fs = fstatfs(&file);
```

~As you can see, `fstatfs(2)` is used in the tests in `test/test_fcntl.rs`:~

```rust
// Test code that involves `fstatfs(2)`

 let tmp: NamedTempFile = NamedTempFile::new().unwrap();

 let fd = tmp.as_raw_fd();
 let statfs = nix::sys::statfs::fstatfs(&tmp).unwrap();
```
~`tmp` is of type [`NamedTempFile`](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html), which does not implement `AsFd` in the current implementation of `tempfile`, but the implementation should be easy as it contains `std::fs::File` internally:~

```rust
pub struct NamedTempFile {
    path: TempPath,
    file: File,
}
```

~So I am thinking about making a PR to `tempfile` to make `NamedTempFile` `AsFd`, any thoughts on this?~

### DIFF
--- a/src/sys/statfs.rs
+++ b/src/sys/statfs.rs
@@ -5,7 +5,7 @@
 use std::ffi::CStr;
 use std::fmt::{self, Debug};
 use std::mem;
-use std::os::unix::io::AsRawFd;
+use std::os::unix::io::{AsFd, AsRawFd};
 
 use cfg_if::cfg_if;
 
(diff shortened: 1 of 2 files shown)


SteveLauC:
Ohh, I am too stupid, we can actually call [`.as_file()`](https://docs.rs/tempfile/3.3.0/tempfile/struct.NamedTempFile.html#method.as_file) on `NamedTempFile` to get `&File` to avoid this...

asomers (approved):
bors r+

PR #1920, opened 2022-12-07
Fix tests with --no-default-features.

plugwash:
While updating nix in Debian, I noticed that the tests fail when run with --no-default-features because the tests are not marked with the features they require, this patch adds those feature markers.

I have tested it with each feature individually, and also with the following feature combinations.

aio,signal
process,ptrace
process,fs
fs,user
fs,zerocopy
mount,sched
mount,fs,sched
mount,sched,user
mount,fs,sched,user
uio,process                 
socket,uio                  
socket,uio,fs               
socket,uio,term             
net,uio                     
net,uio,time
socket,user
socket,user,uio
socket,user,uio,process
socket,process
socket,uio,process
net,uio,process
net,uio,process,feature
net,uio,process,feature,time
poll,signal
process,signal
pthread,signal
ptrace,memoffset
ptrace,signal
ptrace,signal,memoffset
term,fs
term,signal
term,process
term,process,signal
time,process
uio,process,signal

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -1122,7 +1122,7 @@ mod sigevent {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test,any(feature = "signal")))]
 mod tests {
     use super::*;
     #[cfg(not(target_os = "redox"))]
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -2317,6 +2317,7 @@ pub mod vsock {
 mod tests {
     use super::*;
 
+    #[cfg(feature = "net")]
     mod types {
         use super::*;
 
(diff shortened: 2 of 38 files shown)


asomers:
When we originally added features to this crate, I deliberately didn't try to feature gate the tests.  I thought the maintenance effort would be too great, for too little benefit.  How important is it to you that you be able to run the tests with a limited feature set?

SteveLauC:
Hi @plugwash, thanks for your contribution to Nix!

I believe this PR is not what we want, as stated by other maintainers, this is done deliberately to relieve our maintenance stress, and it is documented that we should run `cargo test --all-features` when running tests in [`CONTRIBUTING.md`](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#testing), so I am gonna close this PR, thanks!

PR #1921, opened 2022-12-07
feat: I/O safety for 'sys/termios' & 'pty'

SteveLauC:
#### What this PR does:
1. Adds I/O safety for modules `sys/termios` and `pty`

------

#### Known Problems:

1. [Double free issue on `PtyMaster`](https://github.com/nix-rust/nix/issues/659)
                                            
   I have changed the `RawFd` in `PtyMaster` to `OwnedFd` in this PR, with this
   change, the double-free issue still exists, see this test code snippet 
   (From [this comment](https://github.com/nix-rust/nix/issues/659#issuecomment-315544022))

   ```rust
   use std::io::prelude::*;
   
   fn main() {
       let mut f = {
           let m = nix::pty::posix_openpt(nix::fcntl::OFlag::O_RDWR).unwrap(); // get fd 3
           nix::unistd::close(m.as_raw_fd()).unwrap(); // close fd 3
           std::fs::File::create("foo").unwrap()       // get fd 3 again
       }; // m goes out of scope, `drop(OwnedFd)`, fd 3 closed
       f.write("whatever".as_bytes()).unwrap(); // EBADF
   }
   ```

   I have tested this code with `nix 0.26.1`, and I am still getting `EBADF`, which means the current impl does not prevent this problem either.

   ```shell
   $ cat Cargo.toml | grep nix
   nix = "0.26.1"

   $ cargo r -q
   thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }', src/main.rs:10:36
   ```

   If we still wanna the drop of `PtyMaster` panic when the internal `fd` is invalid 
   as we did in #677, then we have to revert the changes to use `RawFd` and manually impl `Drop`.

2. Some trait implementations for some types are removed

   * `struct OpenptyResult`:
     1. PartialEq 
     2. Eq
     3. Hash
     4. Clone

   * `struct ForkptyResult`:
     1. Clone

   * `struct PtyMaster`:
     1. PartialEq
     2. Eq
     3. Hash

   In the previous implementation, these trait impls are `#[derive()]`ed, due to
   the type change to `OwnedFd`, we can no longer derive them. Should we manually
   implement them?

   I kinda think we should at least impl `PartialEq` and `Eq` for `OpenptyResult`
   and `PtyMaster`.

-----
#### Some Clarifications that may help code review

1. For the basic `fd`-related syscall like `read(2)`, `write(2)` and `fcntl(2)`
   , I am still using the old `RawFd` interfaces, as they will be covered in
   other PRs.

2. Two helper functions

   1. `write_all()` in `test/sys/test_termios.rs`:

      ```rust
      /// Helper function analogous to `std::io::Write::write_all`, but for `RawFd`s
      fn write_all(f: RawFd, buf: &[u8]) {
      /// Helper function analogous to `std::io::Write::write_all`, but for `Fd`s
      fn write_all<Fd: AsFd>(f: &Fd, buf: &[u8]) {
          let mut len = 0;
          while len < buf.len() {
              len += write(f, &buf[len..]).unwrap();
              len += write(f.as_fd().as_raw_fd(), &buf[len..]).unwrap();
          }
      }
      ```

   2. `read_exact()` in `test/test.rs`:

      ```rust
      /// Helper function analogous to `std::io::Read::read_exact`, but for `RawFD`s
      fn read_exact(f: RawFd, buf: &mut [u8]) {
      /// Helper function analogous to `std::io::Read::read_exact`, but for `Fd`s
      fn read_exact<Fd: AsFd>(f: &Fd, buf: &mut [u8]) {
          let mut len = 0;
          while len < buf.len() {
              // get_mut would be better than split_at_mut, but it requires nightly
              let (_, remaining) = buf.split_at_mut(len);
              len += read(f, remaining).unwrap();
              len += read(f.as_fd().as_raw_fd(), remaining).unwrap();
          }
      }
      ```

    I have added I/O safety for them, but it actually does not matter whether
    they use `Fd: AsFd` or `RawFd`. So feel free to ask me to discard these changes
    if you guys don't like it.


test/test_unistd.rs
```diff
@@ -555,16 +555,13 @@ fn test_lseek() {
...
     let offset: off_t = 5;
-    lseek(tmpfd, offset, Whence::SeekSet).unwrap();
+    lseek(tmp.as_raw_fd(), offset, Whence::SeekSet).unwrap();
 
     let mut buf = [0u8; 7];
-    crate::read_exact(tmpfd, &mut buf);
+    crate::read_exact(&tmp, &mut buf);
     assert_eq!(b"f123456", &buf);
-
-    close(tmpfd).unwrap();
```
SteveLauC, line 567: I don't quite understand this `close(2)` operation, `tmpfd` will be closed when `tmp` (of type: `File`) dropped, why do we need this extra `close(2)`?

src/pty.rs
```diff
@@ -16,78 +16,58 @@ use crate::{fcntl, unistd, Result};
 
 /// Representation of a master/slave pty pair
 ///
-/// This is returned by `openpty`.  Note that this type does *not* implement `Drop`, so the user
```
asomers, line 19: Since we explicitly called this out in the documentation before, do you think we should warn people in the CHANGELOG that they no longer need to close the file descriptors?

test/test_unistd.rs
SteveLauC: I get it now, missed that 'into_raw_fd()' yesterday.

Will remove into_raw_fd and close.

asomers (commented):
Soon we will change nix::unistd::close to take an OwnedFd, which will make the double-close situation _much_ harder to trigger.  Given that, I'm cool with this PR.  We don't need to worry overmuch about what happens in `Drop`.

I think it makes sense to drop those particular trait implementations.  They aren't very useful for file descriptors.  PartialEq used to be, but isn't any more now that OwnedFd isn't Clone.  So how would you ever get two PtyMaster objects that were equal?  Equally weird, one might expect that file descriptors returned by `dup2` would be equal to their originals, but they won't be for any likely implementation of PartialEq.

This PR looks pretty good.  I just wonder if we should add explicit instructions about Drop to the CHANGELOG.

src/pty.rs
SteveLauC: Yep, this should be documented, I will take care of it.

SteveLauC: Done.

asomers (approved):
bors r+

PR #1868, opened 2022-11-16
`pidfd_getfd`, `pid_open` and `pidfd_send_signal`

JonathanWoollett-Light:
Implemented [`pidfd_getfd`](https://man7.org/linux/man-pages/man2/pidfd_getfd.2.html), [`pid_open`](https://man7.org/linux/man-pages/man2/pidfd_open.2.html) and [`pidfd_send_signal`](https://man7.org/linux/man-pages/man2/pidfd_send_signal.2.html)

PR #1923, opened 2022-12-07
feat: I/O safety for 'sys/wait'

SteveLauC:
#### What this PR does:
1. Adds I/O safety for `sys/wait`

----------

Actually, I am not sure about which type to use here:
```rust
pub enum Id<'fd> {
    /// Wait for the child referred to by the given PID file descriptor
    PIDFd(RawFd),
    PIDFd(BorrowedFd<'fd>),
}
```

If we use `Fd: AsFd`

```rust
pub enum Id<'fd, Fd: AsFd> {
    /// Wait for the child referred to by the given PID file descriptor
    PIDFd(RawFd),
    PIDFd(&'fd Fd),
}
```

 then the user has to specify that generic type when using this interface, which is kinda user-unfriendly...

------

The typical usage of this interface will be something like:

```rust
// Thought currently we don't have pidfd_open(2) in `Nix`
let fd_referring_to_a_process: OwnedFd  = pidfd_open().unwrap();
let status = waitid(Id::PIDFd(fd_referring_to_a_process), WaitPidFlag::XXXX).unwrap();
```

UPDATE: `pidfd_open(2)` will be added in #1859 or #1868 .

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -10,7 +10,7 @@ use std::convert::TryFrom;
     target_os = "android",
     all(target_os = "linux", not(target_env = "uclibc")),
 ))]
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsRawFd, BorrowedFd};
 
 libc_bitflags!(
     /// Controls the behavior of [`waitpid`].
@@ -343,8 +343,8 @@ pub fn wait() -> Result<WaitStatus> {
     target_os = "haiku",
     all(target_os = "linux", not(target_env = "uclibc")),
 ))]
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub enum Id {
+#[derive(Debug)]
+pub enum Id<'fd> {
     /// Wait for any child
     All,
     /// Wait for the child whose process ID matches the given PID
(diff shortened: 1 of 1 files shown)


SteveLauC:
The reason why CI failed is that `enum Id<'fd>` is available on:
1. Linux (glibc/musl)
2. Android
3. FreeBSD
4. Haiku

And that `'fd` parameter is used only on `Android` and `Linux(glibc/musl)` as the `cfg` here:

```rust
    /// Wait for the child referred to by the given PID file descriptor
    PIDFd(BorrowedFd<'fd>),
```

So `'fd` becomes `unused` on FreeBSD and haiku, and Rust does not like this.

Honestly, I don't have a good solution to this problem, perhaps then `BorrowedFd<'_>` is not an appropriate option for this PR...

-----
Similar Problem: [Type parameter not used on enums](https://internals.rust-lang.org/t/type-parameter-not-used-on-enums/13342)

asomers:
I think the easiest way to fix the tests on non-Linux is to add a `PhantomData<'fd>` field.  For non-Linux systems, the lifetime will always end up being `'static`, so it shouldn't be much of an inconvenience to the users.

src/sys/wait.rs
```diff
@@ -355,7 +355,10 @@ pub enum Id {
     PGid(Pid),
     /// Wait for the child referred to by the given PID file descriptor
     #[cfg(any(target_os = "android", target_os = "linux"))]
-    PIDFd(RawFd),
+    PIDFd(BorrowedFd<'fd>),
+    /// A helper variant to resolve the unused parameter (`'fd`) problem on platforms
+    /// other than Linux and Android.
+    _Unreachable(std::marker::PhantomData<&'fd std::convert::Infallible>),
```
SteveLauC, line 361: Put a [`std::convert::Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) in it to make sure this variant could NEVER be constructed.

SteveLauC: Will mark this variant with a #[doc(hidden)].

asomers (approved):
bors r+

PR #1925, opened 2022-12-08
fix: send ETH_P_ALL in htons format

tzneal:


tzneal:
This was required in my testing on Linux at least.   I *think* the other values need the same treatment, but didn't want to change them as I didn't test on the other platforms.

src/sys/socket/mod.rs
```diff
@@ -217,7 +224,7 @@ pub enum SockProtocol {
     // The protocol number is fed into the socket syscall in network byte order.
     #[cfg(any(target_os = "android", target_os = "linux"))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
-    EthAll = libc::ETH_P_ALL.to_be(),
+    EthAll = htons_i32(libc::ETH_P_ALL),
```
asomers, line 227: Would this work instead?  It's a little bit simpler to inline it.
```suggestion
    EthAll = (libc::ETH_P_ALL as u16).to_be() as i32,
```

tzneal: I think it works the same, but it's a `const fn` so it's evaluated at compile time and won't require inlining, or did you just mean it's simpler to not have the function?

Happy to do it either way, but I *think* the other values need the same treatment on Linux, I just hadn't tested that far.

asomers: Oh!  It would really help if you could test them too.  I suggested inlining that function just for readability's sake.

collinoc:
Any status updates on this? Attaching [this relevant Rust forums post](https://users.rust-lang.org/t/nix-unexpected-behavior-from-af-packet-socket/89761) that looks into some of the issues occurring from this.

tzneal:
From what I can tell, just the other `ETH_` constants need this handling.  The only one in nix at the moment is `ETH_P_ALL`, so I think it's good.

asomers:
@tzneal are you going to inline that function like I suggested?  And this needs a CHANGELOG, too.

tzneal:
Yeah I can, I was going to leave it for the next `ETH_` case, but inlining is fine with me.

asomers (approved):
bors r+

PR #1927, opened 2022-12-08
feat: I/O safety for 'sys/statvfs'

SteveLauC:
#### What this PR does:
1. Adds I/O safety for module `sys/statvfs`.


src/sys/statvfs.rs
```diff
@@ -146,11 +146,11 @@ pub fn statvfs<P: ?Sized + NixPath>(path: &P) -> Result<Statvfs> {
 }
 
 /// Return a `Statvfs` object with information about `fd`
-pub fn fstatvfs<T: AsRawFd>(fd: &T) -> Result<Statvfs> {
+pub fn fstatvfs<Fd: AsFd>(fd: &Fd) -> Result<Statvfs> {
```
asomers, line 149: We should probably be consistent in our use of `AsFd` vs `&AsFd` arguments.  Other recent PRs are mostly going the former way.  Want to argue in favor of `&AsFd`, @SteveLauC ?  cc @SUPERCILEX .

SteveLauC: Yes, they should be consistent. Let's talk about it [here](https://github.com/nix-rust/nix/issues/1750#issuecomment-1343124228) instead of scattering it across these PRs

asomers (approved):
bors r+

PR #1928, opened 2022-12-08
feat: I/O safety for 'sys/memfd' & 'sys/event' & 'sys/eventfd'

SteveLauC:
#### What this PR does:

Adds I/O safety for moduels:
1. `sys/memfd`
2. `sys/event`
3. `sys/eventfd`

-----

BYW, I called `rustfmt` on these 4 files, which introduces some noise, sorry about this.

### DIFF
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -83,7 +83,7 @@ impl EpollEvent {
 /// let epoll = Epoll::new(EpollCreateFlags::empty())?;
 ///
 /// // Create eventfd & Add event
-/// let eventfd = unsafe { OwnedFd::from_raw_fd(eventfd(0, EfdFlags::empty())?) };
+/// let eventfd = eventfd(0, EfdFlags::empty())?;
 /// epoll.add(&eventfd, EpollEvent::new(EpollFlags::EPOLLIN,DATA))?;
 ///
 /// // Arm eventfd & Time wait
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -8,7 +8,7 @@ use libc::{c_int, c_long, intptr_t, time_t, timespec, uintptr_t};
 use libc::{c_long, intptr_t, size_t, time_t, timespec, uintptr_t};
 use std::convert::TryInto;
 use std::mem;
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, OwnedFd};
 use std::ptr;
 
 // Redefine kevent in terms of programmer-friendly enums and bitfields.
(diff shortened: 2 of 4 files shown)


src/sys/event.rs
```diff
@@ -267,8 +267,8 @@ impl KEvent {
     }
 }
 
-pub fn kevent(
-    kq: RawFd,
+pub fn kevent<Fd: AsFd>(
+    kq: &Fd,
```
JonathanWoollett-Light, line 271: It is better to pass `Fd` as not a reference, consider [`impl<T: AsFd> AsFd for &T`](https://doc.rust-lang.org/nightly/std/os/fd/trait.AsFd.html#impl-AsFd-for-%26T).

SteveLauC:
```rust
use std::{fs::File, os::unix::io::AsFd};

fn syscall<Fd: AsFd>(_fd: Fd) {}

fn main() {
    let f1: File = File::open(".").unwrap();
    let f2: File = f1.try_clone().unwrap();

    syscall(f1);  // We can pass T: AsFd
    syscall(&f2); // and ref to T
}
```
I get it, though all my PRs use `Fd: AsFd(fd: &Fd)` in the current implementations...

cc @asomers, any thoughts on this, which type should we use?

SUPERCILEX: Agreed with @JonathanWoollett-Light, see my comment here: https://github.com/nix-rust/nix/issues/1750#issuecomment-1343124228

src/sys/epoll.rs
```diff
@@ -78,22 +78,22 @@ impl EpollEvent {
...
 /// const MILLIS: u64 = 100;
-/// 
+///
 /// // Create epoll
 /// let epoll = Epoll::new(EpollCreateFlags::empty())?;
-/// 
+///
 /// // Create eventfd & Add event
-/// let eventfd = unsafe { OwnedFd::from_raw_fd(eventfd(0, EfdFlags::empty())?) };
+/// let eventfd = unsafe { eventfd(0, EfdFlags::empty())? };
```
asomers, line 86: Why does this need an `unsafe` block?

asomers (changes requested):
Could you please split the formatting changes out into a separate commit?

SUPERCILEX:
Made the formatting changes here: https://github.com/nix-rust/nix/pull/1935. CI checks are clearly still broken, I'll try and figure out why if I have time.

src/sys/epoll.rs
SteveLauC: Missed that, sorry about this.

SteveLauC:
> Could you please split the formatting changes out into a separate commit?

Let't merge #1935 first, then I will rebase my branch.

asomers:
I think you need to rebase to make the formatting changes go away.  And the OSX aarch64 failure is spurious.

SteveLauC:
> I think you need to rebase to make the formatting changes go away. And the OSX aarch64 failure is spurious.

Rebased.

asomers (approved):
bors r+

PR #1936, opened 2022-12-10
feat: I/O safety for 'sys/signal'

SteveLauC:
## What does this PR do

1. Adds I/O safety for `sys/signal`.

## Checklist:

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -1,6 +1,3 @@
-// Portions of this file are Copyright 2014 The Rust Project Developers.
-// See https://www.rust-lang.org/policies/licenses.
-
 //! Operating system signals.
 
 use crate::errno::Errno;
@@ -10,8 +7,6 @@ use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::mem;
 use std::ops::BitOr;
-#[cfg(freebsdlike)]
-use std::os::unix::io::RawFd;
 use std::ptr;
 use std::str::FromStr;
 
@@ -1114,8 +1109,8 @@ pub type type_of_thread_id = libc::pid_t;
 // as a pointer, because neither libc nor the kernel ever dereference it.  nix
 // therefore presents it as an intptr_t, which is how kevent uses it.
 #[cfg(not(any(target_os = "fuchsia", target_os = "hurd", target_os = "openbsd", target_os = "redox")))]
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub enum SigevNotify {
+#[derive(Clone, Copy, Debug)]
+pub enum SigevNotify<'fd> {
     /// No notification will be delivered
     SigevNone,
     /// Notify by delivering a signal to the process.
@@ -1131,7 +1126,7 @@ pub enum SigevNotify {
     #[cfg(freebsdlike)]
     SigevKevent {
         /// File descriptor of the kqueue to notify.
-        kq: RawFd,
+        kq: std::os::fd::BorrowedFd<'fd>,
         /// Will be contained in the kevent's `udata` field.
         udata: libc::intptr_t
     },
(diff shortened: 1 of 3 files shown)


src/sys/signal.rs
```diff
@@ -1015,6 +1015,9 @@ pub enum SigevNotify {
         /// structure of the queued signal.
         si_value: libc::intptr_t
     },
+    /// A helper variant to resolve the unused parameter (`'fd`) problem on platforms
+    /// other than FreeBSD and DragonFly.
```
asomers, line 1019:
```suggestion
    /// A helper variant to resolve the unused parameter (`'fd`) problem on platforms
    /// other than FreeBSD and DragonFly.
    #[doc(hidden)]
```

asomers (approved):
Apart from the doc nit, this LGTM.  However, it may cause borrow checker issues upstack.  The change works for me in mio-aio, but it could cause difficulty if Tokio adopts I/O Safety.  I think we should probably merge it, but work with Tokio to ensure that all of the pieces will fit together before we release.

SteveLauC:
macOS CI failure: network issue

```
    Updating crates.io index
warning: spurious network error (2 tries remaining): SecureTransport error: (null); class=Net (12)
warning: spurious network error (1 tries remaining): SecureTransport error: (null); class=Net (12)
error: failed to get `bitflags` as a dependency of package `nix v0.26.1 (/private/var/folders/hy/gdmd645x1ws410vx266t7py00000gn/T/cirrus-ci-build)`

Caused by:
  failed to load source for dependency `bitflags`

Caused by:
  Unable to update registry `crates-io`

Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  SecureTransport error: (null); class=Net (12)

Exit status: 101
```

SteveLauC:
>  but it could cause difficulty if Tokio adopts I/O Safety.

Hi @asomers, I think we can merge this now, looks like Tokio won't adopt I/O safety.

src/sys/signal.rs
```diff
@@ -1161,6 +1156,14 @@ pub enum SigevNotify {
...
         si_value: libc::intptr_t
     },
+    /// A helper variant to resolve the unused parameter (`'fd`) problem on
+    /// platforms other than FreeBSD and DragonFlyBSD.
+    ///
+    /// This variant can never be constructed due to the usage of an enum with 0
+    /// variants.
+    #[doc(hidden)]
+    #[cfg(not(freebsdlike))]
+    _Unreachable(&'fd std::convert::Infallible),
```
SteveLauC, line 1166: I changed the associated type of this variant to `&'fd Infallible`, since `Infallible` is an enum with 0 variants, this variant can now never be constructed. (As a contrast, `PhantomData` is very easy to create)

src/sys/signal.rs
```diff
@@ -1,6 +1,3 @@
-// Portions of this file are Copyright 2014 The Rust Project Developers.
```
SteveLauC, line 1: Remove this header, looks like it is the only header we have.

asomers (approved):
I just checked again.  There's no way to make this work with Tokio except by using `BorrowedFd::borrow_raw`, which defeats the whole point of I/O safety.  But we can merge it if there are other users who can benefit.

SteveLauC:
>  There's no way to make this work with Tokio except by using `BorrowedFd::borrow_raw`

Ah, right, since Tokio does not adopt I/O safety (in its implementation), all its internal interfaces are using RawFd, so it has to do the tedious conversion when interacting with I/O-safe interfaces.

PR #1937, opened 2022-12-10
feat: I/O safety for 'poll'

SteveLauC:
#### What this PR does:
1. Adds I/O safety for module `poll`.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,5 +1,5 @@
 //! Wait for events to trigger on specific file descriptors
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
 
 use crate::errno::Errno;
 use crate::Result;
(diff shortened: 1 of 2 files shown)


src/poll.rs
```diff
@@ -21,10 +21,24 @@ pub struct PollFd {
...
     /// Creates a new `PollFd` specifying the events of interest
     /// for a given file descriptor.
-    pub const fn new(fd: RawFd, events: PollFlags) -> PollFd {
+    //
+    // Different from other I/O-safe interfaces, here, we have to take `AsFd`
+    // by reference to prevent the case where the `fd` is closed but it is
+    // still in use. For example:
+    //
+    // ```rust
+    // let (reader, _) = pipe().unwrap();
```
asomers, line 30: Taking it by reference isn't sure proof against the file descriptor being closed prematurely.  Maybe we should add a `_fd: PhantomData<BorrowedFd<'a>>` to the struct?

src/poll.rs
```diff
@@ -68,9 +82,17 @@ impl PollFd {
...
-        self.pollfd.fd
+impl AsFd for PollFd {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        // BorrowedFd::borrow_raw() requires that the raw fd being passed must
+        // remain open for the duration of the returned BorrowedFd, this is
+        // guaranteed as the returned BorrowedFd has lifetime parameter same
+        // as `self`:
+        // "fn as_fd<'self>(&'self self) -> BorrowedFd<'self>"
+        // which means that `self` (PollFd) is guaranteed to outlive the returned
+        // BorrowedFd. Since `self` is valid, the fd within `self` is valid,
```
asomers, line 93: Not quite.  As written there's no way we can guarantee that self is valid.  For example
```rust
let f = tempfile();
let p = PollFd::new(&f, ...);
drop(f);
let _ = p.as_fd();    // Invalid fd!
```

asomers (changes requested):

src/poll.rs
SteveLauC: What about defining struct pollfd in Rust instead of using the libc one, just like [what rustix did](https://docs.rs/rustix/latest/rustix/io/struct.PollFd.html#).

With this, we can also resolve the next problem

asomers: I'd rather not.  It's too easy to screw it up , especially if the structure is OS-dependent.  I assume you're suggesting using a BorrowedFd in the struct instead of a `c_int`?  Would that work any better than using PhantomData?

SteveLauC: > It's too easy to screw it up , especially if the structure is OS-dependent. 

This makes sense.

Adding a phantom field is a better approach.

asomers (approved):
bors r+

PR #1940, opened 2022-12-11
feat: I/O safety for 'sys/select'

SteveLauC:
#### What this PR does:
1. Adds I/O safety for module `sys/select`.

### DIFF
--- a/src/sys/select.rs
+++ b/src/sys/select.rs
@@ -7,7 +7,7 @@ use std::convert::TryFrom;
 use std::iter::FusedIterator;
 use std::mem;
 use std::ops::Range;
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd};
 use std::ptr::{null, null_mut};
 
 pub use libc::FD_SETSIZE;
@@ -15,7 +15,10 @@ pub use libc::FD_SETSIZE;
 /// Contains a set of file descriptors used by [`select`]
 #[repr(transparent)]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub struct FdSet(libc::fd_set);
+pub struct FdSet<'fd> {
+    set: libc::fd_set,
+    _fd: std::marker::PhantomData<BorrowedFd<'fd>>,
+}
 
 fn assert_fd_valid(fd: RawFd) {
     assert!(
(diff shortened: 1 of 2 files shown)


src/sys/select.rs
```diff
@@ -24,37 +27,40 @@ fn assert_fd_valid(fd: RawFd) {
...
+                _fd: std::marker::PhantomData,
+            }
         }
     }
 
     /// Add a file descriptor to an `FdSet`
-    pub fn insert(&mut self, fd: RawFd) {
-        assert_fd_valid(fd);
-        unsafe { libc::FD_SET(fd, &mut self.0) };
+    pub fn insert<Fd: AsFd>(&mut self, fd: &'fd Fd) {
```
SteveLauC, line 44: For `FdSet::insert()` and `FdSet::contains()`, taking  `AsFd` by references to make them safer.

src/sys/select.rs
```diff
@@ -24,37 +27,40 @@ fn assert_fd_valid(fd: RawFd) {
...
+        assert_fd_valid(fd.as_fd().as_raw_fd());
+        unsafe { libc::FD_SET(fd.as_fd().as_raw_fd(), &mut self.set) };
     }
 
     /// Remove a file descriptor from an `FdSet`
-    pub fn remove(&mut self, fd: RawFd) {
-        assert_fd_valid(fd);
-        unsafe { libc::FD_CLR(fd, &mut self.0) };
+    pub fn remove<Fd: AsFd>(&mut self, fd: Fd) {
```
SteveLauC, line 50: For `FdSet::remove()`, IMO, it is fine to "consume" that `fd` as it will no long be monitored.

src/sys/select.rs
```diff
@@ -88,11 +97,13 @@ impl FdSet {
...
-    /// set.insert(9);
-    /// let fds: Vec<RawFd> = set.fds(None).collect();
+    /// let fd_four = unsafe {BorrowedFd::borrow_raw(4)};
+    /// let fd_nine = unsafe {BorrowedFd::borrow_raw(9)};
+    /// set.insert(&fd_four);
+    /// set.insert(&fd_nine);
+    /// let fds: Vec<RawFd> = set.fds(None).map(|borrowed_fd|borrowed_fd.as_raw_fd()).collect();
     /// assert_eq!(fds, vec![4, 9]);
     /// ```
     #[inline]
```
SteveLauC, line 109: The `highest` parameter of this `fds()` function, IMO, using a `RawFd`  is better than using a `AsFd` as users may just wanna specify a specific value, and this value (fd) probably is not open. For example, I just wanna construct an iterator whose file decriptors are smaller than 50:

```rust
let set = FdSet::new();
// some insertions..
let fds_smaller_than_fifty = set.fds(Some(50));
```
I don't want to create a `BorrowedFd` first, which is obviously too tedious:

```rust
let fd_fifty = unsafe { BorrowedFd::borrow_raw(50)) };
let fds = set.fds(Some(&fd_fifty));
```

src/sys/select.rs
asomers: I disagree.  It's conceivable that somebody no longer wishes to poll a certain file descriptor with select, but still wishes to have it open.  That's harder to do in this case.  Plus, for consistency's sake, I think remove should take by reference just like the other methods.

SteveLauC: Done.

asomers (approved):
bors r+

PR #1942, opened 2022-12-12
refactor: a new FileStat type

SteveLauC:


src/sys/stat.rs
```diff
@@ -66,13 +65,135 @@ libc_bitflags! {
 }
 
 #[cfg(any(target_os = "macos", target_os = "ios", target_os = "openbsd"))]
-pub type type_of_file_flag = c_uint;
+pub type type_of_st_flag = c_uint;
```
SteveLauC, line 68: I renamed `type_of_file_flag` to `type_of_st_flag` to be consistent with other abstraction layer types.

src/sys/stat.rs
```diff
@@ -236,6 +357,134 @@ pub fn umask(mode: Mode) -> Mode {
...
+    pub fn blocks(&self) -> type_of_st_blocks {
+        self.0.st_blocks
+    }
+
+    /// Returns the time of last access
+    pub fn atime(&self) -> TimeSpec {
+        #[cfg(target_os = "netbsd")]
+        return TimeSpec::new(self.0.st_atime, self.0.st_atimensec);
+        #[cfg(not(target_os = "netbsd"))]
+        return TimeSpec::new(self.0.st_atime, self.0.st_atime_nsec);
```
SteveLauC, line 427: NetBSD has field `st_xtimensec` rather than `st_xtime_nsec`, so we have to use `#cfg()` handle it.

src/sys/stat.rs
```diff
@@ -236,6 +357,134 @@ pub fn umask(mode: Mode) -> Mode {
...
+        target_os = "openbsd",
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "freebsd",
+        target_os = "dragonfly",
+        target_os = "netbsd",
+    ))]
+    /// Return the file flags
+    pub fn flags(&self) -> FileFlag {
+        FileFlag::from_bits(self.0.st_flags as _).unwrap()
```
SteveLauC, line 456: This `as _` cast is necessary as on `FreeBSD/DragonFlyBSD/NetBSD`, `st_flags` and `type_of_st_flag` (`type_of_file_flag`) have different types.

src/sys/stat.rs
SteveLauC: Seems like I should rename it to `type_of_st_flags` to match the field name `st_flags` though...

PR #1943, opened 2022-12-13
Rustier kqueue API

asomers:
* Prefer methods instead of functions.
* Create a newtype for a kqueue.
* Document everything.
* Deprecate EVFILT_SENDFILE, because it was never fully implemented upstream.
* Add support to the libc_enum! macro to be able to deprecate variants.

### DIFF
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -1,5 +1,7 @@
-/* TOOD: Implement for other kqueue based systems
- */
+//! Kernel event notification mechanism
+//!
+//! # See Also
+//! [kqueue(2)](https://www.freebsd.org/cgi/man.cgi?query=kqueue)
 
 use crate::{Errno, Result};
 #[cfg(not(target_os = "netbsd"))]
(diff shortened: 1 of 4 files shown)


src/sys/event.rs
```diff
@@ -37,43 +97,66 @@ libc_enum! {
...
+        /// Establishes a file system monitor.
         EVFILT_FS,
         #[cfg(target_os = "freebsd")]
+        /// Notify for completion of a list of POSIX AIO operations.
+        /// # See Also
+        /// [lio_listio(2)](https://www.freebsd.org/cgi/man.cgi?query=lio_listio)
         EVFILT_LIO,
         #[cfg(any(target_os = "ios", target_os = "macos"))]
+        /// Mach portsets
         EVFILT_MACHPORT,
```
notgull, line 125: This part may be unsound; I'm not sure about mach port support here but it's definitely not just an FD.

src/sys/event.rs
```diff
@@ -8,16 +10,74 @@ use libc::{c_int, c_long, intptr_t, time_t, timespec, uintptr_t};
...
+/// Used by the kernel to notify the process of various types of asynchronous
+/// events.
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct Kqueue(OwnedFd);
+
+impl Kqueue {
+    /// Create a new kernel event queue.
+    pub fn new() -> Result<Self> {
+        let res = unsafe { libc::kqueue() };
```
notgull, line 35: Is there a plan to support the `kqueue1()` function exposed by FreeBSD?

asomers: Do you mean NetBSD?  No.  I don't know of anybody who needs it, and we don't have a dedicated NetBSD maintainer.

src/sys/event.rs
asomers: What might not just be an FD?  And what part might be unsound?  This just defines the filter constant.

src/sys/event.rs
```diff
@@ -8,16 +10,74 @@ use libc::{c_int, c_long, intptr_t, time_t, timespec, uintptr_t};
 use libc::{c_long, intptr_t, size_t, time_t, timespec, uintptr_t};
 use std::convert::TryInto;
 use std::mem;
-use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, OwnedFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, OwnedFd};
 use std::ptr;
 
-// Redefine kevent in terms of programmer-friendly enums and bitfields.
+/// A Kernel event queue.  Used to notify a process of various asynchronous
```
rtzoeller, line 16: Nit.

```suggestion
/// A kernel event queue.  Used to notify a process of various asynchronous
```

src/sys/event.rs
```diff
@@ -293,30 +454,19 @@ type type_of_nchanges = c_int;
...
-            },
-        )
-    };
-
-    Errno::result(res).map(|r| r as usize)
+    kq.kevent(changelist, eventlist, timeout_opt)
 }
 
+/// Modify an existing [`KEvent`].
+// Probably should deprecate.  Would anybody ever use it over `KEvent::new`?
```
rtzoeller, line 469: I think deprecating it makes sense. If there is a use case I expect we'll hear about it after marking it as deprecated - we can always remove the deprecation label and keep it around if a use case does come up.

src/sys/event.rs
```diff
@@ -86,131 +169,194 @@ libc_enum! {
...
         #[cfg(any(target_os = "macos", target_os = "ios"))]
+        #[allow(missing_docs)]
         NOTE_VM_PRESSURE;
         #[cfg(any(target_os = "macos", target_os = "ios"))]
+        #[allow(missing_docs)]
         NOTE_VM_PRESSURE_SUDDEN_TERMINATE;
         #[cfg(any(target_os = "macos", target_os = "ios"))]
+        #[allow(missing_docs)]
         NOTE_VM_PRESSURE_TERMINATE;
+        #[allow(missing_docs)]
```
rtzoeller, line 351: If we find ourselves sprinkling `#[allow(missing_docs)]` over every attribute like this in other structs, we might consider modifying  the `libc_bitflags` macro to make it less tedious.

rtzoeller (approved):
Approved with some minor feedback. Don't forget to update the CHANGELOG PR #.

asomers:
bors r=rtzoeller

PR #1944, opened 2022-12-13
Rework vsock test

### LINKED ISSUE #1934: Multiple failures in test_vsock
When we first added the vsock code in #1091 the test contained an assertion that connecting to the socket would fail, because "The current implementation does not support loopback devices".  But it did not check the particular failure mode.  A recent change (#1915) has caused `connect` to start succeeding, in at least some cases.  It seems that at least some of the original failures were due to a race condition in the test: it closed a socket too early.  Now that that is fixed, we see three behaviors:

* Fails with EINVAL on Linux x86_64, presumably due to the "does not support loopback devices" part.
* Passes on most qemu platforms.  Perhaps it's connecting to the qemu host?
* Fails with ENODEV on mips and mips64

See #1917 to view the behavior on each platform.

@stefano-garzarella could you please clarify what we should expect from `connect`, and fix the test?  For now I'm going to disable it so it doesn't hold up other PRs.

stefano-garzarella:
We mainly provide VsockAddr, so let's try to test well that VsockAddr mapping to libc::sockaddr_vm is correct.

Let's remove all interactions with the socket, since vsock may or may not be available in the environment.
Testing socket(), bind(), listen(), connect(), etc. caused unexpected failures, and it's out of scope of this crate.

So let's simplify the vsock test focussing on VsockAddr. This should work also on graviton, so let's try to re-enable it.

Fixes #1934

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -2040,57 +2040,43 @@ pub fn test_recv_ipv6pktinfo() {
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
-#[cfg_attr(graviton, ignore = "Not supported by the CI environment")]
 #[test]
 pub fn test_vsock() {
-    use nix::errno::Errno;
-    use nix::sys::socket::{
-        bind, connect, listen, socket, AddressFamily, SockFlag, SockType,
-        VsockAddr,
-    };
-    use nix::unistd::close;
-    use std::thread;
+    use nix::sys::socket::SockaddrLike;
+    use nix::sys::socket::{AddressFamily, VsockAddr};
+    use std::mem;
 
     let port: u32 = 3000;
 
-    let s1 = socket(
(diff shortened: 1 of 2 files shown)


test/sys/test_socket.rs
```diff
@@ -2040,57 +2040,43 @@ pub fn test_recv_ipv6pktinfo() {
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
-#[cfg_attr(graviton, ignore = "Not supported by the CI environment")]
```
asomers, line 2043: If you're going to remove this line, then you should also remove the `RUSTFLAGS: --cfg graviton -D warnings` line in .cirrus.yml .

asomers (changes requested):

asomers (approved):
bors r+

PR #1945, opened 2022-12-18
EventFd type

JonathanWoollett-Light:
Added an `EventFd` type that makes usage a little more convenient and explicit.

### DIFF
--- a/src/sys/eventfd.rs
+++ b/src/sys/eventfd.rs
@@ -1,6 +1,6 @@
 use crate::errno::Errno;
-use crate::Result;
-use std::os::unix::io::{FromRawFd, OwnedFd};
+use crate::{Result,unistd};
+use std::os::unix::io::{FromRawFd, OwnedFd, AsRawFd, AsFd, RawFd, BorrowedFd};
 
 libc_bitflags! {
     pub struct EfdFlags: libc::c_int {
@@ -10,8 +10,75 @@ libc_bitflags! {
     }
 }
 
+#[deprecated(since = "0.28.0", note = "Use EventFd::from_value_and_flags() instead")]
 pub fn eventfd(initval: libc::c_uint, flags: EfdFlags) -> Result<OwnedFd> {
     let res = unsafe { libc::eventfd(initval, flags.bits()) };
 
     Errno::result(res).map(|r| unsafe { OwnedFd::from_raw_fd(r) })
+}
+
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct EventFd(OwnedFd);
+impl EventFd {
+    /// [`EventFd::from_value_and_flags`] with `init_val = 0` and `flags = EfdFlags::empty()`.
+    pub fn new() -> Result<Self> {
+        Self::from_value_and_flags(0, EfdFlags::empty())
+    }
+    /// Constructs [`EventFd`] with the given `init_val` and `flags`.
+    /// 
+    /// Wrapper around [`libc::eventfd`].
+    pub fn from_value_and_flags(init_val: u32, flags: EfdFlags) -> Result<Self> {
+        let res = unsafe { libc::eventfd(init_val, flags.bits()) };
+        Errno::result(res).map(|r| Self(unsafe { OwnedFd::from_raw_fd(r) }))
+    }
+    /// [`EventFd::from_value_and_flags`] with `init_val = 0` and given `flags`.
+    pub fn from_flags(flags: EfdFlags) -> Result<Self> {
+        Self::from_value_and_flags(0, flags)
+    }
+    /// [`EventFd::from_value_and_flags`] with given `init_val` and `flags = EfdFlags::empty()`.
+    pub fn from_value(init_val: u32) -> Result<Self> {
+        Self::from_value_and_flags(init_val, EfdFlags::empty())
+    }
+    /// Arms `self`, a following call to `poll`, `select` or `epoll` will return immediately.
+    /// 
+    /// [`EventFd::write`] with `1`.
+    pub fn arm(&self) -> Result<usize> {
+        self.write(1)
+    }
+    /// Defuses `self`, a following call to `poll`, `select` or `epoll` will block.
+    /// 
+    /// [`EventFd::write`] with `0`.
+    pub fn defuse(&self) -> Result<usize> {
+        self.write(0)
+    }
+    /// Enqueues `value` triggers.
+    /// 
+    /// The next `value` calls to `poll`, `select` or `epoll` will return immediately.
+    /// 
+    /// [`EventFd::write`] with `value`.
+    pub fn write(&self, value: u64) -> Result<usize> { 
+        unistd::write(&self.0,&value.to_ne_bytes())
+    }
+    // Reads the value from the file descriptor.
+    pub fn read(&self) -> Result<u64> {
+        let mut arr = [0; std::mem::size_of::<u64>()];
+        unistd::read(self.0.as_raw_fd(),&mut arr)?;
+        Ok(u64::from_ne_bytes(arr))
+    }
+}
+impl AsFd for EventFd {
+    fn as_fd(&self) -> BorrowedFd {
+        self.0.as_fd()
+    }
+}
+impl AsRawFd for EventFd {
+    fn as_raw_fd(&self) -> RawFd {
+        self.0.as_raw_fd()
+    }
+}
+impl From<EventFd> for OwnedFd {
+    fn from(x: EventFd) -> OwnedFd {
+        x.0
+    }
 }
--- a/changelog/1945.added.md
+++ b/changelog/1945.added.md
@@ -1,0 +1,1 @@
+Added `EventFd` type.
--- a/changelog/1945.changed.md
+++ b/changelog/1945.changed.md
@@ -1,0 +1,1 @@
+Deprecated `sys::eventfd::eventfd`.
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -72,7 +72,7 @@ impl EpollEvent {
 
 /// A safe wrapper around [`epoll`](https://man7.org/linux/man-pages/man7/epoll.7.html).
 /// ```
-/// # use nix::sys::{epoll::{EpollTimeout, Epoll, EpollEvent, EpollFlags, EpollCreateFlags}, eventfd::{eventfd, EfdFlags}};
+/// # use nix::sys::{epoll::{EpollTimeout, Epoll, EpollEvent, EpollFlags, EpollCreateFlags}, eventfd::{EventFd, EfdFlags}};
 /// # use nix::unistd::write;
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   (#[1662](https://github.com/nix-rust/nix/pull/1662))
 - Added `CanRaw` to `SockProtocol` and `CanBcm` as a separate `SocProtocol` constant.
   ([#1912](https://github.com/nix-rust/nix/pull/1912))
+  - Added `EventFd` type.
+  ([#1912](https://github.com/nix-rust/nix/pull/1912))
```
rtzoeller, line 15: PR number needs to be updated.

JonathanWoollett-Light: Fixed.

src/sys/eventfd.rs
```diff
@@ -10,8 +10,58 @@ libc_bitflags! {
...
+        let res = unsafe { libc::eventfd(0, flags.bits()) };
+        Errno::result(res).map(|r| Self(unsafe { OwnedFd::from_raw_fd(r) }))
+    }
+    /// [`EventFd::value_and_flags`] with given `init_val` and `flags = EfdFlags::empty()`.
+    pub fn value(init_val: u32) -> Result<Self> {
+        let res = unsafe { libc::eventfd(init_val, EfdFlags::empty().bits()) };
+        Errno::result(res).map(|r| Self(unsafe { OwnedFd::from_raw_fd(r) }))
+    }
+    /// [`EventFd::write`] with `1`.
+    pub fn arm(&self) -> Result<usize> {
```
asomers, line 46: What is the meaning of the return value, here and for defuse?

src/sys/eventfd.rs
```diff
@@ -10,8 +10,58 @@ libc_bitflags! {
...
+impl EventFd {
+    /// [`EventFd::value_and_flags`] with `init_val = 0` and `flags = EfdFlags::empty()`.
+    pub fn new() -> Result<Self> {
+        let res = unsafe { libc::eventfd(0, EfdFlags::empty().bits()) };
+        Errno::result(res).map(|r| Self(unsafe { OwnedFd::from_raw_fd(r) }))
+    }
+    /// Constructs [`EventFd`] with the given `init_val` and `flags`.
+    /// 
+    /// Wrapper around [`libc::eventfd`].
+    pub fn value_and_flags(init_val: u32, flags: EfdFlags) -> Result<Self> {
```
asomers, line 31: It would be more idiomatic to name these constructors from_XXX

src/sys/eventfd.rs
```diff
@@ -10,8 +10,58 @@ libc_bitflags! {
...
+    /// [`EventFd::write`] with `1`.
+    pub fn arm(&self) -> Result<usize> {
+        unistd::write(self.0.as_raw_fd(),&1u64.to_ne_bytes())
+    }
+    /// [`EventFd::write`] with `0`.
+    pub fn defuse(&self) -> Result<usize> {
+        unistd::write(self.0.as_raw_fd(),&0u64.to_ne_bytes())
+    }
+    /// Writes a given `value` to the file descriptor.
+    pub fn write(&self, value: u64) -> Result<usize> {
```
asomers, line 54: What is the meaning of writing a value to the file descriptor?

JonathanWoollett-Light: This is literally what it is doing, although I would agree its not the clearest documentation although I'm not sure what to replace it with.

https://man7.org/linux/man-pages/man2/write.2.html

> write() writes up to count bytes from the buffer starting at buf to the file referred to by the file descriptor fd.

src/sys/eventfd.rs
JonathanWoollett-Light: It passes through the return value of `unistd::write`.

src/sys/eventfd.rs
JonathanWoollett-Light: Renamed with `from_` prefix.

src/sys/eventfd.rs
asomers: Yeah, but what does that mean?  What does it even mean to write to an eventfd ?

src/sys/eventfd.rs
asomers: But what is even the purpose of writing to an eventFD?  Is there any reason why anybody would ever want to do that?

JonathanWoollett-Light: When you poll on an `EventFd` you can consider that the event fires when a non-zero value can be read from the `EventFd`. By writing a value to the `EventFd` you are setting it up to trigger `value` events when polled.

If you write `2` here then poll 3 times on the `EventFd` the 1st 2 polls will return and the 3rd will wait.

src/sys/eventfd.rs
JonathanWoollett-Light: I think this is answered by https://github.com/nix-rust/nix/pull/1945#discussion_r1070627849

src/sys/eventfd.rs
asomers: Ok.  Could you please rewrite the docs then, in terms of "arming" instead of "writing".  We should strive to be as high-level as possible while remaining zero-cost.  Also, is there any point in passing through the return value of `unistd::write`?

JonathanWoollett-Light: I've updated the docs.

src/sys/eventfd.rs
```diff
@@ -10,8 +10,69 @@ libc_bitflags! {
...
+        unistd::read(self.0.as_raw_fd(),&mut arr)?;
+        Ok(u64::from_ne_bytes(arr))
+    }
+}
+impl AsFd for EventFd {
+    fn as_fd(&self) -> BorrowedFd {
+        self.0.as_fd()
+    }
+}
+impl AsRawFd for EventFd {
```
SteveLauC, line 74: Do you need to impl `AsRawFd` for it?

SteveLauC (changes requested):
Left some comments

src/sys/eventfd.rs
```diff
@@ -10,8 +10,69 @@ libc_bitflags! {
...
 
+#[deprecated(since = "0.27.0", note = "Use EventFd::from_value_and_flags() instead")]
 pub fn eventfd(initval: libc::c_uint, flags: EfdFlags) -> Result<OwnedFd> {
     let res = unsafe { libc::eventfd(initval, flags.bits()) };
 
     Errno::result(res).map(|r| unsafe { OwnedFd::from_raw_fd(r) })
 }
+
+#[derive(Debug)]
+pub struct EventFd(pub OwnedFd);
```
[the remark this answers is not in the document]
JonathanWoollett-Light: By making the field public it allows a user to deconstruct it which `AsFd` doesn't allow, so for whatever reason they want they can extract the `OwnedFd` into another type.

I've added `#[repr(transparent)]`.

src/sys/eventfd.rs
```diff
@@ -10,8 +10,69 @@ libc_bitflags! {
...
+    }
+    /// [`EventFd::from_value_and_flags`] with given `init_val` and `flags = EfdFlags::empty()`.
+    pub fn from_value(init_val: u32) -> Result<Self> {
+        Self::from_value_and_flags(init_val, EfdFlags::empty())
+    }
+    /// Arms `self`, a following call to `poll`, `select` or `epoll` will return immediately.
+    /// 
+    /// [`EventFd::write`] with `1`.
+    pub fn arm(&self) -> Result<usize> {
+        unistd::write(&self.0,&1u64.to_ne_bytes())
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I've updated it to use this approach.

src/sys/eventfd.rs
```diff
@@ -10,8 +10,69 @@ libc_bitflags! {
...
+    }
+    /// Defuses `self`, a following call to `poll`, `select` or `epoll` will block.
+    /// 
+    /// [`EventFd::write`] with `0`.
+    pub fn defuse(&self) -> Result<usize> {
+        unistd::write(&self.0,&0u64.to_ne_bytes())
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I've updated it to use this approach.

src/sys/eventfd.rs
JonathanWoollett-Light: I think so, is there any alternative approach to get the raw file descriptor you would suggest?

src/sys/eventfd.rs
```diff
@@ -10,8 +10,69 @@ libc_bitflags! {
     }
 }
 
+#[deprecated(since = "0.27.0", note = "Use EventFd::from_value_and_flags() instead")]
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I've updated it to `0.28.0`.

src/sys/eventfd.rs
SteveLauC: I was thinking that we have `AsFd`, then we can get a `BorrowedFd`, then we can extract the raw fd by calling `.as_raw_fd()` on the `BorrowedFd`, functionally, they have no difference, but this approach may discourage our user from using the raw fd

src/sys/eventfd.rs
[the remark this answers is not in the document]
SteveLauC: What about `impl From<EventFd> for OwnedFd`, this is how the std handles them I think

src/sys/eventfd.rs
JonathanWoollett-Light: I think its generally more ergonomic for users to have `AsRawFd` rather than not have it. I could see someone complaining about not having it, but I can't see someone complaining about having it.

SteveLauC: Yeah, it has no harm, let's keep it

src/sys/eventfd.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: What is the downside of inner item being public? I don't any downside.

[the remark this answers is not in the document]
SteveLauC: Still, functionally, no difference, but I would like to be consistent with the std:

```
impl From<ChildStderr> for OwnedFd
impl From<ChildStdin> for OwnedFd
impl From<ChildStdout> for OwnedFd
impl From<File> for OwnedFd
impl From<PidFd> for OwnedFd
impl From<TcpListener> for OwnedFd
```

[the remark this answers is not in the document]
JonathanWoollett-Light: Okay, I've made the change.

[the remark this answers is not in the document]
SteveLauC: Could you please make that `OwnedFd` field private given that we have `impl From<EventFd for OwnedFd` now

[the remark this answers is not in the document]
JonathanWoollett-Light: Updated to remove `pub`.

SteveLauC (approved):
Thanks!

PR #1946, opened 2022-12-19
SigSet: Fix incorrect implementation of Eq, PartialEq and Hash

### LINKED ISSUE #1749: SigSet Incorrect implementation of Eq and PartialEq
`signal::SigSet` derives `Eq` and `PartialEq`, I assume because `libc::sigset_t` does it too, but in linux, `sigfillset(3)` only initializes the first 64 bits, leaving the rest uninitialized, so two `signal::SigSets` can be different even though both have the same set of signals

germag:
`signal::SigSet` derives `Eq`, `PartialEq` and `Hash`, but on linux, `sigfillset(3)` only initializes the first 64 bits, leaving the rest uninitialized. So two `signal::SigSets` can be different even though both have the same set of signals.

closes #1749

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+- Fix `SigSet` incorrect implementation of `Eq`, `PartialEq` and `Hash`
+  ([#1946](https://github.com/nix-rust/nix/pull/1946))
+
 ### Changed
 
 - The following APIs now take an implementation of `AsFd` rather than a
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -7,6 +7,7 @@ use crate::errno::Errno;
 use crate::{Error, Result};
 use cfg_if::cfg_if;
 use std::fmt;
+use std::hash::{Hash, Hasher};
 use std::mem;
 #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
 use std::os::unix::io::RawFd;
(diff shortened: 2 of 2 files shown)


germag:
v2:
- Fix formatting issue

rtzoeller:
@germag have you done any performance analysis of the auto-derived version (on platforms other than Linux; I acknowledge the Linux impl is broken) vs the manually implemented one?

If this new implementation is measurably slower, we might want to use the auto-derived version on platforms where it is correct.

We could also implement this on Linux by reading only the initialized bits and hashing them all at once, although that is certainly a leaky abstraction.

germag:
> @germag have you done any performance analysis of the auto-derived version (on platforms other than Linux; I acknowledge the Linux impl is broken) vs the manually implemented one?
> 
> If this new implementation is measurably slower, we might want to use the auto-derived version on platforms where it is correct.

No, I haven't, `sigismember()` it's a simple function, but there will certainly be a difference in performance. However, IIRC according to posix we must treat `sigset_t` as an opaque object. Anyway, I can implement it only for linux, but I don't know if is also required in other targets.

> We could also implement this on Linux by reading only the initialized bits and hashing them all at once, although that is certainly a leaky abstraction.

I think it would be better to do that in libc. Also, the `__val[]` field inside `sigset_t` is not public.

asomers (approved):
I think we should merge this.  It's hard for Nix to know which platforms fully initialize sigset_t and which don't.  So unless they start providing `sigeqset`or something like that, we should probably just do things the hard way.

SteveLauC:
Friendly ping @germag, would you like to finish this PR? The CI failure seems to be irrelevant, rebasing your branch should resolve it :)

germag:
Rebase

> Friendly ping @germag, would you like to finish this PR? The CI failure seems to be irrelevant, rebasing your branch should resolve it :)

Sorry, I thought it was already merged

CHANGELOG.md
```diff
@@ -93,6 +93,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Made `clone` unsafe, like `fork`.
   ([#1993](https://github.com/nix-rust/nix/pull/1993))
 
+### Fixed
```
SteveLauC, line 96: The change log should be put under [\[Unreleased\]](https://github.com/nix-rust/nix/blob/master/CHANGELOG.md#unreleased---releasedate)

germag: done

SteveLauC (approved):

PR #1948, opened 2022-12-19
ci: release-please

JonathanWoollett-Light:
Adds release-please automation.


asomers:
What does this do?

JonathanWoollett-Light:
> What does this do?

Added some more information, the effects can be seen on my fork.

asomers:
`cargo-release` already takes care of editting the Changelog for us.  I'm not sure how much extra value release-please provides.

JonathanWoollett-Light:
`cargo-release` still requires every commit to manually edit the changelog does it not?

asomers:
Yes it does.  Does `release-please` generate the changelog based on commit messages instead?

JonathanWoollett-Light:
> Yes it does. Does `release-please` generate the changelog based on commit messages instead?

Yes it generates it based on the commit message. See the PR I linked above https://github.com/JonathanWoollett-Light/nix/pull/1 which shows the automatically generated PR to the change.

rtzoeller:
This might remove the requirement that maintainers be able to push directly to `master` in order to publish releases, which is appealing.

Changelog management is definitely an issue today, but I do really value having a centralized changelog file in the repo. That may not be a hard-requirement though. I'd be curious if there is a way to preview the `release-please` changelog for a release before publishing, to make sure you're happy with how it comes out - it'd be annoying to miss something just because a commit was formatted incorrectly.

JonathanWoollett-Light:
> Changelog management is definitely an issue today, but I do really value having a centralized changelog file in the repo.

The PR it generates just appends to `changelog.md` so it is still centralized.

> I'd be curious if there is a way to preview the release-please changelog for a release before publishing

You can see the changes in the PR it generates and maintains, this is only a base so you could still take this then expand upon it with manual additions.

SteveLauC:
Some thoughts on this thing:

1. release-please generates CHANGELOG from commit message, which requires that commit message should be [semantic](https://www.conventionalcommits.org/)
   
   Not everyone is a fan of this stuff, you can see this from nix's commit history (However, I am)

2. In our CHANGELOG, we have 4 types of logs:

   1. Added
   2. Changed
   3. Fixed
   4. Removed

   Their corresponding semantic commit message types would be

   | nix change log type | commit msg type |
   |----------|-----------------|
   | Added    | feat |
   |Changed   | refactor|
   |Fixed     |fix|
   |Removed   | refactor|

   You can see that both `Changed` and `Removed` would use `refactor` here, 
   which means we cannot distinguish them when generating CHANGELOGs using
   release-please.

   > This seems to be [customizable](https://github.com/googleapis/release-please/blob/main/docs/customizing.md#changelog-types), but needs some work I guess

3. The generated CHANGELOG:

   ```
   * I/O safety for 'kmod' 
     ([2f546c9](https://github.com/JonathanWoollett-Light/nix/commit/2f546c98cb1abb4463f13b084443ae4bf1597cb3)) 
   ```

   has a link to that **commit** instead of PR, honestly, I prefer PR more as I can 
   see the discussion/context of how people were working on it.
  
   > Still, should be customizable, but needs some work

3. When we switch to [`towncrier`](https://github.com/nix-rust/nix/pull/2149), the management of CHANGELOG would be much easier

4. From this comment:

   > This might remove the requirement that maintainers be able to push directly
   > to master in order to publish releases, which is appealing.

   This is indeed appealing

PR #1950, opened 2022-12-20
Mutex

JonathanWoollett-Light:
Added a mutex interface.


asomers:
What is the advantage of pthread mutexes compared to std::sync::Mutex ?

JonathanWoollett-Light:
`std::sync::Mutex` is not `repr(C)` and thus has inconsistent members ordering so can't be used between processes Rust processes (e.g. processes sharing a mutex stored in shared memory) and it can't be used for FFI.

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+/// #   num::NonZeroUsize,
+/// #   os::unix::io::OwnedFd
+/// # };
+/// # use nix::{
+/// #   sys::{pthread::{Mutex, MutexAttr}, mman::{mmap, MapFlags, ProtFlags}},
+/// #   unistd::{fork,ForkResult},
+/// # };
+/// const TIMEOUT: Duration = Duration::from_millis(500);
+/// const DELTA: Duration = Duration::from_millis(100);
+/// # fn main() -> nix::Result<()> {
```
asomers, line 175:
```suggestion
```
It's normally not necessary to put an explicit `main` function in a doc test.

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+/// // The mutex is unlocked, so `try_lock` will lock.
+/// assert_eq!(mutex.try_lock(), Ok(true));
+/// // Unlock the mutex.
+/// assert_eq!(mutex.unlock(), Ok(()));
+/// // The mutex is unlocked, so `lock` will lock and exit immediately.
+/// assert_eq!(mutex.lock(), Ok(()));
+/// // Unlock the mutex.
+/// assert_eq!(mutex.unlock(), Ok(()));
+/// 
+/// // Test across threads
```
asomers, line 190: These different examples should be in separate doc tests, not all combined together.  Or, in unit tests, if they aren't particularly useful as documentation.

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+/// let shared_memory = unsafe { mmap::<OwnedFd>(
+///     None,
+///     NonZeroUsize::new_unchecked(size_of::<Mutex>()),
+///     ProtFlags::PROT_WRITE | ProtFlags::PROT_READ,
+///     MapFlags::MAP_SHARED | MapFlags::MAP_ANONYMOUS,
+///     None,
+///     0
+/// )? };
+/// let mutex_ptr = shared_memory.cast::<Mutex>();
+/// let mutex = unsafe { &*mutex_ptr };
```
asomers, line 217: This line creates a `Mutex` with uninitialized memory, right?  I"m not sure that's ok.  Safer to initialize it via passing a raw pointer to C.  Like with a `fn init(*mut Mutex, attr: Option<MutexAttr>) -> Result<Self>` method.  Then you can trust that the memory will always be initialized before you access it from Rust code.  You'll need `repr(transparent` for that to be guaranteed to work.

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+///         assert_eq!(mutex.lock(), Ok(()));
+///         assert!(now.elapsed() > TIMEOUT && now.elapsed() < TIMEOUT + DELTA);
+///     }
+/// }
+/// 
+/// # Ok(())
+/// # }
+/// ```
+#[cfg(target_os = "linux")]
+#[derive(Debug)]
```
asomers, line 248:
```suggestion
#[derive(Debug)]
#[repr(transparent)]
```

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+            libc::PTHREAD_PRIO_INHERIT => Self::Inherit,
+            libc::PTHREAD_PRIO_PROTECT => Self::Protect,
+            _ => unreachable!()
+        }
+    }
+}
+
+/// Mutex attributes.
+#[cfg(target_os = "linux")]
+#[derive(Debug)]
```
asomers, line 78:
```suggestion
#[derive(Debug)]
#[repr(transparent)]
```

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+                Err(Errno::EBUSY) => Ok(false),
+                Err(err) => Err(err)
+            }
+            
+        }
+    }
+    /// Wraps [`libc::pthread_mutex_unlock`].
+    /// 
+    /// <https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3p.html>
+    pub fn unlock(&self) -> Result<()> {
```
asomers, line 300: According to https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3p.html , unlocking a pthread mutex from a thread which does not own it is undefined behavior.  So this API isn't sound.  Perhaps you can do better if the lock function returns a Guard which is neither `Send` nor `Sync`, and you can only unlock the mutex via the Guard.

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+            Errno::result(libc::pthread_mutexattr_setprotocol(&mut self.0,protocol as i32)).map(drop)
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+impl Default for MutexAttr {
+    fn default() -> Self {
+        let mutex_attr = Self::new().unwrap();
+        debug_assert_eq!(mutex_attr.get_shared(),Ok(true));
```
asomers, line 145: Why is this line here?  From my reading of https://linux.die.net/man/3/pthread_mutexattr_init , the opposite is true.  Mutexes are private by default.  I wouldn't trust that to be the same across all OSes, though.  And why use debug_assert_eq instead of assert_eq ?

test/sys/test_pthread.rs
```diff
@@ -20,3 +20,103 @@ fn test_pthread_kill_none() {
...
+        },
+        unistd::{fork, ForkResult},
+    };
+    use std::{mem::size_of, num::NonZeroUsize, os::unix::io::OwnedFd};
+    struct MutexWrapper {
+        lock: Mutex,
+        data: u128,
+    }
+    impl MutexWrapper {
+        fn add(&mut self) {
```
asomers, line 40: There's little point to using a mutex if one already has a `&mut self` reference.  So this isn't a very realistic test.

test/sys/test_pthread.rs
```diff
@@ -20,3 +20,103 @@ fn test_pthread_kill_none() {
...
+
+    let fork_results = (0..FORKS)
+        .map(|_| unsafe { fork().unwrap() })
+        .collect::<Vec<_>>();
+
+    let handles = (0..THREADS)
+        .map(|_| {
+            let bits = mutex_wrapper as usize;
+            std::thread::spawn(move || {
+                let ptr = bits as *mut MutexWrapper;
```
asomers, line 88: Here you create a mutable pointer from a shared reference, and it looks like you had to cast that shared reference to a `usize` just to fool the compiler into allowing it.  Dereferencing the shared reference on line 90 is UB.  This is exactly the kind of pattern that Mutexes should be used to prevent.

asomers (changes requested):
This is a very low-level, C-ish API.  It's got several problems that I would expect a Rust API to be free of:
* Mutex::unlock can trigger UB within the pthread library
* There are no mutex guards, so the programmer must remember to manually unlock the mutex.
* The mutex doesn't protect any data.  That's the `T` in `std::sync::Mutex<T>`.  So the programmer must remember which data the mutex is supposed to protect.  That's even more awkward in Rust than in C, because the protected data is probably accessed via a shared immutable pointer which the programmer must unsafely cast.

Have you looked at the https://github.com/glandium/flexible-locks library yet?  It might already do what you want.

src/sys/pthread.rs
JonathanWoollett-Light: Added

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+        Some(s) => s as c_int,
         None => 0,
     };
     let res = unsafe { libc::pthread_kill(thread, sig) };
     Errno::result(res).map(drop)
 }
 }
+
+/// Mutex protocol.
+#[cfg(target_os = "linux")]
```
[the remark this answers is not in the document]
JonathanWoollett-Light: My understanding is that the implementation of mutex can vary between systems. I think this restriction can be removed or loosened if it also works like this on other systems (but at the moment I do not know).

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+impl std::ops::Drop for MutexAttr {
+    /// Wraps [`libc::pthread_mutexattr_destroy`].
+    fn drop(&mut self) {
+        unsafe {
+            Errno::result(libc::pthread_mutexattr_destroy(&mut self.0)).unwrap();
+        }
+    }
+}
+
+/// Mutex.
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Added.

src/sys/pthread.rs
JonathanWoollett-Light: I believe it is necessary so as to use a specific error type.

Otherwise it automatically defines it `main -> ()` which doesn't work here.

src/sys/pthread.rs
JonathanWoollett-Light: Fixed.

src/sys/pthread.rs
JonathanWoollett-Light: Changed.

src/sys/pthread.rs
JonathanWoollett-Light: Added.

src/sys/pthread.rs
JonathanWoollett-Light: Added a guard and made `Mutex::unlock` unsafe and documented with this.

src/sys/pthread.rs
JonathanWoollett-Light: Removed.

test/sys/test_pthread.rs
JonathanWoollett-Light: I've changed it to pass a reference.

JonathanWoollett-Light:
> This is a very low-level, C-ish API. It's got several problems that I would expect a Rust API to be free of:
> 
>     * Mutex::unlock can trigger UB within the pthread library
> 
>     * There are no mutex guards, so the programmer must remember to manually unlock the mutex.
> 
>     * The mutex doesn't protect any data.  That's the `T` in `std::sync::Mutex<T>`.  So the programmer must remember which data the mutex is supposed to protect.  That's even more awkward in Rust than in C, because the protected data is probably accessed via a shared immutable pointer which the programmer must unsafely cast.
> 
> 
> Have you looked at the https://github.com/glandium/flexible-locks library yet? It might already do what you want.

---

> -  Mutex::unlock can trigger UB within the pthread library
> -  There are no mutex guards, so the programmer must remember to manually unlock the mutex.

I have used a guard mechanism for `Mutex::lock` and marked `Mutex::unlock` as unsafe.

> - The mutex doesn't protect any data.  That's the `T` in `std::sync::Mutex<T>`.  So the programmer must remember which data the mutex is supposed to protect.  That's even more awkward in Rust than in C, because the protected data is probably accessed via a shared immutable pointer which the programmer must unsafely cast.

The problem is that Rust doesn't allow changing representation of imported structs. I require `repr(C)` to use it across-process but this is less efficient `repr(Rust)`, people would want both.

We could add:
```
struct MutexWrapperReprC<T> {
        lock: Mutex,
        data: UnsafeCell<T>,
}
struct MutexWrapperReprRust<T> {
        lock: Mutex,
        data: UnsafeCell<T>,
}
// etc.
```
But this seems quite awkward.

> Have you looked at the https://github.com/glandium/flexible-locks library yet? It might already do what you want.

I did not know about this. I will look at this.

I think it is still valuable to have `Mutex` in Nix (especially given https://github.com/glandium/flexible-locks appears to no longer be maintained).

test/sys/test_pthread.rs
JonathanWoollett-Light: I've changed this to `&self`.

src/sys/pthread.rs
```diff
@@ -34,10 +41,285 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+/// # Ok(())
+/// # }
+/// ```
+#[cfg(target_os = "linux")]
+#[derive(Debug)]
+pub struct Mutex(UnsafeCell<libc::pthread_mutex_t>);
+#[cfg(target_os = "linux")]
+impl Mutex {
+    /// Wraps [`libc::pthread_mutex_init`].
+    pub fn init(&self, attr: Option<MutexAttr>) -> Result<()> {
```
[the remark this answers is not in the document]
JonathanWoollett-Light: This wasn't a leak since it just initializes the given `Mutex` in-place.

But I have changed this to take `*mut Mutex`.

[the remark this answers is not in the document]
asomers: I don't mean a memory leak, but a resource leak.  `pthread_mutex_init` surely creates some resources in the kernel, the C library, or both.  So it should always be paired with `pthread_mutex_destroy`.  Using the `Mutex::init` method on an already-initialized Mutex would have the effect of calling `pthread_mutex_init` twice with no intervening `pthread_mutex_destroy`.

src/sys/pthread.rs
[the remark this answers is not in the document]
asomers: Except for a few details like default attributes it should be the same everywhere.  For pthread functions, a suffix of `_np` indicates that it is _not_ portable, like `pthread_mutexattr_getkind_np`.

src/sys/pthread.rs
asomers: Oh, ok.  Usually I just `.unwrap()` in doc tests, so return type isn't an issue.

src/sys/pthread.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: From https://man7.org/linux/man-pages/man3/pthread_mutex_destroy.3p.html

> Attempting to initialize an already initialized mutex results in undefined behavior.

I've added this in the docs.

src/sys/pthread.rs
```diff
@@ -34,10 +39,351 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+pub struct Mutex(UnsafeCell<libc::pthread_mutex_t>);
+
+#[cfg(target_os = "linux")]
+impl Mutex {
+    /// Wraps [`libc::pthread_mutex_init`].
+    /// 
+    /// # Safety
+    /// 
+    /// Attempting to initialize an already initialized mutex results in undefined behavior.
+    pub unsafe fn init(mutex: *mut Mutex, attr: Option<MutexAttr>) -> Result<()> {
```
asomers, line 296: What I had in mind was passing the raw pointer directly to libc.  I'm not sure that taking a `*mut Mutex` pointer is sufficient, if the target is not initialized.  Can you find any other examples of initializing objects in shared memory?  Would it be better to construct the mutex elsewhere, and then move it into shared memory?  For that matter, the shared memory will only be useful if it can be accessed from more than one process.  How do you plan for the second process to verify that the mutex is initialized?

src/sys/pthread.rs
JonathanWoollett-Light: Yeah, its not important, its more just ✨ aesthetic ✨.

src/sys/pthread.rs
```diff
@@ -34,10 +39,351 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+    /// Wraps [`libc::pthread_mutex_unlock`].
+    /// 
+    /// <https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3p.html>
+    /// 
+    /// Prefer unlocking by dropping the [`MutexGuard`] returned by [`Mutex::lock`] or [`Mutex::try_lock`].
+    /// 
+    /// # Safety
+    /// 
+    /// Results in UB if not called from the thread that locked the mutex.
+    pub unsafe fn unlock(&self) -> Result<()> {
```
asomers, line 347: Can this be a private method?  Why would a user ever want to call it?

src/sys/pthread.rs
```diff
@@ -34,10 +39,351 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+/// Mutex guard to prevent unlocking a mutex from a different thread than the thread that locked it.
+#[cfg(target_os = "linux")]
+#[derive(Debug)]
+pub struct MutexGuard<'a>(&'a Mutex);
+
+#[cfg(target_os = "linux")]
+impl MutexGuard<'_> {
+    /// Calls [`Mutex::unlock`].
+    pub fn try_unlock(self) -> Result<()> {
+        // Prevent calling `Self::Drop` which would attempt to unlock twice.
```
asomers, line 376: What happens if try_unlock fails?  Won't the MutexGuard get dropped, with no way to unlock the mutex later?

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Fix `SigSet` incorrect implementation of `Eq`, `PartialEq` and `Hash`
   ([#1946](https://github.com/nix-rust/nix/pull/1946))
 
+### Added
+
+- Added mutex interface.
```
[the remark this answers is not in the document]
JonathanWoollett-Light: This is an important feature so I understand it may warrant more writing, but I'm just not sure what to write?

It's mostly self explanatory right? Its a mutex.

src/sys/pthread.rs
```diff
@@ -34,10 +39,351 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+impl std::ops::Drop for MutexAttr {
+    /// Wraps [`libc::pthread_mutexattr_destroy`].
+    fn drop(&mut self) {
+        unsafe {
+            Errno::result(libc::pthread_mutexattr_destroy(&mut self.0)).unwrap();
```
asomers, line 146: Panicing during drop is very bad.  It makes bugs hard to debug.  So you should check if the thread is already panicing, and if so don't unwrap the Result.

src/sys/pthread.rs
```diff
@@ -34,10 +39,351 @@ pub fn pthread_kill<T>(thread: Pthread, signal: T) -> Result<()>
...
+
+#[cfg(target_os = "linux")]
+unsafe impl Sync for Mutex {}
+
+#[cfg(target_os = "linux")]
+impl std::ops::Drop for Mutex {
+    /// Wraps [`libc::pthread_mutex_destroy`].
+    fn drop(&mut self) {
+        unsafe {
+            Errno::result(libc::pthread_mutex_destroy(self.0.get())).unwrap();
```
asomers, line 362: Panicing during drop is very bad.  It makes bugs hard to debug.  So you should check if the thread is already panicing, and if so don't unwrap the Result.

JonathanWoollett-Light: I'm not sure I entirely agree with this approach as it adds unavoidable overhead. Is there some more documentation on why to use this approach?

src/sys/pthread.rs
JonathanWoollett-Light: See  https://github.com/nix-rust/nix/pull/1950#discussion_r1342198679

src/sys/pthread.rs
JonathanWoollett-Light: The mutex could then be unlocked with `Mutex::unlock`.

I'm not sure there is any other to handle this that supports getting this error?

Maybe `pub fn try_unlock(self) -> std::result::Result<(), (Self, nix::Error)>`?

src/sys/pthread.rs
JonathanWoollett-Light: See https://github.com/nix-rust/nix/pull/1950#discussion_r1342199592

src/sys/pthread.rs
JonathanWoollett-Light: > How do you plan for the second process to verify that the mutex is initialized?

I'm not sure there is away to do this in this scope. In practice I'd initialize the mutex before forking, but this cannot be captured here.

I think there is only so much this PR can cover, I would say we shouldn't let perfect be the enemy of good.

src/sys/pthread.rs
asomers: Panicing while already unwinding a panic causes a SIGILL signal.  So rather than seeing a nice stack trace, the user just sees that his program was killed by SIGILL.

JonathanWoollett-Light: To an end user (not the developer) I don't think they really care about debug info. I generally take the approach any panic is disastrous. Maybe it'd make sense to do something like:
```rust
impl std::ops::Drop for MutexAttr {
    /// Wraps [`libc::pthread_mutexattr_destroy`].
    fn drop(&mut self) {
        let res = libc::pthread_mutexattr_destroy(&mut self.0);
        #[cfg(debug_assertions)]
        if std::thread::panicked() {
            return;
        }
        Errno::result(res).unwrap();
    }
}
```

asomers:
> The problem is that Rust doesn't allow changing representation of imported structs. I require `repr(C)` to use it across-process but this is less efficient `repr(Rust)`, people would want both.

Is it really that much less efficient?  I think `repr(C)` would probably satisfy.  And it's certainly far better than no RAII at all.

JonathanWoollett-Light:
> > The problem is that Rust doesn't allow changing representation of imported structs. I require `repr(C)` to use it across-process but this is less efficient `repr(Rust)`, people would want both.
> 
> Is it really that much less efficient? I think `repr(C)` would probably satisfy. And it's certainly far better than no RAII at all.

To be honest, I have never seen it make a difference but I guess LLVM implemented struct re-ordering for a reason.

PR #1962, opened 2023-01-09
fix: Update libc

JonathanWoollett-Light:
To include fix https://github.com/rust-lang/libc/commit/44cc30c6b68427d3628926868758d35fe561bbe6.

This is needed to unblock https://github.com/nix-rust/nix/pull/1868.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


asomers (approved):
Approved, though next time you might consider updating the libc dep directly in the PR that needs it.

bors r+

PR #1965, opened 2023-01-10
impl TryFrom<i32> for `nix::mman::MmapAdvise`

acully-vmware:
We have a use-case where we are intercepting the `madvise` syscall, and wish to handle certain types of advice specially. Without `TryFrom`, we need to write boilerplate to safely translate the application-provided `i32` into a form `nix::mman::madvise` will take, and I'm not sure why `madvise` can't simply implement `TryFrom` itself?


acully-vmware:
(For what it's worth, I can't see how the failed check can possibly be related to this change?)

asomers:
The CI failure was spurious, and I've restarted it.  The change looks reasonable, but could you please add an entry to the CHANGELOG?

acully-vmware:
> The CI failure was spurious, and I've restarted it. The change looks reasonable, but could you please add an entry to the CHANGELOG?

Done.

PR #1966, opened 2023-01-11
Added `mq_timedreceive` to `::nix::mqueue`.

DavidCollard:
Noticed when working on a project that `mq_timedreceive` is missing from the `mqueue` module.

Not entirely familiar with best practices around feature gating in tests - I feel like I might have messed up by not including a `#[cfg(feature = "time")]`. Would like a little guidance there.


asomers (approved):
Seems allright to me.  And you did the right thing with feature gating: we don't attempt to feature gate the functional tests at all.  You can only run the tests with all features enabled.

bors r+

PR #1969, opened 2023-01-12
Fix byte-swapped flowinfo and scope_id

jgallagher:
The `From<SockaddrIn6> for net::SocketAddrV6` implementation used `from_be()` on each of the port/flowinfo/scope_id fields, but the flowinfo and scope_id are not stored in big endian, causing the conversion to produce a `SocketAddrV6` with byte-swapped flowinfo and scope_id. I noticed this when running `getifaddrs()` on my machine and then converting to a standard `SocketAddrV6` created an address with a scope ID of `33554432` (hex `0x02000000`) for an interface that should have been scope id 2.

I added a basic test to round-trip between a standard `SocketAddrV6` and a `SockaddrIn6`; prior to the change to the `From` impl, this test failed with

```
thread 'sys::socket::addr::tests::sockaddr_in6::roundtrip_std_socketaddrv6' panicked at 'assertion failed: `(left == right)`
  left: `[1234:5678:90ab:cdef::1111:2222%67305985]:258`,
 right: `[1234:5678:90ab:cdef::1111:2222%16909060]:258`
```

where `67305985 == 0x4030201` is the byte-swapped value of `16909060 == 0x1020304`, the expected value.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -1180,10 +1180,10 @@ impl From<net::SocketAddrV6> for SockaddrIn6 {
 impl From<SockaddrIn6> for net::SocketAddrV6 {
     fn from(addr: SockaddrIn6) -> Self {
         net::SocketAddrV6::new(
-            net::Ipv6Addr::from(addr.0.sin6_addr.s6_addr),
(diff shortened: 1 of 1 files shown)


asomers:
Duplicate of #1964

PR #1973, opened 2023-01-17
if_nameindex: add `target_os = illumos`

jgallagher:
illumos [also has if_nameindex()](https://illumos.org/man/3XNET/if_nameindex); this change allows calling nix's wrappers when building for it.


asomers (approved):
LGTM.  And this is such a tiny change I think it's ok to skip CHANGELOG.

bors r+

PR #1976, opened 2023-01-23
Deps: update to memoffset 0.8

poliorcetics:


poliorcetics:
Yes, `memoffset` is pretty low in dependency chains so having two of it is often not a problem in terms of API, just in term of rebuilds.

Also, it adds const-eval on stable, which is nice for my use case, but not necessarily used by everyone else

asomers (approved):
bors r+

Issue #1977, opened 2023-01-24
`From` impls that take `libc` types are unsound


djkoloski:
`User` and `Group` implement `From<libc::passwd>` and `From<libc::group>`, but these structs contain public fields which are pointers and are not guaranteed to uphold the invariants guaranteed after fetching with functions like `getpw*` and `getgr*`. These functions should be unsafe with the requirement that the passed struct is the output from one of these functions. For example:

```rust
impl User {
    /// # Safety
    ///
    /// `passwd` must be the output from some `getpw*` call, like `getpwuid`.
    unsafe fn from_passwd(passwd: &libc::passwd) -> Self {
        ...
    }
}
```

PR #1978, opened 2023-01-24
Null-check `libc::group` members before converting

djkoloski:
This mirrors the approach used for the `From<&libc::passwd> for User` impl.


djkoloski:
Related: #1977 to guarantee validity of member pointers.

rtzoeller (approved):
bors r+

PR #1979, opened 2023-01-24
Make `*::from_anything` methods unsafe

djkoloski:
The called function must uphold some invariants about initializing data in order for the calls to `from_anything` to be sound.


rtzoeller (approved):
bors r+

PR #1982, opened 2023-01-26
feat: `readv` `iov` as immutable slice .

JonathanWoollett-Light:
The `iov` parameter for `readv` does not need to be a mutable slice, I have changed it to an immutable slice.

The `iov` parameter for `readv` that was previously a mutable slice is now an immutable slice.


asomers (changes requested):
This came up once before and the consensus was that you can't safely have a mutable reference behind an immutable one.  That's why readv needs a mutable slice of IoSliceMut's.  The compiler only things that immutable slices are fine because we only access that argument by pointer.

PR #1988, opened 2023-01-29
Add functions for iterating over Linux filesystems

KoviRobi:
Hi,

I'm relatively new to Rust and this is my first time interfacing Rust with C, I needed to find out if a path is already mounted and I thought I should contribute back :)


CHANGELOG.md
```diff
@@ -106,7 +108,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#1824](https://github.com/nix-rust/nix/pull/1824))
 - Workaround XNU bug causing netmasks returned by `getifaddrs` to misbehave.
   ([#1788](https://github.com/nix-rust/nix/pull/1788))
-
```
VorpalBlade, line 109: I believe you have your editor set to automatically delete trailing whitespaces. I'm not against it, but that should be done as a separate clean up commit.

src/mount/linux.rs
```diff
@@ -85,7 +88,7 @@ libc_bitflags!(
         MNT_DETACH;
         /// Mark the mount point as expired.
         MNT_EXPIRE;
-        /// Don't dereference `target` if it is a symlink.
```
VorpalBlade, line 88: Again: I believe you have your editor set to automatically delete trailing whitespaces. I'm not against it, but that should be done as a separate clean up commit.

src/mount/linux/mntent.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// [`getmntent`](https://www.man7.org/linux/man-pages/man3/getmntent.3.html)
+    /// [`fopen`](https://www.man7.org/linux/man-pages/man3/fopen.3.html)
+    pub fn new<P: ?Sized + NixPath>(path: &P, mode: String) -> Result<Self> {
+        let mode = CString::new(mode).unwrap();
+        let file = path.with_nix_path(|cstr| unsafe {
+            setmntent(cstr.as_ptr(), mode.as_ptr())
+        })?;
+
+        if file.is_null() {
+            panic!("The setmntent call returned NULL");
```
VorpalBlade, line 30: I don't like panic! in general in library code. And this function already returns Result. Why not return a suitable Err here?

src/mount/linux/mntent.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn drop(&mut self) {
+        unsafe { endmntent(self.file) };
+    }
+}
+
+/// Represents an entry in mtab/fstab.
+#[derive(Debug, Eq, PartialEq)]
+pub struct MountEntry {
+    /// 1. name of the filesystem (e.g. the device)
+    pub fs_name: String,
```
VorpalBlade, line 47: These are all owned strings. Is there a way we could do fancy life times to avoid copying?

src/mount/linux/mntent.rs
```diff
@@ -0,0 +1,136 @@
...
+                pass_no: value.mnt_passno,
+            }
+        }
+    }
+}
+
+impl Iterator for MountEntries {
+    type Item = MountEntry;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
VorpalBlade, line 86: Unfortunately this iterator is not sound, specifically due to it not being thread safe. Quoting the man page:

> The getmntent() function reads the next line of the filesystem description file from stream and returns a pointer to a structure containing the broken out fields from a line in the file.  The pointer points to a static area of memory which is overwritten by subsequent calls to getmntent().

and

> MT-Unsafe race:mntentbuf locale 

At least on GNU systems there is also:

> The reentrant getmntent_r() function is similar to getmntent(), but stores the mntent structure in the provided *mntbuf, and stores the strings pointed to by the entries in that structure in the provided array buf of size buflen.

This should probably be preferred.

There is also the following issue:

> The getmntent() and getmntent_r() functions return a pointer to the mntent structure or NULL on failure.

I don't see that case handled anywhere?

VorpalBlade (changes requested):
First: I'm not affiliated with the nix project, so take this review with a grain of salt.

I believe the thread safety issue is a show stopper. And `getmntent_r` is a GNU extension unfortunately. So it would be possible to make this work on glibc systems. I don't know if musl implements that extension, nor do I know about other platforms.

src/mount/linux/mntent.rs
KoviRobi: Possibly, they are only valid until the next call to the iteration, and I am not sure how to encode that so any pointers are welcome :)

src/mount/linux/mntent.rs
KoviRobi: Thanks, the (glibc) man-page wasn't clear about the return of `setmntent` but it returns the same as `fopen`

src/mount/linux/mntent.rs
KoviRobi: For the `NULL` on failure, the `as_ref` handles it, as it'll return `None`: https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref-1.

For `getmntent_r` and reentrancy, the `libc` doesn't implement (even conditionally on `target_env = "gnu"`) the `getmntent_r`, so I marked the `MountEntries` as `!Sync + !Send`, I think that should ensure the soundness? (I'm not experienced in this area so please point out if I'm wrong.)

KoviRobi:
> First: I'm not affiliated with the nix project, so take this review with a grain of salt.

Of course, but I imagine this is helping the maintainers of nix-rust, and is certainly helping me so thank you for your time :)

src/mount/linux/mntent.rs
KoviRobi: I've also added a `doctest` that ensures it should fail to compile when sending `MountEntries` across threads.

VorpalBlade: > For the `NULL` on failure, the `as_ref` handles it, as it'll return `None`: https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref-1.

Oops, my bad.

> For `getmntent_r` and reentrancy, the `libc` doesn't implement (even conditionally on `target_env = "gnu"`) the `getmntent_r`, so I marked the `MountEntries` as `!Sync + !Send`, I think that should ensure the soundness? (I'm not experienced in this area so please point out if I'm wrong.)

I don't believe so. As I understand the documentation in glibc, it is unsafe to call `getmntent` separately from multiple threads. So it means the whole iterator would have to take a static mutex to prevent another thread from concurrently making an unrelated call to `getmntent`.

KoviRobi: Ah thank you, that has made it all click in my head! So it's unsafe because it has some global struct that it's filling with data from the mtab, and if you call it from multiple threads you can get a race condition. If it were a thread-local struct then it would be `!Send + !Sync` as the dereference of the thread local variable wouldn't make sense in another thread, but at least they wouldn't interfere. The mutex should work then though, and we cannot (with `getmntent`) do anything clever with lifetimes.

With `getmntent_r` we could possibly use the trick from `https://www.reddit.com/r/rust/comments/d4rl3d/a_remarkably_simple_solution_to_the_problem_of/` as there we would be the ones allocating the `mntent` struct, though I think it's not really worth bothering given I imagine the whole accessing a file is probably going to be the slowest part anyway.

KoviRobi: Sorry I forgot about this, why would the mutex not be a good idea? It's more portable, and I don't imagine iterating across mounted filesystems is going to be a performance-critical code. But sure I'll look into that :)

VorpalBlade: > Sorry I forgot about this, why would the mutex not be a good idea? It's more portable, and I don't imagine iterating across mounted filesystems is going to be a performance-critical code. But sure I'll look into that :)

It is probably indeed not performance critical. But this is also assuming that no other code wants to use this API (this include foreign code you may be calling via FFI as well as code using alternatives to `nix` such as  `rustix`). To me, using non-reentrant code that *could* be called without your mutex protection (i.e. is not hidden in a completely private module) is unsafe and too much of a footgun. Really, in my opinion there is *no* sane way for *safe* rust code to use this sort of API. There simply isn't a wrapper that is not unsound.

KoviRobi: Aha! Thank you, that makes a lot of sense :) I'll do that then! And again thank you for your patience and advice, I've definitely learned new stuff making this PR

https://github.com/rust-lang/libc/pull/3148

KoviRobi: Okay I have made it use `getmntent_r`, it uses a buffer of constant size, which I've made a parameter. This is what happens if the parameter isn't big enough, on my system anyway
```
❯ evcxr
Welcome to evcxr. For help, type :help
>> :dep nix = { path = "." }
>> use nix::mount::mntent::{MountEntries, MountEntry};
>> for m in MountEntries::<3>::new("/etc/mtab", "r".to_string())? { println!("{:?}", m); }
MountEntry { fs_name: "de", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "de", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "tm", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "pr", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "tm", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "ra", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "tm", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "sy", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "/d", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "/d", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "se", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "cg", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "ps", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "ef", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "bp", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "mq", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "de", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "hu", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "fu", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "co", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "ra", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "sy", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "ra", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "/d", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "/d", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "ra", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "/d", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
MountEntry { fs_name: "tm", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
```
despite https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_627.html#IDX3451 saying
>  The function returns a NULL pointer in error cases. Errors could be:
>
> * error while reading the file,
> * end of file reached,
> * bufsize is too small for reading a complete new entry.

KoviRobi: I did check though, at least it isn't writing values past `buflen`

```console
…/nix-rust on  add-linux-mntent [$!] is 📦 v0.26.1 via 🦀 v1.67.0 via   (nix-shell-env) took 29s at 00:33:40 nu
❯ git diff
diff --git a/src/mount/linux/mntent.rs b/src/mount/linux/mntent.rs
index 1802205..12a94d0 100644
--- a/src/mount/linux/mntent.rs
+++ b/src/mount/linux/mntent.rs
@@ -104,16 +104,18 @@ impl<const CAPACITY: usize> Iterator for MountEntries<CAPACITY> {
             mnt_passno: 0,
         };
         let mut buffer = [0; CAPACITY];
-        unsafe {
+        let ret = unsafe {
             getmntent_r(
                 self.file,
                 &mut mntbuf,
                 buffer.as_mut_ptr(),
-                CAPACITY as i32,
+                (CAPACITY - 10) as i32,
             )
             .as_ref()
             .map(MountEntry::from)
-        }
+        };
+        println!("{:?}", &buffer);
+        ret
     }
 }


…/nix-rust on  add-linux-mntent [$!] is 📦 v0.26.1 via 🦀 v1.67.0 via   (nix-shell-env) at 00:33:42 nu
❯ evcxr
Welcome to evcxr. For help, type :help
>> :dep nix = { path = "." }
use nix::mount::mntent::{MountEntries, MountEntry};
for m in MountEntries::<15>::new("/etc/mtab", "r".to_string())?.take(2) {
  println!("{:?}", m);
}
[100, 101, 118, 116, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
MountEntry { fs_name: "devt", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
[100, 101, 118, 112, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
MountEntry { fs_name: "devp", mount_dir: "", fs_type: "", options: "", dump_freq: 0, pass_no: 0 }
()
>>
```

VorpalBlade (commented):
I'm still wondering if it is possible to avoid all the owned strings. It feels like exposing `Cow<'_, str>` safely is just on the border of what is possible with rust.

src/mount/linux/mntent.rs
KoviRobi: I have looked into it, the problem is this loop

https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/mntent_r.c;h=5493ed359a3c27833329a72e73892e3f48a0f6f4;hb=HEAD#122

```C
  do
    {
      char *end_ptr;

      if (__fgets_unlocked (buffer, bufsiz, stream) == NULL)
	  return false;

      end_ptr = strchr (buffer, '\n');
      if (end_ptr != NULL)	/* chop newline */
	{
	  /* Do not walk past the start of buffer if it's all whitespace.  */
	  while (end_ptr != buffer
		 && (end_ptr[-1] == ' ' || end_ptr[-1] == '\t'))
            end_ptr--;
	  *end_ptr = '\0';
	}
      else
	{
	  /* Not the whole line was read.  Do it now but forget it.  */
	  char tmp[1024];
	  while (__fgets_unlocked (tmp, sizeof tmp, stream) != NULL)
	    if (strchr (tmp, '\n') != NULL)
	      break;
	}

      head = buffer + strspn (buffer, " \t");
      /* skip empty lines and comment lines:  */
    }
  while (head[0] == '\0' || head[0] == '#');;
```

never exits when the buffer is size 1, so only has space for the null character so it is a `while (true)`. So I'd say it's a glibc bug

KoviRobi:
> I'm still wondering if it is possible to avoid all the owned strings. It feels like exposing `Cow<'_, str>` safely is just on the border of what is possible with rust.

I think that is what they are trying to do https://www.reddit.com/r/rust/comments/d4rl3d/a_remarkably_simple_solution_to_the_problem_of/ with `Rc`. I still need to wrap my brain around this a bit more, especially as it's difficult because the buffer we send to `getmntent_r` comes back to us in pieces (split into the fields)

Cargo.toml
```diff
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc", rev = "44cc30c6b68427d3628926868758d35fe561bbe6", features = [ "extra_traits" ] }
+libc = { git = "https://github.com/KoviRobi/libc", rev = "8517146364cd26b061213841b1298766f47dc550", features = [ "extra_traits" ] }
```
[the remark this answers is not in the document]
KoviRobi: Of course, https://github.com/rust-lang/libc/pull/3148, they have been very fast and already merged it

VorpalBlade (approved):

KoviRobi:
Thanks! I've resolved the conflicts (was just the updating libc, which is no longer necessary in this commit). I haven't had any further thoughts about the buffer we allocate as one, then pass in to C which splits it up, re: how to use CoW but if I do I might even remember to come back to this. I think I still need to chase up the glibc bug but I can do that separately to this :)

Issue #1718, opened 2022-05-08
System V IPC support


imachug:
Bindings for the various functions supplied by sysvipc(7) would be great. This includes `msgctl`, `msgget`, `msgrcv`, `msgsnd`, `semctl`, `semget`, `semop`, `shmat`, `shmctl`, `shmdt`, `shmget` (list copied verbatim from man page).

asomers:
Would you be able to put together a PR?

PR #1989, opened 2023-01-30
Add System V shared memory APIs

arpankapoor:
First draft to add System V shared memory APIs.
I would like some help with these questions:
1. How to decide if an exported function should be `unsafe` or not? For now, I've marked functions taking in any pointer arguments as unsafe.
2. Should we have `libc_bitflags` for `shmflg` or `cmd`? There aren't a set of constants with a fixed prefix as described in the [CONVENTIONS.md](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#bitflags) file. There are a few `IPC_*` and `SHM_*` constants for them

Partially addresses #1718.

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -7,7 +7,7 @@ use crate::Result;
 #[cfg(not(target_os = "android"))]
 #[cfg(feature = "fs")]
 use crate::{fcntl::OFlag, sys::stat::Mode};
-use libc::{self, c_int, c_void, off_t, size_t};
+use libc::{self, c_int, c_void, key_t, off_t, shmid_ds, size_t};
 use std::{num::NonZeroUsize, os::unix::io::{AsRawFd, AsFd}};
 
 libc_bitflags! {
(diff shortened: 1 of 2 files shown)


src/sys/mman.rs
```diff
@@ -603,3 +603,64 @@ pub fn shm_unlink<P: ?Sized + NixPath>(name: &P) -> Result<()> {
...
     Errno::result(ret).map(drop)
 }
+
+/// Creates and returns a new, or returns an existing, System V shared memory
+/// segment identifier.
+///
+/// For more information, see [`shmget(2)`].
+///
+/// [`shmget(2)`]: https://man7.org/linux/man-pages/man2/shmget.2.html
+pub fn shmget(key: key_t, size: size_t, shmflg: c_int) -> Result<c_int> {
```
VorpalBlade, line 613: In general (not just for shmget): I would not expose key_t directly. It is just a type alias, and thus not safe. Looking at the docs, some sort of newtype or enum would be more appropriate.

VorpalBlade, line 613: `shmflg` should be an enum of the valid values, not a raw C integer. I believe that there is a special macro in nix (`libc_enum!` or something like that) that will streamline the creation of this.

src/sys/mman.rs
```diff
@@ -603,3 +603,64 @@ pub fn shm_unlink<P: ?Sized + NixPath>(name: &P) -> Result<()> {
...
+/// # Safety
+///
+/// `shmid` should be a valid shared memory identifier and
+/// `shmaddr` must meet the requirements described in the [`shmat(2)`] man page.
+///
+/// [`shmat(2)`]: https://man7.org/linux/man-pages/man2/shmat.2.html
+pub unsafe fn shmat(
+    shmid: c_int,
+    shmaddr: *const c_void,
+    shmflg: c_int,
```
VorpalBlade, line 631: Again: `shmflg`, `shmid` should be wrapped in safe abstractions. Reading the man pages it even seems like shmat and shmget take different sets of flags. So two different enums probably?

src/sys/mman.rs
VorpalBlade, line 613: At the very least the result type should be a newtype pattern. I.e:

```rust
-> Result<ShmKey>
```

where 

```rust
#[derive(Debug, PartialEq, Eq, Hash)]
struct ShmKey(c_int)

impl ShmKey {
    unsafe fn to_raw(&self) -> c_int {
        self.0
    }
    // unsafe from_raw, other methods
}
```

This helps ensure that you will get sensible compiler errors.

src/sys/mman.rs
```diff
@@ -603,3 +603,64 @@ pub fn shm_unlink<P: ?Sized + NixPath>(name: &P) -> Result<()> {
...
+/// memory segment given by `shmid`.
+///
+/// For more information, see [`shmctl(2)`].
+///
+/// # Safety
+///
+/// All arguments should be valid and meet the requirements described in the [`shmctl(2)`] man page.
+///
+/// [`shmctl(2)`]: https://man7.org/linux/man-pages/man2/shmctl.2.html
+pub unsafe fn shmctl(
```
VorpalBlade, line 660: Again: `shmflg` & `cmd` should be wrapped in safe abstractions. Note that at least on Linux there seems to be some Linux specific commands. So `cfg` compile time guards will be needed to construct the enum.

We also don't want to expose shmid_ds directly to the user I would say. It too needs rustification (though I haven't looked into the details of that).

src/sys/mman.rs
```diff
@@ -603,3 +603,64 @@ pub fn shm_unlink<P: ?Sized + NixPath>(name: &P) -> Result<()> {
...
+    shmid: c_int,
+    cmd: c_int,
+    buf: *mut shmid_ds,
+) -> Result<c_int> {
+    Errno::result(libc::shmctl(shmid, cmd, buf))
```
VorpalBlade, line 665: It also seems the meaning of the return value changes depending on the specific command, so it is not correct to treat it as errno unconditionally:

Quoting man 2 shmctl on my system:

> A successful IPC_INFO or SHM_INFO operation returns the index of the highest used entry in the kernel's internal array recording information about all shared  memory  segments.   (This  information  can  be  used  with  repeated  SHM_STAT  or SHM_STAT_ANY  operations to obtain information about all shared memory segments on the system.)  A successful SHM_STAT operation returns the identifier of the shared memory segment whose index was given in shmid.  Other operations return 0 on success.

VorpalBlade (changes requested):
First off, I'm not affiliated with nix, so take this review with a grain of salt.

However I do see a big issue with your approach. It is a very low level API. It basically
just exposes the raw libc API. Is there a way to build a memory safe API on top instead?

Even if we wrap key_t, the various flags etc in rust abstractions, that is still rather low level. Can we do better? What would a `struct ShmMemory` look like? Is it possible to make something like that will be generally useful (not restricting certain use cases).

PR #1992, opened 2023-01-31
Don't deref uninit addrs in `pack_mhdr_to_receive`

### LINKED ISSUE #1990: `Multiheader::preallocate` causes an uninitialized value to be dereferenced
In `src/sys/socket/mod.rs`, `Multiheader::preallocate` allocates a vec or `MaybeUninit::uninit()` and then passes each `address.as_mut_ptr()` to `pack_mhdr_to_receive`. `pack_mhdr_to_receive` in turn performs `(*address).as_mut_ptr()`, which results in undefined behavior.

To support this desired behavior `SockaddrLikePriv` must change the signature of `as_mut_ptr` to `fn(*mut Self) -> *mut libc::sockaddr` and use `ptr::addr_of_mut!` to access fields in implementations. It does not appear that performing `*mut S as *mut libc::sockaddr` is possible because `()` implements `SockaddrLike` but returns `null_mut()` from its `SockaddrLikePriv` implementation.

djkoloski:
This adds a new method to `SockaddrLikePriv` to get a pointer to some underlying `libc::sockaddr` from a pointer to some type. This avoids dereferencing uninitialized values during `pack_mhdr_to_receive`, which is called from `MultiHeaders::preallocate` and `recvmsg`.

Because this is a private trait, this is not a semver-breaking change.

Fixes #1990

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -910,6 +910,10 @@ impl private::SockaddrLikePriv for () {
     fn as_mut_ptr(&mut self) -> *mut libc::sockaddr {
         ptr::null_mut()
     }
+
+    fn as_mut_ptr_raw(_: *mut Self) -> *mut libc::sockaddr {
+        ptr::null_mut()
+    }
 }
 
 /// `()` can be used in place of a real Sockaddr when no address is expected,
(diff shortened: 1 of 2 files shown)


aarond10:
Can we get this rolled into a release?

djkoloski:
This UB was introduced in https://github.com/nix-rust/nix/pull/1684. Maybe @asomers would be interested in reviewing it?

aarond10:
Ping! @asomers, could we get this merged in?

PR #1993, opened 2023-01-31
Make `clone` unsafe

### LINKED ISSUE #1975: `sched::clone` is unsound and should be unsafe to call
Reproduced on Debian 5.19.11:

```rust
use nix::sched::{clone, CloneFlags};

fn main() {
    let mut bytes = [0u8; 1024];
    let cb = Box::new(|| 0isize) as Box<dyn FnMut() -> isize>;
    let _ = clone(cb, &mut bytes[1024..], CloneFlags::CLONE_VM, None);
    for byte in bytes {
        assert_eq!(byte, 0);
    }
}
```

Output:

```rust
  left: `48`,
 right: `0`', src/main.rs:8:9
*** stack smashing detected ***: terminated
Segmentation fault
```

This function should be unsafe to call. Related: #360

djkoloski:
There are many features of `clone` that may cause memory unsafety when called. This documents one of them and references `fork()`, which is already unsafe to call.

Fixes #1975

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#2014](https://github.com/nix-rust/nix/pull/2014))
 - Added `SO_TS_CLOCK` for FreeBSD to `nix::sys::socket::sockopt`.
   ([#2093](https://github.com/nix-rust/nix/pull/2093))
+- Made `clone` unsafe, like `fork`
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
Um yes, clone is nearly impossible to use safely.  I see that the clone binding has been present in Nix since its very first commit in 2014, back before Rust 1.0.  I'm guessing that almost no one has used it since then, which is why nobody has made it `unsafe` yet.  There's only one problem with this PR: could you add a CHANGELOG entry?

djkoloski:
Sorry for the delay, should be good to go now.

asomers:
Sorry I missed this PR when last you updated it.  I've just hijacked it to fix the merge conflict.

asomers (approved):

PR #1995, opened 2023-02-01
Add `AT_EACCESS` to `AtFlags` on all platforms

MatteoNardi:
On Linux, I need the `AT_EACCESS` flag for `faccessat`:
```
AT_EACCESS
       Perform access checks using the effective user and group IDs.  By default, faccessat() uses the real IDs (like access()).
```

~~This commit is enabling it for all platforms but redox. I'm not really sure if this is correct of I should have just added a target_os = "linux". A did quick git grep -e AT_EACCESS -e 'AT_REMOVEDIR' on libc and it showed they're available on the same targets, so for consistence I just removed the cfg limitation.~~ CI was helpful and showed me I was wrong: it's undefined on android.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 ### Added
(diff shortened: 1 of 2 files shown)


asomers (approved):
bors r+

Issue #1994, opened 2023-02-01
hints for adding sigsuspend().


castilma:
I'm a rust beginner and I'd like to practice a bit by adding sigsuspend. But I need some hints.

I thought adding it to SigSet right after wait. Regarding the name I'd choose `suspend`, since `wait` had the prefix `sig` stripped, too.

If sigsuspend returns, it always returns the same value. Is it necessary to return that value to the caller?

Do I have to add some special documentation because it may not return?

asomers:
> I'm a rust beginner and I'd like to practice a bit by adding sigsuspend. But I need some hints.
> 
> I thought adding it to SigSet right after wait. Regarding the name I'd choose `suspend`, since `wait` had the prefix `sig` stripped, too.

Makes sense.

> 
> If sigsuspend returns, it always returns the same value. Is it necessary to return that value to the caller?

Not in the success case.  But it might also fail with EFAULT.

> 
> Do I have to add some special documentation because it may not return?

No.

castilma:
> Not in the success case. But it might also fail with EFAULT.
There is no 'success' case. It never returns with a success. It's always an 'error' case, where errno is set to distinguish between the errors. And since errno is global, I don't have to return it explicitly, right?

PR #1997, opened 2023-02-06
Add `suspend` to `sys::signal::SigSet`

castilma:
This adds a simple wrapper for libc::sigsuspend.  `sigsuspend` either doesn't return (if the sighandler terminates the process) or returns -1 and sets errno to EITR.  So if this function returns, the return value must be -1 and does not need to be explicitly returned by `SigSet::suspend`.

In issue #1994 I asked for hints. This is what I got.

I want to add a test, but have difficulties with kill and making sure the test never blocks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#1912](https://github.com/nix-rust/nix/pull/1912))
 - Added `mq_timedreceive` to `::nix::mqueue`.
   ([#1966])(https://github.com/nix-rust/nix/pull/1966)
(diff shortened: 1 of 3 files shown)


VorpalBlade (commented):
I'm not affiliated with the nix project, so take this review with a grain of salt. I simply noticed that the project seems a bit overloaded and decided to help.

Overall this seems good. However:
* There are CI failures that needs to be fixed (obviously)
* Errno handling for non-Linux might need further investigation.
* The test feels a bit work in progress to be honest. So I won't really review it in it's current state. TODO comments etc needs to be dealt with.

SteveLauC:
Hi @castilma, thanks for your contribution to Nix, this PR didn't get a review due to the lack of review bandwidth, and #2279 implements this in a more complete way, so I am gonna close this, I am really sorry.

PR #1999, opened 2023-02-08
inotify: Add AsFd to allow using with epoll (issue #1998)

### LINKED ISSUE #1998: Combing Epoll and Inotify?
So apparently the crate is transitioning `Epoll` to a new safer (and better documented!) abstraction. But this has yet to be released. Fine I thought, since I'm starting a new project, I will just use the latest master for now and switch to the next release that contains this.

However, it seems that `Inotify` (which I also need) has lost `AsRawFd`, but not gained a replacement to allow it to be used together with `Epoll`? For my use case I need to epoll a few FDs, one of which is an inotify FD (that monitors a few files). Thus I need a way to combine these. I would prefer to avoid threads for this.

Is there no way to do this currently or am I missing something obvious?

VorpalBlade:
This resolves issue #1998 and allows `Inotify` to be used by `Epoll` by adding AsFd.

I'm not entirely sure about the unit test. Maybe it would be possible to do a more comperhensive check by contructing inotify using `from_raw_fd` and checking that I get the same value back. However, that would basically mean duplicating  `Inotify::new` and that feels a bit pointless.

Another option would be to create an integration test to combine `Inotify` and `Epoll`.

Fixes #1998

### DIFF
--- a/src/sys/inotify.rs
+++ b/src/sys/inotify.rs
@@ -32,7 +32,7 @@ use libc::{c_char, c_int};
 use std::ffi::{CStr, OsStr, OsString};
 use std::mem::{size_of, MaybeUninit};
 use std::os::unix::ffi::OsStrExt;
-use std::os::unix::io::{AsRawFd, FromRawFd, OwnedFd, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, OwnedFd, RawFd};
 use std::ptr;
 
 libc_bitflags! {
@@ -238,5 +238,11 @@ impl Inotify {
 impl FromRawFd for Inotify {
     unsafe fn from_raw_fd(fd: RawFd) -> Self {
         Inotify { fd: OwnedFd::from_raw_fd(fd) }
+    }
+}
+
+impl AsFd for Inotify {
+    fn as_fd(&'_ self) -> BorrowedFd<'_> {
+        self.fd.as_fd()
     }
 }


VorpalBlade:
Huh I don't understand the CI build failures. It works for me on stable rust (1.67.0). Yet it seems to be running on a much older version since it complains about AsFd not being a thing (supported since 1.63.0, which is the version that adds OwnedFd too, which was already used in that file).

VorpalBlade:
I did a change (where AsFd/BorrowedFd are imported from), but I do not understand why it was needed, as it builds without that change on my system. I would appreciate some explanation on this!

It also seems I get other errors building on CI than locally as well. I'm really confused by this. How do I build locally the same was as on CI?

asomers:
I don't remember the exact version numbers, but one version of Rust stabilized the `AsFd` trait, and then a later version moved it to a different module within the standard library.  That's why the first version of your PR failed.  You must've been using a new Rust toolchain locally.

This PR looks good, but you don't need to bother with a test that short.  If a one-line function can't be tested without essentially just restating what that function does, then the test adds no value.

VorpalBlade:
> I don't remember the exact version numbers, but one version of Rust stabilized the `AsFd` trait, and then a later version moved it to a different module within the standard library. That's why the first version of your PR failed. You must've been using a new Rust toolchain locally.

Ah, that explains it. A bit of a footgun though in my opinion!

> This PR looks good, but you don't need to bother with a test that short. If a one-line function can't be tested without essentially just restating what that function does, then the test adds no value.

I have removed the test now and squashed the changes into a single commit. Should be ready to merge!

asomers (approved):
LGTM.

bors r+

PR #2000, opened 2023-02-15
feat: mman `NonNull`

JonathanWoollett-Light:
Updates functions in  `nix::sys::mman` to use `NonNull`.

The case where the returned pointer is null is the error case, therefore in the `Ok` case we can return `NonNull`.

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -537,11 +537,11 @@ pub unsafe fn munmap(addr: NonNull<c_void>, len: size_t) -> Result<()> {
 ///
 /// [`madvise(2)`]: https://man7.org/linux/man-pages/man2/madvise.2.html
 pub unsafe fn madvise(
-    addr: *mut c_void,
+    addr: NonNull<c_void>,
     length: size_t,
     advise: MmapAdvise,
 ) -> Result<()> {
-    Errno::result(libc::madvise(addr, length, advise as i32)).map(drop)
+    Errno::result(libc::madvise(addr.as_ptr(), length, advise as i32)).map(drop)
 }
 
 /// Set protection of memory mapping.
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -375,8 +375,8 @@ libc_bitflags! {
 /// `addr` must meet all the requirements described in the [`mlock(2)`] man page.
 ///
 /// [`mlock(2)`]: https://man7.org/linux/man-pages/man2/mlock.2.html
-pub unsafe fn mlock(addr: *const c_void, length: size_t) -> Result<()> {
-    Errno::result(libc::mlock(addr, length)).map(drop)
+pub unsafe fn mlock(addr: NonNull<c_void>, length: size_t) -> Result<()> {
+    Errno::result(libc::mlock(addr.as_ptr(), length)).map(drop)
 }
 
 /// Unlocks all memory pages that contain part of the address range with
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -8,10 +8,8 @@ use crate::Result;
 #[cfg(feature = "fs")]
 use crate::{fcntl::OFlag, sys::stat::Mode};
 use libc::{self, c_int, c_void, off_t, size_t};
-use std::{
-    num::NonZeroUsize,
-    os::unix::io::{AsFd, AsRawFd},
-};
+use std::{num::NonZeroUsize, os::unix::io::{AsRawFd, AsFd}};
+use std::ptr::NonNull;
 
 libc_bitflags! {
     /// Desired memory protection of a memory mapping.
@@ -430,7 +428,7 @@ pub unsafe fn mmap<F: AsFd>(
     flags: MapFlags,
     f: F,
     offset: off_t,
-) -> Result<*mut c_void> {
+) -> Result<NonNull<c_void>> {
     let ptr = addr.map_or(std::ptr::null_mut(), |a| a.get() as *mut c_void);
 
     let fd = f.as_fd().as_raw_fd();
@@ -439,9 +437,12 @@ pub unsafe fn mmap<F: AsFd>(
 
     if ret == libc::MAP_FAILED {
         Err(Errno::last())
-    } else {
-        Ok(ret)
     }
+    else {
+        // SAFETY: `libc::mmap` returns a valid non-null pointer or `libc::MAP_FAILED`, thus `ret`
+        // will be non-null here.
+        Ok(unsafe { NonNull::new_unchecked(ret) })
+    } 
 }
 
 /// Create an anonymous memory mapping.
@@ -459,7 +460,7 @@ pub unsafe fn mmap_anonymous(
     length: NonZeroUsize,
     prot: ProtFlags,
     flags: MapFlags,
-) -> Result<*mut c_void> {
+) -> Result<NonNull<c_void>> {
     let ptr = addr.map_or(std::ptr::null_mut(), |a| a.get() as *mut c_void);
 
     let flags = MapFlags::MAP_ANONYMOUS | flags;
@@ -468,7 +469,9 @@ pub unsafe fn mmap_anonymous(
     if ret == libc::MAP_FAILED {
         Err(Errno::last())
     } else {
-        Ok(ret)
+        // SAFETY: `libc::mmap` returns a valid non-null pointer or `libc::MAP_FAILED`, thus `ret`
+        // will be non-null here.
+        Ok(unsafe { NonNull::new_unchecked(ret) })
     }
 }
 
@@ -559,8 +562,8 @@ pub unsafe fn madvise(
 /// let mut slice: &mut [u8] = unsafe {
 ///     let mem = mmap_anonymous(None, one_k_non_zero, ProtFlags::PROT_NONE, MapFlags::MAP_PRIVATE)
 ///         .unwrap();
-///     mprotect(mem, ONE_K, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE).unwrap();
-///     std::slice::from_raw_parts_mut(mem as *mut u8, ONE_K)
+///     mprotect(mem.as_ptr(), ONE_K, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE).unwrap();
+///     std::slice::from_raw_parts_mut(mem.as_ptr().cast(), ONE_K)
 /// };
 /// assert_eq!(slice[0], 0x00);
 /// slice[0] = 0xFF;
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -484,25 +484,25 @@ pub unsafe fn mmap_anonymous(
 /// detailed requirements.
 #[cfg(any(target_os = "linux", target_os = "netbsd"))]
 pub unsafe fn mremap(
-    addr: *mut c_void,
+    addr: NonNull<c_void>,
     old_size: size_t,
     new_size: size_t,
     flags: MRemapFlags,
-    new_address: Option<*mut c_void>,
-) -> Result<*mut c_void> {
+    new_address: Option<NonNull<c_void>>,
+) -> Result<NonNull<c_void>> {
     #[cfg(target_os = "linux")]
     let ret = libc::mremap(
-        addr,
+        addr.as_ptr(),
         old_size,
         new_size,
         flags.bits(),
-        new_address.unwrap_or(std::ptr::null_mut()),
+        new_address.map(NonNull::as_ptr).unwrap_or(std::ptr::null_mut()),
     );
     #[cfg(target_os = "netbsd")]
     let ret = libc::mremap(
-        addr,
+        addr.as_ptr(),
         old_size,
-        new_address.unwrap_or(std::ptr::null_mut()),
+        new_address.map(NonNull::as_ptr).unwrap_or(std::ptr::null_mut()),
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 - Added `Icmp` and `IcmpV6` to `SockProtocol`.
   (#[2103](https://github.com/nix-rust/nix/pull/2103))
+- `sys::mman::mmap` to return `NonNull<c_void>`.
+  ([#2000](https://github.com/nix-rust/nix/pull/2000))
```
SteveLauC, line 45: This change should be put under `### Changed` rather than `### Added`?

CHANGELOG.md
```diff
@@ -123,6 +125,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   `sys::kevent::Kqueue::kevent`, and `sys::event::kqueue` is deprecated in
   favor of `sys::kevent::Kqueue::new`.
   ([#1943](https://github.com/nix-rust/nix/pull/1943))
+
+### Removed
```
SteveLauC, line 129: This seems to be added by accident

src/sys/mman.rs
```diff
@@ -425,19 +426,15 @@ pub unsafe fn mmap<F: AsFd>(
...
     let ret =
         libc::mmap(ptr, length.into(), prot.bits(), flags.bits(), fd, offset);
-
-    if ret == libc::MAP_FAILED {
-        Err(Errno::last())
-    } else {
-        Ok(ret)
-    }
+        
+    NonNull::new(ret).ok_or_else(Errno::last)
```
SteveLauC, line 437: [`NonNull::new()`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.new) will ONLY return a `None` if its argument is a null pointer, the error case of `mmap`, whose return value `libc::MAP_FAILED` ((void *) -1 = 0xffffffff) is not null:

```rust
use std::ptr::NonNull;

fn main() {
    let ptr = NonNull::new(libc::MAP_FAILED);
    println!("{:?}", ptr);
}
```

```
$ cargo r -q
Some(0xffffffffffffffff)
```

which means that the current impl will return `Ok(NonNull(0xffffffffff))` even when the underlying `mmap()` failed

SteveLauC (changes requested):

SteveLauC:
Given that `mmap()` won't return a null pointer at all, I am actually not sure if this change is needed, thoughts on this?

cc @asomers

CHANGELOG.md
JonathanWoollett-Light: Changed.

CHANGELOG.md
JonathanWoollett-Light: Removed.

src/sys/mman.rs
JonathanWoollett-Light: I've updated the implementation considering this.

JonathanWoollett-Light:
> n that `mmap()` won't return a null pointer at all, I am actually not sure if this change is needed, thoughts on this?

If it always returns a non-null pointer this seems even more reason to return `NonNull`.

src/sys/mman.rs
```diff
@@ -554,10 +557,10 @@ pub unsafe fn madvise(
...
 /// let one_k_non_zero = std::num::NonZeroUsize::new(ONE_K).unwrap();
 /// let mut slice: &mut [u8] = unsafe {
-///     let mem = mmap_anonymous(None, one_k_non_zero, ProtFlags::PROT_NONE, MapFlags::MAP_PRIVATE)
-///         .unwrap();
-///     mprotect(mem, ONE_K, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE).unwrap();
-///     std::slice::from_raw_parts_mut(mem as *mut u8, ONE_K)
+///     let mem = mmap::<BorrowedFd>(None, one_k_non_zero, ProtFlags::PROT_NONE,
+///                    MapFlags::MAP_ANON | MapFlags::MAP_PRIVATE, None, 0).unwrap();
+///     mprotect(mem.as_ptr(), ONE_K, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE).unwrap();
+///     std::slice::from_raw_parts_mut(mem.as_ptr().cast(), ONE_K)
```
SteveLauC, line 563: Please revert this change as we have a dedicated `mmap_anonymous()` interface now

SteveLauC (changes requested):

SteveLauC:
> > Given that `mmap()` won't return a null pointer at all, I am actually not sure if this change is needed, thoughts on this?
> 
> If it always returns a non-null pointer this seems even more reason to return `NonNull`.


I think both our points are correct here:

From the interface safety perspetive, `mmap()` won't return a NULL pointer, so wrapping it in a `NonNull` is useless. However, for pointers that are not NULL, wrapping it in a `NonNull` will definitely make the interface more Rusty.

So I am not against this now, but still ping @asomers, I would like to hear your thoughts

src/sys/mman.rs
JonathanWoollett-Light: I've removed this change.

Jan561:
Just adding my two cents here.

`NonNull<T>` differs from `*mut T` in two aspects:

- The pointer is guaranteed to be not NULL (obviously). This allows the compiler to perform some [optimizations].
- `NonNull<T>` is covariant in `T`. This is usually desired when dealing with *unique* pointers (i.e. we own `T`) or when dealing with *shared* pointers (e.g. shared ownership `Rc`, `Arc`).

`mmap` allocates new virtual memory and returns a unique pointer to the allocation, which exactly fits the use case of representing ownership with `NonNull`. This alone makes this change worthwhile for me.

Additionally, this function is fallible and we are only one call away of turning the result into an `Option` by calling `Result::ok`, where the aforementioned optimizations apply.

The only reason why `NonNull` is so underused even in the standard library is that many functions have already been stabilized before `NonNull` was invented. The first (?) API in std that will use `NonNull` instead of `*mut` is the upcoming [Allocator] API, which coincidentally is also what `mmap` essentially is.

[optimizations]: https://doc.rust-lang.org/std/option/index.html#representation
[Allocator]: https://doc.rust-lang.org/std/alloc/trait.Allocator.html

SteveLauC:
> Just adding my two cents here.
> 
> `NonNull<T>` differs from `*mut T` in two aspects:
> 
> * The pointer is guaranteed to be not NULL (obviously). This allows the compiler to perform some [optimizations](https://doc.rust-lang.org/std/option/index.html#representation).
> * `NonNull<T>` is covariant in `T`. This is usually desirable when dealing with _unique_ pointers (i.e. we own `T`, e.g. `Vec`, `Box`) or when dealing with _shared_ pointers (e.g. shared ownership `Rc`, `Arc`).
> 
> `mmap` allocates new virtual memory and returns a unique pointer to the allocation, which exactly fits the use case of representing ownership with `NonNull`. This alone makes this change worthwhile for me.
> 
> Additionally, this function is fallible and we are only one call away of turning the result into an `Option` by calling `Result::ok`, where the aforementioned optimizations apply.
> 
> The only reason why `NonNull` is so underused even in the standard library is that many functions have already been stabilized before `NonNull` was invented. The first (?) API in std that will use `NonNull` instead of `*mut` is the upcoming [Allocator](https://doc.rust-lang.org/std/alloc/trait.Allocator.html) API, which coincidentally is also what `mmap` essentially is.

Makes sense to me, if we are gonna merge  this change, should we also update the `munmap(2)` interface to something like?

```rs
pub unsafe fn munmap(addr: NonNull<c_void>, len: size_t) -> Result<()>
```

Jan561:
Yes, good point. For symmetry reasons, I believe all other functions in [mman] should implement this change, too.

[mman]: https://docs.rs/nix/latest/nix/sys/mman/index.html

JonathanWoollett-Light:
> Yes, good point. For symmetry reasons, I believe all other functions in [mman](https://docs.rs/nix/latest/nix/sys/mman/index.html) should implement this change, too.

I've updated the PR with all the other changes.

Jan561:
> > Yes, good point. For symmetry reasons, I believe all other functions in [mman](https://docs.rs/nix/latest/nix/sys/mman/index.html) should implement this change, too.
> 
> I've updated the PR with all the other changes.
> 
> I think a follow-up PR could update the functions to take generic `NonNull<[T]>` instead of taking `NonNull<c_void>` and `size_t`.

I agree, this is a way nicer abstraction, but since this function allocates in pages and doesn't care about the layout of `T`, an idiomatic `mmap` should probably just return a page-aligned `NonNull<[u8]>` and the other functions taking the allocation as an argument should use the page-aligned starting pointer `NonNull<u8>`, similar to the std allocator trait.

The metadata of the slice pointer would then be the `length` rounded up to the next multiple of the page size.

test/sys/test_mman.rs
```diff
@@ -43,7 +47,7 @@ fn test_mremap_grow() {
     let slice: &mut [u8] = unsafe {
         #[cfg(target_os = "linux")]
         let mem = mremap(
-            slice.as_mut_ptr().cast(),
+            NonNull::from(&slice[..]).cast(),
```
asomers, line 50: Here you're mutating through an immutable pointer.  That's not ok.  You should do `Slice::as_mut_ptr` or `&mut slice` instead.  It happens a few places below, too.

asomers (changes requested):
I was on the fence about this, because NonNull is less ergonomic than a plain pointer.  But symmetry with Allocator is a good enough reason to do this, I think.  I just noted one minor problem in some of the tests.

test/sys/test_mman.rs
JonathanWoollett-Light: Updated it to `&mut slice[..]`

src/sys/mman.rs
```diff
@@ -439,9 +437,10 @@ pub unsafe fn mmap<F: AsFd>(
 
     if ret == libc::MAP_FAILED {
         Err(Errno::last())
-    } else {
-        Ok(ret)
     }
+    else {
+        Ok(unsafe { NonNull::new_unchecked(ret) })
```
SteveLauC, line 442: Given that this is unsafe, do you think that we could add a comment stating it is safe, maybe something like:

> `mmap()` will never return a NULL pointer, it is safe to use `NonNull::new_unchecked()` here

src/sys/mman.rs
```diff
@@ -468,7 +467,7 @@ pub unsafe fn mmap_anonymous(
     if ret == libc::MAP_FAILED {
         Err(Errno::last())
     } else {
-        Ok(ret)
+        Ok(unsafe { NonNull::new_unchecked(ret) })
```
SteveLauC, line 470: Also this one

src/sys/mman.rs
```diff
@@ -481,33 +480,33 @@ pub unsafe fn mmap_anonymous(
...
+        new_address.map(NonNull::as_ptr).unwrap_or(std::ptr::null_mut()),
         new_size,
         flags.bits(),
     );
 
     if ret == libc::MAP_FAILED {
         Err(Errno::last())
     } else {
-        Ok(ret)
+        Ok(unsafe { NonNull::new_unchecked(ret) })
```
SteveLauC, line 509: And this one: `mremap()` will never return a NULL pointer, it is safe to use `NonNull::new_unchecked()` here

src/sys/mman.rs
```diff
@@ -519,8 +518,8 @@ pub unsafe fn mremap(
 /// page.
 ///
 /// [`munmap(2)`]: https://man7.org/linux/man-pages/man2/munmap.2.html
-pub unsafe fn munmap(addr: *mut c_void, len: size_t) -> Result<()> {
-    Errno::result(libc::munmap(addr, len)).map(drop)
+pub unsafe fn munmap(mut addr: NonNull<c_void>, len: size_t) -> Result<()> {
+    Errno::result(libc::munmap(addr.as_mut(), len)).map(drop)
```
SteveLauC, line 522: Why we are using `NonNull.as_mut()` for `munmap()`? Other functions are all using `NonNull.as_ptr()`

test/sys/test_mman.rs
SteveLauC: It seems that we can get rid of that `#![allow(clippy::redundant_slicing)]` if we just use the `slice` here, `slice` itself is of type `&mut [u8]`, the constructed `NonNull` will be `NonNull<[u8]>`, then we `cast()` it to `NonNull<c_void>`

JonathanWoollett-Light: I tried this, it doesn't work.

src/sys/mman.rs
JonathanWoollett-Light: Added comment.

src/sys/mman.rs
JonathanWoollett-Light: Added comment.

src/sys/mman.rs
JonathanWoollett-Light: Added comment.

src/sys/mman.rs
JonathanWoollett-Light: Added comment.

test/sys/test_mman.rs
SteveLauC: Oops, sorry, `&mut [u8]` is not `Copy`, manually reborrowing here should do the trick

SteveLauC (approved):

PR #2005, opened 2023-02-24
Revert libc::memfd_create for gnu and musl

### LINKED ISSUE #1972: undefined reference to `memfd_create'
```
= note: /root/build/release/deps/libnix-fca8cee73dd370df.rlib(nix-fca8cee73dd370df.nix.9961a42b-cgu.2.rcgu.o): In function `nix::sys::memfd::memfd_create::hbef039c9926ba4f0':
          nix.9961a42b-cgu.2:(.text._ZN3nix3sys5memfd12memfd_create17hbef039c9926ba4f0E+0x5): undefined reference to `memfd_create'
          collect2: error: ld returned 1 exit status
          
  = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
```

```
root@f0fca0e971de:~/rc3.1# gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/5/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 5.4.0-6ubuntu1~16.04.12' --with-bugurl=file:///usr/share/doc/gcc-5/README.Bugs --enable-languages=c,ada,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-5 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-5-amd64/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-5-amd64 --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-5-amd64 --with-arch-directory=amd64 --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --enable-objc-gc --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.12) 
```

StephenWakely:
Closes #1972 

https://github.com/nix-rust/nix/commit/cf15c2bb5505231bc3128ef61522336e315e2169 introduced changes to force a hard dependency on `libc::memfd_create`. This symbol doesn't exist in Ubuntu 16.04 or CentOS 7 and thus it [breaks our builds](https://github.com/vectordotdev/vector/actions/runs/4252121295/jobs/7395320611). [Neither](https://ubuntu.com/about/release-cycle) of [these](https://wiki.centos.org/About/Product) are EOL for a few more years.

```
 = note: /target/x86_64-unknown-linux-gnu/debug/deps/libnix-0e1e3ef93965abb0.rlib(nix-0e1e3ef93965abb0.nix.194bd3d9-cgu.7.rcgu.o): In function `nix::sys::memfd::memfd_create::h8f53f62aa4e9fe80':
          /cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.26.2/src/sys/memfd.rs:56: undefined reference to `memfd_create'
          collect2: error: ld returned 1 exit status
```

This change pulls out using `libc::memfd_create` for gnu and musl. I believe the original commit was intended just for freebsd, so the original intent of that commit remains unchanged.

### DIFF
--- a/src/sys/memfd.rs
+++ b/src/sys/memfd.rs
@@ -48,9 +48,6 @@ pub fn memfd_create(name: &CStr, flags: MemFdCreateFlag) -> Result<OwnedFd> {
                 not(target_os = "android"),
                 any(
                     target_os = "freebsd",
-                    // If the OS is Linux, gnu and musl expose a memfd_create symbol but not uclibc
-                    target_env = "gnu",
-                    target_env = "musl",
                 )))]
             {
                 libc::memfd_create(name.as_ptr(), flags.bits())


asomers:
I don't want to hold back development for the sake of six year old distros.  The correct solution is to figure out the righth way to enable as-needed in the linker.

jszwedko:
> I don't want to hold back development for the sake of six year old distros.

That is completely reasonable. Do you have a minimum glibc that you are targeting? That would help with setting expectations. With https://github.com/vectordotdev/vector, we have users deploying on systems with older libcs (in particular CentOS 7) and so have a desire to maintain compatibility with them.

asomers:
No, we don't target a minimum glibc.  Have you checked to see what, if any, minimum glibc the Rust standard library targets?

asomers:
Actually, I don't think glibc is really the problem.  The problem is more likely the old linker that you're using.

jszwedko:
> No, we don't target a minimum glibc. Have you checked to see what, if any, minimum glibc the Rust standard library targets?

It looks like 2.17, which matches what we are targeting: https://github.com/rust-lang/rust/blob/7d782b7ff4d57170e110211565209ecc5bbb3907/src/doc/rustc/src/platform-support.md?plain=1#L40

nekopsykose:
does `RUSTFLAGS="-Clink-arg=-Wl,--as-needed"` fix it for you, in case it's not the default?

jszwedko:
> does `RUSTFLAGS="-Clink-arg=-Wl,--as-needed"` fix it for you, in case it's not the default?

We did try that, but it didn't seem to make a difference. See https://github.com/nix-rust/nix/issues/1972#issuecomment-1446649236

nekopsykose:
it's `link-arg` (`linkarg` isn't recognised?), but no worries, just a guess :)

asomers:
Superseded by #2049

PR #2009, opened 2023-03-06
Add more detail for ptrace documentation

thomasqueirozb:
All functions are documented with the equivalent C call example except for `read`, `read_user`, `write` & `write_user`.

I was looking for the function that used `PTRACE_PEEKDATA` and couldn't find it using the documentation and had to go to the source code to find it out. In hindsight, it's pretty obvious that it was read... But still I think it would be nice to have it documented


asomers (approved):
thanks! 

bors r+

Issue #786, opened 2017-10-26
Support for wasm and asmjs ?


lilianmoraru:
I try to cross-compile `tantivy` for the web with(I initially tried directly through "source emsdk_env.sh" + cross-compiling - same results but I wanted to mention the projects to make it easier for you to test) [`wargo`](https://github.com/lord/wargo),[2](https://lord.io/blog/2017/wargo/)(wasm) and [`cargo-web`](https://github.com/koute/cargo-web)(asmjs) and it seems that the current blocking point is that `nix` does not compile on those platforms.

It is clear that this is because of some platform-specific `cfg`s(and of course no code for an API that does not map 1-on-1 with the POSIX API):
```
error[E0432]: unresolved import `self::os`
 --> src/features.rs:1:15
  |
1 | pub use self::os::*;
  |               ^^ Could not find `os` in `self`

error[E0432]: unresolved import `self::platform`
   --> src/sys/ioctl/mod.rs:243:15
    |
243 | pub use self::platform::*;
    |               ^^^^^^^^ Could not find `platform` in `self`
```

Do you guys think that it is feasible to make `nix` compatible with these platforms(the Emscripten API is inspired by POSIX)?
If yes, could you please consider adding support for the `asmjs-unknown-emscripten`/`wasm32-unknown-emscripten` targets? :)

Susurrus:
I've wondered about support for nix, considering nix exposes APIs not covered by POSIX I always wondered if the JS crowd would try to compile it. I don't think any of us are against it, but we are also very much a scratch-your-own-itch library and "we" the devs do little active development outside of areas we care about, but do review PRs. We'd definitely would love PRs to support additional platforms, so feel free to file own if you can. If you can't, I fear you'll be waiting a while with this request and almost no one goes through issues that aren't bugs to resolve them.

Susurrus:
Would you be willing to help test and possibly mentor someone to do this? I could label it a mentored issue, and if you wanted to post it to the TWiR section on things people want help with, maybe it could actually be done relatively soon.

PR #2011, opened 2023-03-07
Initial wasi/wasm support for fs, uio, zerocopy

connor4312:
References https://github.com/nix-rust/nix/issues/786 
~~Depends on https://github.com/rust-lang/libc/pull/3143, hence marking as a draft.~~ Done

This implements WASI support sufficient to build the fs, uio, and zerocopy features (enough to build [coreutils](https://github.com/uutils/coreutils) 🙂)

Lots of stuff is flagged out since file permissions aren't present in [`wasi_snapshot_preview1`](https://github.com/WebAssembly/wasi-http-proxy/blob/main/phases/snapshot/witx/wasi_snapshot_preview1.witx) which is what rust supports out of the box for `wasm32-wasi`.

One kind of weird thing is dealing with `PATH_MAX`. WASI doesn't define one, it's whatever the host system is, but afaik there's not an API to get this. So for WASI I defaulted to libc's "linux_like" value of 4096. Let me know what you think.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { path = "../libc", features = [ "extra_traits" ] }
+libc = { git = "https://github.com/rust-lang/libc", rev = "34e9a5e37856ba122791ce6d5f04cec9c7c24f2b", features = [ "extra_traits" ] }
 bitflags = "1.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc", rev = "44cc30c6b68427d3628926868758d35fe561bbe6", features = [ "extra_traits" ] }
+libc = { path = "../libc", features = [ "extra_traits" ] }
```
connor4312, line 30: todo until libc updates

connor4312, line 30: todo until libc updates

connor4312:
Looks like build failure is a flake

atilag:
I guess these changes could also be applied to `target="emscripten"`,  am I right?
Just asking because I'm targeting this platform :)

PR #2013, opened 2023-03-15
Add `getaddrinfo` from POSIX <netdb.h>

Dante-Broggi:
I found myself implementing this and felt I should make a PR.
Most of the doc comments are quotes from the posix documentation.
I am uncertain how the remaining `AddrInfo` fields should be converted to stronger types so I just left comments. 
I also added the name info bit flags struct, but left it unused.

### DIFF
--- a/src/netdb.rs
+++ b/src/netdb.rs
@@ -95,7 +95,8 @@ libc_bitflags!{
 #[repr(i32)]
 #[non_exhaustive]
 pub enum AddressInfoError {
-    // UnknownErrno = 0,
+    /// An error which is unmapped by this enum
+    Unknown = 0,
     /// The name could not be resolved at this time. Future attempts may succeed.
     EAI_AGAIN = libc::EAI_AGAIN,
     /// The flags had an invalid value.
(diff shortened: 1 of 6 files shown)


src/netdb.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+impl Drop for AddrInfoList {
+    fn drop(&mut self) {
+        // SAFETY: getaddrinfo returns an owned list of addrinfo elements.
+        // SAFETY(take_tail): The freeaddrinfo() function shall support the freeing of arbitrary sublists of an addrinfo list originally returned by getaddrinfo().
+        unsafe { libc::freeaddrinfo(self.0.as_ptr().cast()) }
+    }
+}
+/// Corresponds to an iterator for a list of `AddrInfo`.
+/// Deliberately is not Clone because we want to own indirect data.
```
Dante-Broggi, line 146:
```suggestion
```
This was copied from the owning version

asomers:
What advantages does this have over the existing DNS stuff in libstd, and in other crates?

src/netdb.rs
```diff
@@ -0,0 +1,287 @@
...
+        AI_NUMERICHOST;
+        /// Inhibit service name resolution.
+        AI_NUMERICSERV;
+        // despite being in POSIX, libc only has it defined on macos
+        #[cfg(target_os = "macos")]
+        /// If no IPv6 addresses are found,
+        /// query for IPv4 addresses and return them to the caller as IPv4-mapped IPv6 addresses.
+        AI_V4MAPPED;
+        // despite being in POSIX, libc only has it defined on macos
+        #[cfg(target_os = "macos")]
```
pronebird, line 183: this should work fine on iOS too.

pronebird:
> What advantages does this have over the existing DNS stuff in libstd, and in other crates?

Ability to control the DNS resolution mechanism, i.e resolve to IPv4-mapped addresses on IPv6 networks. I am really not sure how it works in default flags but I assume it doesn't work the same way considering that `to_sock_addrs()` zeroes out the hints struct:

https://github.com/rust-lang/rust/blob/9337f7afa6fda07e60d6aa5ee88e692878446782/library/std/src/sys_common/net.rs#L210

PR #2014, opened 2023-03-15
Add `MSG_WAITFORONE` flag

stevenengler:
I'm not sure what platforms this is defined on, so I listed the ones that I think it's defined on. Below I've listed why I think it's defined on each platform.

- android - android src: https://android.googlesource.com/platform/bionic/+/c7ca8f3/libc/include/sys/socket.h#233
- freebsd - man page: https://man.freebsd.org/cgi/man.cgi?query=recvmmsg&apropos=0&sektion=2&manpath=FreeBSD+11-current&format=html
- fuchsia - defined in rust libc: https://github.com/rust-lang/libc/blob/60bf6d7fa9d561820ea562751ee455ccf67d3015/src/fuchsia/mod.rs#L1768
- illumos - I think this is the illumos src?: https://github.com/TritonDataCenter/illumos-joyent/blob/master/usr/src/uts/common/brand/lx/syscall/lx_socket.c
- linux - man page: https://man7.org/linux/man-pages/man2/recvmmsg.2.html
- netbsd - man page: https://man.netbsd.org/recvfrom.2
- openbsd - man page: https://man.openbsd.org/recvmsg.2
- solaris - man page: https://docs.oracle.com/cd/E88353_01/html/E37843/recvmmsg-3c.html

**Edit:** Removed the flag for freebsd, illumos, and openbsd since it isn't defined in the rust libc library for those platforms. The linux musl failure seems to be unrelated.

**Edit 2:** Added a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Removed `flock` from `::nix::fcntl` on Solaris. ([#2082](https://github.com/nix-rust/nix/pull/2082))
 - Use I/O safety with `copy_file_range`, and expose it on FreeBSD.
   (#[1906](https://github.com/nix-rust/nix/pull/1906))
+- Added `MSG_WAITFORONE` to `MsgFlags` on Android, Fuchsia, Linux, NetBSD,
+  FreeBSD, OpenBSD, and Solaris.
+  ([#2014](https://github.com/nix-rust/nix/pull/2014))
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
If some definitions are missing from libc, then you should add them there rather than leave them out here.  That's a very common thing to do when adding new functionality to Nix.

stevenengler:
> If some definitions are missing from libc, then you should add them there rather than leave them out here. That's a very common thing to do when adding new functionality to Nix.

I've added it for more platforms in [libc](https://github.com/rust-lang/libc/pull/3224), and that change is in the latest [libc release](https://github.com/rust-lang/libc/releases/tag/0.2.143).

asomers:
This PR is still failing because MSG_WAITFORONE isn't defined everywhere.  You need to update the libc spec in Cargo.toml.  Also, I think you need to rebase to fix the failure in Linux armv7 uclibceabihf

stevenengler:
This is rebased and all tests now pass.

asomers (approved):

Issue #1298, opened 2020-09-19
does not support sockaddr_in for libc::accept


fzyzcjy:
Hi thanks for the lib! I find the `accept` does not allow us to know the `struct sockaddr_in`.

asomers:
Could you describe your problem in more detail, please?

fzyzcjy:
@asomers Looking at the nix doc and the linux doc, it seems that the nix's `accept` has one less effective argument, thus we cannot get info of `sockaddr_in`. Thanks!

lk16:
## Workaround

For people reading this in the future: I worked around it by calling [`getpeername()`](https://docs.rs/nix/latest/nix/sys/socket/fn.getpeername.html) on the file descriptor. 

We can prevent this call to `getpeername()` and I'll provide a PR if there's any possibility that it would get merged. 

## Description of the problem:

Posix [requires](https://pubs.opengroup.org/onlinepubs/9699919799/functions/accept.html) that the `accept()` function looks like this in C:
```c
#include <sys/socket.h>

int accept(int socket, struct sockaddr *restrict address,
       socklen_t *restrict address_len);
```

So, in C, we can specify the file descriptor of the socket and get the address of the peer using the last two arguments. 

Using Rust and Nix, we can't do that, because Nix sends NULL pointers, which means the C library won't tell us the peer address.

## Suggested change of signature

```rs
// current signature
pub fn accept(sockfd: RawFd) -> Result<RawFd>

// proposed signature
pub fn accept<T: SockaddrLike>(sockfd: RawFd) -> Result<(RawFd, T)>
```

We would keep the API consistent with [`getpeername()`](https://docs.rs/nix/latest/nix/sys/socket/fn.getpeername.html) and other functions in the `socket` crate.

This change is not backwards compatible, but that seems unavoidable to me.

PR #2020, opened 2023-04-15
Make accept() return a (fd, SockaddrLike) tuple

lk16:
This PR addresses #1298.

On my machine 4 tests are failing, I don't understand what's the exact cause. Any suggestions how to fix them would be appreciated.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -2148,10 +2148,22 @@ pub fn bind(fd: RawFd, addr: &dyn SockaddrLike) -> Result<()> {
 /// Accept a connection on a socket
 ///
 /// [Further reading](https://pubs.opengroup.org/onlinepubs/9699919799/functions/accept.html)
-pub fn accept(sockfd: RawFd) -> Result<RawFd> {
-    let res = unsafe { libc::accept(sockfd, ptr::null_mut(), ptr::null_mut()) };
+pub fn accept<T: SockaddrLike>(sockfd: RawFd) -> Result<(RawFd, T)> {
+    let mut addr = mem::MaybeUninit::<T>::uninit();
+    let mut len = T::size();
 
-    Errno::result(res)
+    let res = unsafe {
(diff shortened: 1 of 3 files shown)


asomers (commented):
Does every caller really want the peer's address?  I think not.  Maybe this should be a new function instead of a change to the signature of the original?  That would be backwards-compatible too.

lk16:
> Maybe this should be a new function instead of a change to the signature of the original? That would be backwards-compatible too.

There is already a function that does this, `getpeername()`. But is a workaround, see also my post [here](https://github.com/nix-rust/nix/issues/1298#issuecomment-1509868632).

> Does every caller really want the peer's address? I think not. 

Not every caller will be interested in the peer's address. But in my opinion that should be decided by the caller. Currently Nix making that decision and that doesn't seem right to me.

If it's not worth breaking backwards compatibility over this, then we can add some note to the documentation of `accept()` to suggest `getpeername()` if the caller is interested in the peer's address. I can make PR for that as well.

src/sys/socket/mod.rs
```diff
@@ -2148,10 +2148,22 @@ pub fn bind(fd: RawFd, addr: &dyn SockaddrLike) -> Result<()> {
...
+    let mut len = T::size();
 
-    Errno::result(res)
+    let res = unsafe {
+        libc::accept(sockfd, addr.as_mut_ptr() as *mut sockaddr, &mut len)
+    };
+
+    Errno::result(res)?;
+
+    let addr = unsafe { T::from_raw(addr.assume_init().as_ptr(), Some(len)) };
```
stevenengler, line 2161: I think this is UB since it's not guaranteed that all bytes of `addr` are initialized. A simple example would be if someone passed in a `SockaddrStorage` for an inet socket. The number of bytes written by the syscall would be less than the size of `SockaddrStorage`, so `assume_init()` would create a `SockaddrStorage` from a partially-initialized `SockaddrStorage`.

PR #2021, opened 2023-04-16
Fixing link in rustdoc comments in User::from_name

austinkeeley:
Link text is correct but the URL goes to the wrong function. Probably just a copy/paste issue from the other function.


asomers (approved):
LGTM; thanks.

asomers:
The lint failures should be fixed on master.

bors r+

PR #2027, opened 2023-04-26
Update to bitflags 2.2.1.

qwandor:
This is a new major version and requires some code changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ targets = [
 
 [dependencies]
 libc = { version = "0.2.141", features = ["extra_traits"] }
-bitflags = "1.1"
+bitflags = "2.3.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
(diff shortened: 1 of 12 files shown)


asomers:
What's the motivation?

qwandor:
Hi, thanks for the reply.

For context, I work on Android. We vendor in all the Rust crates that we use (you can see them at https://cs.android.com/android/platform/superproject/+/master:external/rust/crates/ if you're curious), and `nix` is among them. As much as possible, we try to avoid having multiple versions of a crate in the tree, both because of the maintenance overhead and because of the impact on overall binary size.

bitflags has caused a problem, because we now have some crates updated upstream to version 2, while some are still using version 1. We can maintain downstream patches to work around this, but this is more maintenance burden so I'd like if possible to get every crate we care about using bitflags 2 upstream.

asomers (approved):
bors r+

qwandor:
Rebased and fixed some build warnings, please try again.

PR #2029, opened 2023-04-29
Remove 'static mut' usage in features::os::kernel_version.

### LINKED ISSUE #2028: `kernel_version` can cause a data race when called from multiple threads
In `src/features.rs`, `kernel_version()` is defined as:
```rust
    fn kernel_version() -> Result<usize> {
        static mut KERNEL_VERS: usize = 0;

        unsafe {
            if KERNEL_VERS == 0 {
                KERNEL_VERS = parse_kernel_version()?;
            }

            Ok(KERNEL_VERS)
        }
    }
```

If two threads call this function simultaneously when `KERNEL_VERS` is still `0`, then it will result in a data race. A trivial fix is to replace it with an `AtomicUsize` accessed with `Relaxed`, since the worst case is that a thread performs a redundant `uname()` call.

zachs18:
Resolves #2028 

Note that this is (AFAICT) the first use of `Atomic*` types in `nix` (other than tests). However, this shouldn't be a portability issue, since `nix` is not `#![no_std]`, and (IIUC) [`std` requires](https://doc.rust-lang.org/std/sync/atomic/#portability) at least loads and stores of pointer-sized atomics (i.e. `AtomicUsize`), which is all this PR uses.

### DIFF
--- a/src/features.rs
+++ b/src/features.rs
@@ -6,6 +6,7 @@ mod os {
     use crate::sys::utsname::uname;
     use crate::Result;
     use std::os::unix::ffi::OsStrExt;
+    use std::sync::atomic::{AtomicUsize, Ordering};
 
     // Features:
     // * atomic cloexec on socket: 2.6.27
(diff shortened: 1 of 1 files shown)


zachs18:
(see https://github.com/bytecodealliance/rustix/pull/645 about CI failures)

asomers (approved):
Nice catch.  This function dates all the way back to 2014, which may be before the standard library offered stuff like AtomicUsize and Once.

bors r+

PR #2033, opened 2023-05-09
remove unused mut from two variable declarations in sys/socket/mod.rs

inglorion:
This addresses diagnostics like:

  error: variable does not need to be mutable
      --> src/sys/socket/mod.rs:1537:13
       |
  1537 |         let mut p = &mut mmsghdr.msg_hdr;


asomers (approved):
LGTM.  I assume that this is a new warning in the latest nightly compiler.

bors r+

PR #2036, opened 2023-05-11
sockopt: add netlink options

tammela:
Add all of the netlink socket options for Linux.
Documentation for them is very succinct on purpose as `man 7 netlink` does a much better job.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -1059,6 +1059,138 @@ sockopt_impl!(
     bool
 );
 
+// Documented on `man 7 netlink`
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Add socket to group
+    NetlinkAddMembership,
+    SetOnly,
+    libc::SOL_NETLINK,
+    libc::NETLINK_ADD_MEMBERSHIP,
+    libc::c_uint
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Drop socket from group
+    NetlinkDropMembership,
+    SetOnly,
+    libc::SOL_NETLINK,
+    libc::NETLINK_DROP_MEMBERSHIP,
+    libc::c_uint
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Enable packet info control messages
+    NetlinkPacketInfo,
+    Both,
+    libc::SOL_NETLINK,
+    libc::NETLINK_PKTINFO,
+    bool
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Send broadcast errors
+    NetlinkBroadcastError,
+    Both,
+    libc::SOL_NETLINK,
+    libc::NETLINK_BROADCAST_ERROR,
+    bool
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Avoid receiving 'ENOBUFS' errors
+    NetlinkNoEnobufs,
+    Both,
+    libc::SOL_NETLINK,
+    libc::NETLINK_NO_ENOBUFS,
+    bool
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Listen to notifications from all network namespaces
+    NetlinkListenAllNsid,
+    SetOnly,
+    libc::SOL_NETLINK,
+    libc::NETLINK_LISTEN_ALL_NSID,
+    bool
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Trim the extack if needed
+    NetlinkCapAck,
+    Both,
+    libc::SOL_NETLINK,
+    libc::NETLINK_CAP_ACK,
+    bool
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Always send extack
+    NetlinkExtAck,
+    Both,
+    libc::SOL_NETLINK,
+    libc::NETLINK_EXT_ACK,
+    bool
+);
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Get strict check
+    NetlinkGetStrictCheck,
+    Both,
+    libc::SOL_NETLINK,
+    libc::NETLINK_GET_STRICT_CHK,
+    bool
+);
+
+/// List netlink groups
+/// The returned Vec contains a bit set in chunks of u32.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[derive(Copy, Clone, Debug)]
+pub struct NetlinkListMemberships;
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+impl GetSockOpt for NetlinkListMemberships {
+    type Val = Vec<u32>;
+
+    fn get<F: AsFd>(&self, fd: &F) -> Result<Self::Val> {
+        let mut len = 0;
+        // Get membership length
+        let res = unsafe {
+            libc::getsockopt(
+                fd.as_fd().as_raw_fd(),
+                libc::SOL_NETLINK,
+                libc::NETLINK_LIST_MEMBERSHIPS,
+                ::std::ptr::null_mut(),
+                &mut len,
+            )
+        };
+        Errno::result(res)?;
+        loop {
+            let mut val = vec![0u32; (len as usize) / mem::size_of::<u32>()];
+            if !val.is_empty() {
+                let old_len = len;
+                // Get the members
+                let res = unsafe {
+                    libc::getsockopt(
+                        fd.as_fd().as_raw_fd(),
+                        libc::SOL_NETLINK,
+                        libc::NETLINK_LIST_MEMBERSHIPS,
+                        val.as_mut_ptr() as *mut libc::c_void,
+                        &mut len,
+                    )
+                };
+                if old_len != len {
+                    continue;
+                } else {
+                    return Errno::result(res).and(Ok(val));
+                }
+            } else {
+                return Ok(val);
+            }
+        }
+    }
+}
+
 #[allow(missing_docs)]
 // Not documented by Linux!
 #[cfg(any(target_os = "android", target_os = "linux"))]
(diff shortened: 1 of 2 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -1001,6 +1001,126 @@ sockopt_impl!(
...
+                let mut val = vec![0u32; len as usize];
+                let res = libc::getsockopt(
+                    fd,
+                    libc::SOL_NETLINK,
+                    libc::NETLINK_LIST_MEMBERSHIPS,
+                    val.as_mut_ptr() as *mut libc::c_void,
+                    &mut len,
+                );
+                Errno::result(res).and(Ok(val))
+            }
```
asomers, line 1119:
```suggestion
            Errno::result(res)?;
            let mut val = vec![0u32; len as usize];
            let res = libc::getsockopt(
                fd,
                libc::SOL_NETLINK,
                libc::NETLINK_LIST_MEMBERSHIPS,
                val.as_mut_ptr() as *mut libc::c_void,
                &mut len,
            );
            Errno::result(res).and(Ok(val))
```

asomers (changes requested):
Any reason not to enable these for Android too?  Would it be possible to add a test for NetlinkListMemberships ?  And don't forget a CHANGELOG entry.

tammela:
> Any reason not to enable these for Android too? Would it be possible to add a test for NetlinkListMemberships ? And don't forget a CHANGELOG entry.

Will address your suggestions shortly

tammela:
Oh there's actually a bug in this version, let me respin

tammela:
There was actually a kernel bug lingering as well:
https://lore.kernel.org/all/20230525144609.503744-1-pctammela@mojatatu.com/

Double checking the CI failures

tammela:
It seems like the CI failures are caused by the QEMU version.
Looking at the mailing list traffic, the QEMU linux-user target got netlink support by 2019-2020.
I could skip the test if the socket returns operation not supported.
@asomers Any advice on how to proceed?

tammela:
The kernel patch has been merged a while ago:
https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/commit/?id=f4e4534850a9d18c250a93f8d7fbb51310828110

It has been back ported to stable kernels and will hit distros likely soon.

As for the tests failing due to qemu-user not supporting some netlink ops, the test now bails out if a "not supported" errno is returned. These ops were implemented in more recent versions.

@asomers I believe this is good for another round of review, thanks!

src/sys/socket/sockopt.rs
```diff
@@ -1001,6 +1001,133 @@ sockopt_impl!(
...
 
+// Documented on `man 7 netlink`
+#[cfg(any(target_os = "android", target_os = "linux"))]
+sockopt_impl!(
+    /// Add socket to group
+    NetlinkAddMembership,
+    SetOnly,
+    libc::SOL_NETLINK,
+    libc::NETLINK_ADD_MEMBERSHIP,
+    libc::c_uint
```
asomers, line 1012: The man page says that this should be a signed int.  Any reason why you made it unsigned here?  Ditto for NETLINK_DROP_MEMBERSHIP.

src/sys/socket/sockopt.rs
```diff
@@ -1001,6 +1001,133 @@ sockopt_impl!(
...
+
+/// List netlink groups
+/// The returned Vec contains a bit set in chunks of u32.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[derive(Copy, Clone, Debug)]
+pub struct NetlinkListMemberships;
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+impl GetSockOpt for NetlinkListMemberships {
+    type Val = Vec<u32>;
```
asomers, line 1095: Defining this as `Vec` requires consumers to do allocations that they might not otherwise need to.  If you define it as `AsRef<[u8]>`, then that would eliminate an allocation.  see `AlgSetKey`.

src/sys/socket/sockopt.rs
```diff
@@ -1001,6 +1001,133 @@ sockopt_impl!(
...
+            let res = libc::getsockopt(
+                fd,
+                libc::SOL_NETLINK,
+                libc::NETLINK_LIST_MEMBERSHIPS,
+                ::std::ptr::null_mut(),
+                &mut len,
+            );
+            Errno::result(res)?;
+            let mut val = vec![0u32; (len as usize) / mem::size_of::<u32>()];
+            if val.capacity() > 0 {
```
asomers, line 1110: Why check its capacity here?  You just defined it one line above.

src/sys/socket/sockopt.rs
```diff
@@ -1001,6 +1001,133 @@ sockopt_impl!(
...
+                let old_len = len;
+                let res = libc::getsockopt(
+                    fd,
+                    libc::SOL_NETLINK,
+                    libc::NETLINK_LIST_MEMBERSHIPS,
+                    val.as_mut_ptr() as *mut libc::c_void,
+                    &mut len,
```
asomers, line 1117: This is a buffer overrun,  assuming that the `optlen` argument is the size of the array in elements, as the man page suggests.  You're telling `getsockopt` here that the array is actually four times larger than what you created on line 1109.  But you seem to be interpreting `optval` as the size of the array in bytes.  Could you find a better reference to clear up the ambiguity?

src/sys/socket/sockopt.rs
```diff
@@ -1001,6 +1001,133 @@ sockopt_impl!(
...
+                );
+                if old_len != len {
```
asomers, line 1119: This behavior isn't correct.  If `old_len != len`, that probably indicates that the group membership changed after the first call to `getsockopt`.  You shouldn't return an error in that case.  If the group membership has shrunk, then you should adjust the size of the vec accordingly.  If the list has grown, then you should extend the size of the Vec and try again.
Also, you don't need to zero-initialize the vec.  You can do `Vec::with_capacity` and allow `getsockopt` to initialize it.  See `unistd::getgrouplist` for an example of that pattern.

src/sys/socket/sockopt.rs
```diff
@@ -1424,4 +1551,71 @@ mod test {
...
+            SockProtocol::NetlinkRoute,
+        );
+
+        if let Err(Errno::EPROTONOSUPPORT) = s {
+            return;
+        }
+
+        let s = s.unwrap();
+
+        let groups = || -> Result<Vec<u32>> {
```
asomers, line 1574: Why define a closure and then immediately call it?  That's confusing.

src/sys/socket/sockopt.rs
```diff
@@ -1424,4 +1551,71 @@ mod test {
...
+            setsockopt(s, super::NetlinkAddMembership, &27)?;
+            getsockopt(s, super::NetlinkListMemberships)
+        }();
+
+        let groups = match groups {
+            Err(Errno::ENOPROTOOPT) | Err(Errno::EOPNOTSUPP) => return,
+            _ => groups.unwrap(),
+        };
+
+        assert!(groups[0] & (1 << 21) != 0);
```
asomers, line 1585: Shouldn't you also assert that there aren't any other bits set?

asomers (changes requested):
BTW, lots of tests fail due to lack of qemu support.  If you suspect that QEMU is to blame, then you should add
```rust
#[cfg_attr(qemu, ignore)]
``` to the test function.

src/sys/socket/sockopt.rs
tammela: It will be interpreted as an unsigned int since it's a bitfield.
https://elixir.bootlin.com/linux/latest/source/net/netlink/af_netlink.c#L1641
https://elixir.bootlin.com/linux/latest/source/net/netlink/af_netlink.c#L1620

src/sys/socket/sockopt.rs
tammela: How would this work if an array doesn't fit the length reported by the kernel?
Should I return an error?

src/sys/socket/sockopt.rs
tammela: `len` can be 0. If you believe `val.length()` is better suited, let me know!

src/sys/socket/sockopt.rs
tammela: `len` is not used as a size indicator in the kernel, it's only a out pointer.
That's why we allocate for the length first on line 1109.
The `len` returns in bytes. As it's a bitfield it's up to the user to how interpret it (u8, u16, u32...).

https://elixir.bootlin.com/linux/latest/source/net/netlink/af_netlink.c#L1754

There was a bug in this specific routine where it was under reporting the size of the array.
The maintainers argued that reporting in bytes is a better API. This of course affected code that assumed it was returning u32, but it was harmless.

https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/commit/?id=f4e4534850a9d18c250a93f8d7fbb51310828110

https://github.com/systemd/systemd/blob/main/src/libsystemd/sd-netlink/netlink-socket.c#L37

src/sys/socket/sockopt.rs
tammela: > This behavior isn't correct. If `old_len != len`, that probably indicates that the group membership changed after the first call to `getsockopt`. 
Correct.
> You shouldn't return an error in that case. If the group membership has shrunk, then you should adjust the size of the vec accordingly. If the list has grown, then you should extend the size of the Vec and try again.
ACK

src/sys/socket/sockopt.rs
tammela: ACK

src/sys/socket/sockopt.rs
tammela: ACK

src/sys/socket/sockopt.rs
tammela: I updated to `.len`

src/sys/socket/sockopt.rs
tammela: Am looping back when `old_len != len`, so there's extremely low chance for an endless loop if the groups are changed by some buggy code. Should we add a try counter?

src/sys/socket/sockopt.rs
tammela: I didn't understand this comment, so this is still not addressed in the new version

tammela:
@asomers @SteveLauC
Anyone has time for another round?
We are using these changes at work and really want to upstream them

Issue #1954, opened 2022-12-31
sendmmsg() in v0.26 do not support different content in control messages across messages


hunts:
For a UDP server program, it is very useful to use  `sendmmsg()` and `recvmmsg()` to improve the performance by reducing syscalls. 

And it is not a uncommon setup that a single socket can receive IP packets with difference destination addresses. In this type of setup, we obtain the real destination addresses of messages from the `Ipv4PacketInfo`/`Ipv6PacketInfo` control message when using `recvmmsg()`.  Corresponding to that, `Ipv4PacketInfo`/`Ipv6PacketInfo` control messages are set **source address** when sending out multiple messages at once via `sendmmsg()`.

In the breaking change introduced by 0.26 about `sendmmsg()` api, it seems it was assumed that control messages are shared/same across all the messages in the batch, I think this is a false assumption.

Did I read the code wrong? If so, is there any example of how to use the new API correctly in this use case.

hunts:
@pacak any idea on this. If we can confirm this is truly a problem. I can try to make a fix.

pacak:
The comment says that control messages are shared and that's what the code does I'm afraid. I guess `C` could be an iterator of some sort too.

hunts:
@pacak thanks for the confirmation. I'll try to make a fix to allow individual `msg` to have their own cmsg contents.

asomers:
Could you please paste a sample of some code that worked with the old implementation but not with the new?  Maybe we should revert that PR.

pacak:
A working code sample would be great, I can change current code to support old behavior.

vkrasnov:
I don't have a specific code in mind, but for example it is very useful to be able to send multiple messages with a different TxTime message.

hunts:
Sorry guys, I didn't find time to work on this. For an working example, I cannot copy-paste my code, but this is the pseudo code showing the use case:

```rust
type Item = (bytes, peer_addr, local_addr);

let recv_queue: VecDeque<Item> = VecDeque::with_capacity(BUF_LEN);
let send_queue: VecDeque<Item> = VecDeque::with_capacity(BUF_LEN);

// Get per-msg local addr from CMSG of each msg.
let incoming: Vec<Item> = match recvmmsg(sock.as_raw_fd(), &mut data, msg_flags, None) { ... };

... push incoming items to the recv_queue
... processing in another thread

let responses = send_queue.iter().take(BATCH_LEN);

[...]

PR #2038, opened 2023-05-13
sendmmsg now allows to have different control messages for different addresses

pacak:
Fixes https://github.com/nix-rust/nix/issues/1954


asomers (changes requested):
sendmmsg is notoriously complicated.  So complicated, that I really think we need a good example for the docs.  Could you please add one?

pacak:
I can copy the code I wrote for tests I guess, not sure if it makes a good example though since it doesn't use control messages...

PR #2040, opened 2023-05-18
timerfd: Add TFD_TIMER_CANCEL_ON_SET flag

apohrebniak:
Hi.
This PR adds an TFD_TIMER_CANCEL_ON_SET flag to use with `timerfd` on Linux and Android


asomers (commented):
Looks good, but how about a test?

apohrebniak:
Changed the documentation a bit.

> Looks good, but how about a test?
The only way I can think of is to actually tweak a `CLOCK_REALTIME`, which would interfere with other processes while running tests either locally or in CI(except the ones running on a dedicated VM?). Unfortunately,  it's not possible to hide it under `time_namespaces`:
```
       Note that time namespaces do not virtualize the CLOCK_REALTIME
       clock.  Virtualization of this clock was avoided for reasons of
       complexity and overhead within the kernel.
```
Am I missing something? Will happily add a test if so.

asomers (approved):
Yeah, that makes sense.  We certainly don't want to adjust the system clock during these tests.

bors r+

PR #1695, opened 2022-04-10
PTRACE_GETREGSET and PTRACE_SETREGSET basic implementation on aarch64

irgstg:
As I mentioned in https://github.com/nix-rust/nix/issues/1419#issuecomment-1094251694 I had this code implemented in a personal project of mine.

As some of you might know, to read and modify tracee's registers with `ptrace` on `aarch64`, one need to do so by calling `ptrace` via `iovec`. ([man ptrace](https://linux.die.net/man/2/ptrace) under `PTRACE_GETREGSET`)

This current PR supports `aarch64` linux and android and includes:
- Two generic `ptrace` calls to receive and modify data via `iovec`: `ptrace_get_iovec_data` and `ptrace_set_iovec_data`.
- Two pub functions that supports modify and read of tracee's registers: `ptrace_getregset` and `ptrace_setregset`
- Test function, `test_ptrace_regsets`, that receives tracee's registers, modify one of them, and validate if the modification completed successfully. 

I might will modify some minor changes.

I want to get you're special attention about this line from `man ptrace`:
_On return, the kernel modifies iov.len to indicate the actual number of bytes returned._
In my original implemented I didn't add anything to validate that, and I'm not aware what'd this repo's conventions dictate in such occasion, so leave a comment and I will edit the code as necessery.

Please let me know if I forgot or missed something.

PR #2044, opened 2023-05-23
Add implementation of `PTRACE_{GET,SET}REGSET`

hack3ric:
This is related to https://github.com/nix-rust/nix/pull/1695, but have many issues solved here:
- `{get,set}regset` now allows specifying which set of registers to read/write, as defined in `RegisterSet`. This should be the expected behavior.
- Also added `PTRACE_{GET,SET}REGS` for most platforms other than x86 using aforementioned implementation.

### DIFF
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -17,8 +17,10 @@ pub type AddressType = *mut ::libc::c_void;
             target_arch = "x86_64",
             any(target_env = "gnu", target_env = "musl")
         ),
-        all(target_arch = "x86", target_env = "gnu")
-    )
+        all(target_arch = "x86", target_env = "gnu"),
+        all(target_arch = "aarch64", target_env = "gnu"),
+        all(target_arch = "riscv64", target_env = "gnu"),
+    ),
 ))]
 use libc::user_regs_struct;
 
@@ -170,6 +172,29 @@ libc_enum! {
     }
 }
 
+libc_enum! {
+    #[cfg(all(
+        target_os = "linux",
+        target_env = "gnu",
+        any(
+            target_arch = "x86_64",
+            target_arch = "x86",
+            target_arch = "aarch64",
+            target_arch = "riscv64",
+        )
+    ))]
+    #[repr(i32)]
+    /// Defining a specific register set, as used in [`getregset`] and [`setregset`].
+    #[non_exhaustive]
+    pub enum RegisterSet {
+        NT_PRSTATUS,
+        NT_PRFPREG,
+        NT_PRPSINFO,
+        NT_TASKSTRUCT,
+        NT_AUXV,
+    }
+}
+
 libc_bitflags! {
     /// Ptrace options used in conjunction with the PTRACE_SETOPTIONS request.
     /// See `man ptrace` for more details.
@@ -217,6 +242,12 @@ fn ptrace_peek(
 }
 
 /// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`
+///
+/// Note that since `PTRACE_GETREGS` are not available on all platforms (as in [ptrace(2)]),
+/// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
+/// on aarch64 and riscv64.
+///
+/// [ptrace(2)]: https://www.man7.org/linux/man-pages/man2/ptrace.2.html
 #[cfg(all(
     target_os = "linux",
     any(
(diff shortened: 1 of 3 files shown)


src/sys/ptrace/linux.rs
```diff
@@ -213,6 +238,48 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
...
+/// Get user registers, as with `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)`
+#[cfg(all(
+    target_os = "linux",
+    target_env = "gnu",
+    any(
+        target_arch = "aarch64",
+        target_arch = "riscv64",
+    )
+))]
+pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
```
asomers, line 250: This function seems unnecessary and confusing.  If the RegisterSet really needs to be optional, a more idiomatic way to do it would be to be an Option<RegisterSet>.  That would be better than having two versions of the function.

src/sys/ptrace/linux.rs
```diff
@@ -213,6 +238,48 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
...
+    let mut data = mem::MaybeUninit::<user_regs_struct>::uninit();
+    let mut iov = libc::iovec {
+        iov_base: data.as_mut_ptr() as *mut _,
+        iov_len: mem::size_of::<user_regs_struct>(),
+    };
+    unsafe {
+        ptrace_other(
+            request,
+            pid,
+            set as i32 as _,
```
asomers, line 276: Why the `as _`?  That can mask type errors.  Better not to do that.

asomers (changes requested):
Needs a CHANGELOG

hack3ric:
Changelog added.

`test_ptrace` on Linux aarch64 failed while previously passed. Debugging.

hack3ric:
I've rebased this patch to master. Any updates?

SteveLauC:
From my perspective, all the work is blocked by #2166

canoriz:
Any updates?

src/sys/ptrace/linux.rs
SteveLauC, line 250: I wonder can we implement this method using `PTRACE_GETREGS` on these non-x86 arches?

src/sys/ptrace/linux.rs
```diff
@@ -213,6 +238,48 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
...
+        target_arch = "x86",
+        target_arch = "aarch64",
+        target_arch = "riscv64",
+    )
+))]
+pub fn getregset(pid: Pid, set: RegisterSet) -> Result<user_regs_struct> {
+    let request = Request::PTRACE_GETREGSET;
+    let mut data = mem::MaybeUninit::<user_regs_struct>::uninit();
+    let mut iov = libc::iovec {
+        iov_base: data.as_mut_ptr() as *mut _,
```
SteveLauC, line 269: For pointer casting, [we prefer `.cast()` over the `as` keyword](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#pointer-type-casting):

```suggestion
        iov_base: data.as_mut_ptr().cast(),
```

src/sys/ptrace/linux.rs
```diff
@@ -213,6 +238,48 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
...
+        iov_len: mem::size_of::<user_regs_struct>(),
+    };
+    unsafe {
+        ptrace_other(
+            request,
+            pid,
+            set as i32 as AddressType,
+            &mut iov as *mut _ as *mut c_void,
```
SteveLauC, line 277: Same here:
```suggestion
            (&mut iov as *mut _).cast(),
```

src/sys/ptrace/linux.rs
```diff
@@ -236,6 +303,50 @@ pub fn setregs(pid: Pid, regs: user_regs_struct) -> Result<()> {
...
+        target_arch = "riscv64",
+    )
+))]
+pub fn setregset(
+    pid: Pid,
+    set: RegisterSet,
+    regs: user_regs_struct,
+) -> Result<()> {
+    let iov = libc::iovec {
+        iov_base: &regs as *const _ as *mut c_void,
```
SteveLauC, line 336: Same here:

```suggestion
        (iov_base: &regs as *const _).cast(),
```

src/sys/ptrace/linux.rs
```diff
@@ -236,6 +303,50 @@ pub fn setregs(pid: Pid, regs: user_regs_struct) -> Result<()> {
...
+        iov_len: mem::size_of::<user_regs_struct>(),
+    };
+    unsafe {
+        ptrace_other(
+            Request::PTRACE_SETREGSET,
+            pid,
+            set as i32 as AddressType,
+            &iov as *const _ as *mut c_void,
```
SteveLauC, line 344: Same here:

```suggestion
            (&iov as *const _).cast().cast_mut(),
```

BTW, we are casting an immutable reference to a mutable pointer, is this safe?

SteveLauC (changes requested):
Hi, sorry for the late review, I have left some comments:

And, we switched to a new changelog mode, please see [this](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add one:)

SteveLauC:
BTW, #1695 has a good test, can we port it to this PR?

src/sys/ptrace/linux.rs
hack3ric: Sorry, this is a mistake. Changed those variables to mutable.

hack3ric:
> BTW, #1695 has a good test, can we port it to this PR?

Ported.

hack3ric:
Cannot reproduce aarch64 test errors on real hardware (Radxa Rock 5B, RK3588, rustc 1.70.0, Linux 5.10.110-20-rockchip)

SteveLauC:
> Cannot reproduce aarch64 test errors on real hardware (Radxa Rock 5B, RK3588, rustc 1.70.0, Linux 5.10.110-20-rockchip)

I think I will set up a Linux VM on my m1 Mac to give it a test. BTW, we run Nix's tests with `sudo`, though I am not sure if it is related

src/sys/ptrace/linux.rs
SteveLauC: > In fact, PTRACE_{GET,SET}REGS symbols are missing on aarch64 and riscv64, so we have to use PTRACE_{GET,SET}REGSET to achieve the same effect.

Hi, can we document this in the code (for `getregs` and `setregs`) so that developers who are not familiar with these APIs can understand why we are having 2 different implementations for the same API:)

hack3ric: Done. :)

SteveLauC:
> I think I will set up a Linux VM on my m1 Mac to give it a test. 

Sorry for the late reply, I tested it in a aarch64 Linux VM, and it also failed with `EIO`:

```sh
[steve@fedora steve]$ uname -a
Linux fedora 6.7.11-orbstack-00143-ge6b82e26cd22 #1 SMP Sat Mar 30 12:20:36 UTC 2024 aarch64 GNU/Linux

[steve@fedora test]$ sudo --preserve-env=HOME $(which cargo) test --all-features sys::test_ptrace::test_ptrace_syscall
    Finished test [unoptimized + debuginfo] target(s) in 0.01s
     Running unittests src/lib.rs (/home/steve/Documents/workspace/nix/target/debug/deps/nix-4701d15085df23b3)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 19 filtered out; finished in 0.00s

     Running test/test.rs (/home/steve/Documents/workspace/nix/target/debug/deps/test-637c83b07deea78d)

running 1 test
test sys::test_ptrace::test_ptrace_syscall ... FAILED

failures:

---- sys::test_ptrace::test_ptrace_syscall stdout ----
thread 'sys::test_ptrace::test_ptrace_syscall' panicked at test/sys/test_ptrace.rs:262:59:
called `Result::unwrap()` on an `Err` value: EIO


failures:
    sys::test_ptrace::test_ptrace_syscall

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 347 filtered out; finished in 0.00s

error: test failed, to rerun pass `--test test`
```

Is there anything I can do to help debug the issue?

hack3ric:
It turns out to be my bad. I ran tests on non-root and it just skipped it.

After some research it seems aarch64 and riscv64 does not support PTRACE_PEEKUSER at all and will just fail with EIO. I'll make that part of the test x86-only.

SteveLauC:
The CI is still failing due to an unused import, let me give it a fix:)

SteveLauC (approved):
Thanks, for the PR and your patience:)

PR #2046, opened 2023-05-31
README: Remove Gitter channel info

zeenix:
This channel is not really helping anyone. Nobody is there to help out folks for many months now. Best not to give false hope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,6 @@ changed in the future without bumping the major or minor version.
 Contributions are very welcome.  Please See [CONTRIBUTING](CONTRIBUTING.md) for
 additional details.
 
-Feel free to join us in [the nix-rust/nix](https://gitter.im/nix-rust/nix) channel on Gitter to
-discuss `nix` development.
-
 ## License
 
 Nix is licensed under the MIT license.  See [LICENSE](LICENSE) for more details.


SteveLauC:
> I am sorry about what happened to you, nix was indeed lack of maintanence in the past, in fact, I won't say it is under active development now, but at least, we are moving
>
> For this chat room, I am not sure I will be active there, it might be easier to get maintainers' attention with GitHub issues

Copy my response from the gitter chat room, it would be better to just use the GitHub issue.

-----

For this PR, I am not sure if the link should be removed, but it might be worthwhile documenting that "it might be easier to get maintainers' attention with GitHub issues", would you like to do it in this PR, friendly ping @zeenix

zeenix:
@SteveLauC  Thanks so much for looking into this matter again. Let me also c&p from the chat. :)

> Thanks so much. However, it's not about my message/PR only. People have been asking questions and pinging folks about PRs since Dec 2022 w/o any response. A chat channel where developers don't respond at all for more than 6 months, is not useful at all. The whole point of a chat channel is a chance of quick interactions. People know already about github issues and that's what they typically already use anyway.
> 
> Having said that, if you think you can keep your presence here from now on, I'm more than happy to simply close my PR.
>

PR #2047, opened 2023-06-01
Remove unnecessary static assertion

WhyNotHugo:
This assertion is not required. The docs for std::mem::transmute read:

    Both types must have the same size. Compilation will fail if this is
    not guaranteed.

I have confirmed this to be the case by trying to transmute two different variables of different length:

    error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
      --> src/sys/socket/addr.rs:44:14
       |
    44 |     unsafe { mem::transmute("asdfasd") }
       |              ^^^^^^^^^^^^^^
       |
       = note: source type: `&str` (128 bits)
       = note: target type: `in_addr` (32 bits)

Given that this is the only usage of the `static_assertions` crates, this allows dropping an unnecessary dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ libc = { version = "0.2.141", features = ["extra_traits"] }
 bitflags = "1.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
-static_assertions = "1"
 memoffset = { version = "0.9", optional = true }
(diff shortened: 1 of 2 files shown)


stevenengler:
I'm pretty sure this transmute is actually unsafe. I've opened #2053.

asomers:
Superseded by #2061

PR #2055, opened 2023-06-16
Add support for QNX Neutrino 7.1

flba-eb:
This adds support for QNX Neutrino 7.1.

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -48,6 +48,10 @@ cfg_if! {
         unsafe fn errno_location() -> *mut c_int {
             unsafe { libc::_Errno() }
         }
+    } else if #[cfg(target_os = "nto")] {
+        unsafe fn errno_location() -> *mut c_int {
+            unsafe { libc::__get_errno_ptr() }
+        }
     }
 }
 
@@ -603,10 +607,20 @@ fn desc(errno: Errno) -> &'static str {
         #[cfg(freebsdlike)]
         EDOOFUS => "Programming error",
 
-        #[cfg(any(freebsdlike, target_os = "hurd", target_os = "redox"))]
+        #[cfg(any(
+            freebsdlike,
+            target_os = "hurd",
+            target_os = "redox",
+            target_os = "nto",
+        ))]
         EMULTIHOP => "Multihop attempted",
 
-        #[cfg(any(freebsdlike, target_os = "hurd", target_os = "redox"))]
+        #[cfg(any(
+            freebsdlike,
+            target_os = "hurd",
+            target_os = "redox",
+            target_os = "nto",
+        ))]
         ENOLINK => "Link has been severed",
 
         #[cfg(target_os = "freebsd")]
@@ -615,7 +629,7 @@ fn desc(errno: Errno) -> &'static str {
         #[cfg(target_os = "freebsd")]
         ECAPMODE => "Not permitted in capability mode",
 
-        #[cfg(any(bsd, target_os = "hurd"))]
+        #[cfg(all(not(target_os = "nto"), any(bsd, target_os = "hurd")))]
         ENEEDAUTH => "Need authenticator",
 
         #[cfg(any(bsd, target_os = "redox", solarish))]
@@ -627,17 +641,19 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "netbsd",
             target_os = "redox",
             target_os = "haiku",
+            target_os = "nto",
             target_os = "hurd"
         ))]
         EILSEQ => "Illegal byte sequence",
 
-        #[cfg(any(bsd, target_os = "haiku"))]
+        #[cfg(all(not(target_os = "nto"), any(bsd, target_os = "haiku")))]
         ENOATTR => "Attribute not found",
 
         #[cfg(any(
             bsd,
             target_os = "redox",
             target_os = "haiku",
+            target_os = "nto",
             target_os = "hurd"
         ))]
         EBADMSG => "Bad message",
@@ -670,7 +686,10 @@ fn desc(errno: Errno) -> &'static str {
         ))]
         ENOTSUP => "Operation not supported",
 
-        #[cfg(any(bsd, target_os = "aix", target_os = "hurd"))]
+        #[cfg(all(
+            not(target_os = "nto"),
+            any(bsd, target_os = "aix", target_os = "hurd")
+        ))]
         EPROCLIM => "Too many processes",
 
         #[cfg(any(
@@ -718,10 +737,10 @@ fn desc(errno: Errno) -> &'static str {
         #[cfg(any(bsd, target_os = "hurd"))]
         EPROCUNAVAIL => "Bad procedure for program",
 
-        #[cfg(any(bsd, target_os = "hurd"))]
+        #[cfg(all(not(target_os = "nto"), any(bsd, target_os = "hurd")))]
         EFTYPE => "Inappropriate file type or format",
 
-        #[cfg(any(bsd, target_os = "hurd"))]
+        #[cfg(all(not(target_os = "nto"), any(bsd, target_os = "hurd")))]
         EAUTH => "Authentication error",
 
         #[cfg(any(
@@ -757,7 +776,8 @@ fn desc(errno: Errno) -> &'static str {
             apple_targets,
             target_os = "aix",
             target_os = "netbsd",
-            target_os = "redox"
+            target_os = "redox",
+            target_os = "nto",
         ))]
         ENODATA => "No message available on STREAM",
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -89,13 +89,15 @@ The following targets are supported by `nix`:
     <li>x86_64-unknown-netbsd</li>
    </td>
    <td>
+    <li>aarch64-unknown-nto-qnx710</li>
```
asomers, line 92: You shouldn't add QNX to this table because there's no CI for it.

src/errno.rs
```diff
@@ -3131,3 +3156,219 @@ mod consts {
...
+    #[deprecated(
+        since = "0.22.1",
+        note = "use nix::errno::Errno::ELAST instead"
+    )]
+    pub const ELAST: Errno = Errno::ENOTSUP;
+    #[deprecated(
+        since = "0.22.1",
+        note = "use nix::errno::Errno::EWOULDBLOCK instead"
+    )]
+    pub const EWOULDBLOCK: Errno = Errno::EAGAIN;
```
asomers, line 3269: There's no need to add code if it will be immediately deprecated.
```suggestion
```

src/fcntl.rs
```diff
@@ -297,12 +304,12 @@ fn readlink_maybe_at<P: ?Sized + NixPath>(
                 cstr.as_ptr(),
                 v.as_mut_ptr() as *mut c_char,
                 v.capacity() as size_t,
-            ),
+            ) as _,
```
asomers, line 307: `as _` can mask a wide range of type errors.  Could you please not do that?

src/sys/resource.rs
```diff
@@ -167,7 +171,7 @@ libc_enum! {
         /// or used by all of this user id's processes.
         RLIMIT_SWAP,
 
-        #[cfg(target_os = "freebsd")]
+        #[cfg(any(target_os = "freebsd"))]
```
asomers, line 174: You should revert this change.  It's not useful, and I think Clippy will complain about it.

src/sys/select.rs
```diff
@@ -307,11 +307,25 @@ where
...
-    let timeout = timeout.map(|ts| ts.as_ref() as *const libc::timespec).unwrap_or(null());
     let sigmask = sigmask.map(|sm| sm.as_ref() as *const libc::sigset_t).unwrap_or(null());
 
-    let res = unsafe {
-        libc::pselect(nfds, readfds, writefds, errorfds, timeout, sigmask)
+    #[cfg(target_os = "nto")]
+    let res = match timeout {
+        Some(t) => {
+            let mut timeout = t.clone();
+            unsafe {libc::pselect(nfds, readfds, writefds, errorfds, timeout.as_mut() as *mut libc::timespec, sigmask)}
```
asomers, line 316: This looks like a bug in Neutrino.  POSIX says that the timeout argument should be const.  Frankly, given how complicated this makes Nix's code, I don't think it's worth it for the benefit this provides.  I would rather just not bind pselect on this platform.

asomers (changes requested):

src/fcntl.rs
flba-eb: Unfortunately, [readlink](https://www.qnx.com/developers/docs/7.1/#com.qnx.doc.neutrino.lib_ref/topic/r/readlink.html) returns an `int` on QNX Neutrino. Any suggestion on how to resolve this (maybe one that is also applicable to `src/sys/select.rs`, see https://github.com/nix-rust/nix/pull/2055/files#r1265967722)?

One possibility (similar to the existing redox handling could be):
```rust
    path.with_nix_path(|cstr| unsafe {
        match dirfd {
            #[cfg(target_os = "redox")]
            Some(_) => unreachable!(),
            #[cfg(target_os = "nto")]
            Some(dirfd) => libc::readlinkat(
                dirfd,
                cstr.as_ptr(),
                v.as_mut_ptr() as *mut c_char,
                v.capacity() as size_t,
            ) as libc::ssize_t,
            #[cfg(not(any(target_os = "redox", target_os = "nto")))]
            Some(dirfd) => libc::readlinkat(
                dirfd,
                cstr.as_ptr(),
                v.as_mut_ptr() as *mut c_char,
                v.capacity() as size_t,
            ),
            #[cfg(target_os = "nto")]
            None => libc::readlink(
                cstr.as_ptr(),
                v.as_mut_ptr() as *mut c_char,
                v.capacity() as size_t,
            ) as libc::ssize_t,
            #[cfg(not(target_os = "nto"))]
            None => libc::readlink(
                cstr.as_ptr(),
                v.as_mut_ptr() as *mut c_char,
                v.capacity() as size_t,
            ),
        }
    })
```

But this makes the code a bit hard to read. Would a small OS specific wrapper function help?

README.md
flba-eb: Both nto items have been removed.

src/sys/resource.rs
flba-eb: Sorry for this one, it was not my intention to push it. Its reverted.

src/sys/select.rs
flba-eb: As also described in my [previous comment](https://github.com/nix-rust/nix/pull/2055#discussion_r1279161236), there are API differences that I'm trying to work around as I cannot change the libc API. Would a small wrapper function help (see other comment)?

asomers: We could make it work, but it would definitely ugly-up the code.  And this const-mut conversion is worse than the size widening of readlink.  Do we know that there are even any QNX users who want to use pselect?

src/fcntl.rs
asomers: Either of those options would work, but I kind of like the first one better.

AkhilTThomas:
@flba-eb Hey! Is this PR stuck? would like to have nto support for nix.

SteveLauC:
Hi @AkhilTThomas, please see this comment: https://github.com/tokio-rs/tokio/pull/6421#issuecomment-2028053903

AkhilTThomas:
> Hi @AkhilTThomas, please see this comment: [tokio-rs/tokio#6421 (comment)](https://github.com/tokio-rs/tokio/pull/6421#issuecomment-2028053903)

Awesome thanks!

flba-eb:
> @flba-eb Hey! Is this PR stuck? would like to have nto support for nix.

Adding QNX support to nix is still on my wish-list -- but I'm currently not needing nix in my projects, which is why it didn't get much attention.
I will have more time after the Embedded World (in about two weeks) and can continue, but feel free to "take over" this PR. Maybe it makes sense to fix this together?

AkhilTThomas:
> > @flba-eb Hey! Is this PR stuck? would like to have nto support for nix.
> 
> Adding QNX support to nix is still on my wish-list -- but I'm currently not needing nix in my projects, which is why it didn't get much attention. I will have more time after the Embedded World (in about two weeks) and can continue, but feel free to "take over" this PR. Maybe it makes sense to fix this together?

That would be great! I will contact you over email.

src/fcntl.rs
flba-eb: Should be fixed, please check.

src/sys/select.rs
flba-eb: @AkhilTThomas what do you think about removing pselect or do you know if it is used?

flba-eb:
Some tests are failing, does anybody have any idea why? On my machine other tests fail also with revision 4ab23c32 (current head on master).

AkhilTThomas:
> Some tests are failing, does anybody have any idea why? On my machine other tests fail also with revision [4ab23c3](https://github.com/nix-rust/nix/commit/4ab23c327889ffb4f57dfec883a9c3c596953949) (current head on master).
Can you try this patch file for some of the test cases. With this only 4 tests fail as of now
`test result: FAILED. 116 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.42s`
[nix-nto.patch](https://github.com/nix-rust/nix/files/14816096/nix-nto.patch)

src/sys/socket/mod.rs
```diff
@@ -884,11 +925,12 @@ impl ControlMessageOwned {
                 let cred: libc::cmsgcred = unsafe { ptr::read_unaligned(p as *const _) };
                 ControlMessageOwned::ScmCreds(cred.into())
             }
+            #[cfg(target_os = "nto")]
+            (libc::SOL_SOCKET, libc::SCM_CREDS) => {
+                let cred: libc::sockcred = ptr::read_unaligned(p as *const _);
+                ControlMessageOwned::ScmCreds(cred.into())
+            }
             #[cfg(not(any(target_os = "aix", target_os = "haiku")))]
-            (libc::SOL_SOCKET, libc::SCM_TIMESTAMP) => {
```
yuzhe-wu, line 888: Not quite sure if this is on purpose, but it breaks my Linux case. (We have same code base for both Linux and QNX)

flba-eb:
I'm closing this PR as I currently neither have the time to work on it nor the need for nix.

PR #2057, opened 2023-06-17
Haiku: `speed_t` is defined as `u8` for 32 and 64 bit systems

nielx:
This fixes the build on 32 bit Haiku systems.


asomers (commented):
As 32-bit Haiku is not an officially supported OS, I'll have to take your word for it.

bors r+

PR #2065, opened 2023-06-29
Parse SOL_TLS control message, closes #2064

### LINKED ISSUE #2064: ControlMessage: SOL_TLS control message falls back to ControlMessageOwned::Unknown
When enabling kTLS on a socket (which mostly involves [setting up the SOL_TLS sockopt](https://github.com/hapsoc/ktls/blob/c6d79c1f1f084ab52937723f1bbb88e4d503abee/src/ffi.rs#L184-L192)), that socket may receive some control messages like so:

```
  2023-06-29T11:56:33.154070Z TRACE ktls::ktls_stream: received UnknownCmsg(
    cmsghdr {
        cmsg_len: 17,
        cmsg_level: 282,
        cmsg_type: 2,
    },
    [
        23,
    ],
)
```

The cmsg_level corresponds to `libc::SOL_TLS`, cmsg_type corresponds to [TLS_GET_RECORD_TYPE](https://github.com/hapsoc/ktls-sys/blob/16149163e208dfbdd8789c40633348902bd6f991/src/bindings.rs#L80)

I believe the additional data in the `Vec<u8>` corresponds to TLS message types: 22 is handshake, 23 is application data, see [RFC 5246](https://datatracker.ietf.org/doc/html/rfc5246#section-6.2.1).

This happens because, after the TLS 1.3 handshake, tickets can be sent by the server to the client (for resumption, see [rustls 0.21's flag for that](https://docs.rs/rustls/0.21.2/rustls/server/struct.ServerConfig.html#structfield.send_tls13_tickets)).

If you're doing userland TLS, your deframer handles those messages gracefully and either throws away or stores the tickets for further use. But with kTLS, they have to go _somewhere_, and.. if you've received a ticket and you do a `read()` on the socket, you get [errno 5](https://github.com/hapsoc/ktls/issues/24) (helpful, I know!). If you do a `recvmsg()` you get one of these with the "bonus data" set to 22, which again, is the TLS 1.2/1.3 "handshake" type.

I'm working on a PR that recognizes those message types, adding a variant to `ControlMessageOwned`. The diff is probably a handful of lines and y'all don't need to bother yourselves with those details anyway but since there's fuck-all documentation about kTLS offline I needed to put it down _somewhere_.

fasterthanlime:
cf. #2064 - I've added the absolute minimum I need. We could swap that u8 with something more strongly-typed, but since I'm clearly the first person using ktls & the nix crate together, I'm treading lightly in terms of added lines.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -894,6 +894,7 @@ pub struct Timestamps {
 #[cfg(all(target_os = "linux"))]
 const TLS_GET_RECORD_TYPE: c_int = 2;
 
+#[cfg(all(target_os = "linux"))]
 const SOL_TLS: c_int = 282;
 
 impl ControlMessageOwned {
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1032,7 +1032,7 @@ impl ControlMessageOwned {
                 let dl = ptr::read_unaligned(p as *const libc::sockaddr_in6);
                 ControlMessageOwned::Ipv6OrigDstAddr(dl)
             },
-            #[cfg(all(target_os = "linux"))]
+            #[cfg(any(target_os = "android", target_os = "linux"))]
             (libc::SOL_TLS, libc::TLS_GET_RECORD_TYPE) => {
                 let content_type = ptr::read_unaligned(p as *const u8);
                 ControlMessageOwned::TlsGetRecordType(content_type)
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -868,6 +868,11 @@ pub enum ControlMessageOwned {
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
     Ipv6RecvErr(libc::sock_extended_err, Option<sockaddr_in6>),
 
+    /// `SOL_TLS` messages of type `TLS_GET_RECORD_TYPE`, containing the TLS message content type,
+    /// normally one of change_cipher_spec(20), alert(21), handshake(22) (for TLS 1.3
+    /// resumption tickets), application_data(23)
+    TlsGetRecordType(u8),
+
     /// Catch-all variant for unimplemented cmsg types.
     #[doc(hidden)]
     Unknown(UnknownCmsg),
@@ -884,6 +889,12 @@ pub struct Timestamps {
     /// hardware based timestamp
     pub hw_raw: TimeSpec,
 }
+
+// Defined in `linux/tls.h`
+#[cfg(all(target_os = "linux"))]
+const TLS_GET_RECORD_TYPE: c_int = 2;
+
+const SOL_TLS: c_int = 282;
 
 impl ControlMessageOwned {
     /// Decodes a `ControlMessageOwned` from raw bytes.
@@ -1026,6 +1037,11 @@ impl ControlMessageOwned {
             (libc::IPPROTO_IPV6, libc::IPV6_ORIGDSTADDR) => {
                 let dl = ptr::read_unaligned(p as *const libc::sockaddr_in6);
                 ControlMessageOwned::Ipv6OrigDstAddr(dl)
+            },
+            #[cfg(all(target_os = "linux"))]
+            (SOL_TLS, TLS_GET_RECORD_TYPE) => {
+                let content_type = ptr::read_unaligned(p as *const u8);
+                ControlMessageOwned::TlsGetRecordType(content_type)
             },
             (_, _) => {
                 let sl = std::slice::from_raw_parts(p, len);
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -890,13 +890,6 @@ pub struct Timestamps {
     pub hw_raw: TimeSpec,
 }
 
-// Defined in `linux/tls.h`
-#[cfg(all(target_os = "linux"))]
-const TLS_GET_RECORD_TYPE: c_int = 2;
-
-#[cfg(all(target_os = "linux"))]
-const SOL_TLS: c_int = 282;
-
 impl ControlMessageOwned {
     /// Decodes a `ControlMessageOwned` from raw bytes.
     ///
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -846,6 +851,13 @@ pub struct Timestamps {
     pub hw_raw: TimeSpec,
 }
 
+// Defined in `linux/tls.h`
+#[cfg(all(target_os = "linux"))]
+const TLS_GET_RECORD_TYPE: c_int = 2;
```
asomers, line 856: This definition should be pushed into the libc crate first, and then you should update Nix to use the new libc verison.  It isn't Nix's place to define C APIs.  Also, do you know if this same interface is defined on Android?  If so, include that in the `#[cfg]` directive.

src/sys/socket/mod.rs
```diff
@@ -846,6 +851,13 @@ pub struct Timestamps {
...
+
+#[cfg(all(target_os = "linux"))]
+const SOL_TLS: c_int = 282;
```
asomers, line 859: This one should also go to libc.

src/sys/socket/mod.rs
```diff
@@ -829,6 +829,11 @@ pub enum ControlMessageOwned {
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
     Ipv6RecvErr(libc::sock_extended_err, Option<sockaddr_in6>),
 
+    /// `SOL_TLS` messages of type `TLS_GET_RECORD_TYPE`, containing the TLS message content type,
+    /// normally one of change_cipher_spec(20), alert(21), handshake(22) (for TLS 1.3
+    /// resumption tickets), application_data(23)
+    TlsGetRecordType(u8),
```
asomers, line 835: Should this be an enum type?

asomers (changes requested):
This looks like a cool feature.  I've long wanted to play with ktls myself (though in my case, the FreeBSD version), but haven't found an excuse.  But before we can accept this, you need to move all of the C-API definitions into the libc crate.  And you should also add a CHANGELOG entry.

src/sys/socket/mod.rs
fasterthanlime: I briefly considered it but, for it to be future-proof it'd need an "Unknown" variant. I don't mind it happening, but I feel like there'd be _many_ back-and-forths on the PR until everyone's happy with it.

src/sys/socket/mod.rs
fasterthanlime: done!

src/sys/socket/mod.rs
fasterthanlime: also done.

fasterthanlime:
@asomers twelve years later, the requisite changes have been merged into libc (and released to crates.io), so this PR now bumps the dependency and uses the constants from there.

fasterthanlime:
@asomers is there anything I can do to help move this along? https://github.com/hapsoc/ktls/issues/24 has been blocked on it for a while (..mostly while waiting for libc to merge+release, but the ball's in nix's court now).

fasterthanlime:
Update: I'm maintaining a fork of _just this bit_ of nix now: https://crates.io/crates/ktls-recvmsg

(I'd like to deprecate that fork eventually)

src/sys/socket/mod.rs
asomers: I don't think this is good enough.  Matching against magic numbers isn't Rusty at all.  I think you should add the various types, including an Unknown variant.

asomers (changes requested):

src/sys/socket/mod.rs
```diff
@@ -1015,6 +1020,11 @@ impl ControlMessageOwned {
                 let dl = ptr::read_unaligned(p as *const libc::sockaddr_in6);
                 ControlMessageOwned::Ipv6OrigDstAddr(dl)
             },
+            #[cfg(all(target_os = "linux"))]
```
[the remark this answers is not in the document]
fasterthanlime: Fine, but be warned I have not tested it on Android at all nor do I plan on maintaining/supporting that platform personally. My general opinion is that it's better to wait for someone who needs it, and then they can be a point of contact re: getting it working on that platform. That person isn't me for Android.

src/sys/socket/mod.rs
fasterthanlime: As you wish — there's now an enum with the variants I know/care about.

src/sys/socket/mod.rs
```diff
@@ -885,6 +888,33 @@ pub struct Timestamps {
     pub hw_raw: TimeSpec,
 }
 
+/// These constants correspond to TLS 1.2 message types, as defined in
+/// RFC 5246, Appendix A.1
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(u8)]
+#[non_exhaustive]
+pub enum TlsGetRecordType {
```
Arnavion, line 897: FWIW TLS 1.3 has a lot more of these. In particular you're missing SessionTicket(4), which is an almost-guaranteed message you'll receive when doing a TLS 1.3 connection as a post-handshake message.

[This](https://github.com/arsing/ktls-test/blob/10316091b43eece021238c9602ee312d5a886d1d/src/tls.rs#L42-L78) is the full list.

That said, and I know @asomers asked for it, but I don't think this belongs in the nix crate because it's about TLS, not libc or Linux.

fasterthanlime: fwiw, re: your original comment:

> That said, and I know @asomers asked for it, but I don't think this belongs in the nix crate because it's about TLS, not libc or Linux.

I agree with this, but I'm willing to do whatever to get this PR merged at this point. I agree the API boundaries are kinda unclear right now, after all libc itself is a grab bag 🤷

src/sys/socket/mod.rs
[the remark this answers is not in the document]
fasterthanlime: I had to revert this, since the libc constants for this are not available on android (my fault, since I did that PR too, but I'm not willing to block this PR on this).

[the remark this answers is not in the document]
asomers: Oops.  I thought the constant _was_ available on Android in libc.  I guess I grepped wrong.  Sorry.

asomers (changes requested):
The enum looks good.  But don't forget to add a changelog entry.

SteveLauC:
Please see [`CONTRIBUTING.md`](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add changelog entries

SteveLauC (approved):

SteveLauC:
Thanks!

PR #2067, opened 2023-07-02
Address seeming typo in lib.rs

solidsnack:
I believe the `my` here was intended to be `many` -- but am not sure.


asomers (approved):
Yep.  Good catch.

bors r+

PR #2068, opened 2023-07-07
Switch to SockProtocolInt

### LINKED ISSUE #1903: SockProtocol: No support for redundant protocol values from different families
I wanted to send in a PR to add the CANbus socket protocols to `SockProtocol`, to include `CanRaw` (1) and `CanBcm` (2).

Like this:
```
    /// The Controller Area Network raw socket protocol
    /// ([ref](https://docs.kernel.org/networking/can.html#how-to-use-socketcan))
    CanRaw = libc::CAN_RAW,
    /// The Controller Area Network broadcast manager protocol
    /// ([ref](https://docs.kernel.org/networking/can.html#how-to-use-socketcan))
    CanBcm = libc::CAN_BCM,
```

The problem is that CAN_BCM = 2, which is already used by the discriminant, `NetlinkUserSock = libc::NETLINK_USERSOCK`.

Should `SockProtocol` be converted into an enum of enum's for each protocol family?

Dev380:
Fixes #1903 
Currently, new/unknown protocols cannot be implemented with Unix sockets and this crate, because every protocol integer must be part of the SockProtocol enum. Additionally, some protocols, such as raw IPv4 with AF_PACKET, must hack around this by using an enum member with the same protocol number but entirely different name, leading to confusion.

This PR creates SockProtocolInt, which is a transparent struct wrapping an i32, and uses it in the library for everything. From<SockProtocol> is implemented so that no code doesn't break and it remains semver compatible.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -120,8 +120,40 @@ impl TryFrom<i32> for SockType {
     }
 }
 
+/// Specify a SockProtocol with a cint (i32) instead of an enum.
+/// The enum is kept for comptability with older versions of nix.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct SockProtocolInt(pub i32);
+
+impl SockProtocolInt {
+    /// The Controller Area Network broadcast manager protocol
+    /// ([ref](https://docs.kernel.org/networking/can.html#how-to-use-socketcan))
+    #[cfg(target_os = "linux")]
+    #[cfg_attr(docsrs, doc(cfg(all())))]
+    #[allow(non_upper_case_globals)]
+    pub const CanBcm: SockProtocol = SockProtocol::NetlinkUserSock; // Matches libc::CAN_BCM
+}
+
+/// Automatic conversion from SockProtocol to SockProtocolInt for backward compatability.
+impl From<SockProtocol> for SockProtocolInt {
+    fn from(value: SockProtocol) -> Self {
+        Self(value as i32)
+    }
+}
+
+/// Automatic conversion from SockProtocol to Option<SockProtocolInt> for backward compatability.
+/// This is necessary because the functions that use SockProtocolInt are generic over From<Option<SockProtocolInt>>
+/// and not From<SockProtocolInt>.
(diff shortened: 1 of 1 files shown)


Dev380:
I'm not sure why this errors
```
error: casting raw pointers to the same type and constness is unnecessary (`*const libc::sockaddr` -> `*const libc::sockaddr`)
    --> src/sys/socket/mod.rs:2261:17
     |
2261 |                 addr.assume_init().as_ptr() as *const sockaddr,
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `addr.assume_init().as_ptr()`
     |
```
AFAIK my changes haven't affected the socket address type.

asomers:
This is a step backwards for ergonomics, by exposing the raw int values.  If there's a specific protocol value that's not available now, then you should add it like how was done for `CanBcm`.  Also, the Clippy warnings are fixed on the master branch.

nat-rix:
Would it be better to have something like this?

```rust
fn socket<T: IntoSockProtocol>(/*...*/ protocol: Option<T>) -> Result<RawFd>;

trait IntoSockProtocol {
    fn protocol(&self) -> c_int;
}

impl IntoSockProtocol for SockProtocol {
    fn protocol(&self) -> c_int { *self as _ }
}

impl IntoSockProtocol for c_int {
    fn protocol(&self) -> c_int { *self }
}
```

I'm not sure what you mean by "step backwards for ergonomics".
The usage of `socket` doesn't change in almost all use cases (`socket(..., Some(SockProtocol::Tcp))`).
The only ergonomics problem I see would be the documentation, where
`T: Into<Option<SockProtocolInt>>` doesn't direct the user to the preferred
`SockProtocol`.
Making a PR for every possible protocol is also not an option for every user,
because there could be custom protocols specific to some custom OSes.

asomers:
Yeah, actually, I do like that version better.  What I don't want to do is to abandon type safety in the name of flexibility.  However, instead of `impl IntoSockProtocol for c_int`, could you use a NewType instead?  Just some syntactic salt to make the reader aware that the code is sending custom data to the OS, outside of the Rust type system.  Something like `struct CustomSockProtocol(pub c_int); impl IntoSockProtocol for CustomSockProtocol{...}`.

PR #2072, opened 2023-07-13
Fix setting sockopt::IpMulticastTtl

sgasse:
The socket option `IpMulticastTtl` was implemented with `u8` as type however `i32` seems to be the correct type.

### DIFF
--- a/changelog/2072.fixed.md
+++ b/changelog/2072.fixed.md
@@ -1,1 +1,1 @@
-Fixed `::sys::socket::sockopt::IpMulticastTtl` by fixing the value of optlen passed to `libc::setsockopt`.
+Fixed `::sys::socket::sockopt::IpMulticastTtl` by fixing the value of optlen passed to `libc::setsockopt` and added tests.
--- a/changelog/2072.fixed.md
+++ b/changelog/2072.fixed.md
@@ -1,0 +1,1 @@
+Fixed `::sys::socket::sockopt::IpMulticastTtl` by fixing the value of optlen passed to `libc::setsockopt`.
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -1247,7 +1247,7 @@ impl<'a> Set<'a, bool> for SetBool {
     }
 
     fn ffi_len(&self) -> socklen_t {
-        mem::size_of::<c_int>() as socklen_t
+        mem::size_of_val(&self.val) as socklen_t
     }
 }
 
@@ -1298,7 +1298,7 @@ impl<'a> Set<'a, u8> for SetU8 {
     }
 
     fn ffi_len(&self) -> socklen_t {
-        mem::size_of::<c_int>() as socklen_t
+        mem::size_of_val(&self.val) as socklen_t
     }
 }
 
@@ -1349,7 +1349,7 @@ impl<'a> Set<'a, usize> for SetUsize {
     }
 
     fn ffi_len(&self) -> socklen_t {
-        mem::size_of::<c_int>() as socklen_t
+        mem::size_of_val(&self.val) as socklen_t
     }
 }
 
--- a/test/sys/test_sockopt.rs
+++ b/test/sys/test_sockopt.rs
@@ -303,7 +303,7 @@ fn test_get_mtu() {
 }
 
 #[test]
-#[cfg(any(target_os = "android", target_os = "freebsd", target_os = "linux"))]
+#[cfg(any(linux_android, target_os = "freebsd"))]
 fn test_ttl_opts() {
     let fd4 = socket(
         AddressFamily::Inet,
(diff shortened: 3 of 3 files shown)


asomers:
> The socket option `IpMulticastTtl` was implemented with `u8` as type however `i32` seems to be the correct type.

Why do you say that?  It's documented as being u8 on FreeBSD, at least.

asomers:
Oh, I just saw the more detailed description in the issue.  Please link the issue in your commit message.  And yes, this does seem to be a difference between platforms.  This sockopt was added way back in 2015 with the first batch of sockopt macros, in 979faf097691dc35bb12b91dcde0d804a3d32b79 .  There wasn't good CI at the time, and no tests for that sockopt, so nobody noticed that it was broken.
It would be nice if you could add a test for this one.

sgasse:
Thanks for the context! I pushed and opened the PR first when there was no issue but now I included the issue in the commit message. My initial commit also included a new test `test_multicast_ttl_opts` which fails on `linux` when using `u8`. It tries to set `sockopt::IpMulticastTtl` on both a v4 and v6 socket. Is that OK as a test or do we need something more?

I rebased on the latest `master` and added config guards, using `u8` for all BSD-like systems and `i32` everywhere else. Could you trigger another CI run? Then we can hopefully tell which other systems need `u8` instead of `i32`.

asomers:
@sgasse CI should be automatic.  Maybe Cirrus had a hiccup?  Try force-push again.

sgasse:
Alright I guess it was a hiccup in Github itself - the force-push yesterday did not make it through to the PR for 5mins ^^

After the rebase, some more CI jobs pass but some still have this error on mutable reference not used mutable :thinking: And the test I added fails on FreeBSD both with `u8` and with `libc::c_int` which is really strange.

asomers:
There are three reasons why that test is failing for you.  The first is that it really should be a `u8`, not a `u32`.  Your PR changes that.  But the second is a bug in Nix dating to 979faf097691dc35bb12b91dcde0d804a3d32b79 that has been hitherto unnoticed because of insufficient test coverage.  Even though the sockopt is defined as using `u8`, Nix was passing 4 as its length to libc::setsockopt.  I've pushed a change to fix that.  The third reason is that IP_MULTICAST_TTL doesn't work on IPv6 sockets, at least not on FreeBSD.  You should remove that part from your test.  Also, you may need to rebase to fix CI.

sgasse:
OK thanks for the thorough explanation :pray: I will take a look tomorrow and update.

sgasse:
I took the time now to check it and looks good, your fix solves the problem entirely @asomers and it works with `u8` also on Linux and Android. Therefore, I renamed this PR, reduced my commit to just adding tests, amended your commit with a changelog entry and moved it before mine in the chain. I kept the IPv6 socket tests on Linux and Android because they work there.

There seems to be some congestion at CirrusCI (the failing test lost connection to the runner), but I will force-push again later and hope that the tests pass then.

sgasse:
@asomers I think this would also be ready for merge.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - Fixed the function signature of `recvmmsg`, potentially causing UB
   ([#2119](https://github.com/nix-rust/nix/issues/2119))
+
+- Fixed the value of optlen as passed to `libc::setsockopt`.
+  (#[#2073](https://github.com/nix-rust/nix/issues/2073))
```
asomers, line 16: This is the wrong PR number.  Also, could you make the message a little bit more clear?  It should include the user-visible change (IpMulticastTtl is fixed) rather than the underlying cause.  The underlying cause is mostly just useful to developers.

asomers (changes requested):

CHANGELOG.md
sgasse: Originally, I found an issue with `IpMulticastTtl`, but my fix did not address the underlying issue. You identified the underlying issue yourself @asomers and pushed a fix on this PR. So I just rebased on it and added the tests. Thus I also just moved the CHANGELOG.md message that you wrote originally.

The number here points to the issue which I originally raised and you referred to yourself when pushing the fix. I now updated the changelog message in your commit and moved it to the new `changelog` subdir.

Is it fine now? If not, would you prefer that I close this PR and you just push your fix separately and add tests to get your fix in?

asomers:
Could you please try rebasing this PR to fix CI and resolve the conflict?

sgasse:
Sure, updated :slightly_smiling_face:

sgasse:
Should be ready again @asomers ;)

asomers (approved):

asomers:
I'm afraid there's another conflict, @sgasse .  Could you please rebase?

sgasse:
Sure, I guess that conflict came from my multicast hops PR. Rebased and pushed again :slightly_smiling_face:

asomers (approved):

PR #2074, opened 2023-07-14
Add ControlMessage::Ipv6MulticastHopLimit

sgasse:
When sending IPv6 multicast packets with `sendmsg`, Linux does not use the hop limit set on the socket. Instead, the hop limit has to be specified for each individual message with ancillary data in a cmsg.

This commit adds the enum variant
`ControlMessage::Ipv6MulticastHopLimit` to specify the limit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added `Icmp` and `IcmpV6` to `SockProtocol`.
   (#[2103](https://github.com/nix-rust/nix/pull/2103))
 
+- Added `Ipv6HopLimit` to `::nix::sys::socket::ControlMessage` for Linux,
+  MacOS, FreeBSD, DragonflyBSD, Android, iOS and Haiku.
+  ([#2074](https://github.com/nix-rust/nix/pull/2074))
+
 ## [0.27.1] - 2023-08-28
 
 ### Fixed
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -18,7 +18,7 @@ use libc::{
 use std::io::{IoSlice, IoSliceMut};
 #[cfg(feature = "net")]
 use std::net;
-use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, RawFd, OwnedFd};
+use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, OwnedFd, RawFd};
 use std::{mem, ptr};
 
 #[deny(missing_docs)]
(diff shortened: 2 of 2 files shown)


asomers (changes requested):
The PR looks good but for a few things:
* Why limit it to Linux?  libc defines IPV6_HOPLIMIT for several operating systems.
* It needs a CHANGELOG entry.
* Ideally there would be a test, though that can be hard for some cmsg types.

asomers:
Oh, and most of the CI failures should be fixed if you rebase on master.

sgasse:
Thanks for the feedback @asomers !
- Seeing with which hop limit a packet actually leaves the network interface is hard to test - if you have a suggestion for a test or an existing test to orient myself along, I am happy to write one.
- Because I can only manually test this so far, I only added linux because I could not verify that it works on other systems myself. If I should add another group of devices (android or something) please let me know.
- I added a changelog entry.

Unfortunately, there are still a bunch of CI failures which I think are not related to my patch despite rebasing on latest `upstream/master` :thinking:

asomers:
No, I don't have a good suggestion for how to write a test.  It isn't always possible for weird network options.  But I did fix the CI failures on master, so that should work if you rebase.  And I suggest you go ahead and define this on every platform where libc defines IPV6_HOPLIMIT.  That looks like haiku, Linux, Android, FreeBSD, DragonflyBSD, macos, and ios.

asomers:
Yeah, I raised MSRV.  And it's fixed now, so you can rebase.

sgasse:
Thanks, should be ready now after the rebase :slightly_smiling_face:

asomers (approved):

asomers:
@sgasse there was a merge conflict.

sgasse:
Hm OK during a local rebase, I was not required to resolve anything manually. However seeing that you released version `0.27.0`, I moved my addition to the changelog into a new `Unreleased` block.

sgasse:
@asomers maybe now everything is ready for a merge? :slightly_smiling_face:

sgasse:
@asomers would you be fine with merging this? It would be really useful for me; right now, I have to include this version with a git tag to be able to use the functionality. I think I resolved all issues that you raised, the changelog is up to date and CI is also happy.

asomers:
@sgasse the changelog is out of date again, unfortunately.  We really need to switch to recording it in a less merge-conflict-prone way.

sgasse:
Ah OK :upside_down_face: I have not tried it out, but maybe there are less merge conflicts if we always had the **complete** structure

```md
## [Unreleased] - ReleaseDate

### Fixed

### Changed

### Added
```

above the release changes? I would hope that git can then figure out where to place changes even if other releases are done before. But not sure, how do other crates to that? :thinking: 

Nevertheless, updated the patch.

PR #2077, opened 2023-07-17
Enable socket timestamping options on Android

spencercw:
Requires https://github.com/rust-lang/libc/pull/3267 which landed in libc 0.2.147.


asomers (approved):
bors r+

PR #2079, opened 2023-07-17
Remove sigevent support on Fuchsia

### LINKED ISSUE #1441: Undefined behavior in SigEvent::new on Fuchsia
Beginning with Rust 1.41.0, the build fails on Fuchsia.  It triggers this error:
```
error: the type `libc::sigevent` does not permit zero-initialization
   --> src/sys/signal.rs:855:36
    |
855 |             let mut sev = unsafe { mem::MaybeUninit::<libc::sigevent>::zeroed().assume_init() };
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                                    |
    |                                    this code causes undefined behavior when executed
    |                                    help: use `MaybeUninit<T>` instead, and only call `assume_init` after initialization is done
```
Judging by the comments, the original author of Fuchsia support was aware of this problem and tried to work around it.  However, later versions of the Rust compiler got more strict.  Since Fuchsia is tier-3, I'm just going to disable the warning (on Fuchsia only) to fix the build.  @amanda-tait would you be willing to devise a more permanent solution?

asomers:
It triggers UB, which the compiler warns about beginning with 1.41.0. Remove it, due to lack of a Fuchsia maintainer and lack of feedback from the original Fuchsia porter.

Fixes #1441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - `nix::sys::signalfd::signalfd` is deprecated.  Use
   `nix::sys::signalfd::SignalFd` instead.
   ([#1938](https://github.com/nix-rust/nix/pull/1938))
+- Removed `SigEvent` support on Fuchsia, where it was unsound.
+  ([#2079](https://github.com/nix-rust/nix/pull/2079))
 
 ## [0.26.2] - 2023-01-18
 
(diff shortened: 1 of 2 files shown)


asomers:
cc @amanda-tait @tamird

tamird:
Fuchsia doesn't have signals. LGTM.

asomers:
> Fuchsia doesn't have signals. LGTM.

And yet it defines all of this stuff in libc.  Should those be removed?

tamird:
Probably! I believe @cramertj added those things, or at least moved them around?

asomers:
bors retry

PR #2081, opened 2023-07-18
Extend type of service/traffic class socket options

spencercw:


asomers:
This looks good, but would it be possible to add a test?

spencercw:
Added tests and rebased on master. The Linux aarch64 failure seems to be a transient network error.

SteveLauC:
Hi, sorry for the late reply.

Something at Nix has changed so that this PR needs an update:

1. We switched to a new mode for adding changelogs to avoid conflicts, please see [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to do it
2. Now we prefer cfg alias to avoid big cfg chunks, please see [CONVENTIONS.md](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#cfg-gates)
    
    For example, the following code can be simplified to:

    ```
    #[cfg(any(
        target_os = "android",
        target_os = "freebsd",
        target_os = "ios",
        target_os = "linux",
        target_os = "macos",
        target_os = "netbsd",
        target_os = "openbsd",
    ))]
    ```
    ```
    #[cfg(any(linux_android, netbsdlike, target_os = "freebsd", target_os = "ios", target_os = "macos"))]
    ```

src/sys/socket/mod.rs
```diff
@@ -748,6 +748,14 @@ pub enum ControlMessageOwned {
     #[cfg(feature = "net")]
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
     Ipv6PacketInfo(libc::in6_pktinfo),
+    #[cfg(any(linux_android, target_os = "freebsd", target_os = "ios", target_os = "macos"))]
```
asomers, line 751: Here and elsewhere, unless you'rre very sure that this doesn't work on tvos or watchos, just use `apple_targets`.
```suggestion
    #[cfg(any(apple_targets, linux_android, target_os = "freebsd"))]
```

test/sys/test_socket.rs
```diff
@@ -2257,6 +2257,162 @@ pub fn test_recv_ipv6pktinfo() {
...
+
+    let lo_ifaddr = loopback_address(AddressFamily::Inet);
+    let (_lo_name, lo) = match lo_ifaddr {
+        Some(ifaddr) => (
+            ifaddr.interface_name,
+            ifaddr.address.expect("Expect IPv4 address on interface"),
+        ),
+        None => return,
+    };
+    let receive = socket(
```
asomers, line 2277: Wouldn't it be easier to write the test to use `socketpair` instead of to create two sockets and connect them?

asomers (changes requested):

test/sys/test_socket.rs
spencercw: I just copied this pattern from nearby tests. Might be better to leave it as is for the sake of consistency?

PR #2088, opened 2023-08-06
sockopt add OpenBSD's SO_RTABLE to set the route table.

devnexen:
Also SO_ACCEPTFILTER for FreeBSD/NetBSD.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -766,6 +766,26 @@ sockopt_impl!(
     libc::SO_USER_COOKIE,
     u32
 );
+#[cfg(target_os = "openbsd")]
+sockopt_impl!(
+    /// Set the route table for this socket, needs a privileged user if
+    /// the process/socket had been set to the non default route.
+    Rtable,
+    SetOnly,
+    libc::SOL_SOCKET,
+    libc::SO_RTABLE,
+    i32
+);
+#[cfg(any(target_os = "freebsd", target_os = "netbsd"))]
+sockopt_impl!(
+    /// Get/set a filter on this socket before accepting connections similarly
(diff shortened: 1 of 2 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -766,6 +766,26 @@ sockopt_impl!(
...
+    libc::SOL_SOCKET,
+    libc::SO_RTABLE,
+    i32
+);
+#[cfg(any(target_os = "freebsd", target_os = "netbsd"))]
+sockopt_impl!(
+    /// Get/set a filter on this socket before accepting connections similarly
+    /// to Linux's TCP_DEFER_ACCEPT but after the listen's call.
+    AcceptFilter,
+    Both,
```
asomers, line 784: This is set-only, at least on FreeBSD.

asomers (changes requested):

src/sys/socket/sockopt.rs
devnexen: are you sure ? I see nginx trying to get the filter [here](https://github.com/nginx/nginx/blob/1c6183725247024f1bca73ac9a833098af7558af/src/core/ngx_connection.c#L346). Also freebsd regression tests, the expected failure is not about getsockopt [here](https://github.com/freebsd/freebsd-src/blob/676f02a4823f8a4a6e10a1cba73c6324c19b651b/tools/regression/sockets/accf_data_attach/accf_data_attach.c#L97).

asomers: No, looking at the kernel code I see that you're right.  I was just confused by the wording in the man page.

asomers (approved):
bors r+

PR #2090, opened 2023-08-07
MapFlags netbsd/freebsd constant additions.

devnexen:
Related to page alignment, adding map_aligned for custom alignment requirements.


src/sys/mman.rs
```diff
@@ -184,6 +184,40 @@ libc_bitflags! {
...
+        /// Pages aligned on 256tb
+        #[cfg(target_os = "netbsd")]
+        #[cfg_attr(docsrs, doc(cfg(all())))]
+        MAP_ALIGNMENT_256TB;
+        /// Pages aligned on 64pb
+        #[cfg(target_os = "netbsd")]
+        #[cfg_attr(docsrs, doc(cfg(all())))]
+        MAP_ALIGNMENT_64PB;
+        /// Right operand value for the page alignment bitshift calculation
+        /// FIXME: not present in libc for FreeBSD
```
asomers, line 212: Why don't you go ahead and add defines for these on FreeBSD (and maybe other places) in libc before merging this PR into Nix?

asomers (changes requested):

src/sys/mman.rs
```diff
@@ -184,6 +184,40 @@ libc_bitflags! {
         #[cfg(target_os = "openbsd")]
         #[cfg_attr(docsrs, doc(cfg(all())))]
         MAP_CONCEAL;
+        /// Pages aligned on 64kb
+        #[cfg(target_os = "netbsd")]
+        #[cfg_attr(docsrs, doc(cfg(all())))]
```
SteveLauC, line 189: Please rebase and remove the `#[cfg_attr(docsrs, doc(cfg(all())))]` attributes from all the variants as we have removed them all from the codebase (#2192)

src/sys/mman.rs
SteveLauC: These 2 constants have been added in [libc#3326](https://github.com/rust-lang/libc/pull/3326)

SteveLauC (changes requested):

PR #2091, opened 2023-08-08
Do not enable any features by default

smalis-msft:
The original intention behind https://github.com/nix-rust/nix/pull/1498 was to allow repositories to slim down what amount of nix they consume, saving on compile time. However, with all these features enabled by default, it only takes one crate in your entire dependency tree that doesn't set `default-features = false` and suddenly you're back to square one. The better solution IMHO would be to always require crates to opt in to the functionality they need, ensuring that this footgun can't fire.

This is a breaking change, and would need a new version number before release.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -18,9 +18,9 @@ build: &BUILD
     - . $HOME/.cargo/env || true
     - $TOOL -Vv
     - rustc -Vv
-    - $TOOL $BUILD $ZFLAGS --target $TARGET --all-targets
-    - $TOOL doc $ZFLAGS --no-deps --target $TARGET
-    - $TOOL clippy $ZFLAGS --target $TARGET --all-targets -- $CLIPPYFLAGS
+    - $TOOL $BUILD $ZFLAGS --target $TARGET --all-targets --all-features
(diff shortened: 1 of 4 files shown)


smalis-msft:
It looks like in order for this to be doable a lot of cleanup is needed around putting `using`s and test code behind feature flags, and then updating CI to test with --all-features. Unfortunately I don't have the time to take that on.

smalis-msft:
Clever work around, but we're still seeing unused import errors in CI for some reason. I'll go add all features to the testing.

asomers:
> Clever work around, but it's not quite enough anymore, as default doesn't contain features now. Thoughts on how to fix it? I'll go add all features to the testing.

What about adding a feature named "full" which enables all of the others?

asomers:
Actually, adding a new feature probably isn't good, because it would make the documentation longer and `cargo hack` would take even more time.  Since this pseudofeature is only needed at one place in the code, I think it would be better just to put a very long `#[cfg_attr(not(all(feature = "...", ...)), allow(unused_imports))]`

smalis-msft:
I was afraid you'd say that lol.

asomers (changes requested):
Could you also please mention this in the CHANGELOG ?

asomers (approved):

PR #2092, opened 2023-08-10
Add NFLOG and generic netlink families

nat-rix:
Add the following (linux-only) netlink `SockProtocol`s:
- `NETLINK_GENERIC`
- `NETLINK_NFLOG`


asomers (approved):
LGTM.

PR #2095, opened 2023-08-11
Fix an incorrect lifetime in the return value of recvmsg

### LINKED ISSUE #2083: recvmsg: incorrect lifetime returned
the current signature of recvmsg is:
```rust
pub fn recvmsg<'a, 'outer, 'inner, S>(fd: RawFd, iov: &'outer mut [IoSliceMut<'inner>],
                   mut cmsg_buffer: Option<&'a mut Vec<u8>>,
                   flags: MsgFlags) -> Result<RecvMsg<'a, 'inner, S>>
```

note the iov param:  `iov: &'outer mut [IoSliceMut<'inner>]`

now in the resulting `RecvMsg` we can access the iovs using a method named `iovs()`. This method iterates the iov array provided. But the lifetime of the iov array is `'outer` and not `'inner`.

So the signature of the result should be changed to `Result<RecvMsg<'a, 'outer, S>>`.

If all sounds correct, i can open a PR to fix this.

example fail scenario with tokio (note the move closure):

```rust
            let tokio_stream = /* get a stream somehow... */;
            let mut buffer = vec![0u8; 10];

            let mut iov = [IoSliceMut::new(&mut buffer)];
            
            let res: Result<nix::sys::socket::RecvMsg<'_, '_, ()>, std::io::Error> =
                tokio_stream.try_io(tokio::io::Interest::READABLE, move || {
                    recvmsg::<()>(
                        raw_fd,
                        &mut iov,
                        None,
                        MsgFlags::empty(),
                    )
                    .map_err(|e| std::io::Error::from_raw_os_error(e as i32))
                });
            match res {
                Ok(res) => {
// BUG HAPPENS HERE
// NOTE: here res.iovs() uses the iov array that was moved in the try_io closure
// res.iovs() may point to invalid memory, as iov may have been destroyed already.
                }
                Err(e) => { /* ... */  }
            };
    }

```

asomers:
Fixes #2083

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   `SockaddrIn6::from<std::net::SockaddrV6>`, `IpMembershipRequest::new`, and
   `Ipv6MembershipRequest::new` with future Rust versions.
   ([#2061](https://github.com/nix-rust/nix/pull/2061))
+- Fixed an incorrect lifetime returned from `recvmsg`.
+  ([#2095](https://github.com/nix-rust/nix/pull/2095))
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


asomers:
@yuval-k does this look correct to you?

yuval-k:
looks good to me @asomers

yuval-k (approved):

PR #2096, opened 2023-08-14
Add support for csky

Dirreke:


asomers:
What is csky?

Dirreke:
it's an arch which `e_machine` code is 252. It will be introduced to rust by https://github.com/rust-lang/rust/pull/113658 soon

asomers:
But what is it?  I've never heard of this architecture.  What is the name of some of the processors that use it?

Dirreke:
its organization is https://github.com/c-sky. It's an ISA created by T-head, Alibaba, China. [here](https://xuantie.t-head.cn/store/ecological/cpu) shows the cpu list including xuantie c801, c806, c860, and it also has the old models like ck860, ck810 and so on.
You can get the method for `rust` to compile at https://github.com/rust-lang/rust/pull/113658/files#diff-631a6f87932239efca4448c292286d00ebbf81945baea0de2b728692291219ae ,
the linux support at https://github.com/torvalds/linux/tree/master/arch/csky or https://github.com/c-sky/csky-linux , 
the gcc compiler at [here]https://occ-oss-prod.oss-cn-hangzhou.aliyuncs.com/resource/1356021/1619528643136/csky-linux-gnuabiv2-tools-x86_64-glibc-linux-4.9.56-20210423.tar.gz) , 
the llvm and clang support at https://github.com/llvm/llvm-project/tree/main/llvm/lib/Target/CSKY

Dirreke:
would you like to merge it? @asomers

asomers:
This is even more obscure than loongarch64.  There's virtually no information about this ISA online, at least in English.  I don't want to devote significant resources to supporting ISAs like this.  However, I'm going to push a separate change that should fix this particular build problem for csky and hopefully other obscure ISA, too.

PR #2097, opened 2023-08-14
Implemented all `std::os::fd` traits for `mqueue::MqdT`.

coderBlitz:
Since linux allows polling message queues [^1], these changes enable that use case. Targeted to `linux`, since it was the only one I was able to confirm [^2].

[^1]: [mq_overview(7)](https://linux.die.net/man/7/mq_overview)
[^2]: https://elixir.bootlin.com/linux/latest/source/ipc/mqueue.c#L908


asomers (commented):
Why only for Linux?  Are you sure these traits shouldn't apply more widely?

coderBlitz:
I don't have any reason to artificially limit it, as I mentioned originally it was based on my working knowledge of the `mqueue` POSIX API and the linux implementation of it. Per the POSIX API (see [1] above), there are no guarantees that `mqd_t` is a file descriptor, or even an identically-sized integer; the linux implementation guarantees a valid file descriptor.

At its broadest, the API could be enabled for all unix per the Rust standard lib (see below for exceptions):
```rs
// unix
#[cfg(not(all(
    doc,
    any(
        all(target_arch = "wasm32", not(target_os = "wasi")),
        all(target_vendor = "fortanix", target_env = "sgx")
    )
)))]
#[cfg(target_os = "hermit")]
#[path = "hermit/mod.rs"]
pub mod unix;
#[cfg(not(all(
    doc,
    any(
        all(target_arch = "wasm32", not(target_os = "wasi")),
        all(target_vendor = "fortanix", target_env = "sgx")
    )
)))]
#[cfg(all(not(target_os = "hermit"), any(unix, doc)))]
pub mod unix;
```

However, there appear to be 2 libc families which would not satisfy the type guarantees[^1]: `freebsdlike` and `solarish`. The rest use the `c_int` type, so those should all theoretically compile without issue. In my semi-cursory searches, all of the `c_int` types seem to alias to `i32`, which matches the `RawFd` alias, so everything points to those compiling fine.

[^1]: https://github.com/search?q=repo%3Arust-lang%2Flibc+mqd_t&type=code

asomers:
Man pages suggest that select and poll are supported for message queues on NetBSD and DragonflyBSD.  I'm very surprised to see that libc doesn't even define mq_open on Android, so we can't do it there.  On FreeBSD select and poll are supported but through a different mechanism: one must first use mq_getfd_np to get the file descriptor.  So I think you should enable this for Linux, NetBSD, and DragonflyBSD.

coderBlitz:
I'm slightly concerned the `dragonfly` build succeeded, given the type in the `freebsdlike` files is explicitly a pointer. @asomers Thoughts?

asomers:
> I'm slightly concerned the `dragonfly` build succeeded, given `mqd_t` in the `freebsdlike` files is explicitly a pointer. @asomers Thoughts?

I think you read that wrong.  The `mqd_t` type isn't defined in freebsdlike/mod.rs.  It's defined separately in freebsdlike/freebsd/mod.rs and freebsdlike/dragonfly/mod.rs.

src/mqueue.rs
```diff
@@ -300,3 +307,55 @@ pub fn mq_remove_nonblock(mqd: &MqdT) -> Result<MqAttr> {
...
+impl AsFd for MqdT {
+    /// Borrow the underlying message queue descriptor.
+    fn as_fd(&self) -> BorrowedFd {
+        // SAFETY: [MqdT] will only contain a valid fd by construction.
+        unsafe { BorrowedFd::borrow_raw(self.0) }
+    }
+}
+
+#[cfg(all(
+    unix,
```
asomers, line 324: You can remove the "unix" part.  It's redundant with the other conditions.

asomers (changes requested):

asomers (approved):

PR #2100, opened 2023-08-17
feat: I/O safety `pipe`, `pipe2` & `write`

JonathanWoollett-Light:
Changes `pipe`, `pipe2` and `write` to use `OwnedFd`s.


SteveLauC:
ref: #1750

SteveLauC (approved):

PR #2101, opened 2023-08-20
Economize our Cirrus CI builds

asomers:
Cirrus CI is eliminating unlimited free open source builds.  Open source projects will now be limited to about 16,000 CPU-minutes per month. Tweak our CI configuration to reduce its demands:

* Eliminate the FreeBSD 12 x86_64, iOS x86_64, macOS x86_64, and Linux powerpc builds.  They weren't providing a useful amount of test coverage.
* Reduce each task to 1 CPU and 2 GB of RAM (but 4 GB for OSX).  Our build rarely uses more than a single CPU anyway.
* Split the tasks up into two groups.  The first group of 8 builds contains broad coverage.  Most build failures will be caught by one of these tasks.  The second group of 29 tasks will only run after the first group completes.

https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,12 +40,10 @@ task:
   env:
     TARGET: x86_64-unknown-freebsd
   matrix:
-    - name: FreeBSD 12 amd64 & i686
-      freebsd_instance:
-        image: freebsd-12-4-release-amd64
     - name: FreeBSD 14 amd64 & i686
       freebsd_instance:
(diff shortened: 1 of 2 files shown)


asomers:
Based on the cumulative build time, I think this PR will reduce each build's CPU-minutes by about 2.3x.

PR #2103, opened 2023-08-22
Add `SockProtocol` variants for ICMP{,v6}

gnaaman-dn:
`CanRaw`/`KextEvent` have the same value as `ICMP`, which means one of them has to be a constant.
I chose to make `CanRaw` & `KextEvent` constants since as far as I can tell, `Icmp` is always present (similar to Udp/Tcp, etc).

I started writing a test that sends an Echo-Request, but opening an ICMP/Raw socket requires root permissions, so I'm not sure if it's fit as a test.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -132,11 +132,6 @@ pub enum SockProtocol {
     Udp = libc::IPPROTO_UDP,
     /// Raw sockets ([raw(7)](https://man7.org/linux/man-pages/man7/raw.7.html))
     Raw = libc::IPPROTO_RAW,
-    /// Allows applications and other KEXTs to be notified when certain kernel events occur
-    /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
-    #[cfg(any(target_os = "ios", target_os = "macos"))]
-    #[cfg_attr(docsrs, doc(cfg(all())))]
-    KextEvent = libc::SYSPROTO_EVENT,
     /// Allows applications to configure and control a KEXT
     /// ([ref](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/NKEConceptual/control/control.html))
     #[cfg(any(target_os = "ios", target_os = "macos"))]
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -221,20 +216,33 @@ pub enum SockProtocol {
     #[cfg(any(target_os = "android", target_os = "linux"))]
     #[cfg_attr(docsrs, doc(cfg(all())))]
     EthAll = (libc::ETH_P_ALL as u16).to_be() as i32,
+    /// ICMP protocol ([icmp(7)](https://man7.org/linux/man-pages/man7/icmp.7.html))
+    Icmp = libc::IPPROTO_ICMP,
+    /// ICMPv6 protocol (ICMP over IPv6)
+    IvmpV6 = libc::IPPROTO_ICMPV6,
```
asomers, line 222:
```suggestion
    IcmpV6 = libc::IPPROTO_ICMPV6,
```

src/sys/socket/mod.rs
```diff
@@ -2441,4 +2449,15 @@ mod tests {
         )
         .expect("Failed to open routing socket");
     }
+    #[test]
+    fn sock_protocol_constant_matches() {
```
asomers, line 2453: This part is tautological.  I suggest removing it.

asomers (changes requested):
It's totally possible to write a test that requires privileges.  In fact, several existing tests already do.  See test_so_type_unknown for an example that uses raw sockets.

src/sys/socket/mod.rs
gnaaman-dn: I wouldn't say it is tautological.
We're initializing `CanRaw` to `Icmp`, because these constants happen to match; this is an added sanity to make sure this is correct. (I'd have opted for a `static_assert!` if it was available).

I do think it is unlikely to ever break, so will remove it.

rcloran:
:+1: These also work for creating unprivileged ICMP{,v6} sockets using SOCK_DGRAM on Linux and macOS -- not sure if that helps the testing situation at all (sounds like not really)? Looks like work has stalled a tiny bit -- anything I can do to help get this over the line?

SteveLauC:
> anything I can do to help get this over the line?

Bump received, thanks!

-----

> I started writing a test that sends an Echo-Request, but opening an ICMP/Raw socket requires root permissions, so I'm not sure if it's fit as a test.

Friendly ping @asomers, can I trouble you for another review on this PR, it seems that the test using raw socket is done, and the request changes are resolved

----

hi, @gnaaman-dn Would you like to rebase your branch to resolve the conflict, thanks!

gnaaman-dn:
Sure can :)

asomers (approved):

PR #2106, opened 2023-08-27
inix::sys::socket::sockopt add Linux's TCP_DEFER_ACCEPT.

devnexen:


src/sys/socket/sockopt.rs
```diff
@@ -797,6 +797,17 @@ sockopt_impl!(
     libc::accept_filter_arg
 );
 #[cfg(target_os = "linux")]
+sockopt_impl!(
+    /// Defers completion on accept so only when data arrives.
```
asomers, line 801: Could you try to reword this?  The first sentence isn't grammatically correct.  The second sentence's `x` refers to a variable that isn't anywhere in the prototype.  Probably "the number of seconds" would be better.  And how would one use this after listen(2), given that listen blocks?

asomers (changes requested):

devnexen:
To your last point and why this option can be set before or after listen : 
1/ one example is [HAProxy](https://github.com/haproxy/haproxy/blob/0074c36dd22e4bb09d1ed24de416ab8bb9ce25da/src/proto_tcp.c#L673) listen is [done below](https://github.com/nginx/nginx/blob/24f3cb795e2d822be39387d828ac2ddc28f775fb/src/core/ngx_connection.c#L845).
2/ another example is [NGINX](https://github.com/nginx/nginx/blob/24f3cb795e2d822be39387d828ac2ddc28f775fb/src/core/ngx_connection.c#L903) listen is in this case [above](https://github.com/nginx/nginx/blob/24f3cb795e2d822be39387d828ac2ddc28f775fb/src/core/ngx_connection.c#L860).

Note however SO_ACCEPTFILTER is always after listen.

asomers:
Sorry; when I said "use this after listen", I was thinking "use this after accept".  Yes, it can be used after listen(), but what does that have to do with portability?

devnexen:
I wrote it for two reasons, to highlight this difference between TCP_DEFER_ACCEPT and SO_ACCEPTFILTER but also I think when you want to implement this feature for those platforms it s somehow better to do both after listen like NGINX does.

asomers:
Why would it be better do to it after listen than before, if it only has an effect on the behavior of accept?

devnexen:
feature wise it changes nothing indeed, just grouping implementations look better in my opinion.

asomers:
> feature wise it changes nothing indeed, just grouping implementations look better in my opinion.

Then why did you say "the latter is better for portable workflows"?  It isn't portable at all, regardless of the order.

PR #2113, opened 2023-08-28
Add IPV6_PKTINFO, Correct documentationn for IPV6_RECVPKTINFO

ishanjain28:
`IPV6_PKTINFO` was missing from `sockopt` so I added it and the documentation for `IPV6_RECVPKTINFO` incorrectly said it was enabling the flag for `IPV6_PKTINFO` so I corrected that.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -922,8 +922,8 @@ sockopt_impl!(
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
-    /// Set delivery of the `IPV6_PKTINFO` control message on incoming
-    /// datagrams.
+    /// Pass an `IPV6_PKTINFO` ancillary message that contains a in6_pktinfo
+    /// structure that supplies some information about the incoming packet.
     Ipv6PacketInfo,
     Both,
     libc::IPPROTO_IPV6,
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -905,7 +905,7 @@ sockopt_impl!(
     libc::IP_PKTINFO,
     bool
 );
-#[cfg(any(linux_android))]
+#[cfg(any(linux_android, bsd))]
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -905,16 +905,29 @@ sockopt_impl!(
     libc::IP_PKTINFO,
     bool
 );
-#[cfg(any(linux_android, target_os = "freebsd", apple_targets, netbsdlike))]
+#[cfg(any(linux_android))]
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
-    /// Set delivery of the `IPV6_PKTINFO` control message on incoming
+    /// Set delivery of the `IPV6_RECVPKTINFO` control message on incoming
     /// datagrams.
     Ipv6RecvPacketInfo,
     Both,
     libc::IPPROTO_IPV6,
     libc::IPV6_RECVPKTINFO,
+    bool
+);
+
+#[cfg(any(linux_android, bsd))]
+#[cfg(feature = "net")]
+sockopt_impl!(
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    /// Set delivery of the `IPV6_PKTINFO` control message on incoming
+    /// datagrams.
+    Ipv6PacketInfo,
+    Both,
+    libc::IPPROTO_IPV6,
+    libc::IPV6_PKTINFO,
     bool
 );
 #[cfg(bsd)]
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -909,7 +909,7 @@ sockopt_impl!(
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
-    /// Set delivery of the `IPV6_RECVPKTINFO` control message on incoming
+    /// Set delivery of the `IPV6_PKTINFO` control message on incoming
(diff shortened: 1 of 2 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -858,14 +866,34 @@ sockopt_impl!(
...
 );
+#[cfg(any(
+    target_os = "android",
+    target_os = "freebsd",
+    target_os = "ios",
+    target_os = "linux",
+    target_os = "macos",
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
```
SteveLauC, line 885: This symbol seems to be available on DragonFlyBSD and other Apple targets, if we add them, then this can be changed to:

```suggestion
#[cfg(any(linux_android, bsd))]
```

SteveLauC (changes requested):
Thanks for your interest in contributing to Nix, please rebase your branch so that the CI can work (We have migrated from Cirrus to GitHub Action)

evanrittenhouse:
Hi, I need to be able to set this sockopt as well - should I make a new PR? This one seems a bit stale cc @SteveLauC

SteveLauC:
> Hi, I need to be able to set this sockopt as well - should I make a new PR? This one seems a bit stale cc @SteveLauC

Hi, thanks for your interest in filing a PR for this! Please go ahead:)

ishanjain28:
hey @evanrittenhouse, Can I revive this PR instead? will rebase it today

src/sys/socket/sockopt.rs
ishanjain28: Updated!

src/sys/socket/sockopt.rs
```diff
@@ -905,18 +905,31 @@ sockopt_impl!(
     libc::IP_PKTINFO,
     bool
 );
-#[cfg(any(linux_android, target_os = "freebsd", apple_targets, netbsdlike))]
+#[cfg(any(linux_android))]
```
ishanjain28, line 908: I removed BSD target because I can't find `IPV6_RECVPKTINFO` in `ip6` bsd man pages. https://man.freebsd.org/cgi/man.cgi?query=ip6&sektion=4

ishanjain28:
I must be missing some thing. I can not find `IPV6_RECVPKTINFO` in bsd man pages. Is it still present there and I am looking at the wrong place or do I update the tests ?

src/sys/socket/sockopt.rs
SteveLauC: It is available on FreeBSD: https://github.com/freebsd/freebsd-src/blob/ef4f4a44d9134ddadede0e2e6e658d0688c5ab3c/sys/netinet6/in6.h#L446

SteveLauC: Also defined in the libc crate: https://docs.rs/libc/latest/x86_64-unknown-freebsd/libc/constant.IPV6_RECVPKTINFO.html

SteveLauC: The CI failure on BSDs is introduced by this change, please revert it

src/sys/socket/sockopt.rs
```diff
@@ -905,18 +905,31 @@ sockopt_impl!(
...
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
-    /// Set delivery of the `IPV6_PKTINFO` control message on incoming
+    /// Set delivery of the `IPV6_RECVPKTINFO` control message on incoming
```
SteveLauC, line 912: Emm, why this change? The old document seems to be correct, it asks the incoming packages to contain the `IPV6_PKTINFO` info

src/sys/socket/sockopt.rs
```diff
@@ -905,18 +905,31 @@ sockopt_impl!(
...
     libc::IPV6_RECVPKTINFO,
     bool
 );
+
+#[cfg(any(linux_android, bsd))]
+#[cfg(feature = "net")]
+sockopt_impl!(
+    #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
+    /// Set delivery of the `IPV6_PKTINFO` control message on incoming
+    /// datagrams.
```
SteveLauC, line 926:
```suggestion
    /// Pass an `IPV6_PKTINFO` ancillary message that contains a in6_pktinfo
    /// structure that supplies some information about the incoming packet.
```

Let's make it consistent with the IPv4 one:)

src/sys/socket/sockopt.rs
SteveLauC, line 908:
```suggestion
#[cfg(any(linux_android, bsd))]
```

DragonFlyBSD also has this, so `bsd` would work here

SteveLauC (changes requested):

SteveLauC:
BTW, a changelog is needed, please see [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add one.

If you apply the changes requested by me, then it should be something similar to:

> filename: 2113.added.md

```md
Add socket option `IPV6_PKTINFO` for BSDs/Linux/Android, also `IPV6_RECVPKTINFO` for DragonFlyBSD
```

src/sys/socket/sockopt.rs
ishanjain28: Thank you for the links! I have updated it

src/sys/socket/sockopt.rs
ishanjain28: @stevelauc This code is for the `IPV6_RECVPKTINFO` flag, The documentation incorrectly mentioned `IPV6_PKTINFO`

SteveLauC: It seems to be copied from the Linux manual: https://man7.org/linux/man-pages/man7/ipv6.7.html

> IPV6_RECVPKTINFO (since Linux 2.6.14)
              Set delivery of the IPV6_PKTINFO control message on
              incoming datagrams.  Such control messages contain a
              struct in6_pktinfo, as per RFC 3542.  Allowed only for
              SOCK_DGRAM or SOCK_RAW sockets.  Argument is a pointer to
              a boolean value in an integer.

ishanjain28: Ah my bad, Reverting it now

SteveLauC (approved):
Thanks! :)

PR #2117, opened 2023-08-28
Revert "Merge #2027 #2057"

### LINKED ISSUE #2112: Breaking change in nix 0.26.3  due to bitflags upgrade
In nix 0.26.3, the bitflags crate was updated which is technically a breaking change. 
For example, the function `nix::fcntl::OFlag::from_bits_unchecked` no longer exist.

(Note: while we were using that function in our project, we're going to port to 0.27 anyway so we're not going to be impacted by this change, but i just wanted to report the issue just in case you think this is worth fixing. I can suggest either reverting the change, or add the missing functions in `libc_bitflags` macro. )

asomers:
This reverts commit 0c3afc27d7bdbea7e5c12969e1ba0607c14b1dc9.

This rolls back the version of bitflags used to 1.1.  Upgrading bitflags to 2.0 inadvertently caused breaking changes in the r0.26 branch, by virtue of the methods that the bitflags! macro adds.

Also, mask the bad_bit_mask lint, triggered by bitflags's generated code.  That lint was the motivation for upgrading bitflags in the first place.  The bitflags maintainers have decided not to fix those warnings in the 1.x release series.
https://github.com/bitflags/bitflags/pull/373

Fixes #2112

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ targets = [
 
 [dependencies]
 libc = { version = "0.2.137", features = [ "extra_traits" ] }
-bitflags = "2.3.1"
+bitflags = "1.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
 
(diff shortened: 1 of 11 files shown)


asomers:
cc @ogoffart

ogoffart:
Thanks for taking care of this.

PR #2120, opened 2023-09-02
Fixes UB of `recvmmsg` and simplifies signatures of `recvmmsg` and `sendmmsg`

### LINKED ISSUE #2119: `recvmmsg` may cause UB
## Description
`recvmmsg` currently mutates state through shared references, which is immediately UB. This is caused by using shared references in the function signature, where mutable references are requried.

## Detailed Explanation

Currently, `recvmmsg` essentially does the following:

```rust
pub fn recvmmsg<'a, XS>(
    // elided ...
    slices: XS,
)
where
    XS: IntoIterator<Item = &'a I>,
    I: AsRef<[IoSliceMut<'a>]> + 'a,
{
    for slice in slices.into_iter() {
        let ptr_mut = slice.as_ref().as_ptr() as *mut libc::iovec;

        // Continue using `ptr_mut` in a mutable context
    }
}
```
This boils down to: (stripping one layer of slices)

```rust
pub fn recvmmsg<'a>(
    // elided ...
    slices: &'a [&'a mut [u8]],
) {
    for slice in slices.iter() {
        let ptr = &**slice as *const [u8];
        let ptr_mut = ptr.cast_mut();

        // Continue using `ptr_mut` in a mutable context
    }
}
```
Although we have a mutable reference to the inner slice, it is hidden behind a shared reference, which behaves exactly like it was borrowed immutably in first place. Casting a shared reference to a `*mut _` pointer and mutating the value through that pointer is always UB.

## Example demonstrating this bug
```rust
use std::{
    io::IoSliceMut,
    net::{SocketAddr, UdpSocket},
    os::fd::AsRawFd,
};

use crossbeam::thread::scope;
use nix::sys::socket::{recvmmsg, MsgFlags, MultiHeaders, SockaddrIn};

fn main() {
    let mut buf = [0u8; 2];
    let i = [IoSliceMut::new(&mut buf[..])];
    let xs = [&i];

    scope(|s| {
        s.spawn(|_| loop {
            let socket = UdpSocket::bind("0.0.0.0:42".parse::<SocketAddr>().unwrap()).unwrap();
            let slices = xs.iter();

            let mut h = MultiHeaders::<SockaddrIn>::preallocate(1, None);

            // This thread is mutating `buf` ...
            recvmmsg(socket.as_raw_fd(), &mut h, slices, MsgFlags::empty(), None).unwrap();
        });

        s.spawn(|_| loop {
            let slices = xs.iter();

            for slice in slices {
                // While this thread tries to read `buf`
                println!("{:?}", slice);
            }
        });
    })
    .unwrap();
}
```

## Proposal

Change the function signature of `recvmmsg` to:

```rust
pub fn recvmmsg<'a, XS, S, I>(
    fd: RawFd,
    slices: XS,
    flags: MsgFlags,
    mut timeout: Option<crate::sys::time::TimeSpec>,
) -> crate::Result<MultiResults<'a, S>>
where
    XS: IntoIterator<Item = &'a mut I>,
    I: AsMut<[IoSliceMut<'a>]> + 'a,
```
or (preferably, but inconsistent with `sendmmsg`)

```rust
pub fn recvmmsg<'a, XS, S, I>(
    fd: RawFd,
    slices: XS,
    flags: MsgFlags,
    mut timeout: Option<crate::sys::time::TimeSpec>,
) -> crate::Result<MultiResults<'a, S>>
where
    XS: IntoIterator<Item = I>,
    I: AsMut<[IoSliceMut<'a>]>,
```

Jan561:
The signature of `recvmmsg` gets changed to:

```rust
pub fn recvmmsg<'a, XS, S, I>(
    fd: RawFd,
    slices: XS,
    flags: MsgFlags,
    mut timeout: Option<crate::sys::time::TimeSpec>,
) -> crate::Result<MultiResults<'a, S>>
where
    XS: IntoIterator<Item = I>,
    I: AsMut<[IoSliceMut<'a>]>,
```

The signature of `sendmmsg` gets changed to:

```rust
pub fn sendmmsg<'a, XS, AS, C, I, S>(
    fd: RawFd,
    slices: XS,
    addrs: AS,
    cmsgs: C,
    flags: MsgFlags
) -> crate::Result<MultiResults<'a, S>>
    where
        XS: IntoIterator<Item = I>,
        AS: AsRef<[Option<S>]>,
        I: AsRef<[IoSlice<'a>]>,
        C: AsRef<[ControlMessage<'a>]>,
        S: SockaddrLike,
```

Fixes #2119

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1552,11 +1552,11 @@ pub fn sendmmsg<'a, XS, AS, C, I, S>(
     flags: MsgFlags
 ) -> crate::Result<MultiResults<'a, S>>
     where
-        XS: IntoIterator<Item = &'a I>,
+        XS: IntoIterator<Item = I>,
         AS: AsRef<[Option<S>]>,
-        I: AsRef<[IoSlice<'a>]> + 'a,
-        C: AsRef<[ControlMessage<'a>]> + 'a,
-        S: SockaddrLike + 'a
+        I: AsRef<[IoSlice<'a>]>,
+        C: AsRef<[ControlMessage<'a>]>,
+        S: SockaddrLike,
 {
 
     let mut count = 0;
@@ -1564,11 +1564,11 @@ pub fn sendmmsg<'a, XS, AS, C, I, S>(
 
     for (i, ((slice, addr), mmsghdr)) in slices.into_iter().zip(addrs.as_ref()).zip(data.items.iter_mut() ).enumerate() {
         let p = &mut mmsghdr.msg_hdr;
-        p.msg_iov = slice.as_ref().as_ptr() as *mut libc::iovec;
+        p.msg_iov = slice.as_ref().as_ptr().cast_mut().cast();
         p.msg_iovlen = slice.as_ref().len() as _;
 
         p.msg_namelen = addr.as_ref().map_or(0, S::len);
-        p.msg_name = addr.as_ref().map_or(ptr::null(), S::as_ptr) as _;
+        p.msg_name = addr.as_ref().map_or(ptr::null(), S::as_ptr).cast_mut().cast();
 
         // Encode each cmsg.  This must happen after initializing the header because
         // CMSG_NEXT_HDR and friends read the msg_control and msg_controllen fields.
@@ -1583,9 +1583,16 @@ pub fn sendmmsg<'a, XS, AS, C, I, S>(
             pmhdr = unsafe { CMSG_NXTHDR(p, pmhdr) };
         }
 
-        count = i+1;
+        // Doing an unchecked addition is alright here, as the only way to obtain an instance of `MultiHeaders`
+        // is through the `preallocate` function, which takes an `usize` as an argument to define its size,
+        // which also provides an upper bound for the size of this zipped iterator. Thus, `i < usize::MAX` or in
+        // other words: `count` doesn't overflow
+        count = i + 1;
     }
 
+    // SAFETY: all pointers are guaranteed to be valid for the scope of this function. `count` does represent the
+    // maximum number of messages that can be sent safely (i.e. `count` is the minimum of the sizes of `slices`,
+    // `data.items` and `addrs`)
     let sent = Errno::result(unsafe {
         libc::sendmmsg(
             fd,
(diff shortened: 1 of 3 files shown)


Jan561:
I've completely disabled the `personality::set` doctest on my branch as it seems that it's completely broken now on multiple platforms

src/sys/personality.rs
```diff
@@ -80,10 +80,9 @@ pub fn get() -> Result<Persona> {
 ///
 /// Example:
 ///
-// Disable test on aarch64 until we know why it fails.
```
asomers, line 83: This failure seemed to be due to changes on Google's side.  But they reverted it after a few days, so you can revert this part, too.

src/sys/socket/mod.rs
```diff
@@ -1552,11 +1552,11 @@ pub fn sendmmsg<'a, XS, AS, C, I, S>(
     flags: MsgFlags
 ) -> crate::Result<MultiResults<'a, S>>
     where
-        XS: IntoIterator<Item = &'a I>,
```
asomers, line 1555: Could you explain the motivation for this part?

asomers (changes requested):

src/sys/personality.rs
Jan561: Done

src/sys/socket/mod.rs
Jan561: `AsRef<U>` is implemented for `&T` if `T: AsRef<U>`, which means that this is less restrictive and all previous function calls should be compatible with this new signature (while this still being a breaking change).

The additional reference is simply redundant and could be a hurdle when trying to create the iterator.

Jan561:
Thanks for your review, I've also noticed your PR #2140 and applied it to the remaining pointer conversions in both functions (except [this one](https://github.com/nix-rust/nix/pull/2120/files#diff-fadc3f607885be01f148ca334aee8e2fc0a0d41678a328d39cfdc7d2a621b8c5L1727), which is fine imo).

asomers (approved):
Sadly, there's another merge conflict in CHANGELOG.  Could you fix that, please?

Jan561:
Done. For the `sendmmsg`-change, I've written a small example of what is possible now, which wasn't possible before.

https://gist.github.com/Jan561/1e44db811698d06bb247c30bbe448fc9

asomers (approved):
Your gist there suggests that a different API would be more convenient.  This is something that I've been thinking about too.  It would look something like this:
```rust
struct MsgVec<'a> {
    msgvec: Vec<libc::msghdr>,
}
impl<'a> MsgVec<'a> {
   fn with_capacity(n: usize) -> Self {...}
   fn add<S: SockaddrLike + 'a>(&mut self, to: S, I: [IoSlice<'a>], &mut cmsg: ControlMessage<'a>)
   fn reset(&mut self);
}

fn sendmmsg<F>(fd: F, msgvec: MsgVec<'_>, flags: MsgFlags)
    where F: AsFd
    {}
```
And you could use it something like this:
```rust
let mut msg_vec = MsgVec::with_capacity(2);
msg_vec.add(some_host, &[iovec0, iovec1][..], &[ControlMessage::ScmRights(...)][..]);
sendmmsg(&s, msg_vec);
```
The general idea being that the `add` API keeps related host, data, and control messages together rather than forcing the user to split them into 3 separate arrays.  Plus, the `MsgVec::reset` method allows reusing storage for a `MsgVec`, eliminating the need to allocate on each call of sendmmsg, which was the primary problem of the earlier implementation.

PR #2121, opened 2023-09-03
FreeBSD's rfork wrapper proposal

devnexen:


src/unistd.rs
```diff
@@ -3355,6 +3355,55 @@ mod getres {
     }
 }
 
+feature! {
+#![feature = "process"]
+#[cfg(target_os = "freebsd")]
+pub mod rfork_freebsd {
```
SteveLauC, line 3361: We don't need to put it in a dedicated module, the existing 2 `pub mod` within `unistd.rs` are for better naming

* acct: https://github.com/nix-rust/nix/pull/952#discussion_r224821858
* alarm: https://github.com/nix-rust/nix/pull/830#discussion_r164946670

`rfork(2)`, IMHO, is good enough so we don't need this

src/unistd.rs
```diff
@@ -3355,6 +3355,55 @@ mod getres {
...
+    use crate::errno::Errno;
+    use crate::Result;
+    use super::ForkResult;
+    use super::Pid;
+
+    libc_bitflags! {
+        /// Flags for [`rfork`]
+        /// subset of flags supported by FreeBSD 12.x and onwards
```
SteveLauC, line 3369:
```suggestion
        /// Flags for [`rfork`]
        ///
        /// subset of flags supported by FreeBSD 12.x and onwards
```

We should add a newline here or the they will be rendered into a single paragraph

src/unistd.rs
```diff
@@ -3355,6 +3355,55 @@ mod getres {
...
+        /// with a safe outcome, thus as `RFMEM` can possibly lead to undefined behavior,
+        /// it is not in the list. And `rfork_thread` is deprecated.
+        pub struct RforkFlags: libc::c_int {
```
SteveLauC, line 3372: Seems like there is a `RFSIGSHARE` flag that is currently missing, though it is also not available in `libc`

SteveLauC (changes requested):

SteveLauC:
And please revert the format changes

src/unistd.rs
SteveLauC: Noticed this PR https://github.com/rust-lang/libc/pull/3457, thanks for doing it!

PR #2125, opened 2023-09-08
Add `MapFlags::map_hugetlb_with_size_log2`

### LINKED ISSUE #2124: Expose MAP_HUGE_SHIFT
Right now it's not possible to construct `MapFlags` for huge page mapping with huge page size variable at runtime without constructing a big intermediate enum or by using `from_bits` and `libc::MAP_HUGE_SHIFT`.

It would be nice to have something like this:
```rust
impl MapFlags {
    fn map_hugetlb_with_shift(shift: u8) -> Option<Self> {
        if shift >= 16 && shift <= 63 {
            Some(Self(libc::MAP_HUGETLB | (shift as i32) << libc::MAP_HUGE_SHIFT))
        } else {
            None
        }
    }
}
```

newpavlov:
Closes #2124

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -190,6 +190,31 @@ libc_bitflags! {
     }
 }
 
+impl MapFlags {
+    /// Create `MAP_HUGETLB` with provided size of huge page.
+    ///
+    /// Under the hood it computes `MAP_HUGETLB | (huge_page_size_log2 << libc::MAP_HUGE_SHIFT`).
+    /// `huge_page_size_log2` denotes logarithm of huge page size to use and should be
+    /// between 16 and 63 (inclusively).
+    ///
+    /// ```
+    /// # use nix::sys::mman::MapFlags;
+    /// let f = MapFlags::map_hugetlb_with_size_log2(30).unwrap();
+    /// assert_eq!(f, MapFlags::MAP_HUGETLB | MapFlags::MAP_HUGE_1GB);
+    /// ```
+    #[cfg(target_os = "linux")]
+    #[cfg_attr(docsrs, doc(cfg(all())))]
+    pub fn map_hugetlb_with_size_log2(huge_page_size_log2: u32) -> Option<Self> {
+        if (16..=63).contains(&huge_page_size_log2) {
+            let flag = libc::MAP_HUGETLB
+                | (huge_page_size_log2 << libc::MAP_HUGE_SHIFT) as i32;
+            Some(Self(flag.into()))
+        } else {
+            None
+        }
+    }
+}
+
 #[cfg(any(target_os = "linux", target_os = "netbsd"))]
 libc_bitflags! {
     /// Options for [`mremap`].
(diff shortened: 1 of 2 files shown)


src/sys/mman.rs
```diff
@@ -187,6 +190,26 @@ libc_bitflags! {
...
+impl MapFlags {
+    /// Create `MAP_HUGETLB` with provided size of huge page.
+    ///
+    /// Under the hood it computes `MAP_HUGETLB | (huge_page_size_log2 as i32) << libc::MAP_HUGE_SHIFT`.
+    /// `huge_page_size_log2` denotes logarithm of huge page size to use and should between 16 and 63
+    /// (inclusively). For example, `MapFlags::map_hugetlb_with_shift(30)` returns value equal to
+    /// `Some(MapFlags::MAP_HUGETLB | MAP_HUGE_1GB)`.
+    #[cfg(target_os = "linux")]
+    #[cfg_attr(docsrs, doc(cfg(all())))]
+    pub fn map_hugetlb_with_shift(huge_page_size_log2: u8) -> Option<Self> {
```
newpavlov, line 202: I couldn't make it `const` because I couldn't find a way to convert `i32` into `InternalBitFlags`.

src/sys/mman.rs
```diff
@@ -187,6 +190,31 @@ libc_bitflags! {
...
+    /// Under the hood it computes `MAP_HUGETLB | (huge_page_size_log2 << libc::MAP_HUGE_SHIFT`).
+    /// `huge_page_size_log2` denotes logarithm of huge page size to use and should be
+    /// between 16 and 63 (inclusively).
+    ///
+    /// ```
+    /// # use nix::sys::mman::MapFlags;
+    /// let f = MapFlags::map_hugetlb_with_size_log2(30).unwrap();
+    /// assert_eq!(f, MapFlags::MAP_HUGETLB | MapFlags::MAP_HUGE_1GB);
+    /// ```
+    #[cfg(target_os = "linux")]
```
asomers, line 205: This should probably include android and fuchsia, too.  At least, libc defines MAP_HUGETLB for those platforms.  But it doesn't define MAP_HUGE_SHIFT.  Is that an oversight in libc?

asomers (changes requested):
Don't forget a changelog entry.

src/sys/mman.rs
newpavlov: >Is that an oversight in libc?

Maybe. Either way, we would need first introduce the constant to `libc` and only then add expand list of targets, so I think it's better to do it in a later PR.

SteveLauC: I filed a PR adding these symbols for Android or Linux: https://github.com/rust-lang/libc/pull/3444

SteveLauC:
We can merge this PR as-is, but @newpavlov could you please leave a TODO here: support Andorid and Fuchsia when libc 0.2.151 gets released

SteveLauC (approved):

PR #2127, opened 2023-09-08
Add `mmap_anonymous` function

### LINKED ISSUE #2126: Introduce `mmap_nofd` to work around `<F: AsFd>`?
Today when I use nix's `mmap` for creating anonymous mappings I have to use `mmap::<BorrowedFd>(...)`, which feels... quite awkward. It would be nice to have a variant of `mmap` with the following signature:
```rust
pub unsafe fn mmap_nofd(
    addr: Option<NonZeroUsize>,
    length: NonZeroUsize,
    prot: ProtFlags,
    flags: MapFlags,
) -> Result<*mut c_void>
```
The `offset` argument has meaning only for file-backed mappings, so we can remove it as well.

newpavlov:
Closes #2126

### DIFF
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -411,7 +411,9 @@ pub fn munlockall() -> Result<()> {
     unsafe { Errno::result(libc::munlockall()) }.map(drop)
 }
 
-/// allocate memory, or map files or devices into memory
+/// Allocate memory, or map files or devices into memory
+///
+/// For anonymous mappings (`MAP_ANON`/`MAP_ANONYMOUS`), see [mmap_anonymous].
 ///
 /// # Safety
 ///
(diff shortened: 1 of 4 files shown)


src/sys/mman.rs
```diff
@@ -426,8 +426,7 @@ pub unsafe fn mmap<F: AsFd>(
     f: Option<F>,
```
SteveLauC, line 426: Since we now have a dedicated interface for anonymous `mmap`, seems like we should also update the `f` argument of `mmap()` to `f: F`, and add the following document:

> For anonymous mappings (MAP_ANON/MAP_ANONYMOUS), see [mmap_anonymous](https://docs.rs/rustix/latest/rustix/mm/fn.mmap_anonymous.html).

SteveLauC: BTW, the document of this function:

> **allocate memory**, or map files or devices into memory

Should be updated as well since it is not for memory allocations anymore?

src/sys/mman.rs
```diff
@@ -440,6 +439,34 @@ pub unsafe fn mmap<F: AsFd>(
...
+///
+/// This function is a wrapper around [`mmap`]:
+/// `mmap(ptr, len, prot, MAP_ANONYMOUS | flags, -1, 0)`.
+///
+/// # Safety
+///
+/// See the [`mmap(2)`] man page for detailed requirements.
+///
+/// [`mmap(2)`]: https://man7.org/linux/man-pages/man2/mmap.2.html
+pub unsafe fn mmap_anonymous(
```
SteveLauC, line 452: Test code that uses `mmap()` for anonymous maping should be replaced with this new function:

```shell
$ rg "mmap::<"
src/sys/mman.rs
528:///     let mem = mmap::<BorrowedFd>(None, one_k_non_zero, ProtFlags::PROT_NONE,

test/sys/test_mman.rs
7:        let ptr = mmap::<BorrowedFd>(
32:        let mem = mmap::<BorrowedFd>(
90:        let mem = mmap::<BorrowedFd>(
```

SteveLauC (changes requested):

SteveLauC:
CI failure seems to be unrelated to this PR

----

Probably two change logs need to be added:

```
### Changed
- `mmap()` no longer takes optional file descriptor as there is a dedicated interface `mmap_anonymous()` for that
   ([#2127](https://github.com/nix-rust/nix/pull/2127))

### Added
- A dedicated interface `mmap_anonymous()` for anonymous mmap has been added
   ([#2127](https://github.com/nix-rust/nix/pull/2127))
```

asomers:
The CI failure is fixed on master.  You need to rebase.

src/sys/mman.rs
newpavlov: It can still be used to allocate memory, e.g. for shared memory mappings based on descriptors received from `shm_open`.

SteveLauC: > It can still be used to allocate memory, e.g. for shared memory mappings based on descriptors received from shm_open.

Thanks for showing me this! :)

SteveLauC:
Code LGTM, still, change logs are needed

SteveLauC:
Friendly ping @newpavlov, could you please add change logs for this PR? We have changed the CHANGELOG mode in #2149, please see [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add change logs

newpavlov:
@SteveLauC 
Done. I've reverted the `mmap` changes since they are breaking and additional changelog note (I am not sure if two changelog notes can be added for one PR). I think it's worth to eventually replace `Option<F>` with just `F`, but it can wait until next breaking release.

SteveLauC:
> Done. 

Thanks for doing this!

> I've reverted the mmap changes since they are breaking 

Breaking changes are allowed as we will bump our minor version number in the next release

> and additional changelog note (I am not sure if two changelog notes can be added for one PR). 

I just tested that we can have two change logs with the same PR number in the current mode, so go ahead:)

SteveLauC (approved):

PR #2128, opened 2023-09-12
Fully disable the personality::set test

saethlin:
It looks like this is failing on x86_64 Linux now too :thinking:


asomers:
The test failures in your other PR seem to be spurious.  Maybe GCE briefly changed its seccomp settings, and then changed them back?  I'm going to close this PR.  But you can reopen it if the test starts to fail again.

Issue #2115, opened 2023-08-28
is BorrowedFd the write API for Epoll::add() etc?


dimbleby:
I'm interacting with a C library so I have a `RawFd` all along.

Previously `epoll_ctl()` etc took a `RawFd` so that was very convenient.  But the new API takes `Fd: AsFd` which is not so convenient: I have to unsafely borrow my `RawFd`; and then nix is only going to `.as_raw_fd()` it again anyway.  It's an awkward and unnecessary dance.

`Fd: AsRawFd` might make more sense, I reckon that's what is actually needed?

#1882, cc @JonathanWoollett-Light

JonathanWoollett-Light:
There is an explanation behind the purpose of `OwnedFd` and `BorrowedFd` here https://rust-lang.github.io/rfcs/3128-io-safety.html.

Nix is moving to this approach, if unfortunately slowly.

JonathanWoollett-Light:
Due to limitation in Rust's compile time evaluation I do not beleive it is currently possible to implement `Epoll` in this form fully safely.

It is true you can add a file descriptor with `Epoll::Add` then close it before closing the epoll or removing it from the interest list. You are correct I agree, this is unsafe, this should be better documented.

I am not sure what code changes would make sense here to better illustrate this.

dimbleby:
I don't think it's only a documentation thing: it's unsafe but the API has no `unsafe` on it.

The suggestion - via polling - is that `Epoll::add()` be `unsafe` and accept a `RawFd`.  

That would be the place to put the documentation saying that callers are responsible for making sure that the file descriptor stays alive up until the point that they `Epoll::delete()` it.

JonathanWoollett-Light:
> The suggestion - via polling - is that Epoll::add() be unsafe and accept a RawFd.

I would suggest using `AsRawFd` so its more ergonomic, this way you could still call it with a reference to an `OwnedFd`.

I do not disagree with adding `unsafe` to it.

> That would be the place to put the documentation saying that callers are responsible for making sure that the file descriptor stays alive up until the point that they Epoll::delete() it.

I do not think this actually matters. If you attempt to de-register a file descriptor that isn't registered it is safe.

The safety concern can be written

> You must gurantee that the file descriptor given to `Epoll::add` is de-registered with `Epoll::del` or the `Epoll` is dropped before the file descriptor is dropped.

JonathanWoollett-Light:
With your update

[...]

PR #2134, opened 2023-09-23
Relax lifetime requirements for PollFd::new

### LINKED ISSUE #2118: PollFd::new requires too restrictive lifetime
I was trying to do

```rust
if let Some(cmd_fd) = state.status_cmd_fd() {
    fds.push(PollFd::new(&cmd_fd, PollFlags::POLLIN));
}
```

where `status_cmd_fd()` returns `Option<BorrowedFd<'_>>`. This code does not compile because the _reference_ `&cmd_fd` does not live long enough, but the lifetime of the reference should be irrelevant.

---

Current signature of `PollFd::new`  is

```rust
pub fn new<Fd: AsFd>(fd: &'fd Fd, events: PollFlags) -> PollFd<'fd>
```

I think it should probably be

```rust
pub fn new<Fd: AsFd + 'fd>(fd: Fd, events: PollFlags) -> PollFd<'fd>
```

asomers:
Fixes #2118

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -22,21 +22,37 @@ pub struct PollFd<'fd> {
 impl<'fd> PollFd<'fd> {
     /// Creates a new `PollFd` specifying the events of interest
     /// for a given file descriptor.
-    //
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # use std::os::fd::{AsFd, AsRawFd, FromRawFd, OwnedFd};
+    /// # use nix::{
+    /// #     poll::{PollFd, PollFlags, poll},
+    /// #     unistd::{pipe, read}
+    /// # };
+    /// let (r, w) = pipe().unwrap();
+    /// let r = unsafe { OwnedFd::from_raw_fd(r) };
+    /// let pfd = PollFd::new(&r.as_fd(), PollFlags::POLLIN);
+    /// let mut fds = [pfd];
+    /// poll(&mut fds, -1).unwrap();
+    /// let mut buf = [0u8; 80];
+    /// read(r.as_raw_fd(), &mut buf[..]);
+    /// ```
+    // Unlike I/O functions, constructors like this must take `AsFd` by
+    // reference.  Otherwise, an `OwnedFd` argument would be dropped at the end
+    // of the method, leaving the structure referencing a closed file
+    // descriptor.
     // Different from other I/O-safe interfaces, here, we have to take `AsFd`
     // by reference to prevent the case where the `fd` is closed but it is
     // still in use. For example:
     //
     // ```rust
-    // let (reader, _) = pipe().unwrap();
-    //
-    // // If `PollFd::new()` takes `AsFd` by value, then `reader` will be consumed,
-    // // but the file descriptor of `reader` will still be in use.
-    // let pollfd = PollFd::new(reader, flag);
-    //
+    // let (r, _) = pipe().unwrap();
+    // let reader: OwnedFd = unsafe { OwnedFd::from_raw_fd(r) };
+    // let pollfd = PollFd::new(reader, flag);  // Drops the OwnedFd
     // // Do something with `pollfd`, which uses the CLOSED fd.
     // ```
-    pub fn new<Fd: AsFd>(fd: &'fd Fd, events: PollFlags) -> PollFd<'fd> {
+    pub fn new<Fd: AsFd + 'fd>(fd: &Fd, events: PollFlags) -> PollFd<'fd> {
         PollFd {
             pollfd: libc::pollfd {
                 fd: fd.as_fd().as_raw_fd(),
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -25,26 +25,22 @@ impl<'fd> PollFd<'fd> {
     ///
     /// # Examples
     /// ```no_run
-    /// # use std::os::fd::{AsFd, AsRawFd, FromRawFd, OwnedFd};
+    /// # use std::os::unix::io::{AsFd, AsRawFd, FromRawFd};
     /// # use nix::{
     /// #     poll::{PollFd, PollFlags, poll},
     /// #     unistd::{pipe, read}
     /// # };
     /// let (r, w) = pipe().unwrap();
-    /// let r = unsafe { OwnedFd::from_raw_fd(r) };
-    /// let pfd = PollFd::new(&r.as_fd(), PollFlags::POLLIN);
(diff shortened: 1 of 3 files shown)


asomers:
cc @SteveLauC

MaxVerevkin:
I think the argument should not be a reference at all, there is `impl<T: AsFd> AsFd for &T`.

src/poll.rs
```diff
@@ -36,7 +36,7 @@ impl<'fd> PollFd<'fd> {
     //
     // // Do something with `pollfd`, which uses the CLOSED fd.
     // ```
-    pub fn new<Fd: AsFd>(fd: &'fd Fd, events: PollFlags) -> PollFd<'fd> {
+    pub fn new<Fd: AsFd + 'fd>(fd: Fd, events: PollFlags) -> PollFd<'fd> {
```
SteveLauC, line 39: If it takes `fd` by value, then we should add a warning to the users in the doc comment to not let them pass the types that are owned file descriptors (e.g., File)

```
    //
    // Different from other I/O-safe interfaces, here, we have to take `AsFd`
    // by reference to prevent the case where the `fd` is closed but it is
    // still in use. For example:
    //
    // ```rust
    // let (reader, _) = pipe().unwrap();
    //
    // // If `PollFd::new()` takes `AsFd` by value, then `reader` will be consumed,
    // // but the file descriptor of `reader` will still be in use.
    // let pollfd = PollFd::new(reader, flag);
    //
    // // Do something with `pollfd`, which uses the CLOSED fd.
    // ```
```

The old comment should also be removed

SteveLauC:
Kinda think we are facing this issue everywhere, here is also a thread discussing this stuff: https://github.com/nix-rust/nix/issues/2115

asomers:
> I think the argument should not be a reference at all, there is `impl<T: AsFd> AsFd for &T`.

Actually, this is subtly wrong.  And it's probably wrong for all constructors.  If `PollFd::new` takes `fd` by value, then it immediately drops it, causing the file descriptor to be closed.  The fact that it currently compiles at all means that there is probably something wrong with the lifetime handling; I'll double check that.

Now that I've reflected on this, the original approach of only taking a reference was a pretty good approach, because usually a `PollFd` isn't useful if it owns the file descriptor.  After all, you usually want to do something else with the file other than just poll it.  If we want `PollFd` to actually own it, then we should probably take an `OwnedFd`.

asomers:
@MaxVerevkin Here's an example that demonstrates why `PollFd` must take its argument by reference.  In this example, `PollFd::new` appears to take its argument by ownership, but it actually drops the `OwnedFd` at the end of the `new` method.  This example compiles with the current state of the branch, but it ought to fail.
```rust
    /// ```compile_fail,E0597
    /// # use std::os::fd::{AsFd, AsRawFd, FromRawFd, OwnedFd};
    /// # use nix::{
    /// #     poll::{PollFd, PollFlags, poll},
    /// #     unistd::{pipe, read}
    /// # };
    /// let pfd = {
    ///     let (r, w) = pipe().unwrap();
    ///     let r = unsafe { OwnedFd::from_raw_fd(r) };
    ///     PollFd::new(r, PollFlags::POLLIN)
    /// };
    /// let mut fds = [pfd];
    /// poll(&mut fds, -1).unwrap();
    /// let mut buf = [0u8; 80];
    /// read(pfd.as_fd().as_raw_fd(), &mut buf[..]);
    /// ```
```
So either we must take by reference, or else the struct should include an `OwnedFd`.  Or, it could take an `OwnedFd` and perform `into_raw`, and we could subsequently to a `OwnedFd::from_raw_fd()` during Drop.  But that's complicated, and it wouldn't be generic for any `AsFd` argument.

MaxVerevkin:
> If PollFd::new takes fd by value, then it immediately drops it, causing the file descriptor to be closed.

Good observation. Yeah, by-value only make sense for `BorrowedFd`...

>  If we want PollFd to actually own it, then we should probably take an OwnedFd.

Please no :)

---

By the way, don't know if it deserves its own issue. How do you reuse the same fd buffer with this new API? If I have a `Vec<PollFd>`, it captures the lifetime and prevents me from using IO resources even if I `.clear()` the buffer. Storing this buffer in a struct along with these IO resources is impossible (self-referencing). This forces me to create a new `Vec` each time in most of my projects, which forced me to "downgrade" to `libc::poll`.

asomers:
Ok, I just pushed changes that restore the reference, and also rewords @SteveLauC 's explanatory comment a bit.

MaxVerevkin:
This doesn't work either:

```rust
let mut fds = Vec::new();

{
    let file = std::fs::File::open("/etc/hostname").unwrap();
    fds.push(PollFd::<'static>::new(&file, PollFlags::POLLIN));
    drop(file);
}

poll(&mut fds, -1).unwrap();
```

This compiles because `File: 'static` is satisfied. But previous approach doesn't work with borrowed fds.

asomers:
> This doesn't work either:
> 
> ```rust
> let mut fds = Vec::new();
> 
> {
>     let file = std::fs::File::open("/etc/hostname").unwrap();
>     fds.push(PollFd::<'static>::new(&file, PollFlags::POLLIN));
>     drop(file);
> }
> 
> poll(&mut fds, -1).unwrap();
> ```
> 
> This compiles because `File: 'static` is satisfied. But previous approach (`&'fd Fd`) doesn't work with borrowed fds.

Ouch.  That's a big problem.  Maybe we should change `Poll::new` to take a `BorrowedFd` by value, just to be sure that the lifetime is specified?

MaxVerevkin:
> Maybe we should change Poll::new to take a BorrowedFd by value

That's definitely _a_ solution. I wanted to propose it but I wasn't sure if it's a good one.

SteveLauC:
> Maybe we should change Poll::new to take a BorrowedFd by value

This approach LGTM

src/poll.rs
```diff
@@ -22,24 +22,36 @@ pub struct PollFd<'fd> {
...
     //
     // ```rust
-    // let (reader, _) = pipe().unwrap();
-    //
-    // // If `PollFd::new()` takes `AsFd` by value, then `reader` will be consumed,
-    // // but the file descriptor of `reader` will still be in use.
-    // let pollfd = PollFd::new(reader, flag);
-    //
+    // let (r, _) = pipe().unwrap();
+    // let reader: OwnedFd = unsafe { OwnedFd::from_raw_fd(r) };
```
SteveLauC, line 47: Considering [`pipe()` now returns `(OwnedFd, OwnedFd)`](https://github.com/nix-rust/nix/pull/2100), this also needs to be updated

SteveLauC (approved):

PR #2136, opened 2023-09-23
Relax lifetime requirements for FdSet::{insert, remove, contains}

### LINKED ISSUE #2130: FdSet is difficulty to use
Code like

```rust
use nix::sys::{ select, signalfd::{self, SigSet} };
fn main() {
    let mut fd_sig = signalfd::SignalFd::new(&SigSet::empty()).unwrap();
    let mut fd_tmp = signalfd::SignalFd::new(&SigSet::empty()).unwrap();

    loop {
	let mut fdset = select::FdSet::new();

	fdset.insert(&fd_tmp);
	fdset.insert(&fd_sig);

	select::select(None, &mut fdset, None, None, None).unwrap();

	if fdset.contains(&fd_sig) {
	    fd_sig.read_signal().unwrap();
	}

	if fdset.contains(&fd_tmp) {
	}
    }
}
```

fails to compile with

```
error[E0502]: cannot borrow `fd_sig` as mutable because it is also borrowed as immutable
  --> src/main.rs:17:6
   |
12 |     fdset.insert(&fd_sig);
   |                  ------- immutable borrow occurs here
...
17 |         fd_sig.read_signal().unwrap();
   |         ^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
20 |     if fdset.contains(&fd_tmp) {
   |        ----- immutable borrow later used here
```

Workaround is to store results of all used `fdset.contains()` in temporary variables like

```rust
        ...
	select::select(None, &mut fdset, None, None, None).unwrap();

        let ev_sig = fdset.contains(&fd_sig);
        let ev_tmp = fdset.contains(&fd_tmp);
```

But this is ugly.

Would it be possible to change the FdSet functions to

```patch
-    pub fn insert<Fd: AsFd>(&mut self, fd: &'fd Fd) {
+    pub fn insert<Fd: AsFd + 'fd>(&mut self, fd: Fd) {
```

?

asomers:
Fixes #2130

### DIFF
--- a/src/sys/select.rs
+++ b/src/sys/select.rs
@@ -72,7 +72,7 @@ impl<'fd> FdSet<'fd> {
     /// # Example
     ///
     /// ```
-    /// # use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
+    /// # use std::os::unix::io::{AsRawFd, BorrowedFd};
     /// # use nix::sys::select::FdSet;
     /// let fd_four = unsafe {BorrowedFd::borrow_raw(4)};
     /// let fd_nine = unsafe {BorrowedFd::borrow_raw(9)};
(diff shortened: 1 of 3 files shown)


asomers:
cc @SteveLauC

SteveLauC:
Seems that these functions should also take a `BorrowedFd`, just like #2134

src/sys/select.rs
```diff
@@ -72,13 +72,13 @@ impl<'fd> FdSet<'fd> {
     /// # Example
     ///
     /// ```
-    /// # use std::os::unix::io::{AsRawFd, BorrowedFd};
+    /// # use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
```
SteveLauC, line 75: This `AsFd` seems to be unused?

SteveLauC (changes requested):

SteveLauC (approved):

Issue #2132, opened 2023-09-23
New maintainers wanted


asomers:
This project needs more help!

I've been the primary maintainer of Nix since 2017.  But I no longer have as much free time as this project deserves.  And frankly, I'm burning out.

New candidates should be experienced Rust programmers with an existing track record of contributions to Nix.

Until we get at least one new maintainer, I'll continue to fix bugs.  But I don't have the time for new feature requests.

asomers:
cc @nix-rust/nix-maintainers .

kamalmarhubi:
Thanks so much for all the work @asomers! It’s been a few years since I’ve done any rust, so I’m kind of out of touch with who might be a good fit :/

SteveLauC:
It would be great to pin this issue so that it gets more attention :)

SteveLauC:
I posted this on [r/rust: nix-rust is looking for new maintainers](https://www.reddit.com/r/rust/comments/16qp3jm/nixrust_is_looking_for_new_maintainers/?utm_source=share&utm_medium=web2x&context=3), hope it will help

woodruffw:
I'll re-up my [offer from 2020](https://github.com/nix-rust/nix/issues/1304#issuecomment-724882696) -- we rely on `nix` at my workplace, and I use it in my personal projects as well. I've made a few (relatively small) contributions over the years, but I'm more than happy to both contribute and review further as the project requires.

SteveLauC:
I am definitely not an experienced Rust programmer, but I have made several contributions to nix in the last year, and nix is actually the first big Rust project I have contributed to, so very willing to help keep this project maintained :)

carlosb1:
I will be happy to help... Does it exist some roadmap?

SteveLauC:
> Does it exist some roadmap?

AFAIK, no roadmap exists

carlosb1:
I see a good set of PRs pending for approvals.  I am not sure how this maintainer role works.. but I don't have problems to help in the revision of all this stuff.

asomers:
I've just invited @SteveLauC to join the @nix-rust/nix-maintainers team.  Welcome aboard!  @carlosb1 , @DaviRain-Su , and @woodruffw given your small track records I did not add you.  But that doesn't mean that you can't contribute!  Please, if you're interested subscribe to the repo and start working on issues and PRs.  You can still submit and approve PRs, just not maintain them.
[...]

PR #2137, opened 2023-09-25
unistd: make getpeereid take AsFd rather than a raw FD

woodruffw:
I noticed that the `sockopt` APIs now take `AsFd`, so I figured this API should be made to do the same :slightly_smiling_face: 

`AsFd` returns a `BorrowedFd` that is correct by construction, meaning that this API rejects more error states at the type level (such as passing in `fd = -1`). As such, I've removed the `test_getpeereid_invalid_fd` testcase, since its error state is no longer reachable from Rust users of this API.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -550,7 +550,11 @@ cfg_if! {
             TcpMaxSeg, GetOnly, libc::IPPROTO_TCP, libc::TCP_MAXSEG, u32);
     }
 }
-#[cfg(not(any(target_os = "openbsd", target_os = "haiku", target_os = "redox")))]
+#[cfg(not(any(
+    target_os = "openbsd",
+    target_os = "haiku",
+    target_os = "redox"
+)))]
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
@@ -572,7 +576,11 @@ sockopt_impl!(
     libc::TCP_REPAIR,
     u32
 );
-#[cfg(not(any(target_os = "openbsd", target_os = "haiku", target_os = "redox")))]
+#[cfg(not(any(
+    target_os = "openbsd",
+    target_os = "haiku",
+    target_os = "redox"
+)))]
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -3943,7 +3943,7 @@ feature! {
     target_os = "netbsd",
     target_os = "dragonfly",
 ))]
-pub fn getpeereid<F: AsFd>(fd: &F) -> Result<(Uid, Gid)> {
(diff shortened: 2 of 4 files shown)


src/sys/socket/sockopt.rs
woodruffw, line 1: NB: Unrelated formatting changes in this file (made automatically). Happy to revert if you'd prefer to leave them out.

SteveLauC:
Ref #1750

src/unistd.rs
```diff
@@ -3943,11 +3943,11 @@ feature! {
     target_os = "netbsd",
     target_os = "dragonfly",
 ))]
-pub fn getpeereid(fd: RawFd) -> Result<(Uid, Gid)> {
+pub fn getpeereid<F: AsFd>(fd: &F) -> Result<(Uid, Gid)> {
```
SteveLauC, line 3946: As discussed [here](https://github.com/nix-rust/nix/issues/1750#issuecomment-1343509652), these functions should take `F: AsFd` by value:

```rust
pub fn getpeereid<F: AsFd>(fd: F) -> Result<(Uid, Gid)> 
```

woodruffw: Thanks! I just saw the discussion in #1750, will refactor now.

woodruffw: https://github.com/nix-rust/nix/pull/2137/commits/22713769c030927d6a2e204430044a1497a29d95

SteveLauC:
BTW, the module `unistd` is partially covered in PR #1863, and it does have some work that is not done yet, see [this comment](https://github.com/nix-rust/nix/issues/1750#issuecomment-1340889060), would you like to finish that part in this PR, which would be great!

woodruffw:
> would you like to finish that part in this PR, which would be great!

It looks like there are some unresolved design questions in that PR's discussion, so I don't want to make this changeset harder to merge by adding those in. 

Does that seem reasonable?

SteveLauC:
Sorry for being unclear, I was suggesting finishing the part that is **missing** from that PR (i.e., the following functions):


>  1. `pub fn tcgetpgrp(fd: c_int) -> Result<Pid>`
>  2. `pub fn tcsetpgrp(fd: c_int, pgrp: Pid) -> Result<()>`
>  3. `pub fn fpathconf(fd: RawFd, var: PathconfVar) -> Result<Option<c_long>>`
>  4. `pub fn ttyname(fd: RawFd) -> Result<PathBuf>`
>  5. `pub fn getpeereid(fd: RawFd) -> Result<(Uid, Gid)>`

It should be okay to simply replace the `RawFd` (or `c_int`) with a trait generic argument, just like what's done in this PR:)

woodruffw:
> Sorry for being unclear, I was suggesting finishing the part that is missing from that PR (i.e., the following functions):

Oh, gotcha! Yeah, I can definitely do that; sorry for the confusion on my part!

woodruffw:
Done -- all 5 of those APIs now take an `AsFd` bounds instead, and I've update the tests.

woodruffw:
Gentle ping on this, now that #2132 is semi-resolved 🙂

asomers (approved):
LGTM.

PR #2138, opened 2023-09-28
sys: add MIPS R6 support

chenx97:
Currently, R6 targets are almost identical to their R2/R5 counterparts. So we add target_arch and bump libc.

### DIFF
--- a/test/test_mq.rs
+++ b/test/test_mq.rs
@@ -112,7 +112,15 @@ fn test_mq_getattr() {
 // FIXME: Fix failures for mips in QEMU
 #[test]
 #[cfg_attr(
-    all(qemu, any(target_arch = "mips", target_arch = "mips64")),
+    all(
+        qemu,
+        any(
+            target_arch = "mips",
+            target_arch = "mips32r6",
+            target_arch = "mips64",
+            target_arch = "mips64r6"
+        )
+    ),
     ignore
 )]
 fn test_mq_setattr() {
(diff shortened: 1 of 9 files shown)


SteveLauC:
Thanks for doing this! I think I will give it a check after dealing with our [CI migration](https://github.com/nix-rust/nix/issues/2161)

chenx97:
I'll migrate my log to towncrier soon

SteveLauC:
Upstream PRs:

* [libc: add mips32r6 and mips64r6 target_arch detection](https://github.com/rust-lang/libc/pull/3268)
* [Rust: add mips64r6 and mips32r6 as target_arch values](https://github.com/rust-lang/rust/pull/112374)

Targets docs: [mipsisa\*r6\*-unknown-linux-gnu*](https://doc.rust-lang.org/nightly/rustc/platform-support/mips-release-6.html)

------

Changes LGTM, but in case something could go wrong, I would like to have @asomers give this a double check

test/test_mq.rs
```diff
@@ -112,7 +112,15 @@ fn test_mq_getattr() {
 // FIXME: Fix failures for mips in QEMU
 #[test]
 #[cfg_attr(
-    all(qemu, any(target_arch = "mips", target_arch = "mips64")),
+    all(
+        qemu,
+        any(
+            target_arch = "mips",
```
asomers, line 118: This would be a good candidate for cfg_aliases, which we have discussed for other purposes.
https://docs.rs/cfg_aliases/latest/cfg_aliases/

asomers (approved):

asomers:
@chenx97 you'll need to rebase your PR to fix CI.

test/test_mq.rs
SteveLauC: If we are gonna replace most big-chunk, common `cfg!`s with some aliases, then it might be better to do it in another, dedicated PR, where we can replace them all and get them well documented in `CONTRIBUTING.md`

SteveLauC: #2188 issue filed

SteveLauC:
There seems to be something wrong with our merge queue, the GHA checks simply don't run, then it reaches the timeout and gets removed from the queue

PR #2140, opened 2023-09-29
Cast pointers more carefully

asomers:
Prefer ptr::{cast, cast_const, cast_mut} over `as`.  The latter makes it too easy to accidentally change both the type and mutability when changing only one was intended.

This exercise caught an unintended mutability cast in one function, the BSD version of sendfile.  In this case there's no UB because it isn't possible for anything else to get a reference to the data that was incorrectly cast.

There was also a type cast that wasn't guaranteed to be correct (but probably was) due to memory layout guarantees in if_nametoindex.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -9,10 +9,10 @@ use crate::sys::time::TimeSpec;
 use crate::sys::time::TimeVal;
 use crate::{errno::Errno, Result};
 use cfg_if::cfg_if;
-use libc::{self, c_int, c_void, size_t, socklen_t};
+use libc::{self, c_int, size_t, socklen_t};
 #[cfg(all(feature = "uio", not(target_os = "redox")))]
 use libc::{
-    iovec, CMSG_DATA, CMSG_FIRSTHDR, CMSG_LEN, CMSG_NXTHDR, CMSG_SPACE,
+    c_void, iovec, CMSG_DATA, CMSG_FIRSTHDR, CMSG_LEN, CMSG_NXTHDR, CMSG_SPACE,
 };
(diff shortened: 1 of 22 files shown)


asomers:
cc @jan561 I was inspired to do this by what I saw in your PR #2120 .  I tried to avoid making any conflicts with that PR.

src/sys/socket/mod.rs
```diff
@@ -487,7 +487,7 @@ cfg_if! {
             pub fn groups(&self) -> &[libc::gid_t] {
                 unsafe {
                     std::slice::from_raw_parts(
-                        self.0.cmcred_groups.as_ptr() as *const libc::gid_t,
+                        self.0.cmcred_groups.as_ptr().cast(),
```
SteveLauC, line 490: This cast seems to be unnecessary? `as_ptr()` already gives us `*const libc::gid_t`

Def: [strcut cmsgcred](https://github.com/rust-lang/libc/blob/6e02a329a2a27f6887ea86952f389ca11e06448c/src/unix/bsd/freebsdlike/mod.rs#L230-L237)

```
    pub struct cmsgcred {
        pub cmcred_pid: ::pid_t,
        pub cmcred_uid: ::uid_t,
        pub cmcred_euid: ::uid_t,
        pub cmcred_gid: ::gid_t,
        pub cmcred_ngroups: ::c_short,
        pub cmcred_groups: [::gid_t; CMGROUP_MAX],
    }
```

asomers: good catch.

asomers:
I made that one change to UnixCredentials, and also rebased on master.

SteveLauC (approved):

PR #2141, opened 2023-09-30
Fix SignalFd::set_mask

### LINKED ISSUE #2116: `SignalFd::set_mask` closes the signal file descriptor
https://github.com/nix-rust/nix/blob/b2e45a930acc636ea14c8ee4c8c13f0c92448d16/src/sys/signalfd.rs#L103-L105

This is because `_signalfd` returns an `OwnedFd` as of https://github.com/nix-rust/nix/pull/1874 (cc @JonathanWoollett-Light ):
https://github.com/nix-rust/nix/blob/b2e45a930acc636ea14c8ee4c8c13f0c92448d16/src/sys/signalfd.rs#L52

This has caused issues in [Smithay/calloop](https://github.com/Smithay/calloop), and required backing out of the nix 0.27 update (+ https://github.com/nix-rust/nix/issues/2112, of course).

This can be worked around in user code by using the newly-deprecated `signalfd` function, so it's not too bad:
```diff
-        file.set_mask(&mask)?;
+        // The nix crate is currently broken. We would like to use:
+        // file.set_mask(&mask)?;
+        // but this closes the file descriptor
+        #[allow(deprecated)]
+        signalfd::signalfd(Some(file.as_fd()), &mask, SfdFlags::empty())
+            .map(std::mem::forget)?;
```

asomers:
In 0.27.0 it inadvertently closed the file descriptor, leaving the SignalFd object accessing a stale file descriptor.

Fixes #2116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Fixed the function signature of `recvmmsg`, potentially causing UB
   ([#2119](https://github.com/nix-rust/nix/issues/2119))
 
+- Fix `SignalFd::set_mask`.  In 0.27.0 it would actually close the file
+  descriptor.
+  ([#2141](https://github.com/nix-rust/nix/pull/2141))
+
 ### Changed
 
 - The following APIs now take an implementation of `AsFd` rather than a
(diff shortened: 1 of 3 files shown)


asomers:
@djmcnab could you please test this patch with your application?

DJMcNab:
This doesn't really impact me, as it was only due to the tests of `calloop`. I'm not set up torun those tests at this exact moment

I don't think the updates in `calloop` should be that bad, if you want to use it for testing - https://github.com/Smithay/calloop/pull/145 doesn't compile because of (effectively) #2112, but that's just adding some derives

SteveLauC (approved):

PR #2142, opened 2023-09-30
fcntl add F_GETPATH support for apple/netbsd

devnexen:


src/fcntl.rs
```diff
@@ -489,6 +489,8 @@ pub enum FcntlArg<'a> {
     F_GETPIPE_SZ,
     #[cfg(any(target_os = "linux", target_os = "android"))]
     F_SETPIPE_SZ(c_int),
+    #[cfg(any(target_os = "netbsd", target_os = "macos", target_os = "ios"))]
+    F_GETPATH(Vec<u8>),
```
SteveLauC, line 493: For arguments like `F_GETXXX`, the associated type should be a mutable reference:

```rust
F_GETPATH(&'a mut Vec<u8>),
```

-----

BTW, do you think it would be better to use a `PathBuf` here since this argument semantically is a path:

```rust
F_GETPATH(&'a mut PathBuf),
```

This approach has an obvious drawback that we can not alter the buffer in place, so we probably have to do another allocation:

```rust
F_GETPATH(path /*: &'a mut PathBuf*/) => {
    let mut another__buf: Vec<u8> =  Vec::with_capacity(libc::PATH_MAX);
    another_buf.resize(libc::PATH_MAX, 0);
    let res = libc::fcntl(fd, libc::F_GETPATH, another_buf.as_ptr().cast());
   
    if res == -1 {
        res
    } else {
        let actual_len = another_buf.iter().position(|byte| *byte == 0).unwrap();
        another_buf.resize(actual_size, 0) // 0 doesn't matter here as `actual_len` is smaller than `PATH_MAX`
        let os_string = OsString::from_vec(another_buf);
        let path_buf = Path::from(os_string);
        *path = path_buf;
        
         res
    }
},
```

src/fcntl.rs
```diff
@@ -549,6 +551,11 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
             F_GETPIPE_SZ => libc::fcntl(fd, libc::F_GETPIPE_SZ),
             #[cfg(any(target_os = "linux", target_os = "android"))]
             F_SETPIPE_SZ(size) => libc::fcntl(fd, libc::F_SETPIPE_SZ, size),
+            #[cfg(any(target_os = "netbsd", target_os = "macos", target_os = "ios"))]
+            F_GETPATH(path) => {
+                path.resize(libc::PATH_MAX);
```
SteveLauC, line 556: `Vec::resize()` takes two arguments, this is the reason why [CI](https://github.com/nix-rust/nix/pull/2142/checks?check_run_id=17282030491) failed

test/test_fcntl.rs
```diff
@@ -561,3 +561,21 @@ mod test_posix_fallocate {
...
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_path() {
+        let mut tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let mut path: Vec<u8> = Vec::new();
+        let res = fcntl(fd, FcntlArg::F_GETPATH(path)).expect("get path failed");
+        assert_eq!(
+            path.len() > 0
```
SteveLauC, line 578: This assertion is kinda useless as the length is `libc::PATH_MAX`, which is [definitely](https://docs.rs/libc/latest/x86_64-apple-darwin/libc/constant.PATH_MAX.html) greater than 0

SteveLauC (changes requested):

SteveLauC:
And, a [changelog](https://github.com/nix-rust/nix/blob/master/CHANGELOG.md) is needed

src/fcntl.rs
devnexen: It we were to choose the PathBuf route, I think

- `let mut another__buf = vec![0; libc::PATH_MAX as usize]` is simpler (and faster) ?
- `libc::fcntl(fd, libc::F_GETPATH, another_buf.as_ptr().cast())`, not sure cast() is necessary here.

SteveLauC: > let mut another__buf = vec![0; libc::PATH_MAX as usize] is simpler (and faster) ?

Yep, this is better

> libc::fcntl(fd, libc::F_GETPATH, another_buf.as_ptr().cast()), not sure cast() is necessary here.

indeed, it should be removed

test/test_fcntl.rs
```diff
@@ -561,3 +561,26 @@ mod test_posix_fallocate {
...
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_path() {
+        let mut tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let mut path: PathBuf::new();
+        let res =
+            fcntl(fd, FcntlArg::F_GETPATH(path)).expect("get path failed");
+        assert_ne!(path.as_path(), Path::new(""));
```
SteveLauC, line 584: what about:

```rust
assert_eq(tmp.path(), path);
```

src/fcntl.rs
```diff
@@ -489,6 +496,8 @@ pub enum FcntlArg<'a> {
     F_GETPIPE_SZ,
     #[cfg(any(target_os = "linux", target_os = "android"))]
     F_SETPIPE_SZ(c_int),
+    #[cfg(any(target_os = "netbsd", target_os = "dragonfly", target_os = "macos", target_os = "ios"))]
+    F_GETPATH(&'a PathBuf),
```
SteveLauC, line 500: It should be a mutable reference

SteveLauC:
Feel free to ping me when you think this is ready for the final review :)

src/fcntl.rs
```diff
@@ -549,6 +558,19 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
             F_GETPIPE_SZ => libc::fcntl(fd, libc::F_GETPIPE_SZ),
             #[cfg(any(target_os = "linux", target_os = "android"))]
             F_SETPIPE_SZ(size) => libc::fcntl(fd, libc::F_SETPIPE_SZ, size),
+            #[cfg(any(target_os = "netbsd", target_os = "macos", target_os = "ios"))]
```
SteveLauC, line 561: Here, `target_os = "dragonfly"` should be added

src/fcntl.rs
```diff
@@ -549,6 +558,20 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+                if res < 0 {
+                    return Errno::result(res);
+                } else {
+                    let len = buffer.iter().position(|b| *b == 0).unwrap();
+                    buffer.truncate(len as usize);
+                    buffer.shrink_to_fit();
+                    *path = PathBuf::from(OsString::from_vec(buffer));
+                    return Errno::result(res);
+                }
+            },
```
SteveLauC, line 574: It might be better to write it like this:

```rust
let mut buffer = vec![0; libc::PATH_MAX as usize];
let res = libc::fcntl(fd, libc::F_GETPATH, buffer.as_ptr());
let ok_res = Errno::result(res)?
let len = buffer.iter().position(|b| *b == 0).unwrap();
buffer.truncate(len as usize);
buffer.shrink_to_fit();
*path = PathBuf::from(OsString::from_vec(buffer));

ok_res
```

test/test_fcntl.rs
```diff
@@ -561,3 +561,26 @@ mod test_posix_fallocate {
...
+    use std::os::unix::io::AsRawFd;
+    use std::path::PathBuf;
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_path() {
+        let tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let mut path = PathBuf::new();
+        let res =
```
SteveLauC, line 582: This `res` variable seems to be unused, and could potentially cause a CI failure, we may want to use `let _ = ` instead or simply ignore it

SteveLauC:
PR LGTM now, BTW, a change log is needed

devnexen:
Yes I waited finishing the code first, still draft ;-)

test/test_fcntl.rs
```diff
@@ -561,3 +561,27 @@ mod test_posix_fallocate {
...
+            fcntl(fd, FcntlArg::F_GETPATH(&mut path)).expect("get path failed");
+        assert_ne!(res, -1);
+        assert_eq!(path.as_path().file_name(), tmp.path().file_name());
```
SteveLauC, line 585: Is that assertion failure the expected behavior? I think we need to figure out why they are giving different paths

devnexen: I wonder why tempnamedfile (another macos oddity) does not store the fullpath, even tough in practice they are the same paths.

devnexen:
> PR LGTM now, BTW, a change log is needed

Nice to see asomers found a good help ;-)

test/test_fcntl.rs
SteveLauC: From this [answer's comment](https://apple.stackexchange.com/a/131080/394687), it seems that `/var` is a symlink to `/private/var`, so if `fcntl()` follows symlink, then the behavior is reasonable.

It would be great if we document this behavior in the test, explaining a bit about why we are only comparing the file names

SteveLauC:
```
   Compiling tempfile v3.8.0
error: unused doc comment
   --> test/test_fcntl.rs:585:9
    |
585 | /         /// We purposely check the filenames, since they are supposed to be unique.
586 | |         /// The left operand tend to have `/private/var` whereas the right does not contains
587 | |         /// the `/private` part even tough in practice they lead to the same file.
    | |_________-------------------------------------------------------------------------^
    |           |
    |           rustdoc does not generate documentation for macro invocations
    |
    = note: `-D unused-doc-comments` implied by `-D warnings`
    = help: to document an item produced by a macro, the macro must produce the documentation as part of its expansion

warning: build failed, waiting for other jobs to finish...
```

Comment starts with `///` is a doc comment, we should use code comment  (`// comment`) instead.

And, it is kinda necessary to mention that `/var` is a symlink to `/private/var` and `fcntl()` follows symlinks, then we can say that `in practice they lead to the same file`

src/fcntl.rs
```diff
@@ -549,6 +558,17 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
             F_GETPIPE_SZ => libc::fcntl(fd, libc::F_GETPIPE_SZ),
             #[cfg(any(target_os = "linux", target_os = "android"))]
             F_SETPIPE_SZ(size) => libc::fcntl(fd, libc::F_SETPIPE_SZ, size),
+            #[cfg(any(target_os = "dragonfly", target_os = "netbsd", target_os = "macos", target_os = "ios"))]
+            F_GETPATH(path) => {
+                let mut buffer = vec![0; libc::PATH_MAX as usize];
+                let res = libc::fcntl(fd, libc::F_GETPATH, buffer.as_ptr());
```
asomers, line 564:
```suggestion
                let res = libc::fcntl(fd, libc::F_GETPATH, buffer.as_mut_ptr());
```

src/fcntl.rs
```diff
@@ -549,6 +558,17 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+                let ok_res = Errno::result(res)?;
+                let len = buffer.iter().position(|b| *b == 0).unwrap();
+                buffer.truncate(len as usize);
+                buffer.shrink_to_fit();
+                *path = PathBuf::from(OsString::from_vec(buffer));
```
asomers, line 569: Instead of finding the NUL by hand, you should probably use `CStr::from_bytes_until_nul`.

src/fcntl.rs
```diff
@@ -549,6 +558,17 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+                ok_res
```
asomers, line 570:
```suggestion
                return ok_res;
```
Return early so you don't call Errno::result a second time below.

test/test_fcntl.rs
```diff
@@ -561,3 +561,31 @@ mod test_posix_fallocate {
...
     }
 }
+
+#[cfg(any(
+    target_os = "dragonfly",
+    target_os = "netbsd",
+    target_os = "macos",
+    target_os = "ios"
+))]
+mod test_apple_netbsd {
```
asomers, line 571: The module isn't very accurately named.  And why do you need a module, anyway?

test/test_fcntl.rs
```diff
@@ -561,3 +561,31 @@ mod test_posix_fallocate {
...
+    use nix::fcntl::*;
+    use std::os::unix::io::AsRawFd;
+    use std::path::PathBuf;
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_path() {
```
asomers, line 578:
```suggestion
    fn test_f_get_path() {
```

test/test_fcntl.rs
```diff
@@ -561,3 +561,31 @@ mod test_posix_fallocate {
...
+        let tmp = NamedTempFile::new().unwrap();
+        let fd = tmp.as_raw_fd();
+        let mut path = PathBuf::new();
+        let res =
+            fcntl(fd, FcntlArg::F_GETPATH(&mut path)).expect("get path failed");
+        assert_ne!(res, -1);
+        // We purposely check the filenames, since they are supposed to be unique.
+        // The left operand tend to have `/private/var` whereas the right does not contains
+        // the `/private` part even tough in practice they lead to the same file, indeed
+        // `/var` is a symlink to `/private/var`.
```
asomers, line 588: This comment is confusing.  Are you saying that there are some symlinks in one path or the other?  If so, I suggest just using `Path::canonicalize` on both sizes.

asomers (changes requested):

test/test_fcntl.rs
asomers: The code is better, but the comment is still confusing, especially because it only applies to OSX.  With the canonicalize in place, I don't think the comment is even necessary.

asomers:
You'll have to raise MSRV to 1.69.0 do use CStr::from_byte_with_nul.

SteveLauC:
The MSRV will be bumped in #2144 so @devnexen you don't need to worry about it in this PR :)

SteveLauC:
> > You'll have to raise MSRV to 1.69.0 do use CStr::from_byte_with_nul.
> 
> The MSRV will be bumped in #2144 so @devnexen you don't need to worry about it in this PR :)

#2144 has been merged, @devnexen you can rebase now :)

test/test_fcntl.rs
SteveLauC: Resolve this as the comment has been removed

CHANGELOG.md
```diff
@@ -94,6 +94,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#2097](https://github.com/nix-rust/nix/pull/2097))
 - Add the ability to set `kevent_flags` on `SigEvent`.
   ([#1731](https://github.com/nix-rust/nix/pull/1731))
+- Added `F_GETPATH` FcntlFlags entry on Apple/NetBSD/DragonflyBSD for `::nix::fcntl`.
```
SteveLauC, line 97: This seems to be in the wrong place, it should be put under:

```
## [Unreleased] - ReleaseDate
### Added
```

SteveLauC (changes requested):

src/fcntl.rs
SteveLauC: It should be `return Ok(ok_res)`;

CHANGELOG.md
devnexen: ah true..

SteveLauC:
For the requested changes in `test_fcntl.rs`, would you like to resolve them? If so, I think this PR is good to be merged.

A test function similar to the following one would be good :)

```rust
#[cfg(any(
    target_os = "dragonfly",
    target_os = "netbsd",
    target_os = "macos",
    target_os = "ios"
))]
fn test_f_get_path() {
    use nix::fcntl::*;
    use std::{os::unix::io::AsRawFd, path::PathBuf};
    use tempfile::NamedTempFile;

    let tmp = NamedTempFile::new().unwrap();
    let fd = tmp.as_raw_fd();
    let mut path = PathBuf::new();
    let res =
        fcntl(fd, FcntlArg::F_GETPATH(&mut path)).expect("get path failed");
    assert_ne!(res, -1);
    assert_eq!(
        path.as_path().canonicalize().unwrap(),
        tmp.path().canonicalize().unwrap()
    );
}
```

SteveLauC (approved):

PR #2143, opened 2023-10-01
chore: add PR template

SteveLauC:
This PR adds a PR template to nix.

One obvious benefit is that, with this, we no longer need to manually remind our contributors to add their change log


asomers (approved):
LGTM.  Speaking of CHANGELOG, it's the greatest source of merge conflicts.  It would be nice to change it someday to fix that.  Perhaps by doing what https://github.com/saltstacksalt does?  That project creates a separate file for each PR's changelog entry, and then rolls them up at release time.

PR #2144, opened 2023-10-01
Raise MSRV to 1.69.0

asomers:
This allows us to use the very useful CStr::from_bytes_until_nul

### DIFF
--- a/src/mount/bsd.rs
+++ b/src/mount/bsd.rs
@@ -396,7 +396,11 @@ impl<'a> Nmount<'a> {
         match Errno::result(res) {
             Ok(_) => Ok(()),
             Err(error) => {
-                let errmsg = CStr::from_bytes_until_nul(&errmsg[..]).ok();
+                let errmsg = if errmsg[0] == 0 {
+                    None
+                } else {
+                    CStr::from_bytes_until_nul(&errmsg[..]).ok()
+                };
                 Err(NmountError::new(error, errmsg))
             }
         }
--- a/src/mount/bsd.rs
+++ b/src/mount/bsd.rs
@@ -396,14 +396,7 @@ impl<'a> Nmount<'a> {
         match Errno::result(res) {
             Ok(_) => Ok(()),
             Err(error) => {
-                let errmsg = match errmsg.iter().position(|&x| x == 0) {
-                    None => None,
-                    Some(0) => None,
-                    Some(n) => {
-                        let sl = &errmsg[0..n + 1];
-                        Some(CStr::from_bytes_with_nul(sl).unwrap())
-                    }
-                };
+                let errmsg = CStr::from_bytes_until_nul(&errmsg[..]).ok();
                 Err(NmountError::new(error, errmsg))
             }
         }
--- a/src/sys/prctl.rs
+++ b/src/sys/prctl.rs
@@ -151,10 +151,11 @@ pub fn get_name() -> Result<CString> {
 
     let res = unsafe { libc::prctl(libc::PR_GET_NAME, &buf, 0, 0, 0) };
 
-    let len = buf.iter().position(|&c| c == 0).unwrap_or(buf.len());
-    let name = CStr::from_bytes_with_nul(&buf[..=len]).map_err(|_| Errno::EINVAL)?;
-
-    Errno::result(res).map(|_| name.to_owned())
+    Errno::result(res).map(|_| {
+        CStr::from_bytes_until_nul(&buf)
+            .map(CStr::to_owned)
+            .map_err(|_| Errno::EINVAL)
+    }).flatten()
 }
 
 /// Sets the timer slack value for the calling thread. Timer slack is used by the kernel to group
--- a/src/sys/prctl.rs
+++ b/src/sys/prctl.rs
@@ -151,11 +151,11 @@ pub fn get_name() -> Result<CString> {
 
     let res = unsafe { libc::prctl(libc::PR_GET_NAME, &buf, 0, 0, 0) };
 
-    Errno::result(res).map(|_| {
+    Errno::result(res).and_then(|_| {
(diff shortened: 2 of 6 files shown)


src/mount/bsd.rs
```diff
@@ -396,14 +396,7 @@ impl<'a> Nmount<'a> {
...
             Err(error) => {
-                let errmsg = match errmsg.iter().position(|&x| x == 0) {
-                    None => None,
-                    Some(0) => None,
-                    Some(n) => {
-                        let sl = &errmsg[0..n + 1];
-                        Some(CStr::from_bytes_with_nul(sl).unwrap())
-                    }
-                };
+                let errmsg = CStr::from_bytes_until_nul(&errmsg[..]).ok();
```
SteveLauC, line 399: The behavior slightly changes compared to the previous code, when `errmsg[0]` is `0`, the new impl will return `Some(an_empty_CStr)` rather than `None`, but it is still consistent with the C code behavior I guess

src/sys/prctl.rs
```diff
@@ -151,10 +151,11 @@ pub fn get_name() -> Result<CString> {
...
 
-    let len = buf.iter().position(|&c| c == 0).unwrap_or(buf.len());
-    let name = CStr::from_bytes_with_nul(&buf[..=len]).map_err(|_| Errno::EINVAL)?;
-
-    Errno::result(res).map(|_| name.to_owned())
+    Errno::result(res).map(|_| {
+        CStr::from_bytes_until_nul(&buf)
+            .map(CStr::to_owned)
+            .map_err(|_| Errno::EINVAL)
+    }).flatten()
```
SteveLauC, line 158: [`Result::flatten()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten) is a nightly API and thus causing CI failures:

```
error[E0658]: use of unstable library feature 'result_flattening'
   --> src/sys/prctl.rs:158:8
    |
158 |     }).flatten()
    |        ^^^^^^^
    |
    = note: see issue #70142 <https://github.com/rust-lang/rust/issues/70142> for more information

For more information about this error, try `rustc --explain E0658`.

Exit status: 101
```

What about writing it like this:

```rust
    Errno::result(res).and_then(|_| {
        CStr::from_bytes_until_nul(&buf)
            .map(CStr::to_owned)
            .map_err(|_| Errno::EINVAL)
    })
```

src/unistd.rs
```diff
@@ -3928,9 +3928,9 @@ pub fn ttyname<F: AsFd>(fd: F) -> Result<PathBuf> {
         return Err(Errno::from_i32(ret));
     }
 
-    let nul = buf.iter().position(|c| *c == b'\0').unwrap();
-    buf.truncate(nul);
-    Ok(OsString::from_vec(buf).into())
+    CStr::from_bytes_until_nul(&buf[..])
```
SteveLauC, line 3931: The `[..]` in `&buf[..]` seems to be unnecessary here

SteveLauC:
The MSRV in `README.md` needs to be updated as well:

> nix is supported on Rust 1.65 and higher. Its MSRV will not be changed in the future without bumping the major or minor version.

BTW, it seems that the version of the next release would be `0.28.0` since we bumped our MSRV in this it

src/mount/bsd.rs
asomers: Good catch.  The C code in the mount command matches the old behavior, so I'll restore it.

asomers:
> BTW, it seems that the version of the next release would be `0.28.0` since we bumped our MSRV in it

We would have to do that anyway, since we've changed the signatures of several functions to add I/O safety.

SteveLauC:
The CI failure of [`Fuchsia x86_64`](https://github.com/nix-rust/nix/pull/2144/checks?check_run_id=17299930649):

```
rustup target add $TARGET
error: toolchain '1.69.0-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-fuchsia'
note: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html
```

seems to be because that the alias `x86_64-fuchsia` for `x86_64-unknown-fuchsia` has been removed?

> [Platform support: tier2](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-without-host-tools)

```
$ rustup target list | grep -i fuchsia
aarch64-unknown-fuchsia
x86_64-unknown-fuchsia (installed)

$ rustup target add x86_64-fuchsia
error: toolchain '1.69.0-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-fuchsia'
note: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html

$ rustup target add x86_64-unknown-fuchsia
info: component 'rust-std' for target 'x86_64-unknown-fuchsia' is up to date
```

SteveLauC:
To fix the CI failure in [`Redox x86_64`](https://github.com/nix-rust/nix/pull/2144/checks?check_run_id=17299931615), importing `std::ffi::OsStr` for all the platforms:

```
    Checking nix v0.27.1 (/tmp/cirrus-ci-build)
error[E0433]: failed to resolve: use of undeclared type `OsStr`
    --> src/unistd.rs:3932:18
     |
3932 |         .map(|s| OsStr::from_bytes(s.to_bytes()).into())
     |                  ^^^^^ use of undeclared type `OsStr`
     |
help: consider importing one of these items
     |
3    + use crate::OsStr;
     |
3    + use std::ffi::OsStr;
     |
```

```rs
// old one
use std::ffi::{CStr, OsString};
use std::ffi::{CString, OsStr};
```

```rs
// new one, the CI failure should be resolved by this
use std::ffi::{CStr, OsStr, OsString};
use std::ffi::CString;
```

SteveLauC (approved):

PR #2151, opened 2023-10-03
Modified return type for ipaddr

### LINKED ISSUE #2003: Illogical return value for `SockaddrIn::ip()` when compared with `SockaddrIn6::ip()`
One returns `in_addr_t` which is integer value while other returns `std::net::Ipv6Addr`. It feels weird that for one type we need to do additional work to change `in_addr_t` to `std::net::Ipv4Addr` which seems counterintuitive as it is not needed for `SockaddrIn6`.

carlosb1:
## What does this PR do
Fixes #2003 

## Checklist:

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -1007,7 +1007,7 @@ impl SockaddrIn {
     /// Returns the IP address associated with this socket address, in native
     /// endian.
     pub fn ip(&self) -> net::Ipv4Addr {
-        net::Ipv4Addr::from(addr.0.sin_addr.s_addr.to_ne_bytes())
+        net::Ipv4Addr::from(self.0.sin_addr.s_addr.to_ne_bytes())
     }
 
     /// Creates a new socket address from IPv4 octets and a port number.
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -36,6 +36,8 @@ use std::hash::{Hash, Hasher};
 use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 use std::{fmt, mem, net, ptr, slice};
+use std::net::{Ipv4Addr,Ipv6Addr};
+use core::mem::transmute;
 
 /// Convert a std::net::Ipv4Addr into the libc form.
 #[cfg(feature = "net")]
@@ -1006,8 +1008,10 @@ pub struct SockaddrIn(libc::sockaddr_in);
 impl SockaddrIn {
     /// Returns the IP address associated with this socket address, in native
     /// endian.
-    pub fn ip(&self) -> net::Ipv4Addr {
-        net::Ipv4Addr::from(self.0.sin_addr.s_addr.to_ne_bytes())
+    pub const fn ip(&self) -> net::Ipv4Addr {
+        let bytes = self.0.sin_addr.s_addr.to_ne_bytes();
+        let (a, b, c, d) = (bytes[0], bytes[1], bytes[2], bytes[3]);
+        Ipv4Addr::new(a, b, c, d)
     }
 
     /// Creates a new socket address from IPv4 octets and a port number.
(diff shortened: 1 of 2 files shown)


src/sys/socket/addr.rs
```diff
@@ -1006,8 +1006,8 @@ pub struct SockaddrIn(libc::sockaddr_in);
 impl SockaddrIn {
     /// Returns the IP address associated with this socket address, in native
     /// endian.
-    pub const fn ip(&self) -> libc::in_addr_t {
-        u32::from_be(self.0.sin_addr.s_addr)
+    pub fn ip(&self) -> net::Ipv4Addr {
```
SteveLauC, line 1009: It seems that we can also retain the `const` here if we choose the following impl:

```rs
    pub const fn ip(&self) -> Ipv4Addr {
        let be_bytes = self.0.sin_addr.s_addr.to_be_bytes();
        let (a, b, c, d) = (be_bytes[0], be_bytes[1], be_bytes[2], be_bytes[3]);
        Ipv4Addr::new(a, b, c, d)
    }
```

SteveLauC:
And for `SockaddrIn6::ip()`, we can also make it const with a similar impl:

```rs
    pub const fn ip(&self) -> Ipv6Addr {
        let bytes = self.0.sin6_addr.s6_addr;
        let [a, b, c, d, e, f, g, h] =
            unsafe { transmute::<_, [u16; 8]>(bytes) };
        
        // Ipv6Addr::new() takes segments in native endian, convert them here
        let [a, b, c, d, e, f, g, h] = [
            u16::from_be(a),
            u16::from_be(b),
            u16::from_be(c),
            u16::from_be(d),
            u16::from_be(e),
            u16::from_be(f),
            u16::from_be(g),
            u16::from_be(h),
        ];

        Ipv6Addr::new(a, b, c, d, e, f, g, h)
    }
```

SteveLauC:
cc @asomers, these two types were added in #1684

carlosb1:
> And for `SockaddrIn6::ip()`, we can also make it const with a similar impl:
> 
> ```rust
>     pub const fn ip(&self) -> Ipv6Addr {
>         let bytes = self.0.sin6_addr.s6_addr;
>         let [a, b, c, d, e, f, g, h] =
>             unsafe { transmute::<_, [u16; 8]>(bytes) };
>         
>         // Ipv6Addr::new() takes segments in native endian, convert them here
>         let [a, b, c, d, e, f, g, h] = [
>             u16::from_be(a),
>             u16::from_be(b),
>             u16::from_be(c),
>             u16::from_be(d),
>             u16::from_be(e),
>             u16::from_be(f),
>             u16::from_be(g),
>             u16::from_be(h),
>         ];
> 
>         Ipv6Addr::new(a, b, c, d, e, f, g, h)
>     }
> ```

`SockaddrIn6::ip()`  is already `const` I doesn't need modifications in this case.

asomers:
> And for `SockaddrIn6::ip()`, we can also make it const with a similar impl:
> 
> ```rust
>     pub const fn ip(&self) -> Ipv6Addr {
>         let bytes = self.0.sin6_addr.s6_addr;
>         let [a, b, c, d, e, f, g, h] =
>             unsafe { transmute::<_, [u16; 8]>(bytes) };
>         
>         // Ipv6Addr::new() takes segments in native endian, convert them here
>         let [a, b, c, d, e, f, g, h] = [
>             u16::from_be(a),
>             u16::from_be(b),
>             u16::from_be(c),
>             u16::from_be(d),
>             u16::from_be(e),
>             u16::from_be(f),
>             u16::from_be(g),
>             u16::from_be(h),
>         ];
> 
>         Ipv6Addr::new(a, b, c, d, e, f, g, h)
>     }
> ```

`mem::transmute` should be a last resort.  I don't think we need it here.  We could do the conversion more verbosely just by addressing the individual octets directly.

SteveLauC (approved):

SteveLauC:
LGTM! Sorry for the late review, thanks for doing this!

SteveLauC:
I forgot to mention that we have changed the CHANGELOG mode in #2149, let me fix this later

PR #2153, opened 2023-10-03
Add `unistd::close_range()`

cpick:
## What does this PR do

This adds `close_range(2)` on Linux which allows efficiently closing (or marking "close on exec") many file descriptors at once without needing procfs.

I have added comments, but am unsure if new tests are warranted.  I have manually tested and used test programs and `strace` to validate that it is working as intended (both with and without various CLOSE_RANGE_* flag permutations.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   MacOS, FreeBSD, DragonflyBSD, Android, iOS and Haiku.
   ([#2074](https://github.com/nix-rust/nix/pull/2074))
 
+- Added `close_range(2)` on Linux
+
 ## [0.27.1] - 2023-08-28
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   ([#2074](https://github.com/nix-rust/nix/pull/2074))
 
 - Added `close_range(2)` on Linux
+  ([#2153](https://github.com/nix-rust/nix/pull/2153))
 
 ## [0.27.1] - 2023-08-28
 
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1124,7 +1124,7 @@ libc_bitflags! {
 /// use nix::unistd::{close_range, CloseRangeFlags};
 ///
 /// let f = tempfile::tempfile().unwrap();
-/// close_range(f.as_raw_fd(), f.as_raw_fd(), CloseRangeFlags::empty()).unwrap();   // Bad!  f will also close on drop!
+/// unsafe { close_range(f.as_raw_fd(), f.as_raw_fd(), CloseRangeFlags::empty()).unwrap() };   // Bad!  f will also close on drop!
 /// ```
 ///
 /// ```rust
@@ -1133,10 +1133,10 @@ libc_bitflags! {
 ///
 /// let f = tempfile::tempfile().unwrap();
 /// let fd = f.into_raw_fd(); // Good.  into_raw_fd consumes f
-/// close_range(fd, fd, CloseRangeFlags::empty()).unwrap();
+/// unsafe { close_range(fd, fd, CloseRangeFlags::empty()).unwrap(); }
 /// ```
 #[cfg(target_os = "linux")]
-pub fn close_range(
+pub unsafe fn close_range(
     first: RawFd,
     last: RawFd,
     flags: CloseRangeFlags,
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1111,6 +1111,30 @@ libc_bitflags! {
 
 /// Close a range of raw file descriptors.
 /// [close_range(2)](https://man7.org/linux/man-pages/man2/close_range.2.html).
+///
+/// Be aware that many Rust types implicitly close-on-drop, including
+/// `std::fs::File`.  Explicitly closing them with this method too can result in
+/// a double-close condition, which can cause confusing `EBADF` errors in
+/// seemingly unrelated code.  Caveat programmer.
+///
+/// # Examples
+///
+/// ```no_run
+/// use std::os::unix::io::AsRawFd;
+/// use nix::unistd::{close_range, CloseRangeFlags};
+///
+/// let f = tempfile::tempfile().unwrap();
+/// close_range(f.as_raw_fd(), f.as_raw_fd(), CloseRangeFlags::empty()).unwrap();   // Bad!  f will also close on drop!
+/// ```
+///
+/// ```rust
+/// use std::os::unix::io::IntoRawFd;
+/// use nix::unistd::{close_range, CloseRangeFlags};
+///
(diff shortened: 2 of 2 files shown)


src/unistd.rs
```diff
@@ -1098,6 +1098,36 @@ pub fn close(fd: RawFd) -> Result<()> {
...
+        CLOSE_RANGE_UNSHARE;
+        /// Set close-on-exec instead of closing file descriptors.
+        CLOSE_RANGE_CLOEXEC;
+    }
+}
+
+/// Close a range of raw file descriptors.
+/// [close_range(2)](https://man7.org/linux/man-pages/man2/close_range.2.html).
+#[cfg(target_os = "linux")]
+pub fn close_range(
```
SteveLauC, line 1115: I think this function should be marked `unsafe` as it is kinda easy to get a double close in Rust, let's say you have created 3 `std::fs::File`s, which have file descriptors 3, 4, 5, then you want to close them with `close_range()`, to avoid a double free, your have to disable the close behavior of `std::fs::File` (or other things that are basically an `OwnedFd`), then the code would be something like this:

```rust
let file1: File1;
let file2: File2;
let file3: File3;

let fd1 = into_raw_fd(file1);
// if you forget to handle the file descriptors between [first, last]. then 
// double close would happen
let _  =  into_raw_fd(file2);
let fd3 = into_raw_fd(file3);

close_range(fd1, fd3, 0);
```
Due to its unsafety, it would be necessary to add a doc to remind the users, just like [`close()`'s document](https://docs.rs/nix/latest/nix/unistd/fn.close.html) does.

------

What about `close(2)`, why isn't it marked `unsafe`?

The current signature of `close()` is also not that safe:

```rust
fn close(fd: RawFd) -> Result<()>
```

And nix is working on [I/O safety](https://github.com/nix-rust/nix/issues/1750), in #1863, we will change its interface to:

```rust
fn close<Fd: IntoRawFd>(fd: Fd) -> Result<()>
```

`IntoRawFd::into_raw_fd(self)` would take `self` by value and `std::mem::forget()` the inner `OwnedFd`, which means that ONLY one `close()` would be called on that fd,  and is thus safe

SteveLauC (changes requested):

SteveLauC:
Thanks for your contribution to nix!

I have left some comments :)

---

> but am unsure if new tests are warranted

It would be great to test if this binding is properly working, we can test it with a single file descriptor with both `first` and `last` set to it

src/unistd.rs
asomers: Yeah, unsafe makes sense here.  Especially because close_range is most useful to use post-fork and pre-exec, which is already a highly unsafe environment.

src/unistd.rs
```diff
@@ -1098,6 +1098,36 @@ pub fn close(fd: RawFd) -> Result<()> {
...
+    first: RawFd,
+    last: RawFd,
+    flags: CloseRangeFlags,
+) -> Result<()> {
+    let res = unsafe {
+        libc::syscall(
```
asomers, line 1121: We shouldn't be using raw syscalls here.  They lack type safety, and they aren't portable.  Instead, you should add `close_range` in a PR to libc.  Then , once that merges, you can add support to Nix.  If you're ambitious, you might consider adding `closefrom` while you're at it.

asomers (changes requested):

CHANGELOG.md
```diff
@@ -56,6 +56,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   MacOS, FreeBSD, DragonflyBSD, Android, iOS and Haiku.
   ([#2074](https://github.com/nix-rust/nix/pull/2074))
 
+- Added `close_range(2)` on Linux
```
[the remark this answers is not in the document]
cpick: I had noticed others included this, but it seemed a bit catch-22 with opening the PR.  I guess it's standard to tack on a second commit after opening?

Added in 5a380d45eebe7259d446f6ad9d40346a90ed59ed

src/unistd.rs
cpick: @SteveLauC:
> I think this function should be marked unsafe as it is kinda easy to get a double close in Rust

I think we're all already on the same page, but just to say it explicitly: double-close is not a memory-safety issue (subsequent operations on the closed FDs would start returning EBADF), but it is an [I/O safety](https://github.com/rust-lang/rfcs/blob/master/text/3128-io-safety.md) issue.

@SteveLauC:
> The current signature of close() is also not that safe:
...
And nix is working on https://github.com/nix-rust/nix/issues/1750, in https://github.com/nix-rust/nix/pull/1863, we will change its interface to:
...

If this crate is working towards I/O safety then I agree this function cannot be made I/O safe and thus must be `unsafe`.

@asomers:
> Especially because close_range is most useful to use post-fork and pre-exec

My use case is actually not post-fork/pre-exec; I'm ensuring there are no unexpected (eg: inherited) FDs before enabling a SECCOMP filter and running some less-trusted code.

It's also possible to use this function with the `CLOSE_RANGE_CLOEXEC` flag in a way that doesn't affect I/O safety, but obviously the interface as-is doesn't restrict the caller to that (safe) use.

I've marked it `unsafe` in  3e5d88108d100b0c8f8483a4871d21d323eed1e

src/unistd.rs
cpick: Ah, I see.  I was following the precedent of other functions in this file (eg: `gettid`, `execveat`, and `pivot_root`).

CHANGELOG.md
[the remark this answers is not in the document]
SteveLauC: >  but it seemed a bit catch-22 with opening the PR. I guess it's standard to tack on a second commit after opening?

Yes, I agree

src/unistd.rs
SteveLauC: For `gettid()` and `execveat()`, see this issue #2156, for `pivot_root`, there seems to be no libc wrapper for it, so we have to use the syscall directly.

-----
`close_range()` has a wrapper in glibc (now sure about other libcs), can you please add it to the upstream [libc](https://github.com/rust-lang/libc) crate and then we can add it in nix

SteveLauC:
BTW, the CI failure is because an `unsafe` function needs a `# Safety` section, it can be added right above this doc:

```
/// Be aware that many Rust types implicitly close-on-drop, including
/// `std::fs::File`.  Explicitly closing them with this method too can result in
/// a double-close condition, which can cause confusing `EBADF` errors in
/// seemingly unrelated code.  Caveat programmer.
```

cpick:
At this point it looks like I'd need to push the same `syscall` code [down into libc](https://github.com/nix-rust/nix/pull/2153#discussion_r1347640344) and the resulting wrapper up here would still be (appropriately) `unsafe`.

This would leave very little benefit for my case for having the code in nix/libc and so I believe I'll just call `syscall` from my code and avoid another round of pull requests across the two crates.

Thank you both for taking the time to review this pull request, it was very pleasant to work with you for a moment and interact with the project.

asomers:
Allright, @cpick .  I'll close this PR for now.  Feel free to reopen if you ever change your mind.

PR #2163, opened 2023-10-07
chore: move cargo fmt to pre-commit script

SteveLauC:
## What does this PR do
1. Move `cargo fmt` task to a pre-commit script, and add a doc for it in `CONTRIBUTING.md`
2. Add a doc on how to skip the CI (Cirrus CI, will add the GitHub Action version when we finish our migration) in `CONTRIBUTING.md`

## Checklist:

[skip ci]

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+function red() {
+	echo "\033[0;31m$@\033[0m"
+}
+
+function green() {
+	echo "\033[0;32m$@\033[0m"
+}
+
+function byellow() {
+	echo "\033[1;33m$@\033[0m"
+}
+
+CHANGED_BY_CARGO_FMT=false
+
+
+echo -e "$(green INFO): Checking code format..."
+
+cargo fmt --all -q -- --check 2>/dev/null
+if [ $? -ne 0 ]; then
+    echo -e "$(byellow WARN): Unformatted code detected"
+    echo -e "$(green INFO): Formatting..."
+    cargo fmt --all
+    CHANGED_BY_CARGO_FMT=true
+fi
+
+if ${CHANGED_BY_CARGO_FMT}; 
+then
+	echo -e "$(red FAIL): git commit $(red ABORTED), please have a look and run git add/commit again"
+	exit 1
+else
+    echo -e "$(green INFO): Check passed"
+fi
+
+exit 0
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -151,7 +151,6 @@ task:
         - FreeBSD 14 amd64 & i686
         - Linux x86_64
         - macOS aarch64
-        - Rust Formatter
         - OpenBSD x86_64
         - Minver
         - Rust Stable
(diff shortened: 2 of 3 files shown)


SteveLauC:
Kinda weird, the Cirrus CI still runs even though I explicitly `[skip ci]`

Jan561:
It seems you have to include `[skip ci]` in the title of the PR.

```
If you open a PR, PR title will be checked for [skip ci] or [ci skip]
instead of the last commit message on the PR branch.
```

https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution

asomers:
I don't know about you, but personally I would find it surprising if the pre-commit script actually changed my code, as opposed to merely checking it.  Also, I don't think that the addition of a pre-commit script absolves us from running rustfmt in CI.  Because what if some contributor doesn't run the pre-commit script?  We have no way to force them to.

SteveLauC:
> Also, I don't think that the addition of a pre-commit script absolves us from running rustfmt in CI. Because what if some contributor doesn't run the pre-commit script? We have no way to force them to.

Then it seems that we have to add rustfmt to our CI, because AFAIK, there is no way to make this check compulsory with a pre-commit script

SteveLauC:
> It seems you have to include `[skip ci]` in the title of the PR.
> 
> ```
> If you open a PR, PR title will be checked for [skip ci] or [ci skip]
> instead of the last commit message on the PR branch.
> ```
> 
> https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution
> 
> Still weird though as it worked in #2158.

Yeah, it is indeed weird :<

pre-commit.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+function red() {
```
asomers, line 3: The `function` is bash syntax, not sh.
```suggestion
red() {
```

pre-commit.sh
```diff
@@ -0,0 +1,36 @@
...
+function byellow() {
+	echo "\033[1;33m$@\033[0m"
+}
+
+CHANGED_BY_CARGO_FMT=false
+
+
+echo -e "$(green INFO): Checking code format..."
+
+cargo fmt --all -q -- --check 2>/dev/null
```
asomers, line 20: This command doesn't actually work.  I just tested it by screwing up the format in src/sys/aio.rs, and this command didn't complain.  I think it's caused by this bug: https://github.com/rust-lang/rustfmt/issues/3253 .  That's why we need to do `cargo fmt --all -- --check **/*.rs`

asomers (changes requested):

SteveLauC:
For the format check, are we gonna use a pre-commit script or do it in CI? 

#2166 has migrated it to GHA, so if we choose the latter, then we can simply close this PR

asomers:
Yeah, I think we can close this PR if we merge the other one too.  The pre-commit.sh script _could_ be useful too, but maybe not useful enough to justify its ongoing maintenance.

SteveLauC:
Close this PR

PR #2164, opened 2023-10-07
sys adding subset of procctl support for FreeBSD.

devnexen:
- Using the PROC_TRACE_CTL ability to enable/disable process tracing.
- Using the PROC_PDEATHSIG_CTL ability to deliver a signal on parent process exit.

### DIFF
--- a/changelog/2164.added.md
+++ b/changelog/2164.added.md
@@ -1,0 +1,1 @@
+Added the `procctl` process feature to `::nix::sys` for FreeBSD.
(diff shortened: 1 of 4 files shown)


SteveLauC (commented):
Any reason why we are not binding the `procctl(2)`?

And this syscall is also available on DragonFlyBSD, we can use `#[cfg(freebsdlike)]` here

vishwin:
How involved would integrating this into the existing `prctl` be? Some crates explicitly call for `nix::sys::prctl` for this functionality, so they would need additional scaffolding to support FreeBSD and DragonFly BSD if this was separate.

PR #2169, opened 2023-10-10
Add apple tvos support

lcruz99:
## What does this PR do
This PR allows using nix lib to target [tier 3 *-apple builds](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-tvos.html):

- Apple tvOS on aarch64
- Apple tvOS Simulator on x86_64

## Checklist:


lcruz99:
tvOS build is failing because `parking_lot` 0.12.0 doesn't yet support this target, next release will do though.

Will drop CI job commit.

src/sys/event.rs
```diff
@@ -297,10 +307,13 @@ libc_bitflags!(
...
         #[allow(missing_docs)]
         NOTE_MSECONDS;
-        #[cfg(any(target_os = "macos", target_os = "ios"))]
+        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))]
         #[allow(missing_docs)]
         NOTE_NONE;
-        #[cfg(any(target_os = "macos", target_os = "ios", target_os = "freebsd"))]
+        #[cfg(any(
+            target_os = "macos",
+            target_os = "ios",
```
asomers, line 315: Should there be a tvos here?

asomers (changes requested):
Instead of adding more code to support tvos, how about we remove code?  By replacing every instance of:
`#[cfg(any(target_os = "macos", target_os = "ios"))]`
with
`#[cfg(target_vendor = "apple")]`.
I think that should work.

SteveLauC:
> Instead of adding more code to support tvos, how about we remove code? By replacing every instance of: `#[cfg(any(target_os = "macos", target_os = "ios"))]` with `#[cfg(target_vendor = "apple")]`. I think that should work.

Will `target_vendor = "apple"` contain the `watchos`? I didn't find any doc on what this `target_vendor` will be but per the [Platform Support](https://doc.rust-lang.org/beta/rustc/platform-support/apple-watchos.html) doc, `watchos` is also one of the Apple OSes that Rust supports

Another thing I found while searching is that the Rust team will [deprecate the `target_vendor` attribute](https://github.com/rust-lang/lang-team/issues/102) in the future, though I believe it is still usable for now

asomers:
> Another thing I found while searching is that the Rust team will [deprecate the `target_vendor` attribute](https://github.com/rust-lang/lang-team/issues/102) in the future, though I believe it is still usable for now

That is highly annoying.  And it looks like the proposed `target_family = apple` doesn't exist yet.  So deprecating `target_vendor` is a real step backwards in terms of usability.  I wonder, can we create a build.rs script that defines our own cfg variable?  Something like `#[cfg(apple)]` when building for any apple target?

SteveLauC:
> I wonder, can we create a build.rs script that defines our own cfg variable? Something like #[cfg(apple)] when building for any apple target?

There is a crate [`cfg_aliases`](https://github.com/katharostech/cfg_aliases/tree/master), which seems to be exactly what we want:

```
// in our `build.rs`

use cfg_aliases::cfg_aliases;

fn main() {
    cfg_aliases! {
        ios: { target_os = "ios" },
        macos: { target_os = "macos" },
        watchos: { target_os = "watchos" },
        tvos: { target_os = "tvos" },
        apple_targets: { any(ios, macos, watchos, tvos) },
    }
}
```

Then we can use `cfg(apple_targets)` to conditionally compile for these apple targets

And it seems to be compatiable with the cfg attributes from `std`, we can mix them in our code:

```rs
// For Apple targets and Linux 
#[cfg(any(apple_targets, target_os = "linux")]
fn foo() { }
```

asomers:
Awesome @SteveLauC that looks perfect

SteveLauC:
Friendly ping @lcruz99, would you like to do it in this PR:

1.  Add that code snippet to `build.rs`
2. Replace `#[cfg(any(target_os = "macos", target_os = "ios"))]` with `#[cfg(apple_targets)]`

Please note that this would also make nix support `watchos` as well

------

If I do this, then this PR would be replaced by my PR, and I don't want to ruin your first contribution to nix :)

lcruz99:
@SteveLauC will push those changes asap, thanks for the heads up :)

SteveLauC (approved):

SteveLauC:
Thanks! And congratulations on your first contribution to nix!

PR #2170, opened 2023-10-10
Added `Flock` object for automatic unlock on drop.

coderBlitz:
## What does this PR do
In the spirit of RAII, created the `Flock` newtype to simplify `flock` lock handling. Users can pass around an explicit representation of a held lock, automatically unlocking once it goes out of scope.

## Checklist:

- [ x ] I have read `CONTRIBUTING.md`
- [ x ] I have written necessary tests and rustdoc comments

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -631,6 +631,42 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
 
     Errno::result(res).map(drop)
 }
+
+/// Represents a file lock on a particular [RawFd], which unlocks when dropped.
+///
+/// See [fcntl::flock] for details on locking semantics.
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub struct Flock(RawFd);
+
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+impl Drop for Flock {
+    fn drop(&mut self) {
+        _ = flock(self.0, FlockArg::Unlock);
+    }
+}
+
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+impl Flock {
+    /// Lock the given `fd` using [flock].
+    ///
+    /// # Example
+    /// ```
+    /// # use std::os::fd::RawFd;
+    /// # use nix::fcntl::{Flock, FlockArg};
+    /// fn do_stuff(fd: RawFd) -> nix::Result<()> {
+    ///     let lock = Flock::lock(fd, FlockArg::LockExclusive)?;
+    ///
+    ///     // Do stuff
+    ///
+    ///     Ok(())
+    /// } // File is unlocked once `lock` goes out of scope.
+    pub fn lock(fd: RawFd, args: FlockArg) -> Result<Self> {
+        flock(fd, args)?;
+
+        Ok(Self(fd))
+    }
+}
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]


asomers (changes requested):
There's little point in wrapping a `RawFd`, since `RawFd` implements `Copy`.  It would be way too easy to use accidentally use the `RawFd` after dropping the lock.  If anything, it should wrap `OwnedFd`.  And if we're going to do this, then we should make the `flock` function private.

SteveLauC:
> There's little point in wrapping a `RawFd`, since `RawFd` implements `Copy`. It would be way too easy to use accidentally use the `RawFd` after dropping the lock. If anything, it should wrap `OwnedFd`. And if we're going to do this, then we should make the `flock` function private.

I kinda think wrapping an `OwnedFd` is not that easy to use, Let's say that I locked a file exclusively so that I can safely write to it:

```rs
fn main() {
    let mut file = File::create("foo").unwrap();

    let flock_guard = Flock::lock(file.into(), FlockArg::LockExclusive).unwrap();

    // error[E0382]: borrow of moved value: `file`
    file.write(b"bar").unwrap();

    // guard dropped here
}
```

I have to convert the `File` into an `OwnedFd` (moved), which means I can not use that `File` anymore

---------

What about we use a `BorrowedFd` here:

```rs
pub struct Flock<'fd>(BorrowedFd<'fd>);

impl<'fd> Drop for Flock<'fd> {
    fn drop(&mut self) {
        _ = flock(self.0.as_raw_fd(), FlockArg::Unlock);
    }
}

impl<'fd> Flock<'fd> {
    pub fn lock(fd: BorrowedFd<'fd>, args: FlockArg) -> anyhow::Result<Self> {
        flock(fd.as_raw_fd(), args)?;

        Ok(Self(fd))
    }
}
```

Though this won't make the above code compile either:

```rs
    let mut file = File::create("foo").unwrap();

    let flock_guard =
        Flock::lock(file.as_fd(), FlockArg::LockExclusive).unwrap();
    // E0502: cannot borrow `file` as mutable because it is also borrowed as immutable
    file.write(b"bar").unwrap();
```

But we can do this:

```rs
    let file = File::create("foo").unwrap();
    
    let flock_guard =
        Flock::lock(file.as_fd(), FlockArg::LockExclusive).unwrap();

    // compiles without any issue
    (&file).write(b"bar").unwrap();

```

SteveLauC:
ref: #1750

coderBlitz:
If `flock` is only valid on files, then perhaps we could accept a `File` and implement `Deref` and `DerefMut` instead:

```rs
pub struct Flock(File);

impl Drop for Flock {
    fn drop(&mut self) {
        _ = flock(self.0.as_raw_fd(), FlockArg::Unlock);
    }
}

impl Deref for Flock {
    type Target = File;
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
impl DerefMut for Flock {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}

impl Flock {
    pub fn lock(fd: File, args: FlockArg) -> Result<Self> {
        flock(fd.as_raw_fd(), args)?;

        Ok(Self(fd))
    }
}
```

SteveLauC:
> then perhaps we could accept a File and implement Deref and DerefMut instead

Yeah, I agree this is the Rusty way to implement it

> If flock is only valid on files

`flock(2)` can be used for other types of file descriptors

coderBlitz:
> `flock(2)` can be used for other types of file descriptors

Could you provide some examples or cite what other things it can be used on? I'm looking through a bunch of linux kernel and posix docs and they only specifically call out "files" (and local VFS files at that).

I could understand an argument that `File` isn't the only way such files could be represented, however I have yet to find conclusive evidence that anything other than "files" permits `flock`'s.

SteveLauC:
Just checked the FreeBSD [manual](https://man.freebsd.org/cgi/man.cgi?query=flock&sektion=2) since this syscall is not POSIX but comes from BSD implementations:

```
 [EINVAL]		  The argument fd refers to an	object	other  than  a file.
```

This is not specified in the Linux manual, and I can call `flock(2)` on a file descriptor created by `pipe(2)`:

```c
#include <stdio.h>
#include <stdlib.h>
#include <sys/file.h>
#include <unistd.h>

int main(void) {
  int fd[2] = {0};
  int res = pipe(fd);

  if (res != 0) {
    perror("pipe(2)");
    exit(EXIT_FAILURE);
  }

  res = flock(fd[0], LOCK_EX);
  if (res != 0) {
    perror("flock(2)");
    exit(EXIT_FAILURE);
  }


  printf("flock acquired on a pipe fd\n");
  exit(EXIT_SUCCESS);
}
```
```sh
$ uname -a
Linux fedora 6.5.9-300.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Oct 25 21:39:20 UTC 2023 x86_64 GNU/Linux

$ gcc main.c

$ ./a.out
flock acquired on a pipe fd
```

> I think we will get an EINVAL error if we try this program on FreeBSD

------

I agree that most manuals of `flock(2)` are just talking about files

> I didn't see any conclusive doc on whether it can be used on fds that are not files either, so my original statement shouldn't be considered accurate, I am sorry about this

but for systems that don't prohibit this, given that `flock` is assocaited with the open file objects, if the corresponding file descriptor has a entry in the open file table, then I guess it will work?

asomers:
So we can't wrap `RawFd` since it's `Copy`.  And we can't do `OwnedFd` because it doesn't have enough methods.  We don't want to do `BorrowedFd` because the borrowed fd could still be accessed after dropping the `Flock`.  But we don't want to limit ourselves to `std::fs::File` because that would exclude stuff like `tokio::fs::File`.  What about being generic, and wrapping anything that isn't `Clone`?  Unfortunately, Rust can't have negative type bounds.  But we can get close by defining an unsafe `Flockable` trait and allowing the user to implement it on anything that isn't `Clone`.  Like this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1c1ec124c76302e36f5ae4e39ef2b9f3

coderBlitz:
I think everything is good to go, but looks like the gnu test had a random failure (it succeeded in the previous runs).

SteveLauC:
> But we don't want to limit ourselves to std::fs::File because that would exclude stuff like tokio::fs::File.

If we want to implement `Flockable` for `tokio::fs::File`, then it seems that we have to pull the Tokio crate to Nix, is this what we want?

If we want it to be generic without pulling third-party dependencies, i.e.

```rs
unsafe impl<T: AsRawFd> Flockable for T { }
```

Then the unsafe mark seems to be meaningless

SteveLauC:
BTW, we have migrated our most CI tasks to GHA, the CI in this PR is in a mixed, intermediate state, rebasing your branch should resolve it

src/fcntl.rs
```diff
@@ -631,6 +634,102 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+    ///
+    ///     // Do critical section
+    ///
+    ///     // Unlock
+    ///     let file = lock.unlock();
+    ///
+    ///     // Do anything else
+    ///
+    ///     Ok(())
+    /// } // File is unlocked once `lock` goes out of scope.
```
SteveLauC, line 721:
```suggestion
    /// }
```

src/fcntl.rs
```diff
@@ -629,6 +632,102 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
 
     Errno::result(res).map(drop)
 }
+
+/// Represents valid types for flock.
+///
+/// # Safety
+/// `T` must be `!Clone`.
```
SteveLauC, line 639: Nit: might be better to rephrase this:

> Types implementing this must be `!Clone`

cause there is no `T` defined in this trait

src/fcntl.rs
```diff
@@ -629,6 +632,102 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+    ///     let lock = match Flock::lock(file, FlockArg::LockExclusive) {
+    ///         Ok(l) => l,
+    ///         Err((_, e)) => return Err(e),
+    ///     };
+    ///
+    ///     // Do stuff
+    ///
+    ///     Ok(())
+    /// } // File is unlocked once `lock` goes out of scope.
+    pub fn lock(t: T, args: FlockArg) -> std::result::Result<Self, (T, Errno)> {
```
SteveLauC, line 692: Given that the `FlockArg` type is only used for locking, should we remove those 2 `UnLock` variants from it?

SteveLauC: Well, I don't know why there is a variant `UnlockNonblock`, from the manual of FreeBSD, that `LOCK_NB` flag should be used with `LOCK_SH` or `LOCK_EX`

> A lock	is applied by specifying an operation argument that is one of	`LOCK_SH` or  `LOCK_EX`  with the optional	addition of  `LOCK_NB`.

The Linux manual does not mention what will happen if we use `LOCK_UN` and `LOCK_NB` together

Friendly cc @agrover, this type was added in #173

src/fcntl.rs
```diff
@@ -629,6 +632,102 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+        _ = flock(self.0.as_raw_fd(), FlockArg::Unlock);
+
+        // Safety: `self.0` never used again.
+        unsafe { ManuallyDrop::take(&mut self.0) }
+    }
+}
+
+// Safety: `File` is not [std::clone::Clone].
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+unsafe impl Flockable for std::fs::File {}
```
SteveLauC, line 730: Seems that we should also do this for `OwnedFd`:

```rs
// Safety: `OwnedFd` is not [std::clone::Clone].
#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
unsafe impl Flockable for std::os::fd::OwnedFd {}
```

SteveLauC (changes requested):

src/fcntl.rs
SteveLauC: BTW, if we are gonna change the `FlockArg` type, we can just re-define it, using `libc_bitflags!`:

```rs
libc_bitflags! {
    /// The `args` argument of [`Flock::lock()`].
    pub struct LockOperation: c_int {
        /// Place a shared lock
        LOCK_SH;
        /// Place an exclusive lock
        LOCK_EX;
        /// When included, the locking request will be nonblocking.
        LOCK_NB;
    }
}
```

And there is little point in keeping the `flock()` function, we can just remove it and call `libc::flock()` when implementing the `Flock` type

coderBlitz: > Well, I don't know why there is a variant `UnlockNonblock`, from the manual of FreeBSD, that `LOCK_NB` flag should be used with `LOCK_SH` or `LOCK_EX`
> 
> > A lock	is applied by specifying an operation argument that is one of	`LOCK_SH` or  `LOCK_EX`  with the optional	addition of  `LOCK_NB`.
> 
> The Linux manual does not mention what will happen if we use `LOCK_UN` and `LOCK_NB` together
> 
> Friendly cc @agrover, this type was added in #173

All it says is "To make a nonblocking request, include LOCK_NB (by ORing) with any of the above operations.", so it is a valid request. Having said that, there are no strong indications of a failure case for unlocking (one could argue that EWOULDBLOCK still applies), so it's probably a no-op in most cases (I think flock is a per-driver implementation).

> Given that the `FlockArg` type is only used for locking, should we remove those 2 `UnLock` variants from it?
> ...
> And there is little point in keeping the `flock()` function, we can just remove it and call `libc::flock()` when implementing the `Flock` type

If `nix` provides nice wrappers for things, shouldn't we leave it? We would also want to revert it to `pub` if we keep it, so people have the option (if they are doing something weird).

As an aside, I personally like the enum as-defined vice the libc bitflags, because it's more rust-like imo.

src/fcntl.rs
coderBlitz: Should we make it generic then:
```rust
// Safety: `OwnedFd` is not [std::clone::Clone].
#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
unsafe impl<T: AsRef<OwnedFd>> Flockable for T {}
```

src/fcntl.rs
SteveLauC: > All it says is "To make a nonblocking request, include LOCK_NB (by ORing) with any of the above operations.", so it is a valid request. Having said that, there are no strong indications of a failure case for unlocking (one could argue that EWOULDBLOCK still applies), so it's probably a no-op in most cases (I think flock is a per-driver implementation).

Makes sence

> If nix provides nice wrappers for things, shouldn't we leave it? We would also want to revert it to pub if we keep it, so people have the option (if they are doing something weird).

For the `Flock::lock()` interface, we should remove that 2 unlock variants from it cause we don't want our users to do such things:

```rs
let lock = Flock::lock(file, FlockArg::Unlock).unwrap();
```

>  We would also want to revert it to pub if we keep it, so people have the option (if they are doing something weird).

IMHO, It isn't that good to provide 2 interfaces for the same thing. If people really want to do something weird, they can always choose the libc bindings (and nix has libc re-exported)


> As an aside, I personally like the enum as-defined vice the libc bitflags, because it's more rust-like imo.

I prefer the libc_flags way as it adheres to our [CONVENTIONS](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#bitflags), and there is exactly a [TODO](https://github.com/nix-rust/nix/blob/dbdcfed86e18673c71b54d4c4cd7be289441843c/src/fcntl.rs#L599) for it

coderBlitz: > For the Flock::lock() interface, we should remove that 2 unlock variants...

Agreed.

> IMHO, It isn't that good to provide 2 interfaces for the same thing. If people really want to do something weird, they can always choose the libc bindings (and nix has libc re-exported)

Fair enough.

> I prefer the libc_flags way as it adheres to our [CONVENTIONS](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#bitflags), and there is exactly a [TODO](https://github.com/nix-rust/nix/blob/dbdcfed86e18673c71b54d4c4cd7be289441843c/src/fcntl.rs#L599) for it

Preference and convention aren't inherently linked lol. Regardless, I would argue an enum type fits this use case just as well, if not better than the bitflags. Specifically, because all (sans `LOCK_NB`) the arguments to `flock()` are "constants that are intended as mutually exclusive arguments to parameters of functions", I think an enum is the best fit. Especially since non-blocking is the only change, I think having the extra enum variants is worth it given there's only 4 total.

src/fcntl.rs
SteveLauC: [There seems to be only 3 types](https://doc.rust-lang.org/std/convert/trait.AsRef.html) that implmemt this `AsRef<OwnedFd>` in the std:

1. Box<OwnedFd>
2. Rc<OwnedFd>
3. Arc<OwnedFd>

`OwnedFd` is not one of them

And using `AsRef` would allow us to use `&T` where `T` is `AsRef<OwnedFd>`, `&T` is Copy, this does not seem to be what we want:

> There's little point in wrapping a RawFd, since RawFd implements Copy. It would be way too easy to use accidentally use the RawFd after dropping the lock.

coderBlitz: Ah, I confused blanket implementations and didn't realize AsRef doesn't implement itself for all `T`. Nevermind then.

src/fcntl.rs
SteveLauC: > constants that are intended as **mutually exclusive** arguments to parameters of functions

This makes sense to me, we should use an enum here

SteveLauC: And, please remove that TODO as it is no longer planned

SteveLauC:
CI / tier3 (armv7-unknown-linux-uclibceabihf) failure is not related to this PR, please ignore it

changelog/2170.removed.md
```diff
@@ -0,0 +1 @@
+Made `fcntl::flock` private in favor of new Flock wrapper.
```
SteveLauC, line 1: This needs an update

src/fcntl.rs
```diff
@@ -596,39 +599,120 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+        let flags = match args {
+            FlockArg::LockShared => libc::LOCK_SH,
+            FlockArg::LockExclusive => libc::LOCK_EX,
+            FlockArg::LockSharedNonblock => libc::LOCK_SH | libc::LOCK_NB,
+            FlockArg::LockExclusiveNonblock => libc::LOCK_EX | libc::LOCK_NB,
+        };
+        match unsafe { libc::flock(t.as_raw_fd(), flags) } {
+            0 => Ok(Self(ManuallyDrop::new(t))),
+            e => Err((t, Errno::from_i32(e))),
         }
```
SteveLauC, line 677: We should do this, `e` is not a value of Errno but -1:

```rs
        match Errno::result(unsafe { libc::flock(t.as_raw_fd(), flags) }) {
            Ok(_) => Ok(Self(ManuallyDrop::new(t))),
            Err(errno) => Err((t, errno))
        }
```

SteveLauC (changes requested):

src/fcntl.rs
```diff
@@ -596,39 +599,120 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+// `ManuallyDrop` is necessary to circumvent move out of `Drop` type error.
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+#[derive(Debug)]
+pub struct Flock<T: Flockable>(ManuallyDrop<T>);
+
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+impl<T: Flockable> Drop for Flock<T> {
+    fn drop(&mut self) {
+        // Result is ignored because flock has no documented failure cases.
+        _ = unsafe { libc::flock(self.0.as_raw_fd(), libc::LOCK_UN) };
```
SteveLauC, line 630: We should call `ManuallyDrop::drop(&mut self.0)` here or the inner `T` will never be dropped

SteveLauC (changes requested):

src/fcntl.rs
SteveLauC: Well, we should not call it or the following code will be a double drop:

```rs
let file = File;
let lock = Flock::lock(file, arg);
let file = lock.unlock() // file dropped within `lock.unlock()`


// file dropped here
```

SteveLauC: I kinda of think `ManuallyDrop` is a bad choice here, what about using `Flock<Option<T>>` here, this should solve the issue

coderBlitz: Concur. See latest changes.

src/fcntl.rs
```diff
@@ -596,39 +597,125 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
-                libc::flock(fd, libc::LOCK_SH | libc::LOCK_NB)
-            }
-            LockExclusiveNonblock => {
-                libc::flock(fd, libc::LOCK_EX | libc::LOCK_NB)
-            }
-            UnlockNonblock => libc::flock(fd, libc::LOCK_UN | libc::LOCK_NB),
+/// Represents an owned flock, which unlocks on drop.
+///
+/// See flock(2) for details on locking semantics.
+// `ManuallyDrop` is necessary to circumvent move out of `Drop` type error.
```
SteveLauC, line 619: This comment needs an update

SteveLauC (changes requested):

src/fcntl.rs
```diff
@@ -596,39 +597,125 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
     LockExclusiveNonblock,
-    UnlockNonblock,
 }
 
+/// Represents valid types for flock.
+///
+/// # Safety
+/// Types implementing this must be `!Clone`.
 #[cfg(not(any(target_os = "redox", target_os = "solaris")))]
-pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
```
asomers, line 612: I would rather not remove this function suddenly.  Better would be to leave it in but deprecate it for a release or two.

src/fcntl.rs
```diff
@@ -596,39 +597,125 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+// `ManuallyDrop` is necessary to circumvent move out of `Drop` type error.
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+#[derive(Debug)]
+pub struct Flock<T: Flockable>(Option<T>);
+
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+impl<T: Flockable> Drop for Flock<T> {
+    fn drop(&mut self) {
+       if let Some(ref t) = self.0 {
+           // Result is ignored because flock has no documented failure cases.
```
asomers, line 626: What do you mean?  Several failure cases are described in the man page.  And there's always a chance of an undocumented error.

src/fcntl.rs
```diff
@@ -596,39 +597,125 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+    ///
+    ///     // Unlock
+    ///     let file = lock.unlock();
+    ///
+    ///     // Do anything else
+    ///
+    ///     Ok(())
+    /// }
+    pub fn unlock(mut self) -> T {
+        _ = unsafe { libc::flock(self.0.as_ref().unwrap().as_raw_fd(), libc::LOCK_UN) };
```
asomers, line 704: You should not ignore the error code.  There could still be, for example, an EBADF error if somebody implemented Flockable on a Clone type, contrary to our documentation.

src/fcntl.rs
```diff
@@ -596,39 +596,124 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
-            LockExclusiveNonblock => {
-                libc::flock(fd, libc::LOCK_EX | libc::LOCK_NB)
-            }
-            UnlockNonblock => libc::flock(fd, libc::LOCK_UN | libc::LOCK_NB),
+/// Represents an owned flock, which unlocks on drop.
+///
+/// See flock(2) for details on locking semantics.
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+#[derive(Debug)]
+pub struct Flock<T: Flockable>(Option<T>);
```
asomers, line 620: Why Option?  When will it ever be None?

asomers (changes requested):

src/fcntl.rs
SteveLauC: None is used here so that we can .take() the inner type out of the lock when calling Flock.unlock()

src/fcntl.rs
SteveLauC: One thing I am concerned is that if we keep the old interface, we keep the original FlockArg type, then we keep those 2 FlockArg::Unlock* variants, which means our user can do:

let lock = Flock::lock(file, Flock::Unlock);

Seems like we should define a new type without Unlock* variants just for the Flock

src/fcntl.rs
coderBlitz: I meant that for unlocking specifically, none of the errors definitively seem to apply to unlock. Regardless, if we can't ignore the error then we can't implement `Drop`, which is the entire point of this effort. Further, flocks are "associated with an open file table entry" [^1], and since the inner object is dropped (when present), any remaining flocks will drop along with it.

If we want `Flock::unlock()` not ignore it, that's fine by me.

[^1]: [flock(2)](https://linux.die.net/man/2/flock)

src/fcntl.rs
coderBlitz: See your above comment on drop for full details, but I'll make the changes here.

Just want to note though that if someone ignored the documentation, and it causes some unexpected error, that's explicitly their problem because we marked the trait as `unsafe`. `EBADF` in particular would not be our problem imho, because they've broken our assumption that they have not cloned their fd.

src/fcntl.rs
coderBlitz: It will only ever be none while it is in the process of dropping, and only when after a successful `unlock()` call. It prevents a double-free situation; see above conversations for more details.

src/fcntl.rs
coderBlitz: We could make that an EINVAL case if we really wanted to prevent that, or we could just make a note of it in the docs while it's in the deprecated transition stage.

asomers: EINVAL sounds good to me.

src/fcntl.rs
asomers: EBADF still applies.  I think we should panic on error (unless `std::thread::panicing == true`).

src/fcntl.rs
asomers: Yeah, it wouldn't be "our fault".  But ignoring an EBADF error could lead to some _very_ difficult bugs, such as when the user opens a new file with the same file descriptor and we close that new file on drop, resulting in some sort of race.

src/fcntl.rs
asomers: I don't like Option because it adds extra runtime code.  You can do it without Option like this:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5713996199c6d48d422690bb50fda583

SteveLauC: Nice! `std::mem::forget()` is the way to go!

changelog/2170.removed.md
```diff
@@ -0,0 +1,2 @@
+Removed `fcntl::flock` in favor of new `Flock` wrapper.
+Removed unlock options from `fcntl::FlockArg`
```
asomers, line 2: This changelog entry is no longer accurate.

coderBlitz:
Looks like the extra failures were "spurious HTTP errors", then the one expected error with `armv7-unknown-linux-uclibceabihf`.

src/fcntl.rs
```diff
@@ -609,6 +608,7 @@ pub enum FlockArg {
 }
 
 #[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+#[deprecated = "`fcntl::Flock` should be used instead."]
```
asomers, line 611:
```suggestion
#[deprecated(since = "0.28.0", note = "`fcntl::Flock` should be used instead.")]
```

src/fcntl.rs
```diff
@@ -629,6 +629,125 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
 
     Errno::result(res).map(drop)
 }
+
+/// Represents valid types for flock.
+///
+/// # Safety
+/// Types implementing this must be `!Clone`.
```
asomers, line 636: To me, the `!` is too small to see clearly.  I think it's better to spell out "not".

```suggestion
/// Types implementing this must not be `Clone`.
```

src/fcntl.rs
```diff
@@ -629,6 +629,125 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+pub unsafe trait Flockable: AsRawFd {}
+
+/// Represents an owned flock, which unlocks on drop.
+///
+/// See flock(2) for details on locking semantics.
```
asomers, line 642:
```suggestion
/// See [flock(2)](https://linux.die.net/man/2/flock) for details on locking semantics.
```

src/fcntl.rs
```diff
@@ -629,6 +629,125 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+#[derive(Debug)]
+pub struct Flock<T: Flockable>(T);
+
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+impl<T: Flockable> Drop for Flock<T> {
+    fn drop(&mut self) {
+       if Errno::result(unsafe { libc::flock(self.0.as_raw_fd(), libc::LOCK_UN) }).is_err() && !std::thread::panicking() {
+           panic!("Failed to remove flock.");
```
asomers, line 651: The panic message should show the actual error.

src/fcntl.rs
```diff
@@ -629,6 +629,125 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+    }
+}
+
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
+impl<T: Flockable> Flock<T> {
+    /// Obtain a/an flock.
+    ///
+    /// # Example
+    /// ```
+    /// # use std::fs::File;
```
asomers, line 677: Either here, or on the type documentation I think it'd be helpful to have an example that demonstrates using the deref impls.  You can mark it as `no_run` if you want, so you can use simpler file names.

src/fcntl.rs
```diff
@@ -629,6 +629,125 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+    /// fn do_stuff(file: File) -> nix::Result<()> {
+    ///     let mut lock = match Flock::lock(file, FlockArg::LockExclusive) {
+    ///         Ok(l) => l,
+    ///         Err((_,e)) => return Err(e),
+    ///     };
+    ///
+    ///     // Do critical section
+    ///
+    ///     // Unlock (don't continue until unlocked)
+    ///     let file = loop {
```
asomers, line 718: Why loop?  unlock is inherently non-blocking.  It shouldn't fail intermittently.

src/fcntl.rs
```diff
@@ -629,6 +629,125 @@ pub fn flock(fd: RawFd, arg: FlockArg) -> Result<()> {
...
+    ///             Ok(f) => break f,
+    ///             Err((l,_)) => l,
+    ///         };
+    ///     };
+    ///
+    ///     // Do anything else
+    ///
+    ///     Ok(())
+    /// }
+    pub fn unlock(self, nonblock: bool) -> std::result::Result<T, (Self, Errno)> {
```
asomers, line 729: unlock should be inherently non-blocking.  You should remove the `nonblock` flag.

asomers (changes requested):

SteveLauC (approved):
Thanks!

PR #2171, opened 2023-10-14
sys::mman add netbsd's PROT_MPROTECT macro

devnexen:


test/sys/test_mman.rs
```diff
@@ -120,3 +120,36 @@ fn test_mremap_shrink() {
...
+    use nix::sys::mman::mprotect;
+    use std::num::NonZeroUsize;
+
+    const ONE_K: size_t = 1024;
+    let ten_one_k = NonZeroUsize::new(10 * ONE_K).unwrap();
+    let prot = ProtFlags::PROT_READ;
+    prot.PROT_MPROTECT(ProtFlags::PROT_READ);
+
+    unsafe {
+        let ptr = mmap::<BorrowedFd>(
```
SteveLauC, line 138: Anonymous mapping has been remvoed from the `mmap(2)` interface, please use `mmap_anonymous()` instead

SteveLauC (changes requested):

SteveLauC:
Could you please explain what this is and how it can be properly used to me? 

* [mmap](https://man.netbsd.org/mmap.2)
* [mremap](https://man.netbsd.org/mremap.2)
* [mprotect](https://man.netbsd.org/mprotect.2)

> As a NetBSD extension, the `PROT_MPROTECT` macro can be used to request additional permissions for later use with mprotect(2).  For example `PROT_MPROTECT(PROT_READ)` requests that future `PROT_READ` mappings are allowed and can be enabled using [`mprotect(2)`](https://man.netbsd.org/mprotect.2), but does not currently grant read mappings to the returned memory segment.

After reading the related manual pages and the test code provided in this PR, I am still confused about the functionality of this macro.

Per the `mprotect(2)` manual:

> EACCESS: The new protection is less restrictive than the protection originally set with [mmap(2)](https://man.netbsd.org/mmap.2).

You will get an `EACCESS` error if you reset the permission that is less restrictive than the original one given in the `mmap(2)` call on NetBSD, does this macro solve that error and thus allow additional permissions to be added?

devnexen:
> Could you please explain what this is and how it can be properly used to me?
> 
> * [mmap](https://man.netbsd.org/mmap.2)
> * [mremap](https://man.netbsd.org/mremap.2)
> * [mprotect](https://man.netbsd.org/mprotect.2)
> 
> > As a NetBSD extension, the `PROT_MPROTECT` macro can be used to request additional permissions for later use with mprotect(2).  For example `PROT_MPROTECT(PROT_READ)` requests that future `PROT_READ` mappings are allowed and can be enabled using [`mprotect(2)`](https://man.netbsd.org/mprotect.2), but does not currently grant read mappings to the returned memory segment.
> 
> After reading the related manual pages and the test code provided in this PR, I am still confused about the functionality of this macro.
> 
> Per the `mprotect(2)` manual:
> 
> > EACCESS: The new protection is less restrictive than the protection originally set with [mmap(2)](https://man.netbsd.org/mmap.2).
> 
> You will get an `EACCESS` error if you reset the permission that is less restrictive than the original one given in the `mmap(2)` call on NetBSD, does this macro solve that error and thus allow additional permissions to be added?

This macro is for mprotect so let 's say you do this for mmap with this flag PROT_READ | PROT_WRITE | PROT_MPROTECT(PROT_EXEC) the actual page won t have PROT_EXEC flag allowed until mprotect request it, if that makes sense

PR #2173, opened 2023-10-22
fix: copy_file_range fd_in arg on Linux/Android

SteveLauC:
## What does this PR do

Fix the `fd_in` argument of the `copy_file_range(2)` syscall on Linux/Andorid

## Checklist:

Well, I have no idea how the test passed before this...

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -712,7 +712,7 @@ pub fn copy_file_range<Fd1: AsFd, Fd2: AsFd>(
             let ret = unsafe {
                 libc::syscall(
                     libc::SYS_copy_file_range,
-                    fd_in,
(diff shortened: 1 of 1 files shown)


SteveLauC:
Oh I get it now, our unit is using `BorrowedFd`, which has the same memory representation as `RawFd`, so it passed without any issue, in my use case, I was using a reference to a `File`, so I got the `EBADF` error

asomers:
Good catch, but why `[skip ci]`?

SteveLauC:
> but why `[skip ci]`?

To save our cirrus CI usage, but it should be needless now, let me remove it

asomers (approved):

PR #2175, opened 2023-11-03
Add `SetSockOpt` definitions related to Linux Kernel TLS.

Arnavion:
1. kTLS is implemented as a TCP ULP, so add a `SetSockOpt` impl for it.

2. kTLS parameters are pushed to the kernel using `setsockopt(SOL_TLS)`,
   so add `SetSockOpt` impls for them.

Other test fixes:

1. Change `setsockopt` tests to bind to port 0 so that
   they bind to an unbound port and don't affect each other.

2. Fix `skip!` macro used to skip tests to generate its own block so that
   it can be used as an expr.

---

## What does this PR do

Implement additional definitions needed to set up kTLS.

## Checklist:

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -127,7 +127,7 @@ macro_rules! getsockopt_impl {
 ///    both of them.
 /// * `$name:ident`: name of type `GetSockOpt`/`SetSockOpt` will be implemented for.
 /// * `$level:expr` : socket layer, or a `protocol level`: could be *raw sockets*
-///    (`lic::SOL_SOCKET`), *ip protocol* (libc::IPPROTO_IP), *tcp protocol* (`libc::IPPROTO_TCP`),
+///    (`libc::SOL_SOCKET`), *ip protocol* (libc::IPPROTO_IP), *tcp protocol* (`libc::IPPROTO_TCP`),
 ///    and more. Please refer to your system manual for more options. Will be passed as the second
 ///    argument (`level`) to the `getsockopt`/`setsockopt` call.
 /// * `$flag:path`: a flag name to set. Some examples: `libc::SO_REUSEADDR`, `libc::TCP_NODELAY`,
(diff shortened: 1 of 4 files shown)


Arnavion:
[The Linux CI](https://github.com/nix-rust/nix/pull/2175/checks?check_run_id=18326788974) did fail with `EADDRINUSE`. I did notice when I was writing the test that all the setsockopt tests that did `bind` were using the same port, and I wondered how it was that they never stepped on each other. I guess it was just luck that nobody hit it before.

Should I change my test and all the others to use unique ports?

asomers (changes requested):
Yes, you should change the port number.  Also, would it be possible to add some docs?  It isn't obvious how to use this.  And is there any overlap with PR #2065 ?

Arnavion:
>Yes, you should change the port number.

Will do.

>Also, would it be possible to add some docs? It isn't obvious how to use this.

The documentation is https://docs.kernel.org/networking/tls.html . I'll add comments explaining which setsockopt call each of the structs maps to.

>And is there any overlap with PR https://github.com/nix-rust/nix/pull/2065 ?

No. The steps of using kTLS are:

1. Set up TCP connection.

2. Do TLS handshake however you want, say using rustls or openssl.

3. Enable the TLS ULP on the TCP socket. (`setsockopt(TclUlp, b"tls")` from this PR.)

4. Import the TLS secrets for sending and receiving to the TCP socket. (`setsockopt(TcpTlsTx, ...)` and `setsockopt(TcpTlsRx, ...)` from this PR.)

5. Read and write to the TCP socket using regular socket syscalls. Reading *should* use `recvmsg` because the kernel will send cmsg's with level == `libc::SOL_TLS` and type == `libc::TLS_GET_RECORD_TYPE` whenever there is a non-application data / post-handshake record. (The `recvmsg` changes in #2065 )

asomers:
There are a bunch of tests that don't pass under cross.  If you're sure that this test failure is cross-induced, you could skip the test in CI by adding `#[cfg_attr(qemu, ignore)]`.

Arnavion:
> There are a bunch of tests that don't pass under cross. If you're sure that this test failure is cross-induced, you could skip the test in CI by adding `#[cfg_attr(qemu, ignore)]`.

Yes, that is what I had been battling with. The last commit I pushed finally fixes them.

src/sys/socket/sockopt.rs
```diff
@@ -1119,6 +1119,158 @@ where
...
+                val.as_ref().as_ptr().cast(),
+                val.as_ref().len() as libc::socklen_t,
+            );
+            Errno::result(res).map(drop)
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+/// Value used with the [`TcpTlsTx`] and [`TcpTlsRx`] socket options.
```
SteveLauC, line 1168: Nitpicking about this doc comment, could you please put it on the top of `#[cfg(target_os = "linux")]` so that it will be consistent with other code :)

```rs
/// Value used with the [`TcpTlsTx`] and [`TcpTlsRx`] socket options.
#[cfg(target_os = "linux")]
#[derive(Copy, Clone, Debug)]
pub enum TlsCryptoInfo {
```

Arnavion: Ah yes. Done.

Cargo.toml
```diff
@@ -83,6 +83,9 @@ caps = "0.5.3"
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
 sysctl = "0.4"
 
+[patch.crates-io]
+libc = { git = "https://github.com/Arnavion/libc", branch = "ktls" }
```
Arnavion, line 87: [The libc PR](https://github.com/rust-lang/libc/pull/3422) has been merged, so I'm now waiting for it to be released.

Arnavion:
Done, and verified with my test program.

src/sys/socket/sockopt.rs
```diff
@@ -1115,6 +1115,158 @@ where
...
+/// ```c
+/// setsockopt(sock, SOL_TCP, TCP_ULP, "tls", sizeof("tls"));
+/// ```
+///
+/// ... and the `nix` equivalent is:
+///
+/// ```ignore,rust
+/// setsockopt(sock, TcpUlp::default(), b"tls\0");
+/// ```
+#[cfg(target_os = "linux")]
```
asomers, line 1131: The TCP_ULP and SOL_TCP symbols are defined on Android as well as Linux.  So here and elsewhere, you should do
```suggestion
#[cfg(linux_android)]
```
Those two symbols are also defined on Fuchsia, but stuff like `tls12_crypto_info_aes_gcm_128` isn't, so I guess you should leave Fuchsia out.

test/sys/test_sockopt.rs
```diff
@@ -613,3 +616,99 @@ fn test_tcp_fast_open_connect() {
...
+        // the `setsockopt(SOL_TCP, TCP_ULP, "tls", sizeof("tls"))` call succeeds
+        // but the subsequent `setsockopt(SOL_TLS, TLS_TX, ...)` call fails with `ENOPROTOOPT`.
+        // It's as if the first `setsockopt` call enabled some other option, not `TCP_ULP`.
+        // For example, `strace` says:
+        //
+        //     [pid   813] setsockopt(4, SOL_TCP, 0x1f /* TCP_??? */, [7564404], 4) = 0
+        //
+        // It's not clear why `setsockopt(SOL_TCP, TCP_ULP)` succeeds if the container image libc doesn't support it,
+        // but in any case we can't run the test on such an architecture, so skip it.
+        Err(nix::Error::ENOPROTOOPT) => skip!("TCP ULP are not supported."),
```
asomers, line 691: It might be simpler to delete this part, and just add `#[cfg_attr(qemu, ignore)]` to the top of the function.

asomers (changes requested):

src/sys/socket/sockopt.rs
Arnavion: The TLS ULP is useless without being able to do `TLS_TX` and `TLS_RX`, and I only added all this in libc for `linux`. Someone who uses Android should extend it to Android.

test/sys/test_sockopt.rs
Arnavion: Done.

src/sys/socket/sockopt.rs
Arnavion: A, but you're right that there's no harm in enabling other ULPs at least. Done.

src/sys/socket/sockopt.rs
```diff
@@ -1115,6 +1115,158 @@ where
...
+/// For example, to enable the TLS ULP on a socket, the C function call would be:
+///
+/// ```c
+/// setsockopt(sock, SOL_TCP, TCP_ULP, "tls", sizeof("tls"));
+/// ```
+///
+/// ... and the `nix` equivalent is:
+///
+/// ```ignore,rust
+/// setsockopt(sock, TcpUlp::default(), b"tls\0");
```
SteveLauC, line 1129: I am curious, in the C interface, does this "tls" argument always have the type C string (having a NUL at the end) or it can be some pure random bytes

Arnavion: The value is the name of the ULP. You can look at the ones you have available with `cat /proc/sys/net/ipv4/tcp_available_ulp`.

The string does not need to be NUL-terminated. This is leftover from when I was trying to make the CI work and I thought the lack of NUL terminator might be the reason armv7 was failing. I confirmed from the kernel code that it only copies the user-provided string up to the smaller of user-provided length and 15 bytes (`TCP_ULP_NAME_MAX - 1`) and adds its own NUL terminator, and it also works without the trailing NUL in my test program.

Arnavion: Removed the trailing NUL from this example and from the test.

SteveLauC: >  I confirmed from the kernel code that it only copies the user-provided string up to the smaller of user-provided length and 15 bytes (TCP_ULP_NAME_MAX - 1) and adds its own NUL terminator.

Thanks for digging into the kernel code, do you think it would be great if we could have a comment in the code describing this behavior?

Arnavion: Done.

SteveLauC: Thanks for doing it, and you don't need to squash commits as we will do squash merge.

SteveLauC (approved):
LGTM, but maybe @asomers also wants to finish his review

Arnavion:
@asomers @SteveLauC Ping.

SteveLauC:
Normally, asomers will maintain Nix on Friday/weekend

Arnavion:
@SteveLauC Ping. This has been ready to merge for two weeks now, and I'm tired of fixing all the merge conflicts from the other newer PRs that keep getting merged :)

This is also blocking more fixes I want to make to the `GetSockOpt` / `SetSockOpt` traits.

SteveLauC:
Understand, let's merge this!

PR #2181, opened 2023-11-07
Pass pointers, not slices, to libc::ioctl

maurer:
Slices are not FFI-safe.
&[u8] and &mut [u8] are *wide pointers*, which means that at the moment, we're getting lucky because they're passed via the "ScalarPair" ABI, and this means that passing a `&[u8]` results in passing two arguments, `*const u8` and `usize` for pointer and length. This passes an extra argument to ioctl, which happens to work because the extra vararg is skipped. We're getting lucky right now, and we should explicitly pass the pointer we meant to pass instead.

This is normally checked on `extern "C"` functions, but `ioctl` in particular uses varargs which prevents the compiler check from reporting.

## What does this PR do

Passes the buffer pointer to `libc::ioctl` rather than trying to pass slices across FFI.

## Checklist:

- [x] I have written necessary tests and rustdoc comments (Comment updated, no new tests)
- [x] A change log has been added if this PR modifies nix's API (N/A)


SteveLauC (approved):

SteveLauC:
Good catch! Thanks, and congratulations on your first contribution to nix!

PR #2190, opened 2023-11-14
Added `sigaction_is_default()`, `sigaction_is_ignore()`, and `sigaction_current()` functions

### LINKED ISSUE #2172: Mechanism to query existing signal handler
If NULL is passed as the second (new action) argument to `sigaction()`, then it won't make any changes, but will [return the existing action](https://man7.org/linux/man-pages/man2/sigaction.2.html).

I don't believe nix provides any way to access this behavior.  The type of `nix::sys::signal::sigaction()`s `sigaction` argument could be changed from `&SigAction` to `Option<&SigAction>`, but this would be a breaking change and also would mean doing a read-only query of the current action would only be possible with an `unsafe` function call.

I'd prefer something like a safe `sigaction_get()` (roughly akin to [this proposal](https://github.com/nix-rust/nix/issues/587) for a `sigaction_safe()`).

cpick:
## What does this PR do

Provide safe mechanisms to determine whether a signal's action is the default or ignore.
This covers the majority of real-world calls to `sigaction()` with a NULL action without needing any unsafe code.

Provide an unsafe way to query the currently installed sigaction.

The decision to add `sigaction_current()` instead of just exposing the (new) `sigaction_inner()` function was to avoid any confusion over the semantics of passing in a `None` `sigaction` argument (eg: someone thinking that it meant remove or reset the action).

This closes #2172

## Checklist:

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -911,6 +911,22 @@ pub unsafe fn sigaction_current(signal: Signal) -> Result<SigAction> {
     sigaction_inner(signal, None)
 }
 
+/// Whether the specified signal currently has its default action.
+///
+/// `signal` can be any signal except `SIGKILL` or `SIGSTOP`.
+pub fn sigaction_is_default(signal: Signal) -> Result<bool> {
+    // SAFETY: fetching the current action is safe if the handler isn't called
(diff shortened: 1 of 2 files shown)


cpick:
I've also added a (loosely) related/complementary proposal to add safe functions to ignore and restore signals' actions in #2191.

PR #2191, opened 2023-11-14
Add `signal_ignore()` and `signal_default()`

### LINKED ISSUE #587: Provide safe sigaction functions for setting SIG_IGN and SIG_DFL
#101 made `sigaction` unsafe because of soundness issues for thread-local storage. It's unfortunate that the really common use case of setting signals to default or ignore now requires `unsafe`.

cpick:
## What does this PR do

Provide safe mechanisms to ignore or reset a signal's action to the default.

Fixes #587.

## Checklist:

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -964,6 +964,22 @@ pub unsafe fn signal(signal: Signal, handler: SigHandler) -> Result<SigHandler> 
     })
 }
 
+/// Reset the specified signal to its default action.
+///
+/// `signal` can be any signal except `SIGKILL` or `SIGSTOP`.
+pub fn signal_default(signal: Signal) -> Result<()> {
+    // SAFETY: restoring a default handler is safe if the old handler isn't called
+    unsafe { self::signal(signal, SigHandler::SigDfl) }.map(drop)
+}
+
+/// Ignore the specified signal.
+///
+/// `signal` can be any signal except `SIGKILL` or `SIGSTOP`.
+pub fn signal_ignore(signal: Signal) -> Result<()> {
+    // SAFETY: ignoring a signal is safe if the old handler isn't called
+    unsafe { self::signal(signal, SigHandler::SigIgn) }.map(drop)
+}
+
 fn do_pthread_sigmask(how: SigmaskHow,
                        set: Option<&SigSet>,
                        oldset: Option<*mut libc::sigset_t>) -> Result<()> {
--- a/test/sys/test_signal.rs
+++ b/test/sys/test_signal.rs
@@ -141,3 +141,19 @@ fn test_signal() {
     // Restore default signal handler
     unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap();
 }
+
+#[test]
+fn test_safe_signal() {
+    let _m = crate::SIGNAL_MTX.lock();
+
+    signal_ignore(SIGINT).unwrap();
+    raise(Signal::SIGINT).unwrap();
+
+    signal_default(SIGINT).unwrap();
+
+    // Ensure default was restored
+    assert_eq!(
+        unsafe { signal(Signal::SIGINT, SigHandler::SigDfl) }.unwrap(),
+        SigHandler::SigDfl
+    );
+}


asomers:
I can see the value of this PR; it allows something to be done safely that was previously only possible unsafely.  It also avoids the problem that the existing `sigaction` function ignores the flags and mask field when setting SIG_DFL or SIG_IGN.  But it creates a layering violation by layering one public API atop another.  And I find the names confusing, too.  Maybe this means that the original API wasn't well thought out?   What would you think about making most of these functions into methods of `Signal`, like this:
```rust
pub struct MaybeSigAction {
    /// Cast into a `SigAction`.
    ///
    /// # Safety
    /// There is no guarantee that the old signal handler was installed correctly. If it was installed by this crate, it will be. But
    /// if it was installed by, for example, C code, then there is no guarantee its function pointer is valid. In that case, this
    /// function effectively dereferences a raw pointer of unknown provenance.
    unsafe fn into(self) -> SigAction {...}
}
pub enum Signal {SIGHUP, SIGINT, ...}
impl Signal {
    pub fn kill(...) {...}
    pub fn killpg(...) {...}
    pub fn is_default() {...}
    pub fn set_default(...) {...}
    pub fn set_handler(...) -> Result<MaybeSigAction> {...}
    pub fn get_handler() -> Result<MaybeSigAction> {...}
    ...
}
```
This API provides better namespacing than the original.  With appropriately chosen definitions for types like SigAction and Handler it eliminates the unused mask and flags fields when setting the SIG_IGN or SIG_DFL.  And it eliminates unsafety by moving it to a final casting step.  So you can use it like this:
```rust
/// Set a signal handler safely, ignoring the old value
let _ = Signal::SIGINT.set_handler(...).unwrap();
/// Set a signal handler and unsafely dereference the old handler
let old: SigAction = unsafe{ Signal::SIGINT.set_handler(...).unwrap().into() }
```
What do you think, @SteveLauC ?

cpick:
Thanks for having a look!

> But it creates a layering violation by layering one public API atop another.

To my eye this is reasonable because the upper layer is adding the "safety functionality", but I'm looking at it from a pretty narrow scope.

> And I find the names confusing, too.

My goal is to make it possible to ignore signals/reset actions safely, I'm definitely not precious about the method naming or implementation.  Very happy to adjust for clarity, etc.

> Maybe this means that the original API wasn't well thought out? What would you think about making most of these functions into methods of `Signal`, like this:
> ...

I can see how a wider re-work could end up in a better place overall.

> ```rust
> /// Set a signal handler safely, ignoring the old value
> let _ = Signal::SIGINT.set_handler(...).unwrap();
> ```

I believe this would need to remain unsafe because of all the regular restrictions that POSIX puts on signal handler behavior (reentrancy, minimal interactions with global state, async-signal-safe functions only, etc).

SteveLauC:
Sorry for the late response, I think I accidentally ignored that ping.

-----

I believe we should allow our users to do the following things in safe Rust:

1. set_default
2. set_ignored
3. is_default
4. is_ignored

For the interface, I slightly tend to the functions proposed in this PR, the `impl Signal` way IMHO kind of makes things a little bit complex.

4 functions may be a little more, we can probably do:

> The naming is also bad, we should use some better ones

```rs
enum SafeSignalHanlder {
    Default,
    Ignore,
}

fn signal_is(sig: Signal, handler: SafeSignalHandler) -> Result<bool>;
fn signal_set(sig: Signal, handler: SafeSignalHandler) -> Result<()>;
```

-----

For the mechanism to query existing signal handler, I tend to change the `sigaction` argument of `sigaction()` to be optional:

```rs
pub unsafe fn sigaction(signal: Signal, sigaction: Option<&SigAction>) -> Result<SigAction>
```
-----

> 
> > ```rust
> > /// Set a signal handler safely, ignoring the old value
> > let _ = Signal::SIGINT.set_handler(...).unwrap();
> > ```
> 
> I believe this would need to remain unsafe because of all the regular restrictions that POSIX puts on signal handler behavior (reentrancy, minimal interactions with global state, async-signal-safe functions only, etc).

I think @asomers' s proposal isolates all the unsafety into `MaybeSigAction`, and this function will take a `SigAction`, which has to be cast from a `MaybeSigAction`, when doing this, users should make sure that the signal handler set by them are safe.

PR #2187, opened 2023-11-13
Added FreeBSD's SCM_REALTIME and SCM_MONOTONIC into sys::socket::ControlMessageOwned

recatek:
## What does this PR do

Adds FreeBSD's SCM_REALTIME and SCM_MONOTONIC into sys::socket::ControlMessageOwned to read socket control messages after setting SO_TS_CLOCK to either SO_TS_REALTIME or SO_TS_MONOTONIC. Note that this does not also support SO_TS_BINTIME yet as that requires a new data structure. See [man setsockopt for FreeBSD](https://man.freebsd.org/cgi/man.cgi?setsockopt) for more info on these settings. 

Also adds a proper enum as a parameter to sockopt::TsClock for choosing which clock to timestamp with. Previously this took a raw i32 and required libc as a dependency to properly configure.

Resolves #2186.

## Checklist:

- [x] I have read `CONTRIBUTING.md`
- [x] I have written necessary tests and rustdoc comments
- [x] A change log has been added if this PR modifies nix's API

PR #2194, opened 2023-11-16
Add fanotify API wrappers

ad0:
## What does this PR do

It adds linux fanotify API support: wrappers for libc `fanotify_init` and `fanotify_mark` functions.
Note that it requires libc version 0.2.150 since it introduces necessary fanotify constants (https://github.com/rust-lang/libc/pull/3408).

## Checklist:


src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+        /// Combination of `FAN_CLOSE_WRITE` and `FAN_CLOSE_NOWRITE`.
+        FAN_CLOSE;
+        /// Combination of `FAN_MOVED_FROM` and `FAN_MOVED_TO`.
+        FAN_MOVE;
+    }
+}
+
+libc_bitflags! {
+    /// Configuration options for [`fanotify_init`](fn.fanotify_init.html).
+    pub struct InitFlags: libc::c_uint {
```
SteveLauC, line 81: What about the other `InitFlags`?

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+        /// Remove the limit of 16384 events for the event queue.
+        FAN_UNLIMITED_QUEUE;
+        /// Remove the limit of 8192 marks.
+        FAN_UNLIMITED_MARKS;
+    }
+}
+
+libc_bitflags! {
+    /// File status flags for fanotify events file descriptors.
+    pub struct OFlags: libc::c_uint {
```
SteveLauC, line 105: We have already defined thie `OFlags` type in `src/fcntl.rs`, it might be better to give this type with another name, like `EventFFlags`? And since this type is basically a subset of `OFlags`, we could:

```rs
impl TryFrom<OFlags> for EventFFlags
impl From<EventFFlags> for OFlags
```

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+        /// Mark the filesystem specified by pathname. Since Linux 4.20.
+        FAN_MARK_FILESYSTEM;
+
+        /// Combination of `FAN_MARK_IGNORE` and `FAN_MARK_IGNORED_SURV_MODIFY`.
+        FAN_MARK_IGNORE_SURV;
+    }
+}
+
+/// Compile version number of fanotify API.
+pub const FANOTIFY_METADATA_VERSION: u8 = libc::FANOTIFY_METADATA_VERSION;
```
SteveLauC, line 164: Is it necessary to re-export this constant if we have a function to check the `ver` field on a `FanotifyEvent`?

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+
+#[derive(Debug)]
+/// Abstract over `libc::fanotify_event_metadata`, which represents an event
+/// received via `Fanotify::read_events`.
+pub struct FanotifyEvent {
```
SteveLauC, line 169: It might be better to change this type to:

```rs
#[derive(Debug, Eq, Hash, PartialEq)]
#[repr(transparent)]
pub struct FanotifyEvent(libc::fanotify_event_metadata)

impl FanotifyEvent {
    pub fn version(&self) -> u8 {
        self.0.vers
    }

   pub fn mask(&self) -> MaskFlags {
        ...
    }

   pub fn fd(&self) -> Option<BorrowedFd> {
       if self.0.fd == libc::FAN_NOFD {
           None
       } else {
           // SAFETY:
           // self.0.fd will be open for the lifetime of `Self`, which is longer than the lifetime of the returned
           // BorrowedFd, so it is safe
           unsafe { BorrowedFd::borrow_raw(self.0.fd) }
       }
   }
}
```

as this adheres to our [conventions](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#libc-constants-functions-and-structs)

For that `fd` field, we can `close()` it when `drop()`ing this struct


-------

Well, our [`InotifyEvent`](https://docs.rs/nix/latest/nix/sys/inotify/struct.InotifyEvent.html) does not obey our conventions either, I don't why it does not, and it may also need to be changed

Friendly cc @asomers, `inotify` API was added in #1016

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Checks that compile fanotify API version is equal to the version of the
+    /// event.
+    pub fn has_compile_version(&self) -> bool {
+        self.version == FANOTIFY_METADATA_VERSION
+    }
+}
+
+#[derive(Debug)]
+/// Abstraction over the structure to be sent to allow or deny a given event.
+pub struct FanotifyResponse<'e> {
```
SteveLauC, line 196: And this one:

```rs
#[derive(Debug)]
#[repr(transparent)]
pub struct FanotifyResponse<'a> {
    inner: libc::fanotify_response,
    _borrowed_fd: PhantomData<BorrowedFd<'a>>,
}

impl<'a> FanotifyResponse<'a> {
    /// Create a new response.
    pub fn new(fd: BorrowedFd<'a>, response: Response) -> Self {
        let response = match response {
            Response::Allow => libc::FAN_ALLOW,
            Response::Deny => libc::FAN_DENY,
        };

        Self {
            inner: libc::fanotify_response {
                fd: fd.as_raw_fd(),
                response,
            },
            _borrowed_fd: PhantomData,
        }
    }
}
```

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+    pub fd: Option<OwnedFd>,
+    /// PID of the process that caused the event. TID in case flag
+    /// `FAN_REPORT_TID` was set at group initialization.
+    pub pid: i32,
+}
+
+impl FanotifyEvent {
+    /// Checks that compile fanotify API version is equal to the version of the
+    /// event.
+    pub fn has_compile_version(&self) -> bool {
```
SteveLauC, line 189: Nit: Do you think it might be better to rename this function to something like `check_version`, well, technically it does have a `FANOTIFY_METADATA_VERSION` if the `version` field equal to it, but "has" feels a little bit of weird

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+                None => Ok(f(std::ptr::null())),
+            }
+        }
+
+        let res = with_opt_nix_path(path, |p| unsafe {
+            libc::fanotify_mark(
+                self.fd.as_raw_fd(),
+                flags.bits(),
+                mask.bits(),
+                dirfd.unwrap_or(libc::AT_FDCWD),
```
SteveLauC, line 262: We have a `at_rawfd()` function, which is used by all the `xxx_at(dirfd: Option<RawFd>, ..)` functions, it might be better to use it here to keep consistency, though they are actually the same thing

That function is gated with:

```
#[cfg(any(
    all(feature = "fs", not(target_os = "redox")),
    all(feature = "process", any(target_os = "android", target_os = "linux"))
))]
```

you need to add the `fanotify` feature to it

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Possible errors can be those that are explicitly listed in
+    /// [fanotify(2)](https://man7.org/linux/man-pages/man7/fanotify.2.html) in
+    /// addition to the possible errors caused by `read` call.
+    /// In particular, `EAGAIN` is returned when no event is available on a
+    /// group that has been initialized with the flag `InitFlags::FAN_NONBLOCK`,
+    /// thus making this method nonblocking.
+    pub fn read_events(&self) -> Result<Vec<FanotifyEvent>> {
```
SteveLauC, line 283: Per the man page (man 7 fanotify), the buffer may not just contain the `struct fanotify_event_metadata` type, should we check and expose them here

SteveLauC (changes requested):

SteveLauC:
Hi, thanks for your interest in contributing to Nix, I have left some comments on this PR:)

ad0:
Hi, thank you for the detailled review, I'll treat your comments asap !

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        let nread = read(self.fd.as_raw_fd(), &mut buffer)?;
+
+        while (nread - offset) >= metadata_size {
+            let metadata = unsafe {
+                let mut metadata =
+                    MaybeUninit::<libc::fanotify_event_metadata>::uninit();
+                ptr::copy_nonoverlapping(
+                    buffer.as_ptr().add(offset),
+                    metadata.as_mut_ptr() as *mut u8,
```
SteveLauC, line 298: Forgot to mention this yesterday, for pointer type casts, we prefer [`.cast()`](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast-1) and `.cast_mut()` rather than `as`

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+        };
+        let resp = libc::fanotify_response {
+            fd: response.fd.as_raw_fd(),
+            response: response_value,
+        };
+        write(
+            self.fd.as_fd(),
+            unsafe {
+                std::slice::from_raw_parts(
+                    (&resp as *const _) as *const u8,
```
SteveLauC, line 346: Same as above

SteveLauC (changes requested):

src/sys/fanotify.rs
```diff
@@ -0,0 +1,365 @@
...
+        /// Subfile was deleted. Since Linux 5.1.
+        FAN_DELETE;
+        /// Self was deleted. Since Linux 5.1.
+        FAN_DELETE_SELF;
+        /// Self was moved. Since Linux 5.1.
+        FAN_MOVE_SELF;
+        /// File was opened for exec. Since Linux 5.0.
+        FAN_OPEN_EXEC;
+
+        /// Event queued overflowed.
```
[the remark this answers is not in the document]
ad0: Hah, I copied typos from kernel headers comments... :sweat_smile:

src/sys/fanotify.rs
ad0: Sure

src/sys/fanotify.rs
ad0: Done, also added feature fanotify to `fcntl::OFlag` feature guard.

src/fcntl.rs
```diff
@@ -64,7 +64,11 @@ libc_bitflags! {
...
 
-#[cfg(any(feature = "fs", feature = "term"))]
+#[cfg(any(
+    feature = "fs",
+    feature = "term",
+    all(feature = "fanotify", target_os = "linux")
+))]
 libc_bitflags!(
     /// Configuration options for opened files.
     #[cfg_attr(docsrs, doc(cfg(any(feature = "fs", feature = "term"))))]
```
SteveLauC, line 74: This also needs to be updated:
```suggestion
    #[cfg_attr(docsrs, doc(cfg(any(feature = "fs", feature = "term", all(feature = "fanotify", target_os = "linux")))))]
```

SteveLauC (changes requested):

src/sys/fanotify.rs
ad0: It makes sense to me since we also expose `pub fn version(&self)` method for the `FanotifyEvent`. For the case the check version function returns false, it may be convenient to be able to get and log both version numbers that mismatch.

src/fcntl.rs
ad0: Oops sorry I missed that one!

src/sys/fanotify.rs
ad0: Well, I omitted several flags on purpose. They are used to generate additional data that goes in additional structs (that should be read after `FanotifyEvent`s) but this mechanism is not yet implemented because the corresponding structs are not in the Rust libc yet. Still, I just added two flags, I did not realized at first that those two don't need additional structs but use the field `pid` of event struct.

src/sys/fanotify.rs
ad0: Yes, we eventually should. But for now, I'm working on another libc contribution as the needed structs are not declared in it yet. I let you decide if this first version of fanotify which contains features up to Linux 5.1 is ok to be merged and completed in another incoming contribution or if you prefer to wait for libc to be updated to get the whole fanotify package in this PR.

src/sys/fanotify.rs
ad0: Nice! I also declared `FAN_ALLOW` and `FAN_DENY` directly with `libc_bitflags`. In case you're wondering why not an enum instead, this is because another two values (`FAN_AUDIT` and `FAN_INFO`) will be added later to send additional data with the response, and they are declared in the C header as bitmask on allow/deny values.

src/sys/fanotify.rs
ad0: Yes! I was extra cautious not to mess with event file descriptors lifetime and indeed took inspiration on inotify API but this way of handling it is nicer.

src/sys/fanotify.rs
SteveLauC: >  For the case the check version function returns false, it may be convenient to be able to get and log both version numbers that mismatch.

Makes sense to me, thanks for explaining it

src/sys/fanotify.rs
SteveLauC: This PR looks good, let's merge it first:)

SteveLauC:
Let's merge this PR, mind removing the WIP/Draft status?

SteveLauC (approved):

SteveLauC:
That tier3 (armv7-unknown-linux-uclibceabihf) CI failure is probably a nightly Rust bug, see also this [comment](https://github.com/nix-rust/nix/pull/2187#issuecomment-1820253849)

PR #2196, opened 2023-11-19
Ensure all tests use unique IP ports

asomers:
## Checklist:

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -2550,7 +2550,7 @@ mod tests {
 
         #[test]
         fn ip() {
-            let s = "127.0.0.1:8080";
+            let s = "127.0.0.1:8082";
             let ip = SockaddrIn::from_str(s).unwrap().ip();
             assert_eq!("127.0.0.1", format!("{ip}"));
         }
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -21,7 +21,7 @@ pub fn test_timestamping() {
     };
     use std::io::{IoSlice, IoSliceMut};
 
-    let sock_addr = SockaddrIn::from_str("127.0.0.1:6790").unwrap();
+    let sock_addr = SockaddrIn::from_str("127.0.0.1:6797").unwrap();
 
     let ssock = socket(
         AddressFamily::Inet,
@@ -314,7 +314,7 @@ mod recvfrom {
 
     #[test]
     pub fn udp() {
-        let std_sa = SocketAddrV4::from_str("127.0.0.1:6789").unwrap();
+        let std_sa = SocketAddrV4::from_str("127.0.0.1:6795").unwrap();
         let sock_addr = SockaddrIn::from(std_sa);
         let rsock = socket(
             AddressFamily::Inet,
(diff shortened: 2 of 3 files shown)


SteveLauC:
There still seems to be tests that are using the same port 8080 in `src/sys/socket/addr.rs`:

https://github.com/nix-rust/nix/blob/dbdcfed86e18673c71b54d4c4cd7be289441843c/src/sys/socket/addr.rs#L2560-L2597

-----

BTW, If we want to maintain this state, what about defining some helper type like:

```rs
/// Helper type for ensuring all ports used in tests are unique
struct PortAllocator {
    next: AtomicU16,
}

impl PortAllocator {
    pub fn allocate_port(&self) -> u16 {
         self.next.fetch_add(1, Ordering::Relaxed)
    }
}
```

Though it may be easier to clash (with the ports used on developer's host) comparing to manually-specified ports

asomers:
In the case of the 8080 ports, you'll notice that none of those tests actually bind to the ports in question.  So there's no real conflict.  Regarding the allocator, I've wondered about that too.  But I wonder if the addition of the allocator would obscure the function of some of the tests.  Do you think so?

SteveLauC:
> In the case of the 8080 ports, you'll notice that none of those tests actually bind to the ports in question. So there's no real conflict.

Yeah, that's right, they are not in actual use. I saw the changes in `src/sys/socket/addr.rs`, which made me think we should also handle such cases, but since they are not in use, let's ignore them


>  Regarding the allocator, I've wondered about that too. But I wonder if the addition of the allocator would obscure the function of some of the tests. Do you think so?

Emmm, I am not sure about this:D, would you like show me a brief example demostrating your worries?

asomers:
> > Regarding the allocator, I've wondered about that too. But I wonder if the addition of the allocator would obscure the function of some of the tests. Do you think so?
> 
> Emmm, I am not sure about this:D, would you like show me a brief example demostrating your worries?

I'm just thinking about readability for somebody not highly familiar with this project.  Code like this:
```rust
let addr = SockaddrIn::from_str("127.0.0.1:1234");
```
is a lot more obvious than this:
```rust
let addr = crate::port_allocator::next();
```

SteveLauC:
> I'm just thinking about readability for somebody not highly familiar with this project. Code like this:

Yeah, if we do something like this, we should document it well.


> I'm just thinking about readability for somebody not highly familiar with this project. Code like this:
> 
> ```rust
> let addr = SockaddrIn::from_str("127.0.0.1:1234");
> ```
> 
> is a lot more obvious than this:
> 
> ```rust
> let addr = crate::port_allocator::next();
> ```

Indeed, the old one is clearer, but will we also do this to IP address?   It should be something like this if we only do this to ports:

```rs
let port = PORT_ALLOCATOR.next();
let addr = SockAddrV4::from_str(&format!("127.0.0.1:{port}")).unwrap();
```

asomers:
Either way would work.  What do you think is best?

SteveLauC:
I think we should add a port allocator because manually maintaining the state is too hard, and conflicts are difficult to catch as well since tests are running concurrently.

For the readability issue, it should not be that bad if we only do this to ports:

> ```rs
> let addr = SockAddrV4::from_str(&format!("127.0.0.1:{port}")).unwrap();
> ```

-----

BTW, is there any guideline on when a test should be put in `src/xxx.rs` or `test/test_xxx.rs`, if not, we should probably make this clear, e.g.,

Tests for public Nix interfaces should all be placed under `test`, leaving only tests for private stuff within its own module.

asomers:
> BTW, is there any guideline on when a test should be put in `src/xxx.rs` or `test/test_xxx.rs`, if not, we should probably make this clear, e.g.,

Not really.  Nix has so many contributors; it seems like each has placed tests according to his preference.  For me personally I have a rule that I like to use, but it doesn't really apply to Nix.

> Tests for public Nix interfaces should all be placed under `test`, leaving only tests for private stuff within its own module.

:+1:

SteveLauC:
Should we merge this PR? I think it is ok to do that port allocator thing in the future

asomers:
> Should we merge this PR? I think it is ok to do that port allocator thing in the future

Yeah, I think so.  Would you approve it please?

SteveLauC (approved):

Issue #2031, opened 2023-05-03
`sendmmsg` panics with assertion failed


kalradivyanshu:
I cant get sendmmsg to work with GSO and txtime

This function uses sendmsg and works:

```rust
pub fn send_to_gso_pacing(
    socket: &tokio::net::UdpSocket,
    buf: &[u8],
    send_to: SocketAddr,
    send_at: Instant,
    segment_size: usize,
) -> io::Result<usize> {
    let iov = [IoSlice::new(buf)];
    let segment_size = segment_size as u16;
    let dst = SockaddrStorage::from(send_to);
    let sockfd = socket.as_raw_fd();

    // GSO option.
    let cmsg_gso = ControlMessage::UdpGsoSegments(&segment_size);
    let nanos_per_sec: u64 = 1_000_000_000;

    // Pacing option.
    let mut time_spec = libc::timespec {
        tv_sec: 0,
        tv_nsec: 0,
    };

    unsafe {
        std::ptr::copy_nonoverlapping(
            &send_at as *const _ as *const libc::timespec,
            &mut time_spec,
            1,
        )
    };

    let send_time = time_spec.tv_sec as u64 * nanos_per_sec + time_spec.tv_nsec as u64;

    let cmsg_txtime = ControlMessage::TxTime(&send_time);

    match sendmsg::<SockaddrStorage>(
        sockfd,
        &iov,
        &[cmsg_gso, cmsg_txtime],
        MsgFlags::empty(),
        Some(&dst),
    ) {
        Ok(v) => Ok(v),
        Err(e) => Err(e.into()),
    }
}
```

but the same message using sendmmsg fails:

```rust
pub fn sendmmsg_to_gso_pacing(
    socket: &tokio::net::UdpSocket,
    bufs: &mut [Vec<u8>],
    send_to: SocketAddr,
    send_at: Instant,
    batch_size: usize,
    segment_size: usize,
) -> io::Result<Vec<usize>> {
    let segment_size = segment_size as u16;
    let mut iovs = Vec::with_capacity(batch_size);
    let mut addrs = Vec::with_capacity(batch_size);
    let dst = SockaddrStorage::from(send_to);
    let sockfd = socket.as_raw_fd();

    for buf in bufs {
        iovs.push([IoSlice::new(buf)]);
        addrs.push(Some(dst));
    }

    // GSO option.
    let cmsg_gso = ControlMessage::UdpGsoSegments(&segment_size);
[...]

PR #2199, opened 2023-11-23
New socket API

### LINKED ISSUE #1939: UB in SockaddrStorage::hash
`SockaddrStorage::from_raw` allows creating a variable that isn't fully initialized.  But the `SockaddrStorage::hash` implementation assumes that if the ss_family field is of an unknown type, then the entire structure must be fully initialized.  You could cause it to access uninitialized data by doing something like the following:
```rust
let sa = libc::sockaddr {
    sa_len: 8,
    sa_family: 255,  // Not a valid family
    .. unsafe{ mem::zeroed() }
};
let ss = unsafe{ SockaddrStorage::from_raw((&sa).as_ptr(), Some(sa.sa_len)).unwrap() };
let mut s = DefaultHasher::new();
ss.hash(&mut s);
```
Granted, this requires the use of `unsafe`.  But we can and should still fix it by always validating `sa_family` in `SockaddrStorage::from_raw`.

### LINKED ISSUE #1990: `Multiheader::preallocate` causes an uninitialized value to be dereferenced

### LINKED ISSUE #2030: `sendmmsg` documentation is incorrect
https://docs.rs/nix/latest/nix/sys/socket/fn.sendmmsg.html is incorrect. The doc says that `sendmmsg` takes 3 arguments:

### Arguments

* `fd`:             Socket file descriptor
* `data`:           Struct that implements `IntoIterator` with `SendMmsgData` items
 * `flags`:          Optional flags passed directly to the operating system.

But it takes 6? And as far as I can tell `SendMmsgData` doesn't exist anymore? I would love a working example of how to use `sendmmsg`.

Thanks!

### LINKED ISSUE #2054: `assume_init` on partially initizalized socket addresses
There are a few places in the code that call `assume_init()` on a `MaybeUninit` that may only be partially initialized, which is UB.

https://github.com/nix-rust/nix/blob/89b4976fddf713ca54834612e351ae35d86c430a/src/sys/socket/mod.rs#L2217-L2242

https://github.com/nix-rust/nix/blob/89b4976fddf713ca54834612e351ae35d86c430a/src/sys/socket/mod.rs#L2340-L2369

https://github.com/nix-rust/nix/blob/89b4976fddf713ca54834612e351ae35d86c430a/src/sys/socket/mod.rs#L2045-L2065

The calling code might call these functions with a `S: SockaddrLike` larger than the number of bytes the kernel will write during the syscall. For example if a `SockaddrStorage` is provided to a `getsockname()` call for an inet socket, the syscall will write a `sockaddr_in` which is much smaller than the `SockaddrStorage`, leaving some of the `SockaddrStorage` bytes uninitialized. Those uninitialized bytes will then be "assumed init" by the `assume_init()` call, which is UB.

The simplest solution would be to use `mem::MaybeUninit::<T>::zeroed()` instead since *I think* all socket address types are POD, so zero values should be allowed. You could also instead use `as_ptr()` directly on the `MaybeUninit` without using `assume_init()` first, but this will give you a `*const S` instead of a `*const sockaddr`, but there's no guarantee that a `*const S` can be cast to a `*const sockaddr`.

### LINKED ISSUE #2176: Possible Buffer-Overflow when sending control messages with `sendmmsg`
```rust
pub fn sendmmsg<'a, C, S>(
    // elided ...
    cmsgs: C,
) 
where
    C: AsRef<[ControlMessage<'a>]>,
{

    for mmsghdr in data.items.iter_mut().enumerate() {
        let p = &mut mmsghdr.msg_hdr;

        let mut pmhdr: *mut cmsghdr = unsafe { CMSG_FIRSTHDR(p) };
        for cmsg in cmsgs.as_ref() {
            assert_ne!(pmhdr, ptr::null_mut());

            // This is unsound: it relies on the caller having initialized the control
            // message buffer inside `data` with enough space for holding all
            // control messages inside `cmsgs`, i.e. having used the `cmsg_space!`
            // macro properly in conjunction with `MultiHeaders::preallocate`. Since
            // the macro, the constructor of `MultiHeaders` and this function are
            // all safe to call, unsafe code must not rely on this invariant to be
            // upheld by the caller.
            //
            // Even the above assertion won't catch this bug, as there might
            // be enough space left to send *a* control message, different from `cmsg`.
            //
            // Original comment:
            //
            // > Safe because we know that pmhdr is valid, and we initialized it with
            // > sufficient space

            unsafe { cmsg.encode_into(pmhdr) };

            pmhdr = unsafe { CMSG_NXTHDR(p, pmhdr) };
        }
    }

    // ...
}

```

The security impact of this issue is probably negligible, as

- the control message buffer inside `data` must have a non-zero length and the majority of the users of this function just want to reduce the amount of syscalls and don't need control messages,
- an attacker would have to be able to influence which control messages are being sent, which is a very odd and unlikely scenario, or the program must be buggy in first place and initialize the control message buffer too small.

### LINKED ISSUE #2177: `recvfrom`, `recvmsg` and `recvmmsg` might return unitialized addresses when using connection-mode sockets
When using connection-mode sockets like TCP, the addresses returned by `recvfrom`, `recvmsg` and `recvmmsg` are not initialized by the syscall.

This is how the receive operations create an IPV4-Socketaddr, with a comment which operation is unsound here:

```rust
impl SockaddrLike for SockaddrIn {
    unsafe fn from_raw(
        addr: *const libc::sockaddr,
        len: Option<libc::socklen_t>,
    ) -> Option<Self>
    where
        Self: Sized,
    {
        if let Some(l) = len {
            if l != mem::size_of::<libc::sockaddr_in>() as libc::socklen_t {
                return None;
            }
        }

        // Unsound if `addr` has not been initialized by the syscall, e.g. when using TCP.
        //
        // Currently, this is a numbers game: if we are lucky enough that the value at the
        // memory region of `sa_family` is not `AF_INET`, this check will notice the unitialized
        // state of `addr` and return `None`. Otherwise, this will return junk as an initialized address.
        if (*addr).sa_family as i32 != libc::AF_INET {
            return None;
        }
        Some(Self(ptr::read_unaligned(addr as *const _)))
    }
}
```

The following test will catch this issue (but it is already based on the new API I'm working on):

```rust
#[test]
fn test_connection_mode_recv_address() {
    use std::io::{IoSlice, IoSliceMut};

    use nix::sys::socket::*;

    let mut header = RecvMsgHeader::<Option<SockaddrIn>>::new();

    const MSG: &[u8] = "Hello, world!".as_bytes();
    let mut buf: [u8; 13] = [0u8; MSG.len()];

    let addr = "127.0.0.1:6069".parse::<SockaddrIn>().unwrap();

    // Prefill the address buffer with valid data.
    {
        let send_sock = socket(
            AddressFamily::INET,
            SockType::Datagram,
            SockFlag::empty(),
            Some(SockProtocol::Udp),
        )
        .unwrap();

        let recv_sock = socket(
            AddressFamily::INET,
            SockType::Datagram,
            SockFlag::empty(),
            Some(SockProtocol::Udp),
        )
        .unwrap();

        bind(recv_sock.as_raw_fd(), &addr).unwrap();

        sendmsg(
            send_sock.as_raw_fd(),
            Some(&addr),
            &[IoSlice::new(MSG)],
            CmsgEmpty::write(),
            MsgFlags::empty(),
        )
        .unwrap();

        let res = recvmsg(
            recv_sock.as_raw_fd(),
            &mut header,
            &mut [IoSliceMut::new(&mut buf)],
            CmsgEmpty::read(),
            MsgFlags::empty(),
        )
        .unwrap();

        assert!(res.address().is_some());
    }

    // Now we'll use the address same buffer with an connection-mode socket
    {
        let send_sock = socket(
            AddressFamily::INET,
            SockType::Stream,
            SockFlag::empty(),
            Some(SockProtocol::Tcp),
        )
        .unwrap();

        let recv_sock = socket(
            AddressFamily::INET,
            SockType::Stream,
            SockFlag::empty(),
            Some(SockProtocol::Tcp),
        )
        .unwrap();

        bind(recv_sock.as_raw_fd(), &addr).unwrap();
        listen(&recv_sock, 1).unwrap();

        let res = std::thread::scope(|s| {
            s.spawn(|| {
                connect(send_sock.as_raw_fd(), &addr).unwrap();

                send(send_sock.as_raw_fd(), MSG, MsgFlags::empty()).unwrap();
            });

            let fd = accept(recv_sock.as_raw_fd()).unwrap();

            recvmsg(
                fd,
                &mut header,
                &mut [IoSliceMut::new(&mut buf)],
                CmsgEmpty::read(),
                MsgFlags::empty(),
            )
            .unwrap()
        });

        // This assertion will fail in the current implementation (untested)
        assert!(res.address().is_none());
    }
}
```

Jan561:
Still a soft WIP, but ready for a first review and open for suggestions. A changelog will be provided when we agreed on the changes. 

This PR changes virtually everything related to socket.

Quick Overview:

- Turned the enum `AddressFamily` to a struct with associated constants as enums were the wrong abstraction here
- Made the naming of the socket address structs consistent (every name now ends with either `*Address` if sized or `*Addr` if unsized)
- `SockaddrStorage` got renamed to `Address`, and turned it into a struct.
- Variable sized addresses like `UnixAddress`, `LinkAddress` and `Address` now have dyn-sized variants named `UnixAddr`, `LinkAddr` and `Addr`
- Removed the `SockaddrLike` trait
- Every address implements `AsRef<Addr>`, which also replaces the `SockaddrLike` bound
- Added `RawAddress`, which can be used when the system call manages the address memory. `getifaddrs` uses this new type.
- Basically every system call signature got changed
- Improved the documentation, while still not perfect. I tried a couple things before this iteration, so maybe some documentation could now be stale.

Incomplete list of issues and PRs affected by this PR:

Fixes #2177
Fixes #2176
Fixes #2054
Fixes #2030
Fixes #1990
Fixes #1939

Maybe Fixez #2031

Supersedes #2038
Supersedes #1992
Supersedes #1475

## Checklist:

### DIFF
--- a/src/ifaddrs.rs
+++ b/src/ifaddrs.rs
@@ -2,34 +2,32 @@
 //!
 //! Uses the Linux and/or BSD specific function `getifaddrs` to query the list
 //! of interfaces and their associated addresses.
-
 use cfg_if::cfg_if;
-#[cfg(apple_targets)]
-use std::convert::TryFrom;
 use std::ffi;
 use std::iter::Iterator;
+use std::marker::PhantomData;
 use std::mem;
 use std::option::Option;
 
 use crate::net::if_::*;
-use crate::sys::socket::{SockaddrLike, SockaddrStorage};
+use crate::sys::socket::RawAddr;
 use crate::{Errno, Result};
 
 /// Describes a single address for an interface as returned by `getifaddrs`.
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct InterfaceAddress {
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct InterfaceAddress<'a> {
     /// Name of the network interface
     pub interface_name: String,
     /// Flags as from `SIOCGIFFLAGS` ioctl
     pub flags: InterfaceFlags,
     /// Network address of this interface
-    pub address: Option<SockaddrStorage>,
+    pub address: Option<RawAddr<'a>>,
     /// Netmask of this interface
-    pub netmask: Option<SockaddrStorage>,
+    pub netmask: Option<RawAddr<'a>>,
     /// Broadcast address of this interface, if applicable
-    pub broadcast: Option<SockaddrStorage>,
+    pub broadcast: Option<RawAddr<'a>>,
     /// Point-to-point destination address
-    pub destination: Option<SockaddrStorage>,
+    pub destination: Option<RawAddr<'a>>,
 }
 
 cfg_if! {
@@ -44,54 +42,12 @@ cfg_if! {
     }
 }
 
-/// Workaround a bug in XNU where netmasks will always have the wrong size in
-/// the sa_len field due to the kernel ignoring trailing zeroes in the structure
-/// when setting the field. See https://github.com/nix-rust/nix/issues/1709#issuecomment-1199304470
-///
-/// To fix this, we stack-allocate a new sockaddr_storage, zero it out, and
-/// memcpy sa_len of the netmask to that new storage. Finally, we reset the
-/// ss_len field to sizeof(sockaddr_storage). This is supposedly valid as all
-/// members of the sockaddr_storage are "ok" with being zeroed out (there are
-/// no pointers).
-#[cfg(apple_targets)]
-unsafe fn workaround_xnu_bug(info: &libc::ifaddrs) -> Option<SockaddrStorage> {
-    let src_sock = info.ifa_netmask;
-    if src_sock.is_null() {
-        return None;
-    }
-
-    let mut dst_sock = mem::MaybeUninit::<libc::sockaddr_storage>::zeroed();
-
-    let dst_sock = unsafe {
-        // memcpy only sa_len bytes, assume the rest is zero
-        std::ptr::copy_nonoverlapping(
-            src_sock as *const u8,
-            dst_sock.as_mut_ptr().cast(),
-            (*src_sock).sa_len.into(),
-        );
-
-        // Initialize ss_len to sizeof(libc::sockaddr_storage).
-        (*dst_sock.as_mut_ptr()).ss_len =
-            u8::try_from(mem::size_of::<libc::sockaddr_storage>()).unwrap();
-        dst_sock.assume_init()
-    };
-
-    let dst_sock_ptr =
-        &dst_sock as *const libc::sockaddr_storage as *const libc::sockaddr;
-
-    unsafe { SockaddrStorage::from_raw(dst_sock_ptr, None) }
-}
-
-impl InterfaceAddress {
+impl<'a> InterfaceAddress<'a> {
     /// Create an `InterfaceAddress` from the libc struct.
     fn from_libc_ifaddrs(info: &libc::ifaddrs) -> InterfaceAddress {
         let ifname = unsafe { ffi::CStr::from_ptr(info.ifa_name) };
-        let address = unsafe { SockaddrStorage::from_raw(info.ifa_addr, None) };
-        #[cfg(apple_targets)]
-        let netmask = unsafe { workaround_xnu_bug(info) };
-        #[cfg(not(apple_targets))]
-        let netmask =
-            unsafe { SockaddrStorage::from_raw(info.ifa_netmask, None) };
+        let address = unsafe { RawAddr::new(&*info.ifa_addr) };
+        let netmask = unsafe { RawAddr::new(&*info.ifa_netmask) };
         let mut addr = InterfaceAddress {
             interface_name: ifname.to_string_lossy().to_string(),
             flags: InterfaceFlags::from_bits_truncate(info.ifa_flags as i32),
@@ -103,12 +59,38 @@ impl InterfaceAddress {
 
         let ifu = get_ifu_from_sockaddr(info);
         if addr.flags.contains(InterfaceFlags::IFF_POINTOPOINT) {
-            addr.destination = unsafe { SockaddrStorage::from_raw(ifu, None) };
+            addr.destination = unsafe { RawAddr::new(&*ifu) };
         } else if addr.flags.contains(InterfaceFlags::IFF_BROADCAST) {
-            addr.broadcast = unsafe { SockaddrStorage::from_raw(ifu, None) };
+            addr.broadcast = unsafe { RawAddr::new(&*ifu) };
         }
 
         addr
+    }
+}
+
+/// Holds the results of `getifaddrs`.
+///
+/// Use the function `getifaddrs` to create this struct and [`Self::iter`]
+/// to create the iterator. Note that the actual list of interfaces can be
+/// iterated once and will be freed as soon as the Iterator goes out of scope.
+#[derive(Debug)]
+pub struct InterfaceAddresses {
+    base: *mut libc::ifaddrs,
+}
+
+impl InterfaceAddresses {
+    /// Create an iterator over the list of interfaces.
+    pub fn iter(&self) -> InterfaceAddressIterator<'_> {
+        InterfaceAddressIterator {
+            next: self.base,
+            _a: PhantomData,
+        }
+    }
+}
+
+impl Drop for InterfaceAddresses {
+    fn drop(&mut self) {
+        unsafe { libc::freeifaddrs(self.base) };
     }
 }
 
@@ -118,19 +100,13 @@ impl InterfaceAddress {
 /// actual list of interfaces can be iterated once and will be freed as
 /// soon as the Iterator goes out of scope.
 #[derive(Debug, Eq, Hash, PartialEq)]
-pub struct InterfaceAddressIterator {
-    base: *mut libc::ifaddrs,
+pub struct InterfaceAddressIterator<'a> {
     next: *mut libc::ifaddrs,
+    _a: PhantomData<&'a ()>,
 }
 
-impl Drop for InterfaceAddressIterator {
-    fn drop(&mut self) {
-        unsafe { libc::freeifaddrs(self.base) };
-    }
-}
-
-impl Iterator for InterfaceAddressIterator {
-    type Item = InterfaceAddress;
+impl<'a> Iterator for InterfaceAddressIterator<'a> {
+    type Item = InterfaceAddress<'a>;
     fn next(&mut self) -> Option<<Self as Iterator>::Item> {
         match unsafe { self.next.as_ref() } {
             Some(ifaddr) => {
@@ -154,7 +130,7 @@ impl Iterator for InterfaceAddressIterator {
 /// # Example
 /// ```
 /// let addrs = nix::ifaddrs::getifaddrs().unwrap();
-/// for ifaddr in addrs {
+/// for ifaddr in addrs.iter() {
 ///   match ifaddr.address {
 ///     Some(address) => {
 ///       println!("interface {} address {}",
@@ -167,13 +143,12 @@ impl Iterator for InterfaceAddressIterator {
 ///   }
 /// }
 /// ```
-pub fn getifaddrs() -> Result<InterfaceAddressIterator> {
+pub fn getifaddrs() -> Result<InterfaceAddresses> {
     let mut addrs = mem::MaybeUninit::<*mut libc::ifaddrs>::uninit();
     unsafe {
         Errno::result(libc::getifaddrs(addrs.as_mut_ptr())).map(|_| {
-            InterfaceAddressIterator {
+            InterfaceAddresses {
                 base: addrs.assume_init(),
-                next: addrs.assume_init(),
             }
         })
     }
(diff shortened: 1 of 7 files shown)


Jan561:
Okay, so I have added a test that performs equality checks on the netmasks returned by `getifaddrs`, and as expected, they fail on macos.

I see two ways how to proceed from here:

- We define a variable sized `*Addr` variant for all addresses, and make the conversion methods in `RawAddr` return only `*Addr` types
- We enforce that fixed size addresses have always a length equal to the underlying libc type.
  - Let the conversion methods fail if the length is too small. MacOS users can still use `RawAddr::as_ptr` to access the netmask
  - Revert the changes with `RawAddr`, and make `getifaddrs` return something like `Option<Address>`

asomers (changes requested):
This is way too big to review.  There are thousands of lines of changes here.
Some of them are purely formatting changes, which should never be mixed with
content changes.  And specifically:

* The sendmsg changes should be broken out into a separate PR from the sockaddr changes.
* Your name changes don't match with existing precedent.  For example, std::net:: always uses "Addr" instead of "Address".  And within std::net, anything that combines an IP address with a port has "Sock" in the name.  We should maintain that precedent.  This PR's `Ipv4Address` type sounds like it is equivalent to std::net::Ipv4Addr instead of std::net::SocketAddrV4.
* Why is enum the wrong abstraction for AddressFamily?
* Functions like `getsockname` now return a full sized sockaddr_storage unconditionally.  That allocates a lot of extra memory that might not be needed.  Previously, being generic, getsockname could be used with smaller sized types.
* It seems to me that the main point of the PR is to replace `SockaddrLike` with `AsRef<Addr>`. Could you please explain the motivation for that?
* The `RawAddr` type only wraps `libc::sockaddr`.  That doesn't even have enough room for a `sockaddr_in6`.  That means that `RawAddr::to_ipv6` reads beyond the end of its reference.  Instead, you should perform the cast from sockaddr to sockaddr_in6 while operating on raw pointers, before you turn anything into references.  The fact that you had to add `#[allow(unused)]` to `RawAddr::new` is a clue that it's doing something wrong.

Jan561:
The current implementation has many issues. The goal of these changes is to not only fix these issues, but to make it harder to make similar mistakes again.

> This is way too big to review. There are thousands of lines of changes here.

I agree, and I expected this response. I thought that for discussions it is easier to have everything in one PR, so you can see the whole picture. We can use this PR just for discussion and eventually create smaller, atomic PRs on the changes we agree on.

>  Your name changes don't match with existing precedent.  For example, std::net:: always uses "Addr" instead of "Address".  And within std::net, anything that combines an IP address with a port has "Sock" in the name.  We should maintain that precedent.  This PR's `Ipv4Address` type sounds like it is equivalent to std::net::Ipv4Addr instead of std::net::SocketAddrV4.

I just thought that `UnixAddress`, `Ipv4Address` and `Address` is way prettier and easier to memorize than `UnixAddr`, `SockaddrIn` and `SockadddrStorage`.

> Why is enum the wrong abstraction for AddressFamily?

`AddressFamily` isn't a closed set, e.g. custom vendor-defined address families are currently impossible to represent. Also our implementation doesn't need (and likely won't ever need) exhaustive pattern matching on all family variants.

> Functions like `getsockname` now return a full sized sockaddr_storage unconditionally.  That allocates a lot of extra memory that might not be needed.  Previously, being generic, getsockname could be used with smaller sized types.

That's a tradeoff between performance and ergonomics. I also have an [implementation] that keeps the address type as a generic, but I ultimately opted for this approach as I valued ergonomics higher. In the end, it is your choice to make, I can live with both.

I also took inspiration from [socket2] which also just returns a wrapper around `sockaddr_storage` in all their functions.

> It seems to me that the main point of the PR is to replace `SockaddrLike` with `AsRef<Addr>`. Could you please explain the motivation for that?

With the `Addr` type, `SockaddrLike` wasn't needed anymore, so I removed it.

The big advantage of `Addr` is that the length information is embedded into the pointer, so types like (the new) `UnixAddress` can be cheaply converted to this lower-level abstraction, without losing any information.

In the current implementation, we instead have the `SockaddrStorage` union that also has `UnixAddr` as a member. `UnixAddr` can have an additional length field on some platforms, which makes [this](https://github.com/nix-rust/nix/blob/master/src/sys/socket/addr.rs#L1279), [this](https://github.com/nix-rust/nix/blob/master/src/sys/socket/addr.rs#L1419) and [this](https://github.com/nix-rust/nix/blob/master/src/sys/socket/addr.rs#L1449) necessary to make it work. At least to me, it feels very forced and unnatural to have to special case `UnixAddr` every time when downcasting. Also, what if one day Linus Torvalds decides that `sockaddr_un` should have the same size as `sockaddr_storage`? The length field wouldn't fit anymore.

> The `RawAddr` type only wraps `libc::sockaddr`.  That doesn't even have enough room for a `sockaddr_in6`.  That means that `RawAddr::to_ipv6` reads beyond the end of its reference.  Instead, you should perform the cast from sockaddr to sockaddr_in6 while operating on raw pointers, before you turn anything into references.  The fact that you had to add `#[allow(unused)]` to `RawAddr::new` is a clue that it's doing something wrong.

The safety documentation of the unsafe `RawAddr::new` demands that the provided argument must be castable to the more specific pointers.

The `allow_unused` is there because the only place where it is currently used is in `ifaddrs.rs`, which isn't supported by all platforms. I tried to design it with reusability in mind, so that if more system calls emerge that manage the address memory for us and don't provide us with a length, we can just use this type.

And yes, this is the API I'm the most skeptical about.

[implementation]: https://github.com/Jan561/nix/blob/socket_api/src/sys/socket/mod.rs#L3642
[socket2]: https://docs.rs/socket2/latest/socket2/struct.SockAddr.html

Jan561:
My motivation for the unsized address types was initially because of freebsd's `sockaddr_dl`.

Take a look at [this] test. `getifaddrs` returns a link layer address with a length of 56, but the size of the struct is only 54.
Of course, the "true" length is way smaller, that's why I truncate the address in this test to only include header + nlen + alen + slen.

A user unaware of this, could run into the following issue, thinking this is totally safe, but actually accesses memory that is out of bounds of `sockaddr_dl`:

```rust
let link_addr = getifaddrs(...).xzy();
let cloned = link_addr.clone();
let cloned_ptr = cloned.as_ptr();

unsafe {
    println!("{}", *cloned_ptr.cast::<u8>().add(cloned.len() - 1));
}
```

Of course, we can also just add another workaround for this, next to the macOS workaround.

Hopefully this explains my motivations a bit better.

[this]: https://github.com/nix-rust/nix/blob/300c2ab75e1c0605d115194f0998050e21db3b75/src/sys/socket/addr.rs#L3723

Jan561:
Whether we want to use unsized types for addresses or not ultimately boils down to this "philosophical" question:

Do we want it to be a *bug*, if functions like `getifaddrs` behave unexpected? Then we need to apply workarounds for all these cases.

Or do we want it to be a *missing feature*? With my implementation, these things can be fixed by adding new functions to the types, like `Ipv4Address::new_netmask` in case of the macOS bug, or `LinkAddr::true_len` alongside with `LinkAddr::to_owned_truncate` in case of the freebsd bug.

I think both approaches are valid and have their own strengths, and even I don't know which is better, or if my provided arguments are strong enough to justify these breaking changes.

asomers:
> AddressFamily isn't a closed set, e.g. custom vendor-defined address families are currently impossible to represent. Also our implementation doesn't need (and likely won't ever need) exhaustive pattern matching on all family variants.

That makes sense.  If users have a need for a custom value, then we should make it a struct.

> That's a tradeoff between performance and ergonomics. I also have an implementation that keeps the address type as a generic, but I ultimately opted for this approach as I valued ergonomics higher. In the end, it is your choice to make, I can live with both.

A primary goal of Nix is zero-cost abstractions.  We try to make the C APIs accessible to Rust but without imposing anything unnecessary.  So let's keep these functions generic.

> what if one day Linus Torvalds decides that sockaddr_un should have the same size as sockaddr_storage? The length field wouldn't fit anymore.

That won't happen, because it would break backwards-compatibility with every binary that uses the old sockaddr_un definition.  And if Linux does introduce a sockaddr_un2, then in Nix we'll just have to exapand `SockaddrStorage`.  That's fine.

> The allow_unused is there because the only place where it is currently used is in ifaddrs.rs, which isn't supported by all platforms.

Oh, ok.  I thought it was because the `unsafe` was unused.

> A user unaware of this, could run into the following issue, thinking this is totally safe, but actually accesses memory that is out of bounds of sockaddr_dl:

Well, it's obviously not "totally safe" since it's `unsafe `;) .  I think this is a problem on the user's side.  Pointer arithmetic is always risky.

> Do we want it to be a bug, if functions like getifaddrs behave unexpected? 

I think yes.  Also, FreeBSD documents sockaddr_dl's sdl_data lenght as "minimum work area, can be larger".  So if we have any examples where it actually overflows, maybe we should create a larger container type of our own.

Jan561:
Sounds good to me, I can agree on this.

PR #2202, opened 2023-11-23
Epoll timeout

JonathanWoollett-Light:
## What does this PR do

Makes the `PollTimeout` from https://github.com/nix-rust/nix/pull/1876 generic so it can be used by `Epoll`.

## Checklist:


JonathanWoollett-Light:
Blocked on https://github.com/nix-rust/nix/pull/1876

asomers:
Can you please rebase to resolve the conflict?

JonathanWoollett-Light:
> Can you please rebase to resolve the conflict?

Rebased.

src/lib.rs
```diff
@@ -196,6 +196,11 @@ feature! {
 #[allow(missing_docs)]
 pub mod unistd;
 
+#[cfg(any(feature = "poll", feature = "event"))]
+mod timeout;
+#[cfg(any(feature = "poll", feature = "event"))]
+pub use timeout::*;
```
asomers, line 202: `nix::Timeout` is too generic.  There are plenty of other functions with timeouts that don't use the same data type.  For example, `aio_suspend` and `recvmmsg`.  Instead, can you nest it under the `poll` and `epoll` namespaces?  It's ok if it's the same type, as long as it's not exported from the crate's root.

asomers (changes requested):

src/lib.rs
SteveLauC: Keeping the `mod timeout` private, re-exporting them in `src/poll.rs` and `src/sys/epoll.rs` would do the trick.

BTW, it might be better to rename `timeout` to `poll_timeout` given it is not that generic

JonathanWoollett-Light: I've implemented this.

asomers (changes requested):
Looks good.  Could you please add a CHANGELOG?

JonathanWoollett-Light:
> Looks good. Could you please add a CHANGELOG?

I've added a CHANGELOG.

asomers (approved):

PR #2203, opened 2023-11-23
Support for socket options to get/set TOS/TTL info

larseggert:
Add support for getting/setting IPv4 type-of-service (TOS) and time-to-live (TTL) information, and IPv6 traffic class (TC) and hop count information.

Depends on https://github.com/rust-lang/libc/pull/3450, which was recently merged.

(This is my first PR to nix, and I'm also reasonably new to Rust. Please let me know if I missed anything I should have changed; am happy to rework this.)


SteveLauC:
Hi! Thanks for your interest in contributing to Nix! :)

A libc with that patch hasn't been released yet, so we have to wait.

And a CHANGELOG entry is needed, please see [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add one.

larseggert:
@SteveLauC Thanks! I will convert this to a draft PR until libc has been release. Will also update CHANGELOG.

asomers (commented):
Interesting feature.  Would it be possible to add a test?

SteveLauC:
Hi @larseggert, we now use the libc dependency from git, so you don't need to patch it yourself, rebasing your branch should make things work

larseggert:
@asomers I added a test. Note that there is a `FIXME` in there - `cmsg_space!` returns too little space when called with the option types. I can't figure out why that is.

@SteveLauC I added a CHANGELOG entry.

src/sys/socket/mod.rs
```diff
@@ -983,6 +1015,55 @@ impl ControlMessageOwned {
                 let content_type = unsafe { ptr::read_unaligned(p as *const u8) };
                 ControlMessageOwned::TlsGetRecordType(content_type.into())
             },
+            #[cfg(any(
+                apple_targets,
+                target_os = "freebsd",
+            ))]
+            #[cfg(feature = "net")]
+            (libc::IPPROTO_IP, libc::IP_RECVTOS) => {
+                let tos = unsafe { ptr::read_unaligned(p as *const libc::c_int) };
```
SteveLauC, line 1024: I realized that all these pointer type castings are using the `as` keyword, we should replace them with `.cast()` in a separate PR

test/sys/test_socket.rs
```diff
@@ -2828,3 +2828,87 @@ fn test_icmp_protocol() {
...
+            .expect("sendmsg failed");
+    }
+
+    {
+        let mut buf = [0u8; 8];
+        let mut iovec = [IoSliceMut::new(&mut buf)];
+
+        // FIXME: When passing in IpTos and IpTtl, cmsg_space returns too little space?
+        // let mut space = cmsg_space!(IpTos, IpTtl);
+        let mut space = cmsg_space!(libc::c_int, libc::c_int);
```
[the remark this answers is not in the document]
larseggert: @asomers I'm trying to look into this, but I am new to Rust. What I have figured out is that `mem::size_of::< nix::sys::socket::sockopt::IpTos>()` is `0`, and that causes the issue, because `cmsg_space!` allocates too little space. What I don't understand is _why_ it is `0`, given that I have (tried to) make it a `libc::c_int`...

test/sys/test_socket.rs
```diff
@@ -2828,3 +2828,87 @@ fn test_icmp_protocol() {
...
+            &mut iovec,
+            Some(&mut space),
+            MsgFlags::empty(),
+        )
+        .expect("recvmsg failed");
+        assert!(!msg
+            .flags
+            .intersects(MsgFlags::MSG_TRUNC | MsgFlags::MSG_CTRUNC));
+
+        let mut found = 0;
```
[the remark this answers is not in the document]
larseggert: Will do.

src/sys/socket/mod.rs
larseggert: Should I do this in this PR for the additions I made, or leave as is?

test/sys/test_socket.rs
```diff
@@ -2828,3 +2828,87 @@ fn test_icmp_protocol() {
...
+
+    let lo_ifaddr = loopback_address(AddressFamily::Inet);
+    let (_, lo) = match lo_ifaddr {
+        Some(ifaddr) => (
+            ifaddr.interface_name,
+            ifaddr.address.expect("Expect IPv4 address on interface"),
+        ),
+        None => return,
+    };
+    let receive = socket(
```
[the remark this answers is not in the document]
larseggert: I went with how the other `AF_INET` tests that go over loopback do things. I tried using `socketpair` with `AF_INET`, but I don't think that's supported?

test/sys/test_socket.rs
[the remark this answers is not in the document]
larseggert: Also `test_recv_ipv4pktinfo` also allocates via `let mut space = cmsg_space!(libc::in_pktinfo)` and not via `let mut space = cmsg_space!(nix::sys::socket::sockopt::Ipv4PacketInfo)`, so I guess that's the way to do it?

src/sys/socket/mod.rs
SteveLauC: Leaving them as-is should be ok, we will clear them in another PR

src/sys/socket/mod.rs
```diff
@@ -803,6 +803,38 @@ pub enum ControlMessageOwned {
     #[cfg(any(target_os = "linux"))]
     TlsGetRecordType(TlsGetRecordType),
 
+    #[cfg(any(
+        apple_targets,
+        linux_android,
+        target_os = "freebsd",
+    ))]
+    #[cfg(feature = "net")]
+    IpTos(libc::c_int),
```
asomers, line 812: Does this need to be `u8` on linux_android?

larseggert: I don't think so.

larseggert: Actually, hang on. The tests pass but my actual app is not.

src/sys/socket/mod.rs
```diff
@@ -803,6 +803,38 @@ pub enum ControlMessageOwned {
...
+    #[cfg(feature = "net")]
+    IpTtl(libc::c_int),
+
+    #[cfg(any(
+        apple_targets,
+        linux_android,
+        target_os = "freebsd",
+    ))]
+    #[cfg(feature = "net")]
+    Ipv6HopLimit(libc::c_int),
```
SteveLauC, line 836: This variant is gated with
```rs
    #[cfg(any(linux_android, freebsdlike, apple_targets, target_os = "haiku"))]
```

 in `ControlMessage`, seems that we should also gate it in this way here?

test/sys/test_socket.rs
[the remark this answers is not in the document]
asomers: Here's the answer: you were trying to calculate the cmsg_space of something unrelated.  Instead of `nix::sys::socket::sockopt::IpTos`, you really wanted `nix::sys::socket::ControlMessageOwned::IpTos`.  The former really does have a size of 0.  And you can't pass the latter to `cmsg_space!`, because `mem::size_of` doesn't work on enum variants.  Does that explanation change anything about the way you would write your test?

SteveLauC:
Hi, sorry for the late response, I think I will take another look at this PR tomorrow:)

src/sys/socket/sockopt.rs
```diff
@@ -1015,6 +1039,50 @@ sockopt_impl!(
...
+#[cfg(feature = "net")]
+sockopt_impl!(
+    /// Retrieve the current time-to-live field for every
+    /// IPv4 packet received on this socket.
+    IpTtl,
+    Both,
+    libc::IPPROTO_IP,
+    libc::IP_TTL,
+    libc::c_int
+);
```
SteveLauC, line 1063: There is an `Ipv4Ttl` above this, and they seem to be the same thing?

src/sys/socket/sockopt.rs
```diff
@@ -1015,6 +1039,50 @@ sockopt_impl!(
...
+#[cfg(feature = "net")]
+sockopt_impl!(
+    /// Retrieve the current time-to-live field for every
+    /// IPv4 packet received on this socket.
+    IpRecvTtl,
+    Both,
+    libc::IPPROTO_IP,
+    libc::IP_RECVTTL,
+    bool
+);
```
SteveLauC, line 1052: From the [man](https://man.freebsd.org/cgi/man.cgi?query=ip&apropos=0&sektion=4&manpath=FreeBSD+14.0-RELEASE&arch=default&format=html) page, this option is also available on FreeBSD.

src/sys/socket/sockopt.rs
```diff
@@ -1015,6 +1039,50 @@ sockopt_impl!(
...
+#[cfg(feature = "net")]
+sockopt_impl!(
+    /// Retrieve the current hop limit field for every
+    /// IPv6 packet received on this socket.
+    Ipv6RecvHopLimit,
+    Both,
+    libc::IPPROTO_IPV6,
+    libc::IPV6_RECVHOPLIMIT,
+    bool
+);
```
SteveLauC, line 1074: Even though this option is not present in the corresponding `ip6` man page, there is [a bug report](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=197256), which seems to indicate that FreeBSD has this option supported

src/sys/socket/mod.rs
```diff
@@ -982,6 +1020,53 @@ impl ControlMessageOwned {
...
+            (libc::IPPROTO_IP, libc::IP_RECVTTL) => {
+                let ttl = unsafe { ptr::read_unaligned(p as *const libc::c_int) };
+                ControlMessageOwned::IpTtl(ttl)
+            },
+            #[cfg(linux_android)]
+            #[cfg(feature = "net")]
+            (libc::IPPROTO_IP, libc::IP_TTL) => {
+                let ttl = unsafe { ptr::read_unaligned(p as *const libc::c_int) };
+                ControlMessageOwned::IpTtl(ttl)
+            },
```
SteveLauC, line 1059: Do we miss a FreeBSD impl here?

test/sys/test_sockopt.rs
SteveLauC, line 1: We have a test for `Ipv4Ttl` (`test_ttl_opts()`), which should be updated.

SteveLauC (changes requested):
Hi, I have left some comments:)

And, since `IpDontFrag` is also supported in this PR, it will be good if we can have a test for this option.

-----

> And, since `IpDontFrag` is also supported in this PR

This change should be recorded in the `changelog`, something like:

1. Add `IpDontFrag` for Linux/Android/FreeBSD
2. Add `Ipv6DontFrag` for FreeBSD

src/sys/socket/sockopt.rs
larseggert: Yes, that was there already, but the `Ipv4Ttl` name is unfortunate, because it's not obvious that this is the socket option that should be used for per-packet CMSG use. What do you prefer I do?

test/sys/test_sockopt.rs
larseggert: Updated in which way? Those tests are for setting the TTL with a a sockopt for all packets, not via CMSG per packet.

src/sys/socket/sockopt.rs
larseggert: `libc` doesn't seem to export it: https://github.com/nix-rust/nix/pull/2203/checks?check_run_id=20299209049

src/sys/socket/sockopt.rs
larseggert: `libc` doesn't seem to export it: https://github.com/nix-rust/nix/pull/2203/checks?check_run_id=20299209049

src/sys/socket/sockopt.rs
SteveLauC: Ahh, they are the same thing under the hood, but for different usages, then this is a really unfortunate naming, do you mind give them some documents on their typical use cases and how they should be properly used, some examples would be great!

test/sys/test_sockopt.rs
SteveLauC: Then let's leave it unchanged:)

larseggert: I don't think so: https://github.com/nix-rust/nix/actions/runs/7463986033/job/20309921408?pr=2203#step:4:122

SteveLauC: The CI failed due to the reason that `Ipv6Ttl` is not enabled on apple targets, I added it in #2287, once it is merged, we can enable that test for apple targets.

src/sys/socket/sockopt.rs
SteveLauC: I [tried to add these 2 constants to the libc crate](https://github.com/rust-lang/libc/pull/3543), though I have no idea about the CI failure...

SteveLauC: You can also adjust our libc git dependency and enable this option for FreeBSD as that libc PR is merged

PR #2204, opened 2023-11-23
fix(vendor): declare build.rs as included file in Cargo.toml

ad0:
## What does this PR do

It adds the file `build.rs` to the package included files in `Cargo.toml`. Without it, when vendoring a project that depends on `nix`, `build.rs` is not added in vendor cache, then macos targets are broken since the `cfg_aliases` configuration added in https://github.com/nix-rust/nix/pull/2169 is missing.

## Checklist:


asomers (approved):
Thanks for the fix!  And thanks for finding it before it made it into a release.

ad0:
No problem, I'm glad it can be fixed before the next release!

PR #2207, opened 2023-11-24
sys::sendfile adding solaris' sendfilev wrapper proposal.

devnexen:


src/sys/sendfile.rs
```diff
@@ -120,6 +120,49 @@ cfg_if! {
                 )
             }
         }
+    } else if #[cfg(any(target_os = "solaris", target_os = "illumos"))] {
```
SteveLauC, line 123: Please rebase and use `#[cfg(solarish)]`

src/sys/sendfile.rs
```diff
@@ -287,5 +330,31 @@ cfg_if! {
             };
             (Errno::result(return_code).and(Ok(())), len)
         }
+    } else if #[cfg(any(target_os = "solaris", target_os = "illumos"))] {
```
SteveLauC, line 333: Same here: `#[cfg(solarish)]`

src/sys/sendfile.rs
devnexen: yes I just saw your PR, will do.

src/sys/sendfile.rs
```diff
@@ -287,5 +330,31 @@ cfg_if! {
...
+        ///  from an offset `off` and for a given amount of data `len`.
+        ///
+        /// The latter allows to send data from the process' address space, from an offset `off`
+        /// and for a given amount of data `len`.
+        ///
+        /// For more information, see
+        /// [the sendfilev{3) man page.](https://docs.oracle.com/cd/E86824_01/html/E54768/sendfilev-3ext.html)
+        pub fn sendfilev<F: AsFd>(
+            out_sock: F,
+            vec: Vec<SendfileVec>
```
asomers, line 350: If you accept my Newtype suggest above, then you can eliminate the `collect` below.  That would also allow you to change this argument from `Vec` to slice.

test/test_sendfile.rs
```diff
@@ -204,3 +205,57 @@ fn test_sendfile_darwin() {
...
+        .write_all(trailer_strings.concat().as_bytes())
+        .unwrap();
+    let (mut rd, wr) = UnixStream::pair().unwrap();
+    let vec: Vec<SendfileVec> = vec![
+        SendfileVec::new(
+            &header_data,
+            0,
+            header_strings.iter().map(|s| s.len()).sum(),
+        ),
+        SendfileVec::new(&body_data, body_offset as off_t, body.len() - 1),
```
asomers, line 237:
```suggestion
        SendfileVec::new(&body_data, body_offset as off_t, body.len() - body_offset),
```

test/test_sendfile.rs
```diff
@@ -204,3 +205,57 @@ fn test_sendfile_darwin() {
...
+    assert!(res.is_ok());
+    wr.shutdown(Shutdown::Both).unwrap();
+
+    // Prepare the expected result
+    let expected_string = header_strings.concat()
+        + &body[body_offset..]
+        + &trailer_strings.concat();
+
+    // Verify the message that was sent
+    //assert_eq!(bytes_written as usize, expected_string.as_bytes().len());
```
asomers, line 255: Why is this commented out?

asomers (changes requested):

test/test_sendfile.rs
devnexen: I forgot to .. especially it passes.

src/sys/sendfile.rs
```diff
@@ -120,6 +120,38 @@ cfg_if! {
...
+            pub fn newself(
+                off: off_t,
+                len: usize
+            ) -> SendfileVec {
+                SendfileVec(libc::sendfilevec_t{sfv_fd: SFV_FD_SELF, sfv_flag: 0, sfv_off: off, sfv_len: len})
+            }
+
+            /// initialises SendfileVec to send data from `fd`.
+            pub fn new<F: AsFd>(
+                fd: F,
```
asomers, line 142: This still doesn't capture anything about fd's lifetime.  With this API, it's possible to use a RawFd after the file has been closed.  That's why I suggested using BorrowedFd.  You might be able to do it, without additional runtime cost, like this:
```rust
pub struct SendfileVec<'a> {
    raw: libc::sendfilevec_t,
    phantom: PhantomData<BorrowedFd<'a>>
}
pub fn new<F: AsFd, 'a>(fd: &'a F, off: off_t, len: usize) -> Self<'a> {
    let bfd = fd.as_fd();
    Self{raw: libc::sendfilevec_t{sfv_fd: bfd.as_raw_fd(), ...}, phantom: PhantomData}}
}
```

src/sys/sendfile.rs
```diff
@@ -287,5 +319,30 @@ cfg_if! {
...
+        ///
+        /// For more information, see
+        /// [the sendfilev(3) man page.](https://illumos.org/man/3EXT/sendfilev)
+        pub fn sendfilev<F: AsFd>(
+            out_sock: F,
+            vec: &[SendfileVec]
+        ) -> (Result<()>, usize) {
+            let mut len = 0usize;
+            let return_code = unsafe {
+                libc::sendfilev(out_sock.as_fd().as_raw_fd(), vec.iter().map(|&v| v.into()).collect::<Vec<_>>().as_ptr(), vec.len() as i32, &mut len)
```
asomers, line 343: You're still using `collect` here.  The intention behind using a Newtype for SendfileVec was to eliminate this `collect`.

asomers (changes requested):

src/sys/sendfile.rs
```diff
@@ -125,6 +121,44 @@ cfg_if! {
...
+        use std::os::unix::io::BorrowedFd;
+
+        #[derive(Debug, Copy, Clone)]
+        /// Mapping of the raw C sendfilevec_t struct
+        pub struct SendfileVec<'a> {
+            raw: libc::sendfilevec_t,
+            phantom: PhantomData<BorrowedFd<'a>>
+        }
+
+        const SFV_FD_SELF: i32 = -2;
```
asomers, line 134: Constants like this should be defined in the libc crate, not in Nix.

asomers (changes requested):
Nice!  It looks a lot more efficient now.

src/sys/sendfile.rs
devnexen: Yes I know I just did a PR for it tough :)

devnexen:
> libc#3452

Do we need to tough ? I can just use directly the value if that s a problem.

src/sys/sendfile.rs
```diff
@@ -125,6 +120,45 @@ cfg_if! {
...
+
+        impl<'a> SendfileVec<'a> {
```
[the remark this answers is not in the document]
devnexen: > We should probably use BorrowedFd<'a> here so that it is consistent with FdSet and PollFd

BTW, I think we should also do this to Epoll

You re seing an old version.

SteveLauC:
> Do we need to tough ?

Could you please elaborate on this? I don't think I fully understand this sentence

devnexen:
> > Do we need to tough ?
> 
> Could you please elaborate on this? I don't think I fully understand this sentence

I meant blocking because of libc.

src/sys/sendfile.rs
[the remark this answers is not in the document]
SteveLauC: Emmm, why did you revert the definition of `SendfileVec`, for that `BorrowedFd`, I was just suggesting changing the function parameter type to `BorrowedFd<'fd>`

devnexen:
> > > > Do we need to tough ?
> > > 
> > > 
> > > Could you please elaborate on this? I don't think I fully understand this sentence
> > 
> > 
> > I meant blocking because of libc.
> 
> Nix does not define these constants ourselves, we add them to libc, and wait for a release of libc with the patch included, then we merge the Nix PR

I no longer define this constant but just the value directly for the time being. but if you insist on waiting the next libc crate release, fine by me.

src/sys/sendfile.rs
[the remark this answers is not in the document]
SteveLauC: Just like what we did to `FdSet`: https://github.com/nix-rust/nix/blob/39ad47bc80af68fb9ad67d37ab699684bea0e2a9/src/sys/select.rs#L18-L47

SteveLauC:
> I no longer define this constant but just the value directly for the time being. but if you insist on waiting the next libc crate release, fine by me.

Yes, we should wait for the libc release rather than using a magic number here

SteveLauC:
Hi @devnexen, We now use the lib dependency from git, so you use the libc constant in this PR

devnexen:
> Hi @devnexen, We now use the libc dependency from git, so you use the libc constant in this PR

hmmm ok sure.

PR #2210, opened 2023-11-25
Converts `AddressFamily` to struct with associated consts

Jan561:
Split off from #2199

Script I used to generate the variants: https://gist.github.com/Jan561/067d64641c93017a9bf6b346ac078ec1

## Checklist:

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -45,233 +45,434 @@ pub(crate) const fn ipv6addr_to_libc(addr: &net::Ipv6Addr) -> libc::in6_addr {
     }
 }
 
-/// These constants specify the protocol family to be used
-/// in [`socket`](fn.socket.html) and [`socketpair`](fn.socketpair.html)
-///
-/// # References
-///
-/// [address_families(7)](https://man7.org/linux/man-pages/man7/address_families.7.html)
-// Should this be u8?
-#[repr(i32)]
-#[non_exhaustive]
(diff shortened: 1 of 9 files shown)


Jan561:
I'll provide a changelog when rebasing on top of #2211

src/sys/socket/addr.rs
```diff
@@ -45,233 +45,434 @@ pub(crate) const fn ipv6addr_to_libc(addr: &net::Ipv6Addr) -> libc::in6_addr {
...
-    #[cfg(not(any(linux_android, target_os = "redox")))]
-    Route = libc::PF_ROUTE,
-    /// Low level packet interface (see [`packet(7)`](https://man7.org/linux/man-pages/man7/packet.7.html))
+/// A possible error when converting a c_int to [`AddressFamily`].
+#[derive(Debug, Clone, Copy)]
+pub struct InvalidAddressFamilyError;
+
+/// Address families, corresponding to `AF_*` constants in libc.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct AddressFamily(libc::c_int);
```
asomers, line 54: All Newtypes should use repr(transparent).  You might even be able to cast directrly from AddressFamily to other int types just by using `as`, instead of `.family()`.
```suggestion
#[repr(transparent)]
pub struct AddressFamily(libc::c_int);
```

asomers (changes requested):
This is pretty intrusive change.  Could you please explain the justification in more detail?

PR #2211, opened 2023-11-25
cargo fmt --all -- "**/*.rs"

Jan561:
## Checklist:


SteveLauC (approved):
Thanks! 

It is indeed confusing to me why the `cargo fmt` in the CI would give different results compared to a local toolchain.

PR #2212, opened 2023-11-26
test: put test_prctl and test_clearenv under test.rs

SteveLauC:
## What does this PR do

Move `test_prctl` and `test_clearenv` to `test.rs`

I actually have no idea why they are put under their own sections, but it seems to be ok to let them live with other tests.

No actual code changes in `test/sys/test_prctl.rs`, I just removed 

```rs
#[cfg(target_os = "linux")]
#[cfg(feature = "process")]
mod test_prctl {
}
```

and shifted the code

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,14 +96,6 @@ name = "test-aio-drop"
 path = "test/sys/test_aio_drop.rs"
 
 [[test]]
-name = "test-clearenv"
-path = "test/test_clearenv.rs"
-
-[[test]]
 name = "test-mount"
 path = "test/test_mount.rs"
 harness = false
-
-[[test]]
-name = "test-prctl"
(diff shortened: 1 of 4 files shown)


asomers:
The prctl stuff is in its own test program because it changes process-wide settings that might affect other tests.
https://github.com/nix-rust/nix/pull/1550#discussion_r917456288 .  clearenv also affects the process-wide environment.

SteveLauC:
> The prctl stuff is in its own test program because it changes process-wide settings that might affect other tests. [#1550 (comment)](https://github.com/nix-rust/nix/pull/1550#discussion_r917456288) . clearenv also affects the process-wide environment.

Makes sense, they should be put in their own processes, close this PR

Issue #2188, opened 2023-11-13
Tracking Issue: Replace big `cfg!` chunks with shorter aliases


SteveLauC:
## Background

With the crate [`cfg_aliases`](https://crates.io/crates/cfg_aliases), we can replace common, chunky `cfg!`s with shorter aliases, here are the aliases defined by us:

* apple_targets: `any(ios, macos, watchos, tvos)`
* freebsdlike: `any(freebsd, dragonfly)`
* bsd: `any(freebsd, dragonfly, netbsd, openbsd, apple_targets)`
* linux_android: `any(linux, android)`
* solarish: `any(illumos, solaris)`

#2205 has done most of the replacements, but there are still `cfg!`s that haven't been updated, we want to clear them all.

## TODO

Scan our source code, and replace the long `cfg!` chunks with shorter aliases:

* [x] `src/dir.rs` `src/env.rs` (#2214)
* [x] `src/errno.rs` (#2214)
* [ ] `src/fcntl.rs` (#2215)
* [ ] `src/features.rs` `src/ifaddrs.rs` (#2215)
* [ ] `src/kmod.rs` `src/lib.rs` (#2215)
* [ ] `src/macro.rs` (#2215)
* [ ] `src/mqueue.rs` (#2215)
* [ ] `src/poll.rs` (#2215)
* [ ] `src/pty.rs` (#2215)
* [ ] `src/sched.rs` (#2215)
* [ ] `src/time.rs` (#2215)
* [x] `src/unistd.rs` (#2216)
* [x] `src/mount/bsd.rs` (#2217)
* [x] `src/mount/linux.rs` `src/mount/mod.rs` (#2217)
* [x] `src/net/if_.rs` (#2217)
* [x] `src/sys/aio.rs` (#2218)
* [x] `src/sys/epoll.rs` (#2218)
* [x] `src/sys/event.rs` (#2218)
* [x] `src/sys/mman.rs` (#2218)
* [x] `src/sys/mod.rs` (#2218)
* [x] `src/sys/resource.rs` (#2218)
* [x] `src/sys/sendfile.rs` (#2218)
* [x] `src/sys/signal.rs` (#2218)
* [x] `src/sys/stat.rs` (#2219)
* [x] `src/sys/statfs.rs` (#2219)
* [x] `src/sys/termios.rs` (#2220)
* [ ] `src/sys/time.rs` (#2229)
* [x] `src/sys/uio.rs` `src/sys/wait.rs` (#2221)
[...]

PR #2214, opened 2023-11-26
refactor: cfg for dir/env/errno

SteveLauC:
## What does this PR do

Apply the cfg aliases for:

* dir.rs
* env.rs
* errno.rs

ref: #2188 

## Checklist:

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -225,15 +225,13 @@ impl Entry {
     pub fn ino(&self) -> u64 {
         cfg_if! {
             if #[cfg(any(target_os = "aix",
-                         target_os = "android",
                          target_os = "emscripten",
                          target_os = "fuchsia",
                          target_os = "haiku",
-                         target_os = "illumos",
+                         solarish,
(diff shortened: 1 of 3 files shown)


src/dir.rs
```diff
@@ -225,15 +225,13 @@ impl Entry {
...
                          target_os = "fuchsia",
                          target_os = "haiku",
-                         target_os = "illumos",
+                         solarish,
+                         linux_android,
                          apple_targets,
-                         target_os = "l4re",
-                         target_os = "linux",
-                         target_os = "solaris"))] {
+                         target_os = "l4re"))] {
```
asomers, line 234: Honestly we could probably eliminate l4re.  I'm not sure why it's there in the first place.  This is the only reference to it in Nix, and it doesn't come remotely close to compiling.

asomers (approved):

src/dir.rs
SteveLauC: Will do it

PR #2232, opened 2023-11-27
refactor: make Epoll take BorrowedFd

SteveLauC:
## What does this PR do

Make `Epoll`-associated functions take `BorrowedFd` so that it is consistent with `FdSet` and `PollFd`

## Checklist:

### DIFF
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -3,7 +3,8 @@ pub use crate::poll_timeout::PollTimeout as EpollTimeout;
 use crate::Result;
 use libc::{self, c_int};
 use std::mem;
-use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, OwnedFd, RawFd};
+use std::os::fd::{AsFd, BorrowedFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, OwnedFd, RawFd};
 
 libc_bitflags!(
     pub struct EpollFlags: c_int {
@@ -85,7 +86,7 @@ impl EpollEvent {
 ///
 /// // Create eventfd & Add event
 /// let eventfd = eventfd(0, EfdFlags::empty())?;
-/// epoll.add(&eventfd, EpollEvent::new(EpollFlags::EPOLLIN,DATA))?;
+/// epoll.add(eventfd.as_fd(), EpollEvent::new(EpollFlags::EPOLLIN,DATA))?;
 ///
 /// // Arm eventfd & Time wait
 /// write(&eventfd, &1u64.to_ne_bytes())?;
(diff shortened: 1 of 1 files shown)


src/sys/epoll.rs
```diff
@@ -102,37 +103,50 @@ impl EpollEvent {
 /// # }
 /// ```
 #[derive(Debug)]
-pub struct Epoll(pub OwnedFd);
-impl Epoll {
+#[repr(transparent)]
+pub struct Epoll<'fd> {
+    epoll_fd: OwnedFd,
+    _fds: std::marker::PhantomData<*mut &'fd ()>,
```
SteveLauC, line 109: We can not use `_fds: std::marker::PhantomData<BorrowedFd<'fd>>` here or the following code would compile without any issue:

```rs
use std::{ fs::File, os::fd::AsFd };
use nix::sys::epoll::{Epoll, EpollCreateFlags, EpollEvent};

fn main() {
    let file = File::open("Cargo.toml").unwrap();
    let epoll = Epoll::new(EpollCreateFlags::empty()).unwrap();
    epoll.add(file.as_fd(), EpollEvent::empty()).unwrap();

    drop(file);
    drop(epoll);
}
```

I haven't figured out the reason, but using an invariant type here would work

test/sys/test_epoll.rs
```diff
@@ -1,26 +1,23 @@
...
-pub fn test_epoll_ctl() {
-    let efd = epoll_create1(EpollCreateFlags::empty()).unwrap();
-    let mut event =
-        EpollEvent::new(EpollFlags::EPOLLIN | EpollFlags::EPOLLERR, 1);
-    epoll_ctl(efd, EpollOp::EpollCtlAdd, 1, &mut event).unwrap();
-    epoll_ctl(efd, EpollOp::EpollCtlDel, 1, None).unwrap();
+pub fn test_epoll_add_delete() {
+    let epoll = Epoll::new(EpollCreateFlags::empty()).unwrap();
+    let event = EpollEvent::new(EpollFlags::EPOLLIN | EpollFlags::EPOLLERR, 1);
+    let fd_1 = unsafe { BorrowedFd::borrow_raw(1) };
```
asomers, line 19: Just to get rid of the unsafe, I suggest:
```suggestion
    let stdin = std::io::stdin();
    let fd_1 = stdin.as_fd();
```

asomers (changes requested):

test/sys/test_epoll.rs
SteveLauC: will do

src/sys/epoll.rs
SteveLauC: Friendly ping @asomers, do you have any idea about this?

PR #2236, opened 2023-11-30
More cfg alias targets + disabled related clippy lint

Jan561:
Adds the remaining supported platforms as cfg alias.

This PR also disables this [lint].

## Checklist:

[lint]: https://rust-lang.github.io/rust-clippy/master/index.html#/mismatched_target_os


SteveLauC (approved):
Thanks!

SteveLauC:
> Why is there a need for those aliases?

Yeah, they are currently not in use, should I revert this PR?

I merged this PR because it is trivial and maybe there will be a use case for it in the future, when we don't want those target_os = "xxx" stuff and fully switch to aliases

asomers:
Yes, I think we should revert it , just so we can reenable that lint.  I've seen that lint fire correctly before.

Jan561:
1. Either we use aliases for *all* supported targets, or we don't use them at all. Something in between is the worst option. I don't want to double check every time if there is a alias configured for the target.

2. #2210 already uses these aliases, and having to configure these in a PR when needing them feels wrong.

3. This lint is necessary if we want to use `#[cfg(linux)]` instead of `#[cfg(target_os = "linux")]`

Jan561:
> > #2210 already uses these aliases, and having to configure these in a PR when needing them feels wrong.
> 
> This is my fault, I am sorry about it

All good, just need to tweak my codegen script a bit.

PR #2241, opened 2023-12-01
refactor: enable most macOS funcs for apple_targets

### LINKED ISSUE #2238: Enable most macos functions on all of apple_targets
A quick search through libc shows that there's only one function we use that's defined on macos but isn't defined on tvos, watchos, and ios as well: `clock_settime`.  Even that, I suspect, is a mistake, given that `clock_gettime` is defined for all of them.
We should replace `#[cfg(target_os = macos)]` with `#[cfg(apple_targets)]` everywhere.
Honestly, I don't know why Apple needs so many different Rust targets.

SteveLauC:
## What does this PR do

1. Enable most macOS functions for apple_targets
2. Enable `ClockId::setime()` and `time::clock_settime()` on macOS

Closes #2238

## Checklist:

### DIFF
--- a/changelog/2241.changed.md
+++ b/changelog/2241.changed.md
@@ -1,0 +1,1 @@
+`ClockId::set_time()` has been disabled on iOS/tvOS/watchOS given that it is only available on macOS
--- a/changelog/2241.changed.md
+++ b/changelog/2241.changed.md
@@ -1,1 +1,0 @@
-`ClockId::set_time()` has been disabled on iOS/tvOS/watchOS given that it is only available on macOS
--- a/src/time.rs
+++ b/src/time.rs
@@ -40,7 +40,11 @@ impl ClockId {
     }
 
     /// Sets time to `timespec` on the clock id
-    #[cfg(not(any(apple_targets, target_os = "redox", target_os = "hermit",)))]
+    #[cfg(not(any(
+        target_os = "macos",
+        target_os = "redox",
+        target_os = "hermit"
+    )))]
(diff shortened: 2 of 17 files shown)


src/time.rs
```diff
@@ -40,7 +40,11 @@ impl ClockId {
     }
 
     /// Sets time to `timespec` on the clock id
-    #[cfg(not(any(apple_targets, target_os = "redox", target_os = "hermit",)))]
+    #[cfg(not(any(
```
SteveLauC, line 43: Enable it only on macOS

changelog/2241.changed.md
```diff
@@ -0,0 +1 @@
+`ClockId::set_time()` has been disabled on iOS/tvOS/watchOS given that it is only available on macOS
```
asomers, line 1: It was already disabled on iOS/tvOS/watchOS.  Shouldn't this read "ClockId::set_time is now enabled on macos" ?

src/time.rs
```diff
@@ -40,7 +40,11 @@ impl ClockId {
...
+        target_os = "macos",
```
asomers, line 44: Did you not notice the `not` ?
```suggestion
    #[cfg(not(any(
        target_os = "ios",
        target_os = "tvos",
        target_os = "watchos",
```

asomers (changes requested):

src/time.rs
SteveLauC: Oops, I am sorry I missed that, fixed!

asomers (approved):

PR #2242, opened 2023-12-01
sys::socket fixes for haiku

devnexen:


src/sys/socket/addr.rs
```diff
@@ -447,7 +447,7 @@ impl UnixAddr {
         cfg_if! {
             if #[cfg(any(linux_android,
                      target_os = "fuchsia",
-                     target_os = "illumos",
+                     solarish,
```
asomers, line 450: Why this change?  It doesn't match the commit message.

src/sys/socket/addr.rs
```diff
@@ -366,7 +366,7 @@ impl UnixAddr {
             .try_into()
             .unwrap();
 
-            #[cfg(bsd)]
+            #[cfg(any(bsd, target_os = "haiku"))]
```
asomers, line 369: I think the cfg expression here should mirror the one at line 292.  Either change this one to:
```suggestion
            #[cfg(not(any(
                linux_android,
                target_os = "fuchsia",
                target_os = "illumos",
                target_os = "redox",
           )))]
```
or change the one at line 292 to:
```rust
#[cfg(not(any(bsd, target_os = "haiku")))]
```

src/sys/socket/addr.rs
```diff
@@ -2327,7 +2327,10 @@ mod tests {
 
         #[test]
         fn from_sockaddr_un_named() {
+            #[cfg(not(target_os = "haiku"))]
```
asomers, line 2330: This socket never actually gets constructed.  The test just exercises sockaddr type conversions.  So why special case it for Haiku?

asomers (commented):
Looks better.  But why the illumos -> Solarish change?

devnexen:
Because it s better using new cfg and solaris was forgotten here.

asomers:
> Because it s better using new cfg and solaris was forgotten here.

But that's a user-visible change, and you didn't put it in the changelog entry or even in the commit message.

devnexen:
I updated the commit message but did not do the changelog change indeed..

asomers (approved):
It looks like the Solaris build is broken in some other ways, too.  But no need to fix them all in this PR.

PR #2244, opened 2023-12-02
Added IpBindAddressNoPort sockopt

xonatius:
From [man page](https://man7.org/linux/man-pages/man7/ip.7.html):
```
IP_BIND_ADDRESS_NO_PORT (since Linux 4.2)
       Inform the kernel to not reserve an ephemeral port when
       using bind(2) with a port number of 0.  The port will
       later be automatically chosen at connect(2) time, in a way
       that allows sharing a source port as long as the 4-tuple
       is unique.
```

## Checklist:

### DIFF
--- a/changelog/2244.added.md
+++ b/changelog/2244.added.md
@@ -1,0 +1,1 @@
+Added `IpBindAddressNoPort` sockopt to support `IP_BIND_ADDRESS_NO_PORT` available on linux.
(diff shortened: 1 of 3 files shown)


asomers (changes requested):
You should enable this for Android, too.

xonatius:
> You should enable this for Android, too.

ah, didn't realize that `linux` doesn't include android. Fixed!

asomers (approved):

PR #2248, opened 2023-12-03
fix, mostly, solaris build.

devnexen:


changelog/2248.fixed.md
```diff
@@ -0,0 +1,2 @@
+Fixed solaris build globally, also disabling a subset of `InterfaceFlags` values
```
asomers, line 1: You're mixing past and present tense in the same sentence here.  And what does the second part mean?  Without reading the code, I can't understand why you disabled a subset of `InterfaceFlags`.

src/net/if_.rs
```diff
@@ -213,26 +213,29 @@ libc_bitflags!(
         /// Interface is offline
         #[cfg(solarish)]
         IFF_OFFLINE;
-        #[cfg(target_os = "solaris")]
+        /*
+        // FIXME: The following are from 64 bits long long type.
```
asomers, line 217: What's wrong here?  Do these not work correctly at runtime?  You should open a new issue for this and explain what the problem is.

asomers (changes requested):

src/net/if_.rs
devnexen: what do you have in mind ?

SteveLauC: > what do you have in mind ?

Cast them with `as c_int` if they are in the range `int`

devnexen: The thing is, for most of them it would overflow [from this one at least](https://github.com/rust-lang/libc/blob/497ac428bc010b5db9682ecf94cd567b31d53e5c/src/unix/solarish/mod.rs#L1891).

SteveLauC: > Can we cast them correctly?

Then we can not do this

changelog/2248.fixed.md
```diff
@@ -0,0 +1,2 @@
+Fixed solaris build globally, also removed a subset of `InterfaceFlags` values
+which are of i64 type.
```
asomers, line 2: "which are of i64 type" still doesn't make sense to a typical reader.  Why were they removed?  Did they not work at runtime, or did they generate build errors?
Also, the ", also" creates a run-on sentence.  You should use a proper conjunction or split it into two sentences.

asomers: That changelog edit still doesn't answer the question.  Did the interfaceflags fail at runtime, or generate build errors?  Also, why did you elect to remove some of them, rather than just change the field width to 64 bits on Solarish?

asomers (approved):
That changelog edit

PR #2249, opened 2023-12-03
Fixes recvm(m)sg blindly assuming correct initialization of received address

Jan561:
`recvm(m)sg` now calling `SockaddrLike::from_raw` instead of `assume_init` on the received address

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1718,7 +1718,7 @@ where
 
         // as long as we are not reading past the index writen by recvmmsg - address
         // will be initialized
-        let address = unsafe { self.rmm.addresses[self.current_index].assume_init() };
+        let address = self.rmm.addresses[self.current_index];
 
         self.current_index += 1;
         Some(unsafe {
@@ -1726,7 +1726,7 @@ where
                 mmsghdr.msg_hdr,
                 mmsghdr.msg_len as isize,
                 self.rmm.msg_controllen,
-                address,
+                &address,
             )
         })
     }
@@ -1778,7 +1778,7 @@ unsafe fn read_mhdr<'a, 'i, S>(
     mhdr: msghdr,
     r: isize,
     msg_controllen: usize,
-    mut address: S,
+    address: &mem::MaybeUninit<S>,
 ) -> RecvMsg<'a, 'i, S>
     where S: SockaddrLike
 {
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -1901,15 +1901,16 @@ unsafe fn read_mhdr<'a, 'i, S>(
...
     };
 
-    // Ignore errors if this socket address has statically-known length
-    //
-    // This is to ensure that unix socket addresses have their length set appropriately.
-    let _ = unsafe { address.set_length(mhdr.msg_namelen as usize) };
+    let addr_len = mhdr.msg_namelen;
+
+    let address = unsafe {
+        S::from_raw(address.as_ptr().cast(), Some(addr_len))
```
asomers, line 1907: Are you aware that this creates a data copy?  The original code was carefully designed to avoid such a thing.

src/sys/socket/mod.rs
```diff
@@ -2215,7 +2216,7 @@ pub fn recvfrom<T: SockaddrLike>(
             &mut len as *mut socklen_t,
         ))? as usize;
 
-        Ok((ret, T::from_raw(addr.assume_init().as_ptr(), Some(len))))
+        Ok((ret, T::from_raw(addr.as_ptr().cast(), Some(len))))
```
Jan561, line 2219: This change isn't technically necessary, both versions are equivalent.

But I think my version better expresses what is happening here. Before, it was

`MaybeUninit<S>` -> `S` -> `*const sockaddr` -> `Option<S>`.

Converting `S` to `Option<S>` makes no sense. Now it is

`MaybeUninit<S>` -> `*const S` -> `*const sockaddr` -> `Option<S>`.

src/sys/socket/mod.rs
asomers: Why does it matter that some other function also performs a data copy?  Are you pointing that out to suggest that we improve it?  I'm requesting that you not add an additional data copy to `recvmsg`.

asomers (changes requested):

src/sys/socket/mod.rs
Jan561: Ah okay yeah, that makes sense, I think I misunderstood you before. I'll change the signature of `read_mhdr` back to expecting a pointer, and perform the conversion there.

PR #2250, opened 2023-12-04
refactor: some clippy warnings

SteveLauC:
## What does this PR do

Fix some clippy warnings

1. Some deprecated `type::max_valus()` functions, replace them with `type::MAX`
2. Two unnecessary parentheses

## Checklist:

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1520,7 +1520,7 @@ pub fn getgroups() -> Result<Vec<Gid>> {
     // equal to the value of {NGROUPS_MAX} + 1.
     let ngroups_max = match sysconf(SysconfVar::NGROUPS_MAX) {
         Ok(Some(n)) => (n + 1) as usize,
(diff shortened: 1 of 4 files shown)


src/unistd.rs
```diff
@@ -1520,7 +1520,7 @@ pub fn getgroups() -> Result<Vec<Gid>> {
     // equal to the value of {NGROUPS_MAX} + 1.
     let ngroups_max = match sysconf(SysconfVar::NGROUPS_MAX) {
         Ok(Some(n)) => (n + 1) as usize,
-        Ok(None) | Err(_) => <usize>::max_value(),
```
asomers, line 1523: In this case, the reason the code used to use max_value instead of MAX is that the latter only stabilized in Rust 1.43.0.  max_value is older.

asomers (approved):

src/unistd.rs
SteveLauC: Yes, that's right

PR #2251, opened 2023-12-04
feat: reboot(2) for OpenBSD/NetBSD

### LINKED ISSUE #2174: reboot api on openbsd
issue in libc, but cross post incase its just here that is the issue: https://github.com/rust-lang/libc/issues/3403
would like to use reboot api open openbsd

SteveLauC:
## What does this PR do

Implements `reboot(2)` for NetBSD/OpenBSD, closes #2174.

Manual:

* NetBSD: https://man.netbsd.org/reboot.2
* OpenBSD: https://man.openbsd.org/reboot.2

The `RebootMode` type, different from the one defined on Linux, is a bitflag rather than an enum, so I defined a new type. On NetBSD, there is a `RebootMode`  `RB_RDONLY`:

> RB_RDONLY  0x0080    
>
> Initially mount the root file system read-only. This is currently the default, and this option has been **deprecated**.

which is deprecated, so I didn't add it.

## Checklist:

### DIFF
--- a/src/sys/reboot.rs
+++ b/src/sys/reboot.rs
@@ -1,48 +1,141 @@
-//! Reboot/shutdown or enable/disable Ctrl-Alt-Delete.
+//! Reboot/shutdown
+//!
+//! On Linux, This can also be used to enable/disable Ctrl-Alt-Delete.
 
 use crate::errno::Errno;
 use crate::Result;
+use cfg_if::cfg_if;
 use std::convert::Infallible;
-use std::mem::drop;
 
-libc_enum! {
-    /// How exactly should the system be rebooted.
-    ///
-    /// See [`set_cad_enabled()`](fn.set_cad_enabled.html) for
-    /// enabling/disabling Ctrl-Alt-Delete.
-    #[repr(i32)]
-    #[non_exhaustive]
-    pub enum RebootMode {
-        /// Halt the system.
-        RB_HALT_SYSTEM,
-        /// Execute a kernel that has been loaded earlier with
-        /// [`kexec_load(2)`](https://man7.org/linux/man-pages/man2/kexec_load.2.html).
-        RB_KEXEC,
-        /// Stop the system and switch off power, if possible.
-        RB_POWER_OFF,
-        /// Restart the system.
-        RB_AUTOBOOT,
-        // we do not support Restart2.
-        /// Suspend the system using software suspend.
-        RB_SW_SUSPEND,
+cfg_if! {
+    if #[cfg(target_os = "linux")] {
+        use std::mem::drop;
+
+        libc_enum! {
+            /// How exactly should the system be rebooted.
+            ///
+            /// See [`set_cad_enabled()`](fn.set_cad_enabled.html) for
+            /// enabling/disabling Ctrl-Alt-Delete.
+            #[repr(i32)]
+            #[non_exhaustive]
+            pub enum RebootMode {
+                /// Halt the system.
+                RB_HALT_SYSTEM,
+                /// Execute a kernel that has been loaded earlier with
+                /// [`kexec_load(2)`](https://man7.org/linux/man-pages/man2/kexec_load.2.html).
+                RB_KEXEC,
+                /// Stop the system and switch off power, if possible.
+                RB_POWER_OFF,
+                /// Restart the system.
+                RB_AUTOBOOT,
+                // we do not support Restart2.
+                /// Suspend the system using software suspend.
+                RB_SW_SUSPEND,
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


src/sys/reboot.rs
```diff
@@ -1,48 +1,163 @@
...
+                RB_TIMEBAD;
+            }
+        }
+
+        /// Reboot system or halt processor
+        ///
+        /// For more information, see [`reboot(2)`](https://man.netbsd.org/reboot.2).
+        #[cfg(target_os = "netbsd")]
+        pub fn reboot<S: AsRef<OsStr>>(how: RebootMode, bootstr: S) -> Result<Infallible> {
+            let bootstr = bootstr.as_ref().as_bytes().as_ptr().cast_mut().cast();
```
SteveLauC, line 148: If the underlying shutdown procedure will modify `bootstr`, then we will get a UB here. I am not sure if `reboot(2)` would change it or not, the manual does not mention this:

> `bootstr` is a string passed to the firmware on the machine, if possible, if this option is set.  Currently this is only implemented on the sparc and the sun3 ports.

But practically, it has little point in writing to that piece of memory given that everything will be gone after shutdown.

src/sys/reboot.rs
```diff
@@ -1,48 +1,163 @@
...
+                /// Power off the system if the system hardware supports the
+                /// function, otherwise it has no effect.
+                ///
+                /// Should be used in conjunction with `RB_HALT`.
+                RB_POWERDOWN;
+                /// An option allowing the specification of an init program (see
+                /// `init(8)`) other than `/sbin/init` to be run when the system
+                /// reboots.
+                ///
+                /// This switch is not currently available.
```
asomers, line 89: What does it mean that it isn't currently available?  Does that mean that NetBSD defines it but does not implement it?  If so, I think that Nix shouldn't expose it.

src/sys/reboot.rs
```diff
@@ -1,48 +1,163 @@
...
+                /// When no options are given (i.e., `RB_AUTOBOOT` is used), the
+                /// system is rebooted from file /bsd in the root file system of
+                /// unit 0 of a disk chosen in a processor specific way. An automatic
+                /// consistency check of the disks is normally performed (see `fsck(8)`).
+                RB_SINGLE;
+                /// The `bootstr` argument is a string passed to the firmware
+                /// on the machine, if possible, if this option is set.
+                ///
+                /// Currently this is only implemented on the sparc and the sun3 ports.
+                #[cfg(target_os = "netbsd")]
```
asomers, line 129: Let's define it for sparc64 only.  It looks like NetBSD doesn't have a Rust package for sun3, so we don't have to worry about that. https://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/lang/rust/index.html
```suggestion
                #[cfg(all(target_os = "netbsd", target_arch = "sparc64"))]
```

src/sys/reboot.rs
asomers: The man page seems to suggest that bootstr is entirely ignored unless `RB_STRING` is specified, and that can only happen on sparc64.  So I suggest that we simply remove this argument and the `RB_STRING` flag until such a time as there is a least one known user of Nix on sparc64-unknown-netbsd.

src/sys/reboot.rs
```diff
@@ -1,48 +1,163 @@
...
+                /// Restart the system.
+                RB_AUTOBOOT,
+                // we do not support Restart2.
+                /// Suspend the system using software suspend.
+                RB_SW_SUSPEND,
+            }
+        }
+
+        /// Reboots or shuts down the system.
+        #[cfg(target_os = "linux")]
```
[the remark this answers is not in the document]
SteveLauC: I am sorry about this, removed

src/sys/reboot.rs
SteveLauC: From this [man page](https://man.netbsd.org/NetBSD-8.0/boothowto.9), seems that it is obsolete and replaced with `RB_ASKNAME`, removed

> RB_INITNAME     0x00000010          This flag is obsolete.  It was previously used to cause the kernel to prompt for the name of the [init(8)](https://man.netbsd.org/NetBSD-8.0/init.8) program, but that function is now controlled by the RB_ASKNAME flag.

src/sys/reboot.rs
SteveLauC: This flag and that `bootstr` argument are not removed

src/sys/reboot.rs
SteveLauC: argument removed

asomers (approved):

PR #2255, opened 2023-12-09
chore: more platofrms to CI

SteveLauC:
## What does this PR do

Pick up the work of #986 as the author is no longer active on GitHub

Add the following targets to CI, #986 want to make them tier 3 in Nix, should we still do this?

* aarch64-unknown-linux-musl
 
  `x86_64-unknown-linux-musl` is tier 1 in Nix, both `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` are tier 2 in the Rust.

   I added the check for this target in Cirrus Linux native builds, which is not the place a tier 3 should go though.

* ~~x86_64-unknown-linux-gnux32~~
  
  tier 2 in Rust, this **has already been added** to Nix, with tier 3 support, and is tested under qemu.
  
* sparc64-unknown-linux-gnu

  tier 2 in Rust, maybe we should also test it with qemu but make it a tier 3 in Nix.
 

## Checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -68,7 +68,7 @@ task:
 # Tasks for Linux aarch64 native builds 
 task:
   matrix:
-    - name: Linux aarch64
+    - name: Linux aarch64 glibc
       arm_container:
         image: rust:1.69.0
         cpu: 1
@@ -76,6 +76,14 @@ task:
         - FreeBSD 14 amd64 & i686
       env:
         TARGET: aarch64-unknown-linux-gnu
+    - name: Linux aarch64 musl
+      arm_container:
+        image: rust:1.69.0
+        cpu: 1
+      depends_on:
+        - FreeBSD 14 amd64 & i686
+      env:
+        TARGET: aarch64-unknown-linux-musl
   setup_script:
     - rustup target add $TARGET
     - rustup component add clippy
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -68,14 +68,22 @@ task:
 # Tasks for Linux aarch64 native builds 
 task:
   matrix:
-    - name: Linux aarch64
+    - name: Linux aarch64 glibc
       arm_container:
         image: rust:1.69.0
         cpu: 1
       depends_on:
         - FreeBSD 14 amd64 & i686
```
SteveLauC, line 76: BTW, is it still necessary to make Linux native builds depend on the FreeBSD test? This dramatically slows down our CI

asomers (commented):
We could easily add dozens of targets.  But it's only worthwhile if we think we'll get additional test coverage out of them.  Do you think that aarch64 musl and sparc64-unknown-linux-gnu will do that?

SteveLauC:
> We could easily add dozens of targets. But it's only worthwhile if we think we'll get additional test coverage out of them. Do you think that aarch64 musl and sparc64-unknown-linux-gnu will do that?

For `aarch64-unknown-linux-musl`, we do have some tests that are disabled for x86_64 musl (in `test_aio.rs`), so aarch64 musl will possibly help, but I agree, we are not desperate for that, it is reasonable to add it until we really need it, I will close this PR and #986.

.cirrus.yml
SteveLauC: > And while we don't have a full month worth of data yet

Right, I prefer to do this until we know the appropriate usage of a month, just in case

PR #2258, opened 2023-12-10
feat: expose the inner fd of Kqueue

### LINKED ISSUE #2183: FreeBSD  nix::sys::event::Kqueue access to inner type
An important functionality is missing in 0.27 for Kqueue.

1) **"field `0` of `Kqueue` is private"** 
There is no way it is possible to close the fd of the kqueue (OwnedFd should do it but in some cases it a raw fd is needed). as_raw_fd() is needed.
2) An access to the inner type (a raw fd) is needed otherwise it breaks functionality of programs which depends on this functionality. Some programs uses fd number to identify the instance when there are multiple sources.

`pub struct Kqueue(OwnedFd);
pub struct Epoll(pub OwnedFd);`
It would be great to have the same for KQueue.

SteveLauC:
## What does this PR do

1. impl AsFd for Kqueue
2. impl From\<Kqueue\> for OwnedFd

The author of #2183 wants to use the inner fd to identify a Kqueue instance, this PR implements it, so closes #2183

Should we do `impl AsRawFd` for it as well, I don't have any strong view over this, but if we have made decision on this, we should do it to all the I/O-safe types

## Checklist:

### DIFF
--- a/changelog/2258.added.md
+++ b/changelog/2258.added.md
@@ -1,0 +1,4 @@
+Expose the inner fd of `Kqueue` through:
+
+* impl AsFd for Kqueue
+* impl From\<Kqueue\> for OwnedFd
--- a/src/sys/event.rs
+++ b/src/sys/event.rs
@@ -10,6 +10,7 @@ use libc::{c_int, c_long, intptr_t, time_t, timespec, uintptr_t};
 use libc::{c_long, intptr_t, size_t, time_t, timespec, uintptr_t};
 use std::convert::TryInto;
 use std::mem;
+use std::os::fd::{AsFd, BorrowedFd};
 use std::os::unix::io::{AsRawFd, FromRawFd, OwnedFd};
 use std::ptr;
 
@@ -28,6 +29,18 @@ pub struct KEvent {
 #[repr(transparent)]
 #[derive(Debug)]
 pub struct Kqueue(OwnedFd);
+
+impl AsFd for Kqueue {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.0.as_fd()
+    }
+}
+
+impl From<Kqueue> for OwnedFd {
+    fn from(value: Kqueue) -> Self {
+        value.0
+    }
+}
 
 impl Kqueue {
     /// Create a new kernel event queue.


asomers:
I think this PR is probably ok.  But I still don't understand @eesekaj's use case.  It sounds like he wants to close the file descriptor without dropping the Kqueue, which is very unsafe.

eesekaj:
> I think this PR is probably ok. But I still don't understand @eesekaj's use case. It sounds like he wants to close the file descriptor without dropping the Kqueue, which is very unsafe.

Calling	close()	on a file  descriptor	will remove any	kevents	that reference the descriptor.

There is no way to drop Kqueue in kernel form userland. And I don't need the instance Kqueue which is created by nix crate, because we have our own abstraction above raw FD and FD is used to identify the instance. If I extract FD from Kqueue, the Kqueu should be consumed and dropped automatically. But in your previous implementation, it was made private and no way to even read the FD number.

asomers:
> > I think this PR is probably ok. But I still don't understand @eesekaj's use case. It sounds like he wants to close the file descriptor without dropping the Kqueue, which is very unsafe.
> 
> Calling close() on a file descriptor will remove any kevents that reference the descriptor.
> 
> There is no way to drop Kqueue in kernel form userland.

Are you aware that dropping the `Kqueue` will automatically close its file descriptor?

> And I don't need the instance Kqueue which is created by nix crate, because we have our own abstraction above raw FD and FD is used to identify the instance. If I extract FD from Kqueue, in this case the Kqueu should be consumed and dropped automatically. But in previous implementation, it was made private and no way to even read the FD number.

If you have your own abstraction anyway and don't use our `Kqueue::kevent` then yes I agree with what you said in the issue.  You should just use libc directly because our abstraction isn't providing any benefit to you.

SteveLauC:
Should we merge this?

asomers:
I think your PR is reasonable.  The only troubling part is @eesekaj 's explanation for why he wanted it.  He never did fully explain it.  I think he was trying to close the kqueue, and didn't understand that closing it happens automatically on drop.

SteveLauC:
> The only troubling part is @eesekaj 's explanation for why he wanted it. He never did fully explain it. I think he was trying to close the kqueue, and didn't understand that closing it happens automatically on drop.

Maybe that usage is very specific to their project, and [they switched to the libc crate](https://github.com/nix-rust/nix/issues/2183#issuecomment-1858211155)

asomers (approved):

PR #2260, opened 2023-12-10
sys::mman adding MMAP_TRYFIXED for netbsdlike platforms.

devnexen:


src/sys/mman.rs
```diff
@@ -154,6 +154,10 @@ libc_bitflags! {
         /// Pages will be discarded in the core dumps.
         #[cfg(target_os = "openbsd")]
         MAP_CONCEAL;
+        /// Attempt to place the mapping at exactly the address specified in `addr`.
+        /// it's a default behavior on openbsd.
```
SteveLauC, line 158: Nit: Might be better if we change it to:

```suggestion
        /// It's the default behavior on OpenBSD.
```

SteveLauC (changes requested):

SteveLauC (approved):

PR #2261, opened 2023-12-11
refactor: impl EventFd::read/write with eventfd_read/write

SteveLauC:
## What does this PR do

Implement `EventFd::read/write` with `libc::eventfd_read()/eventfd_write()` functions instead of implementing them ourselves so that we can survive even when libc has its behavior changed

## Checklist:


asomers (approved):
LGTM.  A quick online search showed that these functions have been available in Ubuntu since at least 16.04, and in Android since at least 2008.

SteveLauC:
> A quick online search showed that these functions have been available in Ubuntu since at least 16.04, and in Android since at least 2008.

Well, then I am not sure if we should merge this since I don't want to be in trouble like the one of #2146, let's me investigate the Glibc version where they were added.

Issue #1361, opened 2020-12-15
aio_suspend returns EINVAL on OSX in CI


asomers:
After working fine for years, sys::test_aio::test_aio_suspend has begun to reliably fail in CI, when run on OSX.  The trigger seems to be the recent MSRV upgrade to 1.40.0.  Unfortunately, none of the current Nix maintainers have access to a Mac, so we aren't able to debug this one.  Help wanted.
```
---- sys::test_aio::test_aio_suspend stdout ----
thread 'sys::test_aio::test_aio_suspend' panicked at 'aio_suspend returned Sys(EINVAL)', test/sys/test_aio.rs:170:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

SteveLauC:
Seems that this issue has been solved, close as completed in https://github.com/nix-rust/nix/pull/2262, but if it still occurs in the future, we can reopen it at any time.

Issue #1099, opened 2019-07-13
test_aio_suspend crashes on Linux with rust nightly


asomers:
With the latest nightly compiler, Travis's Linux builds (but not my local linux builds) fail with this error:
```
../sysdeps/pthread/aio_suspend.c:102: do_aio_misc_wait: Assertion `status == 0 || status == -11' failed
```
That error is coming from within glibc itself.

asomers:
On Cirrus-CI this test is now failing with a different error: `The futex facility returned an unexpected error code.`.  But it's still a glibc bug.

SteveLauC:
Seems that this issue has been solved, close as completed in #2262, but if it still occurs in the future, we can reopen it at any time.

PR #2262, opened 2023-12-11
test: enable test_aio_suspend on both Linux and macOS

SteveLauC:
## What does this PR do

Ref:

* [test_aio_suspend crashes on Linux with rust nightly](https://github.com/nix-rust/nix/issues/1099)
* [aio_suspend returns EINVAL on OSX in CI](https://github.com/nix-rust/nix/issues/1361)

Since we migrated our macOS CI to GHA, Let's try if it works on GHA. For that glibc bug, let's see if it got fixed in the past 2 years.

## Checklist:

### DIFF
--- a/test/sys/test_aio.rs
+++ b/test/sys/test_aio.rs
@@ -569,12 +569,6 @@ fn test_aio_cancel_all() {
 }
 
 #[test]
-// On Cirrus on Linux, this test fails due to a glibc bug.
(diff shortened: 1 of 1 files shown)


SteveLauC:
Look good, at least for now, I just realized that our GHA is using Ubuntu 20.04, probably we should bump it to 22.04 in order to get a newer libc

asomers (approved):

PR #2266, opened 2023-12-15
add the TCP_INFO constant for NetBSD/FreeBSD.

devnexen:


asomers (commented):
tcp_info is tricky, because it is a version-dependent structure.  The size at least is constant for FreeBSD 6 through 14.  But it grows on FreeBSD 15.  It isn't really meant to be used across different OS versions.  Probably the best way to use it from Rust would be to run bindgen at build time (which of course precludes cross-compiling).  Given that, is it really safe to include in Nix?

devnexen:
For FreeBSD, it ensures the size remains the same through versions (at least until it reaches the limit of the padding) or do you mean something else ?

asomers:
That's just the thing; in FreeBSD 15 the size increases.

devnexen:
oh I see yes the layout had been changed a bit this year.

PR #2267, opened 2023-12-16
feat: AT_EMPTY_PATH for fchownat on Linux/Android/FreeBSD/Hurd

### LINKED ISSUE #1029: Missing Flag For nix::unistd::fchownat

SteveLauC:
## What does this PR do

Picks up #1128

Closes #1128 Closes #1029

Enable the `AT_EMPTY_PATH` flag for `fchownat()` on Linux/Android/FreeBSD/Hurd

## Checklist:

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -737,11 +737,17 @@ pub fn fchown(fd: RawFd, owner: Option<Uid>, group: Option<Gid>) -> Result<()> {
     Errno::result(res).map(drop)
 }
 
-/// Flags for `fchownat` function.
-#[derive(Clone, Copy, Debug)]
-pub enum FchownatFlags {
-    FollowSymlink,
-    NoFollowSymlink,
+// Just a wrapper around `AtFlags` so that we can help our users migrate.
+#[cfg(not(target_os = "redox"))]
+pub type FchownatFlags = AtFlags;
+#[cfg(not(target_os = "redox"))]
+impl FchownatFlags {
+    #[deprecated(since = "0.28.0", note = "The variant is deprecated, please use `AtFlags` instead")]
+    #[allow(non_upper_case_globals)]
+    pub const FollowSymlink: FchownatFlags = FchownatFlags::empty();
+    #[deprecated(since = "0.28.0", note = "The variant is deprecated, please use `AtFlags` instead")]
+    #[allow(non_upper_case_globals)]
+    pub const NoFollowSymlink: FchownatFlags = FchownatFlags::AT_SYMLINK_NOFOLLOW;
 }
 
 /// Change the ownership of the file at `path` to be owned by the specified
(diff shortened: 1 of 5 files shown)


src/unistd.rs
```diff
@@ -735,11 +735,21 @@ pub fn fchown(fd: RawFd, owner: Option<Uid>, group: Option<Gid>) -> Result<()> {
...
 }
 
-/// Flags for `fchownat` function.
-#[derive(Clone, Copy, Debug)]
-pub enum FchownatFlags {
-    FollowSymlink,
-    NoFollowSymlink,
+libc_bitflags! {
+    /// Flags for the `fchownat` function.
+    pub struct FchownatFlags: c_int {
```
SteveLauC, line 740: For this type, I am not sure if we should define a new type or just use the [AtFlags](https://docs.rs/nix/latest/nix/fcntl/struct.AtFlags.html) type.

It is safer to define a new type so that we can prevent users from passing invalid arguments, but this will add our maintainence burgen. We are already using the AtFlags type for other functions like [fstatat](https://docs.rs/nix/latest/nix/sys/stat/fn.fstatat.html)

asomers: If we're already using AtFlags for more than one function, I think we can use it here, too.

src/unistd.rs
```diff
@@ -735,11 +735,22 @@ pub fn fchown(fd: RawFd, owner: Option<Uid>, group: Option<Gid>) -> Result<()> {
...
+        /// If  pathname is an empty string, operate on the file referred to
+        /// by `dirfd` (which may have been obtained using the `open(2)`
+        /// `O_PATH` flag). In this case, `dirfd` can refer to any type of
+        /// file, not just a directory.
+        ///
+        /// If `dirfd` is `None`, the call operates on the current working
+        /// directory.
+        #[cfg(linux_android)]
```
asomers, line 749: It's also available on FreeBSD, beginning in 13.1.

asomers (commented):
This is a rare case of an API change in Nix where we can help users migrate.  What about defining some deprecated constants like this?
```rust
impl FchownatFlags {
     #[deprecated]
    pub const FollowSymlink = FchownatFlags::empty();
    #[deprecated]
    pub const NoFollowSymlink = FchownatFlags::AT_SYMLINK_NOFOLLOW;
}
```

src/unistd.rs
SteveLauC: Let me do this in a separate PR

src/unistd.rs
SteveLauC: Get it

SteveLauC:
> This is a rare case of an API change in Nix where we can help users migrate. What about defining some deprecated constants like this?
> 
> ```rust
> impl FchownatFlags {
>      #[deprecated]
>     pub const FollowSymlink = FchownatFlags::empty();
>     #[deprecated]
>     pub const NoFollowSymlink = FchownatFlags::AT_SYMLINK_NOFOLLOW;
> }
> ```

Done

----

Well, fixing imports is so painful, I really hope Nix never do

```
use xxx
```

but directly use the items when needed.

asomers (approved):

PR #2268, opened 2023-12-16
refactor: make dup3 wrapper call the real dup3

### LINKED ISSUE #939: `unistd::dup3` never uses the `dup3` syscall (even if it might be available)

SteveLauC:
## What does this PR do

Pick up #960, closes #939

Before this PR, we were emulating `dup3()` with `dup2()` and `fcntl()`, and thus can cause a race condition, this PR makes the `dup3()` wrapper involve the real `dup3()`.

## Checklist:

### DIFF
--- a/changelog/2268.changed.md
+++ b/changelog/2268.changed.md
@@ -1,0 +1,1 @@
+Changed the `dup3` wrapper to perform a real call to `dup3` instead of emulating it via `dup2` and `fcntl` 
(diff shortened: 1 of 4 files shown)


changelog/2268.changed.md
```diff
@@ -0,0 +1 @@
+Changed the `dup3` wrapper to perform a real call to `dup3` instead of emulating it via `dup2` and `fcntl`
```
asomers, line 1: I would call this "Fixed"

asomers (approved):

changelog/2268.changed.md
SteveLauC: Good catch! done

asomers (approved):

Issue #2264, opened 2023-12-14
listen does not allow specifying a negative backlog


asomers:
The `listen` function, in C, takes a "backlog" argument as an `int`.  It normally must be positive.  However, it may be ngative:
* On FreeBSD, since version 2.2, to use the maximum value
* On Linux, as an undocumented feature, to use a default value, at least according to @devnexen .

But when `listen` was added to Nix in change f85e9bfbe5bfef4d8ce5985d29d9a173aa4d5fb0 , the backlog argument's type was specified as `usize`, which precludes negative values.  It also will rollover at `i32::MAX_VALUE` with unpredictable consequences.

asomers:
See https://github.com/rust-lang/rust/pull/97963 for a discussion of the behavior on Linux.

SteveLauC:
So we know that on FreeBSD, [a negative backlog or a value bigger than the max one means the default value](https://github.com/freebsd/freebsd-src/blob/e7d02be19d40063783d6b8f1ff2bc4c7170fd434/sys/kern/uipc_socket.c#L1069). And from that PR, Linux will see `-1` as the default value, what about other negative values, will they also be seen the default value?

Maybe we should use a wrapper type, just like the one introduced in #1876

SteveLauC:
Ok, from the Linux source code, they [directly cast the `backlog` argument from `int` to `unsigned int`](https://github.com/torvalds/linux/blob/c7402612e2e61b76177f22e6e7f705adcbecc6fe/net/socket.c#L1874-L1875), for a `somaxconn` of value 4096, all the negative values will be bigger than it and then set to it:

> This behavior depends on the value of `somaxconn`, so it is different from the FreeBSD one, we should probably not use other negative value

```
int __sys_listen(int fd, int backlog)
{
	struct socket *sock;
	int err, fput_needed;
	int somaxconn;

	sock = sockfd_lookup_light(fd, &err, &fput_needed);
	if (sock) {
		somaxconn = READ_ONCE(sock_net(sock->sk)->core.sysctl_somaxconn);
		if ((unsigned int)backlog > somaxconn)
			backlog = somaxconn;

		err = security_socket_listen(sock, backlog);
		if (!err)
			err = READ_ONCE(sock->ops)->listen(sock, backlog);

		fput_light(sock->file, fput_needed);
	}
	return err;
}
```

devnexen:
[...]

PR #2276, opened 2023-12-28
sys::socket listen's Backlog wrapper type addition.

devnexen:
changing the sys::socket::listen backlog type from `usize` to a `i32` wrapper, offering known sane values, from -1, SOMAXCONN to
 511.

close gh-2264

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -2009,12 +2009,48 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
     unsafe { Ok((OwnedFd::from_raw_fd(fds[0]), OwnedFd::from_raw_fd(fds[1]))) }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Backlog(i32);
+
+impl Backlog {
+    /// Sets the listen queue size to system `SOMAXCONN` value
+    pub const MAXCONN: Self = Self(libc::SOMAXCONN);
+    /// Sets the listen queue size to -1 for system supporting it
+    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+    pub const MAXALLOWABLE: Self = Self(-1);
+
+    /// Create a `Backlog`, an `EINVAL` will be returned if `val` is invalid.
+    pub fn new<I: Into<i32>>(val: I) -> Result<Self> {
+        cfg_if! {
+            if #[cfg(any(target_os = "linux", target_os = "freebsd"))] {
+                const MIN: i32 = -1;
+            } else {
+                const MIN: i32 = 0;
+            }
+        }
+
+        let val = val.into();
+
+        if !(MIN..Self::MAXCONN.0).contains(&val) {
+            return Err(Errno::EINVAL);
+        }
+
+        Ok(Self(val))
+    }
+}
+
+impl From<Backlog> for i32 {
+    fn from(backlog: Backlog) -> Self {
+        backlog.0
+    }
+}
+
 /// Listen for connections on a socket
 ///
 /// [Further reading](https://pubs.opengroup.org/onlinepubs/9699919799/functions/listen.html)
-pub fn listen<F: AsFd>(sock: &F, backlog: usize) -> Result<()> {
+pub fn listen<F: AsFd>(sock: &F, backlog: Backlog) -> Result<()> {
     let fd = sock.as_fd().as_raw_fd();
-    let res = unsafe { libc::listen(fd, backlog as c_int) };
+    let res = unsafe { libc::listen(fd, backlog.into()) };
 
     Errno::result(res).map(drop)
 }
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -1611,7 +1611,9 @@ fn test_impl_scm_credentials_and_rights(mut space: Vec<u8>) {
 // Test creating and using named unix domain sockets
 #[test]
 pub fn test_named_unixdomain() {
-    use nix::sys::socket::{accept, bind, connect, listen, socket, UnixAddr};
+    use nix::sys::socket::{
+        accept, bind, connect, listen, socket, Backlog, UnixAddr,
+    };
     use nix::sys::socket::{SockFlag, SockType};
     use nix::unistd::{read, write};
     use std::thread;
@@ -1627,7 +1629,7 @@ pub fn test_named_unixdomain() {
     .expect("socket failed");
     let sockaddr = UnixAddr::new(&sockname).unwrap();
     bind(s1.as_raw_fd(), &sockaddr).expect("bind failed");
-    listen(&s1, 10).expect("listen failed");
+    listen(&s1, Backlog::new(10).unwrap()).expect("listen failed");
 
     let thr = thread::spawn(move || {
         let s2 = socket(
@@ -1648,6 +1650,14 @@ pub fn test_named_unixdomain() {
     thr.join().unwrap();
 
     assert_eq!(&buf[..], b"hello");
+}
+
+#[test]
+pub fn test_listen_wrongbacklog() {
+    use nix::sys::socket::Backlog;
+
+    assert!(Backlog::new(libc::SOMAXCONN + 1).is_err());
+    assert!(Backlog::new(-2).is_err());
 }
 
 // Test using unnamed unix domain addresses
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { version = "0.2.151", features = ["extra_traits"] }
+libc = { git = "https://github.com/rust-lang/libc", rev = "6a203e955b60cca48562f020f0e4e003079f3199", features = ["extra_traits"] }
 bitflags = "2.3.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
--- a/changelog/2276.added.md
+++ b/changelog/2276.added.md
@@ -1,0 +1,1 @@
+Added the `Backlog` wrapper type for the `listen` call.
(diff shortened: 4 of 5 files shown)


src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,116 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
     unsafe { Ok((OwnedFd::from_raw_fd(fds[0]), OwnedFd::from_raw_fd(fds[1]))) }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Backlog(i32);
+
+impl Backlog {
+    /// Sets the listen queue size to system `SOMAXCONN` value
+    #[cfg(target_os = "redox")]
+    pub const DEFAULT: Self = Self(128);
```
SteveLauC, line 2018: Redox has `SOMAXCONN` defined, we should use it:

```sh
$ pwd
/home/steve/Documents/workspace/libcs/libc/src/unix/redox

$ rg "SOMAXCONN"
mod.rs
816:pub const SOMAXCONN: ::c_int = 128;
```

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,116 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+    #[cfg(not(target_os = "redox"))]
+    pub const DEFAULT: Self = Self(libc::SOMAXCONN);
```
SteveLauC, line 2020: Why are we calling this value `DEFAULT`? But not `MAX` or simply `SOMAXCONN`

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,116 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+    /// Sets the listen queue size to -1 for system supporting it
+    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+    pub const MAXALLOWABLE: Self = Self(-1);
+    /// Reasonable upper limit, while in theory we might be able
+    /// to set it to a higher boundary, comes with a risk
+    /// of resource exhaustion in return
+    pub const MAXVALUE: Self = Self(511);
```
SteveLauC, line 2027: Is it really necessary to expose a `MAXVALUE` that is specified by us rather than the system?

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,116 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+        backlog.0
+    }
+}
+
+impl TryFrom<i64> for Backlog {
+    type Error = BacklogTryFromError;
+    fn try_from(backlog: i64) -> std::result::Result<Self, Self::Error> {
+        match backlog {
+            ..=-2 => Err(BacklogTryFromError::TooNegative),
+            -1..=511 => Ok(Self(i32::try_from(backlog).map_err(|_| BacklogTryFromError::TooPositive)?)),
```
SteveLauC, line 2076: If a backlog of value `-1` is only valid on Linux and FreeBSD, why are we allowing it for other OSes here?

SteveLauC (changes requested):
Thanks for this PR! I have left some comments. BTW, I would appreciate it if we can have doc comments for these new public types.

A changelog is also needed.

src/sys/socket/mod.rs
devnexen: It is mostly what people use we can call it differently sure.

src/sys/socket/mod.rs
devnexen: We might refine it a bit (e.g. linux/glibc which has SOMAXCONN to 4096, for this case we might go up to 65535).

src/sys/socket/mod.rs
devnexen: sure. still a draft :) just wanted first impressions.

src/sys/socket/mod.rs
SteveLauC: IMHO, we should not put a restriction that the OS does not have, so I kinda tend to remove this and use SOMAXCONN as the upper bound

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,130 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+    fn from(backlog: u16) -> Self {
+        Self(i32::from(backlog))
+    }
+}
+
+impl From<u8> for Backlog {
+    fn from(backlog: u8) -> Self {
+        Self(i32::from(backlog))
+    }
+}
```
SteveLauC, line 2059: Just realized that we need to check the value even the value is in range of `i32`, so what about defining a `new(val: i32)` function:

```rs
/// The `backlog` argument of [`listen()`].
pub struct Backlog(i32);
impl Backlog {
    /// Sets the listen queue size to system `SOMAXCONN` value
    pub const MAXCONN: Backlog = Backlog(libc::SOMAXCONN);
    /// Sets the listen queue size to -1 for system supporting it
    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
    pub const MAXALLOWABLE: Backlog = Backlog(-1);

    /// Create a `Backlog`, an `EINVAL` will be returned if `val` is invalid.
    pub fn new(val: i32) -> Result<Self> {
        cfg_if! {
            if #[cfg(any(target_os = "linux", target_os = "freebsd"))] {
                const MIN: i32 = -1;
            } else {
                const MIN: i32 = 0;
            }
        }

        if !(MIN..libc::SOMAXCONN).contains(&val) {
            return Err(Errno::EINVAL);
        }

        Ok(Self(val))
    }
}
```

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,145 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+
+impl Backlog {
+    /// Sets the listen queue size to system `SOMAXCONN` value
+    pub const MAXCONN: Self = Self(libc::SOMAXCONN);
+    /// Sets the listen queue size to -1 for system supporting it
+    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+    pub const MAXALLOWABLE: Self = Self(-1);
+
+    // Create a `Backlog`, an `EINVAL` will be returned if `val` is invalid.
+    pub fn new<I: Into<i32> + PartialOrd<I> + From<i32>>(val: I) -> Result<Self> {
```
SteveLauC, line 2023: If `I` is `Into<i32>`, why not just cast it to `i32` and do the comparison:

```rs
let val = val.into();
```

And, with this generic function, the other `From/TryFrom` impls are no longer needed, they are making things overly complicated.

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,48 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+pub struct Backlog(i32);
+
+impl Backlog {
+    /// Sets the listen queue size to system `SOMAXCONN` value
+    pub const MAXCONN: Self = Self(libc::SOMAXCONN);
+    /// Sets the listen queue size to -1 for system supporting it
+    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+    pub const MAXALLOWABLE: Self = Self(-1);
+
+    // Create a `Backlog`, an `EINVAL` will be returned if `val` is invalid.
```
SteveLauC, line 2022: Please use a doc comment:

```suggestion
    /// Create a `Backlog`, an `EINVAL` will be returned if `val` is invalid.
```

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,48 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+    fn from(backlog: Backlog) -> Self {
+        backlog.0
+    }
+}
+
 /// Listen for connections on a socket
 ///
 /// [Further reading](https://pubs.opengroup.org/onlinepubs/9699919799/functions/listen.html)
-pub fn listen<F: AsFd>(sock: &F, backlog: usize) -> Result<()> {
+pub fn listen<F: AsFd, B: Into<Backlog>>(sock: &F, backlog: B) -> Result<()> {
```
SteveLauC, line 2051: With the current impl, we can just use the `Backlog` type here:

```suggestion
pub fn listen<F: AsFd>(sock: &F, backlog: Backlog) -> Result<()> {
```

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,48 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
     let fd = sock.as_fd().as_raw_fd();
-    let res = unsafe { libc::listen(fd, backlog as c_int) };
+    let res = unsafe { libc::listen(fd, i32::from(backlog.into())) };
```
SteveLauC, line 2053: `Backlog` is `Into<i32>`, so we can simply do this:

```suggestion
    let res = unsafe { libc::listen(fd, backlog.into()) };
```

test/sys/test_socket.rs
```diff
@@ -1650,6 +1652,15 @@ pub fn test_named_unixdomain() {
     assert_eq!(&buf[..], b"hello");
 }
 
+#[test]
+pub fn test_listen_wrongbacklog() {
+    use nix::sys::socket::Backlog;
+
+    assert!(Backlog::new(5012).is_err());
+    assert!(Backlog::new(65535).is_err());
```
SteveLauC, line 1660: For these 2 assertions, why not use the `libc::SOMAXCONN`, like this:

```rust
assert!(Backlog::new(libc::SOMAXCONN + 1).is_err());
```

SteveLauC (changes requested):

SteveLauC (approved):
Thanks!

src/sys/socket/mod.rs
```diff
@@ -2009,12 +2009,48 @@ pub fn socketpair<T: Into<Option<SockProtocol>>>(
...
+            if #[cfg(any(target_os = "linux", target_os = "freebsd"))] {
+                const MIN: i32 = -1;
+            } else {
+                const MIN: i32 = 0;
+            }
+        }
+
+        let val = val.into();
+
+        if !(MIN..Self::MAXCONN.0).contains(&val) {
```
gabiganam, line 2034: this check does not include `MAXCONN` so `Backlog::new(4096)` returns an error, while it should be valid.

PR #2281, opened 2024-01-06
fix: unused struct tuple fields on DragonFlyBSD

SteveLauC:
## What does this PR do

~~Test the DragonFlyBSD CI failure in #2277, theoretically, I should be able to reproduce it in this PR.~~

Fix that CI failure by converting the type `SendfileHeaderTrailer` into a real struct and explicitly mark the fields that are not in use as unused with a `_` prefix.

## Checklist:

### DIFF
--- a/src/sys/sendfile.rs
+++ b/src/sys/sendfile.rs
@@ -81,11 +81,11 @@ cfg_if! {
         use std::io::IoSlice;
 
         #[derive(Clone, Debug)]
-        struct SendfileHeaderTrailer<'a>(
-            libc::sf_hdtr,
-            Option<Vec<IoSlice<'a>>>,
-            Option<Vec<IoSlice<'a>>>,
-        );
+        struct SendfileHeaderTrailer<'a> {
+            raw: libc::sf_hdtr,
+            _headers: Option<Vec<IoSlice<'a>>>,
+            _trailers: Option<Vec<IoSlice<'a>>>,
+        }
 
         impl<'a> SendfileHeaderTrailer<'a> {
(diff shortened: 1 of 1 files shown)


SteveLauC:
Maybe this is not a bug but a feature, such a check was introduced in [rust#118297: Merge unused_tuple_struct_fields into dead_code](https://github.com/rust-lang/rust/pull/118297)

asomers (commented):
It looks like those fields only exist in order to keep a slice from dropping while we have a pointer to it.  What about changing the function signature so we keep the user's original slice around?  Something like:
```rust
fn send_file_header_trailer(headers: &Option<Vec<IoSlice<'a>>>, trailers: ...) -> Option<libc::sf_hdtr> {
    ...
}
```

asomers:
> It looks like those fields only exist in order to keep a slice from dropping while we have a pointer to it. What about changing the function signature so we keep the user's original slice around? Something like:
> 
> ```rust
> fn send_file_header_trailer(headers: &Option<Vec<IoSlice<'a>>>, trailers: ...) -> Option<libc::sf_hdtr> {
>     ...
> }
> ```

Ahh, I see now why it was done this way.  Because sf_hdtr needs its headers and trailers to be in the form of a slice of `iovec`, which has a different layout than `&[&[u8]]`.

asomers (approved):
LGTM, except that the commit message is inaccurate.  It applies to FreeBSD and Apple as well.  But we don't test those on CI with a nightly compiler, which is why we only noticed for DragonflyBSD.

PR #2283, opened 2024-01-06
Introduce set_errno

PolyMeilex:
## What does this PR do

This adds a way to set errno, it makes it possible to fully drop errno crate in crates that already use nix. 

## Checklist:

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -79,8 +79,8 @@ impl Errno {
     ///
     /// assert_eq!(Errno::last(), Errno::EIO);
     /// ```
-    pub fn set(errno: Self) {
-        Self::set_raw(errno as i32)
+    pub fn set(self) {
+        Self::set_raw(self as i32)
     }
 
     /// Sets the raw i32 value of errno.
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -4,7 +4,7 @@
 //! ```
 //! use nix::errno::Errno;
 //!
-//! Errno::set(Errno::EIO);
+//! Errno::EIO.set();
 //! assert_eq!(Errno::last(), Errno::EIO);
 //!
 //! Errno::clear();
@@ -75,7 +75,7 @@ impl Errno {
     /// ```
     /// use nix::errno::Errno;
     ///
-    /// Errno::set(Errno::EIO);
+    /// Errno::EIO.set();
     ///
     /// assert_eq!(Errno::last(), Errno::EIO);
     /// ```
@@ -113,7 +113,7 @@ impl Errno {
     /// ```
     /// use nix::errno::Errno;
     ///
-    /// Errno::set(Errno::EIO);
+    /// Errno::EIO.set();
     ///
     /// Errno::clear();
     ///
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -52,6 +52,14 @@ pub fn errno() -> i32 {
     unsafe { *errno_location() }
 }
 
+/// Sets the platform-specific value of errno.
+pub fn set_errno(errno: Errno) {
+    // Safe because errno is a thread-local variable
+    unsafe {
+        *errno_location() = errno as i32;
+    }
+}
+
 impl Errno {
     pub fn last() -> Self {
         last()
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -53,7 +53,7 @@ cfg_if! {
 }
 
 /// Returns the platform-specific value of errno
-#[deprecated(note = "please use `Errno::last_raw()` instead")]
+#[deprecated(since = "0.28.0", note = "please use `Errno::last_raw()` instead")]
 pub fn errno() -> i32 {
     Errno::last_raw()
 }
@@ -91,7 +91,10 @@ impl Errno {
         }
     }
 
-    #[deprecated(note = "please use `Errno::from_raw()` instead")]
+    #[deprecated(
+        since = "0.28.0",
+        note = "please use `Errno::from_raw()` instead"
+    )]
     pub const fn from_i32(err: i32) -> Errno {
         Self::from_raw(err)
     }
@@ -981,7 +984,10 @@ mod consts {
         pub const ENOTSUP: Errno = Errno::EOPNOTSUPP;
     }
 
-    #[deprecated(note = "please use `Errno::from_raw()` instead")]
+    #[deprecated(
+        since = "0.28.0",
+        note = "please use `Errno::from_raw()` instead"
+    )]
     pub const fn from_i32(e: i32) -> Errno {
         use self::Errno::*;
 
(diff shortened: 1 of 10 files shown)


asomers (changes requested):
How about a usage example?  And FTR, what function from the errno crate does this replace?  Oh, and don't worry about the test failure on DragonflyBSD; it's unrelated.

PolyMeilex:
> How about a usage example? 

If you are asking for a use case, I simply noticed that fish shell depends both on nix and errno, so I digged into it and that's how I found out that nix does not have a way to set errnos.
As to why they need a way to set errno, the project is transitioning from C++ so errno is being used on the API boundary in some places I suppose.
But same need would arise if one is writing rust libs meant for consumption from C.

And if you are asking me to add rustdoc usage example, sure I can do that, just let me know that I misunderstood the question.

> And FTR, what function from the errno crate does this replace?

Simply [errno::set_errno](https://docs.rs/errno/latest/errno/fn.set_errno.html)

asomers:
Yeah, a rustdoc example would be great.  Also, you'll have to rebase to fix the CI failure.

PolyMeilex:
Rebased, examples added to module docs, and to set and clear functions

PolyMeilex:
I'm not sure if this should be top level function, if clear is a static method already?

Currently we have:
```rs
pub fn errno() -> i32;
pub fn set_errno(errno: Errno);

impl Errno {
    pub fn clear(); // Does the same'ish action as set but lives in diferent scope
    pub fn last() -> Errno;
}
```

So that's not consistent.

But moving set to a method does not seem entirely right either?
```rs
impl Errno {
    pub fn set(&self); // Like so?
    pub fn set(errno: Errno) // Or perhaps like this?
}
```

Personally I would just move clear to top level as well to make it consistent, but that's already part of public API, so I would rather not break it for no reason.

asomers:
Excellent question.  I agree that the current situation is inconsistent.  I'm a fan of namespaces, so my preference would be:
```rust
impl Errno {
    fn set(self) {...}
    fn clear() {...}
    fn last() { ...}
}
```

SteveLauC:
> I'm not sure if this should be top level function, if clear is a static method already?
> 
> Currently we have:
> 
> ```rust
> pub fn errno() -> i32;
> pub fn set_errno(errno: Errno);
> 
> impl Errno {
>     pub fn clear(); // Does the same'ish action as set but lives in diferent scope
>     pub fn last() -> Errno;
> }
> ```
> 
> So that's not consistent.

It is indeed not consistent. 

I think the following approach is ok:

```rs
impl Errno {
    pub fn set(errno: Errno) {}
```

So when people use it, `Errno::` would just behave like a namespace.

------

For `nix::errno::errno() -> i32`, Maybe we should also move it to:

```rs
impl Errno {
    /// Returns the platform-specific value of errno
    pub fn last_raw() -> i32 { }
}
```

And `nix::errno::from_i32()` (`consts::from_i32()`) should be marked `pub(crate)` so that we don't expose 2 interfaces for 1 thing.

@asomers thoughts on this?

asomers:
I agree with renaming `errno()` to `Errno::last_raw`.  But what exactly do you want to rename `from_raw`, and why?

SteveLauC:
> I agree with renaming `errno()` to `Errno::last_raw`. But what exactly do you want to rename `from_raw`, and why?

Just for symmetry:

```rs
impl Errno {
    pub fn from_raw(raw: i32) -> Errno
    pub fn last_raw() -> i32
}

**Comment by SteveLauC** [comment]
To deprecate an interface, we do it like [this](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#removedeprecate-an-interface).

**Comment by PolyMeilex** [comment]
> To deprecate an interface, we do it like [this](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#removedeprecate-an-interface).
> 
> The version of the next release will be `0.28.0`.

Done.

Although I have no idea why the CI is failing, I can't figure out where is the code that causes this error:
```
error: use of deprecated associated function `errno::<impl errno::consts::Errno>::from_i32`: please use `Errno::from_raw()` instead
   --> src/time.rs:251:20
    |
251 |         Err(Errno::from_i32(ret))
    |                    ^^^^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(deprecated)]`
```
There is no line 251 in nix/time module :shrug:

**Comment by SteveLauC** [comment]
> > I agree with renaming `errno()` to `Errno::last_raw`. But what exactly do you want to rename `from_raw`, and why?
> 
> Just for symmetry:
> 
> ```rust
> impl Errno {
>     pub fn from_raw(raw: i32) -> Errno
>     pub fn last_raw() -> i32
> }
> ```

LGTM! Thanks!

But I would like to know @asomers's view on the new APIs before merging this PR.

File: src/errno.rs [hunk 1]
```diff
@@ -39,34 +52,80 @@ cfg_if! {
...
+    ///
+    /// # Example
+    /// ```
+    /// use nix::errno::Errno;
+    ///
+    /// Errno::set(Errno::EIO);
+    ///
+    /// assert_eq!(Errno::last(), Errno::EIO);
+    /// ```
+    pub fn set(errno: Self) {
```
asomers, line 82: This API is redundant: the user must type "Errno" twice.  I think it would be better like this:
```rust
pub fn set(self);
```
that way a user could call it simply by:
```rust
Errno::EINVAL.set();
```

asomers (changes requested):

src/errno.rs
```diff
@@ -1,3 +1,16 @@
+//! Safe wrappers around errno functions
+//!
+//! # Example
+//! ```
+//! use nix::errno::Errno;
+//!
+//! Errno::set(Errno::EIO);
```
SteveLauC, line 7: Doc needs an update

src/errno.rs
PolyMeilex: I find this confusing, but using self does not stop one from still doing `Errno::set(err)` syntax, so I'm fine with this.
Done

src/errno.rs
PolyMeilex: Updated anyway already, but yeah I initially left that intentionally as a matter of preference.

src/errno.rs
SteveLauC: > ```rs
> Errno::EINVAL.set();
> ```
> I find this confusing,

This indeed feels a little bit weird, but anyway, it is ok, no strong view on this.

test/test_errno.rs
```diff
@@ -0,0 +1,8 @@
+use nix::errno::Errno;
+
+#[test]
+fn errno_set_and_read() {
+    Errno::clear();
```
asomers, line 5: This test doesn't actually verify that clear does what it's supposed to.  You should probably have one test for set and another for clear.

asomers (changes requested):

test/test_errno.rs
PolyMeilex: Done

asomers: But why leave the Errno::clear there on line 5?  You should just delete it.

PolyMeilex: Oh, yeah, set will override it anyway, not sure why I even added it in the first place :facepalm: 
Done

asomers (approved):

asomers (approved):

SteveLauC (approved):
Thanks!

PR #2289, opened 2024-01-14
Add docs to the time and fcntl modules

asomers:
Also, deprecate FlockArg::UnlockNonblock.  That combination of flags doesn't make sense.  And indeed, reviewing the kernel source code shows that the kernel will ignore LOCK_NB when LOCK_UN is set.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1,3 +1,4 @@
+//! file control options
 use crate::errno::Errno;
 #[cfg(all(target_os = "freebsd", target_arch = "x86_64"))]
 use core::slice;
@@ -45,15 +46,28 @@ pub use self::posix_fadvise::{posix_fadvise, PosixFadviseAdvice};
 #[cfg(not(target_os = "redox"))]
 #[cfg(any(feature = "fs", feature = "process", feature = "user"))]
 libc_bitflags! {
+    /// Flags that control how the various *at syscalls behave.
     #[cfg_attr(docsrs, doc(cfg(any(feature = "fs", feature = "process"))))]
     pub struct AtFlags: c_int {
+        /// Used with [`unlink`](crate::unistd::unlink) to indicate that the path to remove is a
+        /// directory, and not a regular file.
         AT_REMOVEDIR;
+        /// Used with [`linkat`](crate::unistd::linkat`) to create a link to a symbolic link's
+        /// target, instead of to the symbolic link itself.
         AT_SYMLINK_FOLLOW;
+        /// Used with [`linkat`](crate::unistd::linkat`) to create a link to a symbolic link
+        /// itself, instead of to the symbolic link's target.
         AT_SYMLINK_NOFOLLOW;
+        /// Don't automount the terminal ("basename") component of pathname if it is a directory
+        /// that is an automount point.
         #[cfg(linux_android)]
         AT_NO_AUTOMOUNT;
+        /// If the provided path is an empty string, operate on the provided directory file
+        /// descriptor instead.
         #[cfg(any(linux_android, target_os = "freebsd", target_os = "hurd"))]
         AT_EMPTY_PATH;
+        /// Used with [`faccessat`](crate::unistd::faccessat), the checks for accessibility are
+        /// performed using the effective user and group IDs instead of the real user and group ID
         #[cfg(not(target_os = "android"))]
         AT_EACCESS;
     }
@@ -186,6 +200,10 @@ pub(crate) fn at_rawfd(fd: Option<RawFd>) -> raw::c_int {
 feature! {
 #![feature = "fs"]
 
+/// open or create a file for reading, writing or executing
+///
+/// # See Also
+/// [`open`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html)
 // The conversion is not identical on all operating systems.
 #[allow(clippy::useless_conversion)]
 pub fn open<P: ?Sized + NixPath>(
@@ -200,6 +218,14 @@ pub fn open<P: ?Sized + NixPath>(
     Errno::result(fd)
 }
 
+/// open or create a file for reading, writing or executing
+///
+/// The `openat` function is equivalent to the [`open`] function except in the case where the path
+/// specifies a relative path.  In that case, the file to be opened is determined relative to the
+/// directory associated with the file descriptor `fd`.
+///
+/// # See Also
+/// [`openat`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/openat.html)
 // The conversion is not identical on all operating systems.
 #[allow(clippy::useless_conversion)]
 #[cfg(not(target_os = "redox"))]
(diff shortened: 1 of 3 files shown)


src/fcntl.rs
```diff
@@ -45,15 +46,28 @@ pub use self::posix_fadvise::{posix_fadvise, PosixFadviseAdvice};
...
+    /// Flags that control how the various *at syscalls behave.
     #[cfg_attr(docsrs, doc(cfg(any(feature = "fs", feature = "process"))))]
     pub struct AtFlags: c_int {
+        /// Used with [`unlink`](crate::unistd::unlink) to indicate that the path to remove is a
+        /// directory, and not a regular file.
         AT_REMOVEDIR;
+        /// Used with [`linkat`](crate::unistd::linkat`) to create a link to a symbolic link's
+        /// target, instead of to the symbolic link itself.
         AT_SYMLINK_FOLLOW;
+        /// Used with [`linkat`](crate::unistd::linkat`) to create a link to a symbolic link
```
SteveLauC, line 58: `linkat()` does not support this flag as it is the default behavior per the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html)

src/fcntl.rs
```diff
@@ -45,15 +46,28 @@ pub use self::posix_fadvise::{posix_fadvise, PosixFadviseAdvice};
 #[cfg(not(target_os = "redox"))]
 #[cfg(any(feature = "fs", feature = "process", feature = "user"))]
 libc_bitflags! {
+    /// Flags that control how the various *at syscalls behave.
     #[cfg_attr(docsrs, doc(cfg(any(feature = "fs", feature = "process"))))]
     pub struct AtFlags: c_int {
+        /// Used with [`unlink`](crate::unistd::unlink) to indicate that the path to remove is a
```
SteveLauC, line 52: It should be `unlinkat()`:

```suggestion
        /// Used with [`unlinkat`](crate::unistd::unlinkat) to indicate that the path to remove is a
```

And, our `unlinkat()` function does not take a `AtFlags` but a [`UnlinkatFlags`](https://github.com/nix-rust/nix/blob/bfcb2269d1f5e97723a1639cf6a33a85814afdf3/src/unistd.rs#L1312-L1317).

src/fcntl.rs
```diff
@@ -239,16 +273,30 @@ pub fn renameat<P1: ?Sized + NixPath, P2: ?Sized + NixPath>(
...
+        /// This operation makes sense only for overlay/union filesystem implementations.
         RENAME_WHITEOUT;
     }
 }
 
 feature! {
 #![feature = "fs"]
+/// Like [`renameat`], but with an additional `flags` argument.
+///
+/// A `renameat2` call with a zero flags argument is equivalent to `renameat`.
```
SteveLauC, line 296: Maybe the following one is more suitable for our interface:
```suggestion
/// A `renameat2` call with an empty flags argument is equivalent to `renameat`.
```

src/fcntl.rs
```diff
@@ -584,17 +679,27 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
     Errno::result(res)
 }
 
+/// Operations for use with [`flock`].
```
SteveLauC, line 682: The `flock()` function is deprecated so let's use the new `Flock::lock()` interface:
```suggestion
/// Operations for use with [`Flock::lock()`].
```

src/fcntl.rs
```diff
@@ -584,17 +679,27 @@ pub fn fcntl(fd: RawFd, arg: FcntlArg) -> Result<c_int> {
...
+#[cfg(not(any(target_os = "redox", target_os = "solaris")))]
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 #[non_exhaustive]
 pub enum FlockArg {
+    /// shared file lock
     LockShared,
+    /// exclusive file lock
     LockExclusive,
+    /// Unlock file
     Unlock,
```
SteveLauC, line 692: Maybe we can also deprecate the `Unlock` variant given that this `FlockArg` type is only used with the lock interface: `Flock::lock()`

src/time.rs
```diff
@@ -1,3 +1,4 @@
+//! Sleep, query system clocks, and set system closk
```
SteveLauC, line 1: Seems to be a typo:

```suggestion
//! Sleep, query system clocks, and set system clock
```

src/time.rs
```diff
@@ -57,20 +57,30 @@ impl ClockId {
     }
 
     #[cfg(any(linux_android, target_os = "emscripten", target_os = "fuchsia"))]
+    /// Starts at zero when the kernel boots and increments monotonically in SI seconds while the
+    /// machine is running.
     pub const CLOCK_BOOTTIME: ClockId = ClockId(libc::CLOCK_BOOTTIME);
+    /// Like [`CLOCK_BOOTTIME`](ClockId::CLOCK_BOOTTIME), but will wake the  system  if  it  is
```
SteveLauC, line 63:
```suggestion
    /// Like [`CLOCK_BOOTTIME`](ClockId::CLOCK_BOOTTIME), but will wake the system if it is
```

src/time.rs
```diff
@@ -57,20 +57,30 @@ impl ClockId {
...
     #[cfg(freebsdlike)]
+    /// Like [`CLOCK_MONOTONIC`](ClockId::CLOCK_MONOTONIC), but optimized for execution time at the expense of accuracy.
     pub const CLOCK_MONOTONIC_FAST: ClockId =
         ClockId(libc::CLOCK_MONOTONIC_FAST);
     #[cfg(freebsdlike)]
+    /// Like [`CLOCK_MONOTONIC`](ClockId::CLOCK_MONOTONIC), but optimized for accuracy at the expense of execution time.
     pub const CLOCK_MONOTONIC_PRECISE: ClockId =
         ClockId(libc::CLOCK_MONOTONIC_PRECISE);
+    /// Similar to [`CLOCK_MONOTONIC`](ClockId::CLOCK_MONOTONIC), but provides access to a raw
+    /// hardware-based time that  is  not subject  to  NTP adjustments.
```
SteveLauC, line 83:
```suggestion
    /// hardware-based time that is not subject to NTP adjustments.
```

src/time.rs
```diff
@@ -81,30 +91,43 @@ impl ClockId {
...
+    #[allow(missing_docs)] // Undocumented on Linux!
     #[cfg(any(
         target_os = "emscripten",
         target_os = "fuchsia",
         all(target_os = "linux", target_env = "musl")
     ))]
     pub const CLOCK_SGI_CYCLE: ClockId = ClockId(libc::CLOCK_SGI_CYCLE);
+    /// International Atomic Time.
+    ///
+    /// A  nonsettable  system-wide  clock derived from wall-clock time but ignoring leap seconds.
```
SteveLauC, line 130:
```suggestion
    /// A nonsettable system-wide clock derived from wall-clock time but ignoring leap seconds.
```

SteveLauC (changes requested):

src/fcntl.rs
asomers: In that case, this flag is totally unused.  I'll mark it as hidden and deprecated.

SteveLauC (approved):

PR #2292, opened 2024-01-14
deny missing docs in the kmod module

asomers:


SteveLauC (approved):
LGTM, though it should be "warn missing docs" since that is the [default attribute we use](https://github.com/nix-rust/nix/blob/1acc5e86138ee796f4c6b4a407cdcfc9b70edac0/src/lib.rs#L91)

PR #2300, opened 2024-01-28
Enable fcntl OFD commands on macos

anacrolix:
## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 name        = "nix"
 description = "Rust friendly bindings to *nix APIs"
 edition     = "2021"
+resolver    = "2"
 version     = "0.27.1"
 rust-version = "1.69"
 authors     = ["The nix-rust Project Developers"]
@@ -28,7 +29,6 @@ targets = [
 ]
 
 [dependencies]
-libc = { git = "https://github.com/rust-lang/libc", rev = "2f93bfb7678e18a9fc5373dec49384bd23f601c3", features = ["extra_traits"] }
 bitflags = "2.3.1"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
@@ -70,6 +70,17 @@ ucontext = ["signal"]
 uio = []
 user = ["feature"]
 zerocopy = ["fs", "uio"]
+
+[dependencies.libc]
+git = "https://github.com/anacrolix/rust-libc"
+branch = "macos-ofd-fcntl"
+features = ["extra_traits"]
+
+[dev-dependencies.libc]
+git = "https://github.com/anacrolix/rust-libc"
+branch = "macos-ofd-fcntl"
+# const-extern-fn is needed for tests
+features = ["extra_traits", "const-extern-fn"]
 
 [dev-dependencies]
 assert-impl = "0.1"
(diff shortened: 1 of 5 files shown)


SteveLauC:
Hi, thanks for your interest in contributing to Nix!

If you want to bump the MSRV, I think this should be done in a separate PR, and you should also update this info in:

1. `.cirrus.yml`
2. `README.md`

Cargo.toml
```diff
@@ -2,8 +2,9 @@
 name        = "nix"
 description = "Rust friendly bindings to *nix APIs"
 edition     = "2021"
+resolver    = "2"
```
SteveLauC, line 5: It seems that [this is already enabled since we are using Rust 2021](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html)?

src/sys/select.rs
```diff
@@ -110,7 +106,7 @@ impl<'fd> FdSet<'fd> {
     pub fn fds(&self, highest: Option<RawFd>) -> Fds {
         Fds {
             set: self,
-            range: 0..highest.map(|h| h as usize + 1).unwrap_or(FD_SETSIZE),
+            range: 0..highest.map(|h| h + 1).unwrap_or(FD_SETSIZE) as usize,
```
SteveLauC, line 109: Interesting, they changed the type of this constant from `usize` to `c_int`: https://github.com/rust-lang/libc/pull/3356/files

src/sys/epoll.rs
```diff
@@ -6,7 +6,7 @@ use std::mem;
 use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, OwnedFd, RawFd};
 
 libc_bitflags!(
-    pub struct EpollFlags: c_int {
+    pub struct EpollFlags: u32 {
```
SteveLauC, line 9: Hmm, another type change: https://github.com/rust-lang/libc/pull/3466

SteveLauC (changes requested):

SteveLauC:
And, for changes (logic changes and clippy fixes) requested by a newer version libc, I also recommend doing them in a separate PR.

> Nix uses squash merge so multiple commits in a PR would be treated as a single commit

anacrolix:
I'll follow this up when the libc PR goes through, as it depends heavily on that. Also there are some issues in using 64bit apis on 32bit systems with nix which requires that I work around it for now.

anacrolix:
Feel free to run this to completion. I might revisit it when the libc stuff is merged. I have stopped depending on nix in the meanwhile as it deviates too much from libc.

SteveLauC:
>  it deviates too much from libc

You mean those libc breaking changes? They only exist in branch `main`, which targets libc version 1.0, we are still using branch `libc-0.2` and won't switch to 1.0 in the near future.

anacrolix:
I that why I was having trouble? Am I trying to use nix with the wrong branch of libc?

SteveLauC:
> I that why I was having trouble? Am I trying to use nix with the wrong branch of libc?

I am not sure about the trouble you have, but if you want to use libc, you can use the one re-exported by Nix, which is guaranteed to be compatible with Nix.

anacrolix:
Ok this is ready for review again, thanks.

PR #2301, opened 2024-01-28
Fix compilation error with target_env=ohos

Chilledheart:
OpenHarmony(alias ohos) is tie 3 supported target for rust. see https://github.com/rust-lang/compiler-team/issues/568.

In implementation it sits close to musl instead of bionic and napi instead jni when comparing with android.

## Checklist:

### DIFF
--- a/src/sys/statfs.rs
+++ b/src/sys/statfs.rs
@@ -55,6 +55,8 @@ type fs_type_t = libc::c_ulong;
 type fs_type_t = libc::c_uint;
 #[cfg(all(target_os = "linux", target_env = "musl"))]
 type fs_type_t = libc::c_ulong;
+#[cfg(all(target_os = "linux", target_env = "ohos"))]
+type fs_type_t = libc::c_ulong;
 #[cfg(all(target_os = "linux", target_env = "uclibc"))]
 type fs_type_t = libc::c_int;
 #[cfg(all(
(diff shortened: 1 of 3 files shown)


src/sys/statfs.rs
```diff
@@ -23,7 +23,7 @@ pub type fsid_t = libc::__fsid_t;
 pub type fsid_t = libc::fsid_t;
 
 cfg_if! {
-    if #[cfg(any(linux_android, target_os = "fuchsia"))] {
+    if #[cfg(any(linux_android, target_os = "fuchsia", target_env = "ohos"))] {
```
SteveLauC, line 26: If this `ohos` is a Linux libc, isn't this already included in `linux_android`?

SteveLauC (changes requested):

src/sys/statfs.rs
Chilledheart: Nice catch!

It is a bit awkward that ohos is kind of target_env, but not target_os like android. Anyway, I fixed it along with another two places. 

Tested:
```
cargo build --target aarch64-unknown-linux-ohos --features acct,aio,dir,env,fs
```

SteveLauC (approved):

SteveLauC:
Merge this PR because I don't want Nix to be broken on this target.

Though I believe we should avoid `cfg(not)` as much as possible since targets are constantly added to Rust, using `cfg(not)` makes Nix inherently broken on these newly-added targets.

PR #2311, opened 2024-02-07
fix: unaligned read panic on macos

anonkey:
## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,32 @@ jobs:
       - name: before_cache_script
         run: rm -rf $CARGO_HOME/registry/index
 
+  macos-aarch64:
+    runs-on: macos-14
+    env:
+      TARGET: aarch64-apple-darwin
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: setup Rust
+        uses: dtolnay/rust-toolchain@master
+        with: 
+          toolchain: '${{ env.MSRV }}'
+          components: clippy
+
+      - name: build
+        uses: ./.github/actions/build
+        with:
+          TARGET: "${{ env.TARGET }}"
+
+      - name: test
+        uses: ./.github/actions/test
+        with:
+          TARGET: "${{ env.TARGET }}"
+
+      - name: before_cache_script
+        run: rm -rf $CARGO_HOME/registry/index
 
   # Use cross for QEMU-based testing
   # cross needs to execute Docker, GitHub Action already has it installed
--- a/changelog/2311.fixed.md
+++ b/changelog/2311.fixed.md
@@ -1,0 +1,1 @@
+Fixed `::unistd::Group::members` using read_unaligned to avoid crash on misaligned pointers
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -3476,18 +3476,17 @@ impl Group {
         let mut ret = Vec::new();
 
         for i in 0.. {
-            let u = unsafe { mem.offset(i) };
-            if unsafe { (*u).is_null() } {
+            let u = unsafe { mem.offset(i).read_unaligned() };
+            if u.is_null() {
                 break;
             } else {
-                let s = unsafe {CStr::from_ptr(*u).to_string_lossy().into_owned()};
+                let s = unsafe {CStr::from_ptr(u).to_string_lossy().into_owned()};
                 ret.push(s);
             }
         }
 
         ret
     }
-
     /// # Safety
     ///
     /// If `f` writes to its `*mut *mut libc::group` parameter, then it must
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -1372,3 +1372,14 @@ fn test_eaccess_file_exists() {
     eaccess(&path, AccessFlags::R_OK | AccessFlags::W_OK)
         .expect("assertion failed");
 }
+
+#[test]
+#[cfg(bsd)]
+fn test_group_from() {
+    let group = Group::from_name("wheel").unwrap().unwrap();
+    assert!(group.name == "wheel");
+    let group_id = group.gid;
+    let group = Group::from_gid(group_id).unwrap().unwrap();
+    assert_eq!(group.gid, group_id);
+    assert_eq!(group.name, "wheel");
+}


changelog/2311.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed `::unistd::Group::members` crash on misaligned read by read_unaligned use
```
asomers, line 1: This sentence is hard to understand.  Could you please try rewording it?

asomers (changes requested):
This PR only contains changes to the doc test.  Where is the actual code change?

changelog/2311.fixed.md
anonkey: Done

anonkey:
> This PR only contains changes to the doc test. Where is the actual code change?

Wanted to check if misalignment was triggered in CI but wasn't the case so I factorize BSD and Apple test

src/unistd.rs
```diff
@@ -3569,13 +3568,29 @@ impl Group {
     ///
     /// # Examples
     ///
+    // Disable this test on all OS except BSD as wheel group may not exist.
+    #[cfg_attr(not(any(apple_targets, bsd)), doc = " ```no_run")]
```
SteveLauC, line 3572: `bsd` actually contains `apple_targets`, see [this](https://github.com/nix-rust/nix/blob/0dfcf3228489a5060a53e69783a1eea6f43b11a9/build.rs#L19).

SteveLauC (changes requested):

SteveLauC:
> Wanted to check if misalignment was triggered in CI but wasn't the case so I factorize BSD and Apple test


Did you successfully reproduce it in CI? Currently, there is no macOS aarch64 CI in Nix, but since GitHub Action has introduced [this](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/), we can enable it if needed.

> I am not sure if macOS x86_64 would allow that misaligned read so...

src/unistd.rs
```diff
@@ -3569,13 +3568,29 @@ impl Group {
     ///
     /// # Examples
```
SteveLauC, line 3569: BTW, I think we should move this new test to `tests/test_unistd.rs` since we don't want 2 duplicate examples

src/unistd.rs
```diff
@@ -3569,13 +3568,29 @@ impl Group {
...
     /// use nix::unistd::Group;
     /// // Returns an Result<Option<Group>>, thus the double unwrap.
-    /// let res = Group::from_name("root").unwrap().unwrap();
-    /// assert!(res.name == "root");
+    /// let group = Group::from_name("root").unwrap().unwrap();
+    /// assert!(group.name == "root");
+    /// let group_id = group.gid;
+    /// let group = Group::from_gid(group_id).unwrap().unwrap();
+    /// assert_eq!(group.gid, group_id);
+    /// assert_eq!(group.name, "root");
```
SteveLauC, line 3593: And this change should be reverted as this example is only for the API `Group::from_name`, `from_gid` will be covered by its own example.

anonkey:
> > Wanted to check if misalignment was triggered in CI but wasn't the case so I factorize BSD and Apple test
> 
> Did you successfully reproduce it in CI? Currently, there is no macOS aarch64 CI in Nix, but since GitHub Action has introduced [this](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/), we can enable it if needed.
> 
> > I am not sure if macOS x86_64 would allow that misaligned read so...

No i didn't, may because of that, but maybe because memory alignment depends of the OS and lot other random variables.
I can test with aarch64 if you think it's relevant

src/unistd.rs
anonkey: done

src/unistd.rs
anonkey: done

src/unistd.rs
anonkey: done

.github/workflows/ci.yml
```diff
@@ -44,6 +44,32 @@ jobs:
       - name: before_cache_script
         run: rm -rf $CARGO_HOME/registry/index
 
+  macos-silicon:
+    runs-on: macos-13-aarch64
```
SteveLauC, line 48:
```suggestion
    runs-on: macos-14
```

The new runner is `macos-14`, this is my fault as I have linked the wrong blog [in my previous comment](https://github.com/nix-rust/nix/pull/2311#issuecomment-1935203764), I am sorry, [this post](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) should be the right one.

.github/workflows/ci.yml
```diff
@@ -44,6 +44,32 @@ jobs:
       - name: before_cache_script
         run: rm -rf $CARGO_HOME/registry/index
 
+  macos-silicon:
```
SteveLauC, line 47: I suggest renaming this new job to `macos-aarch64`, and since we want to keep both these 2 CIs, we should rename the old `macos` job to `macos-x86_64`.

```suggestion
  macos-aarch64:
```

test/test_unistd.rs
```diff
@@ -1372,3 +1372,25 @@ fn test_eaccess_file_exists() {
...
+#[test]
+#[cfg(target_os = "linux")]
+fn test_group_from() {
+    let group = Group::from_name("root").unwrap().unwrap();
+    assert!(group.name == "root");
+    let group_id = group.gid;
+    let group = Group::from_gid(group_id).unwrap().unwrap();
+    assert_eq!(group.gid, group_id);
+    assert_eq!(group.name, "root");
+}
```
SteveLauC, line 1396: Nit: this test is probably not needed as we already have it covered in our doc tests (examples)

SteveLauC (changes requested):

.github/workflows/ci.yml
anonkey: done

test/test_unistd.rs
anonkey: done

anonkey:
don't trigger in CI, i tried 13 groups but none trigger while lot of them crash on my Mac, maybe because of the virtualization of macos?

.github/workflows/ci.yml
SteveLauC: > we should rename the old macos job to macos-x86_64.

It would be better if we rename the old `macos` job to `macos-x86_64` for symmetry reasons.

SteveLauC:
> don't trigger in CI, i tried 13 groups but none trigger while lot of them crash on my Mac, maybe because of the virtualization of macos?

That's possible. BTW, does this patch fix the issue on your host?

anonkey:
> > don't trigger in CI, i tried 13 groups but none trigger while lot of them crash on my Mac, maybe because of the virtualization of macos?
> 
> That's possible. BTW, does this patch fix the issue on your host?

Definitely, yes

SteveLauC (approved):
Thanks!

SteveLauC:
> Can you say when you are able to release a new version including this fix? We are having users who get this error and would like to use the fix in the dependencies. Thanks for fixing!

Hi, I think I will release 0.28.0 in the near future (possibly within this week)

PR #2314, opened 2024-02-14
System V shared memory APIs (2nd draft)

IzawGithub:
## Checklist:

I've seen issue #1718 and the pull request #1989 made by @arpankapoor from a year ago, that added some support for the System V API.

This is a continuation of this PR, where I tried to implement most of the change suggested by @VorpalBlade.

## What's missing

I haven't added any unit test yet, as somehow I'm getting a lot of compilation error on the test, about some missing symbol, but I've got a closed source project where I'm using this draft as a "eat your own dog food", with unit test, and it seems to work fine.

Some of the bitflags, 2 in `ShmgetFlag` and a lot in `SemctlCmd` do not exist in libc, even though they are documented.
So to merge this, there is a need to create a PR for libc that add those.

## What can be better

`Permissions`: Could maybe be moved elsewhere? I haven't found any struct already implemented that wrap the octal permission system unix use, so I threw this in 5m.

`shmat`: I'm **really** not a fan of returning a void ptr. I've tried to wrap it in a `Weak<RefCell<T>>` but I get a segfault when it gets dropped.
What would be ideal is for the user to give a generic type T that could be wrapped in a non owning smart pointer that could also be null. That's why I though Weak would be the answer, but I'm not good enough to find the problem.
Currently, the user has the cast it in unsafe themselves to be able to use the value in it, which is awkward and error prone.

I'm not too sure about the linux only bitflags compilation. We should test it for android too, but I don't know how to do that.

### DIFF
--- a/changelog/2313.added.md
+++ b/changelog/2313.added.md
@@ -1,0 +1,1 @@
+Add SystemV IPC support
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -634,14 +634,85 @@ pub fn shm_unlink<P: ?Sized + NixPath>(name: &P) -> Result<()> {
     Errno::result(ret).map(drop)
 }
 
+#[derive(Debug, Default, Clone, Copy)]
+/// Type used to transform a raw number to an octal permission, while performing a clamp to u9
+///
+/// # Example
+///
+/// ```
+/// # use nix::errno::Errno;
+/// # use nix::sys::mman::Permissions;
+///
+/// # fn main() -> Result<(), Errno> {
+/// assert_eq!(Permissions::new(511)?.get_permission(), &(0o0777 as u16));
+/// assert_eq!(Permissions::new(512).expect_err("512 is bigger than what u9 can store"), Errno::E2BIG);
+/// # Ok(())
+/// # }
+/// ```
+pub struct Permissions {
+    permission: u16,
+}
+
+impl Permissions {
+    /// Create a new Permissions object
+    ///
+    /// Clamp to a u9 size, return Errno::E2BIG if it fails
+    ///
+    pub fn new(octal: u16) -> Result<Self> {
+        if octal >= 2_u16.pow(9) {
+            return Err(Errno::E2BIG);
+        }
+        Ok(Permissions { permission: octal })
+    }
+
+    pub fn get_permission(&self) -> &u16 {
+        &self.permission
+    }
+
+    /// Using the current stored permission, do a bitor operation on the
+    /// bitflags enums given
+    ///
+    pub fn to_octal<T: bitflags::Flags<Bits = i32>>(
+        &self,
+        vec_flags: Vec<T>,
+    ) -> c_int {
+        let mut flags: c_int = T::empty().bits();
+        for flag in vec_flags {
+            flags |= flag.bits();
+        }
+        flags |= self.permission as i32;
+        flags
+    }
+}
 
 libc_bitflags! {
+    /// Different flags for the command `shmget`
     pub struct ShmgetFlag: c_int
     {
+        /// A new shared memory segment is created if key has this value
         IPC_PRIVATE;
+        /// Create a new segment. If this flag is not used, then shmget() will
+        /// find the segment associated with key and check to see if the user
+        /// has permission to access the segment.
         IPC_CREAT;
+        /// This flag is used with IPC_CREAT to ensure that this call creates
+        /// the segment.  If the segment already exists, the call fails.
         IPC_EXCL;
+        /// Allocate the segment using "huge" pages.  See the Linux kernel
+        /// source file Documentation/admin-guide/mm/hugetlbpage.rst for
+        /// further information.
+        #[cfg(any(target_os = "linux"))]
         SHM_HUGETLB;
+        // Does not exist in libc but should
+        // SHM_HUGE_2MB;
+        // SHM_HUGE_1GB;
+        /// This flag serves the same purpose as the mmap(2) MAP_NORESERVE flag.
+        /// Do not reserve swap space for this segment. When swap space is
+        /// reserved, one has the guarantee that it is possible to modify the
+        /// segment. When swap space is not reserved one might get SIGSEGV upon
+        /// a write if no physical memory is available. See also the discussion
+        /// of the file /proc/sys/vm/overcommit_memory in proc(5).
+        #[cfg(any(target_os = "linux"))]
         SHM_NORESERVE;
     }
 }
@@ -655,11 +726,9 @@ pub fn shmget(
     key: key_t,
     size: size_t,
     shmflg: Vec<ShmgetFlag>,
+    permission: Permissions,
 ) -> Result<i32> {
-    let mut flags: c_int = ShmgetFlag::empty().bits();
-    for flag in shmflg {
-        flags |= flag.bits;
-    }
+    let flags = permission.to_octal(shmflg);
     Errno::result(unsafe { libc::shmget(key, size, flags) })
 }
 
@@ -688,11 +757,9 @@ pub fn shmat(
     shmid: c_int,
     shmaddr: *const c_void,
     shmflg: Vec<ShmatFlag>,
+    permission: Permissions,
 ) -> Result<*mut c_void> {
-    let mut flags: c_int = ShmatFlag::empty().bits();
-    for flag in shmflg {
-        flags |= flag.bits;
-    }
+    let flags = permission.to_octal(shmflg);
     Errno::result(unsafe { libc::shmat(shmid, shmaddr, flags) })
 }
 
@@ -707,7 +774,7 @@ pub fn shmat(
 ///
 /// [`shmdt(2)`]: https://man7.org/linux/man-pages/man2/shmdt.2.html
 pub fn shmdt(shmaddr: *const c_void) -> Result<()> {
-    Errno::result(unsafe {libc::shmdt(shmaddr)}).map(drop)
+    Errno::result(unsafe { libc::shmdt(shmaddr) }).map(drop)
 }
 
 libc_bitflags! {
@@ -742,8 +809,11 @@ libc_bitflags! {
 /// [`shmctl(2)`]: https://man7.org/linux/man-pages/man2/shmctl.2.html
 pub fn shmctl(
     shmid: c_int,
-    cmd: c_int,
+    cmd: ShmctlFlag,
     buf: *mut shmid_ds,
+    permission: Permissions,
 ) -> Result<c_int> {
-    Errno::result(unsafe {libc::shmctl(shmid, cmd, buf)} )
+    let command = permission.to_octal(vec![cmd]);
+    Errno::result(unsafe { libc::shmctl(shmid, command, buf) })
 }
+
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -633,3 +633,117 @@ pub fn shm_unlink<P: ?Sized + NixPath>(name: &P) -> Result<()> {
 
     Errno::result(ret).map(drop)
 }
+
+
+libc_bitflags! {
+    pub struct ShmgetFlag: c_int
+    {
+        IPC_PRIVATE;
+        IPC_CREAT;
+        IPC_EXCL;
+        SHM_HUGETLB;
+        SHM_NORESERVE;
+    }
+}
+/// Creates and returns a new, or returns an existing, System V shared memory
+/// segment identifier.
+///
+/// For more information, see [`shmget(2)`].
+///
+/// [`shmget(2)`]: https://man7.org/linux/man-pages/man2/shmget.2.html
+pub fn shmget(
+    key: key_t,
+    size: size_t,
+    shmflg: Vec<ShmgetFlag>,
+) -> Result<i32> {
+    let mut flags: c_int = ShmgetFlag::empty().bits();
+    for flag in shmflg {
+        flags |= flag.bits;
+    }
+    Errno::result(unsafe { libc::shmget(key, size, flags) })
+}
+
+libc_bitflags! {
+    pub struct ShmatFlag: c_int
+    {
+        SHM_EXEC;
+        SHM_RND;
+        SHM_RDONLY;
+        #[cfg(any(target_os = "linux"))]
+        SHM_REMAP;
+    }
+}
+/// Attaches the System V shared memory segment identified by `shmid` to the
+/// address space of the calling process.
+///
+/// For more information, see [`shmat(2)`].
+///
+/// # Safety
+///
+/// `shmid` should be a valid shared memory identifier and
+/// `shmaddr` must meet the requirements described in the [`shmat(2)`] man page.
+///
+/// [`shmat(2)`]: https://man7.org/linux/man-pages/man2/shmat.2.html
+pub fn shmat(
+    shmid: c_int,
+    shmaddr: *const c_void,
+    shmflg: Vec<ShmatFlag>,
+) -> Result<*mut c_void> {
+    let mut flags: c_int = ShmatFlag::empty().bits();
+    for flag in shmflg {
+        flags |= flag.bits;
+    }
+    Errno::result(unsafe { libc::shmat(shmid, shmaddr, flags) })
+}
+
+/// Performs the reverse of [`shmat`], detaching the shared memory segment at
+/// the given address from the address space of the calling process.
+///
+/// For more information, see [`shmdt(2)`].
+///
+/// # Safety
+///
+/// `shmaddr` must meet the requirements described in the [`shmdt(2)`] man page.
+///
+/// [`shmdt(2)`]: https://man7.org/linux/man-pages/man2/shmdt.2.html
+pub fn shmdt(shmaddr: *const c_void) -> Result<()> {
+    Errno::result(unsafe {libc::shmdt(shmaddr)}).map(drop)
+}
+
+libc_bitflags! {
+    pub struct ShmctlFlag: c_int {
+        #[cfg(any(target_os = "linux"))]
+        IPC_INFO;
+        IPC_SET;
+        IPC_STAT;
+        IPC_RMID;
+        // not available in libc but should be?
+        // #[cfg(any(target_os = "linux"))]
+        // SHM_INFO;
+        // #[cfg(any(target_os = "linux"))]
+        // SHM_STAT;
+        // #[cfg(any(target_os = "linux"))]
+        // SHM_STAT_ANY;
+        #[cfg(any(target_os = "linux"))]
+        SHM_LOCK;
+        #[cfg(any(target_os = "linux"))]
+        SHM_UNLOCK;
+    }
+}
+/// Performs control operation specified by `cmd` on the System V shared
+/// memory segment given by `shmid`.
+///
+/// For more information, see [`shmctl(2)`].
+///
+/// # Safety
+///
+/// All arguments should be valid and meet the requirements described in the [`shmctl(2)`] man page.
+///
+/// [`shmctl(2)`]: https://man7.org/linux/man-pages/man2/shmctl.2.html
+pub fn shmctl(
+    shmid: c_int,
+    cmd: c_int,
+    buf: *mut shmid_ds,
+) -> Result<c_int> {
+    Errno::result(unsafe {libc::shmctl(shmid, cmd, buf)} )
+}
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -7,7 +7,10 @@ use crate::Result;
 #[cfg(not(target_os = "android"))]
 #[cfg(feature = "fs")]
 use crate::{fcntl::OFlag, sys::stat::Mode};
-use libc::{self, c_int, c_void, off_t, size_t};
+use libc::{
+    self, c_int, c_short, c_void, key_t, off_t, semid_ds, seminfo, shmid_ds,
+    size_t,
+};
 use std::ptr::NonNull;
 use std::{
     num::NonZeroUsize,
@@ -733,6 +736,30 @@ pub fn shmget(
 }
 
 libc_bitflags! {
+    pub struct SemgetFlag: c_int
+    {
+        IPC_PRIVATE;
+        IPC_CREAT;
+        IPC_EXCL;
+    }
+}
+/// Creates and return a new, or returns an existing, System V shared memory
+/// semaphore identifier.
+///
+/// For more information, see [`semget(2)`].
+///
+/// [`semget(2)`]: https://man7.org/linux/man-pages/man2/semget.2.html
+pub fn semget(
+    key: key_t,
+    size: i32,
+    semflg: Vec<SemgetFlag>,
+    permission: Permissions,
+) -> Result<i32> {
+    let flags = permission.to_octal(semflg);
+    Errno::result(unsafe { libc::semget(key, size, flags) })
+}
+
+libc_bitflags! {
     pub struct ShmatFlag: c_int
     {
         SHM_EXEC;
@@ -784,7 +811,7 @@ libc_bitflags! {
         IPC_SET;
         IPC_STAT;
         IPC_RMID;
-        // not available in libc but should be?
+        // not available in libc/linux, but should be?
         // #[cfg(any(target_os = "linux"))]
         // SHM_INFO;
         // #[cfg(any(target_os = "linux"))]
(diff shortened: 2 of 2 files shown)


SteveLauC:
I am trying to add support for System V message queue in #2318

IzawGithub:
Hello @SteveLauC,

I think this PR is pretty good now. I've added some basic test, with around 30% coverage of the new function.
What's missing is testing `shmdt`, which I don't really know how I could do that, and all of the different bitflags, which is a bit overwhelming.

I also didn't implement the `Semop` function because I have no clue at what it's even supposed to do. If anyone got an idea about how to implement it, it would be more than welcome.

I don't understand why BSD isn't able to build successfully, when it should be working fine.
<br>

Is there anything left to be done before a review and an eventual merge?

SteveLauC:
Hi @IzawGithub, sorry for the late reply, I plan to take a look at this PR this weekend :)

src/sys/system_v.rs
```diff
@@ -0,0 +1,385 @@
...
+/// Creates and returns a new, or returns an existing, System V shared memory
+/// segment identifier.
+///
+/// For more information, see [`shmget(2)`].
+///
+/// [`shmget(2)`]: https://man7.org/linux/man-pages/man2/shmget.2.html
+pub fn shmget(
+    key: key_t,
+    size: size_t,
+    shmflg: Vec<ShmgetFlag>,
```
SteveLauC, line 105: Do we need to use a `Vec` here, considering that `ShmgetFlag` is a bitflag so that it is ORable, for example, you can do:

```rs
shmget(key, size, ShmgetFlag::IPC_CREAT | ShmgetFlag::IPC_EXCL, permission)

File: src/sys/mod.rs [hunk 2]
```diff
@@ -143,6 +143,10 @@ feature! {
 #[allow(missing_docs)]
 pub mod sysinfo;
 
+#[allow(missing_docs)]
```
SteveLauC, line 146: I would appreciate it if these new APIs have nice documents or even better, examples:)

src/sys/mod.rs
```diff
@@ -143,6 +143,10 @@ feature! {
...
+#[cfg(any(bsd, target_os = "linux", not(target_os = "android")))]
```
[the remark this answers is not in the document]
SteveLauC: And it is always ok to not support all the platforms, we can add them incrementally

SteveLauC (changes requested):

changelog/2313.added.md
```diff
@@ -0,0 +1 @@
+Add SystemV IPC support
```
SteveLauC, line 1: It seems that this PR only covers:

1. semaphore set
2. shared memory segments

so we should probably annotate this

src/sys/system_v.rs
```diff
@@ -0,0 +1,385 @@
...
+/// # use nix::errno::Errno;
+/// # use nix::sys::system_v::Permissions;
+///
+/// # fn main() -> Result<(), Errno> {
+/// assert_eq!(Permissions::new(511)?.get_permission(), &(0o0777 as u16));
+/// assert_eq!(Permissions::new(512).expect_err("512 is bigger than what u9 can store"), Errno::E2BIG);
+/// # Ok(())
+/// # }
+/// ```
+pub struct Permissions {
```
SteveLauC, line 25: For the last 9 bits of those `flg` arguments, Nix already has a type for it,
   `nix::sys::stat::Mode`, so a new type is probably not necessary.

SteveLauC (changes requested):

SteveLauC:
Some thoughts about these APIs:

1. Module names

   I kinda think we should separate `semaphore set` and `shared memory segments`
   into 2 modules, following the names of their C header files, i.e, `sys/sem.rs` for 
   `semaphore set` and `sys/shm.rs` for `shared memory segments`.

2. For those APIs, do you think it will be better if we do something like:

   Take `semaphore set` as an example:

   ```rs
   /// A System V semaphore set.
   pub struct SemaphoreSet(libc::c_int);

   pub enum SemCmd {
       IPC_STAT(&mut libc::semid_ds),
       IPC_SET(&libc::semid_ds),
   }

   impl SemaphoreSet {
       pub fn new(key: key_t, n_sem: usize, semflg: SemFlg) -> Result<Self>;
       pub fn op<O: AsRef<[Sembuf]>>(&self, ops: O) -> Result<()>;
       pub fn op_timeout<O: AsRef<[Sembuf]>>(&self, ops: O, timeout: TimeSepc) -> Result<()>;
       pub fn ctl(&self, sem_num: usize, cmd: SemCmd) -> Result<Option<libc::c_int>>;
   }
   ```

   And to remove a `semaphore set`, I am thinking about making our wrapper type 
   non-Copy, and do this:

   ```rs
   #[allow(missing_copy_implementations)]
   pub struct SemaphoreSet(libc::c_int);

   impl SemaphoreSet {
       pub fn remove(self) -> Result<(), (Errno, Self)> {
          let res = unsafe {
              libc::semctl(self.0, /*whatever*/, IPC_RMID)
          };

          match Errno::result(res) {
              Ok(_) => Ok(()),
              Err(errno) => Err((errno, self))
          }
       }
   }
   ```

IzawGithub:
Hello,

Thanks for the comprehensive review! Those are all fair issue, I'll work on them this week.

For your last question about wrapping all the functions, this is certainly doable, but I wasn't sure if it was prefered to match libc as closely as possible or not. I can definitely try to make something more Rusty. Worst case scenario, if it's not ok we could always revert to what's currently there.

Also, once the PR is done, should I clean the history with a rebase before a merge or not?

SteveLauC:
> Also, once the PR is done, should I clean the history with a rebase before a merge or not?

Nix uses squash merge, so it's fine to leave the commit history here:)

IzawGithub:
Hello @SteveLauC, sorry for not answering sooner, I was pretty sick.

I have refactored the shared memory segment portion of this PR, to fit more with Rust style of doing things. If you feel that this is the right direction to go, I'll work on making something similar for the semaphore portion.

This also resolve the issue of returning a raw pointer. Before, the end user had to use `unsafe` for deref. This is now fixed because `SharedMemory` acts like a smart pointer to the non owning data. This makes for a much nicer interface.

changelog/2313.added.md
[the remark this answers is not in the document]
IzawGithub: Fixed

src/sys/system_v.rs
IzawGithub: You're right, I forgot about that.
Fixed by changing to taking a `Flag` directly

src/sys/mod.rs
IzawGithub: Fixed, removed the allow lint and added some doctest, please tell me if anything isn't clear in it.

changelog/2313.added.md
IzawGithub: Kinda fixed, I'm not sure if I should put more detail in it?

src/sys/system_v.rs
IzawGithub: Fixed, wasn't aware of this type. This makes it depend on the `fs` module but I'm assuming it's fine?

src/sys/mod.rs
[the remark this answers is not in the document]
IzawGithub: Fixed the redundant `android` target. BSD is still not building and I don't really understand why. Something to do with macro? I'm not proficient with them yet.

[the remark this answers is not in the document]
SteveLauC: > BSD is still not building and I don't really understand why.

The OpenBSD and NetBSD CI failed because some missing constants in the [libc](https://github.com/rust-lang/libc) crate, which means we need to add them before merging this PR if we want to have `#[cfg(bsd)`s supported.

Let me take a closer look at this and probably give it a fix!

changelog/2313.added.md
SteveLauC: > I'm not sure if I should put more detail in it?

No need for details, a brief intro will be sufficient.

src/sys/system_v.rs
SteveLauC: >  This makes it depend on the fs module but I'm assuming it's fine?

Yeah, this is unfortunate, ideally, this `Mode` type should be available if `#[cfg(any(feature="fs", feature="system_v")]`, doing this will require some extra work, so I think it is ok to make `system_v` depend on `fs`.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+    ///     None,
+    ///     ShmatFlag::empty(),
+    ///     Mode::empty(),
+    /// )?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    pub fn new(
+        shmid: c_int,
+        shmaddr: Option<c_void>,
```
SteveLauC, line 108: It seems that one cannot create a variable of type `c_void`? (unless with `std::mem::transmute()`) Which means that the case `Some(a_value_of_type_c_void)` will never happen?

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+}
+
+impl<T> Drop for SharedMemory<T> {
+    fn drop(&mut self) {
+        Self::shmdt(self).expect("SharedMemory detach from SysV IPC");
+    }
+}
+
+impl<T> SharedMemory<T> {
+    /// Create a new SharedMemory object
```
SteveLauC, line 81: This should probably removed considering that it won't create a new segment but attach to an existing one.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+    ///     Mode::empty(),
+    /// )?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    pub fn new(
+        shmid: c_int,
+        shmaddr: Option<c_void>,
+        shmat_flag: ShmatFlag,
+        mode: Mode,
```
SteveLauC, line 110: Looks like that the `shmflg` argument of `shmat(2)` should not contain the permission info

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+    ///
+    /// let _ = shared_memory.shmctl(ShmctlFlag::IPC_RMID, None, Mode::empty())?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    /// [`shmctl(2)`]: https://man7.org/linux/man-pages/man2/shmctl.2.html
+    pub fn shmctl(
+        &self,
+        shmctl_flag: ShmctlFlag,
+        buf: Option<shmid_ds>,
```
SteveLauC, line 190: What about passing it by reference:

```suggestion
        buf: Option<&shmid_ds>,
```

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+    /// )?;
+    ///
+    /// let _ = shared_memory.shmctl(ShmctlFlag::IPC_RMID, None, Mode::empty())?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    /// [`shmctl(2)`]: https://man7.org/linux/man-pages/man2/shmctl.2.html
+    pub fn shmctl(
+        &self,
+        shmctl_flag: ShmctlFlag,
```
SteveLauC, line 189: This name can be confusing, I think we should stick with the C variable name, i.e., something like `cmd` or `shm_cmd`.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+        buf: Option<shmid_ds>,
+        mode: Mode,
```
SteveLauC, line 191: Same here, the `shmctl(2)` syscall does not seem to take a `mode` argument

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+    /// address space of the calling process.
+    ///
+    /// This is called automatically on [`SharedMemory::new`].
+    ///
+    /// For more information, see [`shmat(2)`].
+    ///
+    /// [`shmat(2)`]: https://man7.org/linux/man-pages/man2/shmat.2.html
+    fn shmat(
+        shmid: c_int,
+        shmaddr: Option<c_void>,
```
SteveLauC, line 213: Same here, see above

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,448 @@
...
+        shmat_flag: ShmatFlag,
+        mode: Mode,
```
SteveLauC, line 215: Same here, `mode` argument seems to be unneeded

SteveLauC (changes requested):

src/sys/system_v/shm.rs
IzawGithub: Huh, I didn't think about that. I don't really see a way to fix it. Most use I'm finding online of `shmat` is with shmaddr set to null.
So maybe not having that feature is ok? If I remove it, I'll add it to the doc that this isn't supported.

src/sys/system_v/shm.rs
IzawGithub: Fixed by removing the ctor and letting `Shm` wraps around it.

src/sys/system_v/shm.rs
IzawGithub: Oops, yeah, copy and paste error, my bad. Fixed

src/sys/system_v/shm.rs
IzawGithub: Yes, it's even transformed to a pointer after. Fixed

src/sys/system_v/shm.rs
IzawGithub: No problem, renamed to `shm_cmd`

src/sys/system_v/shm.rs
IzawGithub: Fixed.

src/sys/system_v/shm.rs
IzawGithub: Fixed.

src/sys/system_v/shm.rs
IzawGithub: Fixed.

IzawGithub:
Hello, all of the issue listed above should be fixed, expect  for the `c_void` where I don't really know how to go forward with.

Nice catch on the UB!
To resolve it, I had to make `shmget` an unsafe function. This is because, if the user is connecting to a shared memory zone that we haven't created ourselves, then there is no way to make sure the types match.
If we're always creating the new memory zone, then we're sure the type match, so I've added a safe `create_and_connect` for that.

I'll work on the semaphores during the coming week.

src/sys/system_v/shm.rs
SteveLauC: I hope we can have it if possible as there will always be someone who wants to use it, so if we are going to keep this argument, what is the proper type? 

1. `Option<&libc::c_void>`
  
     If `SHM_RND` is not specified in the flag argument, then the address must be page-aligned, which means in most cases, users need to manually specify a page-aligned address, and reference, is rarely page-aligned, so it is not that useful.

2. `Option<*const libc::c_void>`
    
    `Option` is rusty, but if the inner type is a raw pointer, why do we need it?

3. a simple raw pointer `*const libc::c_void`
   
   Simple, unrusty, but work.

Your thoughts on this?

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+
+impl<T> Drop for SharedMemory<T> {
+    fn drop(&mut self) {
+        Self::shmdt(self).expect("SharedMemory detach from SysV IPC");
+    }
+}
+
+impl<T> SharedMemory<T> {
+    /// Performs control operation specified by `cmd` on the System V shared
+    /// memory segment given by `shmid`.
```
SteveLauC, line 236: This `shmid` identifier is not a parameter but stored in `SharedMemory`

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+            libc::shmat(self.id, ptr::null(), shmat_flag.bits())
+        })
+        .map(|ok| ok.cast::<T>())
+    }
+}
+
+#[derive(Debug)]
+/// Safe wrapper around a SystemV shared memory segment data
+///
+/// This is a smart pointer, and so implement the [`Deref`] and [`DerefMut`] traits.
```
SteveLauC, line 181: Nit: seems to be a mistake:

```suggestion
/// This is a smart pointer, and so implements the [`Deref`] and [`DerefMut`] traits.
```

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+    ///     Mode::S_IRWXU | Mode::S_IRWXG | Mode::S_IRWXO,
+    /// )?;
+    /// let shared_memory = mem_segment.attach(ShmatFlag::empty())?;
+    ///
+    /// let _ = shared_memory.shmctl(ShmctlFlag::IPC_RMID, None)?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    /// [`shmctl(2)`]: https://man7.org/linux/man-pages/man2/shmctl.2.html
+    pub fn shmctl(
```
SteveLauC, line 263: Should we move this method to `impl Shm { }`, it is more like a method associated with `Shm`

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+            .map(drop)
+    }
+}
+
+libc_bitflags!(
+    /// Valid flags for the third parameter of the function [`shmget`]
+    pub struct ShmgetFlag: c_int
+    {
+        /// A new shared memory segment is created if key has this value.
+        IPC_PRIVATE;
```
SteveLauC, line 297: If this is a key rather than a value of flag, then we should probably remove it from `ShmgetFlag`?

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+    /// [`shmdt(2)`]: https://man7.org/linux/man-pages/man2/shmdt.2.html
+    fn shmdt(&self) -> Result<()> {
+        let shmaddr_ref: *const T = &**self;
+        Errno::result(unsafe { libc::shmdt(shmaddr_ref.cast::<c_void>()) })
+            .map(drop)
+    }
+}
+
+libc_bitflags!(
+    /// Valid flags for the third parameter of the function [`shmget`]
```
SteveLauC, line 293: This makes it render correctly:

```suggestion
    /// Valid flags for the third parameter of the function [`Shm::shmget`].
```

SteveLauC (changes requested):

src/sys/system_v/shm.rs
SteveLauC: Yeah, I think this should be done or users will have to attach to a shared memory segment in order to remove it, which is not needed in the C API

IzawGithub:
Hello, this is harder than I though!

For your example, after a bit of debugging, I think that pointing to 0x0 is because we're trying to store a type that's a ref, when SharedMemory actually expect a non ref.
`T` being a generic, I would like to restrict it to only owned data, which would resolve the issue.
However, I don't really see a way to do that. I didn't see any "not a pointer" trait, maybe `Default`? But that's annoying for the end user if their data struct does not implement default.

The rest of the issue are pretty simple fix I'll work on right away.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+    /// const MY_KEY: i32 = 1337;
+    ///
+    /// let mem_segment = Shm::<MyData>::create_and_connect(
+    ///     MY_KEY,
+    ///     Mode::S_IRWXU | Mode::S_IRWXG | Mode::S_IRWXO,
+    /// )?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    pub fn create_and_connect(key: key_t, mode: Mode) -> Result<Self> {
```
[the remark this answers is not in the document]
IzawGithub: Yeah I didn't really put too much though on the name. Maybe `create_and_get`? We're creating the segment by using shm*get*. I'm not the best at naming things, so if you've got a better idea, feel free!

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+
+impl<T> Shm<T> {
+    /// Attach to the current SystemV shared memory segment.
+    ///
+    /// To create a new segment, use [`Shm::create_and_connect`].\
+    /// If you need more customisation, use the unsafe version,
+    /// [`Shm::shmget`], with the key [`ShmgetFlag::IPC_CREAT`].
+    ///
+    /// Attaching a segment to a specific adress isn't supported. This is
+    /// because there is no way to create a void pointer on rust.
```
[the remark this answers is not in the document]
IzawGithub: I was worried of a potential deref of a null pointer, but I remembered that the libc shmat already handle that, so this is actually fine and not that bad to use as a end user. Added

src/sys/mod.rs
```diff
@@ -143,6 +143,12 @@ feature! {
 #[allow(missing_docs)]
 pub mod sysinfo;
 
+#[cfg(any(bsd, target_os = "linux"))]
+feature! {
+    #![feature = "system_v"]
+    pub mod system_v;
```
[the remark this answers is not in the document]
IzawGithub: Oh I didn't realise that this would all go in `sys`, I though you meant a sysv specific module, my bad. Fixed

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+        /// Allow the contents of the segment to be executed. The caller must
+        /// have execute permission on the segment.
+        #[cfg(linux)]
+        SHM_EXEC;
+        /// Attach the segment for read-only access. The process must have read
+        /// permission for the segment. If this flag is not specified, the
+        /// segment is attached for read and write access, and the process must
+        /// have read and write permission for the segment.
+        /// There is no notion of a write-only shared memory segment.
+        SHM_RDONLY;
```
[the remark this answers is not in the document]
IzawGithub: Yeah, it was there before I removed it with the shmat pointer adress var. Added back

src/sys/system_v/shm.rs
IzawGithub: Forgot to edit that, nice catch. Fixed.

src/sys/system_v/shm.rs
IzawGithub: Fixed

src/sys/system_v/shm.rs
IzawGithub: This is even better, that way SharedMemory really is just a smart pointer and has no associated function except the private detach. Fixed.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+        // SHM_HUGE_2MB;
+        // TODO: Same for this one
+        // SHM_HUGE_1GB;
+        /// This flag serves the same purpose as the mmap(2) MAP_NORESERVE flag.
+        /// Do not reserve swap space for this segment. When swap space is
+        /// reserved, one has the guarantee that it is possible to modify the
+        /// segment. When swap space is not reserved one might get SIGSEGV upon
+        /// a write if no physical memory is available. See also the discussion
+        /// of the file /proc/sys/vm/overcommit_memory in proc(5).
+        #[cfg(linux)]
```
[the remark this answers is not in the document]
IzawGithub: No problem, fixed.

src/sys/system_v/shm.rs
IzawGithub: Fixed

Cargo.toml
```diff
@@ -64,6 +64,7 @@ resource = []
 sched = ["process"]
 signal = ["process"]
 socket = ["memoffset"]
+system_v = ["fs"]
```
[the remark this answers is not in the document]
IzawGithub: Sure, fixed

src/sys/system_v/shm.rs
IzawGithub: After re-reading the doc, you're right. This isn't a bitflag but a value the user can pass to the `key` argument. Fixed.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+        SHM_LOCK;
+        /// Unlock the segment, allowing it to be swapped out.
+        #[cfg(linux)]
+        SHM_UNLOCK;
+    }
+);
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
IzawGithub: I used private field before, but not anymore, so this can indeed be an independent test. Fixed.

src/sys/system_v/shm.rs
```diff
@@ -0,0 +1,476 @@
...
+    ///     Mode::S_IRWXU | Mode::S_IRWXG | Mode::S_IRWXO,
+    /// )?;
+    /// # Ok::<(), Errno>(())
+    /// ```
+    ///
+    pub fn create_and_connect(key: key_t, mode: Mode) -> Result<Self> {
+        let size = std::mem::size_of::<T>();
+        // This is the main difference between this function and [`Shm::shmget`]
+        // Because we are always creating a new segment, we can be sure that the size match
+        let shmget_flag = ShmgetFlag::IPC_CREAT | ShmgetFlag::IPC_EXCL;
```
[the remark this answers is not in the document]
IzawGithub: I do not understand what you mean here.
Why would `.attach()` use `shmget_flag`, when it only connect to an existing memory segment?

This here is guaranteed safe because we're 100% sure that the new memory segment either has the size of `T`, or if it fails to be created, return an error. If something more fine grained is needed, isn't `unsafe .shmget()` the function that fulfill this purpose?

SteveLauC:
Hi, sorry for the late response, been busy with my work and can barely do open source things on work days.

-----

> For your example, after a bit of debugging, I think that pointing to 0x0 is because we're trying to store a type that's a ref, when SharedMemory actually expect a non ref. T being a generic, I would like to restrict it to only owned data, which would resolve the issue.

Even if we restrict `T` to be owned types, it is still not guaranteed that the value stored in that shared memory is valid for T, so the problem still exists.

From my observation, that memory segment will be zero-initialized, 0 is a valid type for `std::num::NonZeroU8`:

```rs
use nix::sys::{
    stat::Mode,
    system_v::shm::{Shm, ShmatFlag},
};

fn main() {
    let mode = Mode::S_IRWXU;

    let shm = Shm::<std::num::NonZeroU8>::create_and_connect(65536, mode).unwrap();
    let mem_segment = shm.attach(ShmatFlag::empty()).unwrap();
    println!("number = {}", *mem_segment);
}
```

```sh
$ cargo r                      
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/rust`
number = 0
```

PR #2315, opened 2024-02-16
refactor: redefine ForkptyResult to avoid uninited master field

SteveLauC:
## What does this PR do

Before this PR, our `ForkptrResult` was defined like this:

```rs
pub struct ForkptyResult {
    pub master: OwnedFd,
    pub fork_result: ForkResult,
}
```

The problem with this definition is that the `master` field won't be initialized (actually, it won't be accessed at all) within the forked child process, which means that our `forkpty()` was calling `OwnedFd::from_raw_fd(an uninitialized memory)`, if the dirty value stored in that uninitialized memory is not a valid value for a file descriptor (say -1), then this function will panic, making the child process panic.

This PR redefines the `ForkptyResult` type so that the `master` field is only present in the parent process.

cc @Xeom 

related to #2312

## Checklist:

### DIFF
--- a/changelog/2315.changed.md
+++ b/changelog/2315.changed.md
@@ -1,0 +1,12 @@
+Change the `ForkptyResult` type to the following repr so that the uninitialized 
+`master` field won't be accessed in the child process:
+
+```rs
+pub enum ForkptyResult {
+    Parent {
+        child: Pid,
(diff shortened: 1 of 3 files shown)


asomers (approved):
Yes, I think this is exactly right.

PR #2324, opened 2024-02-28
Change the type of data of ptrace::write to make it safe

### LINKED ISSUE #2323: Change the signature of `ptrace::write` to make it safe
For now, the whole impl of write is

``` rs
pub unsafe fn write(
    pid: Pid,
    addr: AddressType,
    data: *mut c_void,
) -> Result<()> {
    unsafe { ptrace_other(Request::PTRACE_POKEDATA, pid, addr, data).map(drop) }
}
```

In ptrace(2), it's mentioned data is "a word", which should be the same as the return type of read(`c_long`, in current code).

So, why don't we change the type of data to `c_long,` and make it safe?

Or maybe the type could be a byte array like `[u8; sizeof::<c_long>()]`

ShuiRuTian:
## What does this PR do
Fix #2323

Change the signature of `ptrace::write` and `ptrace::write_user` to make them safe

## Checklist:

### DIFF
--- a/changelog/2324.changed.md
+++ b/changelog/2324.changed.md
@@ -1,0 +1,1 @@
+Change the signature of `ptrace::write` and `ptrace::write_user` to make them safe
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -543,17 +543,12 @@ pub fn read(pid: Pid, addr: AddressType) -> Result<c_long> {
 
 /// Writes a word into the processes memory at the given address, as with
 /// ptrace(PTRACE_POKEDATA, ...)
-///
-/// # Safety
-///
-/// The `data` argument is passed directly to `ptrace(2)`.  Read that man page
-/// for guidance.
-pub unsafe fn write(
+pub fn write(
     pid: Pid,
     addr: AddressType,
-    data: *mut c_void,
+    data: c_long,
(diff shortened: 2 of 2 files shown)


ShuiRuTian:
I use the same command, but could not reproduce the issue, does anyone know the reason?

I ran `cargo clippy  --target x86_64-unknown-linux-gnu --all-targets --all-features -- -D warnings -A unknown-lints`

SteveLauC:
Well, it is indeed weird, I didn't reproduce it either...

Will investigate more.

SteveLauC:
I kinda think this error is a false positive?

My guess on the reason why it is triggerred is that we are directly casting a `c_long` to a raw pointer, but I think thisx is fine as the kernel won't treat it as a raw pointer but an [`unsigned long`](https://github.com/torvalds/linux/blob/805d849d7c3cc1f38efefd48b2480d62b7b5dcb7/kernel/ptrace.c#L1310), so we can:

1. Allow it `#[allow(clippy::not_unsafe_ptr_arg_deref)]`
2. Write a comment stating why the usage of `ptrace_other()` within `write()` and `write_user()` is safe
   
    ```
    // Safety:
    ```

ShuiRuTian:
Thanks for the suggestions, it's past now!

SteveLauC (approved):
Thanks!

PR #2326, opened 2024-03-03
feat: make SigAction repr(transparent)&From<raw>&Into<raw>

### LINKED ISSUE #2319: Support custom signals for sigaction
`sigaction` currently only accepts the `Signal` enum and directly passes it to `libc::sigaction` as a c_int.

I'm working on a linux system that has custom signals superior to 32. And thus not declared in the Signal enum.

The idea would be to have another `sigaction` function that accepts directly a `c_int`.

Something like 
```rust
pub unsafe fn sigaction_raw_signal(signal_raw: libc::c_int sigaction: &SigAction) -> Result<SigAction> {
    let mut oldact = mem::MaybeUninit::<libc::sigaction>::uninit();

    let res = libc::sigaction(signal_raw,
                              &sigaction.sigaction as *const libc::sigaction,
                              oldact.as_mut_ptr());

    Errno::result(res).map(|_| SigAction { sigaction: oldact.assume_init() })
}
```

It might be a bit out of scope for this library ... let me know!

I looked into implementing the function myself, but since Sigaction's fields are not public, I cannot have this function inside my own library. So another way would be to make `sigaction` public, but I feel like it's not ideal.

SteveLauC:
## What does this PR do

Make SigAction repr(transparent) & can be converted to/from the libc raw type.

This is the first PR towards the goal described in this [comment](https://github.com/nix-rust/nix/issues/2319#issuecomment-1967988916), closes #2319.

I will handle the remaining ones in the future PRs.

refe: #2327

## Checklist:

### DIFF
--- a/changelog/2326.added.md
+++ b/changelog/2326.added.md
@@ -1,0 +1,1 @@
+make SigAction repr(transparent)&From<libc::sigaction>&Into<libc::sigaction>
--- a/changelog/2326.added.md
+++ b/changelog/2326.added.md
@@ -1,1 +1,1 @@
-make SigAction repr(transparent)&From<libc::sigaction>&Into<libc::sigaction>
+make SigAction repr(transparent) & can be converted to/from the libc raw type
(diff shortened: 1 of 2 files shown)


SteveLauC:
Let's merge this!

Issue #2327, opened 2024-03-03
Make all NIx libc wrapper types `repr(transparent)`/`From<libc::xx>`


SteveLauC:

SteveLauC:
My current thought on this is that `impl From<libc::xx> for Wrapper` can be `unsafe` because we cannot ensure the  values set in those libc types are valid

asomers:
> My current thought on this is that `impl From<libc::xx> for Wrapper` can be `unsafe` because we cannot ensure the values set in those libc types are valid

Yes.  Especially for cases where the libc type is a union, for example.

asomers:
> There are cases, where the data does not originate from C code, but from Rust code directly. Wouldn't it make sense to have a native Rust type that is guaranteed to be complete as alternative to the libc types? Something like:

That involves doing data copies during every transition from C to Rust or vice versa.  Nix strives for 0-cost abstractions instead.  Wherever possible, Nix's Rust accessors should manipulate the raw C struct.

jonatanzeidler:
My suggestion does not replace any existing approach, but adds an alternative way for initialization that happens once for a socket. I'd argue that this optional one-time copy is a valid trade-off for getting rid of `unsafe` code.

jonatanzeidler:
In order to get rid of the copy, we could also make `SockAddrLl` a newtype around `libc::sockaddr_ll` and let the user move all values into the constructor. Would you prefer that?

SteveLauC:
> In order to get rid of the copy, we could also make `SockAddrLl` a newtype around `libc::sockaddr_ll` and let the user move all values into the constructor (hinting to and hoping for in-lining). Would you prefer that?

The constructor way has been discussed in this [comment](https://github.com/nix-rust/nix/issues/2059#issuecomment-1627663963), and:

> I don't like the proposed constructors because they're potentially too limited. On Linux and OSX, the libc structure has 7 fields. On the BSDs, it has between 8 and 10. We don't want to define a new constructor for every combination of fields that the user might want to initialize.

SteveLauC:
Related: #1977

PR #2328, opened 2024-03-06
Improve ergonomics between SockaddrIn and underlying libc type

pronebird:
## Checklist:

- Add `Into<libc::sockaddr_in6> for SockaddrIn6`
- Add `Into<libc::sockaddr_in> for SockaddrIn`

My typical use case for these `Into` traits is when I am bridging from safe world into the world of pain (C). For example,  presently I have to go over `as_ref()` to access the underlying `sockaddr_inX` . This is a bit verbose for my liking, so I figured it would be great to have `into()` to do the job for me.  Usually `From` is preferred but I am not certain if we want that, hence I have implemented `Into` but can easily flip to `From` if that makes sense.

```diff
let addr_sa = SockaddrIn6::from(SocketAddrV6::new(addr, 0, 0, 0));
let mut req: in6_aliasreq = unsafe { mem::zeroed() };
-req.addr = *addr_sa.as_ref();
+req.addr = addr_sa.into();
```

I also noticed that `repr(transparent)` is set on these structs and naturally I thought I could use `addr_sa as sockaddr_in6` but that doesn't seem to compile.

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -920,6 +920,19 @@ impl From<SockaddrIn> for net::SocketAddrV4 {
 }
 
 #[cfg(feature = "net")]
+impl From<SockaddrIn> for libc::sockaddr_in {
+    fn from(sin: SockaddrIn) -> libc::sockaddr_in {
+        sin.0
+    }
+}
+#[cfg(feature = "net")]
+impl From<libc::sockaddr_in> for SockaddrIn {
+    fn from(sin: libc::sockaddr_in) -> SockaddrIn {
+        SockaddrIn(sin)
+    }
+}
+
+#[cfg(feature = "net")]
 impl std::str::FromStr for SockaddrIn {
     type Err = net::AddrParseError;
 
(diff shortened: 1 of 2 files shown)


src/sys/socket/addr.rs
```diff
@@ -920,6 +920,15 @@ impl From<SockaddrIn> for net::SocketAddrV4 {
 }
 
 #[cfg(feature = "net")]
+#[allow(clippy::from_over_into)]
```
SteveLauC, line 923: Would you like to do:

```
impl From<libc::sockaddr_in> for SockaddrIn
impl From<SockaddrIn> for libc::sockaddrin

impl From<libc::sockaddr_in6> for SockaddrIn6
impl From<SockaddrIn6> for libc::sockaddrin6
```

So that we can do roundtrips and also make clippy happy:)

SteveLauC:
This is indeed something we should do to allow Nix to be used in such scenarios, see #2327, and thanks for your interest in contributing to Nix!

src/sys/socket/addr.rs
pronebird: Done.

pronebird:
CI failure in `test_sigsuspend` is probably unrelated.

SteveLauC (approved):
Thanks! That CI failure is not related:)

PR #2333, opened 2024-03-10
sys::resource update max_rss accessor doc to reflect the unit per platform

### LINKED ISSUE #2331: `ru_maxrss` in the result of `getrusage` is in bytes on macOS
According to the manpage of `getrusage` on macOS, `ru_maxrss` is in bytes, which contradicts the statement in nix's doc.

Quote from `man getrusage` on macOS:

> ru_maxrss    the maximum resident set size utilized (in bytes).

nix's doc on `Usage`:

> The resident set size at its peak, in kilobytes.

devnexen:
on Apple's devices. this field returns the value in byte. close #2331

### DIFF
--- a/src/sys/resource.rs
+++ b/src/sys/resource.rs
@@ -293,7 +293,9 @@ impl Usage {
         TimeVal::from(self.0.ru_stime)
     }
 
-    /// The resident set size at its peak, in kilobytes.
+    /// The resident set size at its peak,
+    #[cfg_attr(apple_targets, doc = " in bytes.")]
+    #[cfg_attr(not(apple_targets), doc = " in kilobytes.")]
     pub fn max_rss(&self) -> c_long {
         self.0.ru_maxrss
     }


src/sys/resource.rs
```diff
@@ -295,7 +295,10 @@ impl Usage {
 
     /// The resident set size at its peak, in kilobytes.
     pub fn max_rss(&self) -> c_long {
-        self.0.ru_maxrss
+        #[cfg(not(apple_targets))]
+        return self.0.ru_maxrss;
+        #[cfg(apple_targets)]
+        return self.0.ru_maxrss >> 10;
```
SteveLauC, line 301: What about updating the doc on `apple_targets` rather than tweaking the code so that we can be more resilient to this kind of [changes](https://github.com/nix-rust/nix/issues/2331#issuecomment-2001858705):

```rs
    /// The resident set size at its peak,
    #[cfg_attr(apple_targets, doc = " in bytes.")]
    #[cfg_attr(not(apple_targets), doc = "in kilobytes.")]
    pub fn max_rss(&self) -> c_long {
        self.0.ru_maxrss
```

src/sys/resource.rs
```diff
@@ -293,7 +293,9 @@ impl Usage {
         TimeVal::from(self.0.ru_stime)
     }
 
-    /// The resident set size at its peak, in kilobytes.
+    /// /// The resident set size at its peak,
```
SteveLauC, line 296: Seems to be redundant:

```suggestion
    /// The resident set size at its peak,
```

src/sys/resource.rs
```diff
@@ -293,7 +293,9 @@ impl Usage {
...
+    #[cfg_attr(apple_targets, doc = " in bytes.")]
+    #[cfg_attr(not(apple_targets), doc = "in kilobytes.")]
```
SteveLauC, line 298: Nix: might be better to have a space here:

```suggestion
    #[cfg_attr(not(apple_targets), doc = " in kilobytes.")]
```

SteveLauC (approved):
Thanks! Are you willing to update the PR title since Nix's merge queue uses squash merge, the PR title will be used as the commit message

PR #2335, opened 2024-03-16
ci: disable the DragonFlyBSD CI

SteveLauC:
## What does this PR do

Disable the tier3 DragonFlyBSD CI because the latest nightly rust toolchain is broken on it, see the upstream issue: https://github.com/rust-lang/rust/issues/122585

Nix issue: https://github.com/nix-rust/nix/issues/2337

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,7 +292,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: x86_64-unknown-dragonfly
+          # Temporarily disable DragonFlyBSD
+          # https://github.com/nix-rust/nix/issues/2337
+          # - target: x86_64-unknown-dragonfly
           - target: x86_64-unknown-openbsd
           # Temporarily disable armv7-unknown-linux-uclibceabihf
           # https://github.com/nix-rust/nix/issues/2200


SteveLauC:
Can reproduce this in both CI and my local Linux machine with nightly `cargo 1.78.0-nightly (7065f0ef4 2024-03-12)`.

To reproduce:

```sh
$ cargo new foo
$ cd foo
$ cargo +nightly --version
cargo 1.78.0-nightly (7065f0ef4 2024-03-12)

$ cargo +nightly check -Zbuild-std --target x86_64-unknown-dragonfly
   Compiling std v0.0.0 (/home/steve/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std)
error[E0425]: cannot find function `getrandom` in crate `libc`
  --> /home/steve/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/pal/unix/rand.rs:73:24
   |
73 |         unsafe { libc::getrandom(buf.as_mut_ptr().cast(), buf.len(), 0) }
   |                        ^^^^^^^^^ not found in `libc`

For more information about this error, try `rustc --explain E0425`.
error: could not compile `std` (lib) due to 1 previous error
```

And that function is indeed unavailable on DragonFlyBSD with libc 0.2.153.

PR #2339, opened 2024-03-20
Add support for openat2.

### LINKED ISSUE #1400: Support for openat2()
On Linux, the recent ``openat2()`` system call provides a broader API surface than ``openat()``. However, it's not yet wrapped in GLIBC, so it would be necessary to make the system call directly. Would this be a problem in terms of maintainability and portability, or is this something that the Nix crate is already used to dealing with?

blinsay:
## What does this PR do

Adds support for openat2 on linux. Includes a new `ResolveFlag` struct to pass resolve flags, which is passed directly to the new `fcntl::openat2` function. `libc::open_how` isn't exposed in any way here, which will mean this API needs to change if there's an update that extends `libc::open_how` with new fields.

Given that the whole point of `libc::open_how` is that it's extensible, I'm not sure that this is the right way to go about adding this API. I could be convinced to go back and add something that looks more like the stdlib's `OpenOptions` that would allow adding more arguments to this function without a breaking change. I would love some input/direction here.

Fixes #1400.

## Checklist:

### DIFF
--- a/changelog/2339.added.md
+++ b/changelog/2339.added.md
@@ -1,0 +1,1 @@
+Added support for openat2 on linux.
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -242,6 +242,119 @@ pub fn openat<P: ?Sized + NixPath>(
     Errno::result(fd)
 }
 
+cfg_if::cfg_if! {
+    if #[cfg(target_os = "linux")] {
+        libc_bitflags! {
+            /// Path resolution flags.
+            ///
+            /// See [path resolution(7)](https://man7.org/linux/man-pages/man7/path_resolution.7.html)
+            /// for details of the resolution process.
+            pub struct ResolveFlag: libc::c_ulonglong {
+                /// Do not permit the path resolution to succeed if any component of
+                /// the resolution is not a descendant of the directory indicated by
+                /// dirfd.  This causes absolute symbolic links (and absolute values of
+                /// pathname) to be rejected.
+                RESOLVE_BENEATH;
+
+                /// Treat the directory referred to by dirfd as the root directory
+                /// while resolving pathname.
+                RESOLVE_IN_ROOT;
+
+                /// Disallow all magic-link resolution during path resolution. Magic
+                /// links are symbolic link-like objects that are most notably found
+                /// in proc(5);  examples include `/proc/[pid]/exe` and `/proc/[pid]/fd/*`.
+                ///
+                /// See symlink(7) for more details.
+                RESOLVE_NO_MAGICLINKS;
+
+                /// Disallow resolution of symbolic links during path resolution. This
+                /// option implies RESOLVE_NO_MAGICLINKS.
+                RESOLVE_NO_SYMLINKS;
+
+                /// Disallow traversal of mount points during path resolution (including
+                /// all bind mounts).
+                RESOLVE_NO_XDEV;
+            }
+        }
+
+        /// Specifies how [openat2] should open a pathname.
+        ///
+        /// See <https://man7.org/linux/man-pages/man2/open_how.2type.html>
+        #[repr(transparent)]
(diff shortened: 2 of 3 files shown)


blinsay:
Also, I haven't added the changelog yet because I wasn't sure what my PR number would be. I'll go ahead and add that once I get some input on what the API should look like.

SteveLauC:
Hi, thanks for this, I will take a look at this PR this weekend:)

SteveLauC:
Hi, thanks for the PR and thoughts on the API design:)

>  which will mean this API needs to change if there's an update that extends libc::open_how with new fields.

Yeah, this is bad and we should avoid it.

> I could be convinced to go back and add something that looks more like the stdlib's `OpenOptions` that would allow adding more arguments to this function without a breaking change.

Yeah, a builder pattern for `OpenHow` would work well:

1. Nix won't introduce breaking changes when the Linux kernel extends this structure, we can simply add new `.with_new_field(field_value)` methods.
2. The [manual](https://man7.org/linux/man-pages/man2/open_how.2type.html) mentions that:
  
     > Therefore, a user **must** zero-fill this structure on initialization. 

    We can enforce this with a builder pattern.


-----

A draft code would be something like this:

```rs
#[repr(transparent)]
pub struct OpenHow(libc::open_how);

pub struct OpenHowBuilder {
    open_how: libc::open_how,
}

impl OpenHowBuilder {
    pub fn new() -> Self {
        // SAFETY:
        // From the manual: https://man7.org/linux/man-pages/man2/open_how.2type.html
        //
        // > Therefore, a user must zero-fill this structure on initialization.
        //
        // Initializing `open_how` with 0s is valid and a necessity.
        let zero_inited = unsafe {
            std::mem::MaybeUninit::zeroed().assume_init()
        };
        Self {
            open_how: zero_inited
        }
    }

    pub fn with_flags(&mut self, flags: nix::fcntl::OFlag) -> &mut Self {
        let flags = flags.bits();
        self.open_how.flags = flags.try_into().expect("failed to cast it to u64");

        self
    }

    pub fn with_mode(&mut self, mode: nix::sys::stat::Mode) -> &mut Self {
        let mode = mode.bits();
        self.open_how.mode = mode.try_into().expect("failed to cast it to u64");

        self
    }

    pub fn with_resolve(&mut self, resolve: nix::fcntl::ResolveFlag) -> &mut Self {
        let resolve = resolve.bits();
        self.open_how.resolve = resolve;

        self
    }
    
    // We can add new methods for future kernel extensions

    pub fn build(self) -> OpenHow {
        OpenHow(self.open_how)
    }
}
```

blinsay:
> Therefore, a user must zero-fill this structure on initialization.

Whoops, that's a big oversight on my part! Thanks for catching it.

I went with a`#[repr(transparent)]` struct with builder-style method, the extra layer of builder didn't feel worth it.

I think the slightly more interesting choice is having `openat2` taking an owned `OpenHow` instead of a `&mut OpenHow`. It felt right given that the `#![deny(missing_copy_implementations)]` lint suggested I make `OpenHow` be `Copy`, but we may want to just make it `Clone` in case any of the future fields aren't easily `Copy`-able.

blinsay:
Sorry for all the force pushes. I switched to using `cfg_if` to try and make it harder to mess up platform support.

blinsay:
Looking at CI it looks like there are tier3 failures for unrelated things and some failures where the syscall is returning `ENOSYS`. Is there `nix` way to detect (lack of) support or should I document that `openat2` requires at least 5.6 and move on?

SteveLauC:
> Looking at CI it looks like there are tier3 failures for unrelated things 

Yeah, they are unrelated and fixed in #2344.

SteveLauC (approved):
The patch looks great! Thanks! Don't forget to add a changelog entry and rebase your branch to have CI issues resolved:)

PR #2343, opened 2024-03-22
Fix the build with the next version of libc

### LINKED ISSUE #2342: Build breakage with libc 0.2.154 from git
Nix builds successfully with the most recently released libc: 0.2.153.  But the version currently in git contains some backwards-incompatible changes.

Here is a simplified build output, on FreeBSD amd64:
```
error[E0308]: mismatched types
   --> src/sys/select.rs:113:65
    |
113 |             range: 0..highest.map(|h| h as usize + 1).unwrap_or(FD_SETSIZE),
    |                                                       --------- ^^^^^^^^^^ expected `usize`, found `i32`
    |                                                       |
    |                                                       arguments to this method are incorrect
    |
help: the return type of this call is `i32` due to the type of the argument passed
   --> src/sys/select.rs:113:23
    |
113 |             range: 0..highest.map(|h| h as usize + 1).unwrap_or(FD_SETSIZE),
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^----------^
    |                                                                 |
    |                                                                 this argument influences the return type of `unwrap_or`
note: method defined here
   --> /localhome/somers/.rustup/toolchains/nightly-x86_64-unknown-freebsd/lib/rustlib/src/rust/library/core/src/option.rs:953:12
    |
953 |     pub fn unwrap_or(self, default: T) -> T {
    |            ^^^^^^^^^
help: you can convert an `i32` to a `usize` and panic if the converted value doesn't fit
    |
113 |             range: 0..highest.map(|h| h as usize + 1).unwrap_or(FD_SETSIZE.try_into().unwrap()),
    |                                                                           ++++++++++++++++++++

error[E0308]: mismatched types
   --> src/unistd.rs:826:41
    |
826 |     unsafe { libc::execv(path.as_ptr(), args_p.as_ptr()) };
    |              -----------                ^^^^^^^^^^^^^^^ types differ in mutability
    |              |
    |              arguments to this function are incorrect
    |
    = note: expected raw pointer `*const *mut i8`
               found raw pointer `*const *const i8`
note: function defined here
   --> /usr/home/somers/src/rust/libc/src/unix/mod.rs:843:12
    |
843 |     pub fn execv(prog: *const c_char, argv: *const *mut c_char) -> ::c_int;
    |            ^^^^^
```

See
https://github.com/rust-lang/libc/pull/3597
https://github.com/rust-lang/libc/pull/3356
https://github.com/rust-lang/libc/pull/3466

asomers:
The next version of libc includes some backwards-incompatible changes.

Fixes #2342


asomers:
This PR sacrifices a little bit of type safety in order to build successfully with both old and new versions of libc.  I'm not sure if we should merge it, just because I'm not sure if the libc folks are actually planning to release these changes as part of the 0.2 series.

SteveLauC:
> This PR sacrifices a little bit of type safety in order to build successfully with both old and new versions of libc.

Given that the development of `libc` has been divided into 2 branches:

* 0.2 (the `libc-0.2` branch)
* 1.0 (the `main` branch, previously it was 0.3)

I think we need to choose one to follow, from the [tracking issue](https://github.com/rust-lang/libc/issues/3248), there are still a lot of work to do to make libc 1.0 ready so I guess it won't happen in the near future, and it is recommended to file PRs against the `libc-0.2` branch if one wants to make the patch available quickly, so we should probably go with 0.2?

However, most PRs are targeting the main branch, i.e., a lot of new APIs will be only available with libc 0.2.x, which makes it really hard to choose...


> I'm not sure if we should merge it, just because I'm not sure if the libc folks are actually planning to release these changes as part of the 0.2 series.

These breaking changes are in the main branch, so I guess they won't be included in the future 0.2.x release.

Cargo.toml
SteveLauC, line 1: There is a `package.rust-version` entry that also needs to be updated if we are gonna bump our MSRV

PR #2347, opened 2024-03-28
feat: introduce macos mount API support

oowl:
## Checklist:


oowl:
I do not know how to write the MacOS test because MacOS can not support `tmpfs` and `nullfs`, Is there any workaround for the MacOS mount test?  cc @asomers @SteveLauC

src/mount/mod.rs
```diff
@@ -5,8 +5,14 @@ mod linux;
 #[cfg(linux_android)]
 pub use self::linux::*;
 
-#[cfg(bsd)]
+#[cfg(bsd_without_macos)]
```
SteveLauC, line 8:
```suggestion
#[cfg(bsd_without_apple)]
```

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+use libc::c_int;
+
+libc_bitflags!(
+    /// Used with [`Nmount::nmount`].
+    pub struct MntFlags: c_int {
+        /// Do not interpret special files on the filesystem.
+        MNT_NODEV;
+        /// file system supports content protection
+        MNT_CPROTECT;
+        /// filesystem is stored locally
```
SteveLauC, line 11: Seems to be a copy-paste error:

```suggestion
        /// file system is quarantined
```

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+        MNT_QUARANTINE;
+        /// filesystem is stored locally
+        MNT_LOCAL;
+        /// quotas are enabled on filesystem
+        MNT_QUOTA;
+        /// identifies the root filesystem
+        MNT_ROOTFS;
+        /// FS supports volfs (deprecated flag in Mac OS X 10.5)
```
SteveLauC, line 19: Should we remove it as it has already been deprecated for years?

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+        MNT_JOURNALED;
+        /// Don't allow user extended attributes 
+        MNT_NOUSERXATTR;
+        /// filesystem should defer writes 
+        MNT_DEFWRITE;
+        /// don't block unmount if not responding
+        MNT_NOBLOCK;
+        /// file system is exported
+        MNT_EXPORTED;
+        /// All I/O to the file system should be done asynchronously.
```
SteveLauC, line 37: The kernel code says:

> file system written asynchronously


https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/sys/mount.h#L298

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+        MNT_ASYNC;
+        /// Force a read-write mount even if the file system appears to be
+        /// unclean.
```
SteveLauC, line 40: No idea what this flag is:

1. Kernel code says: force unmount or readonly change
2. Man page says: the filesystem should be forcibly unmounted even if files are still active

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+        MNT_NOEXEC;
+        /// Do not honor setuid or setgid bits on files when executing them.
+        MNT_NOSUID;
+        /// Mount read-only.
+        MNT_RDONLY;
+        /// Causes the vfs subsystem to update its data structures pertaining to
+        /// the specified already mounted file system.
+        MNT_RELOAD;
+        /// Create a snapshot of the file system.
+        ///
```
SteveLauC, line 56: Nit: We can remove this extra line

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+        match p {
+            Some(path) => path.with_nix_path(|p_str| f(p_str.as_ptr())),
+            None => Ok(f(std::ptr::null())),
+        }
+    }
+
+    let res = source.with_nix_path(|s|
+        target.with_nix_path(|t| {
+            with_opt_nix_path(data, |d| unsafe {
+                println!("mounting {:?} to {:?}", s, t);
```
SteveLauC, line 102: We should remove it as it is for debugging purposes

src/mount/apple.rs
```diff
@@ -0,0 +1,125 @@
...
+                libc::mount(
+                    s.as_ptr(),
+                    t.as_ptr(),
+                    flags.bits(),
+                    d as *mut libc::c_void,
```
SteveLauC, line 107: For pointer casting, [we prefer `.cast()` and `.cast_mut()` over the `as` keyword](https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#pointer-type-casting):

```suggestion
                    d.cast().cast_mut(),
```

BTW, we are casting an immutable reference (`&P3`) to a muable pinter (`*mut c_void`), is this safe?

SteveLauC (changes requested):
Hi, thanks for your interest in contributing to Nix! I have left some comments.

And:

1. A changelog entry is needed, please take a look at [doc](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add one.
2. There are still some `#[cfg(apple_targets)]` attributes in `bsd.rs`, I think we should remove them.

SteveLauC:
> I do not know how to write the MacOS test because MacOS can not support `tmpfs` and `nullfs`, Is there any workaround for the MacOS mount test? cc @asomers @SteveLauC

For tests, I think we can just do something simple to ensure that the binding works, e.g., using the following stupid code:

```sh
$ cat src/main.rs
use nix::{errno::Errno, mount::{mount, MntFlags}};

fn main() {
    let res = mount::<str, str, str>("", "", MntFlags::empty(), None);
    assert_eq!(res, Err(Errno::ENOENT));
}

$ cargo r -q
mounting "" to ""

$ echo $?
0
```

src/mount/apple.rs
oowl: Yeah, MacOS has some gap doc sometimes, I just follow BSD doc in this flag. Should we need to change?

src/mount/apple.rs
oowl: `libc:mount()` needs to recv `libc:c_void` type, It's hard to cast type to `libc:c_void` by the `cast` function, So I have changed it to `d.cast_mut() as *mut libc::c_void,`

oowl:
@SteveLauC I have made some changes depending on your comment, Let's continue.

src/mount/apple.rs
SteveLauC: > Should we need to change?

We cannot do things right if Apple doesn't take their things seriously, so I guess no

src/mount/mod.rs
```diff
@@ -5,8 +5,14 @@ mod linux;
 #[cfg(linux_android)]
 pub use self::linux::*;
 
-#[cfg(bsd)]
+#[cfg(bsd_without_apple)]
 mod bsd;
```
SteveLauC, line 9: We may want to change module and file name as well, for the sake of consistency:)

```suggestion
mod bsd_without_apple;
```

src/mount/apple.rs
SteveLauC: >  It's hard to cast type to libc:c_void by the cast function, So I have changed it to d.cast_mut() as *mut libc::c_void,

Yeah, `.cast().cast_mut()` should not work as the pointee type is unknown, `.cast_mut().cast()` should work.

> BTW, we are casting an immutable reference (&P3) to a muable pinter (*mut c_void), is this safe?

Will `mount(2)` write to that pointer, if so, when a UB could happen:<

oowl: From https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/mount.2.html man page
> 
     int
     mount(const char *type, const char *dir, int flags, void *data);

     int
     unmount(const char *dir, int flags);


>     
       Data is a pointer to a structure that contains the type specific argu-ments arguments
       ments to mount.  The format for these argument structures is described in
       the manual page for each filesystem.

I think it's not be written, but nobody knows how Apple doing I think.

SteveLauC (approved):
Thx!

Xuanwo:
Hello, @SteveLauC, is there a plan to get this PR released?

SteveLauC:
> Hello, @SteveLauC, is there a plan to get this PR released?

Personally, I don't have a plan to do the next release in the near future. Do you guys need to patch? If so, can you use a git dependency?

Xuanwo:
> Personally, I don't have a plan to do the next release in the near future.

Is there a specific reason? I believe it's just a patch update that should have minimal impact.

> Do you guys need to patch? 

Yes.

> If so, can you use a git dependency?

No, we can't release with git dependency.

SteveLauC:
> Is there a specific reason?

Not a strong reason, I simply want to get more work done in a release.

>  I believe it's just a patch update that should have minimal impact.

Yeah, I agree, Nix always releases on demand, so I guess I will do a release as requested.

Xuanwo:
> Not a strong reason, I simply want to get more work done in a release.

That will be very helpful!

PR #2352, opened 2024-04-02
Update bitflags to latest version

neumann-paulmann:
## What does this PR do
Update dependency of `bitflags` to latest release to avoid duplicate crate versions in [serialport](https://github.com/serialport/serialport-rs).

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ targets = [
 
 [dependencies]
 libc = { version = "0.2.153", features = ["extra_traits"] }
-bitflags = "2.5.0"
+bitflags = "2.3"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
 memoffset = { version = "0.9", optional = true }


asomers (changes requested):
Firstly, this PR combines formatting changes with content changes.  Don't do that.  Secondly, instead of requiring the latest bitflags, can you relax the range?  Maybe make the requirement "2.3" instead?  That would allow cargo to choose anything from 2.3.0 or later, but not 3.0 .

asomers (changes requested):
That looks better.  Can you confirm that with this setting, cargo will let you do either `cargo update -p bitflags --precise 2.3.1` or `cargo update -p bitflags --precise 2.5.0`?  And can you please squash your commits?

As for style settings, they're right in rustfmt.toml.  But they aren't applied to Cargo.toml, only to .rs files.

Finally, you should probably add a "added" entry in the changelog dir.

asomers:
> Thank you for the hint. Alas, both commands fail:
> 
> ```
> ~/RustroverProjects/nix> cargo update -p bitflags --precise 2.3.1
> error: There are multiple `bitflags` packages in your project, and the specification `bitflags` is ambiguous.
> Please re-run this command with one of the following specifications:
>   bitflags@1.3.2
>   bitflags@2.5.0
> ~/RustroverProjects/nix> cargo update -p bitflags --precise 2.5.0
> error: There are multiple `bitflags` packages in your project, and the specification `bitflags` is ambiguous.
> Please re-run this command with one of the following specifications:
>   bitflags@1.3.2
>   bitflags@2.5.0
> ~/RustroverProjects/nix>                         
> ```

That means that you have some other dependency pulling in bitflags-1.3.2.  In that case, modify the commands I posted to specify the exact version of bitflags that you want to update, like this for example: `cargo update -p bitflags@2.3.1 --precise 2.5.0`

PR #2364, opened 2024-04-13
ci: check newly-added CHANGELOGs in CI

### LINKED ISSUE #2357: Check `CHANGELOG` PR ID in CI
We should add a CI workflow to check if the PR ID in the new `CHANGELOG` file name matches the actual PR ID, I hasn't investigate if this is something we can do with GHA, but I think we should do it if we can.

SteveLauC:
## What does this PR do

This PR adds a new composite Github Action `check_changelog`(.github/actions/check_changelog/action.yml) and applies it in our `ci.yml`, for those newly-added changelog files under the `changelog` directory, it will check:

1. the CHANGELOG kind belongs to [added, changed, fixed, removed]
2. the PR number matches the real one
3. the file extension is `md`

Closes #2357

## Checklist:

### DIFF
--- a/.github/actions/check_changelog/action.yml
+++ b/.github/actions/check_changelog/action.yml
@@ -1,0 +1,41 @@
+name: 'Check CHANGELOG'
+description: 'Check CHANGELOG kind and PR number'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Get newly added CHANGELOGs
+        id: new-changelogs
+        uses: tj-actions/changed-files@v44
+        with:
+          # Only checek the files under the `changelog` directory 
+          files: changelog/** 
+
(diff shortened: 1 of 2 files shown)


.github/actions/check_changelog/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  using: "composite"
+  steps:
+    - name: Get newly added CHANGELOGs
+      id: new-changelogs
+      uses: tj-actions/changed-files@v44
+      with:
+        # Only checek the files under the `changelog` directory 
+        files: changelog/** 
+
+    - name: List added CHANGELOGs
```
SteveLauC, line 14: Should be something like `Check them`

PR #2373, opened 2024-04-14
feat: Use type parameters to allow `{get,set}regset` to use different register set structs

hack3ric:
## What does this PR do

Previous implementations of `{get,set}regset` are incorrect when register set value is anything other than `NT_PRSTATUS`. This PR adds the ability to return corresponding struct to the value using type parameters `S: RegisterSet`, so the behaviour is correct.

Added `NT_PRSTATUS` as well as `NT_PRFPREG` types so we can also probe floating-point registers. Other register sets may be inspected as well, such as arch-specific vector registers, but I don't have a use case right now, and it is easy to implement afterwards with the framework in this PR.

## Checklist:

This is an improvement over https://github.com/nix-rust/nix/pull/2044, so changelog remains the same.

### DIFF
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -172,26 +172,89 @@ libc_enum! {
     }
 }
 
+#[cfg(all(
+    target_os = "linux",
+    target_env = "gnu",
+    any(
+        target_arch = "x86_64",
+        target_arch = "x86",
+        target_arch = "aarch64",
+        target_arch = "riscv64",
+    )
+))]
 libc_enum! {
-    #[cfg(all(
-        target_os = "linux",
-        target_env = "gnu",
-        any(
-            target_arch = "x86_64",
-            target_arch = "x86",
-            target_arch = "aarch64",
-            target_arch = "riscv64",
-        )
-    ))]
     #[repr(i32)]
-    /// Defining a specific register set, as used in [`getregset`] and [`setregset`].
+    /// Defines a specific register set, as used in `PTRACE_GETREGSET` and `PTRACE_SETREGSET`.
     #[non_exhaustive]
-    pub enum RegisterSet {
+    pub enum RegisterSetValue {
         NT_PRSTATUS,
         NT_PRFPREG,
         NT_PRPSINFO,
         NT_TASKSTRUCT,
         NT_AUXV,
+    }
+}
+
+#[cfg(all(
+    target_os = "linux",
+    target_env = "gnu",
+    any(
+        target_arch = "x86_64",
+        target_arch = "x86",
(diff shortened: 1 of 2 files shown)


src/sys/ptrace/linux.rs
```diff
@@ -195,6 +195,63 @@ libc_enum! {
...
+        target_arch = "riscv64",
+    )
+))]
+/// Register sets used in [`getregset`] and [`setregset`]
+pub mod regset {
+    use super::*;
+
+    #[derive(Debug, Clone, Copy)]
+    /// General-purpose registers.
+    pub enum NT_PRSTATUS {}
```
SteveLauC, line 233: Nit: maybe using a unit struct here would be more suitable?

```suggestion
    pub struct NT_PRSTATUS;
```

src/sys/ptrace/linux.rs
```diff
@@ -195,6 +195,63 @@ libc_enum! {
...
+    target_env = "gnu",
+    any(
+        target_arch = "x86_64",
+        target_arch = "x86",
+        target_arch = "aarch64",
+        target_arch = "riscv64",
+    )
+))]
+/// Represents register set areas, such as general-purpose registers or floating-point registers.
+pub trait RegisterSet {
```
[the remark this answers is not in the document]
SteveLauC: Yeah, I agree, users can mistakenly specify a wrong type and call `assume_init()` on it, then UB could happen

TheJonny:
It could be more direct, if the `RegisterSet` would be implemented on the actual register set types. but the current suggestion has the advantage of having names one could recognize from the man page.

hack3ric:
Sorry for the late response. Reviews have been addressed.

> It could be more direct, if the RegisterSet would be implemented on the actual register set types. but the current suggestion has the advantage of having names one could recognize from the man page.

I'd add that it could be confusing to implement `RegisterSet` on actual register set types. `user_regs_struct` may be universal, but every platform has its own floating-pointer register space types (see `impl RegisterSet for NT_PRFPREG`).

SteveLauC (approved):
Thanks!

src/sys/ptrace/linux.rs
SteveLauC: Just realized that using a zero-variants enum would be better than a unit struct if we don't want our users to construct this type, so I guess this is the reason why you were using it here?

hack3ric: Yes. Unit struct is OK since it should be impossible to misuse it, but enum would probably be more elegant. Up to you :)

SteveLauC: Yeah, I think your idea is better and we should use a zero-variants enum, so I will file a PR for this:)

PR #2378, opened 2024-04-20
sys::prctl: Adding set_vma_anon_name.

devnexen:
to set a name for an `anonymous` region for Linux/Android.

## Checklist:

### DIFF
--- a/src/sys/prctl.rs
+++ b/src/sys/prctl.rs
@@ -9,9 +9,11 @@ use crate::errno::Errno;
 use crate::sys::signal::Signal;
 use crate::Result;
 
-use libc::{c_int, c_ulong};
+use libc::{c_int, c_ulong, c_void};
 use std::convert::TryFrom;
 use std::ffi::{CStr, CString};
+use std::num::NonZeroUsize;
+use std::ptr::NonNull;
 
 libc_enum! {
     /// The type of hardware memory corruption kill policy for the thread.
@@ -212,4 +214,15 @@ pub fn set_thp_disable(flag: bool) -> Result<()> {
 /// Get the "THP disable" flag for the calling thread.
 pub fn get_thp_disable() -> Result<bool> {
     prctl_get_bool(libc::PR_GET_THP_DISABLE)
+}
+
+/// Set an identifier (or reset it) to the address memory range.
+pub fn set_vma_anon_name(addr: NonNull<c_void>, length: NonZeroUsize, name: Option<&CStr>) -> Result<()> {
+    let nameref = match name {
+        Some(n) => n.as_ptr(),
+        _ => std::ptr::null()
+    };
+    let res = unsafe { libc::prctl(libc::PR_SET_VMA, libc::PR_SET_VMA_ANON_NAME, addr.as_ptr(), length, nameref) };
+
+    Errno::result(res).map(drop)
 }
--- a/test/sys/test_prctl.rs
+++ b/test/sys/test_prctl.rs
@@ -122,4 +122,38 @@ mod test_prctl {
 
         prctl::set_thp_disable(original).unwrap();
     }
+
+    #[test]
+    fn test_set_vma_anon_name() {
+        use nix::errno::Errno;
+        use nix::sys::mman;
+        use std::num::NonZeroUsize;
+
+        const ONE_K: libc::size_t = 1024;
+        let sz = NonZeroUsize::new(ONE_K).unwrap();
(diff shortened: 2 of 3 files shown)


src/sys/prctl.rs
```diff
@@ -213,3 +215,10 @@ pub fn set_thp_disable(flag: bool) -> Result<()> {
 pub fn get_thp_disable() -> Result<bool> {
     prctl_get_bool(libc::PR_GET_THP_DISABLE)
 }
+
+/// Set an identifier (or reset it) to the address memory range.
+pub fn set_addr_name(addr: NonNull<c_void>, length: NonZeroUsize, name: &CStr) -> Result<()> {
+    let res = unsafe { libc::prctl(libc::PR_SET_VMA, libc::PR_SET_VMA_ANON_NAME, addr.as_ptr(), length, name.as_ptr()) };
```
SteveLauC, line 221: From the manual, to reset the name, the arg5 has to be NULL:

>  If arg5 is NULL, the name of the appropriate anonymous virtual memory areas will be reset.

Maybe we should change the `name` argument to `Option<&CStr>`?

src/sys/prctl.rs
```diff
@@ -213,3 +215,10 @@ pub fn set_thp_disable(flag: bool) -> Result<()> {
 pub fn get_thp_disable() -> Result<bool> {
     prctl_get_bool(libc::PR_GET_THP_DISABLE)
 }
+
+/// Set an identifier (or reset it) to the address memory range.
+pub fn set_addr_name(addr: NonNull<c_void>, length: NonZeroUsize, name: &CStr) -> Result<()> {
```
SteveLauC, line 220: I kinda think we should rename this function to `set_vma_anon_name()` so that it is consistent with the attribute name  `SET_VMA_ANON_NAME`, though I do agree that `set_addr_name` is the clearer one.

test/sys/test_prctl.rs
```diff
@@ -122,4 +122,40 @@ mod test_prctl {
...
+        )
+        .unwrap_err();
+        assert_eq!(err, Errno::EINVAL);
+        // `CONFIG_ANON_VMA_NAME` kernel config might not be set
+        prctl::set_addr_name(
+            ptr,
+            sz,
+            CStr::from_bytes_with_nul(b"Nix\0").unwrap(),
+        )
+        .unwrap_or_default();
```
SteveLauC, line 157: Why are we using `.unwrap_or_default()` here

SteveLauC (changes requested):

test/sys/test_prctl.rs
devnexen: because of the comment above, the kernel might not support the feature.

src/sys/prctl.rs
```diff
@@ -213,3 +215,14 @@ pub fn set_thp_disable(flag: bool) -> Result<()> {
 pub fn get_thp_disable() -> Result<bool> {
     prctl_get_bool(libc::PR_GET_THP_DISABLE)
 }
+
+/// Set an identifier (or reset it) to the address memory range.
+pub fn set_vma_anon_name(addr: NonNull<c_void>, length: NonZeroUsize, name: Option<&CStr>) -> Result<()> {
+    let nameref = match name {
+        Some(n) => n.as_ptr(),
+        _ => std::ptr::null_mut()
```
SteveLauC, line 223: Looks like we should use `null()` here

```suggestion
        _ => std::ptr::null()
```

test/sys/test_prctl.rs
SteveLauC: Sorry about missing that!

Google search shows that this option is a C macro defined in the kernel code, and if it is not defined, then this prctl call would return  `EINVAL`:

```diff
> +#else /* CONFIG_ANON_VMA_NAME */
> +static int prctl_set_vma(unsigned long opt, unsigned long start,
> +                        unsigned long size, unsigned long arg)
> +{
> +       return -EINVAL;
> +}
> +#endif /* CONFIG_ANON_VMA_NAME */
```

Seems that there is no explicit way to check if this option is set or not?

SteveLauC (changes requested):

devnexen:
> Seems that there is no explicit way to check if this option is set or not?

No and in this case it does not really matter, this feature is more the "icing of top of cake" kind, so if it does not work it s just a no-op :)

SteveLauC (approved):
Thx!

PR #2380, opened 2024-04-21
unistd: fdatasync support for apple.

devnexen:
is not present in system headers, however the syscall is actually implemented.


SteveLauC (approved):
Thanks!

PR #2381, opened 2024-04-22
test: grab the SIGNAL_MTX lock in test_sigaction

SteveLauC:
## What does this PR do

This test changes the process-wide signal handler so we should use a lock to serialize it with other tests.

cc @TheJonny

## Checklist:


SteveLauC:
This issue was found in #2375 by @TheJonny

PR #2386, opened 2024-04-22
test: Grab FORK_MTX when unmounting or forking in the mount tests

TheJonny:
fork() duplicates all open file descriptors, which then could prevent a successful umount().

So either when creating file descriptors or when requiring a file descriptor to be closed, we must make sure that no child processes have copies. since wile descriptors are created all over the place, It is more practical to wait for all child processes before unlock.

# TODO:
 - [x] verify the semantic of FORK_MTX: how long must it be held? This PR assumes that the mutex is locked until all child processes have exited.

### DIFF
--- a/test/mount/test_mount.rs
+++ b/test/mount/test_mount.rs
@@ -53,6 +53,8 @@ fn test_mount_tmpfs_without_flags_allows_rwx() {
         .unwrap_or_else(|e| panic!("read failed: {e}"));
     assert_eq!(buf, SCRIPT_CONTENTS);
 
+    // while forking and unmounting prevent other child processes
+    let _m = FORK_MTX.lock();
     // Verify execute.
     assert_eq!(
         EXPECTED_STATUS,
@@ -89,6 +91,8 @@ fn test_mount_rdonly_disallows_write() {
             .unwrap()
     );
 
+    // wait for child processes to prevent EBUSY
+    let _m = FORK_MTX.lock();
     umount(tempdir.path()).unwrap_or_else(|e| panic!("umount failed: {e}"));
 }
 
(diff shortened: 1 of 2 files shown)


SteveLauC:
Hi, thanks for working on this!

What about we set the close-on-exec flag for the test files in that bind-mount fs (as well as the fds in other mount tests), I kinda think this would be more straightforward, thoughts?

TheJonny:
The close-on exec-flag will not prevent the issue, because the issue is not exec, but fork: it copies any opened file descriptors and then they live until exit or exec. if a file descriptor to something on the mountpoint is forked, unmount will fail.

SteveLauC:
> it copies any opened file descriptors and then they live until exit or exec.

You are right, there is still a timespan between `fork()` and `execxxx()` that could theoretically make the test fail

test/mount/test_mount.rs
```diff
@@ -89,6 +91,8 @@ fn test_mount_rdonly_disallows_write() {
             .unwrap()
     );
 
+    // wait for child processes to prevent EBUSY
+    let _m = FORK_MTX.lock();
```
SteveLauC, line 95: Is this lock necessary, this test does not open file descriptors, so theoretically no one could access the file system mounted at `tempdir`

test/test.rs
```diff
@@ -58,7 +58,8 @@ fn read_exact<Fd: AsFd>(f: Fd, buf: &mut [u8]) {
 }
 
 /// Any test that creates child processes must grab this mutex, regardless
```
SteveLauC, line 60: Maybe we should update this comment to something like:

```suggestion
/// Any test that creates child processes or can be affected by child processes must grab this mutex, regardless
```

test/mount/test_mount.rs
TheJonny: You are right, the `File::create` has to fail, so this is not necessary

SteveLauC (approved):
Thanks!

PR #2387, opened 2024-04-26
fcntl: change (vm)splice, tee using AsFd instead.

devnexen:


src/fcntl.rs
```diff
@@ -1080,13 +1080,13 @@ pub fn splice(
 /// # See Also
 /// *[`tee`](https://man7.org/linux/man-pages/man2/tee.2.html)
 #[cfg(linux_android)]
-pub fn tee(
-    fd_in: RawFd,
-    fd_out: RawFd,
+pub fn tee<F: AsFd>(
```
SteveLauC, line 1083: For this syscall, even though it requests the 2 fd arguments are all pipes, but technically they still can be different Rust types, so maybe we should use 2 generic types here? Thoughts?

devnexen: makes sense yes.

SteveLauC (approved):
Thanks!

PR #2393, opened 2024-05-05
fcntl adding apple F_PREALLOCATE flag.

devnexen:
## Checklist:

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -802,7 +802,12 @@ pub enum FcntlArg<'a> {
     F_RDADVISE(libc::radvisory),
     /// Turn read ahead off/on
     #[cfg(apple_targets)]
-    F_RDAHEAD(bool)
+    F_RDAHEAD(bool),
+    /// Pre-allocate storage with different policies on fd.
+    /// Note that we want a mutable reference for the OUT
+    /// fstore_t field fst_bytesalloc.
+    #[cfg(apple_targets)]
+    F_PREALLOCATE(&'a mut libc::fstore_t),
     // TODO: Rest of flags
 }
 
@@ -919,7 +924,11 @@ pub fn fcntl<Fd: std::os::fd::AsFd>(fd: Fd, arg: FcntlArg) -> Result<c_int> {
             F_RDAHEAD(on) => {
                 let val = if on { 1 } else { 0 };
                 libc::fcntl(fd, libc::F_RDAHEAD, val)
-            }
+            },
+            #[cfg(apple_targets)]
+            F_PREALLOCATE(st) => {
+                libc::fcntl(fd, libc::F_PREALLOCATE, st)
+            },
         }
     };
 
--- a/changelog/2393.added.md
+++ b/changelog/2393.added.md
@@ -1,0 +1,1 @@
+Add `fcntl`'s `F_PREALLOCATE` constant for Apple targets.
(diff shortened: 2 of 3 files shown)


SteveLauC:
I am working on the CI issues, they are not related to this PR

SteveLauC:
Gentle ping on this PR, would you like to rebase and finish it?

devnexen:
seems libc 0.2.154 not yet available (just the tag on the repo but not yet published maybe) ?

SteveLauC:
> seems libc 0.2.154 not yet available (just the tag on the repo but not yet published maybe) ?

Right, it has been yanked, we have to wait for https://github.com/rust-lang/libc/pull/3682

SteveLauC:
The issue has been fixed, please rebase your branch:)

src/fcntl.rs
```diff
@@ -692,6 +692,9 @@ pub enum FcntlArg<'a> {
     /// Return the full path without firmlinks of the fd.
     #[cfg(apple_targets)]
     F_GETPATH_NOFIRMLINK(&'a mut PathBuf),
+    /// Pre-allocate storage with different policies on fd.
+    #[cfg(apple_targets)]
+    F_PREALLOCATE(&'a libc::fstore_t),
```
SteveLauC, line 697: The associated buffer has an `OUT` field: 

```c
typedef struct fstore {
	    u_int32_t fst_flags;      /* IN: flags word */
	    int       fst_posmode;    /* IN: indicates offset field */
	    off_t     fst_offset;     /* IN: start of the region */
	    off_t     fst_length;     /* IN: size of the region */
	    off_t     fst_bytesalloc; /* OUT: number of bytes allocated */
} fstore_t;
```

i.e., the buffer can be written, if we use an immutable reference here, then UB could happen


Ref: https://keith.github.io/xcode-man-pages/fcntl.2.html#F_PREALLOCATE

SteveLauC: I just tried implementing this on my mac, I think we should create a helper type for those `IN` fields, and return the `OUT` field through `fcntl()`.

For that `fst_falgs` field, we would create a bitflags struct, for `fst_posmode`, a Rust enum would suffice.


And [F_ALLOCATEPERSIST](https://keith.github.io/xcode-man-pages/fcntl.2.html#F_ALLOCATEPERSIST) is not available in libc, we should add it first.

SteveLauC (changes requested):

src/fcntl.rs
SteveLauC: libc PR filed: https://github.com/rust-lang/libc/pull/3713

src/fcntl.rs
```diff
@@ -800,6 +800,9 @@ pub enum FcntlArg<'a> {
     /// Issue an advisory read async with no copy to user
     #[cfg(apple_targets)]
     F_RDADVISE(libc::radvisory),
+    /// Pre-allocate storage with different policies on fd.
```
SteveLauC, line 803: Do you think we shoud document the reason why a mutable reference is needed here? i.e., there is a OUT field in the associated sturcutre🤔

SteveLauC (approved):
LGTM, thanks!

SteveLauC:
Looks like you need to resolve the conflict in `src/fcntl.rs`.

PR #2394, opened 2024-05-05
ci: fix CI by correcting wrong cfg value & removing cache with sudo

SteveLauC:
## What does this PR do

This PR tries to fix the CI, there are 2 issues here:

1. New nightly toolchain added checks for cfg values, and they work pretty well
   1. there are few typos in our cfg values (which is really unfortunate) `fushsia -> fuchsia`
   2. `target_arch = x32` does not exist

2. For CI that are using our MSRV toolchain, the last step `before_cache_script` would fail with `Permission denied` because the files that it wants to delete are owned by root: https://github.com/nix-rust/nix/actions/runs/8958096003/job/24601956503?pr=2394
   
    This is kinda weird because it didn't happen in the past, and for now, it only happens with our MSRV toolchain.

```
uid=1001(runner) gid=127(docker) groups=127(docker),4(adm),101(systemd-journal)
/home/runner/.cargo/registry/index
├── [root     root    ]  github.com-1ecc6299db9ec823
│   ├── [root     root    ]  .cache
│   │   ├── [root     root    ]  3
│   │   │   └── [root     root    ]  s
│   │   │       └── [root     root    ]  syn
│   │   ├── [root     root    ]  as
│   │   │   └── [root     root    ]  se
│   │   │       └── [root     root    ]  assert-impl
│   │   ├── [root     root    ]  au
│   │   │   └── [root     root    ]  to
│   │   │       └── [root     root    ]  autocfg
```

## Checklist:

### DIFF
--- a/test/test.rs
+++ b/test/test.rs
@@ -15,7 +15,7 @@ mod test_fcntl;
 mod test_kmod;
 #[cfg(any(
     freebsdlike,
-    target_os = "fushsia",
+    target_os = "fuchsia",
     target_os = "linux",
     target_os = "netbsd"
 ))]
--- a/test/test.rs
+++ b/test/test.rs
@@ -13,11 +13,7 @@ mod test_errno;
 mod test_fcntl;
 #[cfg(linux_android)]
 mod test_kmod;
-#[cfg(any(
-    freebsdlike,
-    target_os = "linux",
-    target_os = "netbsd"
-))]
+#[cfg(any(freebsdlike, target_os = "linux", target_os = "netbsd"))]
 mod test_mq;
 #[cfg(not(target_os = "redox"))]
 mod test_net;
--- a/test/test.rs
+++ b/test/test.rs
@@ -15,7 +15,6 @@ mod test_fcntl;
 mod test_kmod;
 #[cfg(any(
     freebsdlike,
-    target_os = "fuchsia",
     target_os = "linux",
     target_os = "netbsd"
 ))]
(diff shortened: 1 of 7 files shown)


SteveLauC:
Well, I reproduced issue 2 on my host, both with our MSRV toolchain and the latest stable toolchain 1.78:

```sh
$ l ~/.cargo/registry/index
Permissions Links Size User Group Date Modified Name
drwxr-xr-x@     1    - root root   5 May 20:35  index.crates.io-6f17d22bba15001f
```

SteveLauC:
> This is kinda weird because it didn't happen in the past, and for now, it only happens with our MSRV toolchain.

Still have no idea why this didn't happen in the past, I fixed it by adding `sudo` to the last step for jobs that run test with `sudo`.

test/test.rs
```diff
@@ -13,12 +13,7 @@ mod test_errno;
 mod test_fcntl;
 #[cfg(linux_android)]
 mod test_kmod;
-#[cfg(any(
-    freebsdlike,
-    target_os = "fushsia",
```
SteveLauC, line 18: Before this PR, this test module was enabled for target `fushsia`(note it is a typo), it does not exist, and since module `mqueue` does not exist on Fuchsia, simply remove this line

test/test_fcntl.rs
```diff
@@ -13,7 +13,6 @@ use nix::fcntl::{openat2, OpenHow, ResolveFlag};
     target_env = "gnu",
     any(
         target_arch = "x86_64",
-        target_arch = "x32",
```
SteveLauC, line 16: This target does not exist, so remove it

Issue #2059, opened 2023-06-26
Unclear how to work with LinkAddr / sockaddr_ll


internetionals:
Currently I can't seem to find any way to create a new LinkAddr or modify a LinkAddr.

The only way I'm able to get a LinkAddr is through `getifaddrs()`. But after that I'm unable to change eg. `sll_protocol` to some specific protocol. I can turn `LinkAddr` into a `sockaddr_ll` (`as_ref().to_owned()`) and change it that way, but then I can't turn it back into a `LinkAddr` without using `unsafe`.

Is there a good reason why one shouldn't be able to get mutable access to the `sockaddr_ll` inside an owned `LinkAddr`? Or should that be done through mutators on the `LinkAddr` (I'll make an MR in a heartbeat :-))

Or would adding a constructor or `impl From<sockaddr_ll> for LinkAddr` be more idiomatic to the way these types are meant to be used?

asomers:
What are you trying to do anyway?  Change an interface's mac address?  There is a good reason why the inner `sockaddr_ll` cannot be exposed; changing its `sll_family` field could lead to UB, since that field functions as the discriminator for the `SockaddrStorage` union.  Mutators would be possible.  But currently, the only way to custom craft a `LinkAddr` would be to craft a `libc::sockaddr_ll`, then cast it to a `const *libc::sockaddr` and use `<LinkAddr as SockaddrLike>::from_raw`.

internetionals:
Working with packet(7) sockets basically needs options to at least modify `sll_protocol` as one can bind on an interface and choose which physical layer protocol one is interested in.

From the manpage:

```txt
By  default, all packets of the specified protocol type are passed to a packet socket.  To get packets only from
a specific interface use bind(2) specifying an address in a struct sockaddr_ll to bind the packet socket  to  an
interface.  Fields used for binding are sll_family (should be AF_PACKET), sll_protocol, and sll_ifindex.
```

When sending this structure is used to specify destination infromation when using `SOCK_DGRAM` (when using `SOCK_RAW` you have to craft the l2-layer header yourself). This involves being able to specify `sll_protocol`, `sll_halen` and `sll_addr`. All other fields should be cleared according to the documentation, though I recall it not really matter too much.

From the manpage:

```txt
When  you  send packets, it is enough to specify sll_family, sll_addr, sll_halen, sll_ifindex, and sll_protocol.
The other fields should be 0.  sll_hatype and sll_pkttype are set on received packets for your information.
```

[...]

PR #2396, opened 2024-05-06
Add `From<libc::sockaddr_ll> for LinkAddr`

jonatanzeidler:
## What does this PR do

This PR adds a `impl From<libc::sockaddr_ll> for LinkAddr` to help users of nix avoid `unsafe` code. See https://github.com/nix-rust/nix/issues/2059#issuecomment-1878463448 for reference.

## Checklist:

- ~~[ ] I have written necessary tests and rustdoc comments~~
- ~~[ ] A change log has been added if this PR modifies nix's API~~

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -1864,6 +1864,12 @@ mod datalink {
         }
     }
 
+    impl From<libc::sockaddr_ll> for LinkAddr {
+        fn from(sll: libc::sockaddr_ll) -> Self {
+            Self(sll)
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


jonatanzeidler:
Do you consider the `From` implementation a change to the API that should be noted in the changelog?

SteveLauC:
My current concert of this kind of conversion is that it can be unsafe since we have no guarantee from the values set in those raw libc types, see https://github.com/nix-rust/nix/issues/2327#issuecomment-2053587739

> Do you consider the From implementation a change to the API that should be noted in the changelog?

No since they are newly-added things

PR #2397, opened 2024-05-06
Add Rust type to represent a valid `libc::sockaddr_ll`

jonatanzeidler:
## What does this PR do

This PR adds a SockAddrLl struct to represent a `libc::socketaddr_ll` with the additional guarantee of valid data, that can be made into a `LinkAddr` without `unsafe` code. See #2327 and https://github.com/nix-rust/nix/issues/2059#issuecomment-1878463448 for reference.

## Checklist:

- ~~[ ] A change log has been added if this PR modifies nix's API~~


jonatanzeidler:
Just for clarification: Is the `unsafe` around constructing the `LinkAddr` due to the invariant (like `len` must match the other data) or about the variably sized types in C?

asomers:
> Just for clarification: Is the `unsafe` around constructing the `LinkAddr` due to the invariant (like `len` must match the other data) or about the potentially unsound libc types coming in via an FFI?

If you're referring to `LinkAddr::from_raw`, it must be `unsafe` because it reads from a raw pointer.  All methods that read from a raw pointer are unsafe.  And the reason that it reads from a raw pointer is because that's how many C APIs, like `getpeeraddr` provide sockaddrs.  C basically treats the various sockaddr types as a variable length tagged union.

This PR's implementation of `From<SockAddrLl>` is incorrect.  It doesn't guarantee valid data.  It doesn't validate the family or length fields, so it's easy to construct an invalid `sockaddr_ll`.  Is all you want a safe constructor for `sockaddr_ll`?  If so, I suggest you simply add one.  There's no need for a new type.

PR #2406, opened 2024-05-15
fix ControlMessageOwned::UdpGroSegments UDP packets processing type.

### LINKED ISSUE #2403: ControlMessageOwned::UdpGroSegments size mismatches the actual size
UdpGroSegments is declared as a u16 cmsg in the library but the Linux kernel actually puts an int in the cmsg: 

https://elixir.bootlin.com/linux/latest/source/include/linux/udp.h#L145

So applications using this option will break on big endian Linux machines. To fix it, the size may need to be adjusted to the size of an int for related kernel versions (doubt kernel maintainers might not change this in hope to not break user space: https://lore.kernel.org/netdev/CANn89iL2PCnC=6dOrozW0309W==tWcKpj2iwZgZAD_s0amvzLA@mail.gmail.com/T/)

devnexen:
The kernel, since 5.0, process the segmentation offload with a signed int.

close #2403.

### DIFF
--- a/changelog/2406.fixed.md
+++ b/changelog/2406.fixed.md
@@ -1,0 +1,1 @@
+Fixed ControlMessageOwned::UdpGroSegments wrapped type from u16 to i32 to reflect the used kernel's one.
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -773,7 +773,7 @@ pub enum ControlMessageOwned {
     #[cfg(target_os = "linux")]
     #[cfg(feature = "net")]
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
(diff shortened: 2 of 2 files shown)


changelog/2406.changed.md
SteveLauC, line 1: Kinda think this changelog should be named "2406.fixed.md" considering that reading an `u16` using `std::ptr::read_unaligned()` from an `i32` buffer would not work on big-endian machines

SteveLauC:
CI failure is not related to this PR, libc 0.2.154 has been yanked, we have to wait for the release of 0.2.155.

SteveLauC:
The issue has been fixed, please rebase your branch:)

SteveLauC (approved):
Thanks!

PR #2413, opened 2024-05-21
recvmsg: Check if CMSG buffer was too small and return an error

### LINKED ISSUE #2400: RecvMsg.cmsg() should check MSG_CTRUNC
If the cmsg buffer is too small, MSG_CTRUNC will be set in the header flags. This should be checked before returning `CmsgIterator`, or else the iterator might return invalid data.

agrover:
If MSG_CTRUNC is set, it is not safe to iterate the cmsgs, since they could have been truncated. Change `RecvMsg::cmsgs()` to return a `Result`, and to check for this flag (an API change).

Update tests for API change. Add test for too-small buffer.

fixes #2400

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -13,6 +13,7 @@ use libc::{self, c_int, size_t, socklen_t};
 #[cfg(all(feature = "uio", not(target_os = "redox")))]
 use libc::{
     c_void, iovec, CMSG_DATA, CMSG_FIRSTHDR, CMSG_LEN, CMSG_NXTHDR, CMSG_SPACE,
+    MSG_CTRUNC,
 };
 #[cfg(not(target_os = "redox"))]
 use std::io::{IoSlice, IoSliceMut};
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -601,11 +602,16 @@ pub struct RecvMsg<'a, 's, S> {
 impl<'a, S> RecvMsg<'a, '_, S> {
     /// Iterate over the valid control messages pointed to by this
     /// msghdr.
```
SteveLauC, line 604: Do you think we should document this error case given that it originates from Nix rather than the underlying syscall, users might be confused about it

test/sys/test_socket.rs
```diff
@@ -1509,11 +1509,20 @@ fn test_scm_credentials_and_rights() {
...
-    test_impl_scm_credentials_and_rights(space);
+    test_impl_scm_credentials_and_rights(space).unwrap();
 }
 
 #[cfg(linux_android)]
-fn test_impl_scm_credentials_and_rights(mut space: Vec<u8>) {
+#[test]
+fn test_too_small_cmsgspace() {
+    let space = vec![0u8; 4];
+    assert!(test_impl_scm_credentials_and_rights(space).is_err());
```
SteveLauC, line 1519: Let's assert it with `ENOBUF` if that is the expected value:

```suggestion
    assert_eq!(test_impl_scm_credentials_and_rights(space), Err(nix::errno::Errno::ENOBUF));
```

SteveLauC (changes requested):
Hi, thanks for the fix! Generally looks good!

agrover:
Proposed fix accepted and squashed. Took a swag at updating docstring to mention the new error condition.

SteveLauC (approved):
Thanks!

PR #2414, opened 2024-05-22
test: remove test of inode count in test_statfs.rs

### LINKED ISSUE #2411: Test `statfs_call()` fails on macOS because `statfs()` and `statvfs()` report different number of i-nodes
Sometimes, test `statfs_call()` fails in our macOS CI, e.g., [this one](https://github.com/nix-rust/nix/actions/runs/9143853411/job/25141049909):

```
failures:

---- sys::test_statfs::statfs_call stdout ----
thread 'sys::test_statfs::statfs_call' panicked at 'assertion failed: `(left == right)`
  left: `1609964262`,
 right: `1609964300`', test/sys/test_statfs.rs:47:5
```

It failed in [this line](https://github.com/nix-rust/nix/blob/ec4beb5a2229159021298d56af16e18a9151fb4e/test/sys/test_statfs.rs#L47), where `statfs()` and `statvfs()` were reporting different number of i-nodes.

SteveLauC:
## What does this PR do

This PR removes the test of inode count in `test/sys/test_statfs.rs` since it sometimes fails the CI on macOS, see issue #2411.

We don't want to bother with the issue, and actually we don't need to since we only need to ensure the Rust binding is working, so remove it.

Closes #2411

## Checklist:

### DIFF
--- a/test/sys/test_statfs.rs
+++ b/test/sys/test_statfs.rs
@@ -44,7 +44,6 @@ fn check_statfs_strict(path: &str) {
 // The cast is not unnecessary on all platforms.
 #[allow(clippy::unnecessary_cast)]
 fn assert_fs_equals(fs: Statfs, vfs: Statvfs) {
-    assert_eq!(fs.files() as u64, vfs.files() as u64);
     assert_eq!(fs.blocks() as u64, vfs.blocks() as u64);
     assert_eq!(fs.block_size() as u64, vfs.fragment_size() as u64);
 }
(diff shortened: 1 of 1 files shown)


test/sys/test_statfs.rs
```diff
@@ -93,7 +92,6 @@ fn assert_fs_equals_strict(fs: Statfs, vfs: Statvfs) {
     assert_eq!(fs.files_free() as u64, vfs.files_free() as u64);
     assert_eq!(fs.blocks_free() as u64, vfs.blocks_free() as u64);
     assert_eq!(fs.blocks_available() as u64, vfs.blocks_available() as u64);
-    assert_eq!(fs.files() as u64, vfs.files() as u64);
```
asomers, line 96: You should leave the check in `assert_fs_equals_strict`.  This function contains other assertions that are even more likely to fail, but it's only called from `#[ignore]` tests.

asomers (changes requested):

test/sys/test_statfs.rs
SteveLauC: done

asomers (approved):

PR #2417, opened 2024-05-25
ci: bump GitHub action runner

SteveLauC:
## What does this PR do

Bump GitHub action runner:

* `macos-14` -> `macos-latest`
   
   Currently, `macos-latest` is `macos-14`, so no changes for now.

* `ubuntu-20.04` -> `ubuntu-latest`

Update:

For `cross` test, we would still use Ubuntu 20.04 due to this issue: #2418.

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
   # Use cross for QEMU-based testing
   # cross needs to execute Docker, GitHub Action already has it installed
   cross:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


SteveLauC:
> Created an issue #2418 for that CI failure, do you think we should ignore this test under QEMU? cc @asomers

I think we should still use Ubuntu 20.04 for that cross test as it won't harm anything until 20.04 is no longer available.

When  20.04 becomes deprecated, if this issue still exists, we will ignore it, or we can happily migrate to newer versions.

PR #2420, opened 2024-05-30
sys::termios: adding ControlFlags::ECHO

devnexen:
enable echoing on input characters.

while at it with this new flag, fixing pty open_tty_pair test for solarish. While the test_write_ptty_pair passes, the test_read_ptty_pair fails when filling the master buffer.

Thus, we disable canonical and echo modes on the slave.


SteveLauC (commented):
Hi

> adding ControlFlags::ECHO

Is this cflag available on Solarish? [Document from Oracle](https://docs.oracle.com/cd/E36784_01/html/E36873/termios.h-3head.html) does not have it.

>  the test_read_ptty_pair fails when filling the master buffer.

Would you like to show me the error message shown on Solarish systems and why we need that special setup on them?

PR #2434, opened 2024-06-08
refactor: I/O safety for modules fcntl and dir

SteveLauC:
## What does this PR do

This PR adds I/O safety to `fcntl.rs` and `dir.rs`.

Ref: #1750 

## xxat() interfaces

For those `xxat()` interfaces, I didn't introduce a new trait to represent directory file descriptors, instead, I took the way how rustix implements it:

1. Define `AT_FDCWD: BorrowedFd<'static>`
2. Change the signature to `xxat<Fd: AsFd>(dirfd: Fd, ...)`

 because:

1. rustix has been using this interface for a long time, which means it would work pretty well.
3. It is indeed unfortunate that we cannot catch some errors at compile time, but we will eventually get notified at runtime, e.g., when passing a fd that is not a directory to `dirfd`, or using `AT_FDCWD` with non-`xxat()` functions.

## Checklist:

### DIFF
--- a/src/dir.rs
+++ b/src/dir.rs
@@ -3,7 +3,7 @@
 use crate::errno::Errno;
 use crate::fcntl::{self, OFlag};
 use crate::sys;
-use crate::{Error, NixPath, Result};
+use crate::{NixPath, Result};
 use cfg_if::cfg_if;
 use std::ffi;
 use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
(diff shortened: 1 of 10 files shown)


src/dir.rs
```diff
@@ -54,21 +54,33 @@ impl Dir {
...
+    ///
+    /// It is only safe if `fd` is an owned file descriptor.
     #[inline]
-    pub fn from<F: IntoRawFd>(fd: F) -> Result<Self> {
-        Dir::from_fd(fd.into_raw_fd())
+    #[deprecated(
+        since = "0.30.0",
+        note = "Deprecate this since it is not I/O-safe, use from_fd instead."
+    )]
+    pub unsafe fn from<F: IntoRawFd>(fd: F) -> Result<Self> {
```
SteveLauC, line 66: I think this function should be removed given its `IntoRawFd` usage, so deprecate it in the next release and remove it in futhre versions.

test/test_dir.rs
```diff
@@ -57,9 +57,3 @@ fn rewind() {
     assert_eq!(entries1, entries2);
     assert_eq!(entries2, entries3);
 }
-
-#[cfg(not(target_os = "haiku"))]
-#[test]
-fn ebadf() {
```
SteveLauC, line 63: This test is removed because this should not happen after the adoption of I/O safety.

PR #2439, opened 2024-06-09
refactor: I/O safety for sys/stat.rs

SteveLauC:
## What does this PR do

Add I/O safety to module `sys/stat`.

## Checklist:


thomaseizinger:
What is the intended migration for when I am currently calling `fstatat` with `None`? I am providing an absolute path as the argument and according to https://linux.die.net/man/2/fstatat, that means the first argument will be ignored. What should I be passing here now that I can no longer pass `None`?

SteveLauC:
Hi @thomaseizinger

Before this PR, `None` will be translated to `AT_FDCWD` under this hood. After this PR, you can use https://docs.rs/nix/latest/nix/fcntl/constant.AT_FDCWD.html. 

> I am providing an absolute path as the argument and according to https://linux.die.net/man/2/fstatat, that means the first argument will be ignored.

Since the kernel will ignore this argument, it doesn't matter what it is, so technically, you can use anything.

But, AT_FDCWD is what you want if you want do a 1:1 migration.

PR #2440, opened 2024-06-09
refactor: I/O safety for unistd.rs

SteveLauC:
## What does this PR do

Add I/O safety to module unistd.

## File descriptor duplication

The safe interface for `dup2()` proposed in this PR looks like this:

```rs
pub fn dup2<Fd: AsFd>(oldfd: Fd, newfd: &mut OwnedFd) -> Result<()>
```

We take `newfd` by mutable reference so that we can ensure we have exclusive access. This interface stops users from specifying `newfd` with arbitrary fd values, `unsafe fn dup2_raw()` comes as a remedy:

```rs
pub unsafe fn dup2_raw<Fd1: AsFd, Fd2: IntoRawFd>(
    oldfd: Fd1,
    newfd: Fd2,
) -> Result<OwnedFd>
```

`dup2_raw()` is unsafe because when specifying a fd that is open, one has to ensure the returned `OwnedFd` is the ONLY owner of this fd, or a double close will happen.

image
 

Using the above `dup2()` interface for stdin/stdout/stderr redirection is troublesome, one has to construct an `OwnedFd` from stdin/stdout/stderr, then `mem::forget()` it. We provide 3 helper functions to make this easier:

```rs
pub fn dup2_stdin<Fd: AsFd>(fd: Fd) -> Result<()>
pub fn dup2_stdout<Fd: AsFd>(fd: Fd) -> Result<()>
pub fn dup2_stderr<Fd: AsFd>(fd: Fd) -> Result<()>
```

Since the only flag that can be used with `dup3()` is `O_CLOEXEC`, which does not make much sense to stdin/stdout/stderr, I didn't provide the dup3 versions of these utilities.

## Checklist:

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -429,7 +429,6 @@ feature! {
 /// * [`dup2()`]
 /// * [`dup2_raw()`]
 /// * [`dup3()`]
-/// * [`dup3_raw()`]
 #[inline]
 pub fn dup<Fd: std::os::fd::AsFd>(oldfd: Fd) -> Result<std::os::fd::OwnedFd> {
     use std::os::fd::AsRawFd;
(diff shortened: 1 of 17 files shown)


src/unistd.rs
```diff
@@ -1295,19 +1504,21 @@ impl LinkatFlags {
 /// # References
 /// See also [linkat(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/linkat.html)
 #[cfg(not(target_os = "redox"))] // RedoxFS does not support symlinks yet
```
SteveLauC, line 1506: Not sure if hard link is supported by RedoxFS, but linkat() is for hard link rather than symlink

PR #2448, opened 2024-06-14
Fix inconsistent reg struct layout for 64bit tracer and 32bit tracee

### LINKED ISSUE #2447: `ptrace::getregs` & `ptrace::getregset` may lead to UB of uninitialized data read
`ptrace::getregs` and `ptrace::getregset` are used to get register values of certain process (tracee), which utilizes the `ptrace(PTRACE_GETREGSET, ..)` syscall:

```rust
pub fn getregset<S: RegisterSet>(pid: Pid) -> Result<S::Regs> {
    let request = Request::PTRACE_GETREGSET;
    let mut data = mem::MaybeUninit::<S::Regs>::uninit();
    let mut iov = libc::iovec {
        iov_base: data.as_mut_ptr().cast(),
        iov_len: mem::size_of::<S::Regs>(),
    };
    unsafe {
        ptrace_other(
            request,
            pid,
            S::VALUE as i32 as AddressType,
            (&mut iov as *mut libc::iovec).cast(),
        )?;
    };
    Ok(unsafe { data.assume_init() })
}
```

 In amd64 Linux, if the tracer is 64bit process, tracee is 32bit process, the `S::Regs` will be resolved to 64bit version of [`libc::user_regs_struct`](https://docs.rs/libc/0.2.155/libc/struct.user_regs_struct.html), while after the syscall return, the `S::VALUE` buffer is filled by kernel with 32bit version of such struct. (The `iovec.iov_len` field will be 68 after returning, which confirms such thing). Reading the returned `S::Regs` will lead to uninitialized data, which is UB.

Evian-Zhang:
## What does this PR do

Fix #2447 

## Checklist:

- [ x] I have read `CONTRIBUTING.md`
- [ x] I have written necessary tests and rustdoc comments

### DIFF
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -310,6 +310,9 @@ fn ptrace_peek(
 /// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
 /// on aarch64 and riscv64.
 ///
+/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, this function
+/// will return an error.
+///
 /// [ptrace(2)]: https://www.man7.org/linux/man-pages/man2/ptrace.2.html
 #[cfg(all(
     target_os = "linux",
@@ -342,6 +345,9 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
 }
 
 /// Get a particular set of user registers, as with `ptrace(PTRACE_GETREGSET, ...)`
+///
+/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, this function
+/// will return an error.
 #[cfg(all(
     target_os = "linux",
     target_env = "gnu",
@@ -367,6 +373,9 @@ pub fn getregset<S: RegisterSet>(pid: Pid) -> Result<S::Regs> {
             (&mut iov as *mut libc::iovec).cast(),
         )?;
     };
+    if iov.iov_len != mem::size_of::<S::Regs>() {
+        return Err(Errno::EIO);
+    }
     Ok(unsafe { data.assume_init() })
 }
 
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -304,7 +304,7 @@ fn ptrace_peek(
     }
 }
 
-/// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`. Call [`getregset`] for safe version
+/// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`. Call `getregset` for safe version
 ///
 /// Note that since `PTRACE_GETREGS` are not available on all platforms (as in [ptrace(2)]),
 /// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -304,14 +304,14 @@ fn ptrace_peek(
     }
 }
 
-/// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`
+/// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`. Call [`getregset`] for safe version
 ///
 /// Note that since `PTRACE_GETREGS` are not available on all platforms (as in [ptrace(2)]),
 /// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
 /// on aarch64 and riscv64.
 ///
-/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, this function
-/// will return an error.
+/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, the return value is
+/// undefined.
 ///
 /// [ptrace(2)]: https://www.man7.org/linux/man-pages/man2/ptrace.2.html
 #[cfg(all(
@@ -324,7 +324,7 @@ fn ptrace_peek(
         all(target_arch = "x86", target_env = "gnu")
     )
 ))]
-pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
+pub unsafe fn getregs(pid: Pid) -> Result<user_regs_struct> {
     ptrace_get_data::<user_regs_struct>(Request::PTRACE_GETREGS, pid)
 }
 
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -310,6 +310,8 @@ fn ptrace_peek(
 /// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
 /// on aarch64 and riscv64.
 ///
+/// # Safety
+///
 /// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, the return value is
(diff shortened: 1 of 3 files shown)


src/sys/ptrace/linux.rs
```diff
@@ -310,6 +310,9 @@ fn ptrace_peek(
 /// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
 /// on aarch64 and riscv64.
 ///
+/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, this function
+/// will return an error.
```
asomers, line 314: Does it?  Did you forget to add a size check below?

Evian-Zhang: Oh yes, I forgot this check. But as stated in the man page, PTRACE_GETREGS does not return the size of bytes written, so it seems that there is no easy way to check this?

asomers: Do you know how C programs use PTRACE_GETREGS for 32-bit targets, then?

Evian-Zhang: As far as I know, we should check the target tracee bitness in advance, instead of checking after the syscall returns. Is it better to mark this method unsafe and provide a safe wrapper which does the bitness check?

asomers: If it cannot guarantee that the return value will be correctly constructed, then yes it must be unsafe.  Either that, or it could return a union (which is unsafe to read from).

Evian-Zhang: Updated. I prefer to mark it as unsafe, since it is hard to import the 32bit user_regs_struct in `libc` crate.

src/sys/ptrace/linux.rs
```diff
@@ -304,12 +304,15 @@ fn ptrace_peek(
...
-/// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`
+/// Get user registers, as with `ptrace(PTRACE_GETREGS, ...)`. Call [`getregset`] for safe version
 ///
 /// Note that since `PTRACE_GETREGS` are not available on all platforms (as in [ptrace(2)]),
 /// `ptrace(PTRACE_GETREGSET, pid, NT_PRSTATUS, ...)` is used instead to achieve the same effect
 /// on aarch64 and riscv64.
 ///
+/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, the return value is
+/// undefined.
+///
```
asomers, line 315: The docs should have a `# Safety` section.

asomers (changes requested):
Could you please add a changelog entry?  Also, have you audited the other ptrace functions for similar behavior?

src/sys/ptrace/linux.rs
Evian-Zhang: Updated

Evian-Zhang:
> Could you please add a changelog entry? Also, have you audited the other ptrace functions for similar behavior?

I'll add one. For other ptrace functions, it is the `setregs` and `setregset` that should be considered. It does not involves UB in Rust side (no uninitialized memory is read by Rust), but it is clear that in the 64bit tracer and 32bit tracee situation, those functions does not do the right thing, and even worse, it may modify the tracee's regs with wrong data. Should we mark it unsafe or add some docs about this?

src/sys/ptrace/linux.rs
```diff
@@ -342,6 +347,9 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
 }
 
 /// Get a particular set of user registers, as with `ptrace(PTRACE_GETREGSET, ...)`
+///
+/// Currently, in x86_64 platform, if the tracer is 64bit and tracee is 32bit, this function
+/// will return an error.
```
SteveLauC, line 352: Nit: we should document that it will return an EIO or users may think this errno comes from the underlying ptrace syscall.

src/sys/ptrace/linux.rs
SteveLauC: > As far as I know, we should check the target tracee bitness in advance, instead of checking after the syscall returns. 

From this [post](https://unix.stackexchange.com/questions/106234/determine-if-a-specific-process-is-32-or-64-bit), it is not trivial to check the bitness, so I would also prefer to mark this function unsafe as well if we don't have a better option.

 > since it is hard to import the 32bit user_regs_struct in libc crate

Would you like to elaborate on this a bit?

Evian-Zhang: If we return a union, then this union should involve both 64bit and 32bit version of user_regs_struct, which is provided by libc crate instead of the nix itself. However, the struct definition is bound to the compile target of libc crate, we cannot import a 32bit user_regs_struct in libc if we compiles to x86_64

SteveLauC: > However, the struct definition is bound to the compile target of libc crate, we cannot import a 32bit user_regs_struct in libc if we compiles to x86_64

Right, the actual type is decided at compile time, so we cannot use the 32-bit version unless we define it in Nix, which is not something Nix should do.

```sh
$ pwd
libc/src/unix/linux_like/linux/gnu

$ rg "struct user_regs_struct"
b64/x86_64/mod.rs
178:    pub struct user_regs_struct {

b32/x86/mod.rs
74:    pub struct user_regs_struct {

b64/loongarch64/mod.rs
193:    pub struct user_regs_struct {

b64/riscv64/mod.rs
196:    pub struct user_regs_struct {

b32/riscv32/mod.rs
200:    pub struct user_regs_struct {

b64/aarch64/mod.rs
146:    pub struct user_regs_struct {
```

------

> As far as I know, we should check the target tracee bitness in advance, instead of checking after the syscall returns.

After figuring out the bitness, how do you use a 32-bit version structure in C if you know the tracee is a 32-bit process? Since in C, these types are also chosen at compile time with `#ifdef` macros.

asomers:
What about the ptrace functions in the BSD module?  Did you look at them?

src/sys/ptrace/linux.rs
Evian-Zhang: In GDB, it is done without the user_regs_struct. The regs are fetched [here](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/amd64-linux-nat.c;h=aa9295d57230d4f559f6b5c346658b786c0b0688;hb=HEAD#l222), and is used [here](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/amd64-nat.c;h=5c5c368c93d64671a86bf1778cc8cef497a7368e;hb=HEAD#l86)

In short, it provides a 64bit version buffer to receive the regs, and when storing it to internal structure, the tracee arch is checked and collect corresponding register values.

Evian-Zhang:
> What about the ptrace functions in the BSD module? Did you look at them?

I do not have a BSD machine, and cannot check whether it has the same behavior with Linux

src/sys/ptrace/linux.rs
Evian-Zhang: Updated

atalii:
Hi all - sorry to necro, but I'm looking at similar-ish code and wondered if any of you had a good reference for how PTRACE_{GET,SET}REGS{,ET} behave on x86 with a 64-bit tracer and 32-bit tracee. In my work, PTRACE_GETREGSET appears to allow for reliable reading of both 64-bit and 32-bit target GP registers (one must only check how much data was written into the iovec), but I'm only able to set registers from a 64-bit tracee to any target by reading, modifying, and writing the 64-bit user_regs_struct with PTRACE_{GET,SET}REGS. Needless to say, this is super confusing to me, and I don't know what operations guarantee what semantics. Curious if you'd have more info? Thanks (:

Issue #1143, opened 2019-10-24
Add SIGRTMIN and SIGRTMAX constants


kacejot:
Or at least remove Error on Signal::from_c_int, if signal is not between 0 and NSIG, so I can pass custom signals. Option here is also excess.

kacejot:
I use transmute for my purpose now and it works fine, but I with getting rid of unsafe block.
```rust
let signal: Signal = unsafe { std::mem::transmute(signal::SIGRTMIN() + 1) };
match kill(pid, signal) {
    ...
}
```

asomers:
I think this would be a reasonable change.  However, you'll have to add support to libc first.  And you'll need to deal with the fact that `SIGRTMIN` is a constant on some platforms and a function on others.

mkroening:
FYI, the SIGRTMIN and SIGRTMAX have been added to libc in https://github.com/rust-lang/libc/pull/2304.

n0toose:
FYI in case anybody comes across this as well: The `transmute` trick doesn't work anymore in more recent nightly builds (https://github.com/rust-lang/rust/pull/141759) because it is no longer possible to "shove" an arbitrary i32 inside of enums with an i32 repr anymore. 

See: https://github.com/hermit-os/uhyve/issues/1083

Issue #495, opened 2017-01-07
Realtime signals


SdNssr:
Realtime signals seem to be missing from `nix::sys::signal::Signal`, and quite a few functions seem to panic on receiving realtime signals:

- `nix::sys::signal::SigSet::wait`
- `nix::sys::wait::waitpid` (when the child died because of a realtime signal)
- and probably many others that I haven't yet used

So is this something that won't be supported, or is it just incomplete?

fiveop:
It is just incomplete. Feel free to add them. You might need to add the constants to libc first.

Susurrus:
@SdNssr Any interest in filing a PR for this functionality?

SdNssr:
No, not really.

ObsidianMinor:
I want to take a shot at this. Only thing I see being an issue is how much of the existing API we want to keep.

If we want to keep the existing `Signal` type (because conversions to and from strings is nice), then I'd propose new types `RtSignal` and `Signum`.

`RtSignal` would be used to define new real-time signal values, with the function `RtSignal::new(i32)` being used to create the values, offset from `SIGRTMIN`. So `RtSignal::new(0)` would be the minimum value, `RtSignal::new(1)` would be `SIGRTMIN() + 1`, etc. This would return an optional value and can't be converted to a string except via debug impl.

`Signum` would take the place of most usage of `Signal`, it being a less typed variant that can represent a `RtSignal` or `Signal` value. `SigSet`'s API would be based on `Signum`, accepting and returning instances of the type. You can attempt to unwrap an `RtSignal` or a `Signal` value from a `Signum`, for instance, for converting to a string, but functions wouldn't be defined on it.

Because `Signum` would be the primary type being used by `SigSet`, it would have it's own set of associated constants that correspond to the libc constants. These would be re-exported in place of the `Signal` enumerants. This is because I'm assuming that most people are using the `Signal` re-exported values for pattern matching on the result of `SigSet::wait`, so this would be the last painful upgrade path.

Alternatively, we rewrite the whole `Signal` type. We keep `FromStr`, but nothing can be converted into a real-time signal. `as_str` becomes something like `try_as_str`, which returns an optional value, none when a real-time signal is the value. `AsRef<str>` is removed, `fmt::Display` is removed, and `SignalIterator` is adjusted to iterate all valid real-time signal values as well.

Thoughts?

PR #2449, opened 2024-06-15
Add sigaction() versions that allow specify the signal by its number

ChrysoliteAzalea:
Currently, the ```nix::sys::signal::sigaction()``` function uses the ```Signal``` enum, which only covers standard signals, but not real-time signals. Using real-time signals with ```nix``` requires using the ```std::mem::transmute()``` function to assign an impossible value to the ```Signal``` enum, which is [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).

## What does this PR do

In this PR, I add two new functions: ```nix::sys::signal::sigaction_numeric()```, that uses signal numbers instead of the enum, and ```nix::sys::signal::sigaction_noretrieve()```, which also doesn't retrieve the old signal handler by supplying the null pointer as the ```oldact``` argument (to avoid trouble if the handler has been installed by the C code).

## Checklist:

- [yes] I have written necessary tests and rustdoc comments
- [?] A change log has been added if this PR modifies nix's API

### DIFF
--- a/src/sys/signal.rs
+++ b/src/sys/signal.rs
@@ -901,6 +901,56 @@ pub unsafe fn sigaction(signal: Signal, sigaction: &SigAction) -> Result<SigActi
     Errno::result(res).map(|_| SigAction { sigaction: unsafe { oldact.assume_init() } })
 }
 
+/// Changes the action taken by a process on receipt of a specific signal, which is specified by a signal number.
+///
+/// `signal` can be any signal except `SIGKILL` or `SIGSTOP`. On success, it returns the previous
+/// action for the given signal. If `sigaction` fails, no new signal handler is installed.
+///
+/// # Safety
+///
+/// * Signal handlers may be called at any point during execution, which limits
+///   what is safe to do in the body of the signal-catching function. Be certain
+///   to only make syscalls that are explicitly marked safe for signal handlers
+///   and only share global data using atomics.
+///
+/// * There is also no guarantee that the old signal handler was installed
+///   correctly.  If it was installed by this crate, it will be.  But if it was
+///   installed by, for example, C code, then there is no guarantee its function
+///   pointer is valid.  In that case, this function effectively dereferences a
+///   raw pointer of unknown provenance.
+pub unsafe fn sigaction_numeric(signal: libc::c_int, sigaction: &SigAction) -> Result<SigAction> {
+    let mut oldact = mem::MaybeUninit::<libc::sigaction>::uninit();
+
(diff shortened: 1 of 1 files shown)


SteveLauC:
Hi, thanks for the work on realtime signals.

> [?] A change log has been added if this PR modifies nix's API

Nix needs a CHANGELOG entry for those PRs that modify the API, please refer to [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to do this.

-------

Related to:

* #495 
* #1143

From my side, the thing that blocks us from supporting realtime signals is that some platforms define the signal number as constants, while other platforms use function. And signal numbers on Linux/glibc is a runtime thing since glibc uses the first 2 or 3 rt signals for internal usage. So we cannot easily add them to the existing `enum Signal` type. 

This PR proposes APIs that use raw `c_int` types with `sigaction()`, instead of adding new interfaces as workaround, I think we should also think about how to make rt signals work with existing interfaces, e.g., as mentioned in #495, `sigwait()` and `waitpid()` will panic when encountering rt signal, I am thinking about changing `enum Signal` to something like this:


```rs
#[repr(transparent)]
pub struct Signal(libc::c_int);

impl Signal {
    pub const SIGINT: Signal = Signal(libc::SIGINT);

    pub fn realtime(sig_num: libc::c_int) -> Result<Self, nix::errno::Errno> {
        let min = libc::SIGRTMIN();
        let max = libc::SIGRTMAX();

        if sig_num < min || sig_num > max {
            return Err(nix::errno::Errno::EINVAL);
        }

        Ok(Signal(sig_num))
    }
}
```

By using a wrapper of `c_int`, rt signals can be accommodated in it, though we will lost the convenient pattern matching after this conversion.

cc @asomers, would like to hear your thoughts on rt signal support.

SteveLauC:
```rs
pub enum Signal {
    Standard(nix::signal::Signal),
    Realtime(libc::c_int),
}
```

Or maybe something like this? Though I am not sure if realtime signal is supported by all the OSes supported by Nix.

ChrysoliteAzalea:
I wonder, should I modify existing functions, or add new one to avoid breaking changes to existing applications?

SteveLauC:
> avoid breaking changes to existing applications?

Breaking change is fine, the things I care about most are the usability and ergonomics of the new interfaces.

ChrysoliteAzalea:
Closing this PR since [a better solution was found](https://github.com/nix-rust/nix/pull/2451).

PR #2453, opened 2024-06-23
Fix the build with -Zdirect-minimal-versions

asomers:
The difference between -Zminimal-versions and -Zdirect-minimal-versions is subtle.  But the important point is that with the former, a new release published by one of our dependencies can cause our CI to fail. With the latter, it cannot.  That's why I think it's better to test with the latter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -345,7 +345,7 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
 
       - name: setup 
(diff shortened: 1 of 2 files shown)


SteveLauC:
> a new release published by one of our dependencies can cause our CI to fail.

So I guess that's one of the reasons why this flag is not recommended for usage:)

Screenshot 2024-06-24 at 8 20 04 AM

SteveLauC (approved):
Thanks!

PR #2469, opened 2024-08-09
nix::InterfaceAddress: adding interface information if it is UP.

devnexen:


SteveLauC (approved):
Looks good, thanks! CI failures are not related, I will handle them.

SteveLauC (commented):
Well, perhaps I need a reason for this change even though it is ok and trivial, we already have this info in `addr.flags.contains(InterfaceFlags::IFF_UP)`.

I think we need to consider what should be stored in the struct, currently, all the stored fields come from the libc struct `libc::ifaddrs`.

devnexen:
well, we can consider extra fields as info access aliases/shortcuts wdyt ?

SteveLauC:
> we can consider extra fields as info access aliases/shortcut

I am thinking that if getting this info is already straightforward, then there is no need to add a shortcut for it 😶

SteveLauC:
CI fixed in #2470

devnexen:
Sure, anyway up to you if you want to merge it, no hard feeling if you don't :)

SteveLauC:
Thanks for the understanding ❤️, I will close this PR then:)

PR #2472, opened 2024-08-16
Fix a clippy::unnecessary_cast lint on the latest nightly

asomers:
The Ipv4Ttl control message has a different size on Linux than on FreeBSD, and needs to be cast differently.


SteveLauC (approved):
Thanks!

PR #2474, opened 2024-08-22
Add direct conversion from SockAddrIn(6) to SocketAddr

pronebird:
## What does this PR do

At the moment nix only offers conversions between `SockaddrIn` -> `SocketAddrV4` and `SockaddrIn6` -> `SocketAddrV6` but no conversions to the container type, `SocketAddr` itself.

While it's trivial to wrap either `SocketAddrV4` or `SocketAddrV6` into `SocketAddr`, I think having a shortcut would improve ergonomics and essentially align nix with std, which supports `From` conversions from `SocketAddrV{4,6}` to `SocketAddr`.

## Checklist:

### DIFF
--- a/changelog/2474.added.md
+++ b/changelog/2474.added.md
@@ -1,0 +1,1 @@
+Add `From` trait implementation between `SocketAddr` and `Sockaddr`, `Sockaddr6`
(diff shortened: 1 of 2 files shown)


SteveLauC (approved):
LGTM, thanks!

PR #2476, opened 2024-08-29
Rename Flags EventFlag and MemFdCreateFlag

howjmay:
relates to #317

## What does this PR do

Rename Flags 

## Checklist:

### DIFF
--- a/test/sys/test_event.rs
+++ b/test/sys/test_event.rs
@@ -1,5 +1,5 @@
 use libc::intptr_t;
-use nix::sys::event::{EventFilter, EventFlag, FilterFlag, KEvent};
+use nix::sys::event::{EvFlags, EventFilter, FilterFlag, KEvent};
 
 #[test]
 fn test_struct_kevent() {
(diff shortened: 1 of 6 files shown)


SteveLauC (commented):
Existing changes look good, thanks!

Though you need to address the CI failures, i.e., rename the usages on other platforms as well.

And can we have something like this to help users migrate:

```rs
#[deprecated(since = "0.30.0", note = "Use `EvFlags instead`")]
pub type EventFlag = EvFlags;
```

SteveLauC:
Also, a changelog entry is needed, see [this](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add one.

test/sys/test_event.rs
```diff
@@ -11,7 +11,7 @@ fn test_struct_kevent() {
     let actual = KEvent::new(
         0xdead_beef,
         EventFilter::EVFILT_READ,
```
SteveLauC, line 13: @asomers, any idea what we should do to things like this `enum EventFilter`? 

```
use nix::sys::event::{EvFlags, EventFilter};
```

Honestly, it feels weird to me when putting them together 🥲, one aims to be short, typical C-style, one aims to be self-explanatory.

SteveLauC (approved):
Looks good, thank you:)

test/sys/test_event.rs
asomers: If does seem odd now. I think they'll need hard to remember like this. We should probably make the two names more consistent.

SteveLauC: Yeah

howjmay:
thank you too

SteveLauC:
Let's merge this first.

PR #2480, opened 2024-09-03
fcntl: adding macOs F_RDADVISE flag

devnexen:
Hints to prefetch blocks of the file into memory when possible. Can be useful to process large files.


SteveLauC (commented):
Interesting prefetch feature, can we have a test for this flag, just create a temp file and invoke the call against it:)

changelog/2113.added.md
SteveLauC, line 1: This was accidentally modified, please revert it.

src/fcntl.rs
```diff
@@ -797,6 +797,9 @@ pub enum FcntlArg<'a> {
     /// Return the full path without firmlinks of the fd.
     #[cfg(apple_targets)]
     F_GETPATH_NOFIRMLINK(&'a mut PathBuf),
+    /// Issue an advisory read async with no copy to user
+    #[cfg(apple_targets)]
+    F_RDADVISE(libc::radvisory),
```
SteveLauC, line 802: I saw this was changed in the "testing" commit, the previous API is indeed a bit unclear:

```
F_RDADVISE(off_t, c_int),
```

I am actually thinking about this:

```rs
F_RDADVISE {
    offset: off_t,
    count: c_int,
}
```

So semantically, the API will be more clear and users won't need to bother with the libc raw structure. The drawback is, we need to copy these 2 numbers, considering they are simply numbers, IMHO, it is fine to copy them. Your thoughts? 🤔

devnexen: I was being consistent with flags passing `libc::flock`.

SteveLauC: Get it, makes sense to me then:)

SteveLauC (approved):
Thanks!

SteveLauC:
Let's try if the Cirrus CI will be ok in this PR, it fails at cloning the repo in other PRs😵‍💫

PR #2483, opened 2024-09-05
fcntl adding F_LOG2PHYS* flags.

devnexen:
to get the current device address, in practice its current offset only.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -808,6 +808,17 @@ pub enum FcntlArg<'a> {
     /// fstore_t field fst_bytesalloc.
     #[cfg(apple_targets)]
     F_PREALLOCATE(&'a mut libc::fstore_t),
+    #[cfg(apple_targets)]
+    /// Get disk device information. In practice,
+    /// only the file offset data is set.
+    F_LOG2PHYS(&'a mut libc::off_t),
+    #[cfg(apple_targets)]
+    /// Get disk device information. In practice,
+    /// only the file offset data is set.
+    /// The difference with F_LOG2PHYS is the struct passed
+    /// is used as both IN/OUT as both its l2p_devoffset and
+    /// l2p_contigbytes can be used for more specific queries.
+    F_LOG2PHYS_EXT(&'a mut libc::log2phys),
     // TODO: Rest of flags
 }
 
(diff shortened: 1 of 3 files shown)


src/fcntl.rs
```diff
@@ -808,6 +808,16 @@ pub enum FcntlArg<'a> {
...
+    #[cfg(apple_targets)]
+    /// Get disk device information. In practice,
+    /// only the file offset data is set.
+    F_LOG2PHYS(&'a mut libc::off_t),
+    #[cfg(apple_targets)]
+    /// Get disk device information. In practice,
+    /// only the file offset data is set.
+    /// difference with F_LOG2PHYS is the struct passed
+    /// is used as both IN/OUT.
+    F_LOG2PHYS_EXT(&'a mut libc::off_t),
```
SteveLauC, line 820: The `F_LOG2PHYS` looks good to me, I am confused about the meaning and usage of `F_LOG2PHYS_EXT`:

> > From man page
>
> The F_LOG2PHYS_EXT command operates on the same structure as F_LOG2PHYS but treats it as an in/out:
>
> ```c
>         struct log2phys {
>             u_int32_t l2p_flags;        /* unused so far */
>             off_t     l2p_contigbytes;  /* IN: number of bytes to be queried;
>                                            OUT: number of contiguous bytes allocated at this position */
>             off_t     l2p_devoffset;    /* IN: bytes into file;
>                                            OUT: bytes into device */
>         };
> ```

1. The interface proposed in this PR only exposes that `l2p_devoffset` argument, but unlike `F_LOG2PHYS`, `l2p_contigbytes` is also used, should we expose it?

2. What does "difference with F_LOG2PHYS is the struct passed is used as both IN/OUT" mean? From the unit test, looks like it is similar to `lseek(2)`, if so, why are we using a mutable reference to `off_t` here? Or it is possible that the returned offset could be different from the value passed in?

devnexen: to be honest F_LOGPHYS_EXT was kind of "late addition" I only noticed it just before I commited :)

> The interface proposed in this PR only exposes that l2p_devoffset argument, but unlike F_LOG2PHYS, l2p_contigbytes is also used, should we expose it?

Yes. Better just exposing the struct altogether in this case.

> What does "difference with F_LOG2PHYS is the struct passed is used as both IN/OUT" mean? From the unit test, looks like it is similar to lseek(2), if so, why are we using a mutable reference to off_t here? Or it is possible that the returned offset could be different from the value passed in?

I may need to rephrase it, was just explaining the comment above about those fields being both IN/OUT for `F_LOG2PHYS_EXT`. Value of `l2p_devoffset` can change thus why I add `mut` qualifier.

SteveLauC: > to be honest F_LOGPHYS_EXT was kind of "late addition" I only noticed it just before I commited :)

Got it:)

> Yes. Better just exposing the struct altogether in this case.

Yeah, that makes sense to me.


> I may need to rephrase it, 

Thanks!

> Value of l2p_devoffset can change thus why I add mut qualifier.

Then what's the `IN` semantic here? i.e., the meaning of the value passed in🤔

SteveLauC: Do you mind telling me the functionality of that `_EXT` flag, I think I am still confused about it:<

devnexen: It is the same as LOG2PHYS only l2p_devoffset/l2p_contigbytes are set after call but also can be set before the call whereas LOG2PHYS set l2p_devoffset after the call.

SteveLauC: Though I am still unclear about this flag, but it should not matter as we expose everything needed in this interface, so merging:)

SteveLauC (approved):
Thanks!

PR #2491, opened 2024-09-08
fix: allow safer PosixSpawnFileActions usage

CameronNemo:
Many functions used for PosixSpawnFileActions were demanding fds passed implement the AsFd trait, but because these actions are meant to be taken in the child process, that trait doesn't offer much benefit and actually often leads to the caller needing to do an unsafe operation: instantiating an OwnedFd from a RawFd. All of these functions need a RawFd anyway, so just let the caller pass a RawFd directly rather than have to unsafely create an OwnedFd first, which itself could have unintended side effects like closing the FD in the parent when no parent-side actions were intended.

## What does this PR do

Followup to #2475, which added the API and has not been in a released version yet.

## Checklist:

- [x] A change log has been added if this PR modifies nix's API (NOTE: changelog was added in #2475, and the changelog does not need any updates)

### DIFF
--- a/src/spawn.rs
+++ b/src/spawn.rs
@@ -1,10 +1,6 @@
 //! Safe wrappers around posix_spawn* functions found in the libc "spawn.h" header.
 
-use std::{
-    ffi::CStr,
-    mem,
-    os::unix::io::{AsFd, AsRawFd},
-};
+use std::{ffi::CStr, mem, os::fd::RawFd};
 
 #[cfg(any(feature = "fs", feature = "term"))]
 use crate::fcntl::OFlag;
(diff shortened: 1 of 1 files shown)


SteveLauC:
Thanks for the feedback!

I agree that I/O-safe interfaces do not make sense here.

Regarding your example, you can actually use the `Stdin` and `Stdout` from std as they have `AsFd` implemented, though I think we should use `RawFd` in these interfaces.

src/spawn.rs
```diff
@@ -281,16 +277,16 @@ impl PosixSpawnFileActions {
     /// Add a [dup2](https://pubs.opengroup.org/onlinepubs/9699919799/functions/dup2.html) action. See
     /// [posix_spawn_file_actions_adddup2](https://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_spawn_file_actions_adddup2.html).
     #[doc(alias("posix_spawn_file_actions_adddup2"))]
-    pub fn add_dup2<Fd1: AsFd, Fd2: AsFd>(
+    pub fn add_dup2<Fd1: AsRawFd, Fd2: AsRawFd>(
```
SteveLauC, line 280: Taking a look at the types that implement `AsRawFd`, they are mostly owned file descriptors, e.g., `File`, passing them by value would have them closed within these `add_xxx()` functions, and we don't have an impl like:


```rs
impl<T: AsRawFd> AsRawFd for &T
```


what about:

```rs
add_xxx<Fd: AsRawFd>(_: &AsRawFd)
```

Though one drawback with this interface is that when passing `RawFd`s, we still need to add a layer of reference to it:

```rs
let stdin_raw_fd: RawFd = 0;

add_xxx(&stdin_raw_fd);
```

thoughts on this?

SteveLauC: What about removing the generic type and just using `RawFd`?

SteveLauC (approved):

PR #2492, opened 2024-09-09
style: deny(unexpected_cfgs)

SteveLauC:
## What does this PR do

The default level for this lint is `warn`, let's deny it. Though I am not sure if this is only available since Rust 1.80, let's verify this in our CI.

## Checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,7 @@
 #![allow(clippy::too_long_first_doc_paragraph)]
 #![recursion_limit = "500"]
 #![deny(unused)]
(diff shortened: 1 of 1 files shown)


SteveLauC:
> Though I am not sure if this is only available since Rust 1.80, let's verify this in our CI.

Ok, I get it now, we should get a Clippy warning of unknown lint, but this is allowed in our CI:

> cargo clippy  --target aarch64-apple-darwin --all-targets --all-features -- -D warnings -A unknown-lints

PR #2498, opened 2024-09-12
Add preallocated buffer version of sendmsg

### LINKED ISSUE #2457: Allow `sendmsg` to accept a preallocated `Vec`

howjmay:
closes #2457


src/sys/socket/mod.rs
```diff
@@ -1631,6 +1641,23 @@ pub fn sendmsg<S>(fd: RawFd, iov: &[IoSlice<'_>], cmsgs: &[ControlMessage],
 }
 
 
+/// Send data in scatter-gather vectors to a socket, possibly accompanied
+/// by ancillary data. Optionally direct the message at the given address,
+/// as with sendto.
+/// sendmsg_pre_alloc is the same as sendmsg but it accepts a preallocated
+/// cmsg buffer vector.
```
SteveLauC, line 1648: Let's emphasize that `pre_alloc` part and link the original API:

```suggestion
/// `sendmsg_pre_alloc()` is the same as [`sendmsg()`] but it accepts a preallocated
/// `cmsg` buffer vector.
///
/// Send data in scatter-gather vectors to a socket, possibly accompanied
/// by ancillary data. Optionally direct the message at the given address,
/// as with sendto.
```

src/sys/socket/mod.rs
```diff
@@ -1631,6 +1641,23 @@ pub fn sendmsg<S>(fd: RawFd, iov: &[IoSlice<'_>], cmsgs: &[ControlMessage],
...
+pub fn sendmsg_pre_alloc<S>(fd: RawFd, iov: &[IoSlice<'_>], cmsgs: &[ControlMessage],
+               flags: MsgFlags, addr: Option<&S>, cmsg_buffer: &mut Vec<u8>) -> Result<usize>
+    where S: SockaddrLike
+{
+    let mhdr = pack_mhdr_to_send(&mut cmsg_buffer[..], iov, cmsgs, addr);
```
SteveLauC, line 1653: We need to ensure that `cmsg_buffer` has enough length to accommodate `cmsgs` before passing it to `pack_mhdr_to_send()`

SteveLauC (changes requested):
Hi, thanks for the patch!

Besides that 2 comments, I hope we can also:

1. Have a test for `sendmsg_pre_alloc()`, you can use the test [here](https://docs.rs/nix/latest/nix/sys/socket/fn.sendmsg.html#examples), and do some assertions after the `sendmsg_pre_alloc()` call. We should add this test in `test/sys/test_socket.rs`.
   

3. Document that one can use `sendmsg_pre_alloc()` in the doc of `sendmsg()`:
   
    ```diff
    -/// Allocates if cmsgs is nonempty.
    +/// Allocates if cmsgs is nonempty, use [`sendmsg_pre_alloc()`] if you want to use a pre-allocated buffer.
    ```
4. Add a changelog entry for this PR, please take a look at [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to do this.

SteveLauC:
For the CI failure, you can allow the lint if necessary:

```
error: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
    --> src/sys/socket/mod.rs:1650:64
     |
1650 |                flags: MsgFlags, addr: Option<&S>, cmsg_buffer: &mut Vec<u8>) -> Result<usize>
     |                                                                ^^^^^^^^^^^^ help: change this to: `&mut [u8]`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
     = note: `-D clippy::ptr-arg` implied by `-D warnings`
```

src/sys/socket/mod.rs
SteveLauC: The current impl of `pack_mhdr_to_send()` requires that `cmsg_buffer.len()` equals the length needed by control messages, which may not be true with the pre-alloc API, we need to change this:

https://github.com/nix-rust/nix/blob/7d75bbc9d6cfe91b903c91f39ed8b500ec65b8d1/src/sys/socket/mod.rs#L2044

howjmay: Maybe we can add a if-condition in `sendmsg_pre_alloc()` which will panic if `cmsg_buffer.len() != cmsgs .len()`?

src/sys/socket/mod.rs
howjmay: btw can I rename the new API as `sendmsg_prealloc()`, I feel the underscore there is a little bit of redundant

SteveLauC: > btw can I rename the new API as sendmsg_prealloc()

Sure

src/sys/socket/mod.rs
SteveLauC: I kinda think that would be too limited? There should be no issue if `cmsg_buffer.len()`> `cmsg.len()`, my preferred way is to: check if `cmsg_buffer.len()` is greater than or equal to `cmsg.len()`, if not, return `Err(Errno::ENOBUFS)`. For the `pack_mhdr_to_send()` function, we can add another `cmsg_len` argument, and document that `cmsg_buffer.len()` could be greater than `cmsg_len` due to the this `sendmsg_pralloc()` function.

PR #2499, opened 2024-09-13
Add MAP_SHARED_VALIDATE & MAP_SYNC

sectordistrict:
This pull adds two flags: MAP_SHARED_VALIDATE & MAP_SYNC implemented on https://github.com/rust-lang/libc

The MAP_SHARED_VALIDATE flag was added to linux since 4.15, its a flag that does two things: provide backwards compatibility with old mmap implementations that don't check for unknown flags, and allow new flags to be added.

MAP_SYNC, on the other hand, was added to allow mmap to utilize Direct Accses (DAX) on hardware that support it (non-volatile memory devices) or in general: any ram-shaped filesystem.

both flags are available on both linux and android. (Edit: The android kernel does not support the flags)

Notice:
mmap always returns ENOTSUPP when MAP_SYNC is used on filesystems that are not mounted with DAX enabled (normal everyday filesystems), this means that this kind of test would not be possible on github CI.
 
## Checklist:

### DIFF
--- a/changelog/2499.added.md
+++ b/changelog/2499.added.md
@@ -1,0 +1,1 @@
+Added `MAP_SHARED_VALIDATE` & `MAP_SYNC` for x86.
--- a/changelog/2499.added.md
+++ b/changelog/2499.added.md
@@ -1,1 +1,1 @@
-Added `MAP_SHARED_VALIDATE` & `MAP_SYNC` for x86.
+`MAP_SHARED_VALIDATE` was added for all linux targets. & `MAP_SYNC` was added for linux with the exclusion of mips architecures, and uclibc
--- a/src/sys/mman.rs
+++ b/src/sys/mman.rs
@@ -42,7 +42,7 @@ libc_bitflags! {
         /// Share this mapping. Mutually exclusive with `MAP_PRIVATE`.
         MAP_SHARED;
         /// Force mmap to check and fail on unknown flags. This also enables `MAP_SYNC`.
-        #[cfg(any(target_os = "linux"), not(linux_android))]
+        #[cfg(all(any(target_arch = "x86"), not(linux_android)))]
         MAP_SHARED_VALIDATE;
         /// Create a private copy-on-write mapping. Mutually exclusive with `MAP_SHARED`.
         MAP_PRIVATE;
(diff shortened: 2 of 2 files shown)


sectordistrict:
I was using a local fork of nix that had these updates. Didn't realize I didn't create a branch for this pull request and it's very cluttered, @SteveLauC, I'm thinking we can close this PR and I'll create a new one?

src/sys/mman.rs
```diff
@@ -41,6 +41,9 @@ libc_bitflags! {
         MAP_FILE;
         /// Share this mapping. Mutually exclusive with `MAP_PRIVATE`.
         MAP_SHARED;
+        /// Force mmap to check and fail on unknown flags. This also enables `MAP_SYNC`.
+        #[cfg(all(any(target_arch = "x86"), not(linux_android), not(target_os = "hurd"), not(target_env = "uclibc"), not(freebsdlike)))]
```
SteveLauC, line 45:
```suggestion
        /// Force mmap to check and fail on unknown flags. This also enables `MAP_SYNC`.
        #[cfg(target_os = "linux")]
```

This flag is available on all the Linux targets.

src/sys/mman.rs
```diff
@@ -142,6 +145,9 @@ libc_bitflags! {
         /// Region grows down, like a stack.
         #[cfg(any(linux_android, freebsdlike, target_os = "openbsd"))]
         MAP_STACK;
+        /// Do not write through the page caches, write directly to the file. Used for Direct Access (DAX) enabled file systems.
+        #[cfg(all(any(target_arch = "x86"), not(linux_android), not(target_os = "hurd"), not(target_env = "uclibc"), not(freebsdlike)))]
```
SteveLauC, line 149: This flag is not that lucky,  uClibc does not have it, for the glibc and musl, seems that they all lack the support for the MIPS target, so this should work for now:


```suggestion
        /// Do not write through the page caches, write directly to the file. Used for Direct Access (DAX) enabled file systems.
        // Available on Linux glibc and musl, MIPS* target excluded.
        #[cfg(all(target_os = "linux", not(any(target_arch = "mips", target_arch = "mips64", target_arch = "mipsel", target_arch = "mips64el")), not(target_env = "uclibc")))]
```

changelog/2499.added.md
```diff
@@ -0,0 +1 @@
+Added `MAP_SHARED_VALIDATE` & `MAP_SYNC` for x86.
```
SteveLauC, line 1: When you apply the suggested changes, changelog should also be updated:)

SteveLauC (changes requested):

SteveLauC:
> I'm thinking we can close this PR and I'll create a new one?

You can, of course. But the current PR looks not bad👀

src/sys/mman.rs
sectordistrict: That's great, thanks for the suggestions and apologies again for all the clutter

sectordistrict:
They've all passed! again thanks for the CI magic! @SteveLauC

SteveLauC (approved):
LGTM, thanks for correcting the MIPS `target_arch` value and the patch!

Issue #2436, opened 2024-06-09
Nix 0.30.0 TODO


SteveLauC:
In Nix 0.30.0, most interfaces should have I/O safety adopted, so there are a lot of braking changes included. We should help Nix users migrate to 0.30.0.

For those popular crates, we should probably test the usability before releasing Nix 0.30.0 in case something unexpected happens.

Berrysoft:
May I ask what's the progress of 0.30? It seems that all PRs listed in https://github.com/nix-rust/nix/issues/1750 has been merged.

SteveLauC:
Yeah, let's roll for a new release

https://github.com/nix-rust/nix/issues/2624#issuecomment-2785669307

PR #2502, opened 2024-09-14
Add `getregs()`/`setregs()`/`getregset()`/`setregset()` for Linux/musl/aarch64

orhun:
## What does this PR do

I needed this specific API for packaging my project for Alpine Linux distribution:

- https://github.com/JakWai01/lurk/issues/4
- https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/71833
- https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/71875

I added support for `aarch64-unknown-linux-musl` for `getregs()` function and it builds fine now, but I don't have a way to test it.
Also, I'm not sure if this is the correct way to add support for this architecture - but should be fine!

Let me know what you think and I can try adding support for other functions (such as `setregs()`) as well.

BTW this is my first PR to the `nix` crate so please let me know if I'm doing something wrong :bear:

## Checklist:

### DIFF
--- a/src/sys/ptrace/linux.rs
+++ b/src/sys/ptrace/linux.rs
@@ -348,12 +348,17 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
 /// Get a particular set of user registers, as with `ptrace(PTRACE_GETREGSET, ...)`
 #[cfg(all(
     target_os = "linux",
-    target_env = "gnu",
     any(
-        target_arch = "x86_64",
-        target_arch = "x86",
-        target_arch = "aarch64",
-        target_arch = "riscv64",
+        all(
+            target_env = "gnu",
+            any(
+                target_arch = "x86_64",
+                target_arch = "x86",
+                target_arch = "aarch64",
+                target_arch = "riscv64"
+            )
+        ),
+        all(target_env = "musl", any(target_arch = "aarch64"))
     )
 ))]
 pub fn getregset<S: RegisterSet>(pid: Pid) -> Result<S::Regs> {
@@ -412,8 +417,13 @@ pub fn setregs(pid: Pid, regs: user_regs_struct) -> Result<()> {
 /// [ptrace(2)]: https://www.man7.org/linux/man-pages/man2/ptrace.2.html
 #[cfg(all(
     target_os = "linux",
-    target_env = "gnu",
-    any(target_arch = "aarch64", target_arch = "riscv64")
+    any(
+        all(
+            target_env = "gnu",
+            any(target_arch = "aarch64", target_arch = "riscv64")
+        ),
+        all(target_env = "musl", target_arch = "aarch64")
+    )
 ))]
 pub fn setregs(pid: Pid, regs: user_regs_struct) -> Result<()> {
     setregset::<regset::NT_PRSTATUS>(pid, regs)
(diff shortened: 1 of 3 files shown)


SteveLauC:
> but I don't have a way to test it.

[Here](https://github.com/nix-rust/nix/blob/6cfbf2df4f5fff6720180fcfb5e896e6bfcf7b0c/test/sys/test_ptrace.rs#L304) is a test, maybe you can take a look and try enabling it for musl as well:)

orhun:
I expanded the scope of this PR a bit and enable the tests. Thought I still wonder why Linux/musl/aarch64 wasn't enabled in the initial implementation in #2044

orhun:
The tests seem to be passing 🤔

SteveLauC:
> Thought I still wonder why Linux/musl/aarch64 wasn't enabled in the initial implementation in https://github.com/nix-rust/nix/pull/2044

I think it is because the author's use case does not target `aarch64-unknown-linux-musl`.

> I expanded the scope of this PR a bit and enable the tests.

Looks like `setregs()` is also added, let me edit the PR title (as it will be used as the commit message, we use squash merge).

SteveLauC (approved):
LGTM, thanks for the patch!

orhun:
Awesome, thanks for the merge!

When should we expect a new release? 😇

SteveLauC:
> When should we expect a new release? 😇

Probably not soon, the I/O safety port is still not done, even though we finish it quickly, we still need to [test it](https://github.com/nix-rust/nix/issues/2436). Can you guys use a git dependency, though this would stop you from releasing to crates.io, I am not sure if this will block lurk from being included in Alpine Linux

orhun:
Thanks for letting me know!

> Can you guys use a git dependency, though this would stop you from releasing to crates.io, 

Yeah, weekly alpha/rc releases would be really useful for these cases. Both for accessing the versions quickly and testing it :)

>  I am not sure if this will block lurk from being included in Alpine Linux

I'm not sure as well, just asked them 👀

src/sys/ptrace/linux.rs
```diff
@@ -344,12 +348,17 @@ pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
...
+                target_arch = "x86_64",
+                target_arch = "x86",
+                target_arch = "aarch64",
+                target_arch = "riscv64"
+            )
+        ),
+        all(target_env = "musl", target_arch = "aarch64")
     )
 ))]
 pub fn getregset<S: RegisterSet>(pid: Pid) -> Result<S::Regs> {
```
SteveLauC, line 364: We forgot to enable the `RegisterSet` trait for `linux/aarch64/musl` in this PR, I found this while reviewing #2507, the CI does not catch it as well because we don't have CI for this triple:<

SteveLauC: > I added support for aarch64-unknown-linux-musl for getregs() function and it builds fine now, but I don't have a way to test it.

I am kinda surprised that Rust does not catch this at compile time. 🤔

@orhun, could you please give this a fix and if possible, give this a try with lurk?

SteveLauC: I try to fix this issue and add the CI for `aarch64/musl` in #2510, but it looks like the constants needed by `enum RegisterSetValue` have not been defined in the libc crate, so we need to add them there first.

SteveLauC: PR filed https://github.com/rust-lang/libc/pull/3938

orhun: Ah damn, good catch, these arch stuff goes deep!

Do you want me to test it out with lurk? Is there any other thing that I can help here?

SteveLauC: > these arch stuff goes deep!

Exactly 😪

> Do you want me to test it out with lurk?

Yeah, that would be great! Once #2510 is merged, we can start the test:)

> Is there any other thing that I can help here?

Currently, no, I may ping you when I do need you 😁

Issue #2418, opened 2024-05-25
ptctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, addr, len, invalid_name) returns EFAULT rather than EINVAL under QEMU


SteveLauC:
When I bump the CI Linux runner from Ubuntu 20.04 to 22.04, [this test fails](https://github.com/nix-rust/nix/actions/runs/9235475942/job/25410368942?pr=2417) under QEMU for some architectures, it failed in this [line](https://github.com/nix-rust/nix/blob/1df106962b227257fbe8994b3712a7848befd898/test/sys/test_prctl.rs#L149):

```rs
failures:

---- test_prctl::test_set_vma_anon_name stdout ----
thread 'test_prctl::test_set_vma_anon_name' panicked at 'assertion failed: `(left == right)`
  left: `EFAULT`,
 right: `EINVAL`', test/sys/test_prctl.rs:149:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_prctl::test_set_vma_anon_name
```

asomers:
@devnexen do you know why this test fails on Ubuntu 22.04, or can you figure out?

devnexen:
That might be because of the types of address and name passed to the syscall. Let me see.

devnexen:
After trying, I can t make much sense of why it fails just because of a container update.

SteveLauC:
> Maybe update all but the cross-compiling part ?

Ubuntu 20.04 will be eof in 2025, so we have to migrate to a new version:(

devnexen:
would you consider trying a more advanced version just to see ? e.g. 23.04

SteveLauC:
> would you consider trying a more advanced version just to see ? e.g. 23.04

23.04 is not available, we have 24.04, though it is still in beta.

devnexen:
I barely something else than that or ignoring the test on qemu.

SteveLauC:
> Yeah, we can give it a try

Tried it in this commit https://github.com/nix-rust/nix/pull/2417/commits/3ea403eda9ec6db0539f446a2f9efe085f38845a, no luck, looks like we have to ignore the test?

SteveLauC:
> we have to bump the image

Done in #2632

PR #2509, opened 2024-09-21
Enable CI for Loongarch64

Fearyncess:
## What does this PR do

This PR is the reimplementation of PR #2461. That will enable CI testing for `loongarch64-unknown-linux-gnu`.

## Checklist:

### DIFF
--- a/test/sys/test_prctl.rs
+++ b/test/sys/test_prctl.rs
@@ -99,6 +99,10 @@ mod test_prctl {
         prctl::set_timerslack(original).unwrap();
     }
 
+    // Loongarch need to use a newer QEMU that disabled these PRCTL subcodes/methods.
+    // https://github.com/qemu/qemu/commit/220717a6f46a99031a5b1af964bbf4dec1310440
+    // So we should ignore them when testing in QEMU environments.
+    #[cfg_attr(all(qemu, target_arch = "loongarch64"), ignore)]
     #[test]
     fn test_disable_enable_perf_events() {
         prctl::task_perf_events_disable().unwrap();
@@ -112,6 +116,10 @@ mod test_prctl {
         assert!(no_new_privs);
     }
 
+    // Loongarch need to use a newer QEMU that disabled these PRCTL subcodes/methods
+    // https://github.com/qemu/qemu/commit/220717a6f46a99031a5b1af964bbf4dec1310440
+    // So we should ignore them when testing in QEMU environments.
+    #[cfg_attr(all(qemu, target_arch = "loongarch64"), ignore)]
     #[test]
     fn test_get_set_thp_disable() {
         let original = prctl::get_thp_disable().unwrap();
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ permissions:
 
 env: 
   MSRV: 1.69.0
+  # Rust's Loongarch support merged in 1.71.0
+  MSRV_LOONGARCH: 1.71.0
   RUSTFLAGS: -Dwarnings
 
 jobs:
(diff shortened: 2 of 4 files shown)


SteveLauC:
@Fearyncess: 

I have rewrite https://github.com/Fearyncess/nix/pull/1 for Loongarch64 CI, and [it works](https://github.com/Fearyncess/nix/pull/1/checks). Unfortunately, I face a big problem with the QEMU and testcases below.
Loongarch64 need a newer QEMU after 7.1 version. But before QEMU supported la64, qemu has a commit that has [disabled many PRCTL subcodes](https://github.com/qemu/qemu/commit/220717a6f46a99031a5b1af964bbf4dec1310440) 3 yrs ago. This commit will cause test_prctl failed in some cases, due to these subcodes has been disabled in newer QEMU.

BTW, I guess some commits in QEMU can change some syscalls' behavior, that brings some weird test results... -> https://github.com/nix-rust/nix/issues/2418

test/sys/test_prctl.rs
```diff
@@ -99,6 +99,9 @@ mod test_prctl {
         prctl::set_timerslack(original).unwrap();
     }
 
+    // Loongarch need to use a newer QEMU that disabled these PRCTL subcodes/methods.
+    // So we should ignore them when testing in QEMU environments.
+    #[cfg_attr(all(qemu, target_arch = "loongarch64"), ignore)]
```
SteveLauC, line 104: > Loongarch64 need a newer QEMU after 7.1 version. But before QEMU supported la64, qemu has a commit that has [disabled many PRCTL subcodes](https://github.com/qemu/qemu/commit/220717a6f46a99031a5b1af964bbf4dec1310440) 3 yrs ago. This commit will cause test_prctl failed in some cases, due to these subcodes has been disabled in newer QEMU.

So due to the use of that custom cross image, the Loongarch64 CI is using a QEMU that is different from other CI, this is the reason why this `ignore` is only needed for Loongarch64, right?

Let link the QEMU commit here:

```suggestion
    // Loongarch need to use a newer QEMU that disabled these PRCTL subcodes/methods:
    // https://github.com/qemu/qemu/commit/220717a6f46a99031a5b1af964bbf4dec1310440
    // So we should ignore them when testing in QEMU environments.
    #[cfg_attr(all(qemu, target_arch = "loongarch64"), ignore)]
```

test/sys/test_prctl.rs
```diff
@@ -112,6 +115,9 @@ mod test_prctl {
         assert!(no_new_privs);
     }
 
+    // Loongarch need to use a newer QEMU that disabled these PRCTL subcodes/methods
+    // So we should ignore them when testing in QEMU environments.
```
SteveLauC, line 119: Also, please link the QEMU commit here as well.

SteveLauC (commented):
Please also add this target to the [target support list](https://github.com/nix-rust/nix/blob/master/README.md#supported-platforms) of Nix as a tier2 target.

SteveLauC:
> BTW, I guess some commits in QEMU can change some syscalls' behavior, that brings some weird test results... -> https://github.com/nix-rust/nix/issues/2418

#2418 happens without a QEMU bump, we encountered it when upgrading the Linux CI base image from Ubuntu 20.04 to 22.04🤔

test/sys/test_prctl.rs
Fearyncess: yes. Loongarch64 needs QEMU 7.1 version, newer than the others. And I've added the link in comments.

Fearyncess:
> Please also add this target to the [target support list](https://github.com/nix-rust/nix/blob/master/README.md#supported-platforms) of Nix as a tier2 target.

Added.

SteveLauC (approved):
Looks good, thanks for adding it!

PR #2512, opened 2024-09-23
Fix incorrect posix_spawn* error checking

kosayoda:
The posix_spawn functions do not use `-1` as the sentinel for errors, rather `0` is return on success and every other return value is an error.

Adds a test case to test posix_spawn failure.

## Checklist:


kosayoda:
Wrong branch, sorry!

PR #2515, opened 2024-09-23
P_PIDFD is supported on linux

sectordistrict:
Hello, The id_type flag: `P_PIDFD` is currently configured to only compile for `linux_android` target, but it's also supported on linux and is available on libc.

This pull adds linux as a target os.

### DIFF
--- a/src/sys/wait.rs
+++ b/src/sys/wait.rs
@@ -340,7 +340,7 @@ pub enum Id<'fd> {
     /// If the PID is zero, the caller's process group is used since Linux 5.4.
     PGid(Pid),
     /// Wait for the child referred to by the given PID file descriptor
-    #[cfg(linux_android)]
+    #[cfg(any(linux_android, target_os = "linux"))]
     PIDFd(BorrowedFd<'fd>),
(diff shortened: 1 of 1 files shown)


src/sys/wait.rs
```diff
@@ -340,7 +340,7 @@ pub enum Id<'fd> {
     /// If the PID is zero, the caller's process group is used since Linux 5.4.
     PGid(Pid),
     /// Wait for the child referred to by the given PID file descriptor
-    #[cfg(linux_android)]
```
SteveLauC, line 343: `linux_android` means `any(target_os = "linux", target_os = "android")`, so you should be able to use this flag on Linux, [here](https://docs.rs/nix/latest/nix/sys/wait/enum.Id.html#variant.PIDFd) is the doc of this flag on Linux/glibc.

FYI, `linux_android` is defined [here](https://github.com/nix-rust/nix/blob/9c0bdaca98db0401545e00bd3fc8c6dcb65df557/build.rs#L24).

sectordistrict: Wow, sorry! I was fairly confused, thanks once again @SteveLauC.

PR #2516, opened 2024-09-24
Developer experience - enable all the features when running tests by default

pacak:
## What does this PR do

https://github.com/nix-rust/nix/pull/2091 left developer experience in a bad state: as far as tools are concerned by default - the project is broken. `cargo test` fails with lots of bogus errors, `rust-analyzer` ignores majority of the code, etc. 

This PR tries to address this problem by explicitly enabling all the features for the `nix` crate in dev mode and making sure they will be added manually to the list by removing explicit `--all-features` from a few places.

According to [cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) path-only dev dependencies are allowed and will be ignored when publishing the crate.

With this change in place I can run tests with usual `cargo test`, `rust-analyzer` and other tools work as expected.

## Checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ test: &TEST
   << : *BUILD
   test_script:
     - . $HOME/.cargo/env || true
-    - $TOOL test --target $TARGET --all-features
+    - $TOOL test --target $TARGET
 
 # Test FreeBSD in a full VM.  Test the i686 target too, in the
 # same VM.  The binary will be built in 32-bit mode, but will execute on a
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -77,6 +77,11 @@ parking_lot = "0.12"
 rand = "0.8"
 tempfile = "3.7.1"
 semver = "1.0.7"
+nix = { path = ".", features = ["acct", "aio", "dir", "env", "event", "fanotify",
```
SteveLauC, line 80: IIUC, this dev dependency is for tests (examples/benchmarks), so I guess we cannot remove the `--all-features` flag of `cargo build` and `cargo clippy`?

pacak: Both work for me locally, CI passes.

SteveLauC: Yeah, that's right, but that means it only builds or runs the clippy check for a small set of code, which is not gated with any feature

pacak:
Good point. Will check what's going on with it tomorrow.

SteveLauC (approved):
Thanks!

SteveLauC:
#2518 is merged.

.cirrus.yml
```diff
@@ -57,9 +57,9 @@ task:
   << : *TEST
   i386_test_script:
     - . $HOME/.cargo/env
-    - cargo build --target i686-unknown-freebsd --all-features
```
SteveLauC, line 60: Looks like this was reverted accidentally

pacak: Hmmm... Good catch. I wonder how I managed that. Time to go to sleep I guess.

PR #2525, opened 2024-10-22
unistd: adding linux(glibc)/freebsd close_range.

devnexen:
Allows to close a range of file descriptors, can set close-on-exec on these and/or unsharing (as having its own file descriptors table
 after fork for the calling process).

## Checklist:

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -3992,3 +3992,49 @@ pub fn chflags<P: ?Sized + NixPath>(path: &P, flags: FileFlag) -> Result<()> {
     Errno::result(res).map(drop)
 }
 }
+
+#[cfg(any(
+    all(target_os = "linux", target_env = "gnu"),
+    target_os = "freebsd"
+))]
+#[cfg(feature = "fs")]
+libc_bitflags! {
+    /// Options for close_range()
+    #[cfg_attr(docsrs, doc(cfg(feature = "fs")))]
+    pub struct CloseRangeFlags : c_uint {
+        #[cfg(all(target_os = "linux", target_env = "gnu"))]
+        /// Unshare the file descriptors range before closing them
+        CLOSE_RANGE_UNSHARE;
+        /// Set the close-on-exec flag on the file descriptors range
+        CLOSE_RANGE_CLOEXEC;
+    }
+}
+
+feature! {
+#![feature = "fs"]
+
+/// Close all the file descriptor from a given range.
+/// An optional flag can be applied to modify its behavior.
+#[cfg(any(
+    all(target_os = "linux", target_env = "gnu"),
+    target_os = "freebsd"
+))]
+pub fn close_range<F: std::os::fd::AsFd>(fdbegin: F, fdlast: F, flags: CloseRangeFlags) -> Result<Option<c_int>> {
+    use std::os::fd::AsRawFd;
+
+    let raw = unsafe {
+        Errno::clear();
+        libc::close_range(fdbegin.as_fd().as_raw_fd() as u32, fdlast.as_fd().as_raw_fd() as u32, flags.bits() as i32)
+    };
+    if raw == -1 {
+        if Errno::last_raw() == 0 {
+            Ok(None)
+        } else {
+            Err(Errno::last())
+        }
+    } else {
+        Ok(Some(raw))
+    }
+
+}
+}
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -1391,3 +1391,18 @@ fn test_group_from() {
     assert_eq!(group.gid, group_id);
     assert_eq!(group.name, "wheel");
 }
+
+#[test]
+#[cfg(any(all(target_os = "linux", target_env = "gnu"), target_os = "freebsd"))]
+fn test_close_range() {
+    use tempfile::NamedTempFile;
+    const CONTENTS: &[u8] = b"abcdef123456";
+    let mut tempfile1 = NamedTempFile::new().unwrap();
+    let mut tempfile2 = NamedTempFile::new().unwrap();
+    let mut tempfile3 = NamedTempFile::new().unwrap();
+    tempfile3.write_all(CONTENTS).unwrap();
+    tempfile2.write_all(CONTENTS).unwrap();
+    tempfile1.write_all(CONTENTS).unwrap();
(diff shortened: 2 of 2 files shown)


SteveLauC:
I restarted the Linux/glibc CI as the error is not related to the PR

src/unistd.rs
```diff
@@ -3992,3 +3992,58 @@ pub fn chflags<P: ?Sized + NixPath>(path: &P, flags: FileFlag) -> Result<()> {
...
+    all(target_os = "linux", target_env = "gnu"),
+    target_os = "freebsd"
+))]
+#[cfg(feature = "fs")]
+libc_bitflags! {
+    /// Options for close_range()
+    #[cfg_attr(docsrs, doc(cfg(feature = "fs")))]
+    pub struct CloseRangeFlags : c_uint {
+        #[cfg(all(target_os = "linux", target_env = "gnu"))]
+        /// Unshare the file descriptors range before closing them
```
SteveLauC, line 4006: Considering that [the doc in the man page is kinda misleading](https://lore.kernel.org/all/CABpewhHabbxi=+h+pXbSBnwmbkpaZvCo5Vm3vujMg4UvxhtmkA@mail.gmail.com/T/), let's fix it on our side:

```suggestion
        /// Unshare the file descriptor table, then close the file descriptors specified in the range.
```

src/unistd.rs
```diff
@@ -3992,3 +3992,58 @@ pub fn chflags<P: ?Sized + NixPath>(path: &P, flags: FileFlag) -> Result<()> {
...
+
+#[cfg(any(
+    all(target_os = "linux", target_env = "gnu"),
+    target_os = "freebsd"
+))]
+#[cfg(feature = "fs")]
+libc_bitflags! {
+    /// Options for close_range()
+    #[cfg_attr(docsrs, doc(cfg(feature = "fs")))]
+    pub struct CloseRangeFlags : c_uint {
```
SteveLauC, line 4004: This is interesting, libc defines these 2 constants as unsigned integers. 🤔

src/unistd.rs
```diff
@@ -3992,3 +3992,58 @@ pub fn chflags<P: ?Sized + NixPath>(path: &P, flags: FileFlag) -> Result<()> {
...
+))]
+pub fn close_range<F: std::os::fd::AsFd>(fdbegin: F, fdlast: F, flags: CloseRangeFlags) -> Result<Option<c_int>> {
+    use std::os::fd::AsRawFd;
+
+    let raw = unsafe {
+        Errno::clear();
+
+        cfg_if! {
+            if #[cfg(all(target_os = "linux", target_env = "gnu"))] {
+                libc::syscall(libc::SYS_close_range, fdbegin.as_fd().as_raw_fd() as u32, fdlast.as_fd().as_raw_fd() as u32, flags.bits() as i32)
```
SteveLauC, line 4030: glibc added the wrapper in 2.34, and the libc crate [has it exposed](https://docs.rs/libc/latest/libc/fn.close_range.html), any reason why we are using a raw syscall here?

src/unistd.rs
```diff
@@ -3992,3 +3992,58 @@ pub fn chflags<P: ?Sized + NixPath>(path: &P, flags: FileFlag) -> Result<()> {
...
+feature! {
+#![feature = "fs"]
+
+/// Close all the file descriptor from a given range.
+/// An optional flag can be applied to modify its behavior.
+#[cfg(any(
+    all(target_os = "linux", target_env = "gnu"),
+    target_os = "freebsd"
+))]
+pub fn close_range<F: std::os::fd::AsFd>(fdbegin: F, fdlast: F, flags: CloseRangeFlags) -> Result<Option<c_int>> {
```
SteveLauC, line 4022: I am curious why it returns a `Result<Option<c_int>>`, and the reason behind the code related to `Errno`. From the man page, it returns -1 on error, and 0 on success, looks like it is just a normal syscall👀

SteveLauC (commented):
Hi, I have left some comments.

About this interface, there was an attempt to add it to Nix, #2153, based on the discussion there, this function should be marked `unsafe` given that it is easy to incur a double close. And, I think it should take raw file descriptors, a common use case is to close all the file descriptors: `close_range(0, u32::MAX, 0)`, using I/O-safe types makes the interface hard to use in this case. Your thoughts on this?

test/test_unistd.rs
```diff
@@ -1391,3 +1391,18 @@ fn test_group_from() {
...
+fn test_close_range() {
+    use tempfile::NamedTempFile;
+    const CONTENTS: &[u8] = b"abcdef123456";
+    let mut tempfile1 = NamedTempFile::new().unwrap();
+    let mut tempfile2 = NamedTempFile::new().unwrap();
+    let mut tempfile3 = NamedTempFile::new().unwrap();
+    tempfile3.write_all(CONTENTS).unwrap();
+    tempfile2.write_all(CONTENTS).unwrap();
+    tempfile1.write_all(CONTENTS).unwrap();
+    let areclosed = close_range(tempfile1, tempfile3, CloseRangeFlags::CLOSE_RANGE_CLOEXEC);
```
asomers: In a multithreaded program, which this is, there is no guarantee that tempfile3's file descriptor will be higher than tempfile1's.

asomers (changes requested):
I agree with @SteveLauC on all points.  Additionally, the "fix tests" commit appears to be unrelated to this PR.  If it is indeed unrelated, then please open a separate PR for it, and use a more descriptive commit message.

src/unistd.rs
devnexen: Originally I call the wrapper but I think it is due to the CI when there is old glibc

SteveLauC: Makes sense, some CIs are still using Ubuntu 20.04 due to [this issue](https://github.com/nix-rust/nix/issues/2418), which has glibc 2.31 😮‍💨, perhaps we can try bumping them to see if we are lucky now

devnexen: we can do that if we do not need to bother supporting older releases out in the field. I may come back to it later.

SteveLauC: Yeah, if we use glibc wrapper, then using this interface would require glibc 2.34 or above. We are still not lucky regarding that test: #2533, perhaps we should disable the test under QEMU then 😪

PR #2526, opened 2024-10-23
sys::mman: enables MAP_STACK flag for netbsd too.

devnexen:
Same status as Linux, a no-op, but in the future can be useful for architectures which need this feature.


SteveLauC (approved):
Thanks!

Issue #2506, opened 2024-09-18
Re-use `MultiHeader<()>` in subsequent `recvmmsg()` calls


OliverNChalk:
## Problem

Before call:
> MultiHeaders { items: [mmsghdr { msg_hdr: msghdr { msg_name: 0x1, msg_namelen: 0, msg_iov: 0x5583b86edbe0, msg_iovlen: 1, msg_control: 0x5583b86edc30, msg_controllen: 32, msg_flags: 0 }, msg_len: 0 }], addresses: [core::mem::maybe_uninit::MaybeUninit<()>], _cmsg_buffers: Some([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), msg_controllen: 32 }

After call:
> MultiHeaders { items: [mmsghdr { msg_hdr: msghdr { msg_name: 0x1, msg_namelen: 28, msg_iov: 0x5583b86edbe0, msg_iovlen: 1, msg_control: 0x5583b86edc30, msg_controllen: 32, msg_flags: 0 }, msg_len: 1200 }], addresses: [core::mem::maybe_uninit::MaybeUninit<()>], _cmsg_buffers: Some([32, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 35, 0, 0, 0, 30, 33, 234, 102, 0, 0, 0, 0, 139, 247, 225, 58, 0, 0, 0, 0]), msg_controllen: 32 }

The main thing to call out is that the kernel set the `msg_namelen` value to `28`. This makes re-using MultiHeader for subsequent reads impossible because the kernel will complain with `err=Bad address (os error 14)` due to the fact we have a `msg_name: NULL` and `msg_namelen: 28` in the input.

I'd like to re-use `MultiHeader` for subsequent reads and have confirmed it works if I instead use `MultiHeaders<UnixAddr>`, as we now have space of the address to be written.

## Potential Solution

Letting the caller set the `msglen` would allow us to re-zero it so the kernel doesn't think we have space available for an address when we don't.

### Side Question

Do you know why msg_name renders like this: `msg_hdr.msg_name: 0x1`, I would have thought `0x0` for `NULL`?

SteveLauC:
> I'd like to re-use MultiHeader for subsequent reads

[...]

PR #2528, opened 2024-10-29
Developer experience, cargo fmt

pacak:
## What does this PR do

Right now `cargo fmt` is broken, CI admits it calling format on each individual file: `cargo fmt --all -- --check **/*.rs`. Second commit shows that CI was not actually enforcing formatting across all the files. This breaks workflow of writing code as is and letting your editor to deal with formatting on save or hotkey.

The problem is caused by the `feature!` macro. `rustfmt` does not perform any macro expansion and ignores everything inside of any unknown macro, in fact nothing inside of the `feature!` macro blocks is formatted. This includes scanning for child modules. `feature!` macro is used to lock some parts of the library behind features as well as explicitly document that in rustdoc using nightly feature `doc_cfg`.

This PR proposes a potential alternative - I replaced some of the `feature!`  macro invocation with plain `#[cfg(feature ...)]` invocation and implicit feature documentation using a different nightly feature: `doc_auto_cfg`. Downside (or an upside) is that documentation will contain the actual combination of feature flags and target platforms, for example features for `dir` from `dir` becomes `Non-Redox and dir`.

If we agree that this is a way forward - I'll remove all the remaining uses of `feature!` macro and remove it. It might cause some merge conflicts.

## Checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,7 +92,7 @@
 #![deny(missing_copy_implementations)]
 #![deny(missing_debug_implementations)]
 #![warn(missing_docs)]
-#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 #![deny(clippy::cast_ptr_alignment)]
 #![deny(unsafe_op_in_unsafe_fn)]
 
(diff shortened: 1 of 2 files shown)


SteveLauC:
> The problem is caused by the feature! macro. rustfmt does not perform any macro expansion and ignores everything inside of any unknown macro,

Yeah, this is indeed annoying!

> This PR proposes a potential alternative - I replaced some of the feature! macro invocation with plain #[cfg(feature ...)] invocation

I am not the person who added this macro, but I think the point of using this macro is to avoid repeated cfg attribute invocation.

> implicit feature documentation using a different nightly feature: doc_auto_cfg.

This is not something we want as discussed here: https://github.com/nix-rust/nix/pull/1611#discussion_r1390843837

SteveLauC:
> cargo fmt --all -- --check **/*.rs. Second commit shows that CI was not actually enforcing formatting across all the files. This breaks workflow of writing code as is and letting your editor to deal with formatting on save or hotkey.

Would setting the option `--all **/*.rs` for your format command in your IDE work? 🤔

pacak:
> Would setting the option `--all **/*.rs` for your format command in your IDE work? 🤔

This doesn't solve all the problems: 

1. This PR made it so `rustfmt` sees that formatting in `src/net/if_.rs` was broken and `cargo fmt` managed to fix it.
2. from `rustfmt` point of view - everything inside unknown macro invocations is just token trees it doesn't know how to format.

Plus as I said in https://github.com/nix-rust/nix/issues/2506 - I don't want to make any changes to my global configuration and figuring how to make it local only is still fairly low priority.

> This is not something we want as discussed here: https://github.com/nix-rust/nix/pull/1611#discussion_r1390843837

Hmmm... Banners are verbose, but I think they are accurate. At least in the docs I generated I saw this: `net and (linux_android or bsd or solarish)` and I'm on Linux (not android).

This is an unstable feature and if there are problems with how it works - it's probably a good idea to at least mention it here: https://github.com/rust-lang/rfcs/pull/3631

asomers:
Yes, as @SteveLauC said we don't want to use doc_auto_cfg.  That would make the docs worse than they are now.  And many more people need to read our docs than format our source code, so we should optimize for their experiences.

SteveLauC:
> This is an unstable feature and if there are problems with how it works - it's probably a good idea to at least mention it here: https://github.com/rust-lang/rfcs/pull/3631

Looks like we can combine `doc_auto_cfg` and `doc_cfg_hide`, e.g., the banner of function foo:

```rust
#![cfg_attr(docsrs, feature(doc_auto_cfg))]
#![cfg_attr(docsrs, feature(doc_cfg_hide))]
#![doc(cfg_hide(target_os = "macos"))]

#[cfg(any(target_os = "macos", feature = "foo"))]
pub fn add(left: u64, right: u64) -> u64 {
    left + right
}
```

will be "available on feature foo" rather than "available on macOS and feature foo" because I hide the `target_os = "macos"` at the crate level.

Screenshot 2024-10-30 at 8 41 32 AM

--------

Regarding this PR, I still think the point of macro `feature!()` is to relieve us of writing repeated `cfg(xxx)`, so removing it does not seem good. Though I do agree that it makes formatting hard and the developer experience bad if you have something like "fmt-on-save" open.

PR #2529, opened 2024-10-30
MacOS socket audit token

PaulDance:
## What does this PR do

This adds a new API for `getsockopt` around `libc::LOCAL_PEERTOKEN` following https://github.com/rust-lang/libc/pull/3929. The return type definition is the low-level [`audit_token_t`](https://developer.apple.com/documentation/kernel/audit_token_t) as defined in [`osfmk/mach/message.h`](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/osfmk/mach/message.h#L741) and taken from [`endpoint-sec-sys`](https://crates.io/crates/endpoint-sec-sys). A small test was added for it.

## Checklist:

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -563,7 +563,7 @@ sockopt_impl!(
     libc::SO_KEEPALIVE,
     bool
 );
-#[cfg(any(freebsdlike, apple_targets))]
+#[cfg(freebsdlike)]
 sockopt_impl!(
     /// Get the credentials of the peer process of a connected unix domain
     /// socket.
@@ -575,10 +575,20 @@ sockopt_impl!(
 );
 #[cfg(apple_targets)]
 sockopt_impl!(
+    /// Get the credentials of the peer process of a connected unix domain
+    /// socket.
+    LocalPeerCred,
+    GetOnly,
+    libc::SOL_LOCAL,
+    libc::LOCAL_PEERCRED,
+    super::XuCred
+);
+#[cfg(apple_targets)]
+sockopt_impl!(
     /// Get the PID of the peer process of a connected unix domain socket.
     LocalPeerPid,
     GetOnly,
-    0,
+    libc::SOL_LOCAL,
     libc::LOCAL_PEERPID,
     libc::c_int
 );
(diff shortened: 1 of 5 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -574,14 +574,34 @@ sockopt_impl!(
     super::XuCred
 );
 #[cfg(apple_targets)]
+sockopt_impl!(
+    /// Get the credentials of the peer process of a connected unix domain
+    /// socket.
+    LocalPeerCred,
+    GetOnly,
+    libc::SOL_LOCAL,
```
SteveLauC, line 582: I was thinking change 0 of the above FreeBSD impl to `libc::SOL_LOCAL` as well, but then I realized that the libc crate does not have it defined. 😢

SteveLauC (approved):
Implementation looks good, thanks for the patch!

SteveLauC:
> but does not seem to be run at all for some reason?

Emm, from the CI log, both [amd64](https://github.com/nix-rust/nix/actions/runs/11592351999/job/32274030313?pr=2529) and [aarch64](https://github.com/nix-rust/nix/actions/runs/11592351999/job/32274031023?pr=2529) CIs have this test executed 🤔

src/sys/socket/sockopt.rs
PaulDance: Yes, that's what I did first in order to see how the build in CI would respond, but fell back to this inlining of sorts. This part is really optional though, I just thought it could help clean things up a bit and make them more readable, so it can be grafted away if need be.

PaulDance:
Hey, thanks for the super quick reply!

> Emm, from the CI log, both [amd64](https://github.com/nix-rust/nix/actions/runs/11592351999/job/32274030313?pr=2529) and [aarch64](https://github.com/nix-rust/nix/actions/runs/11592351999/job/32274031023?pr=2529) CIs have this test executed 🤔

Yes, I double-checked and they actually do run! I initially pushed just to see the result of the CI run before launching my whole macOS machinery (I don't work from one) and fixing any upcoming issue, but when I saw that every test job was green on the first attempt, I was immediately doubtful and when searching with my browser's search, I couldn't find the test name in the logs, but that was just due to GitHub's auto load and unload of logs, so the integrated search is really the source of truth here :sweat_smile:

Anyway, if everything is good for you, then it is for me too!

PaulDance:
Super cool, thanks a bunch!

PR #2530, opened 2024-10-30
initialize msg_name with null pointer when msg_name is empty

pacak:
The msg_name field points to a caller-allocated buffer that is used to return the source address if the socket is unconnected. The caller should set msg_namelen to the size of this buffer before this call; upon return from a successful call, msg_namelen will contain the length of the returned address. If the application does not need to know the source address, msg_name can be specified as NULL.

In case we use () msgname_len gets initialized with 0, but pointer to the array with msg_name. This works for the first iteration somehow, but after that kernel sets msgname_len to a non-zero and second invocation with the same MultiHeader fails

## What does this PR do

Fixes https://github.com/nix-rust/nix/issues/2506

## Checklist:


src/sys/socket/mod.rs
```diff
@@ -2002,7 +2002,7 @@ unsafe fn pack_mhdr_to_receive<S>(
     let mut mhdr = mem::MaybeUninit::<msghdr>::zeroed();
     let p = mhdr.as_mut_ptr();
     unsafe {
-        (*p).msg_name = address as *mut c_void;
+        (*p).msg_name = if S::size() == 0 { ptr::null_mut() } else { (*address).as_mut_ptr() as *mut c_void };
```
SteveLauC, line 2005: > (*address).as_mut_ptr() as *mut c_void

Any idea why it would work after reverting the change made in #2041?

SteveLauC: > [not sure if things fixed by https://github.com/nix-rust/nix/pull/2041 are still working after that.](https://github.com/nix-rust/nix/issues/2506#issuecomment-2366939632)

Considering this, gentle ping on @JarredAllen, the author of #2041, could you please take a look at this PR?

## Context 

@pacak is trying to solve #2506, after digging into the previous Nix versions, [he figured out that the issue is possibly introduced by #2041](https://github.com/nix-rust/nix/issues/2506#issuecomment-2366915152):

Screenshot 2024-10-31 at 8 33 28 AM

pacak: > Any idea why it would work after reverting the change made in https://github.com/nix-rust/nix/pull/2041?

As far as I understand `.as_mut_ptr()` gives a null pointer in case of  `()`, `address as *mut c_void` gives a dangling pointer through magic of rustc optimizations.

pacak: I guess just using `.as_mut_ptr()` would be enough as long as all the instances return the right pointer. Explicit checking for size makes it so it works even if there's a bogus instance.

SteveLauC: > As far as I understand .as_mut_ptr() gives a null pointer in case of (), address as *mut c_void gives a dangling pointer through magic of rustc optimizations.

If so

> I guess just using .as_mut_ptr() would be enough as long as all the instances return the right pointer.

then yes


-----

> not sure if things fixed by https://github.com/nix-rust/nix/pull/2041 are still working after that.

I think this PR won't break the fix done by #2041 as there is a unit test guarding it.

BTW, let's document why we should use `.as_mut_tr()` rather than `address as *mut c_void` because it is so easy to revert the fix if one is not aware of such a **subtle** different behavior . 😮‍💨

SteveLauC (approved):
Thanks! Appreciate the patience when you struggle with the format.

src/sys/socket/mod.rs
JarredAllen: Sorry I wasn't around, but thanks for fixing my mistake! I concur with the assessment that this shouldn't break my fix.

SteveLauC: No need to be sorry. 😆

PR #2534, opened 2024-11-11
clippy cleanup: needless as_bytes() in tests `test_sendfile_dragonfly/test_recverr`

devnexen:


test/sys/test_socket.rs
```diff
@@ -2914,7 +2914,7 @@ mod linux_errqueue {
         )
         .unwrap();
         // The sent message / destination associated with the error is returned:
-        assert_eq!(msg.bytes, MESSAGE_CONTENTS.as_bytes().len());
```
SteveLauC, line 2917: Could you please tell me what error you got before this PR, looks like they are equivalent and both return the length in bytes 🤔

devnexen: it is pedantism

```shell
error: needless call to `as_bytes()`
   --> test/test_sendfile.rs:146:40
    |
146 |     assert_eq!(bytes_written as usize, expected_string.as_bytes().len());
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: `len()` can be called directly on strings: `expected_string.len()`
```

same reason for both.

asomers: In the commit message, you should name what error you're fixing.

SteveLauC: This is a clippy cleanup, not a fix, so let's update the PR title a bit since it will be used as the commit message

SteveLauC (approved):
Thanks!

devnexen:
oh you did already :)

Issue #413, opened 2016-09-02
Add syslog APIs


lambda-fairy:
The [syslog interfaces](http://linux.die.net/man/3/syslog) are useful for writing daemons and such. It would be nice to have bindings to them.

lfairy:
I've published a preliminary set of bindings on [my fork](https://github.com/lfairy/nix/blob/syslog/src/syslog.rs). There are a few issues with it so far:

* According to the [Linux documentation](http://linux.die.net/man/3/syslog), the `ident` argument to `openlog()` should have a static lifetime. I have modeled this using `&'static CStr`, which is correct but kind of fiddly.

* The interface of `syslog()` itself *almost* matches that of `printf`, except the former supports an extra code `%m`. I have no idea how to expose this safely.

Given these issues, I think we'd be better off just using a pure-Rust implementation instead.

fiveop:
> I have modeled this using &'static CStr, which is correct but kind of fiddly.

I think is the correct way to expose it safely.

> The interface of syslog() itself almost matches that of printf, except the former supports an extra code %m. I have no idea how to expose this safely.

I have not encountered a good way of dealing with foreign functions with a variable number of arguments. When strings are involved, it only gets messier.

lfairy:
> I have not encountered a good way of dealing with foreign functions with a variable number of arguments. When strings are involved, it only gets messier.

A syntax extension would work, but I think that would be overkill for this crate.

tisonkun:
> I think we'd be better off just using a pure-Rust implementation instead.

A binding is still helpful.

Today, I tried to write syslogs to macos, and encounter this issue https://github.com/python/cpython/issues/91070, while directly call `syslog(3)` can work.

tisonkun:
Or from the other perspective, how to implement syslog lib as in rust?

I made https://github.com/fast/fasyslog/ as a syslog client library. It works well for remote server and if the syslog daemon listens to a UNIX domain socket. But macos no longer works in either way so it's hard to make it work on macos.

SteveLauC:
> A binding is still helpful.

> while directly call syslog(3) can work.

Considering that macOS bug, I agree it would be still helpful to bind them, if we can safely do so.

tisonkun:
Ref -

* https://github.com/python/cpython/blob/v3.13.0/Modules/clinic/syslogmodule.c.h#L112
* https://github.com/python/cpython/blob/v3.13.0/Modules/syslogmodule.c

[...]

PR #2537, opened 2024-11-12
Add syslog supports

tisonkun:
This refers to https://github.com/nix-rust/nix/issues/413.

Hello world works locally:

```
2024-11-12 19:19:52.264557+0800 0x27405d0  Default     0x0                  27738  0    test-aae182a0654f64e3: Hello, nix!
```

I'd like to send this patch early for any necessary alignments.


SteveLauC:
BTW, a changelog is needed, please see [CONTRIBUTING.md](https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests) on how to add one.

tisonkun:
@SteveLauC Thanks for your information! All updated.

> For enum, [libc_enum!](https://github.com/nix-rust/nix/blob/0e4353a368abfcedea4ebe4345cf7604bb61d238/src/macros.rs#L101) will be used. Types Severity and Facility fall into this category.

Updated.

> If a C string is needed (string with a tailing NUL), [trait NixPath](https://docs.rs/nix/latest/nix/trait.NixPath.html) can be used to fill the gap. You can use it for the void openlog(const char *ident argument.

Updated.

> The priority argument of syslog() and vsyslog() should be constructed by ORing enum Severity and enum Facility, we can probably impl std::ops::BitOr<Output = c_int> for these 2 types, but this is not that fault-resilient, perhaps we should define another wrapper type here🤔

Reasonable. `Priority` struct added.

> Regarding the format void syslog(int priority, const char *format, ...); argument, do you think we should implement syslog as a macro similar to println!?

No. As stated at https://github.com/nix-rust/nix/issues/413#issuecomment-2470194025, I think we just simply use Rust's formatting function and pass the formatted string. This is also how Python wraps syslog functionalities.

> A few tests to ensure the wrapper would work should be sufficient.

Included.

tisonkun:
This patch doesn't wrap `vsyslog` and `setlogmark`.

For `vsyslog`, it's a variant of `syslog` for ergonomically coding in C, irrelevant to Rust.

For `setlogmark`, `libc` doesn't export `LOG_MASK` and/or `LOG_UPTO`, make it hard to wrap reasonable code. I temporarily added the wrapper but later revert it. IMO it's less used and we can revisit this function when there is a real-world necessity.

src/syslog.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::Result;
+use std::ffi::OsStr;
+
+/// Logging options of subsequent [`syslog`] calls can be set by calling [`openlog`].
+///
+/// The parameter `ident` is a string that will be prepended to every message. The `logopt`
+/// argument specifies logging options. The `facility` parameter encodes a default facility to be
+/// assigned to all messages that do not have an explicit facility encoded.
+pub fn openlog<P: NixPath + ?Sized>(
+    ident: &P,
```
SteveLauC, line 13: On Linux, the `ident` argument is suggested to have the static lifetime: 

> The argument ident in the call of `openlog()` is probably stored as-is.  Thus, if the string it points to is changed, `syslog()` may start prepending the changed string, and if the string it points to ceases to exist, the results are undefined.  Most portable is to use a string constant.

I think we will have a different interface for Linux in the future:

```rs
pub fn openlog(
    ident: &'static Cstr,
```

src/syslog.rs
```diff
@@ -0,0 +1,146 @@
...
+    ident.with_nix_path(|ident| unsafe {
+        libc::openlog(ident.as_ptr(), logopt.bits(), facility as libc::c_int);
+    })
+}
+
+/// Writes message to the system message logger.
+///
+/// The message is then written to the system console, log files, logged-in users, or forwarded
+/// to other machines as appropriate.
+pub fn syslog<S: AsRef<OsStr> + ?Sized>(
```
SteveLauC, line 26: Can we have an example here? It would be great if it could show how to use Rust's built-in formatter with this function.

src/syslog.rs
SteveLauC: I just realized that using `S: AsRef<OsStr> + ?Sized>` looks exactly right for the `syslog(..., message)` argument, maybe `ident` should also be done similarly, considering they are not paths but strings.

tisonkun:
`LOG_MASK` and `LOG_UPTO` may be easy to add:

```rust
#[macro_export]
macro_rules! LOG_MASK 
{
    ($($arg:tt)*) => (
        (1 << $($arg)*)
    )
}

#[macro_export]
macro_rules! LOG_UPTO 
{
    ($($arg:tt)*) => (
        ((1 << (($($arg)*) + 1)) - 1)
    )
}
```

But I'd like to wait for this patch getting review and perhaps did it in a followup so that we don't push everything at once.

src/syslog.rs
```diff
@@ -0,0 +1,146 @@
...
+    priority: Priority,
+    message: &S,
+) -> Result<()> {
+    let formatter = OsStr::new("%s");
```
SteveLauC, line 30: Neat! This also conforms to a suggestion in the Linux man page:

> Never pass a string with user-supplied data as a format, use the following instead:
>
> ```c
> syslog(priority, "%s", string);
> ```

changelog/2537.added.md
```diff
@@ -0,0 +1 @@
+Add support for `syslog`, `openlog`, `closelog` on `macos`.
```
SteveLauC, line 1: Looks like `closelog()` is not included in the PR

src/syslog.rs
```diff
@@ -0,0 +1,146 @@
...
+        /// Write the message to standard error output as well to the system log.
+        LOG_PERROR;
+    }
+}
+
+libc_enum! {
+    /// Severity levels for log messages.
+    #[repr(i32)]
+    #[non_exhaustive]
+    pub enum Severity {
```
SteveLauC, line 80: The man pages of macOS, FreeBSD, and Linux all use the word "level", what about renaming this type to simply `LogLevel`?

SteveLauC (changes requested):

src/syslog.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Create a new priority from a facility and severity level.
+    pub fn from(severity: Severity, facility: Facility) -> Self {
+        let priority = (facility as libc::c_int) | (severity as libc::c_int);
+        Priority(priority)
+    }
+}
+
+libc_bitflags! {
+    /// Options for system logging.
+    pub struct LogFlags: libc::c_int {
```
SteveLauC, line 60: Thanks for following the [convention](https://github.com/nix-rust/nix/issues/317)!

tisonkun: I didn't even notice this. Just find a name I feel comfortable, lol.

changelog/2537.added.md
tisonkun: Weird .. Perhaps get reverted in the setlogmark func. Let me add it back.

src/syslog.rs
tisonkun: Severity is defined as in:

* https://en.wikipedia.org/wiki/Syslog
* https://datatracker.ietf.org/doc/html/rfc3164
* https://datatracker.ietf.org/doc/html/rfc3164

I'd prefer to use this name.

tisonkun:
Just noticed that we can't use `with_nix_path` here because NixPath has a limit on length which seems not applicable for ident and message.

tisonkun:
@SteveLauC test on unix and updated.

I have an idea to implement `setlogmask`, but not sure if we can export `libc::c_int` or we have other better types:

```rust
pub fn setlogmask(maskpri: libc::c_int) -> libc::c_int {
    let mask = unsafe { libc::setlogmask(maskpri) };
    mask
}

pub fn log_mask(pri: Severity) -> libc::c_int {
    1 << (pri as libc::c_int)
}

pub fn log_upto(pri: Severity) -> libc::c_int {
    (1 << ((pri as libc::c_int) + 1)) - 1
}
```

src/lib.rs
```diff
@@ -200,6 +202,12 @@ feature! {
     pub mod spawn;
 }
 
+#[cfg(unix)]
```
tisonkun, line 205: I've tested on ubuntu and archlinux also.

Perviously add `cfg(any(macos, linux))` and clippy notice me to make it `cfg(unix)`.

test/test.rs
```diff
@@ -42,6 +42,9 @@ mod test_sendfile;
 ))]
 mod test_spawn;
 
+#[cfg(unix)]
```
tisonkun, line 45: Or may be just removed it? Since nix works only for UNIX already.

src/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+use crate::Result;
+use std::ffi::OsStr;
+
+/// Logging options of subsequent [`syslog`] calls can be set by calling [`openlog`].
+///
+/// The parameter `ident` is a string that will be prepended to every message. The `logopt`
+/// argument specifies logging options. The `facility` parameter encodes a default facility to be
+/// assigned to all messages that do not have an explicit facility encoded.
+pub fn openlog<S>(
+    ident: Option<&S>,
```
SteveLauC, line 13: This interface should be good for UNIXs other than Linux, since [Linux would store the `ident` pointer as-is](https://github.com/nix-rust/nix/pull/2537#discussion_r1844889953) (will be used during whole process lifetime), we need to add a separate interface for Linux, and gate this with `#[cfg(not(target_os = "linux"))]`.

src/syslog.rs
```diff
@@ -0,0 +1,193 @@
...
+/// openlog(None::<&str>, flags, Facility::LOG_USER).unwrap();
+/// syslog(Severity::LOG_EMERG, "Hello, nix!").unwrap();
+///
+/// // use `format!` to format the message
+/// let name = "syslog";
+/// syslog(Severity::LOG_EMERG, &format!("Hello, {name}!")).unwrap();
+/// ```
+pub fn syslog<P, S>(priority: P, message: &S) -> Result<()>
+where
+    P: Into<Priority>,
```
SteveLauC, line 57: Nice!

src/syslog.rs
SteveLauC: Makes sense

src/syslog.rs
SteveLauC: And thanks for adding the Option here! I didn't catch it in my first round of review.

SteveLauC (changes requested):

src/syslog.rs
tisonkun: I'd suggest we remove the `ident` param for now.

It's not easy to just add `Option<&'static str>` because it may not end with `nul` and I don't know how to achieve this.

tisonkun: I don't observe the `ident` on macos and linux, perhaps this is because the allocated string is collected 🤣

tisonkun: Maybe `Option<&'static CStr>`. Let me try.

SteveLauC: Nix does not pursue uniform interfaces across OSes, if they are different, let them be different. I suggest keeping the original interface for non-Linux platforms as they are not constrained by this limit and adding a separate interface for Linux

src/syslog.rs
```diff
@@ -0,0 +1,178 @@
...
+        /// delayed until `syslog` is called.
+        ///
+        /// This is the default, and need not be specified.
+        LOG_ODELAY;
+        /// Open the connection to syslogd(8) immediately. Normally the open is delayed until
+        /// the first message is logged. Useful for programs that need to manage the order in
+        /// which file descriptors are allocated.
+        LOG_NDELAY;
+        /// Write the message to standard error output as well to the system log.
+        #[cfg(all(not(target_os = "redox"), not(target_os = "illumos")))]
```
SteveLauC, line 103: Let's do:

```suggestion
        #[cfg(not(any(target_os = "redox", target_os = "illumos")))]
```

This is easier to read and feels more natural.

src/syslog.rs
tisonkun: Could you send a patch to this PR or just edit this PR with the suggested change?

SteveLauC: > Could you send a patch to this PR

Yeah, let me send a commit

tisonkun:
> @SteveLauC test on unix and updated.
> 
> I have an idea to implement `setlogmask`, but not sure if we can export `libc::c_int` or we have other better types:
> 
> ```rust
> pub fn setlogmask(maskpri: libc::c_int) -> libc::c_int {
>     let mask = unsafe { libc::setlogmask(maskpri) };
>     mask
> }
> 
> pub fn log_mask(pri: Severity) -> libc::c_int {
>     1 << (pri as libc::c_int)
> }
> 
> pub fn log_upto(pri: Severity) -> libc::c_int {
>     (1 << ((pri as libc::c_int) + 1)) - 1
> }
> ```
> 
> Also related test ideas - https://stackoverflow.com/questions/56049852/python-2-7-setlogmask0-not-disabling-syslog

@SteveLauC Here is an open question on whether to include `setlogmask` in this PR.

SteveLauC:
Regarding the `setlogmask()` interface, I think we can implement it in the following way:

```rs
/// System log priority mask.
#[repr(transparent)]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct LogMask(libc::c_int);

/// Set the process-wide priority mask to `mask` and return the previous mask 
/// value. 
/// 
/// Calls to `syslog()` with a priority level not set in `mask` are ignored. The 
/// default is to log all priorities.
/// 
/// If the `mask` argument is `None`, the current logmask is not modified, this
/// can be used to query the current log mask.
pub fn setlogmask(mask: Option<LogMask>) -> LogMask {
    let mask = match mask {
        Some(mask) => mask.0,
        None => 0,
    };
    
    let prev_mask = unsafe {
        libc::setlogmask(mask)
    };
    
    LogMask(prev_mask)
}

impl LogMask {
    /// Creates a mask of all priorities up to and including `priority`.
    #[doc(alias("LOG_UPTO"))]
    pub fn up_to(priority: Severity) -> Self {
        todo!()
    }
    
    /// Creates a mask for the specified priority.
    #[doc(alias("LOG_MASK"))]
    pub fn of_priority(priority: Severity)  -> Self {
        todo!()
    }
    
    /// Returns if the mask for the specified `priority` is set.
    pub fn contains(&self, priority: Severity) -> bool {
        todo!()
    }
}

impl std::ops::BitOr for LogMask {
    type Output = Self;
    fn bitor(self, rhs: Self) -> Self::Output {
        Self(self.0 | rhs.0)
    }
}
``` 

Though I think the C macros `LOG_MASK` and `LOG_UPTO` should be added to the libc crate first, so perhaps we can include these things in the next PR?

src/syslog.rs
SteveLauC: See https://github.com/nix-rust/nix/pull/2537/commits/596280b34a64018923bf9eb85e3ec7fb679ced55

tisonkun:
> Though I think the C macros LOG_MASK and LOG_UPTO should be added to the libc crate first, so perhaps we can include these things in the next PR?

Yes. Let's review this PR as is now.

@SteveLauC Thanks for your following commits!

SteveLauC (approved):
Looks good, thanks for the PR, let's merge it! 🚀

src/syslog.rs
AlexTMjugador: I just stumbled upon this review comment when writing a portable interface that relies on `openlog` and I found the different signature among different Unix flavors quite strange, and potentially inappropriate.

`openlog` is a function provided by the environment C library, which may be implemented in different ways even for the same Unix flavor (Linux targets routinely use both `musl` and `glibc` as C library implementations), so changing a function signature just because the target is Linux is something that should be done with lots of caution, as it's likely to not be proper.

Moreover, the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9799919799/functions/syslog.html) says that "the application shall ensure that the string pointed to by `ident` remains valid during the `syslog()` calls that will prepend this identifier". Because we can't really know in this function at what times `syslog()` will be further called, the only safe lifetime for the `ident` parameter is `'static`. And because the string is assumed to be a standard C string with no NUL characters in the middle, it should be `CStr`. Thus, I think the Linux signature of this function is the one that should be used for all Unix flavors.

What do you think about this? Would a PR to tweak this be welcome? :smile:

PR #2548, opened 2024-11-27
feat: impl Into<OwnedFd> for owned fd types

SteveLauC:
## What does this PR do

Implements `Into<OwnedFd> for x` (Actually, it is `From<x> for OwnedFd` due to the clippy rule [clippy::from_over_into](https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into)) for types that are **simply owned file descriptors**.

Types like [Dir](https://docs.rs/nix/latest/nix/dir/struct.Dir.html), do not only contain a file descriptor, so they are not included in this PR.

## Checklist:

### DIFF
--- a/changelog/2548.added.md
+++ b/changelog/2548.added.md
@@ -1,0 +1,1 @@
+Implements `Into<OwnedFd>` for `PtyMaster/Fanotify/Inotify/SignalFd/TimerFd`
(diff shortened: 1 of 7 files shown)


SteveLauC:
Hmm, I did not expect to see hurd CI broken here

PR #2556, opened 2024-12-04
Make sockopt_impl! public

### LINKED ISSUE #577: Ability to add custom socket options

asomers:
This allows users to define their own sockopts, instead of needing to make a PR to Nix for every single one.

Fixes #577

## Checklist:

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -1,18 +1,18 @@
 //! Socket options as used by `setsockopt` and `getsockopt`.
-use super::{GetSockOpt, SetSockOpt};
-use crate::errno::Errno;
+#[cfg(linux_android)]
+use super::SetSockOpt;
 use crate::sys::time::TimeVal;
-use crate::Result;
+#[cfg(linux_android)]
+use crate::{errno::Errno, Result};
 use cfg_if::cfg_if;
 use libc::{self, c_int, c_void, socklen_t};
-#[cfg(apple_targets)]
 use std::ffi::{CStr, CString};
-#[cfg(any(target_os = "freebsd", linux_android))]
 use std::ffi::{OsStr, OsString};
 use std::mem::{self, MaybeUninit};
-#[cfg(any(target_os = "freebsd", linux_android))]
 use std::os::unix::ffi::OsStrExt;
-use std::os::unix::io::{AsFd, AsRawFd};
+#[cfg(linux_android)]
+use std::os::unix::io::AsFd;
+use std::os::unix::io::AsRawFd;
 
 // Constants
 // TCP_CA_NAME_MAX isn't defined in user space include files
@@ -26,8 +26,8 @@ const TCP_CA_NAME_MAX: usize = 16;
 /// This macro aims to help implementing `SetSockOpt` for different socket options that accept
 /// different kinds of data to be used with `setsockopt`.
 ///
-/// Instead of using this macro directly consider using `sockopt_impl!`, especially if the option
-/// you are implementing represents a simple type.
+/// Instead of using this macro directly consider using [`sockopt_impl!`](crate::sockopt_impl),
+/// especially if the option you are implementing represents a simple type.
 ///
 /// # Arguments
 ///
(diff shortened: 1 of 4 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -42,25 +42,33 @@ const TCP_CA_NAME_MAX: usize = 16;
...
-
-                    let res = libc::setsockopt(
+            fn set<F: std::os::unix::io::AsFd>(
+                &self,
+                fd: &F,
+                val: &$ty,
+            ) -> $crate::Result<()> {
+                use $crate::sys::socket::sockopt::Set;
+                let setter: $setter =
+                    $crate::sys::socket::sockopt::Set::new(val);
```
SteveLauC, line 59: Considering that it has already been imported:

```suggestion
                use $crate::sys::socket::sockopt::Set;
                let setter: $setter = Set::new(val);
```

src/sys/socket/sockopt.rs
```diff
@@ -42,25 +42,33 @@ const TCP_CA_NAME_MAX: usize = 16;
...
+                let level = $level;
+                let flag = $flag;
+                let res = unsafe {
+                    libc::setsockopt(
                         fd.as_fd().as_raw_fd(),
```
SteveLauC, line 64: We can import `use std::os::unix::io::AsRawFd` in the scope as well so that users won't need to do this manually:

```
/// Users should be able to define their own sockopts.
mod sockopt_impl {
    use nix::sys::socket::{
        getsockopt, setsockopt, socket, AddressFamily, SockFlag, SockProtocol,
        SockType,
    };
    use std::os::unix::io::AsRawFd; // Users do not need to handle this
```

src/sys/socket/sockopt.rs
```diff
@@ -88,43 +96,52 @@ macro_rules! setsockopt_impl {
...
-                    let mut getter: $getter = Get::uninit();
-
-                    let res = libc::getsockopt(
+            fn get<F: std::os::unix::io::AsFd>(
+                &self,
+                fd: &F,
+            ) -> $crate::Result<$ty> {
+                use $crate::sys::socket::sockopt::Get;
+                let mut getter: $getter =
+                    $crate::sys::socket::sockopt::Get::uninit();
```
SteveLauC, line 112: Same here

src/sys/socket/sockopt.rs
```diff
@@ -88,43 +96,52 @@ macro_rules! setsockopt_impl {
...
+                let level = $level;
+                let flag = $flag;
+                let res = unsafe {
+                    libc::getsockopt(
                         fd.as_fd().as_raw_fd(),
```
SteveLauC, line 117: Same here

src/sys/socket/sockopt.rs
```diff
@@ -1727,15 +1775,14 @@ impl<'a> Set<'a, OsString> for SetOsString<'a> {
 }
 
 /// Getter for a `CString` value.
-#[cfg(apple_targets)]
-#[cfg(feature = "net")]
-struct GetCString<T: AsMut<[u8]>> {
+// Hide the docs, because it's an implementation detail of `sockopt_impl!`
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct GetCString<T: AsMut<[u8]>> {
```
SteveLauC, line 1781: Currently this `Get` impl is directly used in the `sockopt_impl!` implementation:

```rs
#[cfg(apple_targets)]
#[cfg(feature = "net")]
sockopt_impl!(
    /// Get the utun interface name.
    UtunIfname,
    GetOnly,
    libc::SYSPROTO_CONTROL,
    libc::UTUN_OPT_IFNAME,
    CString,
    GetCString<[u8; libc::IFNAMSIZ]> // NOTE
);
```

because we don't have such conversion branches for it:

```rs
macro_rules! sockopt_impl {
    ($(#[$attr:meta])* $name:ident, GetOnly, $level:expr, $flag:path, CString) => {
        sockopt_impl!($(#[$attr])*
                      $name, GetOnly, $level, $flag, bool, $crate::sys::socket::sockopt::GetCString);
    ($(#[$attr:meta])* $name:ident, SetOnly, $level:expr, $flag:path, CString) => {
        sockopt_impl!($(#[$attr])*
                      $name, GetOnly, $level, $flag, bool, $crate::sys::socket::sockopt::SetCString);
    };
```

So we should probably not hide it in the doc, but it is barely used, and we can fix this if becomes commonly used in the future so hiding it still looks good to me

test/sys/test_sockopt.rs
```diff
@@ -1047,3 +1047,54 @@ fn test_ipv6_recv_traffic_class_opts() {
...
     );
 }
+
+/// Users should be able to define their own sockopts.
+mod sockopt_impl {
+    use nix::sys::socket::{
+        getsockopt, setsockopt, socket, AddressFamily, SockFlag, SockProtocol,
+        SockType,
+    };
+    use std::os::unix::io::AsRawFd;
```
SteveLauC, line 1057: This can be removed once we import it in the Get/Set function body

SteveLauC (commented):
Left some comments

src/sys/socket/sockopt.rs
asomers: Oh good catch.  I didn't notice that.

src/sys/socket/sockopt.rs
asomers: done

src/sys/socket/sockopt.rs
asomers: Oh, I guess I didn't need to make that one pub.

SteveLauC (approved):
Thanks!

PR #2557, opened 2024-12-05
Exclude NUL characters from OSStrings returned by getsockopt

asomers:
On FreeBSD, getsockopt appends a single NUL to the returned string.  On Linux, it pads the entire buffer with NULs.  But both of those behaviors may be version-dependent.  Adjust GetOsString to handle any number of NUL characters.

## Checklist:

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -1746,9 +1746,11 @@ impl<T: AsMut<[u8]>> Get<OsString> for GetOsString<T> {
         let len = self.len as usize;
         let mut v = unsafe { self.val.assume_init() };
         if let Ok(cs) = CStr::from_bytes_until_nul(&v.as_mut()[0..len]) {
+            // It's legal for the kernel to return any number of NULs at the
+            // end of the string.  C applications don't care, after all.
             OsStr::from_bytes(cs.to_bytes())
         } else {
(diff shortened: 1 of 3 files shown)


SteveLauC (approved):
> But both of those behaviors may be version-dependent.

So this is a defensive patch, but anyway, it harms nothing:)

src/sys/socket/sockopt.rs
```diff
@@ -1745,7 +1745,13 @@ impl<T: AsMut<[u8]>> Get<OsString> for GetOsString<T> {
     unsafe fn assume_init(self) -> OsString {
         let len = self.len as usize;
         let mut v = unsafe { self.val.assume_init() };
-        OsStr::from_bytes(&v.as_mut()[0..len]).to_owned()
+        if let Ok(cs) = CStr::from_bytes_until_nul(&v.as_mut()[0..len]) {
+            OsStr::from_bytes(cs.to_bytes())
+        } else {
+            // The OS returned a non-NUL-terminated string
+            OsStr::from_bytes(&v.as_mut()[0..len])
+        }
```
SteveLauC, line 1753: Considering that this behavior is version-dependent, i.e., some contributors, with specific kernel versions, may find this check useless, perhaps we should document why we do this here 🤔

PR #2558, opened 2024-12-07
sys::socket adding GetOnly `TCP_FUNCTION_ALIAS` for freebsd.

devnexen:
to complete `TCP_FUNCTION_BLK` as to get the alias name, if existent, of the TCP stack.

### DIFF
--- a/test/sys/test_sockopt.rs
+++ b/test/sys/test_sockopt.rs
@@ -283,7 +283,7 @@ fn test_tcp_congestion() {
 
 #[test]
 #[cfg(target_os = "freebsd")]
-fn test_tcp_function_blk() {
+fn test_tcp_function_blk_alias() {
     use std::ffi::CStr;
 
     let fd = socket(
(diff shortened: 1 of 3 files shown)


SteveLauC (commented):
Can we have a test for this socket option?

devnexen:
sure but then that would be for TCP_FUNCTION_BLK too as there is none for it.

test/sys/test_sockopt.rs
```diff
@@ -298,8 +298,15 @@ fn test_tcp_function_blk() {
...
 
+    let aliastfs = getsockopt(&fd, sockopt::TcpFunctionAlias).unwrap();
+    let aliasname =
+        unsafe { CStr::from_ptr(aliastfs.function_set_name.as_ptr()) };
+    // freebsd default tcp stack has no alias.
+    assert!(aliasname.to_bytes().is_empty());
+
     // We can't know at compile time what options are available.  So just test the setter by a
     // no-op set.
+    // TODO: test if we can load for example BBR tcp stack kernel module.
```
SteveLauC, line 309: We just need to ensure the wrapper works, so it is not necessary to do further tests, but having them harms nothing:)

SteveLauC (approved):
Thanks

test/sys/test_sockopt.rs
devnexen: it is more in case future related flag additions, some of which need this alternative stack to work.

PR #2559, opened 2024-12-11
sys::mman: adding few madvise flags.

devnexen:
- `MADV_PAGEOUT` which reclaim the address range, swapping it out if the page is anonymous or written back to disk if it's backed up by a file.
- `MADV_COLD` which deactivate the address range but as a hint.
- `MADV_WIPEONFORK` after `fork()`, the address range is wiped out to avoid sharing sensible data between processes.
- `MADV_KEEPONFORK` disables `MADV_WIPEONFORK` workflow.

### DIFF
--- a/changelog/2559.added.md
+++ b/changelog/2559.added.md
@@ -1,1 +1,1 @@
-Add `sys::mman::MmapAdvise` `MADV_PAGEOUT`, `MADV_COLD`, `MADV_WIPEONFORK`, `MADV_KEEPONFORK` for Linux target
+Add `sys::mman::MmapAdvise` `MADV_PAGEOUT`, `MADV_COLD`, `MADV_WIPEONFORK`, `MADV_KEEPONFORK` for Linux and Android targets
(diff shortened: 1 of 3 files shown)


devnexen:
@SteveLauC is it ok if, eventually, I do not add new tests there is no previous madvise test.

SteveLauC (approved):
Thanks

SteveLauC:
>  is it ok if, eventually, I do not add new tests there is no previous madvise test.

Yeah, I think so. Having tests is great, but Nix, as a wrapper crate, only needs to ensure the wrappers work. These flags are just constants, not syscalls, they are guarded by the libc crate, so it is ok not to have tests for every constant 😄

changelog/2559.added.md
```diff
@@ -0,0 +1 @@
+Add `sys::mman::MmapAdvise` `MADV_PAGEOUT`, `MADV_COLD`, `MADV_WIPEONFORK`, `MADV_KEEPONFORK` for Linux target
```
SteveLauC, line 1:
```suggestion
Add `sys::mman::MmapAdvise` `MADV_PAGEOUT`, `MADV_COLD`, `MADV_WIPEONFORK`, `MADV_KEEPONFORK` for Linux and Android targets
```

SteveLauC: Let me commit this and merge the PR

SteveLauC (approved):

PR #2562, opened 2024-12-16
Add Sctp protocol and MSG_NOTIFICATION

gnaaman-dn:
## What does this PR do
 - Adds the `Sctp` variant to the `SockProtocol` enum, allowing the user to create SCTP sockets with either `SockType::Stream` or `SockType::SeqPacket`.
 - (Linux Only) Expose `MSG_NOTIFICATION` through `MsgFlags` so that it won't be truncated when converting from `i32`.

`IPPROTO_SCTP` is defined in [RFC-6458](https://datatracker.ietf.org/doc/html/rfc6458), but it is not supported by all OSs.
I've cfg'd-out targets that failed in CI.

`MSG_NOTIFICATION` is Linux-specific, as far as I can tell, but is not exposed through Android's libc, so I've limited it to `linux` and not `linux_android`.

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -213,6 +213,14 @@ pub enum SockProtocol {
     Icmp = libc::IPPROTO_ICMP,
     /// ICMPv6 protocol (ICMP over IPv6)
     IcmpV6 = libc::IPPROTO_ICMPV6,
+    /// SCTP ([sctp(7)](https://man7.org/linux/man-pages/man7/sctp.7.html))
+    #[cfg(any(
+        apple_targets,
+        linux_android,
+        target_os = "freebsd",
+        target_os = "netbsd"
+    ))]
+    Sctp = libc::IPPROTO_SCTP,
 }
 
 impl SockProtocol {
@@ -356,6 +364,9 @@ libc_bitflags! {
                   target_os = "fuchsia",
                   target_os = "freebsd"))]
         MSG_WAITFORONE;
+        /// Indicates that this message is not a user message but an SCTP notification.
+        #[cfg(target_os = "linux")]
+        MSG_NOTIFICATION;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/sys/socket/mod.rs
```diff
@@ -356,6 +366,9 @@ libc_bitflags! {
                   target_os = "fuchsia",
                   target_os = "freebsd"))]
         MSG_WAITFORONE;
+        /// Indicates that this message is not a user message but an SCTP notification.
+        #[cfg(linux)]
```
SteveLauC, line 370: If the alias only contains 1 target, we don't use it:

```suggestion
        #[cfg(target_os = "linux")]
```

src/sys/socket/mod.rs
```diff
@@ -213,6 +213,16 @@ pub enum SockProtocol {
...
     IcmpV6 = libc::IPPROTO_ICMPV6,
+    /// SCTP ([sctp(7)](https://man7.org/linux/man-pages/man7/sctp.7.html))
+    #[cfg(not(any(
+        target_os = "haiku",
+        target_os = "redox",
+        target_os = "solaris",
+        target_os = "openbsd",
+        target_os = "illumos",
+        target_os = "dragonfly",
+    )))]
```
SteveLauC, line 224: If possible, we prefer not to use `cfg(not)` unless the # of targets that do not have this symbol is quite few:

```suggestion
    #[cfg(any(
        apple_targets,
        linux_android,
        target_os = "freebsd",
        target_os = "netbsd"
    ))]
```

I hope I do not miss any targets:)

SteveLauC (commented):
Thanks for the PR, just 2 cfg nits:)

src/sys/socket/mod.rs
gnaaman-dn: Yeah, this makes much more sense as whitelist.
Thank you :)

SteveLauC (approved):
Thanks!

gnaaman-dn:
Thank you @SteveLauC !

PR #2563, opened 2024-12-16
Add from_owned_fd constructor to EventFd

### LINKED ISSUE #2561: EventFd has no conversion from OwnedFd
We pass raw fds into a Rust program via FFI, unix domain sockets, and inheritance. There's no way to use `EventFd` in those situations because I don't see even an unsafe API to convert `OwnedFd` into `EventFd`.

Since it's unsafe, maybe a `From` instance is not appropriate, but perhaps `EventFd::from_fd`?

What do you think?

chadaustin:
Allow creating an `EventFd` from an `OwnedFd` acquired from another process.

Fixes #2561

### DIFF
--- a/changelog/2563.added.md
+++ b/changelog/2563.added.md
@@ -1,0 +1,1 @@
+Added `from_owned_fd` constructor to `EventFd`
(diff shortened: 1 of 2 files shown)


chadaustin:
Let me know if a test would make sense.

SteveLauC (approved):
Thanks

SteveLauC:
> Let me know if a test would make sense.

No need to have tests, this logic here is quite simple:)

PR #2564, opened 2024-12-17
Parse unix address from sockaddr without length

pronebird:
## What does this PR do

I noticed that feeding `*const sockaddr_un` into `SockaddrStorage` without length returns `None`. I had impression that it should actually work.

However extracting `sa_len` manually and feeding it into `SockaddrStrorage` works as expected and I am able to cast it to `UnixAddr` later on by calling `SockaddrStorage::as_unix_addr()`.

```rs
 SockaddrStorage::from_raw(sa_ptr, Some((*sa_ptr).sa_len.into()))
```

## Checklist:

### DIFF
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -1189,6 +1189,9 @@ impl SockaddrLike for SockaddrStorage {
                 libc::AF_ALG => unsafe {
                     AlgAddr::from_raw(addr, l).map(|alg| Self { alg })
                 },
+                libc::AF_UNIX => unsafe {
+                    UnixAddr::from_raw(addr, l).map(|su| Self { su })
+                },
                 #[cfg(feature = "net")]
                 libc::AF_INET => unsafe {
                     SockaddrIn::from_raw(addr, l).map(|sin| Self { sin })
--- a/src/sys/socket/addr.rs
+++ b/src/sys/socket/addr.rs
@@ -2392,6 +2392,19 @@ mod tests {
             assert_eq!(ss.len(), ua.len());
         }
 
+        #[test]
+        fn from_sockaddr_un_named_no_length() {
+            let ua = UnixAddr::new("/var/run/mysock").unwrap();
+            assert!(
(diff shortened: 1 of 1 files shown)


asomers:
What new behavior is enabled, exactly?  A test would be helpful to understand your aim.  Also, you need a changelog entry.

pronebird:
> What new behavior is enabled, exactly? A test would be helpful to understand your aim. Also, you need a changelog entry.

That `SocketaddrStorage` can actually parse a pointer to `sockaddr_un` of unknown length. Hint: the length is actually held inside of the pointer.

I can add a test and change entry.

asomers:
There's a reason that it doesn't work.  It's deliberate.  The purpose of the length argument is to tell Nix how much of the struct is actually initialized.  In C, pointers to sockaddrs are often not fully initialized, or even allocated.  But reading uninitialized memory in Rust is UB.  So that argument is important.

pronebird:
The length argument is `Option<socklen_t>` so it's ok to pass `None` if I don't know the length when a syscall returns a non-null `*sockaddr` that has the `AF_UNIX` family set and a length inside of it.

The else branch is happily converting the whole range of types, i.e: 
https://github.com/nix-rust/nix/blob/master/src/sys/socket/addr.rs#L1184-L1222

`UnixAddr::from_raw()` also happily accepts `None` as `len`:
https://github.com/nix-rust/nix/blob/master/src/sys/socket/addr.rs#L535-L569

So all this functionality is already there and it feels that `SockaddrStorage` just forgot to handle `AF_UNIX` in the else branch.

pronebird:
So it seems that the following is not true for all platforms:

```rs
assert!(
  UnixAddr::size() <= mem::size_of::<libc::sockaddr_storage>() as libc::socklen_t
);
```

pronebird:
I think this could be useful on apple platforms where `sockaddr_un` should fit into `sockaddr_storage` but I don't have time to pursue the completion of this PR.

PR #2567, opened 2024-12-18
tests: disable mknod/mknodat on openbsd too (EPERM errno).

devnexen:


SteveLauC:
You got EPERM even with sudo?🤔

devnexen:
but then I get EINVAL for both.

SteveLauC:
EINVAL means there is probably something wrong with the test 🫠, can you show me the detailed errors?

devnexen:
```shell
---- sys::test_stat::test_mknod stdout ----
thread 'sys::test_stat::test_mknod' panicked at test/sys/test_stat.rs:380:54:
called `Result::unwrap()` on an `Err` value: EPERM

---- sys::test_stat::test_mknodat stdout ----
thread 'sys::test_stat::test_mknodat' panicked at test/sys/test_stat.rs:403:71:
called `Result::unwrap()` on an `Err` value: EPERM
```

```shell
---- sys::test_stat::test_mknod stdout ----
thread 'sys::test_stat::test_mknod' panicked at test/sys/test_stat.rs:380:54:
called `Result::unwrap()` on an `Err` value: EINVAL

---- sys::test_stat::test_mknodat stdout ----
thread 'sys::test_stat::test_mknodat' panicked at test/sys/test_stat.rs:403:71:
called `Result::unwrap()` on an `Err` value: EINVAL
```

PR #2569, opened 2024-12-23
nix::fcntl add FcntlArg::F_READAHEAD for freebsd.

devnexen:
Set/clear the amount of read ahead for the file descriptor.

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -825,6 +825,11 @@ pub enum FcntlArg<'a> {
     /// Both descriptors must reference regular files in the same volume.
     #[cfg(apple_targets)]
     F_TRANSFEREXTENTS(RawFd),
+    /// Set or clear the read ahead (pre-fetch) amount for sequential access or
+    /// disable it with 0 or to system default for any value < 0.
+    /// It manages how the kernel caches file data.
+    #[cfg(target_os = "freebsd")]
+    F_READAHEAD(c_int),
     // TODO: Rest of flags
 }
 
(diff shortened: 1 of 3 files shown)


test/test_fcntl.rs
```diff
@@ -805,3 +805,22 @@ fn test_f_transferextents() {
...
+#[cfg(target_os = "freebsd")]
+#[test]
+fn test_f_readahead() {
+    use nix::fcntl::*;
+
+    let tmp = NamedTempFile::new().unwrap();
+    // With TMPDIR set with UFS, the vnode name is not entered
+    // into the name cache thus path is always empty.
+    // Therefore, we reopen the tempfile a second time for the test
+    // to pass.
```
SteveLauC, line 818: Why would an empty path fail our test?

src/fcntl.rs
```diff
@@ -825,6 +825,10 @@ pub enum FcntlArg<'a> {
     /// Both descriptors must reference regular files in the same volume.
     #[cfg(apple_targets)]
     F_TRANSFEREXTENTS(RawFd),
+    /// Set or clear the read ahead amount for sequential access or
+    /// disable it with 0 or to system default for any value < 0.
+    #[cfg(target_os = "freebsd")]
+    F_READAHEAD(c_int),
```
SteveLauC, line 831: Is this something related to the kernel page cache, something used to control the pre-fetch strategy?

test/test_fcntl.rs
devnexen: ah nvm I think this is copy paste from another test..

src/fcntl.rs
devnexen: it manages file data caching yes. will adjust the comment.

SteveLauC (approved):

SteveLauC:
Thanks

PR #2572, opened 2024-12-27
sys::socket adding sockopt::LingerSec for Apple targets.

devnexen:
```c
#define SO_LINGER 0x0080 /* linger on close if data present (in ticks) */
#define SO_LINGER_SEC 0x1080 /* linger on close if data present (in seconds) */
```

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -372,6 +372,15 @@ sockopt_impl!(
     libc::SO_LINGER,
     libc::linger
 );
+#[cfg(apple_targets)]
+sockopt_impl!(
+    /// Same as `SO_LINGER`, but the duration is in seconds rather than kernel ticks.
+    LingerSec,
+    Both,
+    libc::SOL_SOCKET,
+    libc::SO_LINGER_SEC,
+    libc::linger
+);
 #[cfg(feature = "net")]
 sockopt_impl!(
     #[cfg_attr(docsrs, doc(cfg(feature = "net")))]
--- a/test/sys/test_sockopt.rs
+++ b/test/sys/test_sockopt.rs
@@ -1063,12 +1063,48 @@ fn test_ipv6_recv_traffic_class_opts() {
     );
 }
 
+#[cfg(apple_targets)]
+#[test]
+fn test_linger_sec() {
+    let fd = socket(
+        AddressFamily::Inet,
+        SockType::Stream,
+        SockFlag::empty(),
+        None,
+    )
+    .unwrap();
+
+    let set_linger = libc::linger {
+        l_onoff: 1,
+        l_linger: 1,
+    };
+    setsockopt(&fd, sockopt::LingerSec, &set_linger).unwrap();
+
+    let get_linger = getsockopt(&fd, sockopt::Linger).unwrap();
+    assert_eq!(get_linger.l_linger, set_linger.l_linger * 100);
+}
+
 /// Users should be able to define their own sockopts.
 mod sockopt_impl {
     use nix::sys::socket::{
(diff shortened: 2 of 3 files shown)


test/sys/test_sockopt.rs
```diff
@@ -1097,6 +1097,36 @@ mod sockopt_impl {
...
     }
 
+    #[cfg(apple_targets)]
+    sockopt_impl!(
+        LingerSec,
+        Both,
+        libc::SOL_SOCKET,
+        libc::SO_LINGER_SEC,
+        libc::linger
+    );
```
SteveLauC, line 1107: Looks like you forgot to remove this:)

src/sys/socket/sockopt.rs
```diff
@@ -372,6 +372,16 @@ sockopt_impl!(
     libc::SO_LINGER,
     libc::linger
 );
+#[cfg(apple_targets)]
+sockopt_impl!(
+    /// `SO_LINGER_SEC` on apple targets is the genuine equivalent to
+    /// other platforms `SO_LINGER`. Indeed, the latter uses ticks.
```
SteveLauC, line 378: > is the genuine equivalent to other platforms `SO_LINGER`. 

This makes me feel that macOS does not provide `SO_LINGER` but only `SO_LINGER_SEC`, can we just say something like:

> Same as `SO_LINGETR`, but the duration is in seconds rather than kernel ticks.

test/sys/test_sockopt.rs
devnexen: No I did the same than with the above test.

src/sys/socket/sockopt.rs
devnexen: > 
> This makes me feel that macOS does not provide `SO_LINGER` but only `SO_LINGER_SEC`

oh apple does provide it (see my commit message, that s the C header snapshot).

, can we just say something like:
> 
> > Same as `SO_LINGETR`, but the duration is in seconds rather than kernel ticks.

Sure.

test/sys/test_sockopt.rs
SteveLauC: Emmm, why do we need to define it here?

devnexen: I can try to "fix" both tests if you like ?

SteveLauC: Hmm, this module `mod sockopt_impl {` is for testing the `sockopt_impl!` macro, we made it and related stuff public in #2556 so that users can define socket options without contributing to Nix. We should not put the `fn test_linger_sec()` test here

SteveLauC: Please put the test [before that module](https://github.com/nix-rust/nix/blob/e789a7cd2d6b21d8962de482b9a5da412b66017a/test/sys/test_sockopt.rs#L1065), and if so, I guess we do not need to re-define the `LingerSec` socket option

test/sys/test_sockopt.rs
```diff
@@ -1063,40 +1063,54 @@ fn test_ipv6_recv_traffic_class_opts() {
...
-        let set_linger = libc::linger {
-            l_onoff: 1,
-            l_linger: 42,
-        };
-        setsockopt(&fd, Linger, &set_linger).unwrap();
-
-        let get_linger = getsockopt(&fd, Linger).unwrap();
-        assert_eq!(get_linger.l_linger, set_linger.l_linger);
-    }
-
```
SteveLauC, line 1099: Other changes look good. Could you please not remove this test from this module, it is for user-defined socket options rather than the one provided by Nix

SteveLauC (approved):
Thanks

PR #2573, opened 2024-12-30
sys::socket::sockopt, adding `ExclBind` for solarish systems.

devnexen:
Disallow multiple sockets to bind to an address/port combination.


SteveLauC (approved):
Thanks

PR #2575, opened 2025-01-02
I propose exposing the ```std::os::fd::AsRawFd``` trait method for ```nix::sys::fanotify::Fanotify``` struct

ChrysoliteAzalea:
Hello everyone! I propose exposing the ```AsRawFd``` API for ```Fanotify```struct.

## What does this PR do

This PR forwards the ```as_raw_fd()``` on ```Fanotify``` instance to the underlying ```OwnedFd``` instance. This is necessary to register the struct with Tokio using the [```AsyncFd```](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html) generic struct, which allows working with fanotify asynchronously -- instead of blocking the current thread, or manually polling file descriptors, an ```async``` function can use the ```await``` operator to read events if they are ready, or switch to other task if no events have yet arrived.

## Checklist:

- [yes] I have written necessary tests and rustdoc comments (no rustdoc comments are necessary, ```as_raw_fd()``` is already documented by the standard library)
- [yes] A change log has been added if this PR modifies nix's API


asomers (approved):
LGTM!  Thanks for your contribution.

PR #2579, opened 2025-01-05
Impl syslog::setlogmask

tisonkun:
This follows up https://github.com/nix-rust/nix/pull/2537 and refers to https://github.com/nix-rust/nix/issues/413.

cc @SteveLauC

### DIFF
--- a/src/syslog.rs
+++ b/src/syslog.rs
@@ -138,8 +138,9 @@ impl LogMask {
 
     /// Returns if the mask for the specified `priority` is set.
     pub fn contains(&self, priority: Severity) -> bool {
-        let pri = priority as libc::c_int;
-        (self.0 & (1 << pri)) != 0
+        let priority = Self::of_priority(priority);
+        let and_result = self & priority;
+        and_result != 0
     }
 }
 
--- a/src/syslog.rs
+++ b/src/syslog.rs
@@ -95,9 +95,85 @@ where
     Ok(())
 }
 
+/// Set the process-wide priority mask to `mask` and return the previous mask
+/// value.
+///
+/// Calls to `syslog()` with a priority level not set in `mask` are ignored. The
+/// default is to log all priorities.
+///
+/// If the `mask` argument is `None`, the current logmask is not modified, this
+/// can be used to query the current log mask.
+pub fn setlogmask(mask: Option<LogMask>) -> LogMask {
+    let mask = match mask {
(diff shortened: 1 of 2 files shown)


src/syslog.rs
```diff
@@ -95,11 +95,87 @@ where
...
+    #[doc(alias("LOG_MASK"))]
+    pub fn of_priority(priority: Severity) -> Self {
+        let pri = priority as libc::c_int;
+        Self(1 << pri)
+    }
+
+    /// Returns if the mask for the specified `priority` is set.
+    pub fn contains(&self, priority: Severity) -> bool {
+        let pri = priority as libc::c_int;
+        (self.0 & (1 << pri)) != 0
```
SteveLauC, line 142:
```suggestion
        let priority = Self::of_priority(priority);
        let and_result = self & priority;
        and_result != 0
```

Let's use our wrapper function

tisonkun: Updated.

SteveLauC (approved):
Thanks!

Regarding those 2 syslog C macros, I am working on sending them to the libc crate, let's merge this PR first.

tisonkun:
@SteveLauC May we cut a release recently?

SteveLauC:
Honestly, hard to tell, I plan to maintain open-source projects for only 7 hours per week, and there are still some I/O safety leftovers to clean, if we are lucky, we will have a release in 2 or 3 weeks

tisonkun:
@SteveLauC Good to know. Do you have a list of the release blockers? Or we have a label/milestone/tracking issue to list them. I may try to see if anything I can help.

SteveLauC:
No detailed to-do lists, grep `RawFd` in the codebase, and if a raw fd can be replaced with I/O-safe fd types (OwnedFd, BorrowedFd), do it. [Here](https://github.com/nix-rust/nix/pull/2443/files) is an example PR. 

I think 90% of interfaces are already I/O-safe, the leftovers are harder to do, e.g., the [ScmRights](https://github.com/nix-rust/nix/pull/2494) control message, original I thought the interface included in that PR should work just fine, then I realized you could send file descriptors within the same process (kinda meaningless, but nothing stops you from doing so), in this case, my interface could lead to double-close quite easily. 

Once this is done, do a [crates-test](https://github.com/nix-rust/nix/issues/2436), if we still have time:)

PR #2581, opened 2025-01-11
sys::socket adding Linux packet filtering on ipv4/ipv6/loopback traffics

devnexen:
Respectively `sys::socket::SockProtocol::EthIp`,
`sys::socket::SockProtocol::EthIpv6` and
`sys::socket::SockProtocol::EthLoop` if we want more refined filtering as opposed to the existing `sys::socket::SockProtocol::EthAll` which captures everything.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -209,6 +209,16 @@ pub enum SockProtocol {
     // The protocol number is fed into the socket syscall in network byte order.
     #[cfg(linux_android)]
     EthAll = (libc::ETH_P_ALL as u16).to_be() as i32,
+    #[cfg(linux_android)]
+    /// Packet filter on loopback traffic
+    EthLoop = (libc::ETH_P_LOOP as u16).to_be() as i32,
+    /// Packet filter on IPv4 traffic
+    #[cfg(linux_android)]
+    #[cfg(target_endian = "big")]
+    EthIp = libc::ETH_P_IP,
+    /// Packet filter on IPv6 traffic
+    #[cfg(linux_android)]
+    EthIpv6 = (libc::ETH_P_IPV6 as u16).to_be() as i32,
     /// ICMP protocol ([icmp(7)](https://man7.org/linux/man-pages/man7/icmp.7.html))
     Icmp = libc::IPPROTO_ICMP,
     /// ICMPv6 protocol (ICMP over IPv6)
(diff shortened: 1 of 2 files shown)


src/sys/socket/mod.rs
```diff
@@ -241,6 +247,12 @@ impl SockProtocol {
     #[cfg(apple_targets)]
     #[allow(non_upper_case_globals)]
     pub const KextEvent: SockProtocol = SockProtocol::Icmp; // Matches libc::SYSPROTO_EVENT
+
+    /// Packet filter on IPv4 traffic
+    /// NOTE: placed here due to conflict (little endian arch) with SockProtocol::NetLinkISCI
```
devnexen, line 252: @SteveLauC let me know which direction you prefer me to take here. e.g. using this for all these Eth* flags to avoid future conflicts, not using it all ...

SteveLauC: The current approach looks good, i.e., when add them via associated constants when we see conflicts

SteveLauC, line 252:
```suggestion
    // NOTE: placed here due to conflict (little endian arch) with SockProtocol::NetLinkISCI
```

This should be a code comment, not a doc comment. 

> conflict (little endian arch)

BTW, wouldn't this conflict always occur regardless of the byte order since they have the same value

SteveLauC: > BTW, wouldn't this conflict always occur regardless of the byte order since they have the same value

I was wrong, they only have the same value on Little-endian machines. With big-endian, EthIp is 2048

SteveLauC:
Looks like we need to do conditional compilation to fix the UB issue that fails our CI:

```rust
    /// Packet filter on IPv4 traffic
    // NOTE: placed here due to conflict (little endian arch) with `SockProtocol::NetLinkISCI`
    #[cfg(linux_android)]
    #[allow(non_upper_case_globals)]
    #[cfg(target_endian = "little")]
    pub const EthIp: SockProtocol = unsafe { std::mem::transmute::<i32, SockProtocol>((libc::ETH_P_IP as u16).to_be() as i32) };
    
     /// Packet filter on IPv4 traffic
    #[cfg(linux_android)]
    #[cfg(target_endian = "big")]
    EthIp = (libc::ETH_P_IP as u16).to_be() as i32,
```

src/sys/socket/mod.rs
```diff
@@ -241,6 +247,19 @@ impl SockProtocol {
...
+    #[cfg(linux_android)]
+    #[allow(non_upper_case_globals)]
+    #[cfg(target_endian = "little")]
+    pub const EthIp: SockProtocol = unsafe { std::mem::transmute::<i32, SockProtocol>((libc::ETH_P_IP as u16).to_be() as i32) };
+
+    /// Packet filter on IPv4 traffic
+    #[cfg(linux_android)]
+    #[allow(non_upper_case_globals)]
+    #[cfg(target_endian = "big")]
+    pub const EthIp: i32 = (libc::ETH_P_IP as u16).to_be() as i32;
```
SteveLauC, line 262: Sorry for being unclear, this should be an enum variant

SteveLauC: On big-endian targets, no enum variant has value 2048, so that `std::mem::transmute()` is UB, which is the reason why I said we should make it an enum variant

devnexen: yes it should work now

SteveLauC (approved):
Thanks

PR #2582, opened 2025-01-13
unistd: allow `linkat` to take multiple path types

hdonnay:
## What does this PR do

Add a second type parameter to `linkat` to allow the two paths to have different backing types.

## Checklist:

----

I'm unsure if this warrants additional tests, or if it qualifies as an API change. (I guess it technically is, but it doesn't change the usage in any way.)

LMK if I need to add either, or if there are other instances of this (unintentional(?) type parameter re-use) that I can clean up.

### DIFF
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -929,6 +929,40 @@ fn test_linkat_file() {
 
 #[test]
 #[cfg(not(any(target_os = "redox", target_os = "haiku")))]
+/// This test is the same as [test_linkat_file], but ensures that two different types can be used
+/// as the path arguments.
+fn test_linkat_pathtypes() {
+    use nix::fcntl::AtFlags;
+
+    let tempdir = tempdir().unwrap();
+    let oldfilename = "foo.txt";
+    let oldfilepath = tempdir.path().join(oldfilename);
+
+    let newfilename = "bar.txt";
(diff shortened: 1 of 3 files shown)


asomers:
This is technically a breaking change, even though most users' code will compile just fine.  But it needs a changelog entry, and ideally a test that wouldn't have passed without this patch.

SteveLauC (approved):
LGTM

test/test_unistd.rs
```diff
@@ -927,6 +927,40 @@ fn test_linkat_file() {
...
+
+    // Get file descriptor for base directory
+    let dirfd =
+        fcntl::open(tempdir.path(), fcntl::OFlag::empty(), stat::Mode::empty())
+            .unwrap();
+
+    // Attempt hard link file at relative path
+    linkat(
+        &dirfd,
+        oldfilename,
```
asomers, line 955: Is this supposed to be oldfilepath?  If not, then both file paths are of the same type.
```suggestion
        oldfilepath,
```

asomers (changes requested):

test/test_unistd.rs
hdonnay: Ah, whoops. Will fix.

asomers (approved):

PR #2585, opened 2025-01-18
Fix warnings on nightly regarding "extern" functions

asomers:
"extern" without an ABI is deprecated.  So we need to explicitly use "extern C".


SteveLauC (approved):
Thanks

PR #2591, opened 2025-02-02
docs: doc that signal-unsafe fns can be called after execve

SteveLauC:
## What does this PR do

Resolves the issue reported in [this comment](https://github.com/nix-rust/nix/issues/586#issuecomment-2620124046)

## Checklist:

### DIFF
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -334,9 +334,9 @@ feature! {
 /// # Safety
 ///
 /// In a multithreaded program, only [async-signal-safe] functions like `pause`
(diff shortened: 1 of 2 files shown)


interacsion:
I believe this needs to be added to https://docs.rs/nix/latest/nix/pty/fn.forkpty.html as well

SteveLauC:
> I believe this needs to be added to https://docs.rs/nix/latest/nix/pty/fn.forkpty.html as well

Right, sorry for missing that🫠, thanks for catching it!

PR #2595, opened 2025-02-02
fix: posix_spawn returns errno on error, not -1

SteveLauC:
## What does this PR do

According to the POSIX man page:

> Otherwise, no child process shall be created, the value stored into the variable pointed to by a non-NULL pid is unspecified, and **an error number shall be returned as the function return value to indicate the error**.

`Error::result()` assumes the function returns -1 on error, so it won't work here.

Actually, this is my fault 🤦‍♂️, the implementation in PR #2007 handled this correctly, but my erroneous review comment led the PR in the wrong direction.

Also, since `posix_spawn()` uses `fork()` (can be other syscalls depending on the impl) under the hood, its tests should grab the `FORK_MTX` lock.

## Checklist:

These 2 posx_spawn*() interfaces have not been released, so no changelog needed.

### DIFF
--- a/src/spawn.rs
+++ b/src/spawn.rs
@@ -370,7 +370,7 @@ pub fn posix_spawn<SA: AsRef<CStr>, SE: AsRef<CStr>>(
 ) -> Result<Pid> {
     let mut pid = 0;
 
-    let res = unsafe {
+    let ret = unsafe {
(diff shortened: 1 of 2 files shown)


SteveLauC:
Tests for this fix have been added in https://github.com/nix-rust/nix/pull/2598

PR #2601, opened 2025-02-10
refactor recvmmsg lifetimes

pacak:
## What does this PR do

Currently `recvmmsg` uses a single lifetime to represent several different things:
- preallocated `MultiHdr`
- separately allocated `IoSliceMut`
- iterator used to give access to all the `IoSliceMut`

While correct this makes it impossible to reuse preallocated headers in a loop - modified test fails to compile without this change, but compiles and runs as expected after.

As for the actual reason - I suspect this is a bug in the borrow checker, at least this `RUSTFLAGS=-Zpolonius cargo +nightly check` find no problems with modified test without requiring changes to `recvmmsg`.

Hmm... And there's an unrelated formatting only commit - current formatting is wrong, but ignored by CI.

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -259,8 +259,11 @@ impl SockProtocol {
     #[cfg(linux_android)]
     #[allow(non_upper_case_globals)]
     #[cfg(target_endian = "little")]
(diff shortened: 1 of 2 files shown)


pacak:
Clipy/rustdoc looks very unhappy. Anyway, we've been using this in production since I implemented it. Renamed lifetimes a bit and replaced `'data` with `'iter`- to cover all the references it needs to hold. What needs to be done to get this merged?

PR #2602, opened 2025-02-10
Reexport crate::poll_timeout::PollTimeoutTryFromError;

pacak:
## What does this PR do

`impl TryFrom<i16> for PollTimeout` uses a custom error (`PollTimeoutTryFromError`), but the error type is not public so handling this error is not really possible. All this PR does is making the type public.

Normally something complains when you try to use private type in a public interface, not sure why is it not happening here.

## Checklist:

### DIFF
--- a/src/sys/epoll.rs
+++ b/src/sys/epoll.rs
@@ -1,5 +1,6 @@
 use crate::errno::Errno;
 pub use crate::poll_timeout::PollTimeout as EpollTimeout;
+pub use crate::poll_timeout::PollTimeoutTryFromError as EpollTimeoutTryFromError;
 use crate::Result;
 use libc::{self, c_int};
 use std::mem;
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -2,7 +2,7 @@
 use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
 
 use crate::errno::Errno;
-pub use crate::poll_timeout::PollTimeout;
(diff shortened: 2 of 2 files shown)


SteveLauC:
> not sure why is it not happening here.

Yeah, werid

Thanks!

SteveLauC (approved):

src/sys/epoll.rs
```diff
@@ -1,5 +1,6 @@
 use crate::errno::Errno;
 pub use crate::poll_timeout::PollTimeout as EpollTimeout;
+pub use crate::poll_timeout::PollTimeoutTryFromError;
```
SteveLauC, line 3:
```suggestion
pub use crate::poll_timeout::PollTimeoutTryFromError as EpollTimeoutTryFromError;
```

SteveLauC:
Mind doing it to `poll` too?

https://github.com/nix-rust/nix/blob/003bfe48546d3174d6c107511880756124f40bd7/src/poll.rs#L5

```rust
pub use crate::poll_timeout::PollTimeoutTryFromError;
```

src/sys/epoll.rs
pacak: I could, but I think this would lead in even more confusion. Consider this page: https://docs.rs/nix/latest/nix/sys/epoll/struct.EpollTimeout.html

It mentions `EpollTimeout` 3 times, it mentions the original name (`PollTimeout`) 68 times.

pacak: (did it anyway, it's not _that_ important)

PR #2603, opened 2025-02-11
Enable TcpMaxSeg `setsockopt` option for apple targets

hulthe:
## What does this PR do

Enables the TcpMaxSeg socket option for apple targets. I've tested this on MacOS, not sure about iOS but I imagine it would work just the same.

Here's some code I used to verify that it works:
```rust
use nix::sys::socket::{getsockopt, setsockopt, sockopt::TcpMaxSeg};
use std::{io::Write, net::TcpStream};

fn main() {
    // Open a socket to somewhere
    let mut socket = TcpStream::connect(("127.0.0.1", 1234)).unwrap();
    
    // Set the MSS
    setsockopt(&socket, TcpMaxSeg, &523).unwrap();
    
    // Read the MSS
    let maxseg = getsockopt(&socket, TcpMaxSeg).unwrap();
    println!("maxseg: {maxseg}");
    
    // Send a large amount of data and observe that the segments are not bigger than 523 bytes
    socket.write_all(&[0x77; 2000]).unwrap();
}
```

## Checklist:

### DIFF
--- a/changelog/2603.added.md
+++ b/changelog/2603.added.md
@@ -1,0 +1,1 @@
+Added the TcpMaxSeg `setsockopt` option for apple targets
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -698,7 +698,7 @@ sockopt_impl!(
     u32
 );
 cfg_if! {
-    if #[cfg(linux_android)] {
+    if #[cfg(any(linux_android, apple_targets))] {
         sockopt_impl!(
             /// The maximum segment size for outgoing TCP packets.
             TcpMaxSeg, Both, libc::IPPROTO_TCP, libc::TCP_MAXSEG, u32);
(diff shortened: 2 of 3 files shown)


changelog/2603.added.md
```diff
@@ -0,0 +1 @@
+Added the TcpMaxSeg socket option for apple targets
```
SteveLauC, line 1: Actually, this socket option was available for macOS, but [for getsockopt() only](https://docs.rs/nix/latest/x86_64-apple-darwin/nix/sys/socket/sockopt/struct.TcpMaxSeg.html#impl-GetSockOpt-for-TcpMaxSeg). This PR adds the `setsockopt()` part

hulthe: Ah, right you are. Fixed.

hulthe:
@SteveLauC I updated the test. Had to change it around a bit though since it seems MaxSeg works slightly differently on macos:
1. Can't set MSS on a non-connected socket.
2. Can't set the MSS on a connected socket and then read the same MSS using getsockopt on the other end.

Still I hope the test is good enough :)

SteveLauC (approved):
Thanks!

PR #2610, opened 2025-02-25
feat: add cygwin support

Berrysoft:
## What does this PR do

This PR adds cygwin target support.

Tracking:
- [x] https://github.com/rust-lang/libc/pull/4279
- [x] https://github.com/rust-lang/libc/pull/4309
- [x] https://github.com/rust-lang/libc/pull/4324
- [x] https://github.com/rust-lang/rust/pull/137621

## Checklist:

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -24,7 +24,8 @@ cfg_if! {
         unsafe fn errno_location() -> *mut c_int {
             unsafe { libc::__error() }
         }
-    } else if #[cfg(any(target_os = "android", netbsdlike))] {
+    } else if #[cfg(any(target_os = "android", netbsdlike, target_os = "cygwin"))] {
+        #[cfg_attr(target_os = "cygwin", no_mangle)]
         unsafe fn errno_location() -> *mut c_int {
             unsafe { libc::__errno() }
         }
@@ -280,6 +281,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         ECHRNG => "Channel number out of range",
 
@@ -289,6 +291,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         EL2NSYNC => "Level 2 not synchronized",
 
@@ -298,6 +301,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         EL3HLT => "Level 3 halted",
 
@@ -307,6 +311,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         EL3RST => "Level 3 reset",
 
@@ -316,6 +321,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         ELNRNG => "Link number out of range",
 
@@ -325,6 +331,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         EUNATCH => "Protocol driver not attached",
 
@@ -334,6 +341,7 @@ fn desc(errno: Errno) -> &'static str {
             target_os = "aix",
             target_os = "fuchsia",
             target_os = "emscripten",
+            target_os = "cygwin",
         ))]
         ENOCSI => "No CSI structure available",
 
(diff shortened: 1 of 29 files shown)


Berrysoft:
The CI is broken:(

Berrysoft:
> Weird, let me take a look

Hello, @SteveLauC , if you are still looking at this:) I think it's because the latest version of `once_cell` updates its MSRV to 1.70, larger than the MSRV of `nix`, 1.69.

SteveLauC:
>  I think it's because the latest version of once_cell updates its MSRV

So was this newer version of `once_cell` updated by the `libc` crate?

Berrysoft:
> So was this newer version of `once_cell` updated by the `libc` crate?

I don't think so. `libc` doesn't depend on `once_cell`. It is `tempfile`.

This repo doesn't commit the `Cargo.lock`. In the CI, cargo tries to find the latest version that satisifies the `Cargo.toml`, without considering the `rust-version` field in `Cargo.toml` (this problem has been solved with resolver 3, which is the default resolver in edition 2024).

However, it seems that `once_cell` published a new release recently to decrease the MSRV to 1.65. So a re-run of CI should solve the problem:)

SteveLauC:
Just re-ran the CI, and they failed because `SOCK_PACKET` was deprecated by https://github.com/rust-lang/libc/pull/4267, but since it is used explicitly for testing so I think we can `#[allow(deprecated)]` [here](https://github.com/search?q=repo%3Anix-rust%2Fnix%20SOCK_PACKET&type=code)

SteveLauC:
Weird, the FreeBSD CI was still using once_cell 1.21.0 🤔

Berrysoft:
> Weird, the FreeBSD CI was still using once_cell 1.21.0

Failed 5 days ago. Not today.

SteveLauC:
> Failed 5 days ago. Not today.

Hmm, sorry for missing that:D

Berrysoft:
Thank you @SteveLauC , now the CI passes again.

src/errno.rs
```diff
@@ -3526,3 +3589,271 @@ mod consts {
...
+        ESTRPIPE = libc::ESTRPIPE,
+    }
+
+    impl Errno {
+        pub const EWOULDBLOCK: Errno = Errno::EAGAIN;
+        pub const EDEADLOCK: Errno = Errno::EDEADLK;
+        pub const EOPNOTSUPP: Errno = Errno::ENOTSUP;
+    }
+
+    pub const fn from_i32(e: i32) -> Errno {
```
SteveLauC, line 3730: This function, on other platforms, [was deprecated](https://github.com/nix-rust/nix/blob/e4895d34d9e92b6d5693a7959bd5ef8bc98b53ba/src/errno.rs#L2110-L2115). On Cygwin, maybe we can simply make it private?

test/sys/test_resource.rs
```diff
@@ -10,6 +10,7 @@ use nix::sys::resource::{getrusage, UsageWho};
 /// to put the new soft limit in effect, and then getrlimit() once more to ensure the limits have
 /// been updated.
 #[test]
+#[cfg_attr(target_os = "cygwin", ignore)]
```
SteveLauC, line 13: Why is this test ignored? looks like the symbol it is testing is available on Cygwin

test/test_fcntl.rs
```diff
@@ -243,6 +243,7 @@ fn test_renameat2_noreplace() {
 
 #[test]
 #[cfg(not(target_os = "redox"))]
+#[cfg_attr(target_os = "cygwin", ignore)]
```
SteveLauC, line 246: Why is this ignored? `readlink()` and `readlinkat()` are available on Cygwin

SteveLauC (commented):
Sorry for the late review, the code part looks good. For the tests, some of them are "ignored" even though the corresponding symbols are available, I would like to know why, and if possible, it would be great to comment on the reason there, which would help maintenance a lot! :)

test/sys/test_resource.rs
Berrysoft: `getrlimit` always returns const value for `RLIMIT_NOFILE`.

https://github.com/cygwin/cygwin/blob/8e50c7af7c49819245739d6f626f6fecc394ef7f/winsup/cygwin/resource.cc#L278

test/test_fcntl.rs
Berrysoft: cygwin cannot create a symlink that points to a non-exist path.

dr0p-out: > cygwin cannot create a symlink that points to a non-exist path.

image

are you using MSYS2? afaik Cygwin itself dont have this limitation, see "winsymlinks" in https://cygwin.com/cygwin-ug-net/using-cygwinenv.html, only `nativestrict` mode seems to require the destination to exist (this which is not the default):

image

MSYS2 probably added a new mode and as its default it copies the file instead of creating a symlink, u can change the behavior using `MSYS=winsymlinks:XYZ` though, but this is just MSYS2-specific stuff

image

image

SteveLauC:
Let me fix the unrelated nightly clippy warning

test/test_fcntl.rs
Berrysoft: Seems that you're right. I'll try again.

SteveLauC (approved):
Thanks

ognevny:
@Berrysoft why unistd::sync was not added in this PR? it seems that it's present in [libc](https://github.com/rust-lang/libc/blob/b841a7822732dc10f05d1f9c022a7326dc7b1bc6/src/unix/cygwin/mod.rs#L2091)

Berrysoft:
I think I have checked most features, but sorry if I missed something...

PR #2611, opened 2025-03-02
socket::sockopt adding SOL_FILTER level options for illumos.

devnexen:
Respectively FIL_ATTACH/FIL_DETACH to bind/unbind a filter to a socket to for example delay connection acceptance until data (datafilt) is received.

### DIFF
--- a/src/sys/socket/sockopt.rs
+++ b/src/sys/socket/sockopt.rs
@@ -1,8 +1,8 @@
 //! Socket options as used by `setsockopt` and `getsockopt`.
-#[cfg(linux_android)]
+#[cfg(any(linux_android, target_os = "illumos"))]
 use super::SetSockOpt;
 use crate::sys::time::TimeVal;
-#[cfg(linux_android)]
+#[cfg(any(linux_android, target_os = "illumos"))]
 use crate::{errno::Errno, Result};
 use cfg_if::cfg_if;
 use libc::{self, c_int, c_void, socklen_t};
@@ -11,7 +11,7 @@ use std::ffi::CString;
 use std::ffi::{CStr, OsStr, OsString};
 use std::mem::{self, MaybeUninit};
 use std::os::unix::ffi::OsStrExt;
-#[cfg(linux_android)]
+#[cfg(any(linux_android, target_os = "illumos"))]
 use std::os::unix::io::{AsFd, AsRawFd};
 
 // Constants
@@ -987,25 +987,6 @@ sockopt_impl!(
     libc::SO_ACCEPTFILTER,
     libc::accept_filter_arg
 );
-#[cfg(target_os = "illumos")]
-sockopt_impl!(
-    /// Attach a named filter to this socket to be able to
-    /// defer when anough byte had been buffered by the kernel
-    FilAttach,
-    SetOnly,
-    libc::SOL_FILTER,
-    libc::FIL_ATTACH,
-    SetOsString<'static>
-);
-#[cfg(target_os = "illumos")]
-sockopt_impl!(
-    /// Detach a socket filter previously attached with FIL_ATTACH
-    FilDetach,
-    SetOnly,
-    libc::SOL_FILTER,
-    libc::FIL_DETACH,
-    SetOsString<'static>
-);
 #[cfg(target_os = "linux")]
 sockopt_impl!(
     /// Set the mark for each packet sent through this socket (similar to the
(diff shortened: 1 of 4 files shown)


src/sys/socket/sockopt.rs
```diff
@@ -987,6 +987,25 @@ sockopt_impl!(
...
 );
+#[cfg(target_os = "illumos")]
+sockopt_impl!(
+    /// Attach a named filter to this socket to be able to
+    /// defer when anough byte had been buffered by the kernel
+    FilAttach,
+    SetOnly,
+    libc::SOL_FILTER,
+    libc::FIL_ATTACH,
+    SetOsString<'static>
```
SteveLauC, line 998: I am thinking if this `'static` lifetime would be too restrictive? I do not see a way to specify lifetime parameter using the macro, maybe we can implement the `nix::sys::socket::SetSockOpt` trait manually?

src/sys/socket/sockopt.rs
```diff
@@ -1482,6 +1482,52 @@ impl SetSockOpt for TcpTlsRx {
...
+    type Val = [u8; libc::FILNAME_MAX as usize];
+
+    fn set<F: AsFd>(&self, fd: &F, val: &[u8; libc::FILNAME_MAX as usize]) -> Result<()> {
+        unsafe {
+            let res = libc::setsockopt(
+                fd.as_fd().as_raw_fd(),
+                libc::SOL_FILTER,
+                libc::FIL_ATTACH,
+                val.as_ptr().cast(),
+                val.len() as libc::socklen_t,
```
SteveLauC, line 1502: We are passing the whole buffer to the syscall, is this something expected?

BTW, the below code should mirror the original implementation:

```rust
#[derive(Clone)]
pub struct FilterAttach<'a> {
    _marker: std::marker::PhantomData<&'a ()>
}

impl<'a> FilterAttach<'a> {
    pub fn new() -> Self {
        Self {
            _marker: std::marker::PhantomData,
        }
    }
}

impl<'a> SetSockOpt for FilterAttach<'a> {
    type Val = &'a OsStr;

    fn set<F: std::os::unix::prelude::AsFd>(&self, fd: &F, val: &Self::Val) -> nix::Result<()> {
        unsafe {
            let res = libc::setsockopt(
                fd.as_fd().as_raw_fd(),
                libc::SOL_FILTER,
                libc::FIL_ATTACH,
                val.as_ptr().cast(),
                val.len() as libc::socklen_t,
            );
            Errno::result(res).map(drop)
        }
    }
}
```

src/sys/socket/sockopt.rs
```diff
@@ -20,6 +20,10 @@ use std::os::unix::io::{AsFd, AsRawFd};
 #[cfg(feature = "net")]
 const TCP_CA_NAME_MAX: usize = 16;
 
+#[cfg(target_os = "illumos")]
+#[cfg(feature = "net")]
+const FILNAME_MAX: usize = 32;
```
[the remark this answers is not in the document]
devnexen: yes I tried to avoid the cast ;) but ok

devnexen:
> Can we have tests for this?

things is you need to register the filter(s) in the system prior otherwise it will fail (No such file or directory)

e.g. `soconfig -F httpf httpf prog 2:2:0,2:2:6,26:2:0,26:2:6`

SteveLauC:
> thing is you need to register the filter(s) in the system prior otherwise it will fail (No such file or directory)
>
> e.g. soconfig -F httpf httpf prog 2:2:0,2:2:6,26:2:0,26:2:6 (that is if the kernel module is present)

Having a test that would fail is also acceptable as we only test the binding/interface, not the libc/kernel implementation

SteveLauC:
Hi, I just realized that the `set()` function takes a **reference** to `Self::Val`, which means we do not need to specify the lifetime parameter`&'a` ourselves as long as we remove the `Sized` bound from `SetSockOpt::Val`:

```diff
/// Represents a socket option that can be set.
pub trait SetSockOpt: Clone {
-     type Val: Sized;
+     type Val: ?Sized;

    /// Set the value of this socket option on the given socket.
```

Then, our socket option impl can be:


```rust
#[cfg(target_os = "illumos")]
#[derive(Copy, Clone, Debug)]
/// Attach a named filter to this socket to be able to
/// defer when anough byte had been buffered by the kernel
pub struct FilterAttach;

#[cfg(target_os = "illumos")]
impl SetSockOpt for FilterAttach {
    type Val = OsStr;

    fn set<F: AsFd>(&self, fd: &F, val: &Self::Val) -> Result<()> {
        if val.len() > libc::FILNAME_MAX as usize {
            return Err(Errno::EINVAL);
        }
        unsafe {
            let res = libc::setsockopt(
                fd.as_fd().as_raw_fd(),
                libc::SOL_FILTER,
                libc::FIL_ATTACH,
                val.as_bytes().as_ptr().cast(),
                val.len() as libc::socklen_t,
            );
            Errno::result(res).map(drop)
        }
    }
}

#[cfg(target_os = "illumos")]
#[derive(Copy, Clone, Debug)]
/// Detach a socket filter previously attached with FIL_ATTACH
pub struct FilterDetach;

#[cfg(target_os = "illumos")]
impl SetSockOpt for FilterDetach {
    type Val = OsStr;

    fn set<F: AsFd>(&self, fd: &F, val: &Self::Val) -> Result<()> {
        if val.len() > libc::FILNAME_MAX as usize {
            return Err(Errno::EINVAL);
        }
        unsafe {
            let res = libc::setsockopt(
                fd.as_fd().as_raw_fd(),
                libc::SOL_FILTER,
                libc::FIL_DETACH,
                val.as_bytes().as_ptr().cast(),
                val.len() as libc::socklen_t,
            );
            Errno::result(res).map(drop)
        }
    }
}
```

Note that I removed your use of `to_string_lossy()`, instead, we use [`OsStrExt::as_bytes`](https://doc.rust-lang.org/stable/std/os/unix/ffi/trait.OsStrExt.html).

Then, in our tests, note that I documented the reason why we accept `ENOENT` here, which could help maintenance a lot:

```rust
#[cfg(target_os = "illumos")]
#[test]
fn test_solfilter() {
    use nix::errno::Errno;
    let s = socket(
        AddressFamily::Inet,
        SockType::Stream,
        SockFlag::empty(),
        SockProtocol::Tcp,
    )
    .unwrap();
    let data = std::ffi::OsStr::new("httpf");
    let attach = sockopt::FilAttach;
    let detach = sockopt::FilterDetach;

    // These 2 options won't work unless the needed kernel model is installed:
    // https://github.com/nix-rust/nix/pull/2611#issuecomment-2750237782
    //
    // So we only test the binding here
    assert_eq!(Err(Errno::ENOENT), setsockopt(&s, attach, &data));
    assert_eq!(Err(Errno::ENOENT), setsockopt(&s, detach, &data));
}
```

SteveLauC (approved):
Thanks

Issue #2389, opened 2024-04-28
MSRV 1.77.0 TODO


asomers:
Whenever we raise the MSRV to 1.77.0 for any reason, we should
[ ] Replace the memoffset::offset_of! macro with the builtin macro of the same name (https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#offset_of)

SteveLauC:
* [ ] for the following code snippet in `build.rs`, use `cargo::` rather than `cargo:` in `build.rs` (new syntax introduced in 1.77)

```rs
    // Below are Nix's custom cfg values that we need to let the compiler know
    println!("cargo:rustc-check-cfg=cfg(apple_targets)");
    println!("cargo:rustc-check-cfg=cfg(bsd)");
    println!("cargo:rustc-check-cfg=cfg(bsd_without_apple)");
    println!("cargo:rustc-check-cfg=cfg(linux_android)");
    println!("cargo:rustc-check-cfg=cfg(freebsdlike)");
    println!("cargo:rustc-check-cfg=cfg(netbsdlike)");
    println!("cargo:rustc-check-cfg=cfg(solarish)");
    println!("cargo:rustc-check-cfg=cfg(fbsd14)");
    println!("cargo:rustc-check-cfg=cfg(qemu)");
```

PR #2617, opened 2025-03-23
NEEDS MSRV >=1.77 | chore: replace `memoffset` with std API

paolobarbolini:
## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ libc = { version = "0.2.166", features = ["extra_traits"] }
 bitflags = "2.3.3"
 cfg-if = "1.0"
 pin-utils = { version = "0.1.0", optional = true }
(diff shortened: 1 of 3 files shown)


SteveLauC (commented):
Nix is used by a lot of crates, bumping MSRV should be done carefully. Any reason why you want to do this?

SteveLauC:
>  but I'm also ok if you decide to stay on and order version for longer and this gets merged some time later.

I do not have an MSRV bumping strategy in my mind, but I think we should keep it close to the libc crate's MSRV, so I will stay, thanks:)

xtqqczze:
https://github.com/nix-rust/nix/issues/2389 covers removal of the `memoffset::offset_of!` macro.

SteveLauC:
Closing the PR as it has been tracked by https://github.com/nix-rust/nix/issues/2389

PR #2619, opened 2025-03-24
doc(socket): add a note on how to enable NONBLOCK / CLOEXEC on macOS

kskalski:
## What does this PR do
Adds rustdoc comment on `SockFlag` options for `NONBLOCK` and `CLOEXEC` hinting on how to achieve that functionality on macOS

See discussion in https://github.com/nix-rust/nix/pull/863

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -290,6 +290,9 @@ libc_bitflags! {
     /// Additional socket options
     pub struct SockFlag: c_int {
         /// Set non-blocking mode on the new socket
+        ///
+        /// Note: macOS supports `NONBLOCK` through separate call to `fcntl`,
+        /// i.e. `nix::fcntl::fcntl(fd, F_SETFL(O_NONBLOCK))`
         #[cfg(any(linux_android,
                   freebsdlike,
                   netbsdlike,
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -292,13 +292,16 @@ libc_bitflags! {
         /// Set non-blocking mode on the new socket
         ///
         /// Note: macOS supports `NONBLOCK` through separate call to `fcntl`,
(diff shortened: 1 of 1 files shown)


src/sys/socket/mod.rs
```diff
@@ -290,12 +290,18 @@ libc_bitflags! {
     /// Additional socket options
     pub struct SockFlag: c_int {
         /// Set non-blocking mode on the new socket
+        ///
+        /// Note: macOS supports `NONBLOCK` through separate call to `fcntl`,
+        /// e.g. `nix::fcntl::fcntl(fd, F_SETFL(O_NONBLOCK))`
```
SteveLauC, line 295: The variant is not available on macOS, so macOS users won't see this hint when reading Nix's doc.

BTW, do we really need to document this? Googling "set file descriptor non-blocking" directly leads me to this: https://stackoverflow.com/questions/27266346/how-to-set-file-descriptor-non-blocking

SteveLauC: I do get what you said here, but Nix should not do such cross-platform documentation, I do not remember such things ever happened since I maintained this lib.

Regarding Nix's documentation, we try to be well-documented, but platform's man pages are the right place that you should refer to when figuring out something, [`$ man 2 socket`](https://keith.github.io/xcode-man-pages/socket.2.html). And if you believe what Apple man pages say, `SOCK_NONBLOCK` is a symbol that macOS does not provide. (I say so because Apple man pages are usually outdated, you need to search darwin's source code if you wanna do serious research)

I think I will document why Nix's [`socket()`](https://docs.rs/nix/latest/nix/sys/socket/fn.socket.html) function has an extra `flags` argument, this can confuse users when they read the doc.

kskalski: Yap, those are valid points. From the perspective of `nix` user I think some hint on "why" specific flag variants are selectively available only on some platforms would be useful, but as you say it might belong to the level of flags enum or somewhere else.

PR #2621, opened 2025-03-29
socket::sockopt AttachReusePortCbpf for Linux addition.

devnexen:
Set a BPF program to the socket to set how packets are handled within the reuseport group.


SteveLauC (approved):
Thanks!

PR #2629, opened 2025-04-11
sys::prctl: adding calls about core scheduling for isolation based on

devnexen:
cookie.

## Checklist:

### DIFF
--- a/src/sys/prctl.rs
+++ b/src/sys/prctl.rs
@@ -7,6 +7,7 @@
 
 use crate::errno::Errno;
 use crate::sys::signal::Signal;
+use crate::unistd::Pid;
 use crate::Result;
 
 use libc::{c_int, c_ulong, c_void};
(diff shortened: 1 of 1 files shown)


SteveLauC (changes requested):
These APIs do not seem right? From the kernel [commit message](https://lore.kernel.org/lkml/20210422123309.039845339@infradead.org/), they should look like this:

```c
prctl(PR_SCHED_CORE, PR_SCHED_CORE_GET, tgtpid, pidtype, &cookie)
prctl(PR_SCHED_CORE, PR_SCHED_CORE_CREATE, tgtpid, pidtype, NULL)
prctl(PR_SCHED_CORE, PR_SCHED_CORE_SHARE_TO, tgtpid, pidtype, NULL)
prctl(PR_SCHED_CORE, PR_SCHED_CORE_SHARE_FROM, srcpid, pidtype, NULL)
```

The third argument should always be a process ID, and the fourth argument should be a kernel enum [`pid_type`](https://github.com/torvalds/linux/blob/7cdabafc001202de9984f22c973305f424e0a8b7/include/linux/pid_types.h#L5):

```c
enum pid_type {
	PIDTYPE_PID,
	PIDTYPE_TGID,
	PIDTYPE_PGID,
	PIDTYPE_SID,
	PIDTYPE_MAX,
};
```

For the `PR_SCHED_CORE_GET` sub-command, the cookie should be of type `unsigned long long` according to the [kernel doc](https://docs.kernel.org/admin-guide/hw-vuln/core-scheduling.html):

> userspace pointer to an unsigned long long for storing the cookie returned by PR_SCHED_CORE_GET command. Should be 0 for all other commands.

SteveLauC:
I think we should:

1. Add `pid_type` to the libc crate
2. Expose the full APIs, e.g., for the `PR_SCHED_CORE_GET` sub-command, it should be something like:

```rust
pub fn sched_core_get(tgtpid: Pid, pid_type: PidType) -> Result<libc::c_ulonglong>;
```

PR #2631, opened 2025-04-14
Remove Copy from PollFd

### LINKED ISSUE #2630: PollFd implementing Copy is a footgun
I think this code speaks for itself:
```rust
fn example(file: &std::fs::File) -> io::Result<Option<()>> {
    use nix::poll::{poll, PollFd, PollFlags, PollTimeout};
    use std::os::fd::AsFd;

    let poll_fd = PollFd::new(file.as_fd(), PollFlags::POLLIN);

    let mut poll_fds = [poll_fd];                          // <<<< Copy happens here

    let poll_res = poll(&mut poll_fds, PollTimeout::NONE)?;
    if poll_res == 0 {
        // Timeout
        return Ok(None);
    }
    assert_eq!(poll_res, 1);

    // This fails because poll_fd is not the object that was actually updated
    //assert_eq!(poll_fd.any(), Some(true));

    // You have to access the item in the array
    assert_eq!(poll_fds[0].any(), Some(true));

    Ok(Some(()))
}
```

The fact that `PollFd` [implements `Copy`](https://github.com/nix-rust/nix/blob/b561476e1da958ef249ce203f348c56d9fe24627/src/poll.rs#L17) makes it easy to make this mistake.

Normally, Rust doesn't let you make mistakes like this. If `PollFd` _didn't_ implement `Copy`, rust would complain if you tried to write this code. You'd have to call `.clone()` and then there would be an explicit clue that you're not operating on the same object.

```
error[E0382]: borrow of moved value: `poll_fd`
  --> src/main.rs:23:16
   |
11 |     let poll_fd = PollFd::new(file.as_fd(), PollFlags::POLLIN);
   |         ------- move occurs because `poll_fd` has type `PollFd<'_>`, which does not implement the `Copy` trait
12 |
13 |     let mut poll_fds = [poll_fd];                          // <<<< Copy happens here
   |                         ------- value moved here
...
23 |     assert_eq!(poll_fd.any(), Some(true));
   |                ^^^^^^^ value borrowed here after move
   |
help: consider cloning the value if the performance cost is acceptable
   |
13 |     let mut poll_fds = [poll_fd.clone()];                          // <<<< Copy happens here
   |                                ++++++++
```

JonathonReinhart:
## What does this PR do

`PollFd` implementing `Copy` makes it easy to accidentally refer to the wrong object after putting one into an array.

This PR removes `Copy` from `PollFd`. This fixes #2630.

## Checklist:

### DIFF
--- a/changelog/2631.removed.md
+++ b/changelog/2631.removed.md
@@ -1,0 +1,1 @@
+Removed the `Copy` trait from `PollFd`
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -11,10 +11,10 @@ use crate::Result;
 /// [`ppoll`](fn.ppoll.html) functions to specify the events of interest
 /// for a specific file descriptor.
 ///
-/// After a call to `poll` or `ppoll`, the events that occurred can be
-/// retrieved by calling [`revents()`](#method.revents) on the `PollFd`.
+/// After a call to `poll` or `ppoll`, the events that occurred can be retrieved by calling
+/// [`revents()`](#method.revents) on the `PollFd` object from the array passed to `poll`.
 #[repr(transparent)]
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct PollFd<'fd> {
     pollfd: libc::pollfd,
     _fd: std::marker::PhantomData<BorrowedFd<'fd>>,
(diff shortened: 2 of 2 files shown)


JonathonReinhart:
I'll wait for a positive signal from maintainers that this is a good solution before I add a changelog entry.

asomers:
This looks like a good approach to me.  Could you please add a doctest that shows how to properly combine `PollFd::new`, `poll`, and `PollFd::any` (or `PollFd::events` or similar)?  You could probably copy the current docs for `PollFd::new` and move them into the `PollFd` structure.

JonathonReinhart:
Thanks @asomers. I believe This is ready for review and merge.

I'm not sure why [tier3 (i686-unknown-hurd-gnu)](https://github.com/nix-rust/nix/actions/runs/14475176027/job/40598974266?pr=2631#logs) is failing; it appears to be unrelated to this change.

SteveLauC:
> I'm not sure why [tier3 (i686-unknown-hurd-gnu)](https://github.com/nix-rust/nix/actions/runs/14475176027/job/40598974266?pr=2631#logs) is failing; it appears to be unrelated to this change.

Let me take a look at this

asomers (approved):

SteveLauC:
Hi @JonathonReinhart, you can rebase your branch to fix the CI issues

asomers (approved):

PR #2632, opened 2025-04-16
ci: downgrade libc to 0.2.171 in CI for hurd

SteveLauC:
## What does this PR do

1. In our CI, pin libc to 0.2.171 on hurd to work around this issue: https://github.com/rust-lang/libc/issues/4421
2. Bump CI runner images, Ubuntu-20.04 is deprecated. 
3. Ignore test `test_prctl::test_set_vma_anon_name` under QEMU due to #2418
4. Allow `clippy::unwrap_or_default` as IMHO it is overly opinionated

## Checklist:

### DIFF
--- a/.github/workflows/check_new_changelog.yml
+++ b/.github/workflows/check_new_changelog.yml
@@ -9,7 +9,7 @@ permissions:
 
 jobs:
   check_new_changelog:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


Berrysoft:
Please don't do that, because cygwin needs 0.2.172

SteveLauC:
>  how about just to pin libc to 0.2.171 on hurd?

Sounds good, I will do it

asomers (commented):
Instead of pinning libc in Cargo.toml, I think you should do it just in the CI configuration . On the hurd target, do an extra step of `cargo update -p libc --precise 0.2.171`.  That way we won't be limiting the libc version that our consumers can use.

PR #2635, opened 2025-04-30
Clippy cleanup: dangerous_implicit_autorefs

asomers:
## What does this PR do
Fixes a compiler warning with the latest nightly
## Checklist:


asomers:
The breakage on Hurd has gotten worse.  Now, building std brings in libc 0.2.172.  Maybe we can fix it by pinning the nightly toolchain to an older version?  But given that Hurd is basically a toy, I don't want to spend too much time on it.

SteveLauC (approved):
Thanks!

PR #2643, opened 2025-06-16
Fix Solaris CI after solaris-vm was updated to CBE release 11.4.81

psumbera:
Fixes: #2642

New Solaris CBE release 11.4.81 comes with updated version of `sudo`.


.github/workflows/ci.yml
```diff
@@ -396,7 +396,7 @@ jobs:
             uname -a
           run: |
             export PATH=$HOME/.rust_solaris/bin:$PATH
-            cargo build --target x86_64-pc-solaris --all-targets --all-features && sudo cargo test
+            cargo build --target x86_64-pc-solaris --all-targets --all-features && sudo env PATH=$PATH cargo test
```
asomers, line 399: This looks like exactly what `sudo --preserve-env=PATH` is for.  Does solaris's sudo have that option?

psumbera: > This looks like exactly what `sudo --preserve-env=PATH` is for. Does solaris's sudo have that option?

It does. But I think it needs to be enabled somewhere in configuration. Which doesn't seem to be the case.

Issue #2641, opened 2025-06-05
nix with ptrace feature fails to build on s390x


alip:
Thanks for the awesome crate and great work! I am the main author of [syd](https://lib.rs/crates/syd) and we've been using nix in our project with great joy. Below is an error we have stumbled upon, thanks to help from Alpine Linux folks on s390x+musl with ptrace feature enabled.

Reference: https://gitlab.alpinelinux.org/alpine/aports/-/jobs/1884950
Excerpt:

```
error[E0308]: mismatched types
   --> /home/buildozer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.30.1/src/sys/ptrace/linux.rs:331:22
    |
331 |         libc::ptrace(request as RequestType, libc::pid_t::from(pid), addr, data)
    |         ------------ ^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |         |
    |         arguments to this function are incorrect
    |
note: function defined here
   --> /home/buildozer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.172/src/unix/linux_like/linux/musl/mod.rs:927:12
    |
927 |     pub fn ptrace(request: c_int, ...) -> c_long;
    |            ^^^^^^
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
331 |         libc::ptrace((request as RequestType).try_into().unwrap(), libc::pid_t::from(pid), addr, data)
    |                      +                      +++++++++++++++++++++
error[E0308]: mismatched types
   --> /home/buildozer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.30.1/src/sys/ptrace/linux.rs:505:13
    |
504 |         libc::ptrace(
    |         ------------ arguments to this function are incorrect
505 |             request as RequestType,
    |             ^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |
note: function defined here
   --> /home/buildozer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.172/src/unix/linux_like/linux/musl/mod.rs:927:12
    |
927 |     pub fn ptrace(request: c_int, ...) -> c_long;
    |            ^^^^^^
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
505 |             (request as RequestType).try_into().unwrap(),
[...]

PR #2648, opened 2025-07-13
Fix GH-2641: ptrace with unsigned int is only for uclibc/Linux glibc.

devnexen:


asomers (commented):
Thanks for the bugfix!  This probably deserves a changelog entry, too.  Could you please add that?

asomers (changes requested):
"fix" is very vague.  What did you fix?  It's important to say whether you're fixing a compile-time issue or a runtime one.  Also, your commit message should end with "Fixes #2641".

PR #2657, opened 2025-08-03
enhanced renameat2 to support raw syscalls

hz2:
## What does this PR do

- replace libc::renameat2() with libc::syscall() for broader compatibility
- define syscall numbers manually for multiple architectures
- update conditional compilation to support all Linux environments
- expand test coverage for additional architectures

context: https://github.com/uutils/coreutils/pull/8391

## Checklist:


asomers (commented):
Nix avoids raw syscalls for good reason.  Why do you need to do this?  If musl does not define renameat2, then renameat2 shouldn't be used in musl.

hz2:
> Nix avoids raw syscalls for good reason.

Do you mind sharing the good reason? I am personally unfamiliar with why it is or is not a good idea, but would love to learn.

> Why do you need to do this? If musl does not define renameat2, then renameat2 shouldn't be used in musl.

The discussion [here](https://github.com/uutils/coreutils/pull/8391#discussion_r2244236476) outlined the desire to use the nix crate, but I can continue with using raw syscalls.

asomers:
It's mainly about type safety.  Using raw syscalls provides absolutely none.  Using the libc bindings is much safer.  Furthermore, there is often important logic in the platform's libc .  For example, Go's insistence on using raw syscalls is the reason why many Go programs are still forced to use fork/exec rather than the higher performance posix_spawn.

hz2:
Thanks for the explanation, makes sense. Closing the issue.

PR #2661, opened 2025-08-14
Fix acct type signature difference for AIX

dalvescb:
## What does this PR do

Fixes acct mod to provide custom methods that account for differences in AIX's `libc::acctl` type interface

## Checklist:

- [X] Does not require a change in tests or rustdoc comments
- [X] Does not require to change to nix' API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,23 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 # Change Log
 
+## [0.31.2] - 2026-02-28
+
+### Added
+
+- Add WatchDescriptor::as_raw, to get libc id of WatchDescriptor.
+  ([#2718](https://github.com/nix-rust/nix/pull/2718))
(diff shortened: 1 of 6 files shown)


dalvescb:
> Why does AIX need a mutable pointer? 

This is just how AIX defines `acct` in its system headers (see https://public.dhe.ibm.com/aix/freeSoftware/aixtoolbox/docs/apidiffs.html#read). On linux the function interface is `int acct(const char *filename);` and on AIX `int acct (char *Path);` The libc crate requires rust accurately reflects the system headers, hence when we ported it we needed to maintain a mutable pointer

> If there is any possibility that it may modify the caller-provided string, then your pointer cast is invalid.

Despite the difference in interface, the Linux and AIX definitions of `acct` are documented as compatible and `acct` does not modify its argument. So the cast should be valid

dalvescb:
We are pursuing getting the interface on AIX fixed, which should resolve this. So discarding this PR for now

PR #2670, opened 2025-09-03
Improve mqueue documentation and documentation example

hicklin:
From reading the [mqueue c implementation](https://github.com/RT-Thread/rt-thread/blob/893ae7d7ba4daecfe4732cb51db8f140dc7211ff/components/libc/posix/ipc/mqueue.c#L141) that this library wraps, it seems that we can't set `attr` to `None` when creating the mqueue `MQ_OFlag::O_CREAT`. If we do, we get an `Errno::EINVAL`. This means that the example in the documentation is incorrect. 

## What does this PR do

This PR updates the mqueue documentation example and adds some helpful documentation to explain the limits of the `MqAttr` input values.

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -2,24 +2,24 @@
 //!
 //! # Example
 //!
-// no_run because a kernel module may be required.
-//! ```no_run
+//! ```
 //! # use std::ffi::CString;
 //! # use nix::mqueue::*;
 //! use nix::sys::stat::Mode;
 //!
 //! const MSG_SIZE: mq_attr_member_t = 32;
-//! let mq_name= "/a_nix_test_queue";
+//! let attr = MqAttr::new(0, 10, MSG_SIZE, 0);
+//! let mq_name= "/a_nix_doc_test_queue";
 //!
 //! let oflag0 = MQ_OFlag::O_CREAT | MQ_OFlag::O_WRONLY;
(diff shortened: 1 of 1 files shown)


src/mqueue.rs
```diff
@@ -9,17 +9,18 @@
 //! use nix::sys::stat::Mode;
```
hicklin, line 9: Is there a reason why `no_run` is required? This is only executed by `cargo test` and if this fails due to missing kernel modules, so would the mqueue tests.

hicklin: OK. I have amended the commit to remove `no_run`.

src/mqueue.rs
```diff
@@ -100,7 +100,9 @@ impl MqAttr {
     ///
     /// - `mq_flags`:   Either `0` or `O_NONBLOCK`.
     /// - `mq_maxmsg`:  Maximum number of messages on the queue.
+    ///   The maximum value allowed by the system can be obtained from `/proc/sys/fs/mqueue/msg_max`.
     /// - `mq_msgsize`: Maximum message size in bytes.
+    ///   The maximum value allowed by the system can be obtained from `/proc/sys/fs/mqueue/msgsize_max`.
```
asomers, line 105: These two comments are Linux-specific.  But the API is cross-platform.  So it's not really appropriate to put these here.

asomers (approved):
Thanks for your contribution.  This change is small enough that I don't think we need a changelog entry.  And I think that the CI failure should be fixed by #2682 .

PR #2687, opened 2025-10-12
Fix IPV6_PKTINFO/IP_PKTINFO sockopts: use in6_pktinfo/in_pktinfo

m-asama:
### Summary
This PR fixes the Ipv6PacketInfo and Ipv4PacketInfo sockopts to use the proper payload types (libc::in6_pktinfo and libc::in_pktinfo) instead of bool. It also updates the docs to state that these options relate to outgoing packets (not “incoming”).

### Rationale
	•	Per RFC 3542 (“Advanced Sockets API for IPv6”), IPV6_PKTINFO is defined as an option that can be used both as a sticky socket option (via setsockopt) and as ancillary data (via sendmsg/recvmsg), carrying a struct in6_pktinfo. When read via getsockopt, the value is an in6_pktinfo structure (with ipi6_addr = in6addr_any and ipi6_ifindex = 0 if no sticky value is set).  ￼
	•	On Linux, enabling delivery of pktinfo on receive is controlled by IPV6_RECVPKTINFO (boolean), while the control message IPV6_PKTINFO itself carries a struct in6_pktinfo. Treating IPV6_PKTINFO as bool conflates these two distinct interfaces.  ￼
	•	RFC 3542 also clarifies the outgoing interface selection order: if an IPV6_PKTINFO ancillary data item specifies an interface, that interface is used. The current docs say “incoming packet”, which is misleading for these options. (Section “6.7. Summary of Outgoing Interface Selection”.)  

### What changes
	•	Change sockopt value types:
	•	Ipv6PacketInfo: bool → libc::in6_pktinfo
	•	Ipv4PacketInfo: bool → libc::in_pktinfo
	•	Doc comments: “incoming packet” → “outgoing packet”

### Compatibility
This is an API breaking change for the affected sockopts (Val = bool → struct). However, the previous signature was incorrect per the standards and typical OS behavior. Code that intended to enable receiving pktinfo should have been using Ipv6RecvPacketInfo (boolean). Code that wants to select the source address / outgoing interface should pass an in6_pktinfo.

## Checklist:

### DIFF
--- a/changelog/2687.fixed.md
+++ b/changelog/2687.fixed.md
@@ -1,0 +1,1 @@
+fixed IPV6_PKTINFO/IP_PKTINFO sockopt type from bool to in6_pktinfo/in_pktinfo


m-asama:
Regarding IPv4, since there seems to be an error in my understanding, I will withdraw it for now.

PR #2689, opened 2025-10-12
Fix CI by relaxing our MSRV constraint

asomers:
Recently a transitive dev-dependency published a new version that is incompatible with Nix's MSRV, causing CI to break.  That's annoying, because there's really no reason why a crate's dev-dependencies ought to respect the MSRV.  Fix Nix's CI by:

* Running all tests with stable Rust (or nightly, for certain targets)
* Adding an additional set of CI checks to ensure that the crate will compile with MSRV, on all of the major operating systems, but don't try to compile its tests.
* Eliminate the now-redundant "rust-stable" CI task.

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -772,7 +772,9 @@ pub enum ControlMessageOwned {
     ///
     /// # Examples
     ///
-    /// ```
+#[cfg_attr(all(target_env = "musl", target_pointer_width = "32"), doc = "See <https://github.com/nix-rust/nix/issues/2698> for notes regarding 32-bit musl")]
+#[cfg_attr(all(target_env = "musl", target_pointer_width = "32"), doc = "```no_run")]
+#[cfg_attr(any(not(target_env = "musl"), target_pointer_width = "64"), doc="```")]
     /// # #[macro_use] extern crate nix;
     /// # use nix::sys::socket::*;
     /// # use nix::sys::time::*;


asomers:
The lio_listio doc tests are still hanging, sometimes, on aarch64 linux and i686-unknown-linux-musl .  I guess to debug that I'll need to install a new Linux VM locally and try to duplicate the i686-unknown-linux-musl failure.

SteveLauC:
> The lio_listio doc tests are still hanging, sometimes, on aarch64 linux and i686-unknown-linux-musl . I guess to debug that I'll need to install a new Linux VM locally and try to duplicate the i686-unknown-linux-musl failure.

It is easy for me to set up an aarch64 Linu VM, so I will work on this

SteveLauC:
Hi @asomers, I would like to know your thoughts on [this comment](https://github.com/rust-lang/api-guidelines/discussions/231#discussioncomment-1537616), maybe we should test the MSRV with deps of minimal versions as well

SteveLauC:
Well, I cannot reproduce the issue locally 🥲:

```sh
steve@ubuntu:~/nix$ git status
On branch msrv-stable

# Run tests 30 times
steve@ubuntu:~/nix$ cat test.sh
count=30
for i in $(seq $count); do
    cargo test --doc --all-features sys::aio::lio_listio
done

steve@ubuntu:~/nix$ bash test.sh > /dev/null 2>/dev/null
steve@ubuntu:~/nix$ echo $?
0
```

SteveLauC:
Interesting, I just noticed the failed CIs are using musl. And testing with `--target aarch64-unknown-linux-musl` can indeed reproduce the issue, let me debug the test

```sh
$ cargo test --all-features --doc  --target aarch64-unknown-linux-musl sys::aio::lio_listio
```

SteveLauC:
Created a repo for this: https://github.com/SteveLauC/musl_aarch64_memory_order

```sh
$ cd musl_aarch64_memory_order/
$ cargo r --target aarch64-unknown-linux-musl
```

Update:

1. I added a write(2) to the signal handler, in the cases where the program hung, I didn't see any output in the stdout, which basically means the signal handler didn't get invoked at all
2. I replaced AtomicBool with a pipe(), write(pipe_tx) in the signal handler and wait for I/O completion via read(pipe_rx). The issue still exists, so I guess it is not realted to memory ordering
3. BTW, as long as I run the program with strace, it works

asomers:
> Hi @asomers, I would like to know your thoughts on [this comment](https://github.com/rust-lang/api-guidelines/discussions/231#discussioncomment-1537616), maybe we should test the MSRV with deps of minimal versions as well

Yes, I've tried the Cargo.lock.msrv before.  However, in this case I don't think it will help us, since our only problematic dependency is a dev dependency.  And we can work around that just by doing "cargo check" or "cargo build" in CI on MSRV, without doing "cargo test".

asomers:
I can now reproduce it locally too, using i686-unknown-linux-musl .  But not x86_64-unknown-linux-musl.  And it would be interesting to try building the tests for i686-unknown-linux-gnu and running them in an x86_64 VM.  But I don't know what Debian packages I need to install for that.  Do you?
It's possible that this is a musl bug.

SteveLauC:
> And it would be interesting to try building the tests for i686-unknown-linux-gnu and running them in an x86_64 VM. But I don't know what Debian packages I need to install for that. Do you?

I am not sure about this


> It's possible that this is a musl bug.

Yeah, I doubt it as well. But it is strange why we didn't encounter this issue in the past. I just tried running that test with our MSRV toolchain, the program still hung

asomers:
> > And it would be interesting to try building the tests for i686-unknown-linux-gnu and running them in an x86_64 VM. But I don't know what Debian packages I need to install for that. Do you?
> 
> I am not sure about this
> 
> > It's possible that this is a musl bug.
> 
> Yeah, I doubt it as well. But it is strange why we didn't encounter this issue in the past. I just tried running that test with our MSRV toolchain, the program still hung

It's not our job to debug musl problems.  I think we should just disable this test for that environment and move on.

SteveLauC:
> It's not our job to debug musl problems. I think we should just disable this test for that environment and move on.

Yeah, let me finish this

asomers:
Do you think all of those lock ordering changes were really necessary, or were we just chasing a musl bug this whole time?

asomers:
Oh, and I think the reason why we only started running into this bug now is because when this PR upgraded the testing image from MSRV to stable, it brought in a whole new container, not just a newer toolchain.  And that container probably included newer versions of musl or some other library.

SteveLauC:
> Do you think all of those lock ordering changes were really necessary, or were we just chasing a musl bug this whole time?

You mean the memory ordering changes in the lio_listio doc test? I think we don't need them, I could still reproduce the issue without using AtomicBool (by replacing it with a pipe).

SteveLauC:
> I think we don't need them, I could still reproduce the issue without using AtomicBool (by replacing it with a pipe).

Reverted in https://github.com/nix-rust/nix/pull/2689/commits/aadefd85347cb56dd6223b738fb436f451f6dc31

SteveLauC:
There is still one issue to look into:

https://github.com/nix-rust/nix/actions/runs/19092747880/job/54546489456

```
 ---- src/sys/socket/mod.rs - sys::socket::ControlMessageOwned::ScmTimestamp (line 775) stdout ----
Test executable failed (exit status: 101).

stderr:

thread 'main' (1509) panicked at src/sys/socket/mod.rs:34:29:
called `Result::unwrap()` on an `Err` value: ENOBUFS
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/core/src/panicking.rs:75:14
   2: core::result::unwrap_failed
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/core/src/result.rs:1855:5
   3: rust_out::main
   4: core::ops::function::FnOnce::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.



failures:
    src/sys/socket/mod.rs - sys::socket::ControlMessageOwned::ScmTimestamp (line 775)
```

I will investigate it today when I have time

asomers:
I think I've figured out the nature of the i686 musl test failure.  It's weird.

It seems that when calling `recvmsg` on i686-musl, at least with an x86_64-gnu host, and setting `SO_TIMESTAMP`, the system will return not one but two cmsgs.  The first is the expected `SCM_TIMESTAMP` msg for a system with 32-bit `long_t`.  It has `cmsg_type==29`. But the second has `cmsg_type==63`.  That corresponds to `SCM_TIMESTAMP` for systems with 64-bit longs[^1].  I guess the kernel is returning two different cmsgs, because it doesn't know which format our executable is expecting?  I can't find any documentation that says that the kernel will return both types.  But within the linux source code, Documentation/networking/timestamping.rst suggests that new programs ought to use `SO_TIMESTAMP_NEW` (63) instead of `SO_TIMESTAMP_OLD` (29).  So maybe we should convert Nix to do that.  In the meantime, I'm inclined to document this behavior somewhere and disable the doc test for i686-musl.

[^1]: https://git.musl-libc.org/cgit/musl/tree/include/sys/socket.h#n237

SteveLauC:
Sorry for the late reply, I have been doing a lot of overtime lately (worked all last weekend) 🫠.

Thanks for figuring that issue out, let's merge this PR:)

asomers:
I cancelled the merge, because I'd like to squash some commits first.

PR #2694, opened 2025-11-03
fix: make nix build on Illumos

extrowerk:
## What does this PR do
This changes lets nix build and work on Illumos based operating systems. Tested on OpenIndiana and OmniOS.

## Checklist:

- [x ] I have read `CONTRIBUTING.md`

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -96,6 +96,7 @@ feature! {
 #[cfg(not(any(
     target_os = "redox",
     target_os = "fuchsia",
+    target_os = "solaris",
     target_os = "haiku"
 )))]
 feature! {


SteveLauC:
Let me add a changelog for this PR

SteveLauC:
Hi @extrowerk, looks like module `sys/resource` should be only enabled on Illumos, and Solaris does not have it. So I disabled it on Solaris, see https://github.com/nix-rust/nix/pull/2694/commits/be2352ff2e7339b016394520ae26d6971c62a78b. 

I guess it should still work, but it would be still great if you could give this patch a re-check by either compiling nix or the fish shell on an Illumos machine.

SteveLauC:
Emm, I forgot that auto-emrge was still enabled when I left my last comment:D

PR #2705, opened 2025-11-27
Fix glibc 2.42 termios panic

### LINKED ISSUE #2672: cfgetospeed(): called `Result::unwrap()` on an `Err` value: EINVAL
Given

```rust
use nix::sys::termios::{
    cfgetospeed, tcgetattr
};

fn main() {
    let stdin = std::io::stdin();
    let termios = tcgetattr(stdin).expect("Could not get terminal attributes");
    let speed = cfgetospeed(&termios);
    println!("{:?}", speed);
}
```

We receive:

```
> cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/demo`

thread 'main' panicked at /home/jak/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.30.1/src/sys/termios.rs:767:70:
called `Result::unwrap()` on an `Err` value: EINVAL
```

with glibc 2.42

(Minimized reproducer of https://github.com/uutils/coreutils/issues/8474)

naoNao89:
stty panics on Ubuntu 25.10 (glibc 2.42) with EINVAL in cfgetospeed/cfgetispeed.

Solution: Replace .unwrap() -> .unwrap_or(BaudRate::B0) to handle conversion errors gracefully.

Fixes #2672, uutils/coreutils#8474

### DIFF
--- a/src/sys/termios.rs
+++ b/src/sys/termios.rs
@@ -753,18 +753,26 @@ cfg_if! {
         /// [cfgetispeed(3p)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/cfgetispeed.html)).
         ///
         /// `cfgetispeed()` extracts the input baud rate from the given `Termios` structure.
+        ///
+        /// On glibc 2.42+, returns `BaudRate::B0` if the actual baud rate cannot be determined.
         pub fn cfgetispeed(termios: &Termios) -> BaudRate {
             let inner_termios = termios.get_libc_termios();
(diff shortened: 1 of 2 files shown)


oech3:
Is there natural `cargo`'s system to backport this patch to uutils's main? It seems this path is working well.

ChrisDryden:
Do we have any documentation on that fallback pattern being the default that we can add to this PR?

oech3:
I think we should have an error instead fallbacking to `B0`.

PR #2711, opened 2025-12-11
refactor to_exec_array

xtqqczze:
the existing implementation in `src/spawn.rs` results in a possible unnecessary allocation

use a common optimised implementation


src/spawn.rs
```diff
@@ -351,12 +351,11 @@ impl Drop for PosixSpawnFileActions {
 // SAFETY:
 // It is safe to add the mutability in types as implementations won't mutable them.
 unsafe fn to_exec_array<S: AsRef<CStr>>(args: &[S]) -> Vec<*mut libc::c_char> {
```
xtqqczze:
```suggestion
fn to_exec_array<S: AsRef<CStr>>(args: &[S]) -> Vec<*mut libc::c_char> {
```

PR #2714, opened 2025-12-13
[AIX] Define timespec_tv_nsec_t as 32-bit.

amy-kwan:
## What does this PR do
On AIX, libc >= 0.2.178 exposes `timespec.tv_nsec` as 32-bit (`c_int`) on both
32- and 64-bit systems to match the platform ABI. This PR updates nix to use
the same width.

This change also eliminates the E0308 mismatched-type errors when building nix
and dependents on AIX with libc 0.2.178+, and keeps FFI layout consistent with the
underlying system ABI.

## Checklist:

### DIFF
--- a/changelog/2714.fixed.md
+++ b/changelog/2714.fixed.md
@@ -1,0 +1,1 @@
+Define `timespec_tv_nsec_t` as 32-bit on both 32- and 64-bit AIX to match the platform ABI.


amy-kwan:
This PR is no longer needed, as we have a follow up libc patch [AIX: Revert timespec.tv_nsec to c_long and use st_timespec for stat](https://github.com/rust-lang/libc/pull/4931) to resolve the timespec related issues we were seeing previously.

Issue #1352, opened 2020-12-07
ENOENT  in af_alg tests on Cirrus-CI


asomers:
The `sys::test_socket::{test_af_alg_cipher, test_af_alg_aead}` tests fail when run on Cirrus CI.  `bind` returns `ENOENT` for both tests, despite /proc/crypto containing both algorithms.  For now, we'll skip those tests on Cirrus.  But we need to root cause this issue.

SteveLauC:
Hit the issue again when I am trying to migrate the aarch64 Linux CIs to Github Action #2600

SteveLauC:
Should be fixed by https://github.com/nix-rust/nix/pull/2796, which does runtime check to see if we should run these tests or not.

PR #2722, opened 2026-01-11
skip test::sys::test_af_alg_cipher on s390x/Linux too

decathorpe:
It appears that s390x-unknown-linux-gnu is also affected by issue https://github.com/nix-rust/nix/issues/1352:

```
---- sys::test_socket::test_af_alg_cipher stdout ----
thread 'sys::test_socket::test_af_alg_cipher' (1118) panicked at test/sys/test_socket.rs:961:39:
bind failed: ENOENT
```

This PR just adds the s390x/Linux configuration to the list of environments where this specific test is skipped.

### DIFF
--- a/test/sys/test_socket.rs
+++ b/test/sys/test_socket.rs
@@ -921,11 +921,21 @@ pub fn test_scm_rights() {
 
 // 1. Disable the test on emulated platforms due to not enabled support of
 //    AF_ALG in QEMU from rust cross
-// 2. Disable the test on aarch64/Linux CI because bind() fails with ENOENT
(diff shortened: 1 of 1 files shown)


decathorpe:
Upon further investigation, this test seems to fail with ENOENT on powerpc64 / little-endian too. :(

SteveLauC (approved):

SteveLauC:
> Upon further investigation, this test seems to fail with ENOENT on powerpc64 / little-endian too. :(


Do you want to include it in this PR as well?

decathorpe:
Done - thanks for taking a look!

PR #2724, opened 2026-01-12
chore: bump libc to 0.2.180

3v1n0:
It's required to expose missing definitions used by coreutils

## Checklist:

### DIFF
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1175,16 +1175,20 @@ libc_bitflags! {
         /// Request that pages be moved instead of copied.
         ///
         /// Not applicable to `vmsplice`.
+        #[cfg(not(target_env = "uclibc"))]
         SPLICE_F_MOVE;
         /// Do not block on I/O.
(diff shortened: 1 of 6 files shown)


3v1n0:
I've replicated the `armv7-unknown-linux-uclibceabihf` issue myself and it's tracked in this rustc upstream issue: https://github.com/rust-lang/rust/issues/151037#issuecomment-3741673121

Should we disable the build in CI for now?

3v1n0:
Hey @SteveLauC, this should be ready now! :)

Cargo.toml
```diff
@@ -28,7 +28,7 @@ targets = [
 ]
 
 [dependencies]
-libc = { version = "=0.2.175", features = ["extra_traits"] }
+libc = { version = "=0.2.180", features = ["extra_traits"] }
```
tisonkun: Why can't we use `version = "0.2.180"` but pin to an exact one?

If `libc` follows semver, this should be fine. But `libc` may remove variables even if the version doesn't increase a breaking one?

PR #2725, opened 2026-01-18
FreeBSD: add pthread_getthreadid_np()

kostikbel:
## What does this PR do
The process::gettid() is missing on FreeBSD.
The PR adds its implementation.

The implementation returns the real tid of the caller's thread, it does
not try to mimic the Linux behavior where initial (main) thread id is
same as the processs id. This might be patched if needed, but pids and
tids are in different namespace on FreeBSD, so better keep it as is
unless there is a pressure.

## Checklist:

### DIFF
--- a/test/test_unistd.rs
+++ b/test/test_unistd.rs
@@ -260,6 +260,19 @@ mod linux_android {
     }
 }
 
+#[cfg(target_os = "freebsd")]
+mod freebsd {
+    use nix::unistd::pthread_getthreadid_np;
+
+    #[test]
+    fn test_pthread_getthreadid_np() {
+        let tid: ::libc::pid_t = pthread_getthreadid_np().into();
+        // FreeBSD has thread id namespace shared with pids, the split
+        // is at PID_MAX = 99999.
+        assert!(tid >= 100000);
+    }
+}
+
 #[test]
 // `getgroups()` and `setgroups()` do not behave as expected on Apple platforms
 #[cfg(not(any(


src/unistd.rs
```diff
@@ -403,6 +403,14 @@ pub fn getpgrp() -> Pid {
 pub fn gettid() -> Pid {
     Pid(unsafe { libc::syscall(libc::SYS_gettid) as pid_t })
 }
+
+/// Get the caller's thread ID (see
+/// [pthread_getthreadid_np(3)](https://man.freebsd.org/cgi/man.cgi?query=pthread_getthreadid_np&sektion=3&manpath=FreeBSD+15.0-RELEASE+and+Ports).
+#[cfg(freebsd)]
+#[inline]
```
SteveLauC:
```suggestion
#[inline]
#[doc(alias = "pthread_getthreadid_np")]
```

Maybe we should add an alias 👀

src/unistd.rs
```diff
@@ -403,6 +403,14 @@ pub fn getpgrp() -> Pid {
...
+pub fn gettid() -> Pid {
```
SteveLauC: Open question, should we call it `gettid()`, i.e., follow the Linux convention here, or we should use its own name?

I am slightly inclined to use its original name as Nix, in most cases, only exposes system stuff in Rusty way, we do not wrap them

src/unistd.rs
SteveLauC: If we use its own name, this alias is not needed.

See the below comment

SteveLauC (commented):
Thanks for the PR. A changelog entry is needed, BTW, see https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests

kostikbel:
> Thanks for the PR. A changelog entry is needed, BTW, see https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests

All done, thanks for the review.

PR #2728, opened 2026-01-21
Allow timerfd use on FreeBSD

jakesfreeland:
FreeBSD has natively supported timerfd(2) system calls since FreeBSD
14. Add missing cfg directives for FreeBSD builds so timerfd is
accessible.

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -187,7 +187,7 @@ feature! {
     pub mod inotify;
 }
 
-#[cfg(linux_android)]
+#[cfg(any(linux_android, target_os = "freebsd"))]
 feature! {
     #![feature = "time"]
     pub mod timerfd;
--- a/changelog/2728.added.md
+++ b/changelog/2728.added.md
@@ -1,0 +1,1 @@
+Added timerfd support on FreeBSD
--- a/src/sys/time.rs
+++ b/src/sys/time.rs
@@ -16,7 +16,10 @@ const fn zero_init_timespec() -> timespec {
 }
 
 #[cfg(any(
-    all(feature = "time", any(target_os = "android", target_os = "linux")),
+    all(
+        feature = "time",
+        any(target_os = "android", target_os = "freebsd", target_os = "linux")
+    ),
     all(
         any(
             target_os = "freebsd",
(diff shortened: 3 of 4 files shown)


src/sys/mod.rs
```diff
@@ -187,7 +187,7 @@ feature! {
     pub mod inotify;
 }
 
-#[cfg(linux_android)]
+#[cfg(any(freebsd, linux_android))]
```
SteveLauC:
```suggestion
#[cfg(any(target_os = "freebsd", linux_android))]
```

We don't use those aliases that only contain 1 OS, see https://github.com/nix-rust/nix/blob/master/CONVENTIONS.md#cfg-gates

src/time.rs
```diff
@@ -56,7 +56,12 @@ impl ClockId {
         self.0
     }
 
-    #[cfg(any(linux_android, target_os = "emscripten", target_os = "fuchsia"))]
+    #[cfg(any(
+        freebsd,
```
SteveLauC:
```suggestion
        target_os = "freebsd",
```

src/time.rs
```diff
@@ -104,7 +114,12 @@ impl ClockId {
     pub const CLOCK_REALTIME_ALARM: ClockId =
         ClockId(libc::CLOCK_REALTIME_ALARM);
     /// Like [`CLOCK_REALTIME`](ClockId::CLOCK_REALTIME), but optimized for execution time at the expense of accuracy.
-    #[cfg(any(linux_android, target_os = "emscripten", target_os = "fuchsia"))]
+    #[cfg(any(
+        freebsd,
```
SteveLauC:
```suggestion
        target_os = "freebsd",
```

src/sys/mod.rs
jakesfreeland: The `mismatched_target_os` lint doesn't seem to exist anymore on the page that the CONVENTIONS document links to: https://rust-lang.github.io/rust-clippy/master/index.html#/mismatched_target_os.

Do you know why using the target alias for a single OS is bad?

jakesfreeland:
Hey @SteveLauC, all comments have been addressed. Let me know if there is anything else that needs to be done to get this in. Thanks

SteveLauC:
> Hey @SteveLauC, all comments have been addressed. Let me know if there is anything else that needs to be done to get this in. Thanks

Sorry the late reply, PR LGTM

PR #2729, opened 2026-01-22
unistd: remove `fork`/`exec` on fuchsia

xtqqczze:
On Fuchsia, these functions are intentionally unimplemented and return `ENOSYS`.

References:

- https://fuchsia.dev/fuchsia-src/development/languages/c-cpp/libc#posix
- https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/src/process/fork.c
- https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/src/process/execv.c
- https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/src/process/execve.c

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -205,6 +205,7 @@ impl fmt::Display for Pid {
 /// When `fork` is called, the process continues execution in the parent process
 /// and in the new child.  This return type can be examined to determine whether
 /// you are now executing in the parent process or in the child.
+#[cfg(not(target_os = "fuchsia"))]
 #[derive(Clone, Copy, Debug)]
 pub enum ForkResult {
     /// This is the parent process of the fork.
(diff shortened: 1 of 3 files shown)


xtqqczze:
Hmm, it’s surprising that the `test_wait_signal test` would have previously passed if `fork` always returned an `Err` on Fuchsia:

```log
error[E0425]: cannot find function `fork` in this scope
  --> test/sys/test_wait.rs:15:20
   |
15 |     match unsafe { fork() }.expect("Error: Fork Failed") {
   |                    ^^^^ not found in this scope
   |
```

stepancheg:
Even if functions are unimplemented, it is often convenient to have them, so you don't have to turn your project into endless cfg-guards.

In fact, this is likely the reason why these function are declared in fuchsia, to make writing cross-platform code easier.

(I have very little to do with this project, so this is my 2 cents.)

PR #2736, opened 2026-02-05
Add interface argument to Ipv6MembershipRequest::new

### LINKED ISSUE #323: ipv6_mreq does not allow interface to be set

Skyb0rg007:
This commit adds an additional argument to the
Ipv6MembershipRequest::new function that allows the user to specify the interface index for which the
membership request is associated with.

Closes #323

## What does this PR do
This PR revives the efforts of #1320 to add proper support to the IPV6_JOIN_GROUP socket option.

Devices (especially routers) have many IPv6 interfaces, and being able to specify which interfaces to join a multicast group on is critical. Specifying `0` (as is currently done) is not portable:

> If the interface index is specified as 0, the kernel chooses local interface.  For example, some kernels look up multicast group in the normal IPv6 routing table and using the resulting interface.

## Checklist:

### DIFF
--- a/changelog/2736.changed.md
+++ b/changelog/2736.changed.md
@@ -1,0 +1,1 @@
+Changed the Ipv6MembershipRequest::new function to accept an interface index.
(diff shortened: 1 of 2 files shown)


Skyb0rg007:
I guess this is also a duplicate of #2688, with a slightly different API choice. It would be great if one of these PRs were at least reviewed.

PR #2747, opened 2026-02-23
Check exit code in Dir::as_raw_fd

stepancheg:
## What does this PR do

`libc::dirfd` should not fail. But if does, we better do some error detection (panic) rather than quietly returning incorrect value.

Alternatively we can stop implementing `AsRawFd`.

## Checklist:


SteveLauC:
I rebased your branch to resolve CI failures.

PR #2748, opened 2026-03-08
add `pidfd_open(2)` and `pidfd_send_signal(2)`

CobaltCause:
## What does this PR do

It adds `pidfd_open(2)` and `pidfd_send_signal(2)`. I personally don't have an immediate need for `pidfd_getfd(2)` so it's not included in this PR. It's particularly important to me that the `pidfd_open(2)` implementation is async-signal-safe, so that it can be used between `fork` and `exec` to achieve race-free PIDFD acquisition without violating any libc's invariants; see the conversation at https://github.com/rust-lang/rust/issues/89522#issuecomment-945840398 for details on that topic.

See https://github.com/nix-rust/nix/pull/1868 for prior art; review comments there have been addressed here.

## Checklist:

### DIFF
--- a/src/sys/pidfd.rs
+++ b/src/sys/pidfd.rs
@@ -3,6 +3,7 @@
 use std::{
     hint::unreachable_unchecked,
     os::fd::{FromRawFd as _, OwnedFd, RawFd},
+    ptr,
 };
 
 use bitflags::bitflags;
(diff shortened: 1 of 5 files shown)


CobaltCause:
[CI / cross (arm-unknown-linux-gnueabi) (pull_request)](https://github.com/nix-rust/nix/actions/runs/22830249828/job/66217351347?pr=2748) and [CI / cross (armv7-unknown-linux-gnueabihf) (pull_request)](https://github.com/nix-rust/nix/actions/runs/22830249828/job/66217351360?pr=2748) are failing because they don't have the syscalls, don't know what to do about that. [CI / cross (powerpc64le-unknown-linux-gnu) (pull_request)](https://github.com/nix-rust/nix/actions/runs/22830249828/job/66217351365?pr=2748) seems to be stuck in the test for `pidfd_send_signal(2)`, don't know what to do about that either[^1]. [CI / redox (pull_request)](https://github.com/nix-rust/nix/actions/runs/22830249828/job/66217351315?pr=2748) and [FreeBSD 14 amd64 & i686](https://github.com/nix-rust/nix/pull/2748/checks?check_run_id=66217196073) seem to be failing for unrelated reasons?

[^1]: Actually, maybe it's stuck because the parent process experienced a race with the child's PID...

CobaltCause:
Changing the test to eliminate any potential races actually made it worse, [CI / cross (arm-unknown-linux-gnueabi) (pull_request)](https://github.com/nix-rust/nix/actions/runs/22837662593/job/66237320294?pr=2748) and [CI / cross (armv7-unknown-linux-gnueabihf) (pull_request)](https://github.com/nix-rust/nix/actions/runs/22837662593/job/66237320280?pr=2748) became stuck in addition to the one mentioned before. I don't seem to be able to manually cancel the runners and there doesn't seem to be any timeout (the previous attempt ran for about five hours before I pushed again), so I'm just going to add a commit to disable the `pidfd_send_signal(2)` test for now.

CobaltCause:
The common problem between all three checks that were getting stuck is that all of them returned `ENOSYS` for the PIDFD syscalls. The second round of deadlocks were probably from `recvmsg` since the corresponding `sendmsg` would never have been reached, but the cause of the first deadlock is still beyond me. I've reverted back to the simpler test and handled `ENOSYS` such that the tests explicitly pass in that case, and now the deadlocks are gone and the new tests pass as expected. But now [CI / macos-aarch64 (pull_request)](https://github.com/nix-rust/nix/actions/runs/22867388629/job/66337739918?pr=2748) is also failing though this seems unrelated.

CobaltCause:
Hmm, pushed again and now [CI / macos-aarch64 (pull_request)](https://github.com/nix-rust/nix/actions/runs/22868376010/job/66341241208?pr=2748) is passing, maybe it's just flaky?

SteveLauC:
Hi, the CI failures should go away if you rebase your branch.

src/sys/pidfd.rs
```diff
@@ -0,0 +1,137 @@
...
+    //
+    // * Arguments passed to the syscall have the correct types.
+    // * The kernel should not return a value that cannot fit in `RawFd`.
+    // * The file descriptor returned by the kernel is open, owned, and requires
+    //   only `close` to release its resources.
+    // * The kernel should not return any negative value other than `-1`.
+    unsafe {
+        match libc::syscall(libc::SYS_pidfd_open, pid.as_raw(), flags.bits()) {
+            fd if fd >= 0 => {
+                Ok(OwnedFd::from_raw_fd(RawFd::try_from(fd).unwrap_unchecked()))
```
Skyb0rg007, line 55: You shouldn't need the `try_from`; `Ok(OwnedFd::from_raw_fd(fd as RawFd))` works.

src/sys/pidfd.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+            -1 => Err(Errno::last()),
+            _ => unreachable_unchecked(),
```
Skyb0rg007, line 58: This logic is the same as `Errno::result(fd)`

src/sys/pidfd.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// See [`pidfd_open(2)`] for details.
+///
+/// [`pidfd_open(2)`]: https://man7.org/linux/man-pages/man2/pidfd_open.2.html
+///
+/// # Safety
+///
+/// This can race with the PID in `pid` getting released or recycled, unless the
+/// PID is that of the calling process, in which case the only way a race could
+/// happen is if the current process exits.
```
Skyb0rg007, line 39: This is also safe when the pid is a child process. It may be worth mentioning that there are stipulations though as mentioned in https://man7.org/linux/man-pages/man2/pidfd_open.2.html

Skyb0rg007 (commented):
Some comments

PR #2752, opened 2026-03-10
mq_open: missing mq_attr doesn't imply opening an existing queue

millert:
It is valid to create a new queue with O_CREAT without specifying an attribute structure, in which case system-specific default attributes will be used.  The mode should always be specified and if the mq attributes are missing, a NULL pointer should be used in its place.

Otherwise, stack garbage will be used for the mode and the attributes, which often results in failure with EINVAL (due to mq_msgsize and mq_maxmsg being too large).

### DIFF
--- a/src/mqueue.rs
+++ b/src/mqueue.rs
@@ -165,7 +165,14 @@ where
                 &mq_attr.mq_attr as *const libc::mq_attr,
             )
         },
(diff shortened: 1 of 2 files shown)


SteveLauC:
I rebased your branch to fix the CI failures.

PR #2770, opened 2026-04-13
Make prctl module available on Android too.

qwandor:
This should work just the same as on other Linux targets.

## What does this PR do

Makes the prctl module available on Android the same as on other Linux targets.

## Checklist:

### DIFF
--- a/changelog/2770.added.md
+++ b/changelog/2770.added.md
@@ -1,0 +1,1 @@
+Made the `sys::prctl` module available on Android the same as on other Linux targets.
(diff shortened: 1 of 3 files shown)


kamalmarhubi:
This feels uncontroversial, but going to get a second opinion from @asomers or @SteveLauC before I start merging random PRs. It's been a while since I was involved here!

SteveLauC (approved):
`cargo check` passes locally, thanks!

PR #2773, opened 2026-04-14
DO NOT MERGE add a FreeBSD/aarch64 CI configuration

asomers:
Yes, I know that Cirrus is going to go away soon.  The purpose of this PR is to measure the speed of FreeBSD/aarch64 CI with cirrus for comparison with the other options.

## Checklist:


xtqqczze:
Maybe it's worth merging a FreeBSD/aarch64 configuration even if it will only be usable for a short time?

asomers:
> Maybe it's worth merging a FreeBSD/aarch64 configuration even if it will only be usable for a short time?

We could.  But I never bothered in the past because I don't think it will gain us much extra coverage.  Unlike Linux, FreeBSD is very consistent between architectures.

SteveLauC:
Closing this PR since we already decided to use vmactions.

PR #2777, opened 2026-04-16
make vmsplice unsafe

### LINKED ISSUE #2633: possible safety violation via vmsplice
It appears you can read freed memory without unsafe code via vmsplice:

```Rust
use nix::fcntl::{fcntl, splice, vmsplice, SpliceFFlags};
use nix::unistd::pipe;
use std::error::Error;
use std::fs::File;
use std::io::{IoSlice, Read};
use std::os::fd::{AsFd, AsRawFd};

type E = Box<dyn Error>;

fn vmsplice_all(fd: &(impl AsFd + std::fmt::Debug), buffer: &[u8]) -> Result<(), E> {
    let mut io = IoSlice::new(buffer);
    while !io.is_empty() {
        io.advance(vmsplice(fd, &[io], SpliceFFlags::empty())?);
    }
    Ok(())
}

fn splice_all(fd_in: impl AsFd, fd_out: impl AsFd, mut size: usize) -> Result<(), E> {
    while size > 0 {
        size -= splice(&fd_in, None, &fd_out, None, size, SpliceFFlags::empty())?;
    }
    Ok(())
}

fn main() -> Result<(), E> {
    const BUF_SIZE: usize = 4096 * 3;
    let pipe1 = pipe()?;
    let pipe_sz = fcntl(
        pipe1.1.as_raw_fd(),
        nix::fcntl::FcntlArg::F_SETPIPE_SZ(4 * 4096),
    )?;
    println!("pipe buffer size = {pipe_sz}");
    let pipe2 = pipe()?;
    let stat = vec![1; BUF_SIZE];
    let stat2 = vec![2; BUF_SIZE];
    let mut after = vec![0; BUF_SIZE];
    vmsplice_all(&pipe1.1, &stat)?;
    splice_all(&pipe1.0, &pipe2.1, stat.len())?;
    vmsplice_all(&pipe1.1, &stat2)?;
    drop(stat);
    File::from(pipe2.0).read_exact(&mut after)?;
    println!("{:?}", &after[..20]);
    Ok(())
}
```

See https://github.com/uutils/coreutils/issues/7625#issuecomment-2816241691 for additional context.

oech3:
## What does this PR do
make vmsplice unsafe. Closes https://github.com/nix-rust/nix/issues/2633
## Checklist:

### DIFF
--- a/changelog/2777.changed.md
+++ b/changelog/2777.changed.md
@@ -1,0 +1,1 @@
+Made `fcntl::vmsplice` an unsafe function.
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1317,8 +1317,10 @@ pub fn tee<Fd1: std::os::fd::AsFd, Fd2: std::os::fd::AsFd>(
 /// # Safety
 ///
 /// * The content of `iov` must not be changed until kernel has finished reading it "even after the caller process finished".
-/// If `SPLICE_F_GIFT` flag is set, the content of `iov` must never be changed
-/// and all of the pointers in `iov` are page aligned, and the lengths are multiples of a page size
+/// * If `SpliceFFlags::SPLICE_F_GIFT` is set:
+///   - The contents of `iov` must never be modified after the call.
+///   - All pointers in `iov` must be page-aligned.
+///   - All lengths must be a multiple of the system page size in bytes.
 /// **Note** Linux kernel does not have an API to know when the kernel finished using the content of RAM
 #[cfg(linux_android)]
 pub unsafe fn vmsplice<F: std::os::fd::AsFd>(
--- a/src/fcntl.rs
+++ b/src/fcntl.rs
@@ -1313,8 +1313,15 @@ pub fn tee<Fd1: std::os::fd::AsFd, Fd2: std::os::fd::AsFd>(
 ///
 /// # See Also
 /// *[`vmsplice`](https://man7.org/linux/man-pages/man2/vmsplice.2.html)
+///
+/// # Safety
+///
+/// * The content of `iov` must not be changed until kernel has finished reading it "even after the caller process finished".
+/// If `SPLICE_F_GIFT` flag is set, the content of `iov` must never be changed
+/// and all of the pointers in `iov` are page aligned, and the lengths are multiples of a page size
+/// **Note** Linux kernel does not have an API to know when the kernel finished using the content of RAM
 #[cfg(linux_android)]
-pub fn vmsplice<F: std::os::fd::AsFd>(
+pub unsafe fn vmsplice<F: std::os::fd::AsFd>(
     fd: F,
     iov: &[std::io::IoSlice<'_>],
     flags: SpliceFFlags,
--- a/test/test_fcntl.rs
+++ b/test/test_fcntl.rs
@@ -360,7 +360,8 @@ mod linux_android {
         let buf2 = b"defghi";
         let iovecs = [IoSlice::new(&buf1[0..3]), IoSlice::new(&buf2[0..3])];
 
-        let res = vmsplice(wr, &iovecs[..], SpliceFFlags::empty()).unwrap();
+        let res = unsafe { vmsplice(wr, &iovecs[..], SpliceFFlags::empty()) }
+            .unwrap();
 
         assert_eq!(6, res);
 


xtqqczze:
Changelog instructions at https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests

oech3:
changelog dir is empty. Is it actually used?

xtqqczze:
> changelog dir is empty. Is it actually used?

It's ingested for each release, e.g. https://github.com/nix-rust/nix/commit/bf1d0e9707189422f546e398594fa1a51a772d9d. So don't change `CHANGELOG.md` here.

I'd also describe this as "fixed" not "changed".

src/fcntl.rs
```diff
@@ -1313,8 +1313,9 @@ pub fn tee<Fd1: std::os::fd::AsFd, Fd2: std::os::fd::AsFd>(
 ///
 /// # See Also
 /// *[`vmsplice`](https://man7.org/linux/man-pages/man2/vmsplice.2.html)
+#[allow(clippy::missing_safety_doc)]
```
xtqqczze: Do you intend to add the safety comment?

oech3: No. vmsplice has no proper safe statement since kernel does not tell when kernel finished to use content.

oech3: Added comment instead.

src/fcntl.rs
```diff
@@ -1313,8 +1313,9 @@ pub fn tee<Fd1: std::os::fd::AsFd, Fd2: std::os::fd::AsFd>(
 ///
 /// # See Also
 /// *[`vmsplice`](https://man7.org/linux/man-pages/man2/vmsplice.2.html)
+#[allow(clippy::missing_safety_doc)] // no safe situation since kernel does not tell when the kernel finished to use content of RAM
```
xtqqczze: comment should be moved to `# Safety` section of doc comment, per [`missing_safety_doc`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc)

oech3:
>I'd also describe this as "fixed" not "changed".

I'm hesitant to use "fixed" for breaking change.

src/fcntl.rs
oech3: This is not safety comment. There is no way to use it safely.

src/fcntl.rs
oech3: If we never change content of RAM we passed to vmsplice, it might safe until process exit.

src/fcntl.rs
asomers: If there's no way to use it safely, then we certainly had better let our users know!

xtqqczze: See https://docs.rs/rustix/latest/rustix/pipe/fn.vmsplice.html

oech3: ok

changelog/2777.changed.md
```diff
@@ -0,0 +1 @@
+Change vmsplice to an unsafe function
```
xtqqczze:
```suggestion
Made `fcntl::vmsplice` an unsafe function.
```

oech3: ok

oech3:
Can we just remove this function?

src/fcntl.rs
```diff
@@ -1313,8 +1313,13 @@ pub fn tee<Fd1: std::os::fd::AsFd, Fd2: std::os::fd::AsFd>(
 ///
 /// # See Also
 /// *[`vmsplice`](https://man7.org/linux/man-pages/man2/vmsplice.2.html)
+/// # Safety
```
xtqqczze, line 1317:
```suggestion

/// # Safety
```

xtqqczze:
is this ready for review?

oech3:
I still don't like current unsafe doc.

src/fcntl.rs
```diff
@@ -1313,8 +1313,15 @@ pub fn tee<Fd1: std::os::fd::AsFd, Fd2: std::os::fd::AsFd>(
 ///
 /// # See Also
 /// *[`vmsplice`](https://man7.org/linux/man-pages/man2/vmsplice.2.html)
+///
+/// # Safety
+///
+/// * The content of `iov` must not be changed until kernel has finished reading it "even after the caller process finished".
+/// If `SPLICE_F_GIFT` flag is set, the content of `iov` must never be changed
+/// and all of the pointers in `iov` are page aligned, and the lengths are multiples of a page size
```
xtqqczze:
```suggestion
/// * If `SpliceFFlags::SPLICE_F_GIFT` is set:
///   - The contents of `iov` must never be modified after the call.
///   - All pointers in `iov` must be page-aligned.
///   - All lengths must be a multiple of the system page size in bytes.
```

asomers (approved):
This is a definite improvement.  But it's still way to unsafe for any but the bravest and most performance-obsessed programmer to use.  It's beyond the scope of this PR, but I wonder if a safe Rust API could be devised?  Something like make vmsplice only accept pointers to memory allocated by a special allocator, and only accept them by move, not by borrow.

oech3:
Someone said RAM allocated by mmap can is bit safer.

oech3:
I'm not familiar with kernel. But how about

input ->(sendfile etc...) memfd ->(mmap, edit and sendfile) -> output
?

SteveLauC (approved):
Thanks! I am surprised that the man page does not mention this (the issue still happens without the gift flag) at all.

oech3:
> I wonder if a safe Rust API could be devised? Something like make vmsplice only accept pointers to memory allocated by a special allocator, and only accept them by move, not by borrow.

We should do that by some way. std's people said splice to memfd & mmap does not provide same performance (not an actual zero-copy). So safe Rust cannot be faster than C currently.

oech3:
https://github.com/uutils/coreutils/issues/7625#issue-2960883882

>This wouldn't happen if the buffer was allocated with mmap outside of memory manager control. Such memory can be unmapped on function exit preventing any chance of it being corrupted.

oech3:
I might not understanding, but is the wrapper only supporting boxed MaybeUninit safe?
So we cannot see/edit it as [u8] without unsafe block, but send it between fd.

PR #2784, opened 2026-04-25
kqueue: use from_bits_retain to avoid panics on unknown flags

kamalmarhubi:
## What does this PR do

KEvent::flags() and KEvent::fflags() used from_bits().unwrap(), which panics if the kernel returns flags unknown to the current bitflags definition. Use from_bits_retain to preserve all bits without panicking.

## Checklist:

### DIFF
--- a/changelog/2784.fixed.md
+++ b/changelog/2784.fixed.md
@@ -1,0 +1,3 @@
+Fixed `KEvent::flags()` and `KEvent::fflags()` to use `from_bits_retain`
+instead of `from_bits().unwrap()`, preventing panics when the kernel
+returns unknown kqueue flags.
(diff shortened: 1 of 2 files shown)


xtqqczze:
`from_bits_retain` only became available with https://github.com/nix-rust/nix/pull/2027, so it's less likely the original intent was to panic.

kamalmarhubi:
> `from_bits_retain` only became available with #2027, so it's less likely the original intent was to panic.

Ah thanks for that context! I just assumed it was originally done in error since the in theory a newer kernel could return bits we don't know about.

asomers (approved):

SteveLauC (approved):

PR #2785, opened 2026-05-06
feat(socket): decode `(SOL_SOCKET, SO_MARK)` cmsg into `SoMark(u32)`

wngr:
## What does this PR do

Add a `ControlMessageOwned::SoMark(u32)` variant and a matching arm in `ControlMessageOwned::decode_from`, modeled on the existing `RxqOvfl(u32)` arm a few lines above. Receivers that have enabled `SO_RCVMARK` (Linux 5.19+, kernel commit
`6fd1d51cfa253b5ee7dae18d7cf1df830e9b6137`) get an `(SOL_SOCKET, SO_MARK)` ancillary message on every datagram carrying the kernel-set `skb->mark`; without this arm the cmsg falls through to the catch-all `Unknown` variant and forces a per-recv `Vec<u8>` allocation to copy four bytes.

`SO_MARK` is in `libc` for `linux_android` and `fuchsia`, but `SO_RCVMARK` (the option that actually causes the kernel to deliver the cmsg on receive) is Linux/Android-only, so the new arm is gated `#[cfg(linux_android)]`. Omitting `fuchsia` seems like the defensive choice.

Let me know if you think this requires additional tests. I validated that behavior locally.

## Checklist:

- [x] I have written ~necessary tests and~ rustdoc comments


wngr:
CI partially failing on clippy complaints seems unrelated to my changes.

SteveLauC:
> CI partially failing on clippy complaints seems unrelated to my changes.

I rebased your branch to fix the CI failures.

SteveLauC (commented):
> Let me know if you think this requires additional tests. I validated that behavior locally.

Yeah, a set-get test would be needed. In order to do that, we should also implement the socket option SO_RCVMARK (pub struct RcvMark).

SteveLauC:
Use [`require_capability`](https://github.com/nix-rust/nix/blob/34b0fbb1e86dab4971af12835618820295da128e/test/common/mod.rs#L17) when you need priviledge.

> Linux 5.19+, 

Use [`require_kernel_version`](https://github.com/nix-rust/nix/blob/34b0fbb1e86dab4971af12835618820295da128e/test/common/mod.rs#L105) to skip the test on older kernels.

SteveLauC:
[31](https://github.com/nix-rust/nix/actions/runs/26091863324/job/76719743328?pr=2785#step:5:135)
error[E0425]: cannot find value `SO_RCVMARK` in crate `libc`
    --> src/sys/socket/sockopt.rs:1183:11
     |
1183 |     libc::SO_RCVMARK,
     |           ^^^^^^^^^^
     
Looks like this constant is not defined for android in the libc crate: https://github.com/search?q=repo%3Arust-lang%2Flibc%20SO_RCVMARK&type=code

wngr:
You're fast :-). Yep, only scoped it now to Linux. Now it's symmetrical.

SteveLauC (approved):
Thanks

wngr:
@SteveLauC could I trouble you to create a new patch release with this feature? That would be extremely helpful for me. Thanks!

PR #2792, opened 2026-05-15
Enable ioctl for Haiku

pulkomandy:
## What does this PR do

Haiku supports ioctl without problems. Enable the feature and basic tests.
There are no "well-behaved" ioctl on Haiku, only "bad" ones with arbitrary numbers. So the corresponding tests for ioctl numbering are not enabled.

## Checklist:


SteveLauC (approved):
Thanks

Issue #2069, opened 2023-07-08
initgroups erroneously not defined on macOS


Elizafox:
Hello,

For some reason, `nix::unistd::initgroups` is explicitly not defined for macOS, with no justification in the docs: https://github.com/nix-rust/nix/blob/75a26cd1e0b44c8d953633ee45cfc189cda7e4d7/src/unistd.rs#L1782

This is strange, because the API [definitely exists](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/initgroups.3.html).

I'm unsure if it exists on iOS; I imagine it does, although in *most* cases it wouldn't be useful (except perhaps for jailbroken systems). But that's probably a separate issue.

Elizafox:
Okay, I dug through the history and docs, and saw that it said:

>"Note: This function is not available for Apple platforms. On those platforms, group membership management should be achieved via communication with the opendirectoryd service."

This is erroneous for this function. OpenDirectory is *just* for querying the directory. It cannot manage the groups of the process.

This also breaks a lot of daemons that use it to drop privileges, that should otherwise work, for no good reason.

asomers:
I don't know enough about OSX to answer your question.  @JayH5 can you and @Elizafox figure out how `setgroups`, `getgrouplist`, and `initgroups` are _really_ supposed to work on OSX?

SteveLauC:
That erroneous comment comes from this [commit](https://github.com/nix-rust/nix/commit/d4e7761923b71aac253675814ae6da723d74b59c), which is a part of #733, I looked through that PR, didn't find anything related to this wrong comment. I think we should update it given that it is not correct.

For these functions, if we accept their weird behaviors (see this [question: `getgroups()` on macOS is not returning all the supplementary groups, but instead just the primary group](https://stackoverflow.com/q/59405719/14092446)), then we should expose them, gentle ping on @Elizafox again, I am curious how you would use these function?


As a reference:

* [man intigroups(3)](https://keith.github.io/xcode-man-pages/initgroups.3.html)
* [man getgroups(2)](https://keith.github.io/xcode-man-pages/getgroups.2.html)
* [man setgroups(2)](https://keith.github.io/xcode-man-pages/setgroups.2.html)

Looks like `setgroups(2)` behaves in the same way as other UNIX implementations, though the manual explicitly says:

[...]

PR #2793, opened 2026-05-15
Add setgroups(2) to macOS/Apple targets

kpcyrd:
## What does this PR do

This is related to #2069 and https://github.com/kpcyrd/acme-redirect/pull/50, the `setgroups` api definitely exists, but it's currently not available in nix for that target, causing compile errors with some projects.

- https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/setgroups.2.html#//apple_ref/doc/man/2/setgroups
- https://keith.github.io/xcode-man-pages/setgroups.2.html
- https://docs.python.org/3/library/os.html#os.setgroups
- https://docs.python.org/3/library/os.html#os.getgroups

The nix commit had some information that was likely not accurate:

> "Note: This function is not available for Apple platforms. On those platforms, group membership management should be achieved via communication with the opendirectoryd service."

Quoting from @Elizafox in #2069 (which I believe to be correct):

> This is erroneous for this function. OpenDirectory is just for querying the directory. It cannot manage the groups of the process.

## Checklist:

### DIFF
--- a/src/unistd.rs
+++ b/src/unistd.rs
@@ -1962,7 +1962,6 @@ pub fn getgroups() -> Result<Vec<Gid>> {
 /// # try_main().unwrap();
 /// ```
 #[cfg(not(any(
-    apple_targets,
     target_os = "redox",
     target_os = "haiku"
 )))]


kpcyrd:
The test failures seem to be related to Linux kernel exploit mitigations (the FreeBSD one I don't know though).

SteveLauC:
> The test failures seem to be related to Linux kernel exploit mitigations (the FreeBSD one I don't know though).

They are unrelated, don't worry about them.


Update: you can rebase your branch after #2796 is merged.

src/unistd.rs
SteveLauC, line 1937: This comment should be updated if we expose it on apple targets.

SteveLauC (commented):
Maybe we should also add the other two functions as well:

* initgroups
* getgroups

I [previously had concerns about the weird semantics of getgtoups](https://github.com/nix-rust/nix/issues/2069#issuecomment-1837966131)￼￼, but that does not really bother me anymore since Nix is ultimately just a bindings crate.

SteveLauC:
A changelog entry is needed, see https://github.com/nix-rust/nix/blob/master/CONTRIBUTING.md#pull-requests.

PR #2799, opened 2026-05-30
Fix/termios from bits retain

snltd:
## What does this PR do

This is a rebase of @qwandor 's work in #2253 to fix bad termios behaviour on illumos. Not only does it fix the code sample in #2071 , but I was able to build [fish](
https://github.com/fish-shell/fish-shell#) 4.7.1 on illumos with my locally patched nix, and have it work fine. (Previously the term got torn down immediately, which is how I found this problem.)

All credit to @qwandor and @nospam3089. All I did was cherry pick some commits.

## Checklist:


iigs:
Added termios.rs file patch to the fish build package in OpenIndiana. https://github.com/OpenIndiana/oi-userland/commit/6a92fb982855d6a9a9611da9c11148e73acaacf9  I can confirm that fish now works with the patch to the nix crate.

plotnick (commented):
Independent confirmation and review from a different code base: we root-caused this exact bug in Oxide's support shell on illumos before finding this PR. Any `tcgetattr` → `tcsetattr` round trip through the wrapper zeroes `CBAUD` in `c_cflag` (baud becomes `B0`), after which `ldterm` half-kills the pty, so output is muted while input still flows (a very odd failure mode). Verified with `truss`, and matches #2071 precisely. We worked around it by switching rustyline to its termios pass-through backend, but this is the correct fix.

Patch covers all four flag fields at both conversion sites, completing the partial `from_bits_retain` that already shipped for `c_cflag` in `update_wrapper`, and the synthetic-bits round-trip test is deterministic on every platform. This fixes real terminal breakage on illumos.

PR #2807, opened 2026-07-25
Remove unnecessary casts

FranciscoTGouveia:
As these expressions already have the type to which they are being cast, the casts are actually redundant and can be safely deleted.

CC @nunoplopes


FranciscoTGouveia:
If possible, could I get a review on this, @SteveLauC ?

I believe the CI failure is unrelated. Am I missing something?

PR #2808, opened 2026-07-27
fix: bump cfg_aliases minimum version to 0.2.2

xtqqczze:
Raise the minimum `cfg_aliases` build-dependency version to 0.2.2.

When running `cargo -Zdirect-minimal-versions`, Cargo pins `cfg_aliases` to its declared minimum version (`0.2.1`). Its `cfg_aliases!` macro expansion contains a trailing semicolon in expression position, which current Rust rejects as a hard error.

Closes https://github.com/nix-rust/nix/pull/2812

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,7 +90,7 @@ caps = "0.5.3"
 sysctl = "0.4"
 
 [build-dependencies]
-cfg_aliases = "0.2.1"
+cfg_aliases = "0.2.2"
 
 [[test]]
 name = "test"


xtqqczze:
@SteveLauC PTAL

xtqqczze:
@asomers Can you merge this?

asomers (approved):

asomers:
@xtqqczze it LGTM, but we need to fix CI first.  Also, even if we merge this, I don't see any need for a patch release.  After all, consumers can easily fix any warnings by updating cfg_aliases in their own projects.

xtqqczze:
> @xtqqczze it LGTM, but we need to fix CI first.

This is required to fix the `minver` job in CI.

xtqqczze:
@asomers `minver` and `msrv` jobs now succeed with this change, would you please merge?

PR #2810, opened 2026-08-13
feat(mount): add `MS_NOSYMFOLLOW` to `MsFlags`

jzuber4:
## What does this PR do

Adds the `MS_NOSYMFOLLOW` flag to `MsFlags` in the `mount` module on Linux and
Android.

`MS_NOSYMFOLLOW` (bit `0x100`) was added to the kernel in Linux 5.10. When
passed to `mount(2)`, it forbids following symbolic links during path
resolution on the mounted filesystem. `nix` did not expose it, so callers that
need it currently have to reach for `unsafe { MsFlags::from_bits_unchecked(0x100) }`
and hard-code the raw bit themselves.

The constant is already provided by `libc` as `libc::MS_NOSYMFOLLOW` since
`0.2.186`, which is `nix`'s current minimum `libc` requirement, so this needs no
dependency bump. The flag is placed in kernel bit order, between `MS_DIRSYNC`
(`0x80`) and `MS_NOATIME` (`0x400`).

## Checklist:


jzuber4:
Heads up for reviewers: the red `minver` check here is **not** caused by this PR. It's a pre-existing breakage on `master` (the `cfg_aliases` 0.2.1 build-dependency floor no longer compiles under `-Zdirect-minimal-versions`), fixed in #2812. Once #2812 lands I'll rebase this PR onto the updated `master` and `minver` will go green.

PR #2811, opened 2026-08-13
fix(build): raise minimum `cfg_aliases` to 0.2.2

jzuber4:
## What does this PR do

Raises the minimum `cfg_aliases` build-dependency from 0.2.1 to 0.2.2.

nix's `build.rs` invokes the `cfg_aliases!` macro. Under
`cargo -Zdirect-minimal-versions`, cargo pins the declared floor (0.2.1),
whose macro expansion contains a trailing semicolon in expression position.
Current Rust promotes that from a warning to a hard error, so the build script
fails to compile and the `minver` CI job is red. (This is currently failing on
`master`, independent of any individual PR.)

`cfg_aliases` 0.2.2 restructured the macro to remove the trailing semicolon, so
raising the floor makes the minimum-versions build compile again. `cfg_aliases`
is a build-dependency only — it is not part of nix's public API — is edition
2018 with no dependencies and no declared MSRV, so this has no effect on nix's
consumers or on nix's MSRV.

## Checklist:

- [x] I have read `CONTRIBUTING.md`
- [x] I have written necessary tests and rustdoc comments
- [x] A change log has been added if this PR modifies nix's API

PR #2812, opened 2026-08-13
Fix minver and msrv CI: bump cfg_aliases floor and MSRV to 1.70

jzuber4:
## What does this PR do

Repairs the `minver` and `msrv` CI jobs, which are currently red on `master`.
These are two small, related build-hygiene fixes; they are in one PR because
neither can demonstrate green CI alone — `msrv` is gated on `minver`
(`needs: minver`), so both are required before the `msrv` job can run and pass.

**1. Raise the minimum `cfg_aliases` build-dependency to 0.2.2.**
`build.rs` invokes the `cfg_aliases!` macro. Under `cargo -Zdirect-minimal-versions`
cargo pins the declared floor (0.2.1), whose macro expansion contains a trailing
semicolon in expression position that current Rust rejects as a hard error, so
the build script fails to compile and `minver` is red. `cfg_aliases` 0.2.2
restructured the macro to remove it. `cfg_aliases` is a build-dependency only,
edition 2018, no dependencies, no MSRV bump — no effect on consumers.

**2. Raise the MSRV from 1.69 to 1.70.**
nix invokes `bitflags!` through its own `libc_bitflags!`. The current `bitflags`
2.13 release line emits macro code that triggers a name-resolver ICE in rustc
1.69.0 (`` `ModuleData::def_id` is called on a block module ``), fixed in rustc
1.70.0. Because nix commits no `Cargo.lock` and its resolver is not rust-version
aware, the `msrv` job resolves the newest `bitflags` and fails to build on 1.69.
1.70.0 is the lowest toolchain on which the current dependency set compiles
(verified locally: 1.69.0 ICEs, 1.70.0 builds clean).

Per the MSRV policy in the README, this implies the next release should be a
minor version bump; I've left the crate `version` for the maintainers to set at
release time.

Supersedes #2811 (which contained only the `cfg_aliases` fix; folding the MSRV
bump in lets CI go fully green).

## Checklist:


maxlapshin:
It would be nice to accept it. My PR is red because of the same problems.

xtqqczze:
I don't think we should bump the MSRV for this; it would be better if the `msrv` job used `direct-minimal-versions`.

asomers:
> I don't think we should bump the MSRV for this; it would be better if the `msrv` job used `direct-minimal-versions`.

Sort of true.  MSRV really just means that "some combination of dependencies will compile using this compiler".  But it's not necessarily the direct-minimal-versions.  For other projects, I've committed a "Cargo.lock.msrv" file just for this purpose.  But Rust 1.70.0 is already more than 3 years old.  I feel comfortable raising the MSRV that far.  It's less effort for maintainers than dealing with a Cargo.lock.msrv file.

changelog/2812.fixed.md
```diff
@@ -0,0 +1,2 @@
+Raised the minimum `cfg_aliases` build-dependency to 0.2.2, which is required to
+build nix's build script under `cargo -Zdirect-minimal-versions` on current Rust.
```
asomers: Not technically true.  It's required to build Nix with -Zdirect-minimal-versions while also denying all warnings.  Consumers won't generally do that.  I think you should remove this changelog entry, because it isn't really something that consumers need to care about.  It's really just to make CI work.

asomers (changes requested):

jzuber4:
Thanks Alan. I've updated the PR to remove the changelog. Note that the red CI appears to be related to rust-nightly rustfmt changes. Also the macos test appears flaky.

asomers (approved):

asomers:
The doc failures should be fixed by #2818 .

xtqqczze:
> `` `ModuleData::def_id` is called on a block module ``

This appears to be related to https://github.com/rust-lang/rust/issues/110848, where rustc 1.69.0 panics while resolving rustdoc links.

xtqqczze:
I can reproduce the `resolve_rustdoc_path` ICE locally with:

```sh
cargo update -p bitflags@2 --precise 2.12.1
```

but not with:

```sh
cargo update -p bitflags@2 --precise 2.12.0
```

The ICE first occurs when the `bitflags` package is patched with https://github.com/bitflags/bitflags/commit/e4ed6354c7f31522a42b84dc2fff97ea7c400282 (https://github.com/bitflags/bitflags/pull/487)

xtqqczze:
It looks like we we'll get a workaround in `bitflags` for the Rust 1.69 ICE, see https://github.com/bitflags/bitflags/pull/496.

xtqqczze:
@jzuber4 `bitflags` 2.13.2 was released with a fix for the Rust 1.69 ICE (https://github.com/bitflags/bitflags/issues/495). This PR can now be closed in favor of https://github.com/nix-rust/nix/pull/2808.

PR #2813, opened 2026-08-16
fix(socket): restore buffer capacities before reusing recvmmsg headers

maxlapshin:
##  Description

recvmmsg loses control messages when MultiHeaders are reused

## Expected

MultiHeaders is documented as a preallocated structure to be reused across recvmmsg calls, so every call should be able to receive as much control data as the buffer given to preallocate holds.

## Actual

only the first call can. msg_namelen and msg_controllen are in-out parameters of recvmsg(2): going in they are the capacity of the buffers, coming out the kernel replaces them with the number of bytes it actually wrote. recvmmsg refills only msg_iov/msg_iovlen between calls, so once a slot receives a datagram carrying no control message, its control capacity stays at zero forever. Every later datagram in that slot comes back with MSG_CTRUNC and cmsgs() returns ENOBUFS — and code that reads control data as an optional extra (.cmsgs().ok()) simply stops seeing it.

## Reproduction (added as test_recvmmsg_cmsgs_after_reuse)

bind a UDP socket, send and receive one datagram with ReceiveTimestampns off, then turn it on and receive a second one through the same MultiHeaders. The timestamp never arrives.

## Found in the wild

a WebRTC load generator that reads with UDP_GRO — the segment size arrives in a control message. After the first non-coalesced datagram the size stopped arriving, each coalesced batch was counted as a single datagram, and the measurement reported 72% packet loss that did not exist.

## Versions

reproduced on 0.29.0 and on master (0.31.3), Linux 6.8 and 7.0.

## Fix

remember the capacities in MultiHeaders at preallocate time and restore msg_namelen/msg_controllen (and clear msg_flags) in the recvmmsg loop. msg_namelen matters on the BSDs, where the input value bounds the address copy; on Linux the kernel resets it itself.

## Checklist:

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -1797,9 +1797,9 @@ pub fn sendmmsg<'a, XS, AS, C, I, S>(
         C: AsRef<[ControlMessage<'a>]> + 'a,
         S: SockaddrLike + 'a,
 {
+    data.restore_capacities();
 
     let mut count = 0;
-
 
     for (i, ((slice, addr), mmsghdr)) in slices.into_iter().zip(addrs.as_ref()).zip(data.items.iter_mut() ).enumerate() {
         let p = &mut mmsghdr.msg_hdr;
@@ -1860,7 +1860,10 @@ pub struct MultiHeaders<S> {
     // while we are not using it directly - this is used to store control messages
     // and we retain pointers to them inside items array
     _cmsg_buffers: Option<Box<[u8]>>,
+    // the capacities of the control and address buffers of every header,
+    // see `restore_capacities`
     msg_controllen: usize,
+    msg_namelen: libc::socklen_t,
 }
 
 #[cfg(any(linux_android, target_os = "freebsd", target_os = "netbsd"))]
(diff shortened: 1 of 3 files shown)


src/sys/socket/mod.rs
```diff
@@ -1861,6 +1861,9 @@ pub struct MultiHeaders<S> {
     // and we retain pointers to them inside items array
     _cmsg_buffers: Option<Box<[u8]>>,
     msg_controllen: usize,
+    // the capacity of every address buffer, needed to restore `msg_namelen`
+    // before reusing the headers, see `reset_for_receive`
```
pacak: What is `reset_for_receive`?

src/sys/socket/mod.rs
```diff
@@ -1943,11 +1947,22 @@ where
...
+    // any control message would be stuck with the shrunken value for good, and
+    // would silently drop what a later datagram carries.
+    let (msg_namelen, msg_controllen) = (data.msg_namelen, data.msg_controllen);
+
     let mut count = 0;
     for (i, (slice, mmsghdr)) in slices.into_iter().zip(data.items.iter_mut()).enumerate() {
         let p = &mut mmsghdr.msg_hdr;
         p.msg_iov = slice.as_mut().as_mut_ptr().cast();
         p.msg_iovlen = slice.as_mut().len() as _;
+        p.msg_namelen = msg_namelen;
```
pacak: Name length is restored, but there's no test for it. Doing something with unix sockets might do the trick.

pacak (commented):
I think sendmmsg suffers from the same problem - it can write control messages to return timestamps for sent messages.

src/sys/socket/mod.rs
maxlapshin: Added test_recvmmsg_address_after_reuse

src/sys/socket/mod.rs
maxlapshin: Changed the code

PR #2818, opened 2026-08-30
fix rustdoc lints

xtqqczze:
Stacked on https://github.com/nix-rust/nix/pull/2808.

### DIFF
--- a/src/sys/socket/mod.rs
+++ b/src/sys/socket/mod.rs
@@ -123,8 +123,7 @@ impl TryFrom<i32> for SockType {
     }
 }
 
-/// Constants used in [`socket`](fn.socket.html) and [`socketpair`](fn.socketpair.html)
-/// to specify the protocol to use.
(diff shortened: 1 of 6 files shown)


src/sys/socket/mod.rs
```diff
@@ -1776,9 +1775,6 @@ pub fn sendmsg<S>(fd: RawFd, iov: &[IoSlice<'_>], cmsgs: &[ControlMessage],
 ///
 /// # Returns
 /// `Vec` with numbers of sent bytes on each sent message.
-///
-/// # References
-/// [`sendmsg`](fn.sendmsg.html)
```
asomers, line 1781: This should probably be an http link to a man page.

asomers (changes requested):

src/sys/socket/mod.rs
xtqqczze: Yes, the public functions should have references to man pages etc. But that's an unrelated issue.

